PR #158, opened 2014-09-14
Add the ability to modify the message sent to individual players through AsyncPlayerChatEvent

blablubbabc:
per-player basis, support getting and modifying the recipients and still
setting the chat message globally for all recipients, if no specific
message is defined.


Zidane:
we'll be doing chat channels which covers the point of this PR.

Going to close this but thanks for bringing it to our attention :).

PR #181, opened 2014-09-19
Add CommandAuthorizationException

jacklin213:
The String permissionNode will need to be changed later on depending how the Permissions API is going to structure "permission nodes"

Deals with [SPONGE-44](http://issues.spongepowered.org/youtrack/issue/SPONGE-44)

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandAuthorizationException.java
+++ b/src/main/java/org/spongepowered/api/command/CommandAuthorizationException.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


jacklin213:
Sorry what should I do with this now?

modwizcode:
Here's what I was planning to do if you didn't have time/reply.

- [ ] Stash changes and do a reset to origin/master, then pop changes and move them into correct packages
- [ ] Leave the existing exceptions, because a ```CommandAuthorizationException``` is a ```CommandException```
- [ ] Handle CommandAuthorizationException with a default permission denied message of some sort. 
- [ ] (Optional) Add possible way to change default message by adding a parameter to command info definitions.

ryantheleach:
Should the default message be customizable to the user, or to the plugin?

modwizcode:
The plugin, if the plugin wants to let the user change the message that's
their choice.

jacklin213:
@modwizcode I'll give that a go now

modwizcode:
Ok thanks.

jacklin213:
@modwizcode Sorry I did a really bad job there, is that what u meant by default message?

modwizcode:
Haven't taken a look yet, mind explaining a bit more?

jacklin213:
@modwizcode Well when u said default message I thoughtt u meant to code the default message into the ```CommandAuthorizationException``` class. So I did it like so 
```private static String defaultMessage = "You do not have the permission to use this command!";```
Which then changed the constructor to
```
public CommandAuthorizationException(String permissionNode, CommandCallable command) {
    this(defaultMessage, permissionNode, command);
}
```

But I'm pretty sure thats not the way to do it, is it?

modwizcode:
Not quite. I'll be more specific in an hour or two.

jacklin213:
@modwizcode That would be nice thanks :D

modwizcode:
Sorry I took way longer than I thought. I meant to change the code in [SimpleCommandServer](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java) to have a second exception handler for CommandAuthorizationException that uses a default message there.

jacklin213:
@modwizcode  Rebased and did what I thought you told me to :D

modwizcode:
@jacklin213 Sorry I just looked at it again, I misread that logging warn as a message to the player. You shouldn't have to make any modifications to SimpleCommandServer.

modwizcode:
Reset SimpleCommandServer and then it should be ready to merge.

Zidane:
@modwizcode 

Finish this up and get it ready for merge please.

modwizcode:
Closed in favor of continuation on https://github.com/SpongePowered/SpongeAPI/pull/455

PR #220, opened 2014-11-20
Statistics/Achievements API

gratimax:
Edit by Deamon:

This PR should be finalized now.

To recap on the changes:

Added `Achievement`s, achievements may be fetched from the `GameRegistry` and then used in any of several methods added to `Player` in order to grant, revoke an achievement to the player. There is also an `AchievementEvent` added for when a player is granted an achievement.

Added Statistics, each `Statistic` has a `StatisticGroup` and an optional `StatisticFormat` override. Groups have default formats defined which are used for each statistic in the group. Statistics are obtainable from the `Statistics` pseudoenum or from `GameRegistry`, and may then be used in one of the new methods provided by `Player` in order to set, add to, or reset a statistic for that player. There is an event for statistics changing, this is no longer extended by statistics related events such as the `PlayerBreakBlockEvent` but exists on its own in order to be able to represent a single change and not have the requirement of multiples. New/Custom statistics may be constructed via the `StatisticBuilder` which is obtainable via the `GameRegistry`. Some statistics are associated with items, blocks, entities, or teams and may be fetched by giving the associated type and group to the proper method in `GameRegistry` and then used against a player.


src/main/java/org/spongepowered/api/stats/achievement/Achievements.java
```diff
@@ -0,0 +1,67 @@
...
+    public static final Achievement OVERKILL = null;
+    public static final Achievement BOOKCASE = null;
+    public static final Achievement EXPLORE_ALL_BIOMES = null;
+    public static final Achievement SPAWN_WITHER = null;
+    public static final Achievement KILL_WITHER = null;
+    public static final Achievement FULL_BEACON = null;
+    public static final Achievement BREED_COW = null;
+    public static final Achievement DIAMONDS_TO_YOU = null;
+    public static final Achievement OVERPOWERED = null;
+
```
AzureusNation: I think getValues() and getForName() should be in the GameRegistry as implementations cant change the methods In classes effectively.

ST-DDT: That's true, the comment is just too old.

ST-DDT:
@gratimax Did you had time to review my PRs yet? Or is @Minecrell responsible for that now?

Zidane:
@gratimax @gabizou @Minecrell 

I want some traction on this and @ST-DDT 's other two PRs...they need to get finished up and merged.

ST-DDT:
@Minecrell I updated my PR's, so they can be pulled/merged automatically.

ST-DDT:
@Minecrell or someone else from the team:
Can you please merge the latest commits from master in here?
There shouldn't be any merge conflicts.
Then i can start working on adding the API for registering new statistics.
(Without polluting my PR with a bunch of commits from master.)

Minecrell:
@ST-DDT Done!

ST-DDT:
@Minecrell Thanks

src/main/java/org/spongepowered/api/stats/StatisticBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+     * @return A new instance of a statistic
+     * @throws IllegalStateException If the statistic is not completed
+     */
+    Statistic buildAndRegister() throws IllegalStateException;
+
+    /**
+     * Represents a builder interface to create new instances of simple
+     * {@link Statistic}s (none grouped).
+     */
+    interface Simple extends StatisticBuilder {
```
ST-DDT: After removing `GroupedStatistic`s this subinterface is obsolete now.

src/main/java/org/spongepowered/api/stats/StatisticType.java
```diff
@@ -0,0 +1,41 @@
...
+ * {@link GroupedStatistic}s are belonging to.
+ */
+public interface StatisticType {
+
+    /**
+     * Gets the {@link StatisticUnit} this statistic is measured in.
+     *
+     * @return The statistic unit this statistic is measured in
+     */
+    StatisticUnit getStatisticUnit();
```
ST-DDT: `StatisticTypes.GENERAL` has no fixed `StatisticUnit`, thats a reason why i didn't add it in the first place.

Deamon5550: Can probably remove this and let it be set on a per-Statistic level (as it already is see [Statistic.java](https://github.com/SpongePowered/SpongeAPI/pull/220/files#diff-1210be3edd5a78c742a49de6c2b74d1cR46))

ST-DDT: Make it Optional, because its quite usefull if you want to print out an entire list of grouped statistics.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -398,6 +410,97 @@
...
+    /**
+     * Gets a collection of all available {@link Statistic}s.
+     *
+     * @return An immutable collection containing all statistics in registry
+     */
+    Collection<Statistic> getStatistics();
+
+    /**
+     * Create a new {@link StatisticBuilder} to create new simple
+     * {@link Statistic}s (none grouped).
```
ST-DDT: Should be 
`Create a new {@link StatisticBuilder} to create new {@link Statistic}s.`

The group stuff is no longer valid.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -398,6 +410,97 @@
...
+    Optional<TeamStatistic> getTeamStatistic(TeamStatisticType statisticType, TextColor teamColor);
+
+    /**
+     * Gets a list of all available {@link Statistic}s that are not grouped.
+     *
+     * @return An immutable collection containing all statistics in registry
+     *         that are not grouped
+     * @see Statistics
+     */
+    Collection<Statistic> getSimpleStatistics();
```
ST-DDT: This method is obsolete now. They can use `Collection<Statistic> getStatistics(StatisticType statisticType)` as an alternative.

src/main/java/org/spongepowered/api/stats/StatisticBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param unit The unit the statistic will be measured in
+     * @return This builder
+     */
+    StatisticBuilder unit(StatisticUnit unit);
+
+    /**
+     * Sets the {@link StatisticType} the {@link Statistic} belongs
+     * to.
+     *
+     * @param type The statistic type the grouped statistic belongs to
```
ST-DDT: Should be `@param type The statistic type the statistic belongs to`
(without group)
Because everything has a type now.

ST-DDT:
@Deamon5550 As an alternative to making it Optional, we can create more `StatisticType`s.

Like 
* `GENERIC_MOVED_DISTANCE` or just `MOVED_DISTANCE` with all elements that measure distances.
* `GENERIC_SPEND_TIME` or `SPEND_TIME`
* `GENERIC_DAMAGE` or `DAMAGE`
* `GENERIC_COUNT` or just `GENERIC`

Since we have `StatisticType`s everywhere, they should have a name attribute on them.

@all is someone working on the implementation already?
If not, i will try to implement it this weekend.

Deamon5550:
@ST-DDT `StatitisticType`s are not so much for logical grouping but rather physical grouping. I think perhaps it would be better to rename it to `StatisticGroup` as this is a bit more representative of what it actually is. And yes StatisticGroup/Type should have a name attached to it I agree.

ST-DDT:
@Deamon5550 Yeah you are right with `StatisticGroup` its defenitly better than `StatisticType`. Which was mostly a name chosen to be similar to `ItemType`and `EntityType`. (IIRC)

>`StatitisticType`s are not so much for logical grouping but rather physical grouping.

Yes, that true, but isn't `GENERIC` a logical grouping?
Maybe instead of making the `StatisticType.getUnit()` return `Optional` we should probably make `Statistic.getType()` return `Optional<? extends StatisticType>`. Using this generic approach make it work for subclasses as well.

Deamon5550:
@ST-DDT No GENERAL is a physical group<sup>1</sup> as well its got all the 'Times played', 'Worlds played' etc. stats in it. Also I already removed Item/Team/Entity StatisticTypes as they didn't really server any purpose other than some minor safety which we can enforce in other ways.

<sup>1</sup>[Statistics screen image](http://i52.tinypic.com/2h4jbdu.png) These are what I define as groups, although the item and block and entity groups are broken down a little further which happens to be rendered in the default client as columns within those umbrellas but that isn't something we need, or should, enforce.

Although you did just get me thinking, perhaps it would be better instead of having the default be optional the `Statistic#getUnit()` should be optional. and if it is absent then the group unit is used instead.

So to recap: Every statistic has a group, every group has a default unit. Statistics can optionally define an override type which will be used in preference to the default type.

ST-DDT:
> Although you did just get me thinking, perhaps it would be better instead of having the default be optional the `Statistic#getUnit()` should be optional. and if it is absent then the group unit is used instead.

Well, this feels kind of wrong.
1) Minecraft stores the unit for each stat (not really a reason i know)
2) If you have a random stat and you just want to display it, you add much unneeded complexity for showing the formatted value. `statistic.getUnit.getOrDefault(statistic.getGroup().getUnit().getOrDefault(???))`

> 1Statistics screen image These are what I define as groups, although the item and block and entity groups are broken down a little further which happens to be rendered in the default client as columns within those umbrellas but that isn't something we need, or should, enforce.

I created the groups without looking at the client at all.
I looked at the shared paths/id parts of the statistics. Thats another reason we i did not add them to the `GENERAL` group. From my point of view there is no general group, because the members of that group have nothing in common.

Deamon5550:
>`statistic.getUnit.getOrDefault(statistic.getGroup().getUnit().getOrDefault(???))`

It would be `statistic.getUnit().or(statistic.getGroup().getDefaultUnit());`. Not exactly overly complex, just allows a little more control on where the unit is from. If we were to have statistic.getUnit defer silently to the group unit then you would have no idea if it had an override set or not.
>From my point of view there is no general group, because the members of that group have nothing in common.

Well there is a general group, and it is the default group, and its members are united by their generalness.

But really, the benefits of having a general group far outweighs the downsides (What even are the downsides?). 

For example, it allows us to do things such as `player.getStatisticsByGroup(StatisticGroups.DEFAULT)` to find these general or default statistics rather than having to do:

```java
Map<Statistic, Long> stats = player.getStatistics();
for(Iterator<Statistic> it = stats.keySet().iterator(); it.hasNext(); )
{
    Statistic stat = it.next();
    if(stat.getGroup().isPresent())
    {
        it.remove();
    }
}
```

ST-DDT:
> But really, the benefits of having a general group far outweighs the downsides (What even are the downsides?). 

The downsides are those changes to the API that only exists in order to allow the creation of the default group and thus complicate the use of the real groups.

> For example, it allows us to do things such as `player.getStatisticsByGroup(StatisticGroups.DEFAULT)` to find these general or default statistics

Well `player.getStatisticsByGroup(null)` asuming `null` is no group would pretty much work as well.
(Yes, i'm one of those guys, who thinks that `null` actually means "none" or "nothing", and i also think that creating an instance of a class just to represent the lack of a meaningful instance of the class is a bad idea)

We could also use a constant for that if you feel more comfortable with them.
````java
public static final StatisticGroup DEFAULT = null; // yes, really null, not just not present yet
player.getStatisticsByGroup(StatisticGroups.DEFAULT):
````
or just `getUngroupedStatistics()`

Maybe we should ask for other people opinions on this topic. Because we both can not agree on one.

Let me try one last comparison/example.
We know for sure:
1. Every `Statistic` has a fixed unit.
2. Every `StatisticGroup` has a fixed unit for all `Statistic`s included in that group. (if we drop the `GENERAL` one)

Which usecase is more likely:
1. Showing a group's name for every `Statistic` in a list
2. Showing the statistics unit or a formatted value for every `Statistic` in a list

There are two things i cannot agree to:
* The `Optional`s for `getUnit()` in both `Statistic` and `StatisticGroup`.
* The dismatch of the statistic's unit and it's group's unit

There are basically 2 ways here i can accept:
1. Drop the `GENERAL` group, make `getGroup()` return `Optional.absent()` 
2. Add a `GENERAL` group per unit (because you will never show those groups members shuffled/mixed)

For me the current solution is not comfortable to use. And your explanations could not make up my mind on that matter. You are a `Collaborator` and i'm just a commiter so you have a stronger position, thats why i ask for other people's opinions on this topic.

**Comment by Deamon5550** [comment]
This PR should be finalized now.

To recap on the changes:

Added `Achievement`s, achievements may be fetched from the `GameRegistry` and then used in any of several methods added to `Player` in order to grant, revoke an achievement to the player. There is also an `AchievementEvent` added for when a player is granted an achievement.

Added Statistics, each `Statistic` has a `StatisticGroup` and an optional `StatisticFormat` override. Groups have default formats defined which are used for each statistic in the group. Statistics are obtainable from the `Statistics` pseudoenum or from `GameRegistry`, and may then be used in one of the new methods provided by `Player` in order to set, add to, or reset a statistic for that player. There is an event for statistics changing, this is no longer extended by statistics related events such as the `PlayerBreakBlockEvent` but exists on its own in order to be able to represent a single change and not have the requirement of multiples. New/Custom statistics may be constructed via the `StatisticBuilder` which is obtainable via the `GameRegistry`. Some statistics are associated with items, blocks, entities, or teams and may be fetched by giving the associated type and group to the proper method in `GameRegistry` and then used against a player.

File: src/main/java/org/spongepowered/api/stats/BlockStatistic.java [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.stats;
+
+import org.spongepowered.api.block.BlockType;
+
+/**
+ * Represents a {@link Statistic} for an {@link BlockType}.
```
simon816: *"for a BlockType"

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -401,6 +410,101 @@
...
+     * Gets the {@link Statistic} for the given {@link StatisticGroup} and
+     * {@link EntityType}. If the statistic group is not a valid
+     * {@link EntityStatistic} group then {@link Optional#absent()} will be
+     * returned.
+     *
+     * @param statisticGroup The type of statistic to return
+     * @param entityType The entity type for the statistic to return
+     * @return The entity statistic or Optional.absent() if not found
+     */
+    Optional<EntityStatistic> getEntityStatistic(StatisticGroup statisticGroup, EntityType entityType);
```
simon816: Could this be a method on `EntityType` instead?
This goes for `ItemType` and `BlockType` too.
I'm thinking it could be similar to the attributes API where these types are 'holders' of statistics.
Something like `EntityType.getStatistic(StatisticGroup)`
There could be an interface like:
```java
public interface StatisticHolder {
    Optional<? extends Statistic> getStatistic(StatisticGroup group);
}

**Comment by ST-DDT** [comment]
@Deamon5550 Apart from our disagreement adding a `HIDDEN` group may be useful as well, so you can check `stat.getGroup() == HIDDEN` and skip it. As you mentioned in your suggestion in the other PR.
What about `StatisticGroup` `DISTANCE_MOVED` there are quite a lot stats bundled in that topic, do you think its worth adding? `TIME_SPEND` and co has really few elements so we may let it stay in the general group from your point of view.

File: src/main/java/org/spongepowered/api/stats/achievement/Achievement.java [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+     * @return The parent of this achievement
+     */
+    Optional<Achievement> getParent();
+
+    /**
+     * Returns the children of this achievement.
+     *
+     * @return An immutable list of all children this achievement has
+     */
+    List<Achievement> getChildren();
```
ST-DDT: Why is this a list?
This should either be a Set or a Collection.
Because there is no usefull ordering in that.

ST-DDT:
Getters for Achievements are missing in the GameRegistry

src/main/java/org/spongepowered/api/stats/StatisticFormats.java
```diff
@@ -0,0 +1,55 @@
...
+ * A utility class for getting all available {@link StatisticFormat}s.
+ */
+public final class StatisticFormats {
+
+    /**
+     * A statistic without a format.
+     */
+    public static StatisticFormat COUNT = null;
+    /**
+     * A statistic measured in centimeters, meters, or kilometers depending on
```
ST-DDT: Maybe we should state that the input is in centimers, to allow custom replacements to format it properly without looking in any impl.

src/main/java/org/spongepowered/api/stats/StatisticFormats.java
```diff
@@ -0,0 +1,55 @@
...
+     * the magnitude.
+     */
+    public static StatisticFormat DISTANCE = null;
+    /**
+     * A statistic measured in 0.1 steps.
+     */
+    public static StatisticFormat FRACTIONAL = null;
+    /**
+     * A statistic measured in seconds, minutes, hours, or days depending on the
+     * magnitude.
```
ST-DDT: Maybe we should state that the input is in ticks, to allow custom replacements to format it properly without looking in any impl.

src/main/java/org/spongepowered/api/stats/Statistics.java
```diff
@@ -0,0 +1,86 @@
...
+    public static final Statistic NOTEBLOCK_PLAYED = null;
+    public static final Statistic NOTEBLOCK_TUNED = null;
+    public static final Statistic JUMP = null;
+    public static final Statistic JUNK_FISHED = null;
+    public static final Statistic LEAVE_GAME = null;
+    public static final Statistic MINECART_DISTANCE = null;
+    public static final Statistic MOB_KILLS = null;
+    public static final Statistic PIG_DISTANCE = null;
+    public static final Statistic PLAYER_KILLS = null;
+    public static final Statistic PLAY_ONE_MINUTE = null;
```
ST-DDT: This one should be renamed to `TIME_PLAYED`, because its measured in ticks and not in minutes.

src/main/java/org/spongepowered/api/stats/Statistics.java
```diff
@@ -0,0 +1,86 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.stats;
+
+/**
+ * A utility class for getting all available {@link Statistic}s that are not
+ * grouped.
```
ST-DDT: `that are not grouped` is no longer valid, because they are grouped in the general group.

src/main/java/org/spongepowered/api/stats/Statistics.java
```diff
@@ -0,0 +1,86 @@
...
+    public static final Statistic DROPPER_INSPECTED = null;
+    public static final Statistic ENDERCHEST_OPENED = null;
+    public static final Statistic FALL_DISTANCE = null;
+    public static final Statistic FISH_CAUGHT = null;
+    public static final Statistic FLOWER_POTTED = null;
+    public static final Statistic FURNACE_INTERACTION = null;
+    public static final Statistic FLY_DISTANCE = null;
+    public static final Statistic HOPPER_INSPECTED = null;
+    public static final Statistic HORSE_DISTANCE = null;
+    public static final Statistic ITEM_ENCHANTED = null;
```
ST-DDT: This should probably be `ITEMS_ENCHANTED`

src/main/java/org/spongepowered/api/stats/Statistics.java
```diff
@@ -0,0 +1,86 @@
...
+    public static final Statistic CHEST_OPENED = null;
+    public static final Statistic CLIMB_DISTANCE = null;
+    public static final Statistic CRAFTING_TABLE_INTERACTION = null;
+    public static final Statistic CROUCH_DISTANCE = null;
+    public static final Statistic DAMAGE_DEALT = null;
+    public static final Statistic DAMAGE_TAKEN = null;
+    public static final Statistic DEATHS = null;
+    public static final Statistic DISPENSER_INSPECTED = null;
+    public static final Statistic DIVE_DISTANCE = null;
+    public static final Statistic DROP = null;
```
ST-DDT: This should probably be `ITEMS_DROPPED` otherwise this may be missleading.

src/main/java/org/spongepowered/api/GameRegistry.java
Deamon5550: Not a fan of this to be honest, the EntityType is not  really a 'holder' of the statistic but rather just an object referenced by the group.

ST-DDT: I agree with Deamon5550 here.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -25,6 +25,8 @@
 
 package org.spongepowered.api;
 
+import org.spongepowered.api.stats.achievement.Achievement;
```
ST-DDT: Wrong import order

ST-DDT:
A Builder for ItemStatistic, EntityStatistic... is missing. When adding this feature please keep in mind that plugins may want create Statistic which supports custom types or multiple types.

ST-DDT:
@Deamon5550 I created a PR https://github.com/SpongePowered/SpongeAPI/pull/507 that addresses the builder issue.

ST-DDT:
@Deamon5550 You don't have to mark the small/last PR as mine. It just complicates the rebasing/merging process for you, doesn't it?.

To sum up the changes in https://github.com/SpongePowered/SpongeAPI/commit/de2a73cd30dac00c5ca39fa5e5f84e25badecd53:

Basically it will be impossible to track the `ZOMBIE` kills while being in `TEAM_BLUE` (with getter support)?
It will also be impossible to create a `ArenaPluginStatistic` with a reference to the arena it should count?
Is both intentional?

Why was the getId() removed?
I would like to use it for persisting it in my plugins config and transmitting it to my stats web page.
Or are we going to add some kind of persistence API support here?
The ids in the builder are most likely required to actually store the value in the stats file.
Or how should the server distinguish between plugin added achievements?
The translation id is probably not a good choice here.
Example:
getId() = `stats.times_chatted`
getTranslation().getId() = `stats.times_chatted.label`
getDescription().getId() = `stats.times_chatted.description` // Not used by MC yet, but quite useful for webpages as hover hint and (most) developers are lazy so they will try to use their existing i18n files everywhere.

Another feature that is not covered by this API yet, is to get the current progress on a specific achievement:
I have created an outline for myself already, but i wasn't sure how to represent the current/required data.
[not released AchievementProgress](https://gist.github.com/ST-DDT/746049bb1828e1d27726)
Or should this also be covered by some kind of Persistance API?

Deamon5550:
> You don't have to mark the small/last PR as mine.

I'll reorder it so that it gets squashed into your earlier PR when I squash this PR for merging.
>Basically it will be impossible to track the ZOMBIE kills while being in TEAM_BLUE (with getter support)?
It will also be impossible to create a ArenaPluginStatistic with a reference to the arena it should count?
Is both intentional?

No you're wrong on both counts, plugins can still create custom statistics simply by implementing statistics, and since they own the class they can simply construct it like normal, no need for our builders to support it, nor is there a need for registration since minecraft will take whatever stats you throw at it (though they won't display in the stats screen without a mod to help it along).
>The translation id is probably not a good choice here.

Alright yeah, but I'm going to rename it to getInternalName to avoid the overloading of the usage of getId.

>the current progress on a specific achievement:

No need for a custom type to handle it, achievements are backed by statistics and I see no need to not represent this as it would just be duplicating functionality. It also means that you can construct custom achievements easily based on existing statistics with limited work needed.

ST-DDT:
>> Basically it will be impossible to track the ZOMBIE kills while being in TEAM_BLUE (with getter support)?
    It will also be impossible to create a ArenaPluginStatistic with a reference to the arena it should count?
    Is both intentional?

> No you're wrong on both counts, plugins can still create custom statistics simply by implementing statistics, and since they own the class they can simply construct it like normal, no need for our builders to support it, nor is there a need for registration since minecraft will take whatever stats you throw at it (though they won't display in the stats screen without a mod to help it along).

But you can't retrieve the custom stats from the GameRegistry and all Stat entries that are not registered in minecraft are automatically dropped by minecraft when the player joins. Or did i miss a method to register a previously/self created `Statistic`?

`[00:39:11 WARN]: Invalid statistic in .\world\stats\uuid.json: Don't know what stat.OceansDrained is`

>> the current progress on a specific achievement:

> No need for a custom type to handle it, achievements are backed by statistics and I see no need to not represent this as it would just be duplicating functionality. It also means that you can construct custom achievements easily based on existing statistics with limited work needed.

How should a statistic backed by an long should represent something like a list of visited biomes?
In addition to that some achievements look entirely different than stats

````js
{
	"stat.flyOneCm" : 14954,
	"stat.jump" : 3,
	"stat.playOneMinute" : 26342,
	"stat.leaveGame" : 1,
	"achievement.exploreAllBiomes" : {
		"value" : 0,
		"progress" : ["Roofed Forest", "Extreme Hills+", "River", "Forest", "Plains", "Extreme Hills"]
	}
}
````

**Comment by Deamon5550** [comment]
>Stat entries that are not registered in minecraft are automatically dropped by minecraft when the player joins

Ah I missed that piece of validation, I'll add a register(Statistic) method to GameRegistry.

>How should a statistic backed by an long should represent something like a list of visited biomes?

I would view those achievements as the exceptions not the rules, after all the explore all biomes achievement is the only one in the vanilla game to use extra data and is handled as a special case. If you are wanting to create achievements with more specialized data then you're going to have to do the handling for granting it yourself, but since it wouldn't be based on an existing statistic you'd have to do this regardless so there is not really much difference.

>What is the difference between getId() and getInternalName()

Nothing I just wanted to avoid having two getId methods (one in the translation).

**Comment by ST-DDT** [comment]
Thanks for adding the method that fast.
Now that we have a method to register new stats are the builders still required? They could be replaced by base classes that plugins can extend easyly.

> Achievement extends Statistic because it always stores a long "value"

What about that? Checking how often a player has a specific could be quite useful.
Although this could be considered a impl detail.
But we could still change `Player.hasAchievement(Achievement)` to return `Optional<Long>`.

File: src/main/java/org/spongepowered/api/event/stats/AchievementEvent.java [hunk 17]
```diff
@@ -0,0 +1,46 @@
...
+package org.spongepowered.api.event.stats;
+
+import org.spongepowered.api.entity.player.Player;
+import org.spongepowered.api.event.entity.living.player.PlayerEvent;
+import org.spongepowered.api.stats.achievement.Achievement;
+import org.spongepowered.api.util.event.Cancellable;
+
+/**
+ * Represents an event that is called when a {@link Player} earns an
+ * {@link Achievement}.
```
ST-DDT: We should qualify whether the event is called every time a player earns this achievement or just for the first time. Or we should add a method with the current value.

src/main/java/org/spongepowered/api/event/stats/StatisticChangeEvent.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public interface StatisticChangeEvent extends PlayerEvent, Cancellable {
+
+    /**
+     * Gets the {@link Statistic} which is being modified by this event.
+     *
+     * @return The statistic being modified
+     */
+    Statistic getChangedStatistic();
+
```
ST-DDT: There should be a method to get the old value as well.

src/main/java/org/spongepowered/api/stats/achievement/AchievementBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public interface AchievementBuilder {
+
+    /**
+     * Sets the translation for the {@link Achievement}.
+     *
+     * @param translation The translation for the achievement
+     * @return This builder, for chaining
+     */
+    StatisticBuilder translation(Translation translation);
```
ST-DDT: Typo `StatisticBuilder` should be `AchievementBuilder`.

src/main/java/org/spongepowered/api/stats/achievement/AchievementBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a builder interface to create new and custom instances of
+ * {@link Achievement}s.
+ */
+public interface AchievementBuilder {
+
+    /**
```
ST-DDT: The setter for internal name is missing.

Deamon5550:
> are the builders still required?

The builders are still good for the more common use case of just wanting to create a generic statistic without needing to implement Statistic yourself to do it.
>Achievement could extend Statistic

The relationship of an achievement to a statistic is more of a 'has a' than an 'is a' relationship. Having it this way is ultimately more useful in my opinion as it allows everything you could do if extending statistic, but doesn't preclude having multiple achievements watching the same statistic.

ST-DDT:
> The builders are still good for the more common use case of just wanting to create a generic statistic without needing to implement Statistic yourself to do it.

If we replace the builders with a base implementation, then the user does not have to implement anything and he has something he can extend when creating his own implementation. But i guess its really rare that a plugin wants to create a fully independent statistic.
>> Achievement could extend Statistic

> The relationship of an achievement to a statistic is more of a 'has a' than an 'is a' relationship. Having it this way is ultimately more useful in my opinion as it allows everything you could do if extending statistic, but doesn't preclude having multiple achievements watching the same statistic.

The suggestion was targeting more to the ability to get the value how often a player has earned the achievement.

Zidane:

Can't merge.

@Deamon5550 Is this ready to go once you fix merging?

src/main/java/org/spongepowered/api/event/stats/AchievementEvent.java
Deamon5550: The event would be for every time that the player is *granted* the achievement. No need for a current value getter imo if you do need it you can always get it from the player.

ST-DDT: > The event would be for every time that the player is granted the achievement.

Granted for the first time? Granted another time? If we don't have a getter here we should mention in the javadocs that the event is only called if the achievement is granted to the player the **first** time/ if he did not have it before. Otherwise we might get inconsistencies with other servers implementing the sponge API/behaviour differently.

> No need for a current value getter imo if you do need it you can always get it from the player.

No, the `Player` does not have a getter for that too.

ST-DDT:
Has someone already started working on the implementation?

PR #242, opened 2014-11-27
Inventory API

gratimax:
An official, better Inventory API.

- [x] Inventory
  - [x] Block Inventory
  - [x] Entity Inventory
  - [x] Equipable Inventory
  - [x] Merchant Inventory
  - [x] Inventory factory
  - [x] Custom inventory
- [x] Recipes
  - [x] Recipe general
  - [x] Shaped recipes
  - [x] Shapeless recipes
  - [x] Ore dictionary
- [x] Javadocs

Note that due to forge mod tinkering we cannot have the nice methods and grids and such that I was proposing and from bukkit. This is on the roadmap though.


src/main/java/org/spongepowered/api/event/inventory/InventoryOpenEvent.java
```diff
@@ -0,0 +1,34 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.inventory;
+
+import org.spongepowered.api.util.event.Cancellable;
+
+/**
+ * Fired when a viewer opens an inventory.
+ */
+public interface InventoryOpenEvent extends ViewerEvent, Cancellable {
```
spaceemotion: Would this be cancellable even if its the normal PlayerInventory? My guess is "no", but would be interesting to know.

nightpool: @spaceemotion I don't know about the minecraft internals, but I'm *pretty* sure
bukkit plugins were able to do this

Aaron1011: I believe it is. You can send a [close window packet](http://wiki.vg/Protocol#Close_Window) with an id 0 0.

ryantheleach: >there is never an Open Window message for inventory.
From the wiki you just posted, so it would appear that you can't even tell that the player *opened* their inventory, let alone stop them.

evilmidget38: Players and inventories are kind of weird, especially in the case of Players and their own inventory.  With most inventories, the Player interacts with a Block and then the server sends them the open window packets.  With a Player's own inventory, however, they don't ask for server side validation before opening the Inventory.  The client stores the contents of its own inventory and is therefore able to open the inventory without needing any additional information from the server.  Up until a few minecraft versions ago, the client actually sent nothing to indicate that the Player had opened their inventory.  I'm not sure which version it was that changed it, but the client now sends a packet for the sake of achievements, letting the server know that it opened its inventory.  We experimented around a bit with this in Bukkit, but ultimately it didn't directly get anywhere(the functionality can be achieved via the achievements API)  The best that you can get is closing the Player's inventory right after they open it, which stands in stark contrast to all other inventories where it can simply be prevented.

Also, I should mention that my knowledge base extends only to Minecraft 1.7.  I know nothing about 1.8, and it could be entirely different(Wasn't Dinnerbone rewriting part of inventories?).

simon816: @evilmidget38 Looking at the Minecraft 1.8 code, it does exactly as you've described, so no changes there AFAIK.

So with that in mind, would it be acceptable to 'hijack' the inventory achievement and send a `S2EPacketCloseWindow` if cancelled?

evilmidget38: @simon816 It's certainly possible to do.  Graphically, it looks a little weird(their inventory opens then suddenly closes), so it's not exactly ideal, but it works.  The biggest issue I see, however, is the potential security issues that are created by naively closing the window.  When you close most inventories, you assume that the player(or really, the client) is unable to interact with the inventory.  Hacked clients, however, could still modify the inventory as no management of whether or not they have it open exists.

If you want to implement this "correctly" and in a manner that's consistent. you'd need to add in complete inventory tracking, presumably by interpreting the achievement packet as an openwindow packet and storing the state of whether they actually have their inventory open.

me4502:
Anyone heard from @Mumfrey on his progress?

Mumfrey:
I'm updating my PR intro document as we speak.

Mumfrey:
Closing in favour of #443

Issue #221, opened 2014-11-20
Ongoing Minor Issue List (OCD list)


phroa:
Track at https://github.com/SpongePowered/SpongeAPI/pull/860

__Minor Issues List__

- [ ] Discover Minor Issues
- [x] SpongeAPI should use quaternions for rotations (provided by flow-math) as they are superior to euler angles and the standard in games. Both options should be provided: what the implementation doesn't support becomes a simple overload (conversions representations are provided by flow-math).
- [ ] Have checkstyle warn on missing Override annotation
- [x] Secondary ordinal or Secondary cardinal? Javadoc and field name conflict. [`Direction:30`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/Direction.java#L30)
- [x] [`GameMode`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/player/gamemode/GameMode.java#L32) javadoc uses fully qualified link to `Player`
- [x] ~~Some method declarations in [`DataHolder`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/DataHolder.java) are generic although they don't have to. This leads to worse usability / more raw casts. See comment: https://github.com/SpongePowered/SpongeAPI/issues/221#issuecomment-107156044~~ appears to have been refactored so I won't touch
- [x] [GameMode](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/player/gamemode/GameMode.java) would be better suited in the [org.spongepowered.api.data.type](https://github.com/SpongePowered/SpongeAPI/tree/master/src/main/java/org/spongepowered/api/data/type) package.
- [x] [`AbstractInventoryProperty`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/item/inventory/property/AbstractInventoryProperty.java) could extend [`AbstractProperty`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/property/AbstractProperty.java). See comment: https://github.com/SpongePowered/SpongeAPI/issues/221#issuecomment-119275472 for more
- [ ] Make UseItemStackEvent.Finish not cancellable, don't allow modifying the duration in Stop and Finish (changes won't be reflected anyway)

__Checkstyle warnings__

File | Problem
---- | ----

[...]

PR #278, opened 2014-12-10
Fix myriad of minor issues.

gabizou:
Fixes https://github.com/SpongePowered/SpongeAPI/issues/221.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/player/PlayerInteractEvent.java
+++ b/src/main/java/org/spongepowered/api/event/player/PlayerInteractEvent.java
@@ -25,7 +25,6 @@
 
 package org.spongepowered.api.event.player;
 
-import com.google.common.base.Optional;
 import org.spongepowered.api.block.BlockLoc;
 import org.spongepowered.api.entity.EntityInteractionType;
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/api/event/player/PlayerInteractEvent.java
```diff
@@ -37,11 +36,11 @@
 public interface PlayerInteractEvent extends PlayerEvent, Cancellable {
 
     /**
-     * Gets the {@link BlockLoc} that the player has clicked, if available.
+     * Gets the {@link BlockLoc} that the player has clicked.
      * 
      * @return The block
      */
-    Optional<BlockLoc> getBlock();
+    BlockLoc getBlock();
```
Bammerbom: Why is this not optional?

gabizou: Please comment on things that are more up to date. This issue does not exist on [master](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/player/PlayerInteractEvent.java#L35).

PR #286, opened 2014-12-15
Added Packet API.

me4502:
Adds an API to deal with Packets.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/network/NetworkEvent.java
+++ b/src/main/java/org/spongepowered/api/event/network/NetworkEvent.java
@@ -1,4 +1,4 @@
-/**
+/*
  * This file is part of Sponge, licensed under the MIT License (MIT).
  *
  * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
--- a/src/main/java/org/spongepowered/api/event/network/NetworkEvent.java
+++ b/src/main/java/org/spongepowered/api/event/network/NetworkEvent.java
@@ -1,0 +1,42 @@
+/**
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.event.network;
+
+import org.spongepowered.api.event.GameEvent;
+import org.spongepowered.api.network.packet.Packet;
+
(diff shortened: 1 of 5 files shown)


Zidane:
I'm going to close this for the following reasons.

A. Adding PacketHandlers reeks of bad API. While people love direct access, this allows plugins to bypass API functions in-which their contract by the implementation strives to keep stability. Packet Handlers also promote API laziness, it is the job of the API to work with plugin developers to add the features they want. We need to work with our community on supporting the plugins they love to the best of our ability. In short, letting plugins handle Minecraft packets is NEVER acceptable.

B. Packets sent to the EventBus is always a bad idea. It would overwhelm the server in a hurry.

modwizcode:
Was there IRC discussion about this?

me4502:
There was a while ago, I agree with them that it is a bad idea to have packets go through the event buffer. One part i disagree with however, is packets being implementation dependant. If a packet is different on some implementation, the game won't recognise it.

kenzierocks:
@me4502 I think it's not so much the protocol differing as it is implementing it for different things, making assumptions about how it works.

modwizcode:
Unless there was more of a consensus on IRC, I think that closing this idea
off is just not the right choice. Some of the coolest features implemented
on Bukkit were done by implementing packet modifications on protocollib.
ProctocolLib cannot be implemented without a nms system (which sponge
doesn't haven given the separation of API and implementation).

If we can't do holograms, because we won't give a versatile API for making
packet changes, then that's going to turn a lot of people away when we have
an alternative.

kenzierocks:
@modwizcode There's just as much separation here as in Bukkit.

modwizcode:
Not at all, if you used Bukkit, sure technically you could run plugins on
Glowstone, but CraftBukkit and Spigot (which are for the most part
compatible with nms) were the only full implementations.  A developer could
write code that unwrapped native minecraft vanilla objects for direct
manipulation. Secure in the knowledge that users would have those classes.

Now I'm not arguing that that is correct, far from it. I'm arguing that
with an official Forge implementation, and a planned Glowstone
implementation, and a third party implementation done by the Granite team.
A plugin can't simply add a packet API that's stable.

By defining a packet API in the official API, we guarantee the maximum
reasonable capacity for plugins. While locking in the functionality so that
its handled safely for plugins.

While the driving force of mc servers is hub servers, and those servers may
stick with Spigot and their existing plugins with a possibly messy codebase
because of the full compatibility offered. The lack of a reasonable way to
implement packet based features is a serious design flaw. Holograms are
important to many servers, that fancy motd animation hack could be
important to servers. We should think about the audience we are writing
this for.

Zidane:
@modwizcode 

If an API requires you to use an implementation, that is a shitty API.

For those plugins, the point is to work with us to get API added. NOT to circumvent that and handle implementation aspects yourself.

modwizcode:
I concede I guess, but when you want to add a feature, sometimes it might
not be a valid inclusion in the API (i.e. animated motd bug, which was
denied from API inclusion by the fact that its a bug). Holograms are also
technically a bug. Some things/ just don't belong in the API, but still
should be able to work cross server implementation.

Do you disagree? Or do you think we can handle what are in essence low
level changes that can be enacted semi/fully transparently with a packet
filter?

Zidane:
@modwizcode Can you direct me to the denial by us in SpongeAPI with the motd?

Bug abuse can be evaluated for validity. Regardless, the 1% doesn't validate an API that literally opens "Pandora's box".

We can't help if people depend on Forge/Sponge, Granite, etc and mess around with the internals. Same as when people did it with CraftBukkit. What we can do is make the best effort in getting things added with a lot more flexibility than in Bukkit.

modwizcode:
#331 @minecrell's comment. I guess the only alternative is basically for a plugin that uses reflection to load the implementation specific classes at runtime, or has separate module that hook into the underlying implementation's with a different module for each implementation. Which may actually be a better system overall. I just don't want to see every little thing that needs an internal change either denied because it doesn't fit, or added and polluting the API with not relevant extras for a single, but popular feature.

Zidane:
@modwizcode 

...and minecrell makes a good point. Introducing changes like that-that are vast and can end up with Mojang fixing it and we are left with a choice of API break or reintroducing bugs aren't really acceptable. This falls into the 1% rule. Yet again this doesn't validate exposing the handlers for the packets.

modwizcode:
@Zidane look at #364, possibly a better way to handle this? It's not actually that far off from our ServiceProvider, except it has additional attributes replacing priority.

ryantheleach:
What about exposing a versioned packet API that breaks when updates are made by Mojang to the protocol?

It would work as long as the sponge project can make the assumption that the client will always be minecraft, that is to say, follows the same protocol as vanilla.

As long as it is known that the packet API is not the expected way to do most things, and that getting the required hooks in sponge should be the first priority, I could see it working for things that are bugs, or subject to Mojang changing.

People are always going to want to play with stuff that can potentially break between versions, as long as the unstable functions are documented as such, I don't see harm in including them? The alternative is that people will rely on implementations and fragment the community further.

modwizcode:
@ryantheleach A possible better solution is having a well designed, but separate plugin that handles packets. With the implementation specifics handled by something like #364.

modwizcode:
@me4502 can we delete this branch?

me4502:
No

PR #306, opened 2014-12-20
[WIP] Asynchronous Scheduler

sibomots:
As work in progress to phase in schedulers for
both asynchronous and synchronous tasks (depending on
the time base referenced). 

See source for implementation notes and pending issues.

Edit: In order to investigate the changes, it is straight-forward to write a simple plugin to initiate Tasks
through this Scheduler:

I have a skeleton Plugin that you can kick around and add to. https://github.com/sibomots/JiveMimic.git

I'll elaborate the plugin later with more use-cases, but the bottom line is that for "Asynchronous" Tasks, we can do that pretty easily and hopefully with the least amount of CPU overhead (the overall intent).

The Synchronous Scheduler on the other hand has more work left to do:

Since I assume that it should use the actual clock-tick is coming from the Minecraft server (via Tick) rather than the simple check of wall-clock time-elapsed there's a gap in the implementation of Sponge to capture a Tick event rather than poll the "runningTicks" in the Mixin.

I'm not in favor of polling, but rather in favor of waiting for events (like the Tick) to initiate the step() function of the SynchronousScheduler. Anyway, that's next on the agenda.


sibomots:
I would prefer to have this PR merged into the main branch for two reasons:

1.  Tasks with "names" is useless.  I don't see a use-case for it.
2.  The JavaDoc information has been updated to reflect that the time unit for both delay and interval are in the time unit of Ticks.
3.  There is no need for two kinds of Task interfaces.  (RepeatingTask and Task).  A Task that repeats has a non-zero period (interval).

PR #316, opened 2014-12-23
[WIP] Ensure API Completeness across the API

me4502:
Ensure API Completeness across the entire API.

The aim of this PR is to identify where we can expose more to plugin developers where we currently haven't.

Other PR's that achieve further completeness, that are seen as a part of this effort:
* [X] BlockType.java
* [X] ItemType.java
* [ ] Player Networks - #363 
* [x] Entity.java - #387
* [x] Item.java - #338 

Input from others is welcome, as well as any criticism about missing API.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -26,6 +26,8 @@
 package org.spongepowered.api.block;
 
 import org.spongepowered.api.service.persistence.data.DataHolder;
+import com.google.common.base.Optional;
+import org.spongepowered.api.item.ItemBlock;
 import org.spongepowered.api.text.translation.Translatable;
 
 /**
@@ -120,5 +122,12 @@ public interface BlockType extends Translatable {
      * @return The amount of light
      */
     float getEmittedLight();
+    
+    /**
+     * Gets the equivalent {@link ItemBlock} for this BlockType.
+     *
+     * @return The equivalent {@link ItemBlock}, if available
+     */
+    Optional<ItemBlock> getHeldItem();
 
 }
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -29,8 +29,11 @@ import com.flowpowered.math.vector.Vector2f;
 import com.google.common.base.Optional;
 import org.spongepowered.api.service.persistence.data.DataHolder;
 import org.spongepowered.api.util.Identifiable;
+import org.spongepowered.api.util.RelativePositions;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
+
+import java.util.EnumSet;
 
 import javax.annotation.Nullable;
 
@@ -74,6 +77,17 @@ public interface Entity extends Identifiable, EntityState, DataHolder {
      * @return True if the teleport was successful
      */
     boolean setLocation(Location location);
+
+    /**
+     * Moves the entity to the specified location, and sets the rotation. {@link RelativePositions}
+     * listed inside the EnumSet are considered relative.
+     *
+     * @param location The location to set
+     * @param rotation The rotation to set
+     * @param relativePositions The coordinates to set relatively
+     * @return True if the teleport was successful
+     */
+    boolean setLocationAndRotation(Location location, Vector2f rotation, EnumSet<RelativePositions> relativePositions);
 
     /**
      * Gets the rotation as a Vector2f.
--- a/src/main/java/org/spongepowered/api/item/EnchantmentSilkTouch.java
+++ b/src/main/java/org/spongepowered/api/item/EnchantmentSilkTouch.java
@@ -25,46 +25,33 @@
 
 package org.spongepowered.api.item;
 
-import org.spongepowered.api.item.inventory.ItemStack;
-import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.block.BlockType;
 
 /**
- * A type of item.
+ * Represents the Silk Touch enchantment, Enchantments.SILK_TOUCH.
  */
-public interface ItemType extends Translatable {
+public interface EnchantmentSilkTouch extends Enchantment {
 
     /**
-     * Gets the id of this item.
+     * Register a block that can be harvested with silk touch.
      *
-     * <p>Ex. Minecraft registers a golden carrot as
-     * "minecraft:golden_carrot".</p>
-     *
-     * @return The id
+     * @param block The block to register
      */
-    String getId();
+    void registerBlock(BlockType block);
 
     /**
-     * Get the default maximum quantity for {@link ItemStack}s
-     * of this item.
+     * Unregister a block so it can no longer be harvested with silk touch.
      *
-     * @return Max stack quantity
+     * @param block The block to unregister
      */
-    int getMaxStackQuantity();
+    void unregisterBlock(BlockType block);
 
     /**
-     * Gets if the item can be damaged.
+     * Gets whether the block can be harvested with silk touch.
      *
-     * <p>Ex. Tools, weapons and armor can all be damaged.</p>
-     *
-     * @return If it can be damaged
+     * @param block The block to test for
+     * @return Whether it can be harvested
      */
-    boolean isDamageable();
-
-    /**
-     * Returns the maximum damage an item can take.
-     * 
-     * @return The max damage
-     */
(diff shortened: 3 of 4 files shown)


src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -121,4 +122,18 @@
      */
     float getEmittedLight();
 
+    /**
+     * Gets if the block is able to be gathered using the silk touch
+     * enchantment.
+     *
+     * @return If the block can be harvested with silk touch
+     */
+    boolean canHarvestWithSilk();
```
simon816: My idea was to have `Enchantments.SILK_TOUCH` a special interface as defined below:
```java
public interface EnchantmentSilkTouch extends Enchantment {

    /**
     * Register a block that can be harvested with silk touch.
     *
     * @param block The block to register
     */
    void registerBlock(BlockType block);

    /**
     * Unregister a block so it can no longer be harvested with silk touch.
     *
     * @param block The block to unregister
     */
    void unregisterBlock(BlockType block);

    /**
     * Gets whether the block can be harvested with silk touch.
     *
     * @param block The block to test for
     * @return Whether it can be harvested
     */
    boolean canHarvest(BlockType block);
}
```
The vanilla blocks are added on start-up, but plugins now have the ability to modify what can and can't be harvested with silk touch.

For the implementation, `canSilkHarvest` would need to be modified to return `Enchantments.SILK_TOUCH.canHarvest(this)` which is straightforward.

src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -121,4 +122,18 @@
...
+
+    /**
+     * Gets the equivalent {@link ItemBlock} for this BlockType.
+     *
+     * @return The equivalent {@link ItemBlock}
+     */
+    ItemBlock getHeldItem();
```
SpaceManiac: Maybe this should be an `Optional`, some blocks don't have corresponding ItemBlocks.

kenzierocks: All blocks have ItemBlocks, some just use the default.

SpaceManiac: Blocks with no directly corresponding ItemBlock include: flowing_water, water, flowing_lava, lava, bed, piston_head, piston_extension, double_stone_slab, fire, redstone_wire, wheat, lit_redstone_ore, unlit_redstone_torch, reeds, portal, cake, unpowered_repeater, powered_repeater, pumpkin_stem, melon_stem, nether_wart, brewing_stand, cauldron, end_portal, lit_redstone_lamp, double_wooden_slab, cocoa, tripwire, flower_pot, carrots, potatoes, skull, unpowered_comparator, powered_comparator, standing_banner, wall_banner, daylight_detector_inverted, double_stone_slab2, spruce_door, birch_door, jungle_door, acacia_door, and dark_oak_door. Many of these have ItemTypes which associate with them, of course, but several simply do not have `ItemBlock` variants.

me4502: For a lot of those, there is an item equivalent.

Eg cakes have the held cake item, reeds have the held item, etc.

ryantheleach: Modded items?

kenzierocks: @SpaceManiac, have you ever used TMI? There are item blocks for all of those, if you can /give yourself it then it has an ItemBlock. That's the thing you're holding.

SpaceManiac: Sure, sure, a lot of them have item *versions* which aren't actually ItemBlocks. Water, lava, the piston pieces, double slabs, stems, lit redstone lamps, signs, banners etc. all have no corresponding item whatsoever.

me4502: As of 1.8 you can't have water blocks in your inventory, there is no inventory item for it.

kenzierocks: Really? That's new. That's versions then. Doesn't matter 'cause their under their corresponding item.

ryantheleach: Also as an example that has *no* mapping even with older versions of minecraft, block 36 has no itemblock counterpart as far as I am aware.
http://minecraft.gamepedia.com/Technical_blocks

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -76,6 +79,16 @@
     boolean setLocation(Location location);
 
     /**
+     * Moves the entity by the specified
```
me4502: Leaving a note here for me to fix this later.

me4502:
@Zidane You able to have a look over this and note any issues?

I'm aware of the half finished javadoc

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -76,6 +79,16 @@
...
+     *
+     * @param location The location to set
+     * @param rotation The rotation to set
+     * @param relativePositions The coordinates to set relatively
+     * @return True if the teleport was successful
+     */
+    boolean setLocationAndRotation(Location location, Vector2f rotation, EnumSet<RelativePositions> relativePositions);
```
Zidane: Explain the enumset here.

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -118,4 +119,10 @@
      */
     void clearTitle();
 
+    /**
+     * Gets the {@link InetSocketAddress} from which this player is connecting from.
+     *
+     * @return The player's address
+     */
+    InetSocketAddress getAddress();
```
Zidane: Remove from API, @gabizou has this in his branch for player-net

src/main/java/org/spongepowered/api/entity/Entity.java
me4502: @Kobata Would be best to explain the reasoning.

src/main/java/org/spongepowered/api/entity/player/Player.java
me4502: Ah, kk - I'll do that

src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -121,4 +123,18 @@
...
+     * @return If the block can be harvested with silk touch
+     */
+    boolean canHarvestWithSilk();
+
+    /**
+     * Gets the equivalent {@link ItemBlock} for this BlockType.
+     *
+     * @return The equivalent {@link ItemBlock}, if available
+     */
+    Optional<ItemBlock> getHeldItem();
```
Zidane: Interesting.

Very VERY few blocks in the base game have no ItemBlock. BlockAir is one off the top of my head that does not. Just an observation.

src/main/java/org/spongepowered/api/block/BlockType.java
Zidane: I think I like simon's idea better. This should be isolated to the SilkTouch enchantment.

src/main/java/org/spongepowered/api/block/BlockType.java
me4502: I think it's safest to just use Optional

src/main/java/org/spongepowered/api/entity/Entity.java
Kobata: Network protocol on the teleport allows setting relative mode for each component of location/pitch/yaw separately.

E.g., /teleport 0 ~ ~3 sends you to (0, y, z+3), and would be Location(0, 0, 3), Rotation(0, 0), EnumSet.of(RelativePositions.Y, RelativePositions.Z, RelativePositions.PITCH, RelativePositions.YAW)

src/main/java/org/spongepowered/api/block/BlockType.java
Deamon5550: @Zidane As of 1.8 quite a few more blocks have no corresponding ItemBlocks, things like fluids, plants, fire, piston parts, beds, etc.

src/main/java/org/spongepowered/api/block/BlockType.java
me4502: Okay, done.

me4502:
This PR has been merged.

PR #322, opened 2014-12-25
Extent Enhancement

Soren025:
This pull request is to make Extent into an object intended to redefine origin relative to a parent Extent up until a root World and make locations represent the local coordinates of the extent it contains.

This would be useful for any region based system, allowing them to reference local coordinates. An example would be an arena object implementing extent, possibly with sub regions for the arena parenting off of the arena. Creating a copy of this arena layout would be as simple as changing the origin of the new copy.

Personally I don't expect this to get merged as I know there are some issues with my way of handling it, however I did this to bring attention to the possibility of what can be done with Extent. This is just something I did to show how it could be done.


gratimax:
I personally don't like the idea of extent "parenting". I would instead consider trying to have an API for relative cartesian coordinate systems. In addition, it might also be cool to get [affine transformations](http://en.wikipedia.org/wiki/Affine_transformation) and the like from linear algebra(e.g. a chunk's coordinate system is simply a translation from the world's coordinate system).

Soren025:
@gratimax Thinking about it I came up with an idea, perhaps have an interface called ```Transform``` that would function similarly to the Transform component in unity. Location would act on Transform and Extent and Entity would both extend Transform as they both have locations to give.

Transform would have another Transform as a parent, and all transforms reference a world in some way. A transform that does not have a parent defaults to that world for calculations.

Extent in this case would still have a ```getParent()``` method in this case  however it would be implemented in both world and chunk to return null to act on the world in question, the purpose would be for Custom transforms to parent off of.

Entity transforms would parent off of the world they are in, a use case for entity being a Transform; say you have a player with a skill that causes explosions around him in a circle, just reference the coordinates local to the player and pass them through the transform to get the world coordinates and your done.

Also I am not sure if this is going to be a thing in the future but back in bukkit ```Location``` was also used to store pitch and yaw to act as a look direction. Transform could possibly also hold the methods for that as well and it would make sense, also giving ```Location``` the look direction as well.

sk89q:
I added Extent with heavy inspiration from WorldEdit, so you can find a lot of examples of Extent's usage in WorldEdit.

You don't actually need to know parents or origin if you implement a transform *as an Extent*.

You can find an example of a transforming extent in WorldEdit:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent/transform/BlockTransformExtent.java

Parent extents are heavily used:
https://github.com/sk89q/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java#L182

Here's a bunch of `Extent` implementations, most of which are proxies:
https://github.com/sk89q/WorldEdit/tree/master/worldedit-core/src/main/java/com/sk89q/worldedit/extent

However, you cannot get a copy of the parent extent in WorldEdit because I don't see the point. There's no reason why the final extent is even a world.

For this PR:

1. I don't think you need / should be able to get the parent.
2. Nor do I think you need to be able to get the origin position.
3. The changes only provide a very simple translation transformation as well.

On a semi-related node:

Please note that we are considering (or actually have for a while, but no one has gotten around to it), changing `BlockLoc` to a proxy object *in the API* and then copying most of its methods to `Extent` (which would make the API more like Minecraft, with `world.setBlock(coords, block)`).

Zidane:
@sk89q 

Can you work with Soren and either get this merged with changes or closed?

Soren025:
As I said above I made without expecting it to get merged. I have ideas but lack the patience to flesh them out to be piratical, thus ask questions and request input from others to make an idea work.

@sk89q @zidane That being said you can go ahead and close it if you want. I really only made this PR as an example of an idea I wish to see in the api.

Soren025:
@sk89q 
> Please note that we are considering (or actually have for a while, but no one has gotten around to it), changing BlockLoc to a proxy object in the API and then copying most of its methods to Extent (which would make the API more like Minecraft, with world.setBlock(coords, block)).

I suddenly wanted to ask for clarity sake for the concept you mentioned. I get the whole moving the block editing to extent, I kinda like that actually. What I am curious about is will BlockLoc become the same as Location but just with a Vector3i? I was recently thinking about it and that't the conclusion I came up with base on what you said.

Zidane:
Can't merge this.

Do we want this @sk89q still?

Soren025:
I don't think so

gratimax:
Closing for now. The most promising alternative is alternate coordinate systems

PR #324, opened 2014-12-26
Added BlockRedstoneUpdateEvent.

me4502:
Adds a redstone event that extends BlockUpdateEvent.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/block/BlockRedstoneUpdateEvent.java
+++ b/src/main/java/org/spongepowered/api/event/block/BlockRedstoneUpdateEvent.java
@@ -1,0 +1,8 @@
+package org.spongepowered.api.event.block;
+
+/**
+ * Created by Matthew on 26/12/2014.
+ */
+public interface BlockRedstoneUpdateEvent {
+
+}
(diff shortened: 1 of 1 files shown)


me4502:
Making this cancellable may cause issues internally - the game does not respond kindly to these kinds of updates not occuring.

modwizcode:
Pretty sure Bukkit let you cancel, which, while justifying an API choice by
saying "Bukkit did it" isn't a real strong support, I think that this is a
case that plugins actively used.

By providing this mechanic would make some plugins completely impossible.

Going to ask @SpongePowered/developers for comment here as I feel we should
decide this as a team.

me4502:
From what I can tell, Bukkit didn't provide this functionality.

modwizcode:
Hmm. Well if it didn't let you cancel, it at least gave you previous state,
new state (strength) and let you set both actually. Which is the
functionality I meant by cancelable.

me4502:
This allows setting new - setting old is quite redundant.

modwizcode:
Wasn't quite sure but after thinking about it some more I realized that's
correct. Setting old is undesirable as well as unnecessary. I missed the
method to set a new value somehow.

Given that, barring any formatting and style concerns I can't check now,
this seems good.

me4502:
The style all checks out.

modwizcode:
Ok will merge

me4502:
I'll do it, already got it locally

modwizcode:
Yeah sorry, I normally work with non-core PRs

me4502:
Yeah, all fine

PR #340, opened 2014-12-29
Implementation of a context-based permissions API

zml2008:
This API provides a lot of functionality for permissions providers,
including transient permissions, various contexts, and a basic
inheritance structure. This API is meant to serve as a flexible base to
build on.

There may be some utility classes added as I get to implementing this API in PEX and in Sponge (the default MC permissions system), but this PR has the API as I think is good

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
+++ b/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
@@ -1,0 +1,210 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.permission;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import org.spongepowered.api.service.permission.context.Context;
+import org.spongepowered.api.util.Tristate;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A subject data implementation storing all contained data in memory.
+ *
+ * <p>This class is thread-safe.
+ */
+public class MemorySubjectData implements SubjectData {
+    private final PermissionService service;
+    private final ConcurrentMap<Set<Context>, NodeTree> permissions = Maps.newConcurrentMap();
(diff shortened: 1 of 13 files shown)


zml2008:
Brief functionality update: The concept of a subject collection requiring parent subjects to be from a certain collection has been removed. Now all methods dealing with parents take/return Subject objects rather than Strings.

On a semi-related note: any thoughts on how best to handle error conditions? A boolean true/false return value doesn't seem like it provides enough information, while declaring an exception thrown feels like too much work for devs dealing with the API.

and note-to-self: add to package-info or Subject(Collection)? javadocs: reminder that all subjects should be managed by the permission service -- and how to integrate this with making Players subjects? unwrap() method?

Zidane:
@zml2008 

How close are we to a merge?

zml2008:
API should be good to merge now

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -0,0 +1,205 @@
...
+    private final ConcurrentMap<Set<Context>, NodeTree> permissions = new ConcurrentHashMap<Set<Context>, NodeTree>();
+    private final ConcurrentMap<Set<Context>, List<Map.Entry<String, String>>> parents = new ConcurrentHashMap<Set<Context>, List<Map.Entry<String, String>>>();
+
+    /**
+     * Creates a new subject data instance, using the provided service to request instances of permission subjects.
+     *
+     * @param service The service to request subjects from
+     */
+    public MemorySubjectData(PermissionService service) {
+        this.service = service;
```
Minecrell: There should be a null check here if it doesn't accept null values

src/test/java/org/spongepowered/api/service/permission/NodeTreeTest.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+package org.spongepowered.api.service.permission;
+
+import org.junit.Test;
+import org.spongepowered.api.util.Tristate;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
```
Minecrell: Static imports are not used in Google's Java code style: https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.1-wildcard-imports

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * A subject data implementation storing all contained data in memory.
+ *
+ * <p>This class is thread-safe.
+ */
+public class MemorySubjectData implements SubjectData {
+    private final PermissionService service;
+    private final ConcurrentMap<Set<Context>, NodeTree> permissions = new ConcurrentHashMap<Set<Context>, NodeTree>();
+    private final ConcurrentMap<Set<Context>, List<Map.Entry<String, String>>> parents = new ConcurrentHashMap<Set<Context>, List<Map.Entry<String, String>>>();
```
Minecrell: This line is too long, should be max. 150. You can use `Maps.newConcurrentMap()` as a shorter way to create the map.

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -0,0 +1,205 @@
...
+    public List<Subject> getParents(Set<Context> contexts) {
+        List<Map.Entry<String, String>> ret = parents.get(contexts);
+        return ret == null ? Collections.<Subject>emptyList() : toSubjectList(ret);
+    }
+
+    @Override
+    public boolean addParent(Set<Context> contexts, Subject parent) {
+        contexts = ImmutableSet.copyOf(contexts);
+        while (true) {
+            Map.Entry<String, String> newEnt = new AbstractMap.SimpleEntry<String, String>(parent.getContainingCollection().getIdentifier(), parent.getIdentifier());
```
Minecrell: Same here, this line is too long

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public boolean removeParent(Set<Context> contexts, Subject parent) {
+        contexts = ImmutableSet.copyOf(contexts);
+        while (true) {
+            Map.Entry<String, String> removeEnt = new AbstractMap.SimpleEntry<String, String>(parent.getContainingCollection().getIdentifier(), parent.getIdentifier());
```
Minecrell: Like above, line too long

src/main/java/org/spongepowered/api/service/permission/NodeTree.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * An immutable tree structure for determining node data. Any changes will create new copies of the necessary tree objects.
+ * Keys are case-insensitive.
+ * Segments of nodes are split by the '.' character
+ */
+public class NodeTree {
+    private static final Pattern SPLIT_REGEX = Pattern.compile("\\.");
```
Minecrell: You could just use `Pattern.compile(".", Pattern.LITERAL)` instead of escaping here

src/test/java/org/spongepowered/api/service/permission/NodeTreeTest.java
zml2008: Pretty sure everybody ever uses wildcard imports for junit Assert.

Minecrell: Well, Google Java style doesn't use any wildcard imports and checkstyle complains about this. If we want to use that we need to add an exception to the contribution guidelines as a difference to the Google style and fix checkstyle from complaining about that (in the minor branch).

simon816: You could use `*` imports when developing, in fact that is sometimes quite convenient, but before committing they should be expanded out.

zml2008: Ok, looks like the formatting guidelines are changing because that's stupid.

PR #350, opened 2014-12-30
Moved getters from Sounds and Rotations to GameRegistry.

AzureusNation:
This has been done as every other Class like this e.g. Blocks, Items, Environments, Biomes do not have getters. The getters for these should be in GameRegistry as they have to be registered like the above examples.


ST-DDT:
There had been a commit that renamed Sound to SoundType. https://github.com/SpongePowered/SpongeAPI/commit/fbdf9e59639b5b20fc2204c166d628b63863a0bf

Issue #237, opened 2014-11-26
ID Constants


Soren025:
This is not much of a concern but more of a convenience measure for those who use java 7 and above. Currently there is no way to switch case ids for blocks and items against a constant due to the lack of use of enums. However in java 7 switch casing strings is possible and final String constants  can be used as the case values.

Further more if just feels right to have access to the vanilla id values as String constants without having to go through the BlockType or ItemType constants located in BlockTypes and ItemTypes.

Something along the lines of a BlockIds and an ItemIds class.

PR #356, opened 2015-01-02
Change sounds to strings

Voltasalt:
In Minecraft, sounds are plain String IDs. Mirroring that in Sponge would make it compatible with resource packs that add custom sounds.


modwizcode:
I support this, because we should have had strings in the first place. If we want to make a big sound list then we can, but it should be a big list of Strings.

However some people might think other solutions are nicer. So I'm asking for input here

kenzierocks:
Can we have a list of all of the vanilla sounds?

TBotV63:
Get listed, sounds!!!
http://minecraft.gamepedia.com/Sounds.json#Sound_events

spaceemotion:
Why did you remove the SoundTypes then? You could have just renamed it to "Sounds" and changed their type to String.

kenzierocks:
@spaceemotion We're discussing this in IRC right now, `String` constants may be inlined by the compiler, preventing insertion at runtime.

gabizou:
I don't see why there should be a change from `SoundType` to `String`. Granted a sound is a string, but why is it necessary to represent a sound as a string?

modwizcode:
Because there's no point in making random wrapper objects for things that
only need to be strings. Furthermore, we need to allow arbitrary strings to
be used, since servers can play custom sounds for clients with the right
resource pack.

ryantheleach:
Smelt like a leaking implementation until I realized that resource packs / minecraft commands pretty much make this almost locked in stone. What would they change it to anyway?

modwizcode:
Right. This is also a protocol, packet based thing, they are sent as strings in the sound play packet. As strings are both more efficient if we need to send them anyway, and more complaint with the existing /playsound command's arguments. Strings are clearly better.

Soren025:
I hate to bring up an old topic, but https://github.com/SpongePowered/SpongeAPI/issues/237...

I had made an issue to request ID string constants for entities and blocks and got so much negativity, how is making string constants for Sounds any different yet there seems to be support for it.

kenzierocks:
Blocks and Entities are more than just Strings, while Sounds are just Strings.

Soren025:
@kenzierocks I merely requested string constants for the ids, not replace the EntityType and BlockType.

kenzierocks:
But if you have those classes, why add more constants for your constants?

Soren025:
@kenzierocks because in java 7, you can switch case strings using string constants for the case statements. Currently there is no way to switch case Block and Entity ids without knowing the id values or making the constants your self.

I am all for Sounds having string constants because of this, just a little disappointed I wont see the same for entities and blocks.

modwizcode:
As we have said before, switch statements, while very nice, are not the
only way to accomplish this. Before Java 7 you couldn't even do switches on
strings. Which are actually implemented as if statements internally.

As SpongeAPI and Sponge are targeting the Java version of Minecraft (1.6)
we're not designing our API based on something nice that is available in a
later Java version.

Additionally, sounds just get played, they require a location and a string
and then their done from the server's point of view. With entities the
entity has a class representation on the server side. We can take advantage
of Java's Generic type safety and prevent down casting if we use Entity
classes to spawn entities rather than string ids.

gabizou:
After discussing for a bit with some fellow developers, Sounds really don't make sense to keep as a wrapper for a String. As far as the server is concerned, it does not know what sounds are available on the Client, so, sounds are simply represented as Strings. The server simply doesn't care whether the client is sent a sound that doesn't exist or not, all it cares is that it sends a packet to play the desired sound attached to the string. 

TLDR; The server shouldn't care about what sound is being played let alone allowing plugins to send custom strings for the sounds they may want to have played, the client easily handles the case when it simply doesn't have the desired resource.

spaceemotion:
Although I don't dislike the decision, why remove the "vanilla" sounds completely then? If one would like to know which sounds are there in the default client, this would be the way to go. Otherwise I can already see "common" libraries adding helper classes for exactly that.

gabizou:
> If one would like to know which sounds are there in the default client, this would be the way to go.

I would agree that the default sounds shouldn't be removed from the API, since there is a subset of sounds by default that all vanilla clients should have.

Bammerbom:
I would like both a way for Strings and SoundTypes

Minecrell:
In general, I agree that sounds are best represented as strings and a wrapper isn't really necessary. However, we should still have the Vanilla sounds listed in our API for both usability and compatibility reasons. A plugin developer should have a list available with all sounds he can (normally) expect to work on most clients.

I wouldn't be too sure that sound IDs can't change in the future, and even if it's just a small minor change in the ID, all plugins using it would no longer be able to play the sound on the client. While this could be basically prevented using a list of constants in the SpongeAPI, the usage of strings for that is a little bit difficult. String constants will get inlined, so this method would be actually compiled like the line below:

```java
public static final String TEST = "sound.test";
player.playSound(Sounds.TEST);
// Decompiled again
player.playSound("sound.test");
```

So using String constants in the API would only make usability easier, but wouldn't prevent problems with ID changes. Of course there are way to prevent the compiler from doing that (like below), but I'm not sure if it wouldn't be easier to just use `SoundType` as a simple wrapper with an additional `getById()` method in `GameRegistry` for example.

```java
public static final String TEST = "sound.test".intern();
player.playSound(Sounds.TEST);
// Decompiled again
player.playSound(Sounds.TEST);
```

```java
player.playSound(Sounds.getById("sound.custom.test"));
```
This would be also more future-proof if sounds might have more properties in the future, not just the sound ID.

modwizcode:
There are easier solutions than new String(...).intern(). But objects are more future-proof.

Minecrell:
@modwizcode `"sound.test".intern()` would be the most efficient way, because it doesn't create a new String.

modwizcode:
It does, but not every time. If you use regular constants in a static {} block, it's identical in functionality because "...".intern() == "..."

modwizcode:
But I agree objects are more appropriate, just wanted to point out that there are other options.

AzureusNation:
I personally think that changing it to a String is a bit dramatic but aslong as the Fields in the Sponge class are near enough the same as the strings e.g. AMBIENCE_CAVE to be AMBIENCE_CAVE_CAVE just so it is easier to register the names.

ST-DDT:
I'm really unsure about this topic as well. But IMO objects are better for two reasons.
First they wont get inlined accidently, and if a sound gets deleted entirely the server implementation can either place a missing/dummy sound there or play an appropriate alternative.
Currently a sound is just a single sound with some random versions. But server implementations may also contain more random sounds that are not available in the client at all such as `PLAY_RANDOM_SOUND` when the server gets this it send one of the registered sounds to the client or a SoundList which plays a short melody using multiple SoundPackets. The object like API is simply more extendible for the future.
In addition to that the servers may use a numeric sound id or an enum that will be passed through the methods and will be resolved only in the package handler using a map which would make it difficult to use it for events safely. 

However it should be easyly possible to `Sound getSoundById(String)` as mentioned above although i'm not sure whether this method should register any requested sound automatically or if we use an extra method for that called `Sound registerSoundId(String)`. Is the sound feature async?

TBotV63:
@Minecrell @modwizcode As I said in IRC, intern() is a bad idea. In Java 6 interned strings are stored in PermGen and we risk an OutOfMemoryException. It's best to not use String interning before Java 8 at all and is considered bad design. Also, @ST-DDT did mention some good points that speak for an Object wrapper.

kenzierocks:
@TBotV63 All string constants are interned already. There is no extra PermGen consumption.

http://theopentutorials.com/tutorials/java/strings/string-literal-pool/

TBotV63:
Dammit. I didn't think about that... Still i think intern() is not the ay to go, just cause of the lookz xD

AzureusNation:
please also look at pr #366 as it is related.

Zidane:
Cannot merge this, has conflicts.

Minecrell:
Closing this as it removes the complete sound listing and creating it again would make the compiler inline the String constants. We want to ensure compatibility with future Minecraft versions and **if an ID will ever change** (and they have in the past) plugins will break so this doesn't really improve our API.

PR #357, opened 2015-01-02
Add getPing() method.

Bammerbom:


src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -117,5 +117,12 @@
      * Removes the currently displayed {@link Title} from the player's screen.
      */
     void clearTitle();
+    
+    /**
+     * Get a player's ping.
+     * 
+     * @return The player's ping
+     * /
```
Minecrell: You need to close the Javadoc comment here: https://travis-ci.org/SpongePowered/SpongeAPI/builds/45701022#L217-L220
> /home/travis/build/SpongePowered/SpongeAPI/src/main/java/org/spongepowered/api/entity/playe/Player.java:121: unclosed comment

Bammerbom: Oops, that space was a typo, fixing it now.

gabizou:
I would think that this would belong in some sort of Packet/Connection API of which we could represent the player connection (which would make sense for Player to include).

Zidane:
This is a part of @gabizou 's player-net branch

Thanks anyways!

Issue #360, opened 2015-01-03
Removing Chunk.


Azatik:
Idea. Add function:
removing Chunk. As if he did not generate. Alternative RegenerateChunk.

Bammerbom:
https://github.com/SpongePowered/SpongeAPI/pull/361/files

Deamon5550:
This already exists in World:

```java
    /**
     * Deletes the given chunk from the world. Returns a {@code boolean}
     * flag for whether the operation was successful.
     * 
     * @param chunk The chunk to delete
     * @return Whether the operation was successful
     */
    boolean deleteChunk(Chunk chunk);
```

@Bammerbom in regards to having a delete method in Chunk, this is just a recipe for memory leaks since as soon as you call Chunk.delete() the reference to the chunk that you have is invalid and needs to be dereferenced.

PR #361, opened 2015-01-03
Add isGenerated, isPopulated, remove, and regenerate to Chunk.

Bammerbom:
https://github.com/SpongePowered/SpongeAPI/issues/360

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Chunk.java
+++ b/src/main/java/org/spongepowered/api/world/Chunk.java
@@ -82,14 +82,14 @@ public interface Chunk extends Extent {
     /**
      * Removes the chunk.
      *
-     * @return Whether or not the chunk is removed.
+     * @return If the chunk was successfully removed
      */
     boolean remove();
     
     /**
      * Reegenerates the chunk.
      *
-     * @return Whether or not the chunk is regenerated
+     * @return If the chunk was successfully regenerated
      */
     boolean regenerate();
 }
--- a/src/main/java/org/spongepowered/api/world/Chunk.java
+++ b/src/main/java/org/spongepowered/api/world/Chunk.java
@@ -54,6 +54,13 @@ public interface Chunk extends Extent {
      * @return Whether or not this chunk is loaded
      */
     boolean isLoaded();
+    
+    /**
+     * Gets whether or not this chunk is generated.
+     *
+     * @return Whether or not this chunk is generated
+     */
+    boolean isGenerated();
 
     /**
      * Loads this chunk, and generates if specified and required.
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/Chunk.java
```diff
@@ -71,4 +78,18 @@
...
+    
+    /**
+     * Removes the chunk.
+     *
+     * @return If the chunk was successfully removed
+     */
+    boolean remove();
+    
+    /**
+     * Reegenerates the chunk.
```
Minecrell: Typo?

Bammerbom: Oops, fixed.

gabizou:
Unless I am mistaken, but a `Chunk` is an instance of a chunk that has already passed through generation. During generation, in the API, we use [`MutableBlockBuffer`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/extent/MutableBlockBuffer.java) to prepare the data that will later be transferred to a chunk. There should never exist a case where a `Chunk` has not been generated already.

Additionally, there already exists [`World.deleteChunk()`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/World.java#L82) to remove a chunk. The method should not exist in the `Chunk` itself since it can not control whether it belongs to a world or not.

Regarding `regenerate()`, you can simply delete the chunk and [load the chunk](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/World.java#L72) again.

Bammerbom:
And how about the isGenerated?

Deamon5550:
>There should never exist a case where a Chunk has not been generated already.

Bammerbom:
I mean, is there already a method for that

gabizou:
@Bammerbom If you have an instance of a Chunk, it's already __*generated*__, so why would you need a method to check if a *generated* chunk `isGenerated`?

Bammerbom:
I mean, is there any other method outside Chunk that does this yet?

Deamon5550:
```java
if(world.loadChunk(Vector3i, false) == Optional.absent())
{
  //do shit
}
```

Bammerbom:
Maybe add a better method to do this. But this pull request seems useless so someone can deny it.

Bammerbom:
Should I create a pull request that adds a isGenerated method?

gabizou:
> Should I create a pull request that adds a isGenerated method?

No. World#loadChunk#isPresent should be sufficient.

PR #366, opened 2015-01-04
SoundTypes Registry

AzureusNation:
Changes Sponge sound names to vanilla's to keep constant.
Removed the getters for the SoundTypes to the GameRegistry.

I think changing the names to minecrafts would benefit all types of implementations as it makes it a lot easier to register the sounds instead of having to hard code them in.


Minecrell:
I like how this separates the different types of sounds a little bit more than before, but basing the names on the sound IDs is a pretty bad idea in my opinion.

This will mean you'll need to change them everytime Mojang renames them and keep the old IDs as well, because the older plugins would break otherwise. To keep the compatibility you would need to hardcode them as an alias for the renamed version again, so you wouldn't save a lot with this.

Sound IDs [have changed in the past](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/cee6a7bab52ebb33c17a76b2b81891111e2f9d93), and I wouldn't say they're not going to be changed again.

ST-DDT:
Well i like your suggestion moving SoundTypes getter to GameRegistry.
But i dont like the renaming of the static variables to something that does state its exact content including repetitions instead of their plain usable name. As @Minecrell mentioned you just push the issue/matter to the next minecraft release.

AzureusNation:
But like I keep saying, it would be beneficial to everyone. They are the minecraft name, it means NO HARD CODING when registering them, and well if they do change a name in another release add a deprecations on them and create the new ones. Also that way when it has a Deprecation annotation the implementation can take that and still give it its new name until it is removed from the API. TBH the way Mojang is at the moment I can see a load of Minecraft being rewritten and a lot of things will have to change anyway.

Minecrell:
You're pretty much hard coding them into the API by changing the field names. Sure, you can easily deprecate them, but then you _need_ to "hard code" these deprecated ones into your implementation anyway. I don't see how having changing field names for the sounds over a longer period of time is beneficial to anyone.

If you compare the lines you need to update with renamed sound IDs with your pull request or with names independent from the sound ID, you need to only update the sound ID without but add a new special rule with the fallback with your pull request. So at the end, you even have more work when updating.

And if you expect changes to be coming soon we shouldn't base the names in our API on any implementation version detail.

Zidane:
Cannot merge this due to conflicts. Please fix.

AzureusNation:
@Zidane Updated and fixed conflicts.

Minecrell:
While you have added a few sounds that are still missing in our implementation, it doesn't make a lot of sense for plugin developers to use `AMBIENT_CAVE_CAVE` instead of just `AMBIENT_CAVE`. We're not trying to make Minecraft modding easier for **the implementation**, we're trying to make Minecraft modding easier for **plugin developers using our API** and this doesn't help with that at all.

See https://github.com/SpongePowered/Sponge/pull/141/files#diff-06f1c717ec93169891baac6d1bf7e03aR1064 for a list of field names matched to their sound IDs if that's the only issue.

PR #371, opened 2015-01-05
Add a way to get and set gamemode of the player

Xephi:
Player api do not have currently a way to get (or set) the gamemode of the player :/

If someone can add implementation of that ^^"

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/player/Player.java
@@ -27,6 +27,7 @@ package org.spongepowered.api.entity.player;
 
 import org.spongepowered.api.effect.Viewer;
 import org.spongepowered.api.entity.living.Human;
+import org.spongepowered.api.entity.player.gamemode.GameMode;
 import org.spongepowered.api.text.chat.ChatType;
 import org.spongepowered.api.text.message.Message;
 import org.spongepowered.api.text.title.Title;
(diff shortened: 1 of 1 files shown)


ST-DDT:
Also included there, https://github.com/SpongePowered/SpongeAPI/pull/368.
But i like the single purpose PRs.

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -117,5 +118,19 @@
      * Removes the currently displayed {@link Title} from the player's screen.
      */
     void clearTitle();
+    
+    /**
+     * Set the player {@link GameMode}.
+     * 
+     * @param gamemode The {@link GameMode} to set
+     */
+     void setGameMode(GameMode gm);
```
simon816: Parameter name does not match up with javadoc `@param`
I suggest changing it to `gameMode`

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -117,5 +118,19 @@
...
+
+    /**
+     * Get the player {@link GameMode}.
```
Phroa: you may want to change `player` to `player's current`

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -117,5 +118,19 @@
      * Removes the currently displayed {@link Title} from the player's screen.
      */
     void clearTitle();
+    
+    /**
+     * Set the player {@link GameMode}.
```
Phroa: `player` -> `player's`?

Zidane:
Already added in #368 

Thanks anyways!

PR #374, opened 2015-01-05
Create a basic config system

dentmaged:
I haven't tested it out, but Eclipse showed up with no bugs. It's quite basic and I need to work on paths

### DIFF
--- a/src/main/java/org/spongepowered/api/util/config/ConfigFile.java
+++ b/src/main/java/org/spongepowered/api/util/config/ConfigFile.java
@@ -119,11 +119,13 @@ public class ConfigFile {
             while ((line = is.readLine()) != null) {
                 line = line.trim();
                 String[] kv = line.split(": ");
-                if (kv.length == 0)
+                if (kv.length == 0) {
                     kv = line.split(":");
+                }
 
-                if (kv.length == 2)
+                if (kv.length == 2) {
                     data.put(kv[0], kv[1]);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


Mumfrey:
There are two immediate problems with this PR:

* Firstly, you seem to have reformatted the entire file to use tabs instead of spaces, this is both incorrect per the style guidelines and also makes your changes impossible to pick out because the diff is basically the whole file. Please fix this.
* Secondly, it doesn't compile because the implicit unboxing conversion is done incorrectly, I suggest you use explicit unboxing on the lines with errors. eg

   ```
if (d instanceof Double) {
        return ((Double)d).doubleValue();
}
```

  your else clauses are also redundant, and fail checkstyle because the if statements are missing braces.

dentmaged:
Ah! Thanks! I don't use doubles/floats that much. I am fixing all those bugs now.

gabizou:
I wasn't aware that SpongeAPI was moving away from using TypeSafe config. @sk89q might have some objections.

Soren025:
looks like it still uses it, just hides it. I'd rather have the full config object.

sk89q:
After some discussion a while ago, we become displeased with TypeSafe config and config is currently a "pending problem." 

However, as for this PR, I do not see what it achieves. It looks like it leaks file handles and you can't even save more than once without duplicating the file.

zml2008:
Also, the format this config implementation uses is basically the .properties format but with : instead of =.

Also the issues sk mentioned exist.

PR #376, opened 2015-01-06
Refactor minecart speed methods to be more implementation-friendly

caseif:
This PR refactors the `getMaxSpeed()` and `setMaxSpeed()` methods in the `Minecart` interface to `getSwiftness()` and `setSwiftness()`, respectively. The reasoning behind this change is that Minecraft implements a `getSpeedMethod()` (as per the Forge deobfuscation mapping) which, while similar, is [fundamentally different from its Sponge counterpart](https://github.com/SpongePowered/Sponge/pull/126#discussion_r22542226). This makes implementation difficult if not impossible and would require use of ASM to rewrite vanilla bytecode.

It also adds a `getPotentialMaxSpeed()` method to the interface which will return the maximum speed at which the minecart may travel at the instant the method is called, i.e. the return value of Minecraft's `getMaxSpeed()` method.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/vehicle/minecart/Minecart.java
+++ b/src/main/java/org/spongepowered/api/entity/vehicle/minecart/Minecart.java
@@ -54,17 +54,20 @@ public interface Minecart extends Entity {
      *
      * <p>The default value is 0.4.</p>
      *
-     * @param swiftness The new max speed
+     * @param swiftness The new maximum speed
      */
     void setSwiftness(double swiftness);
 
     /**
-     * Gets the maximum speed that this cart is allowed to travel at at the
+     * Gets the maximum speed that this cart is allowed to travel at the
      * instant this method is called.
      *
      * <p>This differs from {@link Minecart#getSwiftness()} in that its value
      * is affected by the block/rail beneath the cart. However, it is still
      * impacted and limited by the cart's swiftness.</p>
+     *
+     * @return The maximum speed at which the minecart may travel at the instant
+     * this method is called
      */
     double getPotentialMaxSpeed();
 
--- a/src/main/java/org/spongepowered/api/entity/vehicle/minecart/Minecart.java
+++ b/src/main/java/org/spongepowered/api/entity/vehicle/minecart/Minecart.java
@@ -41,24 +41,32 @@ public interface Minecart extends Entity {
     boolean isOnRail();
 
     /**
-     * Gets the maximum speed that this cart is
-     * allowed to travel at.
+     * Gets the absolute maximum speed that this cart is allowed to travel at.
      *
-     * <p>The Default value is 0.4.</p>
+     * <p>The default value is 0.4.</p>
      *
      * @return The maximum speed
      */
-    double getMaxSpeed();
+    double getSwiftness();
 
     /**
-     * Sets the maximum speed that this cart is
-     * allowed to travel at.
+     * Sets the absolute maximum speed that this cart is allowed to travel at.
      *
-     * <p>The Default value is 0.4.</p>
+     * <p>The default value is 0.4.</p>
      *
-     * @param maxSpeed The new max speed
+     * @param swiftness The new max speed
      */
-    void setMaxSpeed(double maxSpeed);
+    void setSwiftness(double swiftness);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/entity/vehicle/minecart/Minecart.java
```diff
@@ -41,24 +41,32 @@
...
+
+    /**
+     * Gets the maximum speed that this cart is allowed to travel at at the
+     * instant this method is called.
+     *
+     * <p>This differs from {@link Minecart#getSwiftness()} in that its value
+     * is affected by the block/rail beneath the cart. However, it is still
+     * impacted and limited by the cart's swiftness.</p>
+     */
+    double getPotentialMaxSpeed();
```
Minecrell: Missing return tag

gratimax:
ping @SpongePowered/developers

Mumfrey:
I don't think this is the right approach, the API really oughtn't to have to change shape to accommodate the target environment, the specific implementations should take care of any conflicts internally. Whilst mixins absolve us of the need to use wrapper objects in 99% of cases, I think there's an argument here for just boxing the thing in a wrapper object à la CraftBukkit.

Mumfrey:
Also if the decision is made to change it, then why not just use `getTopSpeed` or `getMaximumSpeed` rather than "swiftness", which is a bit tenuous.

modwizcode:
Not to mention the Block implementation is a wrapper.

Zidane:
@Mumfrey 

I think we should prioritize not using wrappers over changing API at this point. While that does let the implementation control the API somewhat, we are also still at a point where we can compromise APIvImpl and not break plugins.

Mumfrey:
@Zidane fair comment, if it doesn't rustle anyone's jimmies to change this bit of the API then nothing's really lost.

modwizcode:
@Mumfrey @Zidane I have something against the idea of saying that we should avoid wrappers to this level. Is there a huge issue with wrappers? In my mind they're a useful tool for things that ASM mixins aren't designed for. Bukkit's implementation in CraftBukkit and most other frameworks tend to also use wrappers exclusively. Obviously this works as that was the only system (besides the incompatible with vanilla forge mod servers) used until now. 

What makes it such a horrific concept to use wrappers?

sk89q:
Note that we already have quite a few existing risks of conflict in the current Sponge API with somewhat generic signatures like `getData(Class<?> clazz)` and `getType()` (on `Dimension`).

ryantheleach:
Also reading the javadocs I still have no idea how swiftness and potentialmaxspeed differ.

kenzierocks:
@ryantheleach `swifteness` is a hard limit on how fast a cart may travel, `potentialMaxSpeed` is some weird voodoo value from Forge.

Actual voodoo:
```
    protected double getMaxSpeed()
    {
        if (!canUseRail()) return func_174898_m();
        BlockPos pos = this.getCurrentRailPosition();
        IBlockState state = this.worldObj.getBlockState(pos);
        if (!BlockRailBase.func_176563_d(state)) return func_174898_m();

        float railMaxSpeed = ((BlockRailBase)state.getBlock()).getRailMaxSpeed(worldObj, this, pos);
        return Math.min(railMaxSpeed, getCurrentCartSpeedCapOnRail());
    }
```

ryantheleach:
OK, so if the API change happens, maybe rename the method? getMaxEnvironmentalSpeed? or something to indicate that it takes position and surrounding blocks into account?

At the very least indicate in the javadocs that this is the max speed that this entity can be travelling at in its current location/environment/local blockstate.

gratimax:
@ryantheleach it says it right there in the javadoc:

```java
<p>This differs from {@link Minecart#getSwiftness()} in that its value
+     * is affected by the block/rail beneath the cart. However, it is still
+     * impacted and limited by the cart's swiftness.</p>
```

Zidane:
@modwizcode 

Object churn is one of the worst enemies of Java. If we can change the name of a method and avoiding creating two objects, no one with any sense would even begin to recommend creating the objects.

I can't stress this enough: we control the API, we control the impl (and therefore direct the impl of others).

PR #377, opened 2015-01-07
Fix branch hiccup

sibomots:
The API version is not synced as the currently available API version
in the implementation. The SpongeAPI submodule is buried.  Hoping to
lift it up as the current version for subsequent builds.


simon816:
This didn't need to be a PR really

PR #378, opened 2015-01-07
Retrying API changes.

sibomots:
Tasks are just Tasks.
RepeatingTask is not necessary.
The SyncScheduler (Scheduler) keeps track of what a RepeatingTask is.
Kept getName() method.


ST-DDT:
Isn't this fixed/merged already?

Zidane:
Already in the API so closing

PR #384, opened 2015-01-11
Moved getters from Rotations to GameRegistry.

AzureusNation:
Title says all really.


Minecrell:
Thanks, added it to #221

PR #385, opened 2015-01-13
Added PlayerExperienceChangeEvent!

thomas15v:
I really think that we need an event for this :smile:.

Forum Post: https://forums.spongepowered.org/t/player-exp-change-event/5573/2


gabizou:
Already available in PR #330, specifically in this [area](https://github.com/SpongePowered/SpongeAPI/pull/330/files#diff-67ac95991a645556c4a0a8374f507ca0R33)

Issue #369, opened 2015-01-04
Replace Optional<X> with Optional<? extends X>


ST-DDT:
I encountered a little problem while looking at the Sponge Implementation and extending existing events.

**The issue**
I can overwrite every method to return a more precise result than the base interface specifies. However using the Optional destroys that benefit if used the way we currerntly do.

Here is an example code
````java
public interface Test {

    Optional<Test> getSelf();

    Optional<? extends Test> getBetterSelf();

}
````
and the second file
````java
public class Proof implements Test {

    @Override
    // public Optional<Proof> getSelf() { - NOT POSSIBLE
    public Optional<Test> getSelf() {
        // return Optional.of(this); - NOT POSSIBLE
        return Optional.of((Test) this);
    }

    @Override
    public Optional<? extends Proof> getBetterSelf() {
        return Optional.of(this);
    }
    
    public void foo() {
        Proof self = getBetterSelf().get();
        // Proof self2 = test.getSelf().get(); - NOT POSSIBLE
        Proof self2 = (Proof) test.getSelf().get();
    }

}
````
Image the Sponge server itself has  a specific subclass for one of its events, which uses a subclass of the original class/interface because a method was added that is required for further use. So if you listen to the given event and you try to get it, you will actually get the base interface because you have to downcast it, but if you use the `<? extends T>` for example this is no longer an issue you class can simply overwrite the return type and everything will be fine.

If we change it we have basically three benefits:
* Avoid unnecessary casting while creating the `Optional`
* Avoid unnecessary casting while getting the `Optional`'s value
* Allow other APIs to extend our API more safely and freely.

I suggest replacing/changing it in the entire API (where the Optional references a class within the  `org.spongepowered.api.**` package) , i also offer my help to change it.
This would also include immutable `Iterable`s, `Collection`s, ...., `Map`s.

Minecrell:
I can confirm this is a problem for some parts of the API, but we need to check where having more precise results are really need it. I had to use it in #367 in order to be able to override the return value in `StatusPingEvent` for a method with `Optional`. https://github.com/SpongePowered/SpongeAPI/pull/367/files#diff-70bb07711fbf5aabad54031ef7fd11aaR59

ST-DDT:
IMO it is need everywhere or at least useful everywhere.
Think of CopyPasting the entire API and just replace all interfaces with implementations, including those in the optionals. 
[...]

PR #386, opened 2015-01-13
Use <? extends X> instead of <X> for most generics.

ST-DDT:
Associated Issue: https://github.com/SpongePowered/SpongeAPI/issues/369

This PR replaces most occurrences of generics with those that are easier to extend in subclasses.

If there are places that you don't want to add the more generic things, i can remove them there.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
+++ b/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
@@ -42,6 +42,7 @@ import org.spongepowered.api.util.command.CommandCallable;
 import org.spongepowered.api.util.command.CommandException;
 import org.spongepowered.api.util.command.CommandMapping;
 import org.spongepowered.api.util.command.CommandSource;
+import org.spongepowered.api.util.command.ImmutableCommandMapping;
 import org.spongepowered.api.util.command.dispatcher.SimpleDispatcher;
 import org.spongepowered.api.util.event.Order;
 import org.spongepowered.api.util.event.Subscribe;
@@ -61,7 +62,7 @@ public class SimpleCommandService implements CommandService {
 
     private final PluginManager pluginManager;
     private final SimpleDispatcher dispatcher = new SimpleDispatcher();
-    private final Multimap<PluginContainer, CommandMapping> owners = HashMultimap.create();
+    private final Multimap<PluginContainer, ImmutableCommandMapping> owners = HashMultimap.create();
     private final Object lock = new Object();
 
     @Inject
@@ -83,17 +84,17 @@ public class SimpleCommandService implements CommandService {
     }
 
     @Override
-    public Optional<CommandMapping> register(Object plugin, CommandCallable callable, String... alias) {
+    public Optional<ImmutableCommandMapping> register(Object plugin, CommandCallable callable, String... alias) {
         return register(plugin, callable, Arrays.asList(alias));
     }
 
     @Override
-    public Optional<CommandMapping> register(Object plugin, CommandCallable callable, List<String> aliases) {
+    public Optional<ImmutableCommandMapping> register(Object plugin, CommandCallable callable, List<String> aliases) {
         return register(plugin, callable, aliases, Functions.<List<String>>identity());
     }
 
     @Override
-    public Optional<CommandMapping> register(Object plugin, CommandCallable callable, List<String> aliases,
+    public Optional<ImmutableCommandMapping> register(Object plugin, CommandCallable callable, List<String> aliases,
                                              Function<List<String>, List<String>> callback) {
         checkNotNull(plugin);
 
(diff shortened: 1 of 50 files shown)


src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
```diff
@@ -61,7 +62,7 @@
 
     private final PluginManager pluginManager;
     private final SimpleDispatcher dispatcher = new SimpleDispatcher();
-    private final Multimap<PluginContainer, CommandMapping> owners = HashMultimap.create();
+    private final Multimap<PluginContainer, ImmutableCommandMapping> owners = HashMultimap.create();
```
nightpool: Isn't this out of score for this PR?

ST-DDT: Maybe, i can remove it if it is not liked (I used a second commit for that so i can revert it if it is not liked.) , but from my point of view the SimpleCommandService is a class that will not be subclasses and thus can be manipulated to match its real contents.

Its more or less an example how the changes to the API could be used. 

I also work on the server implementation, but there are currently some unrelated issues that prevent my IDE from showing any useful compile errors.

Zidane:
Cannot merge, please fix

ryantheleach:
I've only just started learning about wildcard generics and they still confuse me, but do you really need <? extends interface> Wouldn't all classes implementing that interface fill <interface> even if they are subclasses of the original class that implemented the interface?

gabizou:
> I've only just started learning about wildcard generics and they still confuse me, but do you really need <? extends interface> Wouldn't all classes implementing that interface fulfil < interface> even if they are subclasses of the original class that implemented the interface?

The issue arrises when a collection of `Foo` which extends `Bar` can't be returned as `List<Bar>` unless explicitly declared and/or casted as `Bar`.

ryantheleach:
Yes, but that isn't the case we are dealing with in the API is it?

Don't we have a collection of objects that implement a common interface?

gabizou:
> Don't we have a collection of objects that implement a common interface?

Yes, and this is why it's troublesome, when an implementation has `List<MyAnimal>` where `MyAnimal extends Animal`, it can't just return that list, it has to re-cast the list as `List<Animal>`.

On the other hand, there are no valid reasons to have fine tuning of these interfaces since there simply isn't one in the API, a `Human` is always a `Human`, regardless of implementation.

gabizou:
Speaking with some more @SpongePowered/developers in IRC, we've made the comments that this PR specifically has the following:
- It does NOT remove any casting in the implementation since at compile time, nothing implements the API.
- It doesn't add to the API in any way, in fact, it makes certain parts of the API look a bit convoluted/confusing as they don't directly affect the API in any way.
- In implementations, casting is almost always required.

ST-DDT:
I'm not sure whether we should use the sponge server impl as a reference. A CB like impk would defenitly have more casting required.
But this is up to you/the dev team.

Shall I stop working on the server side PR for this?

Zidane:
I've talked with a few and heard both sides. I just don't see the merit in this when a cast is easily performed. As such, closing.

Thanks @ST-DDT for bringing this forward. Very good discussion and learning all around.

Minecrell:
We should keep this in mind because it is and will be required at some places in the API, for example I could imagine this with events. I can't think of a specific example right now, but here is a small example showing where it is necessary:

```java
interface OptionalEntityEvent extends Event {
    // This needs to have the different type argument
    // so we can make it more specific in the extended event.
    // Optional<Entity> getEntity();
    Optional<? extends Entity> getEntity();
}

interface OptionalPlayerEvent extends OptionalEntityEvent {
    @Override
    Optional<Player> getEntity();
}
```

Zidane:
@Minecrell 

Thats fine, it can be handled on a case by case basis.

ryantheleach:
At some point you will want the API to be stable for a release. Handling it on a case-by-case basis means potentially missing some situations where this would be needed, and as far as I can see including this wouldn't hurt except inside the API itself.

To save pain later on wouldn't the change for all optionals be a good idea?

gabizou:
> To save pain later on wouldn't the change for all optionals be a good idea?

The issue with a blanket application of all optionals is generally a bad idea. Just think, does it make sense that there would be a more fine grained implementation of `Game` made available in the API? The same applies to several of the optionals.

ST-DDT:
For more implementation specific APIs it would defenitly makes sense.
SpongeAPI < ImplAPI < Impl

modwizcode:
But it doesn't matter if we never want a more specific type. So Game
wouldn't ever make any sense here.

Axiometry:
The takeaway from this is that Java Generics are terrible :cry:

ST-DDT:
>The takeaway from this is that Java Generics are terrible :cry:

No, they are not terrible, they are logical and safe

> But it doesn't matter if we never want a more specific type. So Game wouldn't ever make any sense here.

So is this idea completely refused or will we change it for only a few occurrences?

PR #387, opened 2015-01-13
ArmourStands and Entities

AzureusNation:
All entities have the option to be invisible or not so I have moved the getter and setter to Entity. Also I have added noGravity to ArmorStand as that was missing when I was implementing.

Only thing with Entities is that they may have their own setInvisible method so in the implementations the setInvisible will need to be Override so that the correct minecraft method can be ran.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/ArmorStand.java
+++ b/src/main/java/org/spongepowered/api/entity/living/ArmorStand.java
@@ -132,18 +132,18 @@ public interface ArmorStand extends Living, ArmorEquipable {
     void setSmall(boolean small);
 
     /**
-     * Returns whether this armor stand is rendered invisible.
+     * Returns whether this armor stand is affected by gravity or not.
      *
-     * @return Whether this armor stand is invisible
+     * @return Whether this armor stand is affected by gravity or not
      */
-    boolean isInvisible();
+    boolean hasGravity();
 
     /**
-     * Sets whether this armor stand is invisible or not.
+     * Sets whether this armor stand is affected by gravity or not.
      *
-     * @param invisible Whether this armor stand is invisible or not
+     * @param gravity Whether this armor stand is affected by gravity
      */
-    void setInvisible(boolean invisible);
+    void setGravity(boolean gravity);
 
     /**
      * Returns whether this armor stand shows arms or not.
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/entity/living/ArmorStand.java
```diff
@@ -132,18 +132,18 @@
     void setSmall(boolean small);
 
     /**
-     * Returns whether this armor stand is rendered invisible.
+     * Returns whether this armor stand is affected by gravity or not.
      *
-     * @return Whether this armor stand is invisible
+     * @return Whether this is a small armor stand
```
kenzierocks: This seems incorrect, is this supposed to be a different method?

src/main/java/org/spongepowered/api/entity/living/ArmorStand.java
```diff
@@ -132,18 +132,18 @@
...
      *
-     * @return Whether this armor stand is invisible
+     * @return Whether this is affected by gravity or not
      */
-    boolean isInvisible();
+    boolean hasNoGravity();
 
     /**
-     * Sets whether this armor stand is invisible or not.
+     * Returns whether this armor stand is affected by gravity or not.
```
Minecrell: The previous Javadocs were correct, this sets something, but doesn't return something.

src/main/java/org/spongepowered/api/entity/living/ArmorStand.java
```diff
@@ -132,18 +132,18 @@
     void setSmall(boolean small);
 
     /**
-     * Returns whether this armor stand is rendered invisible.
+     * Returns whether this armor stand is affected by gravity or not.
      *
-     * @return Whether this armor stand is invisible
+     * @return Whether this is affected by gravity or not
```
Minecrell: `this armor stand is ...` sounds better than `this is ...`, only `this` would refer to the operation of the method, and not armor stand.

Minecrell:
Wouldn't `hasGravity` and `setGravity` make more sense than an additional negation with `setNoGravity`? While the property of the armor stand is defined as `NoGravity`, the method would make more sense the other way around, IMO

AzureusNation:
> Wouldn't  hasGravity  and  setGravity  make more sense than an additional negation with  setNoGravity ? While the property of the armor stand is defined as  NoGravity , the method would make more sense the other way around, IMO

@Minecrell I can change that, no problem

PR #388, opened 2015-01-14
added parseResource method to ConfigFile.

thomas15v:
This will hopefully avoid duplicated code across plug-ins. It would be cleaner to read plug-in code that uses this method :).

Forum post: https://forums.spongepowered.org/t/help-me-understand-configuration/5582/3

### DIFF
--- a/src/main/java/org/spongepowered/api/util/config/ConfigFile.java
+++ b/src/main/java/org/spongepowered/api/util/config/ConfigFile.java
@@ -244,12 +244,14 @@ public final class ConfigFile implements Config {
      *
      * @param resourceName The name of the resourceFile
      * @param saveFile The file where the configFile will be saved
+     * @param pluginObject The plugin object to get the resource from
      * @return
      */
-    public static ConfigFile parseResource(String resourceName, File saveFile){
+    public static ConfigFile parseResource(String resourceName, File saveFile, Object pluginObject){
+
         checkNotNull(resourceName);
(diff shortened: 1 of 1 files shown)


jk-5:
I agree in adding this method, but i don't agree with the name. The name doesn't imply that it saves the configuration file to disk. Maybe something like `parseAndSaveResource` is better

thomas15v:
It doesn't save. But indeed maybe a force-save method will be better. Than could avoid extra duplicated code.

jk-5:
Sorry. I first read the code and then the forum post so i accidentally thought this method saved too. Why not create such a method as mentioned in the forum post?

thomas15v:
Idk, I was thinking. In case the dev wants to add something else before it saves the config. It would save it twice. And that not really efficient.

thomas15v:
I just realise that this can't work without giving the plugin object as parameter. But wouldn't that look to messy?

Zidane:
@sk89q 

Gonna let you make the call on this as you are adjusting the config API currently.

sk89q:
We are contemplating replacing the configuration code entirely with something new.

Something similiar to https://github.com/sk89q/Eduardo/tree/master/src/main/java/com/sk89q/eduardo/util/config but with support for serialization of arbitrary objects.

Zidane:
We've switched to configurate-hocon but thanks anyhow!

PR #389, opened 2015-01-14
[WIP] Commands System Improvement

AlphaModder:
This slightly modifies the command interfaces to allow for more variety in command systems, and adds a more featured default command system The features it will introduce to it are:

* Allows use of /plugin:command in a less hacky way.
* Allows plugins to register conflicting commands, and resolves those conflicts when the command is sent.
* Removes the callback feature from command registration as it is no longer neccessary due to the above change.
* Allows use of /mc or /minecraft:command to run vanilla commands [NYI]
* Creates an API that may be used to set certain aliases to point to certain plugins in the command system.

It is based on EvilSeph's old blog post [here](http://evilseph.com/post/75608917830/bukkit-supporting-minecraft-commands).

Due to configuration files not yet having a standard format yet, I will only create an API that allows the command configuration described in the blog post to be implemented, but this PR does not include the actual implementation of the configuration file. 

I still need help with: a good way to wrap vanilla/sponge commands into the existing system. maybe some kind of VanillaPluginContainer/SpongePluginContainer as wrappers so the command system can use them?


me4502:
This has many formatting issues, as well as making gradlew non-executable.

me4502:
@sk89q - You'd probably have the most to say about this PR.

AlphaModder:
@me4502 Whoops with the gradlew thing. Not used to having other stuff in a repository than code :P 
And I plan on working on the formatting tommorow.

AlphaModder:
@me4502 I'm afraid with the gradlew thing, I use Windows, so I have no idea exactly how to change octal-style permissions.

kenzierocks:
`git update-index --chmod=+x gradlew`

AlphaModder:
That's weird, I thought I set tabs to be converted to spaces...

sk89q:
These changes make the command API pretty much only for use by Sponge, which defeats the purpose of having the command API. SimpleDispatcher cannot have any reference of plugins in it.

AlphaModder:
@sk89q I'm afraid I don't understand. I've modified SimpleDispatcher so that if it's being extended, for example, in a plugin, it disables all but the old parts. Is there another problem?

AlphaModder:
@sk89q If you think it's neccessary, i could restore the old SimpleDispatcher (mostly) and have a SpongeDispatcher or a ComplexDispatcher or something.

me4502:
Having it extended would probably be best, if it is really required.

sk89q:
The API in the util package is supposed to be a generic command dispatcher API -- something that could reasonably be removed from Sponge or used in a plugin, and then the APIs in the service package is supposed to contain the Sponge-specific parts (aliases, etc.). That is why SImpleCommandService merely implemented plugin information on top of SimpleDispatcher.

The point is to decouple plugin access from the commands code.

AlphaModder:
@sk89q Okay, I understand now, So should I perhaps refactor SimpleCommandService to do all the new stuff? Or should I create a new Dispatcher and CommandService. If you actually want this in sponge, that is.

AlphaModder:
@sk89q For me, the second option makes more sense I think, if only because it's much easier and less hacky to implement the dynamic resolution on the side of the Dispatcher.

Zidane:
@sk89q 

Commands are your thing and I'm leaving this to you.

AlphaModder:
I'm gonna close this specific PR and create a new one with the same code to change the branch structure in my fork.

PR #392, opened 2015-01-16
Fix many (many many) minor issues

simon816:
Fixes all issues currently listed in #221 

Some credit goes to the authors of PR's that did one or two fixes.

On the issue "Move remaining type getters to `GameRegistry`", there are some things that may or may not be moved, please suggest whether the getters in the following classes should be in `GameRegistry`:
 * ~~[`LivingMetaFactory`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/living/meta/LivingMetaFactory.java)~~ All methods exist in GameRegistry
 * ~~[`VillagerFactory`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/living/villager/VillagerFactory.java)~~ All methods exist in GameRegistry
 * ~~[`TextStyles`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextStyles.java)~~ Need to decide on factories first
 * ~~[`TextColors`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextColors.java)~~ Need to decide on factories first
 * ~~[`ChatTypes`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/chat/ChatTypes.java)~~ Need to decide on factories first
 * ~~[`TextActions`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/action/TextActions.java)~~ Need to decide on factories first
 * ~~[`Messages`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/message/Messages.java)~~ Need to decide on factories first
 * ~~[ `Selectors`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/selector/Selectors.java)~~ Need to decide on factories first
 * ~~[`SelectorTypes`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/selector/SelectorTypes.java)~~ Need to decide on factories first
 * ~~[`Titles`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/title/Titles.java)~~ Need to decide on factories first
 * ~~[`Translations`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/translation/Translations.java)~~ Need to decide on factories first

The Javadocs in SpongeEventFactory were mostly write once, replace all occurrences and so please comment if the wording could be improved.

~~There may be some more checkstyle issues not on the minor issues list, I will fix them if requested.~~
All fixed


src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -391,6 +408,21 @@
      */
     List<DimensionType> getDimensionTypes();
 
+    /**
+     * Gets the {@link Rotation} with the provided degrees.
+     *
+     * @param degrees The degrees of the rotation
+     * @return The {@link Rotation} with the given degrees or Optional.absent() if not found
+     */
+    Optional<Rotation> getRotationFromDegree(int degrees);
```
me4502: Should we provide one for radians as well? Java Math uses radians.

Minecrell:
There are still some checkstyle problems, can you fix them as well? https://travis-ci.org/SpongePowered/SpongeAPI/builds/47198760#L254

File | Problem
---- | ----
[GameRegistry:61](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/GameRegistry.java#L61) | Wrong lexicographical order for `java.io.InputStream` import.
[BlockType:29](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/block/BlockType.java#L29) | Wrong lexicographical order for `org.spongepowered.api.service.persistence.data.DataHolder` import.
[StatusPingEvent:30](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/server/StatusPingEvent.java#L30) | Wrong lexicographical order for `org.spongepowered.api.GameProfile` import.
[Task:0](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/scheduler/Task.java#L0) | File does not end with a newline.
[StatusResponse:29](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/status/StatusResponse.java#L29) | Wrong lexicographical order for `org.spongepowered.api.GameProfile` import.
[BiomeArea:29](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/gen/BiomeArea.java#L29) | Wrong lexicographical order for `com.flowpowered.math.vector.Vector2i` import.
[BlockBuffer:31](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/gen/BlockBuffer.java#L31) | Wrong lexicographical order for `com.flowpowered.math.vector.Vector3i` import.
[DimensionType:51](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/DimensionType.java#L51) | GenericWhitespace `&lt;` is preceded with whitespace.
[WorldGenerator:31](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/gen/WorldGenerator.java#L31) | Wrong lexicographical order for `com.flowpowered.math.vector.Vector3i` import.
[WorldStorage:29](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/storage/WorldStorage.java#L29) | First sentence should be present.
[WorldStorage:31](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/storage/WorldStorage.java#L31) | At-clause should have a non-empty description.
[WorldStorage:33](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/storage/WorldStorage.java#L33) | Expected an `@return` tag.

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -103,7 +160,19 @@ private SpongeEventFactory() {
         return (T) factories.getUnchecked(type).apply(values);
     }
 
-    public static BlockBreakEvent createBlockBreak(Game game, Cause cause, BlockLoc block, BlockSnapshot replacementBlock, int exp, Collection<Item> droppedItems) {
+    /**
+     * Create a new {@link BlockBreakEvent}.
```
Minecrell: All these Javadocs should be `Creates ...`, not `Create`.

Minecrell:
You can move all the methods to GameRegistry excluding the `text` package. We have not decided what we're going to do with the factories in there yet.

simon816:
Fixed all remaining issues.

Tests (`src/test/java`) have many issues, there are several things that could be done:
 * Ignore warnings
 * Disable checkstyle for tests
 * Use a different checkstyle config for tests that don't check for javadocs

extra/sponge.importorder
```diff
@@ -0,0 +1,7 @@
+#Organize Import Order
```
Minecrell: Is this file used by Eclipse or why did you add this?

simon816: Yeah it's for Eclipse, and a useful reference to know what order imports should be.

me4502: Including this file could be useful. Is there an IntelliJ equivalent worth adding as well?

Minecrell: The IntelliJ definition is already included in the our formatted settings.
Why do you need to include all these packages? Generally the imports are defined like this:

- Static imports
- All other imports
- `java` imports
- `javax` imports

With each group of the above separated by an empty line.

simon816: Eclipse cannot separate import groups automatically (i.e. a group of `org.*` then a group of `net.*`), each group has to be defined so I have defined the most common groups here
I notice that the order was exported upside-down, I'll change that

Minecrell: These groups shouldn't be separated from each other, only `java`, `javax` and the static imports should be in a separate group.

simon816: [They should be](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.3-import-ordering-and-spacing)
> Third-party imports, one group per top-level package, in ASCII sort order

Minecrell: Ah. Well, we don't have this in our IntelliJ configuration, and I don't like adding only the most common packages, it should work correctly for any import. For now I would just do it like our IntelliJ configuration.

simon816: Changed. It should be mentioned in the [Contributing Guidelines](https://docs.spongepowered.org/en/latest/devs/guidelines/) as a difference from Google's style guide

src/main/java/org/spongepowered/api/GameRegistry.java
ST-DDT: I agree on this one as well.

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -230,7 +416,18 @@ public static EntityBreakBlockEvent createEntityBreakBlock(Game game, Cause caus
...
+     *
+     * @param game The game instance for this {@link GameEvent}
+     * @param cause The cause of the event, can be null
+     * @param entity The entity involved in this event
+     * @param block The block affected by this event
+     * @param replacementBlock The block that will replace the existing block
+     * @return A new instance of the event
+     */
+    public static EntityChangeBlockEvent
+            createEntityChangeBlock(Game game, Cause cause, Entity entity, BlockLoc block, BlockSnapshot replacementBlock) {
```
ST-DDT: Why did you push the method name to a new line?
In line [R406] (https://github.com/SpongePowered/SpongeAPI/pull/392/files#diff-1378043d9608b3c46ff7baa499da8e70R406) and line [R174](https://github.com/SpongePowered/SpongeAPI/pull/392/files#diff-1378043d9608b3c46ff7baa499da8e70R174) you moved the method arguments to the new line.

gabizou:
I'm not sure if it's relevant, but the javadoc changes on `SpongeEventFactory` are being changed around with #330 as many events have changed requirement. @SpongePowered/developers would it be wise to keep these changes as they pertain to master, or move these changes as needed for the Events PR?

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
simon816: Hmm, eclipse formatter did that, I'll check for any more occurrences of this and change them.

simon816: The [Google style guide](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.5.1-line-wrapping-where-to-break) states:
> There is no comprehensive, deterministic formula showing exactly how to line-wrap in every situation. Very often there are several valid ways to line-wrap the same piece of code.

So either way of wrapping is allowed in this situation. I'll wrap the arguments to be consistent however IMO it's neater to move the method name down.

extra/sponge.importorder
ST-DDT: I don't know whether you just haven't uploaded it yet, but there are some classes/files with empty lines in the imports, that should not be there. If i got it right.
Here is a list of them (i'll add more if i find more):
* Server
* EnderDragon
* Enderman
* SimpleServiceManager
* StatusResponse
* Subscribe
* ClassGenerator
* BiomeArea
* BlockBuffer
* AccessorFirstStrategy
* WorldGenerator
* ClassGeneratorProviderTest

simon816: @ST-DDT That would be because I (and possibly some other contributors) was organizing imports per top-level group, but now I learn we are doing it differently, I'll change it accordingly.

src/main/java/org/spongepowered/api/GameRegistry.java
simon816: The [`Rotation`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/rotation/Rotation.java) interface only provides the angle in degrees. There is a TODO marker to create an `Angle` interface which would be the appropriate place for adding radians, so that should be done as it's own PR/commit

ST-DDT:
Can you do me a favour and add the empty before the closing brackets and one empty line between the licence header and the package? As specified by SpongeAPI's [code template](https://docs.spongepowered.org/en/latest/devs/guidelines/#the-gist). It's just visuals and usually not worth a PR, so it would be great if you can include it in yours, that already contains a lot of formatting.
Example:
````java
 */
package org.spongepowered;
[..]
    }
}
````
to
````java
 */

package org.spongepowered;
[..]
    }

}
````
Here are some example regex to simplify your work:
Search for: ` \*\/\npackage`
Replace with: `\*\/\n\npackage`

Search for: `\}\n\}`
Replace with: `\}\n\n\}`
On windows use `\r\n` instead of `\n`

extra/sponge.importorder
ST-DDT: Thanks for fixing those. I guess you have missed this one:
`org.spongepowered.api.Server`
That should be the last one for now. The regex does not find any other places.

simon816:
@ST-DDT I have done that now, see https://github.com/simon816/SpongeAPI/commit/e254b3c4f0674ff4370a81229fdf0690812096bd

However it has pushed the total number of modified files from 54 to 248 which it may not be a good idea to make such small changes to that number of files.

ST-DDT:
Wow. I did not realize that it would be that much places. Thanks for your work and time.

gabizou:
> However it has pushed the total number of modified files from 54 to 248 which it may not be a good idea to make such small changes to that number of files.

Please undo that change as it heavily pollutes the history of the files unnecessarily imo. though @SpongePowered/developers probably should chime in whether it's necessary, like @Phroa

simon816:
@gabizou 
`git reset HEAD~1 --hard`
Done

AzureusNation:
@gabizou Can you merge this asap please.

PR #394, opened 2015-01-17
Make use of ListenableFutures and other tweaks for the Scheduler

PieGuy128:
The Guava library provides a set of tools in its "util.concurrent" package that, I believe, could be of great use to plugin developers.

In Bukkit, if a developer wanted to, for example, retrieve information from a database asynchronously then do something with it synchronously, he/she would have to jump back-and-forth using something that resembles the following:

```
runAsyncTask(new Runnable() {
    try {
        final String x = getSomeData();
        runSyncTask(new Runnable() {
            void run(){
                getLogger().info(x);
            }
        });
    } catch (final SomethingTerrible e) {
        runSyncTask(new Runnable() {
            void run(){
                getLogger().info(e.getMessage());
            }
        });
    }
});
```

Using Guava's ListenableFuture, we can instead use...

```
PluginContainer container = getContainerFromSomewhere();
AsynchronousScheduler async = game.getAsyncScheduler();
SynchronousScheduler sync = game.getSyncScheduler();

Task<String> task = async.runTask(container, new Callable<String>() {
	@Override
	public String call() throws Throwable {
		return getSomeData();
	}
}).get();

Futures.addCallback(task.getFuture(), new FutureCallback<String>() {
	@Override
	public void onSuccess(String result) {
		getLogger().info(result);
	}

	@Override
	public void onFailure(Throwable t) {
		getLogger().info(t.getMessage());
	}
}, sync.getExecutorService(container));

```

With Bukkit's system, code could become difficult to follow, especially if you needed to jump back-and-forth multiple times.
With ListenableFuture, we can simply listen for the completion of the task then do something with the information it returns.

Implementation side will come soon!


ST-DDT:
Why doesn't scheduler declare (most of) the methods?
Now you have the overhead of declaring them in both interfaces which makes it harder to read and harder to implement it in the server with a shared base class.

sibomots:
I see the assignment.  I'll take a good look.  There are a lot of good ideas here.  Give me a chance to review and test the suggestions.  Thank you.

sibomots:
@ST-DDT I couldn't agree  more.  The thing that has been bothering me is the obvious duplication of the two interfaces except for the Time Unit involved.   I am toying with a way to introduce a Time Unit type (interface perhaps) to disambiguate one interface for either scheduler.

sibomots:
I'm in the process of merging these into feature/async-scheduler, then modify some issues to address the comments above.  I'll push an updated feature/async-scheduler based on the commits of this PR when that's done.

ryantheleach:
Honestly I don't understand what is asynchronous in this scheduler. as far as  I can tell, this is just a synchronous scheduler that runs off of the wall clock.

If this is true, people coming from Bukkit are going to be painfully surprised that there is a scheduler named similarly, that performs a very different function.

sibomots:
The API for the Scheduler puts the context plainly - the object managed by the Scheduler(s) are Tasks, defined by plugins that execute according to rules afforded by the Scheduler(s) and the parameters defined by the caller (the Plugin).
The term "synchronous" is always relative to something.  With Sponge, the context and relativity of the Synchronous Task means the task is run in sync with the game's own ServerTickEvent.  Ergo, the Synchronous Task is synchronous with the ServerTickEvent.
Asynchronous Tasks are different.  They are not synchronous to the game, not synchronous to the game ServerTickEvent.  They merely run whenever they are supposed to run defined by the parameters set when the Task is requested (run after a certain TimeUnit from now,  run with a period (interval) in TimeUnit time).
If a Bukkit-aware plugin author is interested in events that are (A) synchronous to the game, they are dealing with Events, not Tasks and (B) the events a plugin can be aware of are ever-increasing in number as the Mixin code for the palette of Events is fulfilled.   Events, not Tasks.
The synchronous thing about Events is only that they can safely modify game data.   Asynchronous they are in terms of the source of the event, but the queuing and firing, and resolution -- they are synchronous to game data.


Date: Tue, 20 Jan 2015 15:08:14 -0800

PieGuy128:
I've had to remove the "runRepeatingTask" methods that take Callables as the task body, as (Listenable)FutureTasks will only return a value from the Callable the one time (Gah, should have caught that!).

If I am able to find a solution, I will re-add those methods.  But until then they shouldn't be present.

Regards,
Mark

ryantheleach:
sibo, I deleted my comment almost instantly after I posted it because I realized I was commenting on the wrong pull request.

I wasn't meaning events, but tasks.

but anyway, if the tasks are running on the same thread, no matter the time unit, then calling the scheduler asynchronous given our audience is going to violate the Principal Of Least Astonishment.

PieGuy128:
If we were to create our own time scaling enum/class and use that in place of Java's TimeUnit (for repeating/delayed tasks), we could possibly remove two of the Scheduler interfaces, and implement the differing types on the implementation side.

So, we would have the one Scheduler interface, and implementation-side we would have the AsyncScheduler and the SyncScheduler.

Example of how we could tackle the scaling: https://gist.github.com/PieGuy128/db410a1af4e3378ff4a6

Best regards,
Mark

sibomots:
So there was a bit of discussion on IRC about the bug I introduced in the AsyncScheduler implementation as it stands now.
The bug (not intentional, just a copy and paste mistake) was that when AsyncScheduler begins to actually run Runnables they will run in the same thread as the state machine owned by the AsyncScheduler.  This is not intended.
Instead, the correct action is to make the AsyncScheduler run the pending-due task in a thread group separate from the game main thread, and separate from the state-machine thread of the AsyncScheduler.
That's not too hard to fix.  And @Minecrell offeerd a good suggestion to that issue.
The second point about time unit scale is sort of a moot point.   The internals of the AsyncScheduler bases all decisions on pending tasks turning into running tasks based on the simple calculation that  the timestamp + offset (or timestamp + period) must be equal to or greater than the current time.  It doesn't matter what AsyncScheduler uses (what scale it uses) to store "now" (the current time).  Repeating tasks always reset the timestamp to force the next invocation.
What does matter is when a Task is scheduled, the caller can (with the recent commit) set the TimeUnit so the arguments are interpreted in that time unit. (See use JavaDoc comments in the implementation -- which reminds me they should be copied to the API JavaDoc, my bad).

// The task will run with a period of 120 milliseconds and delay of 302 milliseconds
// (If the scales are the same for both arguments)
runRepeatingTaskAfter(somePlugin, someRunnableTarget, TimeUnit.MILLISECONDS, 120, 302);

// If the two units are in different scales:

// The task will run with a period of 20 seconds and delay of 500 milliseconds:
// Either:

runRepeatingTaskAfter(somePlugin, someRunnableTarget, TimeUnit.MILLISECONDS, TimeUnit.SECOND.toMillis(20), 500);

// OR

runRepeatingTaskAfter(somePlugin, someRunnableTarget, TimeUnit.SECONDS, 20, TimeUnit.MILLISECONDS.toSeconds(500));


Date: Wed, 21 Jan 2015 18:13:17 -0800

ryantheleach:
Oh, that makes much more sense.

PieGuy128:
Seeing the recent changes, should I be recreating this PR, aimed to another branch?

sibomots:
Branch the master and make a PR on your proposed changes to the Schedulers.

Zidane:
Can't merge this.

Re-assigning to @sibomots

PieGuy128:
Aye!  I'll go ahead and close this PR, and open another one aimed at master once I get both API and impl. sides working!

Best regards,
- Mark

PR #395, opened 2015-01-18
Entity Fixes

AzureusNation:
This is pretty self explanatory but just to elaborate.

Sitting needs to extend Tameable as Sitting can only be accomplished by a Tameable animal.
Tameable needs to extend Animal.
Squid only needs Aquatic as it is not an Animal in Minecraft.

This should fix a few issues that would have arisen if not sorted now.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/Sittable.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Sittable.java
@@ -42,5 +42,4 @@ public interface Sittable extends Tameable {
      * @param sitting Whether this entity is sitting
      */
     void setSitting(boolean sitting);
-
 }
--- a/src/main/java/org/spongepowered/api/entity/living/Tameable.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Tameable.java
@@ -26,11 +26,12 @@ package org.spongepowered.api.entity.living;
 
 import com.google.common.base.Optional;
 import org.spongepowered.api.entity.Tamer;
+import org.spongepowered.api.entity.living.animal.Animal;
 
 /**
  * Represents an Agent that can be tamed by a Tamer.
  */
-public interface Tameable extends Agent {
+public interface Tameable extends Animal {
 
     /**
      * Checks if this is tamed
@@ -66,5 +67,4 @@ public interface Tameable extends Agent {
      * @param tamer The Tamer who should own this
      */
     void setOwner(Tamer tamer);
-
 }
--- a/src/main/java/org/spongepowered/api/entity/living/Tameable.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Tameable.java
@@ -29,7 +29,7 @@ import org.spongepowered.api.entity.Tamer;
 import org.spongepowered.api.entity.living.animal.Animal;
 
 /**
- * Represents an Agent that can be tamed by a Tamer.
+ * Represents an Animal that can be tamed by a Tamer.
  */
 public interface Tameable extends Animal {
 
--- a/src/main/java/org/spongepowered/api/entity/living/animal/Horse.java
+++ b/src/main/java/org/spongepowered/api/entity/living/animal/Horse.java
@@ -37,7 +37,7 @@ import javax.annotation.Nullable;
 /**
  * Represents a Horse.
  */
-public interface Horse extends Animal, Tameable {
+public interface Horse extends Tameable {
 
     /**
      * Gets the current style of this Horse.
@@ -102,5 +102,4 @@ public interface Horse extends Animal, Tameable {
      * @param itemStack The saddle item
      */
     void setSaddle(@Nullable ItemStack itemStack);
-
 }
--- a/src/main/java/org/spongepowered/api/entity/living/animal/Ocelot.java
+++ b/src/main/java/org/spongepowered/api/entity/living/animal/Ocelot.java
@@ -31,7 +31,7 @@ import org.spongepowered.api.entity.living.meta.OcelotType;
 /**
  * Represents an Ocelot.
  */
-public interface Ocelot extends Animal, Sittable {
+public interface Ocelot extends Sittable {
 
     /**
      * Gets the current {@link OcelotType} of this ocelot.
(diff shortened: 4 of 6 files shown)


src/main/java/org/spongepowered/api/entity/living/Sittable.java
```diff
@@ -42,5 +42,4 @@
      * @param sitting Whether this entity is sitting
      */
     void setSitting(boolean sitting);
-
```
Minecrell: Why do you need to remove this line?

src/main/java/org/spongepowered/api/entity/living/Tameable.java
```diff
@@ -26,11 +26,12 @@
 
 import com.google.common.base.Optional;
 import org.spongepowered.api.entity.Tamer;
+import org.spongepowered.api.entity.living.animal.Animal;
 
 /**
  * Represents an Agent that can be tamed by a Tamer.
```
simon816: Javadoc would need to be updated.

src/main/java/org/spongepowered/api/entity/living/Sittable.java
AzureusNation: Didn't need to just got in the way.

src/main/java/org/spongepowered/api/entity/living/Tameable.java
AzureusNation: on it.

src/main/java/org/spongepowered/api/entity/living/Sittable.java
Minecrell: Please remove these changes, as mentioned in the [contribution guidelines](https://docs.spongepowered.org/en/latest/devs/guidelines/#code-style) they're supposed to be there:

> Place a blank line before the first member of a class, interface, enum, etc. (i.e. after `class Example {`) as well as after the last member

Mumfrey: Actually, per the style guide they're [completely optional](https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.1-vertical-whitespace). So it's best not to fret about these and just leave them however they happen to be in order to reduce diff pollution.

Minecrell: We have a few changes to the original Google java style in our [contribution guidelines](https://docs.spongepowered.org/en/latest/devs/guidelines/#code-style).

src/main/java/org/spongepowered/api/entity/living/Tameable.java
```diff
@@ -26,11 +26,12 @@
...
 import com.google.common.base.Optional;
 import org.spongepowered.api.entity.Tamer;
+import org.spongepowered.api.entity.living.animal.Animal;
 
 /**
- * Represents an Agent that can be tamed by a Tamer.
+ * Represents an Animal that can be tamed by a Tamer.
  */
-public interface Tameable extends Agent {
+public interface Tameable extends Animal {
```
gabizou: I made absolutely sure that it isn't necessary to have Tameable extend Animal. Please stop this nonsense of re-engineering the entity structure to match Minecraft's entity structure. There isn't a reason for any mod to add a creeper that is also a Tameable creeper. 

Otherwise I don't see where this entire PR is "urgent"

src/main/java/org/spongepowered/api/entity/living/Sittable.java
Mumfrey: It looks like whoever transposed that document from [the original](https://docs.google.com/document/d/1afZMMsU9yODX6d6WZAiKnabMbkuDXa_3ulsuDKZcfxg/edit#heading=h.kfel7xixr2pj) didn't include the statement that that rule is at the author's discretion (which is in agreement with the google style document on that rule). The page should probably be updated to reflect that.

gabizou:
Looking over this PR, I disagree with accepting this into the API under the circumstances, I don't see any proof of "implementation issues" with a `Tameable` extending `Agent` instead of `Animal`. If you wish to still propose this change in the API, please provide code proof of the issue and then I'll consider it. Otherwise, a specific implementation issue is NOT a valid reason to change the API.

AzureusNation:
Points below:
 * A Squid is NOT an ageable entity.
 * A Sitable Entity has to be tameable otherwise you cannot make that entity sit.
 * A Tameable Entity needs to be an animal as an Agent cannot be tamed.

@gabizou I would love to see how you would implement a squid that has an age. And I would like more then 1 persons opinion on this issue @Zidane

me4502:
I believe we once went over this issue, and decided that it was best to follow how it was done internally in the game.

gabizou:
> A Sitable Entity has to be tameable otherwise you cannot make that entity sit.

No changes needed for that....

> A Squid is NOT an ageable entity.

Ok, I'll give you that. But I don't see where this applies in the current discussion.


> A Tameable Entity needs to be an animal as an Agent cannot be tamed.

There is nothing in the contract of `Tameable` that requires `Animal` API's. Even in Minecraft source, there is absolutely no code in `EntityTameable` or any related AI classes that references `EntityAnimal`, only the methods from `EntityLiving` (which is our `Agent`), `Entity` (which, as people know is our `Entity`), and `EntityTameable` (which in our case is `Tameable`). Your point here is entirely invalid and only seeks to copy the *implementation* detail of Minecraft's entities, not even the ***interfaces*** the implementation implements.



> I would love to see how you would implement a squid that has an age.

Again, point is noted.

me4502:
Also, a lot of these changes do not make much sense, such as removing Animal from Wolf, Ocelot, Horse, etc.

gabizou:
> Also, a lot of these changes do not make much sense, such as removing Animal from Wolf, Ocelot, Horse, etc.

He didn't, it's just that he changed `Tameable` to extend `Animal` which those entities inherently already extend by extending `Tameable`.

AzureusNation:
It wont be long before the Sponge implementation devs start realising the issue aswell.

PR #396, opened 2015-01-19
Extracted interface Leashable.

ryantheleach:
Only golems and animals are leashable in vanilla, However other entities are still able to be leashed if forced.

I propose that an entity *other* then what is leashable in vanilla is an implementation specific detail, and should not be included in the API.

However, due to leashable being it's own interface, anyone that wishes to may attempt to cast an not listed as leashable in the API as leashable in order to test whether it happens to be leashable as an implementation detail. This would allow for mods/other server implementations to make anything leashable, whilst leaving room for the sponge mod to refuse to support entities which are currently leashable but not listed as such in the API at any time.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/Leashable.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Leashable.java
@@ -22,15 +22,44 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
+package org.spongepowered.api.entity.living;
 
-package org.spongepowered.api.entity.living.animal;
+import com.google.common.base.Optional;
+import org.spongepowered.api.entity.Entity;
 
-import org.spongepowered.api.entity.living.Ageable;
-import org.spongepowered.api.entity.living.Leashable;
+import javax.annotation.Nullable;
 
 /**
- * Represents an Animal, such as a Cow.
+ * Leashables are entities that can be interacted with by players wielding a {@link org.spongepowered.api.item.ItemTypes#LEAD}
  */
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/entity/living/Leashable.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package org.spongepowered.api.entity.living;
+
+import com.google.common.base.Optional;
+import org.spongepowered.api.entity.Entity;
+
+import javax.annotation.Nullable;
+
+/**
+ * Created by Ryan on 20/01/2015.
```
Minecrell: Don't add the author to the Javadocs (I guess it's automatically generated, you can turn it off), instead explain what this interface is.

Zidane:
Leashable should extend Living at the least.

me4502:
Through gameplay it is not possible to leash the other entities, however it is still possible to leash them using other methods. As all living creatures /can/ be leashed, I don't see why they shouldn't be Leashable.

ryantheleach:
I had tested leashing some entities to each other using commandblocks and the results were weird. not rendering correctly but still following the leash behaviour.

Zidane:
@SpongePowered/developers 

I'd like a consensus on this one.

ryantheleach:
@me4502 to me it's the distinction between saying that all living entities will always be leashable because it's in the API, vs saying that specific entities will always be leashable.

If someone created a LivingEntity that wasn't leashable, what would happen then? Or would it be by definition no longer a LivingEntity because it can't be leashed?

modwizcode:
I don't think LivingEntity is defined by the ability to be leashed, that
wouldn't make much sense, it's too narrow a trait for a class already
fulfilling a defined role (living entities).

PR #397, opened 2015-01-19
Fix Squid inherentance. Fix several OCD Issues.

gabizou:
This fixes all currently known OCD issues in #221 as well as fixing a few minor issues in the API related to Chunks, Horses, and Squid.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/animal/Horse.java
+++ b/src/main/java/org/spongepowered/api/entity/living/animal/Horse.java
@@ -103,4 +103,26 @@ public interface Horse extends Animal, Tameable {
      */
     void setSaddle(@Nullable ItemStack itemStack);
 
+    /**
+     * Gets this horse's temper.
+     *
+     * <p>Temper defines how close a horse can become tamed by
+     * {@link org.spongepowered.api.entity.Tamer}s. The higher the temper,
+     * the more likely a tamer will tame the horse.</p>
+     *
+     * @return This horses current temper
+     */
+    int getTemper();
+
+    /**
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/api/plugin/Plugin.java
```diff
@@ -41,23 +41,17 @@
     /**
      * An ID for the plugin to be used internally. The ID should be unique as to
      * not conflict with other plugins.
-     *
-     * @return The plugin identifier
```
Minecrell: If you remove this the JDK will start complaining, so it doesn't really fix this issue currently. That's a bug in checkstyle, the return statement should be there

src/main/java/org/spongepowered/api/entity/living/animal/Horse.java
```diff
@@ -103,4 +103,26 @@
...
+    /**
+     * Gets this horse's temper.
+     *
+     * <p>Temper defines how close a horse can become tamed by
+     * {@link org.spongepowered.api.entity.Tamer}s. The higher the temper,
+     * the more likely a tamer will tame the horse.</p>
+     *
+     * @return This horses current temper
+     */
+    int getTemper();
```
Zidane: To me it would seem like that a higher temper = angrier horse. Maybe a better name?

gabizou: Maybe tolerance?

Zidane: Sounds great to me.

PR #403, opened 2015-01-22
Add Attributes API

AlphaModder:
Solves #372. 
I wish I had a more detailed description, but I can't really think of anything to write. [This](http://minecraft.gamepedia.com/Attribute) page describes how attributes work. The only remaining issue right now is that i'm not sure of the best way to remove attributes from entities. Please discuss in the comments.


.gitignore
```diff
@@ -63,6 +63,7 @@ Thumbs.db
 .nb-gradle
 .idea
 .project
+.metadata
```
gabizou: Please remove from the PR.

src/main/java/org/spongepowered/api/attributes/Attribute.java
```diff
@@ -0,0 +1,36 @@
+package org.spongepowered.api.attributes;
+
+public interface Attribute {
+	
+	/**
+	 * Gets the ID of this attribute as a string.
+	 * @return The ID of this attribute as a string.
```
gabizou: Please insert a new line between the doc and the return. Applies to the rest of the javadocs.

.gitignore
AlphaModder: Whoops! On it.

src/main/java/org/spongepowered/api/attributes/Attribute.java
gabizou: Also, it is common in the API that a public interface does not denote the provided methods as `public`. So, for conformity, please remove the `public` modifier to the methods.

gabizou: Also, in `@param` and `@return` javadocs, do not end sentences with periods.

src/main/java/org/spongepowered/api/attributes/Attribute.java
```diff
@@ -0,0 +1,40 @@
+package org.spongepowered.api.attributes;
+
+public interface Attribute {
+
+    /**
+     * Gets the ID of this attribute as a string.
+     * 
+     * @return The ID of this attribute as a string.
```
gabizou: Please remove period at the end.

src/main/java/org/spongepowered/api/attributes/AttributeHolder.java
```diff
@@ -0,0 +1,80 @@
...
+     *         Optional.absent() if the attribute is not applicable.
+     */
+    Optional<Double> getAttributeValue(Attribute attribute);
+
+    /**
+     * Sets the base value of a certain {@link Attribute} on this holder.
+     * 
+     * @param attribute The Attribute to set the base value of.
+     * @param base The new base value.
+     * @throws IllegalArgumentException If the base value exceeds the maximum or
```
gabizou: Should mention this within the javadoc description, and possibly some link or reference to where to find the minimum/maximum allowed for the `Attribute`.

src/main/java/org/spongepowered/api/attributes/Attributes.java
```diff
@@ -0,0 +1,15 @@
+package org.spongepowered.api.attributes;
+
+public class Attributes {
```
gabizou: Might want to denote the class as `final` as well. Assuming this is a quasi enumeration class like `DyeColors` and such.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -48,12 +50,16 @@
     /**
      * Returns the health amount.
      *
-     * <p>The range of the health depends on the object on which this
```
gabizou: Undo these changes please.

src/main/java/org/spongepowered/api/attributes/Attributes.java
gabizou: You also might want to add to `GameRegistry` to retrieve `Optional<Attribute> getAttributeByName(String)` and `List<Attribute> getAvailableAttributes()` or something similar to what is available in the registry.

src/main/java/org/spongepowered/api/entity/living/Living.java
AlphaModder: Ah, crap, I thought I had my formatter setup properly.

src/main/java/org/spongepowered/api/attributes/Attribute.java
```diff
@@ -0,0 +1,64 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+public interface Attribute {
+
+    /**
+     * Gets the ID of this attribute as a string
```
gabizou: The Description part of the javadoc DOES require full punctuation. The parameters/returns/throws do not.

src/main/java/org/spongepowered/api/attributes/AttributeHolder.java
```diff
@@ -0,0 +1,104 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+import com.google.common.base.Optional;
+
+import java.util.Collection;
+import java.util.List;
+
+public interface AttributeHolder {
```
gabizou: Need class javadocs.

src/main/java/org/spongepowered/api/attributes/Attribute.java
```diff
@@ -0,0 +1,64 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+public interface Attribute {
```
gabizou: Need class javadocs. In the javadocs, please make sure to fully describe what an `Attribute` is and it's relation to an `AttributeModifier`.

gabizou:
Missing a `package-info.java` for the `org.spongepowered.api.attributes` package.

src/main/java/org/spongepowered/api/attributes/Attribute.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Gets the ID of this attribute as a string.
+     * 
+     * @return The ID of this attribute as a string
+     */
+    String getStringId();
```
simon816: Just `getId()` will do here

src/main/java/org/spongepowered/api/attributes/AttributeHolder.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Calculates the value of a certain {@link Attribute} on this holder.
+     * 
+     * @param attribute The Attribute to find the value of
+     * @return The value of a certain Attribute on this holder, or
+     *         Optionalabsent() if the attribute is not applicable
```
simon816: `Optional.absent()`

src/main/java/org/spongepowered/api/attributes/Attributes.java
```diff
@@ -0,0 +1,39 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+public final class Attributes {
+
```
simon816: Needs private constructor.

src/main/java/org/spongepowered/api/attributes/AttributeTargets.java
```diff
@@ -0,0 +1,34 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+public final class AttributeTargets {
+
```
simon816: Needs private constructor.

src/main/java/org/spongepowered/api/attributes/Operations.java
```diff
@@ -0,0 +1,33 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attributes;
+
+public final class Operations {
+
```
simon816: Needs private constructor.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -346,7 +353,8 @@
      * Gets the villager {@link Profession} with the specified id.
      *
      * @param id The id of the profession to return
-     * @return The profession with the given id or Optional.absent() if not found
+     * @return The profession with the given id or Optional.absent() if not
+     *         found
```
simon816: The formatting changes, although correct, probably shouldn't be done in this PR

AlphaModder: Yeah... It's just that when I save GameRegistry, my formatter wants to fix all the line lengths... I'm looking for a good solution.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -69,17 +72,13 @@
 /**
  * Provides an easy way to retrieve types from a {@link Game}.
  *
- * <p>Note that the registries may be in flux, especially during game
- * initialization. These will be accurate for the time they are called, however
- * they may change at a later point. Do not assume that the contents of a list
- * will be all the entries that will exist.</p>
+ * <p>Note that the registries may be in flux, especially during game initialization. These will be accurate for the time they are called, however
```
simon816: Don't reformat this. Line length for Javadoc is 80 anyway.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -536,4 +535,26 @@
...
+     * 
+     * @param id The id to find
+     * @return An Operation if one could be found, otherwise Optional.absent()
+     */
+    Optional<Operation> getOperation(int id);
+
+    /**
+     * Gets a new {@link AttributeModifierBuilder}.
+     * 
+     * @return A new AttributeModifierBuilder.
```
simon816: Remove period at EOL

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -536,4 +535,26 @@
...
+     */
+    AttributeModifierBuilder getAttributeModifierBuilder();
+}
```
simon816: Needs `\n` newline at EOF

simon816:
Please correctly wrap javadoc comments, they should not exceed 80 characters per line

src/main/java/org/spongepowered/api/attribute/AttributeHolder.java
```diff
@@ -0,0 +1,97 @@
...
+     * @return The base value of a certain Attribute on this holder
+     */
+    double getBase(Attribute attribute);
+
+    /**
+     * Gets a list of all applied modifiers.
+     * 
+     * @return A list of all applied modifiers
+     */
+    List<AttributeModifier> getAllModifiers();
```
simon816: Could this be a `Collection`?

src/main/java/org/spongepowered/api/attribute/AttributeTarget.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attribute;
+
+public interface AttributeTarget {
+
+    /**
+     * If this attribute target can apply to the specified.
```
simon816: Period in wrong place (should be after the `@link` in this sentence)

src/main/java/org/spongepowered/api/attribute/AttributeTarget.java
```diff
@@ -0,0 +1,39 @@
...
+     * {@link AttributeHolder}
+     * 
+     * @param holder The AttributeHolder to check with
+     * @return If this attribute target can apply to the specified
+     *         AttributeHolder
```
simon816: Should either `@link AttributeHolder` or just call it `holder`

src/main/java/org/spongepowered/api/attribute/Operation.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attribute;
+
+public interface Operation {
+
+    /**
+     * Returns the value of an Attribute after it has been operated upon.
```
simon816: `Attribute` should be `@link`ed or lower-case.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -287,4 +281,4 @@
      * @param invisible Whether this entity is invisible or not
      */
     void setInvisible(boolean invisible);
-}
+}
```
simon816: Don't remove the newline at EOF

src/main/java/org/spongepowered/api/event/attribute/AttributeBaseChangedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.event.attribute;
+
+public interface AttributeBaseChangedEvent extends AttributeEvent {
+
+    /**
+     * Gets the old value of the changed
+     * {@link org.spongepowered.api.attribute.Attribute}.
```
simon816: Could import here to avoid fully qualifying.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -130,4 +132,22 @@
      * @return The level of the enchantment, or -1 if the enchantment is not applied to this stack
      */
     int getEnchantment(Enchantment enchant);
+    
+    /**
+     * Gets all {@link AttributeModifier}s on this ItemStack.
+     * @return All AttributeModifiers on this ItemStack.
```
simon816: Remove periods from EOL on `@return` and `@param`. (Also for below methods too)

gratimax:
:+1: Pinging @SpongePowered/developers for final review

Kobata:
I'm not sure if it was mentioned somewhere in the outdated comments, but having AttributeModifier.getName() and .getUUID() would be useful

AlphaModder:
@Kobata I am unsure of the usefulness of that, because no name is stored and the UUIDs are inconsistent at best:
"...Note that some UUIDs are generated on-the-fly, while others are represented as fixed strings in the code..."

Kobata:
Name is a description and somewhat useful for that (although only really for developers).

UUID is 'identity' and is important as a new modifier added with the same UUID will replace any existing applied modifier with the same one -- most of them are fixed (e.g., CB3F55D3-645C-4F38-A497-9C13A33DB5CF is always added damage from active item, and has different names depending on the item), unless it's a completely dynamic modifier that should apply along with all others.

AlphaModder:
@Kobata while that's true, I don't see a need to expose it to plugin developers, and in a way it seems like an implementation detail.

me4502:
I agree with getName being added.

Imagine the case of printing "Added ATTRIBUTE_NAME to player" to the console.

AlphaModder:
@me4502 I see the reasoning behind that. I'll add it soon.

Soren025:
I don't know much about attributes in the base game, but is there not a way for plugins to create their own attributes?

AlphaModder:
@Soren025 It seems, inspecting the attribute code that this would mean modification to the MC entity classes, otherwise it would only be useful for new entities, which plugins can't make anyway.

ST-DDT:
@Soren025 Creating new attributes, does only make sense, if you deal with it all yourself, either in a new entity implementation or have customizations in events based on those attributes.

For example listen to `EntityDamageEvent`s and spawn reinforcements based on the attribute value `GENERIC_REINFORCEMENTS`, but doing this via Attributes is only useful if you also have one or two `PotionEffectType`s that modify the attribute value or that value is persisted.

Another question of mine: Are `Attribute`s (base values) persistent?

> it would only be useful for new entities, which plugins can't make anyway.

Not yet, it is planned for the next big MC release. (1.9) Although i don't know anything about its capabilities, because its totally defined via NBT.

AlphaModder:
@ST-DDT  Yes, they are persistent. And since we'll have some other way to attach tags to entites, we don't really need custom attributes.

AlphaModder:
Wow, I accidentally hit the wrong button.

Soren025:
@ST-DDT @AlphaModder 
> or have customizations in events based on those attributes.

This exactly.

I was hoping to be able to take advantage of the ```ItemStack``` attribute modifiers with attributes created such as "INT" "STR" "DEX" and what not. It would make things a lot simpler for rpg plugins.

AlphaModder:
@Soren025 That would be better served using a metadata system for ItemStack and Entity than making "fake" minecraft Attributes, I think.

AlphaModder:
@Soren025 Especially since attribute display logic is client-side, custom attributes wouldn't display as "+5 Intelligence" just by being there.

Soren025:
@AlphaModder I see that's unfortunate

ST-DDT:
@AlphaModder But that could be covered by a Sponge client plugin easily i guess. It is planned anyway, so we should probably add it to the feature request list.
https://github.com/SpongePowered/SpongeAPI/issues/429

AlphaModder:
@ST-DDT While that's true, the Client API is supposed to be optional, so custom attributes shouldn't be part of this PR, Though they could be part of it.

src/main/java/org/spongepowered/api/attribute/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+@org.spongepowered.api.util.annotation.NonnullByDefault
+package org.spongepowered.api.attribute;
+
```
gratimax: extra newline

src/main/java/org/spongepowered/api/event/attribute/AttributeBaseChangedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.event.attribute;
+
+import org.spongepowered.api.attribute.Attribute;
+
+public interface AttributeBaseChangedEvent extends AttributeEvent {
```
gratimax: Missing javadocs

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -22,21 +22,22 @@
...
+import com.flowpowered.math.vector.Vector3f;
+import com.google.common.base.Optional;
+
 import java.util.Collection;
 import java.util.List;
 
 import javax.annotation.Nullable;
 
-public interface Living extends Entity {
+public interface Living extends Entity, AttributeHolder {
```
gratimax: I realize now that there are missing javadocs here, but don't fix it

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -287,4 +288,5 @@
      * @param invisible Whether this entity is invisible or not
      */
     void setInvisible(boolean invisible);
+
```
gratimax: Formatting fix

ST-DDT:
@AlphaModder That's true, in the entire API there is no possibility to add custom stuff yet. This is most likely the case until the Inventory API is finished. But you are right adding custom stuff is not/should not be part of this PR.

src/main/java/org/spongepowered/api/event/attribute/AttributeEvent.java
```diff
@@ -0,0 +1,49 @@
...
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.event.attribute;
+
+import org.spongepowered.api.attribute.Attribute;
+import org.spongepowered.api.attribute.AttributeHolder;
+import org.spongepowered.api.util.event.Cancellable;
+
+public interface AttributeEvent extends Cancellable {
```
gratimax: Missing javadocs

src/main/java/org/spongepowered/api/event/attribute/AttributeEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Gets the {@link AttributeHolder} who's attributes are being changed.
```
gratimax: `whose`

src/main/java/org/spongepowered/api/event/attribute/AttributeModifiedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.event.attribute;
+
+import org.spongepowered.api.attribute.AttributeModifier;
+
+public interface AttributeModifiedEvent extends AttributeEvent {
```
gratimax: Missing javadocs

src/main/java/org/spongepowered/api/event/attribute/AttributeModifiedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return The AttributeModifier that was added or removed in this event
+     */
+    AttributeModifier getModifier();
+
+    /**
+     * Gets if this event represent the removal of the specified modifier.
+     * 
+     * @return If this event represent the removal of the specified modifier
+     */
+    boolean isRemoval();
```
gratimax: I feel like this would function better as 2 separate events.

src/main/java/org/spongepowered/api/attribute/AttributeTarget.java
```diff
@@ -0,0 +1,43 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attribute;
+
+/**
+ * Represents a function to find out if an {@link AttributeHolder} can have a
+ * certain {@link Attribute} applied to it.
+ */
+public interface AttributeTarget {
```
gratimax: Unecessary, use `Predicate<AttributeHolder>` from guava.

src/main/java/org/spongepowered/api/attribute/AttributeModifier.java
```diff
@@ -0,0 +1,75 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.attribute;
+
+/**
+ * A modifier that changes the value of an {@link Attribute} on an
+ * {@link AttributeHolder}.
+ */
+public interface AttributeModifier {
```
gratimax: Make this interface immutable, e.g. remove `setValue`, `setAttribute`, and `setOperation`. That's why we have builders.

src/main/java/org/spongepowered/api/attribute/AttributeHolder.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Adds multiple {@link AttributeModifier} to this holder.
+     * 
+     * @param modifiers The AttributeModifiers to add
+     * @throws IllegalArgumentException If any of the AttributeModifiers'
+     *         {@link Attribute}s' {@link AttributeTarget}s' cannot apply to
+     *         this holder
+     */
+    void addModifiers(Collection<AttributeModifier> modifiers) throws IllegalArgumentException;
```
gratimax: We don't even need a Collection here, just an Iterable.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -536,4 +539,26 @@
...
+     * @param name The name to find
+     * @return An Attribute if one could be found, otherwise Optional.absent()
+     */
+    Optional<Attribute> getAttribute(String name);
+
+    /**
+     * Gets an {@link Operation} by id.
+     * 
+     * @param id The id to find
+     * @return An Operation if one could be found, otherwise Optional.absent()
```
gratimax: `{@link Optional#absent()}`

src/main/java/org/spongepowered/api/event/attribute/AttributeModifiedEvent.java
ST-DDT: And there should probably be helper methods in the `SpongeEventFactory`.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -536,4 +539,26 @@
      */
     List<BannerPatternShape> getBannerPatternShapes();
 
+    /**
+     * Gets an {@link Attribute} by name.
+     * 
+     * @param name The name to find
+     * @return An Attribute if one could be found, otherwise Optional.absent()
```
gratimax: `{@link Optional#absent()}`

src/main/java/org/spongepowered/api/event/attribute/AttributeModifiedEvent.java
AlphaModder: @gratimax I don't see why, they contain the exact same info.

src/main/java/org/spongepowered/api/GameRegistry.java
ST-DDT: Everywhere else its just `Optional.absent()`.

gratimax:
Attributes can have multiple operations, see http://minecraft.gamepedia.com/Attribute#Vanilla_Modifiers.

src/main/java/org/spongepowered/api/attribute/Attribute.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return The default value for this attribute
+     */
+    double getDefaultBase();
+
+    /**
+     * Gets the target of this attribute.
+     * 
+     * @return The target of this attribute
+     */
+    AttributeTarget getTarget();
```
gratimax: On second thought, even a Predicate is not needed. Use EntityType

src/main/java/org/spongepowered/api/GameRegistry.java
gratimax: @ST-DDT Absolutely not. If it is somewhere, that should be fixed

ST-DDT:
@gratimax I can't see anything in the API that prevents that.
The multiplicity is covered by using multiple `AttributeModifier`s.

src/main/java/org/spongepowered/api/attribute/Attribute.java
AlphaModder: I disagree. Attributes could be on non-entity things with some mods, and there could be attributes that are more specific, say wither vs nonwither skeles or certain sizes of Slime.

ST-DDT: `AttributeTargets.GENERIC` is all, but there is no such `EntityType.ALL` so it should be a `Predicate` or `AttributeTarget`.

src/main/java/org/spongepowered/api/event/attribute/AttributeModifiedEvent.java
gratimax: @AlphaModder yes, but conceptually something different is happening for each event.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -130,4 +132,22 @@
      * @return The level of the enchantment, or -1 if the enchantment is not applied to this stack
      */
     int getEnchantment(Enchantment enchant);
+
```
gratimax: Why doesn't this extend AttributeHolder?

gratimax:
@ST-DDT Our goal is to model the game, and the game simply doesn't work like that. See "Leader zombie bonus".

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
AlphaModder: AttributeHolder is an object with base value that are modified. An ItemStack has no such base values, only modifiers.

gratimax: Then set up some sort of hierarchy. e.g. AttributeHolder extends some interface and adds base values.

AlphaModder:
@gratimax I'll refactor it to allow multiple thingies.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
ST-DDT: `AttributeHolder` is the target where attributes apply to.
`ItemStack` is more a source of attribute modifiers. Maybe a new interface `AttributeSource` is a good addition, this could be used in the events as well.

Soren025:
@AlphaModder 

> Attribute modifiers can have multiple operations, see http://minecraft.gamepedia.com/Attribute#Vanilla_Modifiers.

Wait the link given @gratimax shows a modifier labeled as "Unknown" at the bottom stating that its a modifier sent by the server that is unknown. So custom attributes are possible server side even if the client doesn't see it on the item stack?

gratimax:
@Soren025 Yes.

ST-DDT:
@gratimax Look at minecrafts implementation, they have an own `AttributeModifier` for each attribute they modify. So the current implementation is totally correct.

gratimax:
@ST-DDT Hmm, fair point. Then we probably want to revert 2390a61749

AlphaModder:
@gratimax Alright, reverting.

AlphaModder:
@gratimax I'm afraid i've never used git to revert before, how would I?

gratimax:
```
git revert --no-commit 2b2cc7
git revert --no-commit 2390a61
git commit -m "Revert adding AttributeModifierEntry"
```

ST-DDT:
Here is an example implementation of the `AttributeModifications` class that reflects the curremt Minecraft implementation (using the old system).

````java
public final class AttributeModifications {

public static final List<AttributeModifier> ZOMBIE_LEADER_BONUS = null;
public static final List<AttributeModifier> ZOMBIE_RANDOM_SPAWN_BONUS = null;

static {
    ZOMBIE_LEADER_BONUS = new ArrayList<>();
    ZOMBIE_LEADER_BONUS.add(new AttributeModifier(Attributes.MAXHEALTH, Operations.ADD, 42);
    ZOMBIE_LEADER_BONUS.add(new AttributeModifier(Attributes.ZOMBIE_REINFORMCENTS, Operations.MULTIPLY, 42);
}

}
````

File: src/main/java/org/spongepowered/api/attribute/Attribute.java [hunk 41]
```diff
@@ -0,0 +1,79 @@
...
+     * @return The default value for this attribute
+     */
+    double getDefaultBase();
+
+    /**
+     * Gets the target of this attribute.
+     * 
+     * @return The target of this attribute
+     */
+    Predicate<AttributeHolder> getTarget();
```
gratimax: I feel like `target` implies that there's only one holder this applies to. Maybe `getTargets`?

AlphaModder:
Alright, it's rolled back. I just used git reset --hard HEAD~2, sorry about no commit message.

gratimax:
@ST-DDT While that's nice, I don't think I'd bother. The only usage for that is adding multiple attribute modifiers, trying to detect if `ZOMBIE_LEADER_BONUS` is actually applied is a different story.

src/main/java/org/spongepowered/api/attribute/Attribute.java
AlphaModder: Yeah.

gratimax: Actually, hold that thought. It's not really `targets` either, it's more like `getFilter` or `getPredicate`

ST-DDT: `Targets` is better than `Target`, but look at what this method actualy does.
The target is not really a targets its more an `isApplicable(AttributeHolder)`, and then @gratimax would be right that it can be checked for `EntityType`.

So i suggest changing it to `boolean isApplicable(AttributeHolder)` or `boolean isApplicable(EntityType)`.

ST-DDT:
@gratimax The reason why a `AttributeModifications`class would be cool is not the removal of said attribute from any entity, it's the addition.

````java
entity.addAttributeModifiers(AttributeModifications.ZOMBIE_LEADER_BONUS);
````
But you are right, its a nice to have, but Optional

File: src/main/java/org/spongepowered/api/attribute/AttributeHolder.java [hunk 42]
```diff
@@ -0,0 +1,106 @@
...
+    Collection<AttributeModifier> getAllModifiers();
+
+    /**
+     * Adds a {@link AttributeModifier} to this holder.
+     * 
+     * @param modifier The AttributeModifier to add
+     * @throws IllegalArgumentException If the AttributeModifier's
+     *         {@link Attribute}'s trget cannot apply to this holder
+     */
+    void addModifier(AttributeModifier modifier) throws IllegalArgumentException;
```
ST-DDT: typo `trget` -> `target`
Please state the behaviour when adding it multiple times.
Minecraft throws an `IllegalArgumentException("Modifier is already applied on this attribute!")` in that case.

In addition to that it should be be stated in AttributeModifier class's javadocs, that `AttributeModifier`s can be used multiple times, for the same purpose. Otherwise plugin authors will create a new instance each time, they want to use it and will fail to remove it because they cannot remove the old modifier with the new one.

ST-DDT:
@AlphaModder What about extracting `AttributeSource` or `AttributeModifierSource`?
`PotionEffectType` also uses `AttributeModifier`s so it could extend that interface too.

Notice: `PotionEffectType` is called `Potion` in MCP.

AlphaModder:
@ST-DDT  Could you explain a bit more?

ST-DDT:
@AlphaModder Sure

````java
[..]
public interface AttributeSource {

    /**
     * Gets all {@link AttributeModifier}s on this ItemStack.
     *
     * @return All AttributeModifiers on this ItemStack
     */
    Collection<AttributeModifier> getAttributeModifiers();

    /**
     * Adds a {@link AttributeModifier} to this ItemStack.
     *
     * @param modifier The AttributeModifier to add to this ItemStack
     */
    void addAttributeModifier(AttributeModifier modifier);

    /**
     * Removes a {@link AttributeModifier} from this ItemStack.
     *
     * @param modifier The AttributeModifier to remove from this ItemStack
     */
    void removeAttributeModifier(AttributeModifier modifier);

}
````

[ItemStack](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java)
````java
[..]
public interface ItemStack extends Serializable, DataSerializable, AttributeSource {
[..]
````

[PotionEffectType](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/potion/PotionEffectType.java)
````java
[..]
public interface PotionEffectType extends AttributeSource {
[..]
````

Which will grant access to: 
net.minecraft.potion.Potion.attributeModifierMap (line 86 from MCP 1.8.0)

But i'm not sure whether we should grant more than read access to `PotionEffectType` `AttributeModifier`s.

I haven't found any up to date class for that so i reference a file in Bukkit:
[Potion](https://github.com/Bukkit/mc-dev/blob/c1627dc9cc7505581993eb0fa15597cb36e94244/net/minecraft/server/MobEffectList.java#L14)
The last method call in that  line is the addition of the AttributeModifier.

**Comment by AlphaModder** [comment]
@ST-DDT Got it. It'll be in the next commit.

**Comment by ST-DDT** [comment]
@AlphaModder Yeah, that's exactly what i wanted. Thanks a lot for your work.

**Comment by AlphaModder** [comment]
@ST-DDT  No problem. I hope you think the mutable distinction was a good idea.

**Comment by ST-DDT** [comment]
@AlphaModder Yes, it is a good idea. I thought of it myself, but i wasn't sure how to name the read only variant. But your choice is very good here. In some other PR there was the discussion with MutableX extends ImmutableX. But i'm glad you avoided that issue.

There is basically 2 or 3 points left from my point of view.
- Add helper methods to `SpongeEventFactory`
- Maybe add `Optional<AttributeSource> getSource()` to `AttributeModifiedEvent` to get the PotionEffectType or ItemStack that causes the attribute change.
- Seperate `AddAttributeModifiedEvent` from `RemoveAttributeModifiedEvent` 
  - You have already commented on this and stated you opinion, so we don't have to discuss it over again

I'm not a Teammember, so you don't have to change anything if you don't want to.

File: src/main/java/org/spongepowered/api/attribute/AttributeTargets.java [hunk 43]
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * An enumeration of all attribute targets present in vanilla Minecraft.
+ */
+public final class AttributeTargets {
+
+    public static final Predicate<AttributeHolder> GENERIC = null;
+    public static final Predicate<AttributeHolder> HORSE = null;
+    public static final Predicate<AttributeHolder> ZOMBIE = null;
+    public static final Predicate<AttributeHolder> UNKNOWN = null;
```
ST-DDT: I guess `UNKNOWN` doesn't make any sense anymore.

Maybe the entire `AttributeTargets` class is obsolete now, but thats up to you.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -536,4 +539,28 @@
...
+    Optional<Attribute> getAttribute(String name);
+
+    /**
+     * Gets an {@link Operation} by id.
+     * 
+     * @param id The id to find
+     * @return An Operation if one could be found, otherwise
+     *         {@link Optional#absent()}.
+     */
+    Optional<Operation> getOperation(int id);
```
me4502: Why exactly is this using numerical IDs?

AlphaModder: Because that's how they are registered in vanilla.

progwml6: even though vanilla uses int ID's on alot of things under no circumstances will we be exposing integer formatted ID's

AlphaModder: @progwml6 Then what other way should I do this?

ST-DDT: @AlphaModder Maybe a String name similar to an enum name. This would be abstract enough to be useful for storing in a config as well. But wait for an official decision here.

me4502:
@SpongePowered/developers - does everyone feel this is ready for merging?

src/main/java/org/spongepowered/api/attribute/Attribute.java
```diff
@@ -0,0 +1,81 @@
...
+     * @return The maximum value for this attribute
+     */
+    double getMaximum();
+
+    /**
+     * Gets the default value for this attribute.
+     *
+     * @return The default value for this attribute
+     */
+    double getDefaultBase();
```
modwizcode: Why isn't this called `getDefaultValue` or `getDefault()`, even `getBaseValue()` would be better. Otherwise base is duplicated and sounds more like a numerical base than a base value.

AlphaModder: Good point. I'll fix that.

src/main/java/org/spongepowered/api/attribute/AttributeHolder.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    void setBase(Attribute attribute, double base) throws IllegalArgumentException;
+
+    /**
+     * Gets the base value of a certain {@link Attribute} on this holder.
+     *
+     * @param attribute The attribute to get the base value of
+     * @return The base value of a certain Attribute on this holder
+     */
+    double getBase(Attribute attribute);
```
Minecrell: Put the getter before the setter, will just confuse otherwise

gabizou:
I was hoping that Movement speed would be represented by Attributes. Can we consider adding those?

AlphaModder:
@gabizou Is it not? Attributes.GENERIC_MOVEMENT_SPEED?

gabizou:
> Is it not? Attributes.GENERIC_MOVEMENT_SPEED?

Does that apply for both fly speed and walking speed?

AlphaModder:
@gabizou I'm not sure, but it's the only movement attribute.

gabizou:
@AlphaModder Well, considering there is a value stored for both flight and walk speed, I'd want an attribute for both.

AlphaModder:
@gabizou I'm not creating these attributes, these are the ones present in Vanilla MC. As stated above, creation of new Attributes is out of the scope of this PR.

gabizou:
> I'm not creating these attributes, these are the ones present in Vanilla MC. As stated above, creation of new Attributes is out of the scope of this PR.

I'll allow a few additional Attributes to be added to the API for this case. There isn't a reason we can't emulate the attributes for player movement and flight speed. So, again, please add these two attributes.

gabizou:
@AlphaModder Can you add the following:

For Humans:
 - Food Level
 - Saturation
 - Exhaustion

Reasoning is that they too are attributes of the Human entity, much like movement speeds.

AlphaModder:
@gabizou That's gonna cause problems with AttributeModifiers, unless sponge actually adds these attributes, which would cause client issues.

src/main/java/org/spongepowered/api/block/data/CommandBlock.java
```diff
@@ -24,14 +24,15 @@
...
-import org.spongepowered.api.util.command.source.CommandBlockSource;
+import org.spongepowered.api.util.command.CommandSource;
+
+import com.google.common.base.Optional;
 
 /**
  * Represents a Command Block.
  */
-public interface CommandBlock extends TileEntity, CommandBlockSource {
+public interface CommandBlock extends TileEntity, CommandSource {
```
Deamon5550: Undo this change please.

AlphaModder: Whoops... Did the name of CommandBlockSource change? For some reason I thought it changed to CommandSource, but that's not right.

gabizou:
> That's gonna cause problems with AttributeModifiers

Why would it cause problems?

Granted these aren't actual attributes represented by the implementation of Minecraft, but by all means, they can be represented as attributes for API reasons. The implementation is a minor detail that overall, the API has been avoiding to directly represent, only, if the API is designed to mirror the implementation of Minecraft, well, we'd have more issues with other parts of the API than something as simple as implementing hunger mechanics as an attribute.

> which would cause client issues.

Incorrect. Clients happily accept custom attributes that aren't he default ones. Besides, the actual implementation of the hunger mechanics could easily become a component in the future (of which Attributes would eventually have the capability of evolving into).

> unless sponge actually adds these attributes

We very well could, though I wouldn't imagine the Forge implementation would actually use them as attributes and just bypass attributes and represent them as proxies to the actual field values stored in the entities themselves.

AlphaModder:
@gabizou If you say so.

Deamon5550:
Another thing, it would be good to support plugin added custom attributes. There is nothing stopping plugin added attributes as we can persist them with the rest of sponge's extraneous data, and they'll even be available on the client for client mods to access.

AlphaModder:
@Deamon5550 Custom Attributes were discussed earlier in this discussion, and we decided they were out of scope for this PR, and maybe Client API exclusive.

Deamon5550:
> out of scope for this PR

This is an attributes API PR, they are not out of scope, all you need is a simple builder and an `addAttributes` method
> Client API exclusive

No reason for this, attributes on the server are propagated to the client and the client will store any attributes that it receives, even if it doesn't recognize them.
> it would only be useful for new entities

Not true, attributes added to existing entities can still be used by plugins in place of needing to store their own custom data.

Kobata:
@gabizou -- all your suggestions here are massively out of scope of what an 'attribute' in the context of minecraft means.

Kobata:
@gabizou For more clarity: The API's job is to expose what Minecraft does in a fairly easy to use manner, preferablly in a way that makes it not too bad for alternate servers to use. Its job is not to randomly expand concepts just because you think it looks cleaner. For things like this PR, that correspond directly to things vanilla does that are well exposed to users of vanilla (in this case, itis heavily documented for mapmaker use), we should keep that correspondence as close as possible.

gabizou:
> Not true, attributes added to existing entities can still be used by plugins in place of needing to store their own custom data.

Though, we still have to resolve a better solution with Persistence API for storing custom data ;).

> all your suggestions here are massively out of scope of what an 'attribute' in the context of minecraft means.

They're not out of scope... We're expanding the definition of Attributes as we are the writers of the overall API. Given that we've been dead set on avoiding to design our API around the implementation of Minecraft, we've found that for Attributes, it's most definitely in the interest of the API to design it in such a way that can evolve for the future. Also, it has been in our interest in the past, and still is currently, to eventually move towards a component system for entities, of which Attributes is the prime candidate to facilitate that evolution.

> The API's job is to expose what Minecraft does in a fairly easy to use manner, preferablly in a way that makes it not too bad for alternate servers to use.

Yes, and that is what I'm aiming to ensure this PR does. The Attributes API does not need to remain stuck as a solidified implementation detail with some blueprint API religiously abiding by the vanilla implementation.

> Its job is not to randomly expand concepts just because you think it looks cleaner.

The API isn't at all just expanding concepts because it looks cleaner. It's simply adding more attributes based on the current design.

> For things like this PR, that correspond directly to things vanilla does that are well exposed to users of vanilla (in this case, itis heavily documented for mapmaker use), we should keep that correspondence as close as possible.

Sorry, but I do not see where we are breaking the contract of the defined usage of Attributes, merely we're expanding on the possible usages on what are stored in Attributes. Examples of this expansion include the Inventory API. It's an entirely different setup compared to what vanilla Minecraft/Bukkit has/had and yet facilitates easier usage of the API.

Kobata:
For the record I don't really like the inventory API concept. (I could probably go on and name half the stuff in the API that I don't really like)

sk89q:
I think we agree on that the API is a balance of several goals:

* Something that is greatly expands the ability to modify MC in as many ways as possible
* Something that is consistent and portable between MC versions
* Something that "fits together" in the grand scheme of the entire API
* Something that is within a scope that we, the team, are able to maintain between MC versions (time and effort-wise)

In my opinion, I think we should provide the ability to register new attributes, but I disagree with making some currently non-attributes "virtual" attributes because attributes are a "public-facing" feature that people can use when making maps.

me4502:
See ^, this guy makes sense.

modwizcode:
Agreed, virtual attributes are not the way to do this. A "property" API
might offer a similarly clean representation that isn't muddled with
Minecraft's definition of an attribute.

AlphaModder:
@modwizcode I was thinking there'd be something in the Persistence API, akin to bukkit's metadata, that would be better used for tagging entities.

modwizcode:
Yeah that's the same thing I was thinking.

gabizou:
@AlphaModder Please do the following:

- Remove virtual attributes (They can be discussed and decided upon in the future)
- Add customized attribute registration for Plugins
- Fix merge conflicts.

AlphaModder:
@gabizou Are you sure the second is necessary? As I said before, It seems like plugins should utilize the Persistence API to save data to entities and items, rather than attributes.

modwizcode:
@gabizou I'm with @AlphaModder on this, it doesn't make much sense to put
custom support anywhere near Attributes.

Zidane:
I'm not seeing an issue with custom attributes. Virtual ones I see the point about.

modwizcode:
@gabizou What's the point of custom attributes?

gabizou:
As attributes are implemented in Minecraft, they are various `double` values assigned to entities that can be modified by something as simple as a held item or equipped item that has an attribute modifier for that specific attribute.

Given that this behavior is entirely supported within Minecraft by default, it would make sense to do the following:
- Create a loosely based API around the implementation of Attributes as they are in Minecraft
- Allow the same freedoms Minecraft grants by having custom attributes created as needed

Given that the first is already done in this PR, the second should most definitely be allowed for the same reason that map makers are able to create their own [AttributeModifiers](http://minecraft.gamepedia.com/Attributes#Vanilla_Modifiers). I already have tested custom attributes being registered with the game where the game doesn't crash. Being that custom attributes would allow storage of various bits of information, modifiers are immensely beneficial to applying a change to an attribute without having to even listen for the change of equipment or anything else for that matter.

I'm failing to see where it's so hard to understand why custom attributes and modifiers is something that should be allowed in our API to begin with. I mean, isn't the whole point of our API to allow integration and customization with modding?

src/main/java/org/spongepowered/api/attribute/AttributeTargets.java
```diff
@@ -0,0 +1,42 @@
...
+package org.spongepowered.api.attribute;
+
+import com.google.common.base.Predicate;
+
+/**
+ * An enumeration of all attribute targets present in vanilla Minecraft.
+ */
+public final class AttributeTargets {
+
+    public static final Predicate<AttributeHolder> GENERIC = null;
```
gabizou: What does this mean exactly?

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -25,13 +25,15 @@
...
-import com.flowpowered.math.vector.Vector3f;
-import com.google.common.base.Optional;
+import org.spongepowered.api.attribute.AttributeHolder;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.player.Player;
 import org.spongepowered.api.potion.PotionEffect;
 import org.spongepowered.api.potion.PotionEffectType;
 
+import com.flowpowered.math.vector.Vector3f;
+import com.google.common.base.Optional;
```
gabizou: wrong import order.

src/main/java/org/spongepowered/api/potion/PotionEffectType.java
```diff
@@ -25,10 +25,12 @@
 
 package org.spongepowered.api.potion;
 
+import org.spongepowered.api.attribute.AttributeSource;
+
 /**
  * Represents a possible type of {@link PotionEffect}.
  */
-public interface PotionEffectType {
+public interface PotionEffectType extends AttributeSource {
```
gabizou: Why is a PotionType an AttributeSource?

modwizcode: I see now, I didn't realize exactly how attributes are handled internally.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -59,6 +61,8 @@
 import org.spongepowered.api.world.biome.BiomeType;
 import org.spongepowered.api.world.difficulty.Difficulty;
 
+import com.google.common.base.Optional;
```
Minecrell: Same here, you broke the import order :/

src/main/java/org/spongepowered/api/attribute/AttributeTargets.java
AlphaModder: @gabizou It's a Predicate to tell if a given AttributeHolder matches the target.

gabizou:
@AlphaModder You have some decent PR's going on, but in every single one of them, you have formatting issues. I'd appreciate some courtesy in fixing your auto formatter since we do provide the formatting styles for both [IntelliJ](https://github.com/SpongePowered/SpongeAPI/blob/master/extra/intellij/sponge_intellij_style.xml) and [Eclipse](https://github.com/SpongePowered/SpongeAPI/tree/master/extra/eclipse).

I'd want to handle merging some PRs but I simply can't get over all the formatting issues. 

### FIX YOUR FORMATTING

AlphaModder:
@gabizou I have been working on the formatting. If you could point out what I've missed, that would help. I don't know how that happens though considering I'm using the sponge style, cleanup, and import order.

gabizou:
Oddly, when I made the comment, the files were showing up incorrectly formatted still. Excuse my rudeness. I'll start reviewing this PR for merging.

gabizou:
@AlphaModder If you can, add an `AttributeBuilder` for plugin created attributes. After which, we should be all ready to go.

src/main/java/org/spongepowered/api/attribute/AttributeModifierBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return This builder, for chaining
+     */
+    AttributeModifierBuilder operation(Operation operation);
+
+    /**
+     * Builds the {@link AttributeModifier} with the set parameters.
+     *
+     * @return The AttributeModifier with the set parameters
+     */
+    AttributeModifier build();
```
gabizou: If you can add a `reset` method so that the same builder can be re-used.

src/main/java/org/spongepowered/api/attribute/AttributeTargets.java
gabizou: Yes, but what does `GENERIC` mean?

src/main/java/org/spongepowered/api/potion/PotionEffectType.java
gabizou: I'm still questioning why `PotionEffectType` is an AttributeSource.

ST-DDT: If you drink a Speed Potion for example, MC will add a AttributeModifier to the Player which increases the Move Attribute. And remove it after the potion expires.
Iirc You could have look at MC's `Potion` class source code which is the equivalent in MC of `PotionEffectType` in SpongAPI.
Or maybe have a look at the issue in the description which gives little extra info.

gabizou: @ST-DDT I understand that certain potion types have different attribute modifiers, but I hardly think that a PotionEffectType should be a source. If anything, that complicates finding out the attribute source. A Potion being used is an attribute modifier source, not the potion effect type.

ST-DDT: It may be useful to add it to Potion, but that may be very confusing as well because it would only be a proxy there. And at the time of the removal of the effect the potion is no longer available. 

Un addition to that think of the possebility to change the potion effects.
The potion itself can have lots of stuff associated to it but they do not effect other potions as well. But if you change the attribute modifiers on one potion you will alter it for all potions of the same type. This is a very bad side effect that the devs will not expect.
Of course it would be cool to have attribute modifiers per potion item that wear down after a certain time but this is not related to MCs Potionsystem is probably out of the scope of this PR.

PR #406, opened 2015-01-22
Load or unload plugins based on the ID.

minecast:


progwml6:
Loading and unloading plugins at runtime will not be supported at this time. It has never been very reliable due to some plugin's using other plugins API's.

progwml6:
additional explanations: https://forums.spongepowered.org/t/plugin-reloading-enabling/684/14

PR #407, opened 2015-01-23
Feature/async scheduler2

sibomots:
So spaketh ZIdane: <Zidane> F*($ing gorgeous, scheduler works wonderfully


zml2008:
+zml	         imo tasks should have to be given names when being scheduled
13:17	+sibomots	i considered the Plugin's ID as part of the name like "FooPlugin-#"
13:17	@blood_	from an administration POV, naming a scheduled task makes the most sense
13:17	@blood_	how else would you know wtf is running?
13:17	@blood_	if you have like 100 scheudled tasks
13:17	@blood_	all unnamed
13:17	+zml	like PEX-TimedPermissionCleaner-n
13:17	@blood_	so if we dont require that, we should
13:18	+sibomots	or Plugin-SequenceID-SerialID  like Sequence being the thread # and Serial ID being incremented by scheduler, etc.. no big to me.
13:18	@blood_	PluginID-name + whatever else
13:18	+zml	what are sequence/serial ids even useful for?
13:18	@blood_	PluginID-taskname
13:18	+sibomots	k make a comment to the PR on feature/async-scheduler2 PR so I can bolt that in.
13:18	+zml	yeah
13:19	@blood_	since 2 plugins could name a task the same
13:19	+sibomots	imho, there ought to be a renameTask API method, or something like that. but the original API didn't have it.
13:19	+zml	or if a plugin has non-unique task names, add PluginId-tasknname-number
13:19	+zml	Task.setName()?
13:19	+sibomots	ssds
13:20	+sibomots	same shit different spelling
13:20	 +zml	pretty much

sibomots:
replaced with PR  423

PR #410, opened 2015-01-24
Add CommandSource#sendMessage(Iterable<String>)

lexware:


Minecrell:
This is not possible, because both `Iterable<Message>` and `Iterable<String>` will be just `Iterable` at runtime due to type erasure. Also, as of https://github.com/SpongePowered/SpongeAPI/commit/0d915435ba12b8403c6cea93ddebb36d93f01e40 the String overload was removed, so there is no need for that anymore.

lexware:
:(

—

PR #411, opened 2015-01-24
Retrieve world a block resides in

MattBDev:
In the current API there is no ability to retrieve the world a block resides in. This PR adds this ability which is helpful for plugins that rely on checking blocks and their location.


kenzierocks:
`blockLoc.getExtent()` is your world.

Issue #302, opened 2014-12-18
Player Disconnecting/Bans


gratimax:
This was brought up in https://forums.spongepowered.org/t/sponge-api-disconnecting-a-player-user/5197. Preferably a method to force disconnection of an online player coupled with some sort of ban system is best here, although I think having plain `setBanned` `getBanned` is not enough.

hedgehog1029:
It'd be pretty easy to move the code to a seperate class (package org.spongepowered.api.service.bans exists now) and have a method for getting the bans service. Maybe it could pull data from the vanilla bans file on startup, modify internally, and then write to the vanilla bans file on shutdown.

ST-DDT:
Is there any code implemented for the API yet?
The branch linked above seems outdated.

I like the concept, but you should not enforce any behaviour on a service or implementatio.
Maybe a init and shutdown method for the service is enough. Then the service can do whatever it wants.

hedgehog1029:
hedgehog1029/SpongeAPI@22bb2fd1aa9007185faa97f3b16c93107b807b3d

That's the latest commit to my repo.

gratimax:
Continued in #412

PR #412, opened 2015-01-25
Bans API

gratimax:
This is a continuation of #302, using the current changes that @hedgehog1029 has made.


src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -144,4 +145,56 @@
      * @return The connection
      */
     PlayerConnection getConnection();
+
+    /*
+     * Returns if the player is banned or not (from the vanilla banning system)
+     *
+     * @return Banned state of player.
+     */
+    boolean isBanned();
```
simon816: This already exists in [`User`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/player/User.java#L92)
Maybe some the methods should be in there too, i.e. `ban()`, `pardon()` and `getBanInfo()`

lexware:
You may want some sort of BanManager. Example - dynmap needs to get all the banned ips.

src/main/java/org/spongepowered/api/service/bans/BanInfo.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @return The name of the player.
+	 */
+	String getName();
+
+	/**
+	 * Gets the date of the ban.
+	 *
+	 * @return Creation date of the ban
+	 */
+	String getDateBanned();
```
lexware: Why not have this a a Date?

src/main/java/org/spongepowered/api/service/bans/BanInfo.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @return Banning player.
+	 */
+	String getSource();
+
+	/**
+	 * Gets the expiration time, formatted in long; if forever returns 0
+	 *
+	 * @return Expiration time of the ban.
+	 */
+	long getExpiration();
```
lexware: Why not have this a a Date?

hedgehog1029:
I was thinking of moving this to User, because I guess it makes more sense...
IP bans still need to be done.

ST-DDT:
I thought of adding an enum for that to those methods.

ST-DDT:
Are going to continue this PR?

src/main/java/org/spongepowered/api/entity/player/Player.java
modwizcode: This should be in User for sure, it's possible to ban somebody while they're offline.

src/main/java/org/spongepowered/api/util/bans/BanBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+
+    void setReason(String reason);
+
+    Date setBanDate();
+
+    Date setExpirationDate();
+
+    void setSource(@Nullable CommandSource source);
+
+}
```
Bammerbom: Is this correct layout for a builder? I thougth it was like
.user(User).reason(Reason).banDate(Date) etc.

And why do setBanDate() and setExpirationDat() have no arguments? Change it to something like setBanDate(Date)

gratimax: this commit is a WIP as by the name, but nice catch

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -144,4 +144,17 @@
      * @return The connection
      */
     PlayerConnection getConnection();
+
+    /**
+     * Kicks the player.
```
Aaron1011: The reason displayed to the player should be specified, since none is provided.

gratimax: @Aaron1011 what would you suggest as the default behavior, then? `You were kicked`? Or maybe the javadocs should say that the behavior is undefined.

Mumfrey: This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  adam@eq2.serveftp.net
    Domain eq2.co.uk has exceeded the max emails per hour (61/60 (101%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext1.iad.github.net ([192.30.252.192]:42033 helo=github-smtp2a-ext-cp1-prd.iad.github.net)
	by web3.cov.yesit.co.uk with esmtps (TLSv1.2:DHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.84)
	(envelope-from <noreply@github.com>)
	id 1YJ5A4-0006vy-1y
	for adam@eq2.co.uk; Wed, 04 Feb 2015 18:58:52 +0000
Date: Wed, 04 Feb 2015 10:58:52 -0800
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1423076332;
	bh=km2t+TE7/Qdt29B9RU8GOZuInIEJqk9esXRTWVtf1s8=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=UyDAY5rXAqOEFVFjfsdqZXhDAV+XoZbGspQOy84kAjakkm8D4diiPxvq4tmHWqylO
	 V8QtXNUbvtGDMgBzT5Nc8+GvpcjEKWnyOMrHHgQ58Ux8RanP6beI0liw+/MmjSXZf/
	 pSYIEPmJYvu3bSeo7UIyM52IyFJ7jNvtFQU961aA=

kenzierocks: Oh mail delivery.

Mumfrey: This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  adam@eq2.serveftp.net
    Domain eq2.co.uk has exceeded the max emails per hour (62/60 (103%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext7.iad.github.net ([192.30.252.198]:51283 helo=github-smtp2a-ext-cp1-prd.iad.github.net)
	by web3.cov.yesit.co.uk with esmtps (TLSv1.2:DHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.84)
	(envelope-from <noreply@github.com>)
	id 1YJ5Au-00070m-3R
	for adam@eq2.co.uk; Wed, 04 Feb 2015 18:59:44 +0000
Date: Wed, 04 Feb 2015 10:59:44 -0800
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1423076384;
	bh=cMhV2joVoVSFVpYBUoB+OpU4g5QsM80TimBZf+0uTuo=;
	h=From:Reply-To:To:Cc:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=c1QmJ6jqlDr1shmrcwZlAjitHJnr0AyqpAZPVGXvetODJKgd50UgscPD6tVlWGIpe
	 qXoNYup1HkRtOS/plVa63ep/+RzLkuPi1J+RxhUqSa/IC1erYlCacVT6EtwiRzr/t1
	 RKjR+ZRaF/MoLmStYoUx/ycNw9VMgUYJvs78BvBU=

lexware: lel

ST-DDT: `Kicked by an operator.` is the default.

>Oh mail delivery.

Probably to much change notifications in the Attribute PR

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -144,4 +144,17 @@
...
+     */
+    void kick();
+
+    /**
+     * Kicks the player given a reason.
+     *
+     * @param reason The reason for the kick
+     */
+    void kick(String reason);
```
ST-DDT: Why `String`? Kick messages can be colored. So it should be `Message` although it has reduced support for everything else.

gratimax: @ST-DDT they're stored as String internally

gratimax: I should add on that the kick reason is wrapped in a ChatComponentText, so it doesn't really make sense to have Message

gratimax:
Some things to consider:

- Should a `BanProvider

gratimax:
Some things to consider:

- Should a `BanProvider` be able to 'revoke' bans provided by other providers?
- Should there be a way to ban via IP? E.g. `IpBan`? How would that work?

src/main/java/org/spongepowered/api/entity/player/Player.java
ST-DDT: @gratimax IMO The point is not what is stored somewhere or somehow, because that can change very soon.

We have `Message.Text` or `Text.Literal` for Colored Strings, which supports all required formatting for the given message. I just find it very strange to have only one place in the entire APi that supports colored messages, but is not using Message for that. Using `kick(coloredMessage.toLiteral())` or `kick("§4Do not Spam")` looks weird.

But its probably easier to use it from commands that way... well is there somewhere a `TextStyle.getLegacyChar()` method to do the replacement from `&` to `§` myself?

gratimax: eh, true I guess. There's a whole bunch of methods like fromLegacy in the Messages class.

ST-DDT:
> Should a BanProvider be able to 'revoke' bans provided by other providers?

Yes, if using the BanService to pardon a user, otherwise you will be in trouble if you want to unban a player from other sources.

>Should there be a way to ban via IP? E.g. IpBan? How would that work?

Just add an enum (like final class) `BanTarget` that specifies, what have been banned.

````java
public final class BanTargets {

public static final BanTarget USER = null;
public static final BanTarget IP = null;

}
````

Thanks for "Support using Message.Text in bans", but you only changed it for Kicks and not for `Ban`s, where it would be useful, especially if other than vanilla ban providers may be able to store more complex ban messages.

**Comment by gratimax** [comment]
@ST-DDT 

> Yes, if using the BanService to pardon a user, otherwise you will be in trouble if you want to unban a player from other sources.

The reason I am against this is because it adds an enormous amount of complexity. There's also ways to get all other ban providers and pardon bans from them, so it isn't actually _that_ useful.

> Just add an enum (like final class) BanTarget that specifies, what have been banned.

Eh. Then the 'value' of the ban target is ambiguous, it would be `InetAddress` instead of `User`.

> Thanks for "Support using Message.Text in bans", but you only changed it for Kicks and not for Bans, where it would be useful, especially if other than vanilla ban providers may be able to store more complex ban messages.

Whoops, fixing now

**Comment by ST-DDT** [comment]
So maybe moving `User getUser()` to `UserBan`
So that you have `Ban`, `UserBan` and `IPBan`. 
`InetAddress` only works for active connections and not for past ones, or am I wrong?

**Comment by gratimax** [comment]
@ST-DDT 

> So maybe moving User getUser() to UserBan

I was considering that in addition to something like the enum you provided earlier.

> InetAddress only works for active connections and not for past ones, or am I wrong?

You're probably thinking of `InetSocketAddress`. `InetAddress` has no connection to an existing socket

**Comment by ST-DDT** [comment]
>> So maybe moving User getUser() to UserBan

>I was considering that in addition to something like the enum you provided earlier.

Having both sounds good for me. 

>>InetAddress only works for active connections and not for past ones, or am I wrong?

>You're probably thinking of InetSocketAddress. InetAddress has no connection to an existing socket

Yeah ups. Too much interfaces/classes out there.

File: src/main/java/org/spongepowered/api/event/entity/living/player/PlayerKickEvent.java [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+ * Called when a player is kicked.
+ */
+public interface PlayerKickEvent extends PlayerEvent {
+
+    /**
+     * Gets the reason for the kick.
+     *
+     * @return The kick reason as a String
+     */
+    Optional<Message.Text> getReason();
```
Aaron1011: When wouldn't this be available?

gratimax: When I added this I thought that there was no default kick reason(for when just `/kick` is run). Feel free to change it and add javadocs for what the kick reason is by default.

Minecrell:
@gratimax @modwizcode Is this already finished (and ready to be reviewed)? There is no [WIP] in here. If not @modwizcode, are you going to finish it (as you've assigned it to yourself)?

Minecrell:
@gratimax @modwizcode Bump -  we need to get this finished.

jamierocks:
i can't see anything about ip bans.

src/main/java/org/spongepowered/api/util/bans/BanBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+import org.spongepowered.api.text.message.Message;
+import org.spongepowered.api.util.command.CommandSource;
+
+import javax.annotation.Nullable;
+import java.util.Date;
+
+/**
+ * Represents a builder that creates bans.
+ */
+public interface BanBuilder {
```
ZephireNZ: Why isn't this builder fluent?

Zidane:

Can't Merge.

@modwizcode this needs to get finished up or re-assigned to someone else.

modwizcode:
@Zidane On it, to confirm changes, fixing merge difficulties, adding ip
banning support?

gratimax:
I'm basically done. The only question is where to put events like `IpBanEvent` and `IpPardonEvent`. I have put them in the `org.spongepowered.api.event.net` package for now, but they probably belong better somewhere else.

src/main/java/org/spongepowered/api/util/ban/BanFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    Ban of(User user);
+
+    /**
+     * Creates an indefinite ban with a reason on a user.
+     *
+     * @param user The user
+     * @param reason The reason
+     * @return The created ban
+     */
+    Ban of(User user, Text.Literal reason);
```
ZephireNZ: Perhaps make the same methods for IP bans too?

modwizcode:
I think either event.ban or event.player.ban something like that.

src/main/java/org/spongepowered/api/util/ban/BanFactory.java
gratimax: I didn't think it was useful.

src/main/java/org/spongepowered/api/util/ban/Ban.java
```diff
@@ -0,0 +1,109 @@
...
+     * Gets whether this ban is indefinitely long, e.g. has no expiration date.
+     *
+     * @return True if this ban has no expiration date, otherwise false
+     */
+    boolean isIndefinite();
+
+    /**
+     * Represents a ban made on a user.
+     */
+    interface User {
```
simon816: Shouldn't this extend `Ban`? Same for `Ip` interface too

gratimax: Whoops!

src/main/java/org/spongepowered/api/util/ban/BanBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    BanBuilder address(InetAddress address);
+
+    /**
+     * Sets the type of the ban.
+     *
+     * @param type The type to be set
+     * @return This builder
+     */
+    BanBuilder banType(BanType type);
```
simon816: Possibly remove the `ban` prefix here and on `banDate`?

ZephireNZ: Agreed, it's already implied that a BanBuilder would be wanting a *ban* type, not a something else

gratimax: I'll do it on type but not date, since there are two dates (the date the user is banned and the date the ban expires)

gratimax: Actually, I guess it's called startDate elsewhere, so let me change that.

src/main/java/org/spongepowered/api/entity/player/User.java
```diff
@@ -126,6 +128,20 @@
     void setBedLocation(@Nullable Location location);
 
     /**
+     * Check whether this user has an operator status.
+     *
+     * @return Whether this user is an operator
+     */
+    boolean isOperator();
```
simon816: I'm not sure about this operator get/set short-cut. There are other ways to check of operator status using the permission API, see https://github.com/SpongePowered/SpongeAPI/pull/285#issuecomment-68066983

src/main/java/org/spongepowered/api/event/entity/player/UserEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.spongepowered.api.event.entity.player;
+
+import org.spongepowered.api.entity.player.User;
+import org.spongepowered.api.event.GameEvent;
+
+/**
+ * Represents an event that involves a User.
+ */
+public interface UserEvent extends GameEvent {
```
simon816: I don't see much use of a `UserEvent` since 'user's are not going to be subject to many events, I may be wrong though.

src/main/java/org/spongepowered/api/entity/player/User.java
gratimax: Ah, I missed that. I'll remove it

src/main/java/org/spongepowered/api/event/entity/player/UserEvent.java
gratimax: There aren't many now, but users are still subject to events in terms of changing game data (like bans).

src/main/java/org/spongepowered/api/util/ban/Ban.java
```diff
@@ -0,0 +1,109 @@
...
+ * Represents a ban made on an object.
+ */
+public interface Ban {
+
+    /**
+     * Gets the type of this ban.
+     *
+     * @return The ban type
+     */
+    BanType getBanType();
```
simon816: `getType()`? Follows other parts of the API e.g BlockLoc.getType()

PR #413, opened 2015-01-25
Get the World from Chunk.

lexware:


Deamon5550:
I wonder if this would be better implemented via more generic extent parenting?

See #322

modwizcode:
@Deamon5550 Actually that PR concluded with @sk89q's opinion that Extent parenting isn't something needed/wanted. So I am all for this PR being merged given that this is a specific case where Extent parenting makes sense.

sk89q:
Chunk could have parenting on its own.

gratimax:
Even if we do add more generic extent parenting, this method is pretty harmless since it has very firm standing in how Minecraft works

me4502:
I agree, I see no reason to not merge it.

modwizcode:
I was merely saying that the generic Extent parenting PR wasn't wanted, but this case is an exception. Let's merge this.

PR #420, opened 2015-01-28
Blazing fast ChunkDataStreams

### LINKED ISSUE #256: [Suggestion] List of all generated chunks
A plugin that I developed for bukkit requires me to create large structures across many chunks. These chunks have to have already been generated (whether they are loaded or not) and devoid of vanilla ores. I've also gotten requests that I make this plugin compatible with already generated worlds. To make doing this easier I request that the following methods be added to the world class:

```java
/*
* Returns whether a particular chunk has already generated structures, 'the layer', and 'decorations' 
* regardless of its load status.
*/
public boolean isGenerated(int x, int z)

/*
* Returns a List of all the chunk coordinates that have already been 'decorated' regardless of load 
* status. The list is a snapshot of the world as it was when the method was called. String can be 
* replaced by other objects more suited to coordinate retrieval, but otherwise only contain the x & z 
* coordinate data of the chunks within the world. 
*/
public List<String> getWorldChunkList()
```
Edit: clarified what I meant by 'generated' .@gratimax pointed out that during chunk generation structures are generated first, then the 'layer' (terrain landforms), and then 'decorators'.

gabizou:
Fixes #256 

The previous API did not specify whether `ChunkIterator` would be blocking on the main thread and provided little ways to read specific chunk data.

This adds the ability to get a specific `Chunk`s data via asynchronous [`ListenableFuture`](https://code.google.com/p/guava-libraries/wiki/ListenableFutureExplained) and aids in refining the asynchronous nature of `ChunkDataStream`.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/storage/WorldStorage.java
+++ b/src/main/java/org/spongepowered/api/world/storage/WorldStorage.java
@@ -24,12 +24,58 @@
  */
 package org.spongepowered.api.world.storage;
 
+import com.flowpowered.math.vector.Vector3i;
+import com.google.common.base.Optional;
+import com.google.common.util.concurrent.ListenableFuture;
+import org.spongepowered.api.service.persistence.data.DataContainer;
+import org.spongepowered.api.world.Chunk;
+
 public interface WorldStorage {
 
     /**
-     * Gets a {@link ChunkIterator}.
+     * Gets a {@link ChunkDataStream}.
+     *
+     * <p>Usage of a {@link ChunkDataStream} should be limited to asynchronous
+     * tasks to avoid hanging the main thread.</p>
+     *
      * @return An iterator of generated chunks
      */
-    ChunkIterator getGeneratedChunks();
+    ChunkDataStream getGeneratedChunks();
+
+    /**
+     * Checks if the given chunk coordinates represented by {@link Vector3i}
+     * exist in the world.
+     *
+     * <p>Note that this is an asynchronous check as the storage of chunks
+     * can not be guaranteed to remain in synch with the server, let alone
+     * on the server thread.</p>
+     *
+     * <p>It is imperative to avoid waiting for the {@link ListenableFuture} to complete
+     * its task.</p>
+     *
+     * @param chunkCoords The chunk coordinates
+     * @return Whether the chunk exists or not
+     */
+    ListenableFuture<Boolean> doesChunkExist(Vector3i chunkCoords);
+
+    /**
+     * Gets a {@link DataContainer} including all data related to a {@link Chunk}.
+     *
+     * <p>The container is a read only instance of the data, and therefor
+     * should not be considered as mutable data. Changes are NOT saved, and
+     * the data may not be in synch with the server if the chunk is currently
+     * loaded.</p>
+     *
+     * <p>This may not return a {@link DataContainer} in the event there is no
+     * chunk data generated at the desired coordinates.</p>
+     *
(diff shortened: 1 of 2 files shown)


me4502:
Obligatory #420 'Blazing'

This PR is quite a nice, I don't see any issues with it.

modwizcode:
I like this quite a bit. This solves completely what the Iterator was half-solving.

Kobata:
The comments at least address the major possible issue with doing this outside of the main thread (data is essentially a 'snapshot')

I'm not quite sure what the point of using Future<> is if you're telling people not to wait for it though.

gabizou:
> I'm not quite sure what the point of using Future<> is if you're telling people not to wait for it though.

To pass the future to an async thread that can wait for the future to complete. From the JDK Javadocs:

> The result can only be retrieved using method get when the computation has completed, blocking if necessary until it is ready. 

We definitely do not want to block the main thread when we get from the future. The idea is simply to just pass off the future to another thread and wait on that thread.

sibomots:
Do you have some test-plugin code that demonstrates the capability?

Date: Wed, 28 Jan 2015 11:07:23 -0800

src/main/java/org/spongepowered/api/world/storage/WorldStorage.java
```diff
@@ -24,12 +24,58 @@
...
+    /**
+     * Checks if the given chunk coordinates represented by {@link Vector3i}
+     * exist in the world.
+     *
+     * <p>Note that this is an asynchronous check as the storage of chunks
+     * can not be guaranteed to remain in synch with the server, let alone
+     * on the server thread.</p>
+     *
+     * <p>It is imperative to avoid waiting for the {@link Future} to complete
+     * its task.</p>
```
Kobata: > To pass the future to an async thread that can wait for the future to complete. From the JDK Javadocs:

Except it explictly says *not* to do this.

OniBait:
Perhaps use guava's ```ListenableFuture``` instead? Seems a bit more friendly for the "preferred" async callback scenario. Otherwise you're going to have code that will end up waiting for the future on the main thread at some point (otherwise why bother calling the method?)

gabizou:
> Otherwise you're going to have code that will end up waiting for the future on the main thread at some point (otherwise why bother calling the method?)

Not if the developer is using the future correctly. It's simple to pass in a Future to a new thread and wait for it in said thread, avoiding blocking the main thread.



> Perhaps use guava's ListenableFuture instead? Seems a bit more friendly for the "preferred" async callback scenario.

Sure, we could use that instead, registering callbacks is a nice idea.

Thoughts @SpongePowered/developers ?

modwizcode:
Async event handlers? Crazy but thought I'd mention it.

src/main/java/org/spongepowered/api/world/storage/WorldStorage.java
Zidane: javadocs could be clarified but this is to stop the mass of plugin developers deciding to suspend the main thread waiting for a future.

gabizou: This was already fixed in 99cf3f9

Kobata: You changed line 73, not this line (54). There were two copies of the same text.

gabizou:
@Zidane I believe this is ready for merging.

gratimax:
:+1:

PR #421, opened 2015-01-28
Fix javadocs

lexware:


Minecrell:
While we don't accept small pull requests with minor changes like this normally, I have merged this into the `minor` branch so this is fixed now. Thanks

lexware:
No problem, next time I'll just mention it on irc.

—

PR #422, opened 2015-01-28
Change how silk touch enchantment works

simon816:
If Sponge is going to work with Forge mods, it will need to allow for a location/player sensitive way of testing whether a block can be harvested with silk touch.
As @diesieben07 said in #minecraftforge, a mod may only allow a block to be harvested if the player has been playing for a certain amount of time, or it may depend on what tile entity data the block has etc.

This PR changes the API to be more like vanilla where the method is on the block, not the enchantment. `BlockLoc` is a more appropriate place than `BlockType` in this case.


modwizcode:
This will then be inconsistent with the representation of other enchantments. While I believe this may be correct, shouldn't plugins themselves then be allowed to change this behavior? AFAIK that would be done with some sort of event. This brings up the issue of where mods are on the event chain. Are they combined with vanilla's callback?

modwizcode:
@SpongePowered/developers comments?

me4502:
I was told to do it the way it is currently done for a few reasons. I had originally done it like this PR, and I was told to change it.

simon816:
@me4502 Yeah I originally suggested it. It is fine for vanilla behaviour but to be compatible with forge mods, it cannot be a simple registry of silk touch-able blocks.

me4502:
I believe Zidane also told me to do it that way as well.

gabizou:
I would rather mods to have a chance to be ahead of the event chain with regards to silk touch overriding, mostly because I believe plugins should have the final say overall.

modwizcode:
This needs to be changed to an event based model, because this is really a specific case of a HarvestEvent. Make another PR for that if you want.

See forge event for inspiration/copying https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/event/world/BlockEvent.java#L45

PR #438, opened 2015-02-07
Add Game#getDirectory()

lexware:


Mumfrey:
Congratulations, you forgot to import `java.io.File` so this doesn't compile. In addition to that you need to clarify what you mean by "game directory" since there are a number of possible meanings:

* The dir that the game was launched from (eg. `user.dir`)
* The base directory of the installation
* The base directory of the profile being launched

lexware:
Sorry about that, IntelliJ has proved to have been a nuisance.

ST-DDT:
Isnt this basically the same as calling new File(".")?

kenzierocks:
Depends on the definition of the directory, and some settings like `--universe` may change that.

AzureusNation:
When you say the "Game directory" do you mean the root folder of the SpongeAPI implementation?

lexware:
I mean the folder containing the config folder and world folder etc.

simon816:
`Server#getDirectory()` would be a better place for this method.
I don't really think there should be a way to directly access the root dir, maybe just config dir and world dir are needed. The directory structure may differ in different implementations so would not be reliable to use the root anyway.

lexware:
The reason I put it in Game as opposed to Server was because although it may be used more in Server, some may find a use for it in Game.

gratimax:
I agree with @simon816. The only useful directories to plugins are the config, plugin, and world dirs.

lexware:
@gratimax and @simon816 Some plugins do want to make use of the directory notably LapisBlue/Pore

gratimax:
@lexware If so it should figure that on its own, this is not needed by 95% percent of plugins.

gratimax:
Pinging @SpongePowered/developers for some opinions. While most plugins do not need to use this method, there are some where it is a verifiable use case.

Mumfrey:
> The reason I put it in Game as opposed to Server was because although it may be used more in Server, some may find a use for it in Game.

The point isn't what you've done or what you meant. It's that you need to *clarify what you mean* in the javadoc. As I said above, the "game directory" can mean multiple things, and doesn't mean what you think it means. So you need to clarify what the intention of the method is in some way.

lexware:
@Mumfrey sure thing, I will update now.

lexware:
will
/**
     * Gets the directory containing the world and config folders etc.
     * 
     * @return The game directory
     */ 
do?

AzureusNation:
So in other words it is the root folder of the implementation.

lexware:
yes

lexware:
Infact thats a much better way of putting it

Mumfrey:
> So in other words it is the root folder of the implementation.

That's why I think it needs clarification, because under launchwrapper (which obviously our implementation is using) the "game directory" is arbitrary and specified on the command line. So it needs to be much clearer than saying "implementation root" because the actual files which make up the implementation may be somewhere completely other than that game data. This is what I was clarifying.

In general minecraft parlance, "game directory" is actually the profile root directory passed to the game when it starts by the launcher, and contains "saves, config, screenshots" etc. Whereas the launch directory may be entirely other (for example it's *usually* APPDATA/.minecraft) and contains the implementation (the "versions, libraries" folders etc). Sometimes these two folders are the same and sometimes they are not

Basically, as long as it's crystal clear what is meant by this accessor so that all implementations can return the expected value, that's all that matters afaict

lexware:
So @Mumfrey how would you word it? Being the profile root directory.

Mumfrey:
I think @AzureusNation may be able to add some insight, because a platform-agnostic description is for the best. Like I say on vanilla we have "launch directory" and "game/profile/data directory" depending on how you look at it. Launch directory will always be the equivalent of `System.getProperty("user.dir")` or `new File(".")` whereas the latter is more nebulous but ultimately the directory most things are concerned with.

Do as you please, but just try to factor in these facts when making your description.

Kobata:
In terms of dedicated server, there's also a few possible places: One is where the server.jar itself is, which can be different from the root for worlds (--universe) or the selected primary world (--world)

boformer:
Also mention that there is no guarantee to find a "mods", "plugins" or "logs" folder inside this root folder, because it is implementation-dependent

me4502:
I personally don't believe this method should exist for one main reason. Things like McMyAdmin have the jar in another folder, meaning that if you were to get the server jar directory, it would not be accessible to the user in most cases with that wrapper.

modwizcode:
I would have to agree, the root directory of the server jar is usually in a
very different place for large setups compared to the smaller simpler
setups.

As for a similar usefulness, I think the main use would be direct access to
the world base directory, which can be guaranteed to be a specific single
directory in relation to the world data. That would be a much better
proposal.

ST-DDT:
Please remember the discussion we had while talking about the world API.
It is not guaranteed that a world has a folder at all. It could be stored in a databe as well.

But there is always a root folder where minecraft stores its config and ban and op files. And in cases where it is not the same folder as the jar its important to get that folder in that case it is not new File(".").
 Although you cannot relyably get the bans and world files and folders.

lexware:
Sent from Outlook
_____________________________

boformer:
The bans and op files could also be stored in a database ;)

ST-DDT:
Sure but the server still has the knowlegde of its root folder. Although it may not be used for anything in database driven servers. Except from the db connection config.

modwizcode:
Another perspective, just add the method and place a serious warning on it.
Make it optional and empty if the server's storage directory is changed.

gogobebe2:
FUCKING HOW?

simon816:
@gogobebe2 What the heck is your problem?!

Mumfrey:
This clearly needs a lot more thought than just adding a simple method to `Game` so closing for now, will reopen if you feel you have a watertight solution to the issues mentioned above.

PR #440, opened 2015-02-08
Improve Player and Human interfaces

caseif:
This PR improves the Player and Human interfaces. Most notably, it refactors the food and experience APIs and improves documentation. It also adds methods pertaining to movement mechanics.

It is implemented by [Sponge-164](https://github.com/SpongePowered/Sponge/pull/164).

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/Human.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Human.java
@@ -44,7 +44,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current exhaustion level
      */
-    float getExhaustion();
+    double getExhaustion();
 
     /**
      * Sets the exhaustion value of this human entity.
@@ -53,7 +53,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @param exhaustion The new exhaustion level
      */
-    void setExhaustion(float exhaustion);
+    void setExhaustion(double exhaustion);
 
     /**
      * Gets the current saturation level of this human entity.
@@ -62,7 +62,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current saturation level
      */
-    float getSaturation();
+    double getSaturation();
 
     /**
      * Sets the saturation level of this human entity.
@@ -71,17 +71,18 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @param saturation The new saturation level
      */
-    void setSaturation(float saturation);
+    void setSaturation(double saturation);
 
     /**
      * Gets the current food saturation of this human entity.
+     *
      * <p>Food level has health effects, depending on game difficulty and
      * hunger levels. If the food level is high enough, the human entity
      * may heal. If the food level is at 0, the human entity may starve.</p>
      *
      * @return The current food level
      */
-    float getFoodLevel();
+    double getFoodLevel();
 
     /**
      * Sets the current food saturation of this human entity.
@@ -92,7 +93,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @param foodLevel The new food level
      */
-    void setFoodLevel(float foodLevel);
+    void setFoodLevel(double foodLevel);
 
     /**
      * Gets the current experience accumulated since the last level-up.
@@ -101,7 +102,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current experience accumulated since the last level-up.
      */
-    double getExperienceSinceLevel();
+    int getExperienceSinceLevel();
 
     /**
      * Sets the experience accumulated since the last level-up.
@@ -110,7 +111,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @param experience The experience accumulated since the last level-up.
      */
-    void setExperienceSinceLevel(double experience);
+    void setExperienceSinceLevel(int experience);
 
     /**
      * Gets the experience required since the last level to level up.
@@ -121,7 +122,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return the experience required since the last level to level up
      */
-    double getExperienceBetweenLevels();
+    int getExperienceBetweenLevels();
 
     /**
      * Gets the current experience level of this human.
--- a/src/main/java/org/spongepowered/api/entity/living/Human.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Human.java
@@ -44,7 +44,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current exhaustion level
      */
-    double getExhaustion();
+    float getExhaustion();
 
     /**
      * Sets the exhaustion value of this human entity.
@@ -62,7 +62,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current saturation level
      */
-    double getSaturation();
+    float getSaturation();
 
     /**
      * Sets the saturation level of this human entity.
@@ -71,7 +71,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @param saturation The new saturation level
      */
-    void setSaturation(double saturation);
+    void setSaturation(float saturation);
 
     /**
      * Gets the current food saturation of this human entity.
--- a/src/main/java/org/spongepowered/api/entity/living/Human.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Human.java
@@ -39,6 +39,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
 
     /**
      * Gets the exhastion value of this human entity.
+     *
      * <p>When exhaustion level reaches 0, saturation may begin to diminish.
      * </p>
      *
@@ -48,6 +49,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
 
     /**
      * Sets the exhaustion value of this human entity.
+     *
      * <p>When exhaustion level reaches 0, saturation may begin to diminish.
      * </p>
      *
@@ -57,6 +59,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
 
     /**
      * Gets the current saturation level of this human entity.
+     *
      * <p>Saturation acts as a buffer for food level. Food level will not
      * decrease while the saturation level is greater than zero.</p>
      *
@@ -66,6 +69,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
 
     /**
      * Sets the saturation level of this human entity.
+     *
      * <p>Saturation acts as a buffer for food level. Food level will not
      * decrease while the saturation level is greater than zero.</p>
      *
@@ -157,12 +161,18 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
     /**
      * Returns whether this human entity is sneaking.
      *
+     * <p>Sneaking may entail reduced nametag visibility and movement speed in
+     * relation to the subject Human.</p>
+     *	 
      * @return Whether this human entity is sneaking
      */
     boolean isSneaking();
 
     /**
      * Sets whether this human entity is sneaking.
+     *
+     * <p>Sneaking may entail reduced nametag visibility and movement speed in
+     * relation to the subject Human.</p>
      *
      * @param sneaking Whether this human entity should be sneaking.
      */
--- a/src/main/java/org/spongepowered/api/entity/living/Human.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Human.java
@@ -38,45 +38,67 @@ import org.spongepowered.api.item.inventory.Inventory;
 public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, Carrier {
 
     /**
-     * Gets the hunger value of this human entity.
-     * <p>Hunger is a type of exhaustion such that when the exhaustion
-     * depletes, saturation may begin to diminish.</p>
+     * Gets the exhastion value of this human entity.
+     * <p>When exhaustion level reaches 0, saturation may begin to diminish.
+     * </p>
      *
-     * @return The current hunger
+     * @return The current exhaustion level
      */
-    float getHunger();
+    double getExhaustion();
 
     /**
-     * Sets the hunger value of this human entity.
-     * <p>Hunger is a type of exhaustion such that when the exhaustion
-     * depletes, saturation may begin to diminish.</p>
+     * Sets the exhaustion value of this human entity.
+     * <p>When exhaustion level reaches 0, saturation may begin to diminish.
+     * </p>
      *
-     * @param hunger The hunger level
+     * @param exhaustion The new exhaustion level
      */
-    void setHunger(float hunger);
+    void setExhaustion(float exhaustion);
 
     /**
-     * Gets the current food saturation of this human entity.
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
+     * Gets the current saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
+     * decrease while the saturation level is greater than zero.</p>
      *
      * @return The current saturation level
      */
-    float getSaturation();
+    double getSaturation();
+
+    /**
+     * Sets the saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
+     * decrease while the saturation level is greater than zero.</p>
+     *
+     * @param saturation The new saturation level
+     */
+    void setSaturation(double saturation);
+
+    /**
+     * Gets the current food saturation of this human entity.
+     * <p>Food level has health effects, depending on game difficulty and
+     * hunger levels. If the food level is high enough, the human entity
+     * may heal. If the food level is at 0, the human entity may starve.</p>
+     *
+     * @return The current food level
+     */
+    float getFoodLevel();
 
     /**
      * Sets the current food saturation of this human entity.
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
      *
-     * @param saturation The saturation level to set
+     * <p>Food level has health effects, depending on game difficulty and
+     * hunger levels. If the food level is high enough, the human entity
+     * may heal. If the food level is at 0, the human entity may starve.</p>
+     *
+     * @param foodLevel The new food level
      */
-    void setSaturation(float saturation);
+    void setFoodLevel(float foodLevel);
 
     /**
      * Gets the current experience towards the next level.
+     *
+     * <p>This value ranges from 0 to 1, with the experience bar
+     * being empty at 0, and the player leveling up at 1.</p>
      *
      * <p>This is not the total experience this human has.</p>
      *
@@ -87,7 +109,12 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
     /**
      * Sets the experience accumulated towards the next level.
      *
-     * @param experience The experience
+     * <p>This value ranges from 0 to 1, with the experience bar
+     * being empty at 0, and the player leveling up at 1.</p>
+     *
+     * <p>This is not the total experience this human has.</p>
+     *
+     * @param experience The experience towards the next level
      */
     void setExperience(int experience);
 
@@ -112,7 +139,7 @@ public interface Human extends Living, ProjectileSource, ArmorEquipable, Tamer, 
      *
      * @return The current total accumulated experience
      */
-    int getTotalExperinece();
+    int getTotalExperience();
 
     /**
      * Sets the total accumulated experience starting from zero.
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,85 @@
...
+     * may heal. If the food level is at 0, the human entity may starve.</p>
+     *
+     * @param foodLevel The new food level
+     */
+    void setFoodLevel(float foodLevel);
+
+    /**
      * Gets the current experience towards the next level.
      *
+     * <p>This value ranges from 0 to 1, with the experience bar
```
gabizou: Sorry, but where the heck did you get this `from 0 to 1` information?

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -119,6 +146,79 @@
...
+
+    /**
+     * Gets the speed this human entity may fly at.
+     *
+     * @return the speed this human entity may fly at
+     */
+    double getFlySpeed();
+
+    /**
+     * Sets the speed this human entity may fly at.
```
gabizou: You're missing some integral restrictions on the movement speed in general. Such as negative speeds severely affecting clients (or preventing them from even being able to interact with the server).

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -119,6 +146,79 @@
...
+
+    /**
+     * Gets the speed this human entity may walk at.
+     *
+     * @return the speed this human entity may walk at
+     */
+    double getWalkSpeed();
+
+    /**
+     * Sets the speed this human entity may walk at.
```
gabizou: Same with restrictions on flight speed.

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -138,6 +137,38 @@
...
+     */
+    void setBedPosition(Optional<Vector3i> location);
+
+    /**
+     * Returns whether this player is invisible to another given player.
+     *
+     * @param otherPlayer The player to toggle visibility of this player for
+     * @return Whether this player is invisible to another given player
+     */
+    boolean isInvisible(Player otherPlayer);
```
gabizou: You're adding several methods to `Human` but you keep this one in `Player`? Why is that? Doesn't this also interfere with [`Living#isInvisible()`](https://github.com/SpongePowered/SpongeAPI/blob/c9efe362b74acd456b3a1909285b0f3ceca29657/src/main/java/org/spongepowered/api/entity/living/Living.java#L282)?

Are you intending these methods to be used as 

``` java
if (!myPlayer.isInvisible(myOtherPlayer)) {
  myServer.sendEntityPacket(myOtherPlayer, new EntityPacket(myPlayer));
}
```

To me, this almost sounds like something VanishNoPacket would seek to achieve.

src/main/java/org/spongepowered/api/entity/living/Human.java
caseif: I've based many of the methods in this PR off of Bukkit's spec, and this is how this method is defined. I can't imagine many use cases for this method (or more importantly, its setter) that don't depend on the percentage of the bar filled rather than the particular amount of experience, but I'm open to discussion about it.

src/main/java/org/spongepowered/api/entity/living/Human.java
caseif: Should these details be mentioned in the documentation?

src/main/java/org/spongepowered/api/entity/player/Player.java
caseif: I was more concerned about implementation when I wrote this method, as I wasn't sure whether it could be very easily implemented, but looking at it again I don't think it would be difficult to implement it in Living instead.

caseif:
@boformer Minecraft stores each bed location as a `BlockPos`. The primary bed spawn in dimension 0 (in your example, World1->NORMAL) is stored in the `spawnChunk` of `EntityPlayer`, while the rest are stored in a `HashMap` mapping them to their dimensions. It seems logical to only return the primary bed spawn, making the world implicit, but I see the benefit of explicitly defining it.

caseif:
@boformer Minecraft stores each bed location as a `BlockPos`, which is effectively identical to a `Vector3i` in that it does not contain a world. The primary bed spawn in dimension 0 (in your example, World1->NORMAL) is stored in its own field in `EntityPlayer`, while the rest are stored in a `HashMap`, mapped to their respective dimensions. It seems logical to only return the primary bed spawn, making the world implicit, but I see the benefit of explicitly defining it.

boformer:
The problem is that there could be mutliple "dimension 0". How do you know which world is the primary world?
Returning ``Vector3i`` instead of ``Location`` is like returning ``UUID`` instead of ``Player``...

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,85 @@
...
      *
+     * <p>This value ranges from 0 to 1, with the experience bar
+     * being empty at 0, and the player leveling up at 1.</p>
+     *
      * <p>This is not the total experience this human has.</p>
      *
      * @return The current experience towards the next level
      */
-    double getExperience();
+    double getExpereienceToLevel();
```
Minecrell: Typo? This should be `getExperienceToLevel`

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -57,18 +60,14 @@
...
-    /**
-     * Sets if the {@link Player} can fly via the fly key.
+     * <p>Passing {@link com.google.common.base.Optional#absent()} will set the
+     * player's display name to their name.</p>
      *
-     * @param allowFlight {@code True} if the player is allowed to fly
+     * @param displayName The new display name of this player.
      */
-    void setAllowFlight(boolean allowFlight);
+    void setDisplayName(Optional<Message> displayName);
```
Minecrell: Use `@Nullable` instead of `Optional` parameters

src/main/java/org/spongepowered/api/entity/living/Human.java
caseif: Good catch.

caseif:
@boformer I should clarify: when I say dimension 0, I'm referring to level 0, which there is only one of. That being said, I understand the need for the primary world to be returned, so I'll switch it to return a `Location`.

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -57,18 +60,14 @@
     Message getDisplayName();
 
     /**
-     * Returns whether the {@link Player} can fly via the fly key.
+     * Sets the player's diaplay name.
```
Bammerbom: Typo

gabizou:
Several issues with this PR:

1) Experience by the API is NOT a direct representation of experience as perceived from Vanilla Minecraft. Having said that, your PR breaks several things without fixing them:
 - [`getExperienceToLevel()`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-037dd70c30bad14c23b1b67e0c05d22cR104) changes the entire functionality of getting the current amount of experience towards the next level to return a ratio instead of the actual experience
 - You provide no getter methods for the required experience to complete the current level to compare to the ratio stated above.
 -  You leak implementation details about the experience being a ratio towards the next level but provide no actual "amount of experience" towards the next level.
 - The internal ratio is used for client rendering from my understanding (aside from storing the experience as a ratio towards the next level).

2) You add several methods stating reasons that "Bukkit had it." which is *NOT* a valid reason for additions to the API.

3) You're adding several methods to Human but fail to notice that several methods apply to `Living` as well. Specifically, [`Living#isInvisible()`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/living/Living.java#L282) is already provided in the API, yet your addition of [`Living#isInvisibleTo(Player)`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-6feb4e1e580ec4477aa33986c083923dR298) makes little sense short of applying either a VanishNoPacket functionality or a direct copy of the already provided method from Minecraft source (*which is used by the renderer*). You should note in the PR that you're adding to the Vanish API.

4) Movement speeds belong as [`Attributes`](https://github.com/SpongePowered/SpongeAPI/pull/403) which apply to [`getFlySpeed()`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-037dd70c30bad14c23b1b67e0c05d22cR198) and [`getWalkSpeed`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-037dd70c30bad14c23b1b67e0c05d22cR212). 

5) [`isFlightAllowed()`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-037dd70c30bad14c23b1b67e0c05d22cR184) is better handled by [`GameModes`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/player/gamemode/GameMode.java), not by the player itself. 

6) You add [`Player#getBedLocation()`](https://github.com/SpongePowered/SpongeAPI/pull/440/files#diff-2ab421167ec5d1c019761332abe9dec3R147) but you really should be adding it to [`User`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/player/User.java) so that the information is available in an offline state. Player is specifically an online `User` with a [`PlayerConnection`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/net/PlayerConnection.java), so any persistent data should be maintained in `User`.

caseif:
@gabizou 
1) Noted. I'll revise this method to make a bit more sense practically speaking.

2) I provided further reasoning in a previous comment: "I can't imagine many use cases for this method (or more importantly, its setter) that don't depend on the percentage of the bar filled rather than the particular amount of experience, but I'm open to discussion about it." However, I understand the need for a way to access the actual XP values and will revise my PR accordingly.

3) I've already moved the visibility methods to Living. I understand where you're coming from when you say the method isn't necessary, but I could comprehend a few use cases for such a feature. But, more importantly, this will not be possible without accessing NMS if it is not offered by the API.

4) Noted.

5) Flight can be enabled and disabled for a player independent of gamemode. However, going along with the previous item, it should actually be an attribute as well.

6) Noted.

I'll fix these issues tomorrow, as I'm about to head to bed.

ST-DDT:
Is Flight really an Attribute?
How is it called?

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -57,18 +62,14 @@
...
-     * Returns whether the {@link Player} can fly via the fly key.
+     * Sets the player's display name.
      *
-     * @return {@code True} if the {@link Player} is allowed to fly
-     */
-    boolean getAllowFlight();
-
-    /**
-     * Sets if the {@link Player} can fly via the fly key.
+     * <p>Passing {@link com.google.common.base.Optional#absent()} will set the
```
Minecrell: You need to fix this for `Nullable`, and also add a note about this to the parameter description (like `The new display name, or null to reset it`

ryantheleach:
> that don't depend on the percentage of the bar filled rather than the particular amount of experience, 

I honestly see more usecase for the raw experience values then the percentage. I should be able to treat the experience of a player like a currency to spend on magical stuff. The only thing I see people wanting a percentage for, would be to abuse the experience bar to be some other mechanic.

That said, I don't see why you can't have both.

Deamon5550:
@ryantheleach If you have the raw experience and the target experience for the next level then you can determine the percentage if you need it, no need for us to provide it.

gabizou:
> That said, I don't see why you can't have both.

As much as I'd like the API to be simple to use, I hate to see included redundancy. It's really something not necessary to include `getExperienceToLevel()` with `getExperienceForLevel()` and then possibly include confusion with `getExperienceToLevelRatio()` when it's simply a `double / double`.

caseif:
I understand that the experience methods included in the PR at the moment are bad design choice, and I'll work on revising them as soon as I get the chance.

gabizou:
> Is Flight really an Attribute?

If the Attribute system is designed correctly, it most certainly can represent Flight as an attribute, regardless how it is represented internally in Vanilla Minecraft.

caseif:
@gabizou Expanding on that, I want to clarify that it actually is a field in `PlayerAbilities`, alongside attributes like walk and fly speed.

gabizou:
> Expanding on that, I want to clarify that it actually is a field in PlayerAbilities, alongside attributes like walk and fly speed.

Yes, but again, I don't wish for the API to mimic exactly the representation in Minecraft as that can change at any stage (plus, allows us to eventually move towards a component system).

gabizou:
@caseif Are you going to make the necessary changes?

src/main/java/org/spongepowered/api/entity/player/User.java
```diff
@@ -85,25 +87,27 @@
...
 
     /**
-     * Checks if this player is whitelisted.
+     * Gets the location of this player's bed spawn.
      *
-     * @return True If whitelisted
+     * @return The location of this player's bed spawn, or
+     * {@link Optional#absent} if it is not set
      */
-    boolean isWhitelisted();
```
gabizou: Why were these removed? It's very important they remain.

caseif: @gabizou It's already added by gratimax's [bans branch](https://github.com/SpongePowered/SpongeAPI/tree/feature/bans).

Deamon5550: And by removing them here you will only cause conflicts when that PR is merged in, better to leave those methods untouched for gratimax's branch to change.

caseif: Oh, I see. For whatever reason I had thought that the getters had been added by my PR. I'll read them when I get the chance later today.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -287,4 +288,20 @@
...
+     */
+    boolean isInvisible(Player otherPlayer);
+
+    /**
+     * Sets whether this living entity is invisible to a given player.
+     *
+     * @param otherPlayer The player to toggle visibility of this entity for
+     * @param invisible Whether this entity is invisible to the given player
+     */
+    void setInvisible(Player otherPlayer, boolean invisible);
```
simon816: Is this really in the scope of a `Living` entity?
I don't know where this functionality would go but I don't think it should be here.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
     /**
-     * Gets the current experience towards the next level.
+     * Gets the current food saturation of this human entity.
+     * <p>Food level has health effects, depending on game difficulty and
+     * hunger levels. If the food level is high enough, the human entity
+     * may heal. If the food level is at 0, the human entity may starve.</p>
+     *
+     * @return The current food level
+     */
+    float getFoodLevel();
```
simon816: what about a `getMaxFoodLevel()`?

src/main/java/org/spongepowered/api/entity/living/Living.java
caseif: There's no real reason to restrict it to `Human` or `Player`, and additionally, `Living` holds the existing invisibility methods.

src/main/java/org/spongepowered/api/entity/living/Human.java
caseif: I believe that would be an `Attribute`.

AlphaModder: @caseif No, it would not. Food Level is a value stored in NBT.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-     * depletes, saturation may begin to diminish.</p>
+     * Gets the exhastion value of this human entity.
+     * <p>When exhaustion level reaches 0, saturation may begin to diminish.
+     * </p>
      *
-     * @return The current hunger
+     * @return The current exhaustion level
      */
-    float getHunger();
+    float getExhaustion();
```
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-     * depletes, saturation may begin to diminish.</p>
+     * Sets the exhaustion value of this human entity.
+     * <p>When exhaustion level reaches 0, saturation may begin to diminish.
+     * </p>
      *
-     * @param hunger The hunger level
+     * @param exhaustion The new exhaustion level
      */
-    void setHunger(float hunger);
+    void setExhaustion(float exhaustion);
```
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
+     * Gets the current saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
+     * decrease while the saturation level is greater than zero.</p>
      *
      * @return The current saturation level
      */
     float getSaturation();
```
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-    void setHunger(float hunger);
+    void setExhaustion(float exhaustion);
 
     /**
-     * Gets the current food saturation of this human entity.
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
+     * Gets the current saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
```
gabizou: Insert a new line between `.` and `<p>`.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
+     * Sets the saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
+     * decrease while the saturation level is greater than zero.</p>
      *
-     * @param saturation The saturation level to set
+     * @param saturation The new saturation level
      */
```
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
     float getSaturation();
 
     /**
-     * Sets the current food saturation of this human entity.
-     * <p>Saturation has health effects, depending on game difficulty and
-     * hunger levels. If the saturation is high enough, the human entity
-     * may heal. If the saturation is at 0, the human entity may starve.</p>
+     * Sets the saturation level of this human entity.
+     * <p>Saturation acts as a buffer for food level. Food level will not
```
gabizou: Same as here.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
      *
-     * @param saturation The saturation level to set
+     * @param saturation The new saturation level
      */
     void setSaturation(float saturation);
 
     /**
-     * Gets the current experience towards the next level.
+     * Gets the current food saturation of this human entity.
+     * <p>Food level has health effects, depending on game difficulty and
```
gabizou: And here.

src/main/java/org/spongepowered/api/entity/living/Human.java
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
+    /**
+     * Sets the current food saturation of this human entity.
+     *
+     * <p>Food level has health effects, depending on game difficulty and
+     * hunger levels. If the food level is high enough, the human entity
+     * may heal. If the food level is at 0, the human entity may starve.</p>
+     *
+     * @param foodLevel The new food level
+     */
+    void setFoodLevel(float foodLevel);
```
gabizou: Change this to double.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
+    /**
+     * Gets the current experience accumulated since the last level-up.
      *
      * <p>This is not the total experience this human has.</p>
      *
-     * @return The current experience towards the next level
+     * @return The current experience accumulated since the last level-up.
      */
-    double getExperience();
+    double getExperienceSinceLevel();
```
gabizou: Change this to `int` as experience is all stored as integers.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
-     * Sets the experience accumulated towards the next level.
+     * Sets the experience accumulated since the last level-up.
+     *
+     * <p>This is not the total experience this human has.</p>
      *
-     * @param experience The experience
+     * @param experience The experience accumulated since the last level-up.
      */
-    void setExperience(double experience);
+    void setExperienceSinceLevel(double experience);
```
gabizou: Likewise, to `int`.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -35,58 +35,90 @@
...
+    /**
+     * Gets the experience required since the last level to level up.
+     *
+     * <p>For example, if level 14 reqiures 280 experience points, level 15
+     * requires 315, and the player is at level 14, this method will return 35.
+     * 35.</p>
+     *
+     * @return the experience required since the last level to level up
+     */
+    double getExperienceBetweenLevels();
```
gabizou: As well, `int`.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -109,7 +141,7 @@
      *
      * @return The current total accumulated experience
      */
-    double getTotalExperinece();
+    double getTotalExperience();
```
gabizou: to `int`.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -119,6 +151,37 @@
     void setTotalExperience(double totalExperience);
 
     /**
+     * Returns whether this human entity is sneaking.
+     *
```
gabizou: Possibly mention in a new paragraph that name tags may have decreased visibility, movement may be slowed etc.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -119,6 +151,37 @@
...
+     * @return Whether this human entity is sneaking
+     */
+    boolean isSneaking();
+
+    /**
+     * Sets whether this human entity is sneaking.
+     *
```
gabizou: Likewise, here, about movement and visibility.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -287,4 +288,20 @@
      * @param invisible Whether this entity is invisible or not
      */
     void setInvisible(boolean invisible);
+
+    /**
+     * Returns whether this living entity is invisible to a given player.
+     *
+     * @param otherPlayer The player to toggle visibility of this entity for
+     * @return Whether this entity is invisible to the given player
+     */
```
gabizou: Remove these as it's already been added.

src/main/java/org/spongepowered/api/entity/living/Human.java
```diff
@@ -119,6 +151,37 @@
...
+     * Gets whether this human entity is currently flying.
+     *
+     * @return Whether this human entity is currently flying
+     */
+    boolean isFlying();
+
+    /**
+     * Sets whether this human entity is currently flying.
+     *
+     * <p>Note: this will have no effect if flight is disabled for this human.
```
gabizou: How do you check this? Is it with `GameMode`? Should note this in the javadocs.

caseif: @gabizou Flying can be toggled independent of gamemode via a boolean flag in the `EntityPlayer` class. That being said, I'm not sure of how I would elaborate on this method without leaking implementation details.[

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -294,23 +294,4 @@
...
     void setInvisible(boolean invisible);
-
-    /**
-     * Gets whether this living entity is invisible to the specific
-     * {@link Player} entity.
-     *
-     * @param player The other player to check
-     * @return Whether this entity is invisible to the given player
-     */
-    boolean isInvisibleTo(Player player);
```
gabizou: Why are you removing these changes now?

caseif: [You had said](https://github.com/SpongePowered/SpongeAPI/pull/440#discussion_r25787602) that it had already been added.

kenzierocks: Yes, but that doesn't mean remove it from the `master` branch, which is what you're doing here.

gabizou: Yes, but did you rebase on master? If so, you don't need to make any changes to Living... This is showing you're removing the invisibility methods from Living.

caseif: Oh, my mistake. I'll readd it.

PR #445, opened 2015-02-11
Entity/TileEntity/Player API Completeness

AlphaModder:
This PR just adds a bunch of missing ability to modify Entities, TileEntities, and Players. It still needs some discussion. It also does a large refactor to the Player hierarchy, which is not complete (at the time of posting).

### DIFF
--- a/src/main/java/org/spongepowered/api/block/data/BrewingStand.java
+++ b/src/main/java/org/spongepowered/api/block/data/BrewingStand.java
@@ -24,10 +24,12 @@
  */
 package org.spongepowered.api.block.data;
 
+import org.spongepowered.api.util.Named;
+
 /**
  * Represents a Brewing Stand.
  */
-public interface BrewingStand extends Lockable {
+public interface BrewingStand extends Lockable, Named {
 
     /**
      * Gets the remaining time until the brewing is complete. Will be zero if
--- a/src/main/java/org/spongepowered/api/block/data/Chest.java
+++ b/src/main/java/org/spongepowered/api/block/data/Chest.java
@@ -25,9 +25,11 @@
 
 package org.spongepowered.api.block.data;
 
+import org.spongepowered.api.util.Named;
+
 /**
  * Represents a Chest.
  */
-public interface Chest extends Lockable {
+public interface Chest extends Lockable, Named {
 
 }
--- a/src/main/java/org/spongepowered/api/block/data/CommandBlock.java
+++ b/src/main/java/org/spongepowered/api/block/data/CommandBlock.java
@@ -24,14 +24,16 @@
  */
 package org.spongepowered.api.block.data;
 
-import com.google.common.base.Optional;
 import org.spongepowered.api.text.message.Message;
+import org.spongepowered.api.util.Named;
 import org.spongepowered.api.util.command.source.CommandBlockSource;
+
+import com.google.common.base.Optional;
 
 /**
  * Represents a Command Block.
  */
-public interface CommandBlock extends TileEntity, CommandBlockSource {
+public interface CommandBlock extends Named, CommandBlockSource {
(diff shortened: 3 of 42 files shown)


gabizou:
Several problems with this PR:

1. What?

1. Again, what?

1. Duplication of the API

1. Adding irrelevant TileEntities

1. Adding unnecessary `Named` interface that extends `TileEntity` but applied to entities.

1. Creating unnecessary `HumanInfo` for absolutely no reason.

1. Creating `ConnectedPlayer` idea that is already represented by `Player`.

1. Unecessary [`FlowerPot`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-0f3c7214e046740181b29326b4d3c486R32) creation. It only exists as a TileEntity internally due to implementation limitations in the vanilla code. No reason why it should be a TileEntity.

1. Unecessary [`RecordPlayer`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-b16815f9f449fa3cd5082ac594ff3f1eR34) for most of the same reason as FlowerPot.

1. What in the world is [`Record`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-10a0afaff5f7c3fa4242f11aa8b7ae8bR33)? It doesn't make any sense to have this short of adding a new `Sound` available in the already available API.

1. I'm fairly sure [`ArmorStand`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-2e5865aa7656ad91b4305a37baaf07afR30) already exists in the API.. In fact, IT DOES: [`ArmorStand`](https://github.com/SpongePowered/SpongeAPI/tree/master/src/main/java/org/spongepowered/api/entity/living/ArmorStand.java#L33)

1. [`EquippableSlot`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-1f0cb749b4abf30bd39b6408388e365cR28) leaks implementation details and is an enum. This is out of scope with the PR description and is already being handled via #443 

1. [`Item`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-315feaa3899d9be367faa9a685cdaea4R33) isn't "owned" by any means and it shouldn't be.

1. [`Owned`](https://github.com/SpongePowered/SpongeAPI/pull/445/files#diff-b424e725bec86458c6d44627cb059121R35) relies on string based name ownership, which Minecraft has been moving towards a `UUID` based ownership for a while, as has the rest of the API. 



I've only managed to get through a __*THIRD*__ of the entire change log and refuse to continue pointing out the several outstanding issues with this PR. By far, this PR's biggest problem is redundancy and horrible leak of implementation.

AlphaModder:
@gabizou 
5. Whoops with the Named thing.
6. It's simply the data that applies to real and not real players. Unless i'm interpreting Human wrong, that seems reasonable.
7. If by ConnectedPlayer you mean JoinedUser that's a player that's joined before, but not online. I don't see the problem. Since User can represent non-joined players, it seemed reasonable. 
8. Then what way would there be to manipulate them?
9. ^
10. It's the playing song. It has information about things other than sound, and is indeed seperate from the item in the box.
11. I could swear I looked for that...
12. I was trying not to use int constants in the ArmorStand thing, but that doesn't matter.
13. Item has a thrower. 
14. Owned is like that because that is (stupidly) how the information is stored on those entities. 
This was based on the Chunk and Player.dat formats on the Minecraft Wiki, and was a rough draft at least, why I asked for discussion.

gabizou:
As discussed in IRC, we'll consider these changes over several PRs as there are too many changes to consider for one single PR.

AlphaModder:
@gabizou I think i'll make it into six PRs? Tile ticks, player stuff, living entiies, misc entities, the missing tileentities, and the Named interface.

PR #449, opened 2015-02-13
Fix most current minor issues

Minecrell:
Fixes all current checkstyle warnings in the API and sets it to break the build when violations are found so we can keep the master branch clean from new warnings in the future.

I've also updated the IDE code style configurations to be more exact to the Google style and added a cleanup profile for Eclipse.


src/main/java/org/spongepowered/api/block/BlockProperty.java
```diff
@@ -82,19 +82,19 @@
     Optional<T> getValueForName(String name);
 
     // Subinterface markers for vanilla property types
-    public interface BooleanProperty extends BlockProperty<Boolean> {
```
simon816: Why are these not public?

src/main/java/org/spongepowered/api/event/block/data/FurnaceSmeltItemEvent.java
```diff
@@ -29,7 +29,7 @@
 import org.spongepowered.api.util.event.Cancellable;
 
 /**
- * An event that occurs when a {@link Furnace}
+ * An event that occurs when a {@link Furnace} smelt an item.
```
simon816: "smelt" => "smelts"

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
```diff
@@ -36,8 +36,8 @@
  */
 public interface PermissionService {
 
-    static final String SUBJECTS_USER = "user";
```
simon816: Why aren't these `static final`?

src/main/java/org/spongepowered/api/block/BlockProperty.java
Minecrell: `public` is redundant in interfaces

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
Minecrell: `static final` is redundant in interfaces

Minecrell:
@Zidane Can you take a look at this? (Especially the change to break the build on violations in the API)

Zidane:
@Minecrell 

I'll probably bitch and moan at you later for this but its a solid change solely to keep our code clean. I'll allow it.

Minecrell:
Merged

PR #451, opened 2015-02-13
Make BlockLoc a delegating class like Location

sk89q:
This change changes BlockLoc to be a delegate like Location. All the methods (set block, get block, etc.) are then implemented on Extent rather than BlockLoc.

The general consensus the last time that this was discussed (a few months ago) was to implement this change.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
+++ b/src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
@@ -26,7 +26,16 @@
 package org.spongepowered.api.world.extent;
 
 import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
+import com.google.common.base.Optional;
 import org.spongepowered.api.block.BlockLoc;
+import org.spongepowered.api.block.BlockSnapshot;
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.util.Direction;
+
+import java.util.Collection;
 
 /**
  * A volume containing blocks.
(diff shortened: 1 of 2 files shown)


Soren025:
I feel like overloads for the methods to accept three int values would be useful for when you are manipulating large amounts of blocks at once so you don't end up creating large amounts of Vector3i instances.

sk89q:
Going ahead with this unless someone objects.

pinging @SpongePowered/developers

Zidane:
I almost feel it would be great if Location and BlockLoc could somehow be one class.

modwizcode:
I'm good with a BlockLoc merge into Location or similar. This is not something that needs so much separation. A getBlockX-Z() method set actually makes a lot of sense.

me4502:
I currently don't see the point in having both BlockLoc and Location - a merge would be great

src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
```diff
@@ -51,4 +60,231 @@
...
+    boolean isBlockPassable(Vector3i position);
+
+    /**
+     * Test whether the given face of the block can catch fire.
+     *
+     * @param position The position of the block
+     * @param faceDirection The face of the block to check
+     * @return Is flammable
+     */
+    boolean isBlockFlammable(Vector3i position, Direction faceDirection);
```
Xemiru: Does this value change based on the doFireTick gamerule? (should clarify in docs)

DDoS:
The additions to BlockVolume are way out of date. BlockVolume should only be used to represent the presence of BlockStates organized in a volume. This interface is also used by block buffers, so adding any methods will needlessly complicate their implementation.

Zidane:
@gabizou 

Do you want to finish up this PR by merging BlockLoc and Location together (which I think ought to happen) and update for latest API changes? Unsure if @sk89q will be available currently.

gabizou:
@Zidane what is needed?

DDoS:
I can do it.

DDoS:
Moved to: https://github.com/SpongePowered/SpongeAPI/pull/519

PR #455, opened 2015-02-14
Command Authorization Exception

modwizcode:
Moving from https://github.com/SpongePowered/SpongeAPI/pull/181 since I'm going to be handling the cleanup. 

This is what I was thinking with the changes I wanted made. @Zidane please take a look at the special case for SimpleCommandService.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
+++ b/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
@@ -39,10 +39,7 @@ import org.spongepowered.api.event.message.CommandEvent;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.plugin.PluginManager;
 import org.spongepowered.api.service.event.EventManager;
-import org.spongepowered.api.util.command.CommandCallable;
-import org.spongepowered.api.util.command.CommandException;
-import org.spongepowered.api.util.command.CommandMapping;
-import org.spongepowered.api.util.command.CommandSource;
+import org.spongepowered.api.util.command.*;
 import org.spongepowered.api.util.command.dispatcher.SimpleDispatcher;
 import org.spongepowered.api.util.event.Order;
 import org.spongepowered.api.util.event.Subscribe;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
```diff
@@ -39,10 +39,7 @@
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.plugin.PluginManager;
 import org.spongepowered.api.service.event.EventManager;
-import org.spongepowered.api.util.command.CommandCallable;
-import org.spongepowered.api.util.command.CommandException;
-import org.spongepowered.api.util.command.CommandMapping;
-import org.spongepowered.api.util.command.CommandSource;
+import org.spongepowered.api.util.command.*;
```
gabizou: I thought star imports weren't allowed?

Minecrell: They aren't.

modwizcode: Yeah I've not cleaned the PR up for our guidelines.

src/main/java/org/spongepowered/api/util/command/CommandAuthorizationException.java
```diff
@@ -0,0 +1,80 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Thrown when a command is called without authorization or the required
+ * permissions.
+ */
+public class CommandAuthorizationException extends CommandException {
+
+    private static final long serialVersionUID = 6013332030938687625L;
```
sk89q: Have we actually been implementing these IDs for serialization?

sk89q:
I think this PR is a bit too coupled to permission nodes -- there may be other reasons for lack of authorization.

modwizcode:
I don't see a reason for this to exist anymore.

Xemiru:
If this pushes through then its just making permission checks a little bit more work, having to deal with an exception being thrown instead of just a simple if check.

gabizou:
@modwizcode Closing this as you said there isn't much reason for this to exist. 

As well, it should be noted that (if I remember correctly) @zml2008 is working on a Commands PR for the near future.

PR #456, opened 2015-02-15
Add explosion events

### LINKED ISSUE #416: Missing Events. (Explosion, block to entity event)
There are currently no events that I can see that govern the effects given to an entity when it is exploded upon (knockback/physics, damage) or to blocks (blockDestroyEvent? but what about yield calculation?)

As well as an event for when a block is turned into an entity (Wither/golem spawn / primed tnt)

Or would these events already be covered by other events and not need to be added?

gabizou:
Adds two new events to specifically deal with explosives. Fixes #416 

**EntityExplosionEvent**

Is designed for when an explosive entity has detonated and is about to change blocks in the world. 

**ExplosionPrimeEvent**

Is designed for when an explosive entity is primed to detonate (pre-explosion).

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/EntityExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/EntityExplosionEvent.java
@@ -1,0 +1,64 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/event/entity/EntityExplosionEvent.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Gets the damaging yield of the explosion to affect blocks.
+     *
+     * <p>The higher the yield, the more blocks are broken. The yield
+     * is between 0 and 100.</p>
+     *
+     * @return The damaging yield of the explosion
+     */
+    double getYield();
```
ST-DDT: Does the yield also affect mobs?

An getRadius() method would be useful to get sone metadata about the event.

PR #458, opened 2015-02-16
Add BlockHarvestEvent, changes how silk touch enchant works

simon816:
I have created a new event, `BlockHarvestEvent` see the javadoc. 
This is the result from a previous PR, #422

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -54,6 +54,7 @@ import org.spongepowered.api.event.block.BlockBreakEvent;
 import org.spongepowered.api.event.block.BlockBurnEvent;
 import org.spongepowered.api.event.block.BlockChangeEvent;
 import org.spongepowered.api.event.block.BlockDispenseEvent;
+import org.spongepowered.api.event.block.BlockHarvestEvent;
 import org.spongepowered.api.event.block.BlockIgniteEvent;
 import org.spongepowered.api.event.block.BlockInteractEvent;
 import org.spongepowered.api.event.block.BlockMoveEvent;
@@ -77,6 +78,7 @@ import org.spongepowered.api.event.entity.EntityConstructingEvent;
 import org.spongepowered.api.event.entity.EntityDeathEvent;
 import org.spongepowered.api.event.entity.EntityDismountEvent;
 import org.spongepowered.api.event.entity.EntityDropItemEvent;
+import org.spongepowered.api.event.entity.EntityHarvestBlockEvent;
 import org.spongepowered.api.event.entity.EntityInteractBlockEvent;
 import org.spongepowered.api.event.entity.EntityInteractEntityEvent;
 import org.spongepowered.api.event.entity.EntityInteractEvent;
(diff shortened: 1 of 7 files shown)


me4502:
im not sure about removing the dropped item from the break event, but otherwise I like it.

Deamon5550:
@me4502 At the time of the BlockBreakEvent the dropped items aren't technically known, since that is the point of the BlockHarvestEvent (to determine what blocks should drop).

I wonder though if the BlockHarvestEvent should perhaps not be explicitly cancellable, since (unless I am misunderstanding the point of the harvest event) at that point the BlockBreakEvent has already been thrown and returned uncancelled so at this point it doesn't make much sense to cancel the harvest event (if you wanted no items to drop you could just clear the dropped items). Ehhh on second though perhaps it is less destructive to have it cancellable rather than forcing plugins to just clear the dropped items, makes it easier for another plugin to later uncancel it. Thoughts?

ST-DDT:
So basically you replace the internal silk touch with an event driven one?
Except that one, i like the idea of this PR.

me4502:
I vote keeping it cancellable, it's less destructive.

me4502:
Maybe a way to get what enchantments have affected the output?

Eg, fortune

ST-DDT:
Enchantment is to limited. Some plugins may want to add looting potions. So maybe this is something for our Cause System.
But then we should also use something like AttributeModifiers to keep track of all the changes.
This would also enable use to cancel the silk touch or fortune enchantment alone.

me4502:
Pinging rest of @SpongePowered/developers for opinions.

simon816:
After a draft implementation on top of forge's [`BlockEvent.HarvestDropsEvent`](https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/event/world/BlockEvent.java#L45), I discovered a few potential issues.
Of course we should not be limited to how forge works, but there may be reasons for it among other things.
 - The drop chance is not always used, and sometimes completely ignored
 - The event is not cancellable
 - The event is fired after deciding whether it's a silk touch operation (hence why there is no `setSilkTouch(boolean)` method ATM)
 - Firing the event does nothing, it's only for listening

It'd be useful if someone could evaluate these points to see whether anything should be changed.

me4502:
Our cancellable could just map to emptying the internal list of items in forge.

me4502:
The last commit failed to build on travis

simon816:
@me4502 There's going to be some issues when implementing this that I realized. I think it's because I was copying forge's event too closely.
I'm not sure whether `dropChance` should be part of the event since it's not always taken into account.
I think I should remove the event creation from `SpongeEventFactory` as it cannot be fired manually. (A plugin should either destroy the block or just spawn `Item` entities if they want to get a similar effect)
Was hoping my points above would have got a yay/nay before being merged.

me4502:
Most events do nothing when fired, plugins can use it for support purposes.

modwizcode:
@me4502 What do you mean "most events do nothing when fired"?

me4502:
If a plugin fires a block break event, the block won't break, only the event will be fired.

simon816:
Hmm so why would we want plugins to fire those events manually? It seems like that could confuse any plugins listening to that event.

Xemiru:
could add a check to events to see if its from an actual game event or if its from a plugin. something like vanilla-only=true in the annotation so its not passed. separate pr, though.

me4502:
Let's say a plugin wants to break a block. Before the break it, they send out the event to see if they're allowed to.

Plugins sending events is a very valid use case.

ryantheleach:
Just to add, when plugins have historically thrown fake block place events they usually cancel them at the last possible moment in order to prevent logging plugins from logging the change.

modwizcode:
Yeah I'm concerned about not differentiating between plugin checks and
vanilla events. What about a base event flag for isPlugin() that can be
checked to avoid doing things like adding a bonus when a plugin breaks a
block or something similar?

This needs some thought.

ryantheleach:
This may have been merged, but I'm lost to where it's gone post event refactor. Has it just been removed or is there some other way to access the data, specifically looking at explosion drop rates.

simon816:
@ryantheleach I think what you're after is [`DropItemEvent.Destruct`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/item/inventory/DropItemEvent.java#L79)
> `Called when one or more {@link Item} drops are triggered by an object such as an {@link Entity} or {@link BlockType} destruction.`

PR #466, opened 2015-02-25
Add more entity related events. Enhance several entity related events. 

### LINKED ISSUE #462: Problems with entity events
Over at Pore, we've been focusing on implementing events to the best of our ability within the plugin. However, in doing so, I've noticed a fair number of gaps within SpongeAPI's event system, namely pertaining to entity-related events.
Below are lists of issues I have found.

**Missing events**
- No EntityShootBowEvent
- No EntityUnleashEvent (despite an EntityLeashEvent)
- Nothing resembling an EntityDamageEvent (cannot get damage modifiers)
- No portal-related events (e.g. creation, entrance, exit)
- No event for when a bottle 'o' enchanting is used
- No event for food level changing due to factors other than food consumption
- No event for when a horse jumps (not necessarily needed, but could be helpful)
- No event for when a zombie is destroying a door (may be covered by entity AI system)
- No event for entity despawning
- No sheep-related events (dying, shearing, eating)
- No event for a projectile hitting an entity (would not necessarily be covered by a health change event due to snowballs and such)
- No event for slimes splitting upon death

**Problems with existing events**
- EntityTargetEvent defines no setters
- EntityTeleportEvent defines no setters
- EntityTeleportEvent defines no velocity-related methods
- EntityIgniteEvent defines no methods
- getFlamable and setFlamable in ExplosionPrimeEvent are misspleled
- EntityInteractEvent and children should implement Cancellable
- ProjectileLaunchEvent should implement Cancellable
- EntityExplosionEvent should implement Cancellable
- EntityTameEvent offers no way either to get the tamer

Finally, the getEntity method is ambiguous in certain cases (namely when more than one entity is involved). I would advise overriding and specifically documenting it in events such as these.

gabizou:
Achieves the following from the OCD list #221.

Fixes #462 
- [x] `BiomeType` needs a getter for its name
- [x] `Living` needs a class javadoc
- [x] `PotionEffect#apply` is questionable and should possibly be moved to Living
- [x] `EntityBreedEvent` is missing javadoc
- [x] Missing an `EntityUnleashEvent` to match our `EntityLeashEvent`
- [x] getFlamable and setFlamable in `ExplosionPrimeEvent` are misspelled
- [x] Add potion related events
- [x] `EntityInteractEvent` and children should implement Cancellable
- [x] `EntityTameEvent` offers no way either to get the tamer
- [x] EntityTeleportEvent has no velocity methods (for maintaining or changing the velocity of the player after the teleport)
- [x] `EntityIgniteEvent` getter and setter for fire ticks
- [x] `EntityEnterPortalEvent` and `EntityLeavePortalEvent`
- [x] `ExperienceEvent` is missing a javadoc

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/EntityPotionEffectChangeEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/EntityPotionEffectChangeEvent.java
@@ -1,0 +1,86 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.entity;
+
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.event.cause.CauseTracked;
+import org.spongepowered.api.potion.PotionEffect;
+import org.spongepowered.api.util.event.Cancellable;
+
+import java.util.Collection;
+
+/**
+ * An event when an {@link Entity} gains or removes a {@link PotionEffect}.
+ *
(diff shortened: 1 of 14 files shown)


ST-DDT:
IMO there should be a super interface for `PortalEvent`s.

Because there are several events connected to portals:
* `PortalCreateEvent`
* `PortalDestroyEvent`
* `PortalEnterEvent`
* `PortalExitEvent`
* `PortalTeleportEvent` extends `TeleportEvent`

````java
public interface PortalEvent extends Event {

    Portal getPortal();

}
````

````java
public interface Portal {

    PortalType getPortalType();

    // Optional:
    Location getTarget();

}
````

File: src/main/java/org/spongepowered/api/event/entity/EntityPotionEffectChangeEvent.java [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+     * @return If the potion effect is being added
+     */
+    boolean isBeingAdded();
+
+    /**
+     * Sets the potion effect to be added to the entity.
+     *
+     * @param potionEffect The potion effect to add
+     */
+    void setPotionEffect(PotionEffect potionEffect);
```
ST-DDT: Maybe you should mention, that if you set the `PotionEffect` here while `isBeingAdded() == true`.
This will overwrite the real potion that was actually added.

src/main/java/org/spongepowered/api/world/biome/BiomeType.java
```diff
@@ -79,4 +79,11 @@
      */
     void insertPopulator(Populator populator, int index);
 
+    /**
+     * Gets the name of this biome type.
+     *
+     * @return The name of this biome type
+     */
+    String getName();
```
ST-DDT: I would move this method to the top to the other getters.

src/main/java/org/spongepowered/api/potion/PotionEffect.java
```diff
@@ -25,7 +25,6 @@
 
 package org.spongepowered.api.potion;
 
-import org.spongepowered.api.entity.living.Living;
 import org.spongepowered.api.service.persistence.DataSerializable;
 
 /**
```
ST-DDT: We should add at least some javadocs on this class how to apply the effects.
For example `You can apply it to an entity using {@link Living#apply(PotionEffect)}.`

gabizou:
> IMO there should be a super interface for PortalEvents.

That will have to come in as another PR as I don't believe adding a Teleporter API (involving portals, portal types and other things involved with teleportation) is in the scope of this PR (which is intended to just fix the small OCD issues.

ST-DDT:
@gabizou You are right. It was intended  as a "think about it" (in the future) as well a starting point to check if somebody is interested in something like that. I just don't want to create an issue for every idea i have.

PR #474, opened 2015-03-01
Rework the entity displacement events

gratimax:
This PR creates a new interface, `EntityDisplaceEvent` for `EntityMoveEvent` and `EntityTeleportEvent` to inherit from.

I would also like to add a 'velocity' property to EntityMoveEvent, but that would probably require some input in terms of when the event is fired, and what changing velocity means.


Bammerbom:
Isn't EntityMoveEvent a resource-intensive event?

ryantheleach:
Yes? and?

It is/was in bukkit, but I suspect that the lack of bukkit style wrapping classes will make the sponge equivalents at least perform somewhat better.

Bammerbom:
I thought in a discussion before we came to a conclusion to not have a EntityMoveEvent. But maybe I am wrong?

ryantheleach:
Oh right, I'm getting confused with PlayerMoveEvent.

gratimax:
Closed by 51082ef9e10

PR #480, opened 2015-03-02
Cleanup of buffers and access interfaces

DDoS:
Better setup for the block and biome buffers
Rename WeatherVolume to WeatherUniverse
Add WeatherUniverse to Extent

https://github.com/SpongePowered/Sponge/pull/190

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -142,7 +142,7 @@ import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.gen.Populator;
 import org.spongepowered.api.world.weather.Weather;
-import org.spongepowered.api.world.weather.WeatherVolume;
+import org.spongepowered.api.world.weather.WeatherUniverse;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


src/main/java/org/spongepowered/api/world/extent/BiomeArea.java
```diff
@@ -22,43 +22,50 @@
...
-     * Minecraft are column-based (over the X and Z plane). Therefore, the biome
-     * for all the blocks in a certain column will all be the same and changing
-     * the biome of one block in a column will change the biome for the
-     * entire column.</p>
-     *
      * @param position The position
      * @return The biome
      */
-    BiomeType getBiome(Vector3i position);
+    BiomeType getBiome(Vector2i position);
```
simon816: Using a 3 component vector allows for future-proofing the API which was the reason why it was a `Vector3i`

DDoS: I think it's very unlikely that 3D biomes will ever be required, they use up a lot of memory compared to flat. 

Using 3D coordinates here is misleading because it isn't obvious that the z coordinate is ignored, and could lead to breaks across implementations. A plugin that expects the z coordinate to be ignored and the whole column to be set being used with an implementation that has 3D biomes will be broken as the results of the calls will be vastly different.

If the need for 3D biomes does arise, we can simply add a BiomeVolume interface and implement both; the implementation of the 2D now sets columns instead of a single block.

AlphaModder: @DDoS Why not just have both in BiomeArea? Its not as if having 3D (get, at least) methods has us lose anything. And I would not be unsurprised if a mod already has 3D biomes, memory or not.

DDoS: Because area is 2D and volume is 3D (and universe is arbitrary). Basically, don't add it unless it's requested and don't pretend to have 3D biomes unless you really have them. There's a good solution for adding 3D support over 2D without breaking compatibility, so I suggest not going any further than needed right now.

Deamon5550:
Looks good to me. :thumbsup:

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -1359,16 +1359,17 @@ public static LightningStrikeEvent createLightningStrike(Game game, WeatherVolum
...
      * @return A new instance of the event
      */
-    public static WeatherChangeEvent createWeatherChange(Game game, WeatherVolume weatherVolume, Weather initialWeather, Weather resultingWeather) {
+    public static WeatherChangeEvent createWeatherChange(Game game, WeatherUniverse weatherUniverse, Weather initialWeather,
+            Weather resultingWeather) {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("initialWeather", initialWeather);
-        values.put("weatherVolume", weatherVolume);
+        values.put("weatherVolume", weatherUniverse);
```
Phroa: is there any reason you haven't changed the key here?

DDoS: Didn't notice it, this was done as an automatic refactor. Good catch, I'll fix it.

PR #481, opened 2015-03-04
Expand Persistence API

gabizou:
This implements the commonly used [`DataView`](https://github.com/SpongePowered/SpongeAPI/blob/feature/persistence-memory/src/main/java/org/spongepowered/api/service/persistence/data/DataView.java) as a usable data structure that can be used by both implementation and plugins alike. 

This also expands on the Persistence API by creating a usable `SerializationService` that enables the registration of custom serialization for various `DataSerializable`s, this allows for implementations to register `DataSerializableBuilder`s for various things like `ItemStack`s, `Entity`, etc. This can also be used by plugins to enable serialization and deserialization of their own `DataSerializable`s.

Design Proposal for a usable Persistence API
----------------

Many developers have been eager to use utilities for manipulating Minecraft data with "NBT API
Utility Library API Lib Protocol API Minecraft NBT Library" when in fact, the biggest issue is
that none of these "libraries" really solved any of the problems of manipulating raw data of
entities, blocks, etc. and overall, they have many downsides:

* NBT data is an implementation detail, which involves the developer knowing the (*data format*)
and the volatility of the data they're manipulating, which presents the following problems:
    * Generally, *Minecraft* has a defined data format for many things that can't be changed due to
   backwards compatibility
    * Users of NBT data rely on this as a pseudo *protocol* but they also depend on this to store
     their custom data wherever they please as long as it doesn't break the system
    * Creating custom data involves knowledge of the NBT format and system of how custom data can
  be stored.
* NBT itself is an implementation detail. Creating an API based on NBT and the systems using NBT
means:
    * We can't hope to expand the "customizability" of the storage systems since any NBT API is
  entirely at the whim of the *implementation* of NBT in Minecraft Server.

### Design Goals

The aim of Persistence is to address these shortcomings with the following:

* Design an abstract [data structure](http://en.wikipedia.org/wiki/Data_structure) that
facilitates representing any and all data from any source that can be understood by anything
using the data structure
* Add `DataSerializable` as a marker that something *can* be serialized and deserialized to and
from any `DataSource`.
* Enable utilization of custom `DataSource`s to serialize `DataContainer`s to any storage medium
    * This allows for world data to be serialized and deserialized to alternative storage mediums
 like SQL, MongoDB, H2, etc.
* Create a `SerializationService` that allows dynamic registration of `DataSerializable`s to
allow deserialization of `DataContainer` to the instance of `DataSerializable`
* Facilitate manipulation of `DataSerializable` data with `DataContainer`:
    * Depending on the `DataSerializable`, the `DataContainer` may be mutable and changes may apply
 back to the `DataSerializable` instance, other cases may not
    * Developers can rely on the Persistence API to handle any custom data as necessary
 * The format of the data stored in a `DataContainer` produced from a `DataSerializable` may
 change, but that is all *implementation*

### Let's get Started!

To begin, we have to entirely forget about NBT and the implementation detail of how Minecraft
persists data. Got it? Good.

##### DataSerializable? What's that?

In traditional Java, there is the [`Serializable`](http://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html)
interface. `Serializable` basically defines
that the implementing object can serialize to any ObjectOutputStream and deserialize from any
ObjectInputStream. This is kinda close to what we want to achieve, but it's not at all the best
idea since we don't want to have *instances* of `ItemStack` to deserialize an `ItemStack`.

This is where `DataSerializable` comes in. It presents itself being able to *serialize* all of
its data to a `DataContainer`. The implementation decides the format and content to serialize to
`DataContainer`, but it is always a usable `DataContainer`.

##### Ok, but what is a DataContainer?

`DataContainer` is the abstract data structure that is used throughout the Persistence API. It
allows to storing various data as necessary and can be querried for data with `DataQuery`. The
container has a default implementation that is backed by a map, but again, that is
*implementation* detail.

##### Fine, but tell me about Serialization!
So, we've established that we have `DataSerializable`s that serialize their data to
`DataContainer`s, but how do we use that to store the data to, say, SQL? I present `DataSource`.

`DataSource` is the common man's source for data. It is a representation of a *source* of data.
The implementation of a `DataSource` decides what it can *serialize* and *deserialize*, since not
 all `DataSource`s are meant to handle all data.

##### So, I can use a DataSource, but how do I *get* one?
`DataSourceFactory`. It's also an *implementation* detail, but it is meant for `DataSource`
providers to create, as necessary, `DataSource`s for a particular *source*. Both `DataSource` and
 `DataSourceFactory` are meant to be implemented for any particular cause. It is expected that
 there will be "Persistence Libraries" to create various implementations to have
 `MySQLDataSourceFactory` or `AnvilDataSourceFactory`. The `DataSourceFactory` takes in a
 `ConfigurationNode` that defines where the `DataSource` is aimed at.

##### So now what?
Now, we've established that we can

- Get a `DataContainer` containing all data from a`DataSerializable`
- Get a `DataSource` from an implemenation of `DataSourceFactory` to send a `DataContainer` to
serialize

But now we want to **deserialize** a `DataContainer` into a `DataSerializable`.

##### I present `SerializationService`
`SerializationService` is a "Service" registrar for `DataSerializable`s and their respective
`DataSerializableBuilder`s. In essence, after the game has initialized, various
`DataSerializable`s should have a `DataSerializableBuilder` registered with the
`SerializationService`. The service allows for retrieving a `DataSerializableBuilder` that can
deserialize a `DataContainer` into an instance of `DataSerializable`. This is integral to the
Persistence API as it allows for developers to provide their own custom `DataSerializable`
objects for serialization and deserialization for any `DataSource` medium.

##### Is that it?
As far as Persistence API is concerned, that is the system in a nutshell. Provided that the
registration of all `DataSerializable` with its proper `DataSerializableBuilder` is done,
developers can expect to have minimal issue with creating their own custom `DataSource`s and
their respective `DataSourceFactory`.

Usability
------

The intended use of Persistence API is not simply to allow modifications being made to the
internal data of entities, blocks, etc., it's intended to allow a full API for *persisting* data.

As we come to understand, `DataSerializable`s chuck their data into a `DataContainer` for the
developer to *do anything to their hearts content*.

What this API does allow to do is create a `NBTDataSourceFactory` that allows reading and storing
 data backed by Minecraft's NBT format into a still understandable `DataContainer`. The
 `DataContainer` does not always persist to the stored data, but it does allow for manipulating
 raw data presented by the Minecraft persistence service.

Features *not* included in this PR
-------------
Reading through this description, the following features are intentionally excluded:

- __NBT API__ - As mentioned at the beginning of the description, there is little sense to create
 an `NBTDataContainer` as NBT itself is just another data structure. The `AnvilChunkLoader` to
 become an `AnvilDataSource` where it can contain its own methods to read the raw NBT data into a
  `DataContainer`.
- Representing custom data belonging to entities, blocks, etc. It isn't within the scope of
Persistence to recognize custom data, but it is within the scope of the relative API's to
represent custom data. All Persistence aims to achieve is the ability to fetch the raw data as a
`DataContainer` from any `DataSerializable`. The format of the data is *implementation* detail.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/persistence/data/DataView.java
+++ b/src/main/java/org/spongepowered/api/service/persistence/data/DataView.java
@@ -26,6 +26,8 @@ package org.spongepowered.api.service.persistence.data;
 
 import com.google.common.base.Optional;
 import org.spongepowered.api.service.persistence.DataSerializable;
+import org.spongepowered.api.service.persistence.DataSerializableBuilder;
+import org.spongepowered.api.service.persistence.SerializationService;
 
 import java.util.List;
 import java.util.Map;
@@ -40,7 +42,9 @@ public interface DataView {
 
     /**
      * Gets the parent container of this DataView.
+     *
      * <p>Every DataView will always have a {@link DataContainer}.</p>
+     *
      * <p>For any {@link DataContainer}, this will return itself.</p>
      *
      * @return The parent container
@@ -50,8 +54,10 @@ public interface DataView {
     /**
      * Gets the current path of this {@link DataView} from its root
      * {@link DataContainer}.
+     *
      * <p>For any {@link DataContainer} itself, this will return an
      * empty string as it is the root of the path.</p>
+     *
      * <p>The full path will always include this {@link DataView}s name
      * at the end of the path.</p>
      *
@@ -61,6 +67,7 @@ public interface DataView {
 
     /**
      * Gets the name of this individual {@link DataView} in the path.
+     *
      * <p>This will always be the final substring of the full path
      * from {@link #getCurrentPath()}.</p>
      *
@@ -71,6 +78,7 @@ public interface DataView {
     /**
      * Gets the parent {@link DataView} of this view. The parent directly
      * contains this view according to the {@link #getCurrentPath()}.
+     *
      * <p>For any {@link DataContainer}, this will return an absent parent.</p>
      *
      * @return The parent data view containing this view
@@ -78,10 +86,12 @@ public interface DataView {
     Optional<DataView> getParent();
 
     /**
-     * Gets a set containing all keys in this {@link DataView}.
+     * Gets a collection containing all keys in this {@link DataView}.
+     *
      * <p>If deep is set to true, then this will contain all the keys
      * within any child {@link DataView}s (and their children, etc).
      * These will be in a valid path notation for you to use.</p>
+     *
      * <p>If deep is set to false, then this will contain only the keys
      * of any direct children, and not their own children.</p>
      *
@@ -92,9 +102,11 @@ public interface DataView {
 
     /**
      * Gets a Map containing all keys and their values for this {@link DataView}.
+     *
      * <p>If deep is set to true, then this will contain all the keys and
      * values within any child {@link DataView}s (and their children,
      * etc). These keys will be in a valid path notation for you to use.</p>
+     *
      * <p>If deep is set to false, then this will contain only the keys and
      * values of any direct children, and not their own children.</p>
      *
@@ -152,6 +164,7 @@ public interface DataView {
     /**
      * Creates a new {@link DataView} with the given data at the desired
      * path.
+     *
      * <p>If any data existed at the given path, that data will be overwritten
      * with the newly constructed {@link DataView}.</p>
      *
@@ -163,6 +176,7 @@ public interface DataView {
 
     /**
      * Gets the {@link DataView} by path, if available.
+     *
      * <p>If a {@link DataView} does not exist, or the data residing at
      * the path is not an instance of a {@link DataView}, an absent is
      * returned.</p>
@@ -174,6 +188,7 @@ public interface DataView {
 
     /**
      * Gets the {@link Boolean} by path, if available.
+     *
      * <p>If a {@link Boolean} does not exist, or the data residing at
      * the path is not an instance of a {@link Boolean}, an absent is
      * returned.</p>
@@ -185,6 +200,7 @@ public interface DataView {
 
     /**
      * Gets the {@link Integer} by path, if available.
+     *
      * <p>If a {@link Integer} does not exist, or the data residing at
      * the path is not an instance of a {@link Integer}, an absent is
      * returned.</p>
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/api/service/persistence/data/MemoryDataView.java
```diff
@@ -0,0 +1,591 @@
...
+
+    protected final Map<String, Object> map = Maps.newLinkedHashMap();
+    private final DataContainer container;
+    private final DataView parent;
+    private final DataQuery path;
+
+    protected MemoryDataView() {
+        if (!(this instanceof DataContainer)) {
+            throw new IllegalStateException(
+                    "Cannot construct a root MemoryDataView without a "
```
simon816: No need for such harsh line wrapping, the style guide states it is up to 150 characters.

src/main/java/org/spongepowered/api/service/persistence/data/DataView.java
```diff
@@ -78,23 +86,27 @@
...
      * These will be in a valid path notation for you to use.</p>
+     *
      * <p>If deep is set to false, then this will contain only the keys
      * of any direct children, and not their own children.</p>
      *
      * @param deep Whether or not to get all children keys
      * @return A set of current keys in this container
      */
-    Set<DataQuery> getKeys(boolean deep);
+    Collection<DataQuery> getKeys(boolean deep);
```
simon816: Keys will always be unique, shouldn't this stay as a `Set`?

src/main/java/org/spongepowered/api/service/persistence/data/MemoryDataView.java
```diff
@@ -0,0 +1,592 @@
...
+ */
+public class MemoryDataView implements DataView {
+
+    protected final Map<String, Object> map = Maps.newLinkedHashMap();
+    private final DataContainer container;
+    private final DataView parent;
+    private final DataQuery path;
+
+    protected MemoryDataView() {
+        if (!(this instanceof DataContainer)) {
```
Minecrell: Can you use `checkState` from Preconditions here?

PR #484, opened 2015-03-04
More work on block and biome access interfaces

DDoS:
- Use BlockState for BlockVolume instead of BlockLoc.
- Make mutable buffers implement their corresponding access interface (hence the previous change).
- Use Vector3i instead of Vector3d for block access.
- Move BlockLoc getters to Extent under the name getFullBlock. Since a BlockLoc requires an Extent, it makes sense that the getters be in Extent instead of BlockVolume (a BlockVolume doesn't necessarily have an Extent).

https://github.com/SpongePowered/Sponge/pull/191

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
+++ b/src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
@@ -25,8 +25,8 @@
 
 package org.spongepowered.api.world.extent;
 
-import com.flowpowered.math.vector.Vector3d;
-import org.spongepowered.api.block.BlockLoc;
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.block.BlockState;
 
 /**
  * A volume containing blocks.
@@ -39,7 +39,7 @@ public interface BlockVolume {
      * @param position The position
      * @return The block
      */
-    BlockLoc getBlock(Vector3d position);
+    BlockState getBlock(Vector3i position);
 
     /**
      * Get a representation of the block at the given position.
@@ -49,6 +49,24 @@ public interface BlockVolume {
      * @param z The Z position
      * @return The block
      */
-    BlockLoc getBlock(int x, int y, int z);
+    BlockState getBlock(int x, int y, int z);
+
+    /**
+     * Sets the block at the given position in the world.
+     *
+     * @param position The position
+     * @param block The block
+     */
+    void setBlock(Vector3i position, BlockState block);
+
+    /**
+     * Sets the block at the given position in the world.
+     *
+     * @param x The X position
+     * @param y The Y position
+     * @param z The Z position
+     * @param block The block
+     */
+    void setBlock(int x, int y, int z, BlockState block);
 
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
```diff
@@ -39,7 +39,7 @@
      * @param position The position
      * @return The block
      */
-    BlockLoc getBlock(Vector3d position);
+    BlockState getBlock(Vector3i position);
```
Zidane: Why wouldn't this be BlockLoc?

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -32,4 +34,22 @@
  */
 public interface Extent extends EntityUniverse, WeatherUniverse, BlockVolume, BiomeArea {
 
+    /**
+     * Get a representation of the block at the given position.
+     *
+     * @param position The position
+     * @return The block
+     */
+    BlockLoc getFullBlock(Vector3i position);
```
Zidane: Why "FullBlock"?

src/main/java/org/spongepowered/api/world/extent/BlockVolume.java
DDoS: Because BlockLoc requires an Extent, but a BlockVolume is a super-interface of Extent. This makes BlockVolume essentially useless unless used with Extent. Using BlockState means that buffers can now implement this interface.

src/main/java/org/spongepowered/api/world/extent/Extent.java
DDoS: BlockState contains most of the block information, but is location independent and doesn't do stuff like chest inventories. Full block returns a BlockLoc, which is a state with a location that encapsulates all the block information. It's the best nomenclature I could come up considering that I wanted BlockVolume to keep its method names as "getBlock".

Zidane: Might be better to call this getBlock and the one above getBlockState perhaps.

DDoS: I was originally going to do this, but that would mean that BlockVolume has getBlockState instead of getBlock, which I find odd since it's a BlockVolume, not a BlockStateVolume. I don't think BlockLoc should be considered a Block, but instead an extension of it. BlockState should be what getBlock returns.

Zidane: Explanation is fair

kenzierocks: Perhaps `BlockState getBlockSnapshot` and `BlockLoc getBlockExtended`? Describes it slightly better, I just dislike the wording of "full block."

DDoS: getBlockExtended is still vague, it should be something more like getBlockLocalized, but I find getFullBlock to be more concise, and since BlockLoc is just the equivalent of the Bukkit Block, devs should understand the idea just fine. Also BlockState isn't a snapshot, that's BlockSnapshot.

PR #487, opened 2015-03-06
Return subcommand suggestions. Fixes #472

### LINKED ISSUE #472: Tab complete - SimpleDispatcher
In SimpleDispatcher.java, we are getting a list of tab complete options but never actually adding them to the result list. We should change line [341](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java#L341) from:
```java
mapping.get().getCallable().call(source, parts.length > 1 ? parts[1] : "", Collections.unmodifiableList(passedParents)));
```
to:
```java
suggestions.addAll(mapping.get().getCallable().call(source, parts.length > 1 ? parts[1] : "", Collections.unmodifiableList(passedParents));
```

boformer:
This fixes the issue that the ``SimpleDispatcher`` does not return TAB complete suggestions that were generated by a subcommand.

### DIFF
--- a/src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
+++ b/src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
@@ -29,6 +29,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
+import com.google.common.base.Joiner;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -338,10 +339,25 @@ public synchronized boolean testPermission(CommandSource source) {
...
+                            //Join the alias and the suggestion, then add it to the suggestion list
+                            suggestions.add(joiner.join(alias, suggestion));
+                        }
+                        
+                        break;
+                    }
+                }
             }
         }
-
```
me4502: This change here isn't necessary.

boformer:
This could also be an issue: https://github.com/SpongePowered/Sponge/issues/192
With this kind of TAB completion, there are no permission checks. I will fix it today.

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -327,6 +328,9 @@ public synchronized boolean testPermission(CommandSource source) {
 
             synchronized (this) {
                 for (CommandMapping mapping : this.commands.values()) {
+                    // Skip commands the source is not permitted to call
+                    if(mapping.getCallable().testPermission(source)) continue;
```
ST-DDT: Is there a NOT missing here?
The comment above hints that.

Minecrell: There is also a space missng after the if, maybe reformat your code to fix it

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -337,8 +341,25 @@ public synchronized boolean testPermission(CommandSource source) {
...
+            // Check if subcommand exists and source is permitted to call it
+            if (mapping.isPresent() && mapping.get().getCallable().testPermission(source)) {
+                // A list of suggestions delivered by the subcommand
+                List<String> subSuggestions = mapping.get().getCallable().getSuggestions(source, parts.length > 1 ? parts[1] : "");
+
+                // parts[0] is the alias how it was entered by the source, e.g. "sENd"
+                // Find the alias how it was registered, e.g. "Send"
+                for (String alias : mapping.get().getAllAliases()) {
+                    if (alias.equalsIgnoreCase(parts[0])) {
+                        Joiner joiner = Joiner.on(' ');
```
Minecrell: There is no need to initialize a new `Joiner` on every iteration, make it a constant in the class

me4502:
This is looking quite nice

Zidane:

me4502:
I feel this is mergable.

PR #425, opened 2015-01-30
Text API 2

gratimax:
Big thanks to @Minecrell for his work here. I just wanted to get a PR to resolve this ASAP and get it merged.

Some new features:

- Name of Message/Text has finally been decided on. It will be Text, with Message.Text being Text.Literal
- The generic getContent method has been removed
- A lot of the factory methods have been moved to GameRegistry
- Some more Translatable interfaces
- Javadocs

This fixes #287, #263 and #435.

PR #488, opened 2015-03-06
Add CatalogedBy annotation to get the Catalog of fake enum types

gabizou:
This adds a new annotation that can be used for marking various *types* as *pseudo enumeration types* such that they have a parent class that contains all registered values of that pseudo enum type.

This has repercussions of aiding other utility classes such as [`Coerce`](https://github.com/SpongePowered/SpongeAPI/blob/feature/pseudoenums/src/main/java/org/spongepowered/api/util/Coerce.java#L370). 

Note: This PR does NOT cover the Text API classes as that is due for changes with #425. 

The primary reason for this change is that since our API has avoided using [Java Enums](http://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) due to their limitation of expandability, we have dozens of interfaces signifying a "type" of "something". The problem arises when developers wish to find the available "values" of that "type".

So, the `CatalogedBy` annotation marks the "type" class with a catalog class that contains all default values of that "type". Example:

```java
@CatalogedBy(Itemtypes.class)
public interface ItemType {
    String getId();
}

public final class ItemTypes {
   public static final ItypeType DIAMOND_SWORD = null;
   //... and more
}
```

We can therefor do the following:

```java
public <T> T toPseudoEnum(Class<T> annotated, Object random) {
    Annotation catalog = annotated.getAnnotation(CatalogedBy.class);
    if (catalog instanceof CatalogedBy) {
        Class<?> clazz = ((CatalogedBy) catalog).value();
        // do things with reflection and fields here
        // eventually finding out if the random is of a type of T
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/api/attribute/Attribute.java
+++ b/src/main/java/org/spongepowered/api/attribute/Attribute.java
@@ -26,11 +26,13 @@
 package org.spongepowered.api.attribute;
 
 import com.google.common.base.Predicate;
+import org.spongepowered.api.util.annotation.CatalogedBy;
 
 /**
  * Represents a type of attribute that can be applied to an {@link
  * AttributeHolder}.
  */
+@CatalogedBy(Attributes.class)
 public interface Attribute {
 
     /**
--- a/src/main/java/org/spongepowered/api/attribute/Operation.java
+++ b/src/main/java/org/spongepowered/api/attribute/Operation.java
@@ -25,10 +25,13 @@
 
 package org.spongepowered.api.attribute;
 
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
 /**
  * Represents a function used by an {@link AttributeModifier} to modify the
  * value of an {@link Attribute}.
  */
+@CatalogedBy(Operations.class)
 public interface Operation extends Comparable<Operation> {
 
     /**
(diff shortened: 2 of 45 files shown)


ST-DDT:
I can see the benefit for this during development, but why do you need it at runtime?
Maybe some kind of autocomplete functionality?

gabizou:
> but why do you need it at runtime?

Did you read the description?

simon816:
The following classes recently added to the API will need the annotation:
[`AttributeTargets`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/attribute/AttributeTargets.java)
[`Attributes`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/attribute/Attributes.java)
[`Operations`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/attribute/Operations.java)

gratimax:
Should be `@CatalogedBy` or `@HasCatalog`, imo. I personally prefer the second one, since 'cataloged' as a verb looks pretty weird.

ST-DDT:
Does the class mentioned in `@CatalogedBy` must contain all instances/references of the annotated class?

gratimax:
@ST-DDT We're assuming that all _sponge_-specific instances of a specific types are cataloged by a specific catalog. Sure, when client-side Sponge arrives and allows adding new blocks, this guarantee won't be held up.

Zidane:
@gabizou 

Get this PR fixed up for merge.

gabizou:
@Zidane Do you believe the `ManufacturedBy` should stay? It doesn't quite provide any use that we could find.

Zidane:
@gabizou Not thinking of one either so remove it.

gabizou:
@Zidane Done.

Zidane:
@gabizou 

Can't merge this again.

PR #491, opened 2015-03-06
Add an API for Scheduled Block Updates

AlphaModder:
This is a fairly simple PR. It adds support for "Tile Ticks", renamed as Scheduled Block Updates. They are documented [here](http://minecraft.gamepedia.com/Chunk_format#Tile_tick_format).


src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -25,14 +25,15 @@
...
 import org.spongepowered.api.block.BlockLoc;
 import org.spongepowered.api.world.weather.WeatherUniverse;
 
+import com.flowpowered.math.vector.Vector3i;
+
 /**
  * Contains blocks, entities, biomes and weather.
  */
-public interface Extent extends EntityUniverse, WeatherUniverse, BlockVolume, BiomeArea {
+public interface Extent extends EntityUniverse, WeatherUniverse, BlockVolume, BiomeArea, TileTickVolume {
```
kenzierocks: If TileTickVolume extends BlockVolume, can that be removed from this list?

AlphaModder: @kenzierocks It can, but I elected to leave it to make it easier to understand Extent.

gratimax:
You could have perhaps looked at #475  before you made this PR...

AlphaModder:
@gratimax Why? It doesn't seem to be adding TileTicks...

kenzierocks:
Naming is a bit confusing, I propose a rename to BlockTicks.

src/main/java/org/spongepowered/api/world/TileTick.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.flowpowered.math.vector.Vector3i;
+
+/**
+ * 
+ * Represents a tile tick. Tile ticks cause a block to update after a certain
+ * number of ticks. Tile ticks with higher priorities are processed first.
+ *
+ */
+public interface TileTick {
```
bloodmc: Mojang calls their Block class "Tile", we do not. A better name would be ScheduledBlockUpdate.

gabizou:
I have multiple worries about adding something like this:

- What is the point of a `TileTick`? I understand that it's to represent the *implementation detail* about Minecraft's `TileEntity` ticking system, but does that really belong in API?

- `TileTick` is not at all clear as to what the purpose of priority is. What is priority based on? The larger the number, the higher priority? What *IS* priority used for? Is there only one `TileTick` processed per game tick?

- What is `TileTick` used for in the API? You can't listen to them, and there isn't a clear reason to use them except implying that some tile entities should process faster and possibly break the client?

AlphaModder:
@gabizou This has nothing to do with TileEntities. It causes a block update after n ticks, which is why I'm renaming it to ScheduledBlockUpdate. If multiple tileticks are to be processed in one server tick, higher priorites go first.

src/main/java/org/spongepowered/api/world/TileTick.java
AlphaModder: @bloodmc Yes, this is what I will use.

gabizou:
@AlphaModder Please rename your PR and description for the renaming update from TileTick to ScheduledBlockUpdate.

AlphaModder:
@gabizou I did.

src/main/java/org/spongepowered/api/block/ScheduledBlockUpdate.java
```diff
@@ -0,0 +1,82 @@
...
+    void setPriority(int priority);
+
+    /**
+     * Gets the position of this scheduled block update.
+     * 
+     * @return the position of this scheduled block update
+     */
+    Vector3i getPosition();
+
+}
```
gratimax: missing newline

src/main/java/org/spongepowered/api/block/ScheduledBlockUpdate.java
```diff
@@ -0,0 +1,82 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.block;
+
+import com.flowpowered.math.vector.Vector3i;
+
+/**
+ *
```
Minecrell: These empty lines before the description are not necessary

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -25,10 +25,11 @@
 
 package org.spongepowered.api.world.extent;
 
-import com.flowpowered.math.vector.Vector3i;
 import org.spongepowered.api.block.BlockLoc;
 import org.spongepowered.api.world.weather.WeatherUniverse;
 
+import com.flowpowered.math.vector.Vector3i;
```
Minecrell: The import order was correct and you broke it now

AlphaModder: @Minecrell Aaaargh... why does this keep happening? My eclipse formatter is set to only format lines I edit!

Minecrell: Why not just import our code formatter config and import order? Then Eclipse should place the imports in the correct order automatically

AlphaModder: @Minecrell That's what I've done...

AlphaModder: @Minecrell Yeah, using the sponge format I've been given does it this way...

Minecrell: Are you sure? You need to import import order, cleanup and formatter separately - if that was done correctly it shouldn't screw up import anymore, at least not after you run Organize Imports for the file

AlphaModder: @Minecrell I just realized they were seperate things >.<

src/main/java/org/spongepowered/api/block/BlockLoc.java
```diff
@@ -25,13 +25,14 @@
...
 package org.spongepowered.api.block;
 
-import com.flowpowered.math.vector.Vector3i;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.service.persistence.data.DataHolder;
 import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.extent.Extent;
 
+import com.flowpowered.math.vector.Vector3i;
```
Minecrell: The import order is wrong here too

src/main/java/org/spongepowered/api/world/TileTick.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Gets the block type used to update the block.
+     * 
+     * @return The block type used to update the block
+     */
+    BlockType getBlockType();
```
simon816: Could remove the need for `getBlockType` and `getPosition` if instead there was a `BlockLoc getBlock()` method

src/main/java/org/spongepowered/api/block/ScheduledBlockUpdate.java
```diff
@@ -0,0 +1,80 @@
...
+ * processed first.
+ */
+public interface ScheduledBlockUpdate {
+
+    /**
+     * Gets the block type used to update the block.
+     * 
+     * @return The block type used to update the block
+     */
+    BlockType getBlockType();
```
simon816: Could remove the need for `getBlockType` and `getPosition` if instead there was a `BlockLoc getBlock()` method

me4502:
I agree with @simon816, having BlockLoc (or even BlockState if the state is important) would be more beneficial.

PR #492, opened 2015-03-06
Add events for interaction with hanging entities

caseif:
This PR adds the following events pertaining to hanging entities:

- `EntityInteractHangingEvent`
- `LivingInteractHangingEvent`
- `HumanInteractHangingEvent`
- `PlayerInteractHangingEvent`
- `BlockBreakHangingEvent`


src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -144,6 +144,15 @@
 import org.spongepowered.api.world.weather.Weather;
 import org.spongepowered.api.world.weather.WeatherUniverse;
 
+import com.flowpowered.math.vector.Vector3d;
```
gabizou: These imports are in the wrong order.

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -200,6 +209,15 @@ public static BlockBreakEvent createBlockBreak(Game game, Cause cause, BlockLoc
         return createEvent(BlockBreakEvent.class, values);
     }
 
+    public static BlockBreakHangingEvent createBlockBreakHanging(Game game, Cause cause, BlockLoc block, Hanging hanging) {
```
gabizou: Javadocs.

src/main/java/org/spongepowered/api/event/block/BlockBreakHangingEvent.java
```diff
@@ -0,0 +1,40 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.block;
+
+import org.spongepowered.api.entity.hanging.Hanging;
+
+/**
+ * Called when a hanging entity is broken by a block change.
+ */
+public interface BlockBreakHangingEvent extends BlockEvent {
```
gabizou: Can we make this cancelable?

caseif: Technically we can, but I would be opposed to it seeing as the hanging would be in an invalid state if the event were cancelled.

src/main/java/org/spongepowered/api/event/entity/EntityInteractHangingEvent.java
```diff
@@ -0,0 +1,40 @@
...
+import org.spongepowered.api.entity.hanging.Hanging;
+
+/**
+ * Called when an entity interacts with a hanging entity.
+ */
+public interface EntityInteractHangingEvent extends EntityInteractEntityEvent {
+
+    /**
+     * Gets the {@link Hanging} which has been interacted with.
+     * @return The {@link Hanging} which has been interacted with.
```
Minecrell: Missing empty line and there should be no period at the end of the return tag

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -201,6 +209,23 @@ public static BlockBreakEvent createBlockBreak(Game game, Cause cause, BlockLoc
     }
 
     /**
+     * Creates a new {@link BlockBreakHangingEvent}.
```
Minecrell: Add an empty line between the description and the Javadoc tags here

Deamon5550:
Are not *InteractHangingEvents just special cases of the *InteractEntityEvents, seems rather superfluous to have this kind of expansion for targets as well as sources. Whats next an EntityInteractZombieEvent?

As for the `BlockBreakHangingEvent` would not a more generic event with a cause attached to it be more useful. Or perhaps just as a cause from an EntityExpireEvent as that is the purpose of that already existing event. While a level of specificity is good there is a balance to be maintained.

caseif:
@Deamon5550 I spoke to @bloodmc and @gabizou about this change before starting on it, and the consensus we came was essentially that interactions involving hangings have a special behavior in that the hanging can be despawned and dropped along with the item it contains. The other part of our reasoning was that it would be overly convoluted and tedious to listen to the equivalents of these events if they were not explicitly called.

gabizou:
@caseif While I said it sounded good and give it a shot, I was curious since there wasn't much opposition at the time and I wasn't taking into consideration the maintainability of it all.

As @Deamon5550 did explain, I too mentioned that if an EntityExpireEvent is fired for a Hanging item, well, there is going to be a cause attached. If an entity attacks a hanging entity and causes it to pop off, there is an event for that. 

I must say that filtering through the types of entities by narrowing the focus of events by adding more events is opposite of maintainability in my opinion.

Overall, I'm not against adding a some more events *if the case is properly justified*, but the narrow-mindedness of these events aren't justified, in my opinion.

gabizou:
@caseif Are you going to reply or leave this PR to sit and collect dust and eventually closed?

caseif:
@gabizou After considering your previous comment, I can understand why these events would be redundant.

PR #493, opened 2015-03-07
Fix Eclipse formatter

boformer:
Right now the Eclipse formatter profile is not usable because it changes almost every javadoc entry like this:

Input:
```java
 * <p>In older versions of Minecraft, block properties were encoded in
 * a block's metadata.</p>
```
Output with the current formatter:
```java
 * <p>
 * In older versions of Minecraft, block properties were encoded in a block's
 * metadata.
 * </p>
```

This PR sets the "Format HTML in comments" option to false.

Output with the changed formatter:
```java
 * <p>In older versions of Minecraft, block properties were encoded in a block's
 * metadata.</p>
```

Another problem is that many javadocs were wrapped although the column width was not reached. When I run the formatter, almost every file changes because of this.

I guess this is related to the rule "Feel free to wrap when it will help with readability" in the Contribution Guidelines.

### DIFF
--- a/extra/eclipse/sponge_eclipse_formatter.xml
+++ b/extra/eclipse/sponge_eclipse_formatter.xml
@@ -236,7 +236,7 @@
 <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_constructor_declaration" value="do not insert"/>
 <setting id="org.eclipse.jdt.core.formatter.blank_lines_before_imports" value="1"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_colon_in_assert" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.comment.format_html" value="true"/>
+<setting id="org.eclipse.jdt.core.formatter.comment.format_html" value="false"/>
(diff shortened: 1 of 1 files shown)


Zidane:
@Minecrell 

Yes or no to this?

Minecrell:
@Zidane It seems good to me, although I'm only primarily using the IntelliJ formatter configuration so I'm not too sure about the Eclipse configuration.

bloodmc:
@Minecrell Why would we not want to format javadoc HTML tags? Also, this really isn't a "fix" but rather something that hasn't been applied yet to API hence the problem.

Mumfrey:
> @Minecrell Why would we not want to format javadoc HTML tags?

I can answer this one: because much as I <3 Eclipse, even I have to admit that one thing it is diabolically bad at is formatting HTML. If it came to a choice of letting Eclipse auto-format your HTML, or connecting a pseudo-random number generator to your keyboard, the RNG would win in terms of generating sensible javadoc layout.

bloodmc:
@Mumfrey True, it definitely isn't the greatest. I wonder if there is a plugin that can handle it better.

Mumfrey:
Probably, but it's up to end users if they want to do that. Setting the defaults in the configuration such that running the auto-formatter won't grenade the any HTML javadoc seems like a sensible move.

PR #496, opened 2015-03-09
Make EntityType generic for its corresponding entity class

simon816:
One of the things I thought could be improved slightly is to remove the need to cast when creating an entity instance.
Given this example:
```java
Optional<Entity> optional = world.createEntity(EntityTypes.ZOMBIE, somePosition);
[...]
Zombie zombie = (Zombie) optional.get(); // 'unsafe' cast
```
I would prefer the entity type to be known automatically (by the IDE)
So `optional` becomes `Optional<Zombie> optional  = [...]`
It would make it easier to know the corresponding `Entity` class directly from using `createEntity`
also, you wouldn't need to know the class, for example it may not be obvious that the `COMPLEX_PART` entity is an `EnderDragonPart`. Using generics this can be known.


gabizou:
Wouldn't an issue of this arrive with type erasure and excessive generics to compare `EntityType`s?

While sure it'd be nice to not have to have no casting, but just chucking generics onto `EntityType` isn't the way to go, in my opinion.

PR #500, opened 2015-03-14
Fix StatisticGroups

ST-DDT:
FIxes #220

The currently suggested `StatisticGroup` layout has some serious disadvantages.

In minecraft there are no groups, but they are quite useful if defined properly.

The old suggestion defines the usage something similar to this:
````java
Map<Statistic, Long> stats = player.getStatisticsByGroup(group);
for (Entry entry : stats) {
    Statistic stat = entry.getKey();
    player.sendMessage(asText(stat) + stat.getFormat().or(group.getDefaultFormat()).format(entry.getValue());
}
````
I considers this malicious and error prone.
So i want to change this to simplyfy the process:
````java
Map<Statistic, Long> stats = player.getStatisticsByGroup(group);
StatisticFormat formatter = group.getFormat();
for (Entry entry : stats) {
    Statistic stat = entry.getKey();
    player.sendMessage(asText(stat) + formatter.format(entry.getValue());
}
````

So basically every statistic group defines a fix format all members use. There are no exceptions from that rule. This also allows a more useful displaying because different categories are not merged randomly.
(First show the general_count, that general_fractional, then the distances, than times spend, or however you want it)

This PR removes the format from the statistic and splits the general group into smaller pieces to ensure that each group only contains stazs with the same format.

As an alternative there are 2 other ways that can be used independently from each other.
- Remove only the Optional from `Statistic.getFormat()`
  -  Because in minecraft every statistic has its own reference to a formatter
- Remove the general group
  - Because the group has nothing in common except that they don't have anything in common

### DIFF
--- a/src/main/java/org/spongepowered/api/stats/Statistic.java
+++ b/src/main/java/org/spongepowered/api/stats/Statistic.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.api.stats;
 
-import com.google.common.base.Optional;
 import org.spongepowered.api.text.translation.Translatable;
 
 /**
@@ -38,15 +37,6 @@ public interface Statistic extends Translatable {
      * @return The String ID
      */
     String getId();
-
-    /**
-     * Gets the {@link StatisticFormat} of this statistic. If this is not
-     * present that this statistic's format is deferred to its group's default
-     * format.
-     *
-     * @return The format of this statistic, if available
-     */
-    Optional<StatisticFormat> getStatisticFormat();
 
     /**
      * Gets the {@link StatisticGroup} this {@link Statistic} belongs to.
--- a/src/main/java/org/spongepowered/api/stats/StatisticBuilder.java
+++ b/src/main/java/org/spongepowered/api/stats/StatisticBuilder.java
@@ -27,8 +27,6 @@ package org.spongepowered.api.stats;
 
 import org.spongepowered.api.text.translation.Translation;
 
-import javax.annotation.Nullable;
-
 /**
  * Represents a builder interface to create new and custom instances of
  * {@link Statistic}s.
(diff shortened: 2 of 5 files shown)


Deamon5550:
`'malicious and error prone'` ... please its a bit more verbose but it also gives more information and control.

I disagree with this for these reasons:
- Does not allow hidden or undisplayed statistics due to the removal of the GENERIC group.
- Does not allow Statistics to deviate from the groups format.
- If removing the optional from statistic.getFormat then this hides whether the format is from the group or just the statistic.
- Forces the use of null to define statistics not in those groups, which is counter to the pattern used in the rest of the API concerning null.

The largest issue here is the removal of the GENERIC group, the statistics in this group do have something in common, they are the 'generic' statistics, join times, number of deaths etc. for the player. Removing this forces every other statistic into that 'generic' group. What about the hidden statistics which back achievements, would a system without knowledge display these along with the other 'generic' statistics, this is counter to the intended action for these hidden statistics.

ST-DDT:
> Does not allow hidden or undisplayed statistics due to the removal of the GENERIC group.

I did not remove it. I just split it into smaller pieces. The previous system did not allow hidden or undisplayed statistics either. If im wrong please give me an example.

> Does not allow Statistics to deviate from the groups format.

Why should it in the first place. The group is thete to contain only elements of the same type. Thats why i added the missing groups.

> If removing the optional from statistic.getFormat then this hides whether the format is from the group or just the statistic.

My current PR removes the format entirely. Because its always the same as the groups.
If i just removed the optional then i would  justify it with. You can still check it you just have to use `==` instead of `isPresent()`.

> the statistics in this group do have something in common, they are the 'generic' statistics, join times, number of deaths etc. for the player.

And a subset of this group has more in common than the rest. Thats why i split it into smaller groups to increase the parts they have in common. This way i also avoid sorting by format before sorting by name or something similar.

> Forces the use of null to define statistics not in those groups, which is counter to the pattern used in the rest of the API concerning null.

I did not remove the General group. We avoid null as return type not as method argument.

> What about the hidden statistics which back achievements, would a system without knowledge display these along with the other 'generic' statistics, this is counter to the intended action for these hidden statistics.

They would either not be returned or have their own secret group. I did not change anything related to this topic.

> please its a bit more verbose but it also gives more information and control.

Yes, its more verbose. As mentioned above it does not give any additional infos. And the only controll it does give you (in addition to my suggestion) is to use it the wrong way.

I maybe create another PR with an alternative solution to this issue. But this one sounds better to me thats why i suggested this one first.

PR #503, opened 2015-03-15
Repair Travis Builds

gabizou:
Currently, travis cannot build any branches, PR and even master of SpongeAPI due to some odd interaction with their Docker containers killing the build processes. 

This removes the `sudo: false` line enabling container builds, but this allows branches to build again on travis without issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: java
 jdk:
   - openjdk6


Minecrell:
This will slow down the Travis builds a lot :/

gabizou:
> This will slow down the Travis builds a lot :/

Yes, but at least our branches will build again. Nothing can build with the Docker containers at the moment, and Travis developers are out for the weekend (understandably).

kenzierocks:
http://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error

gabizou:
> http://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error

Our builds don't consume more than 3GB of memory, we've checked, and the builds don't have more than two "tasks". Please be considerate when suggesting a thread that doesn't exactly apply to our situation (that link specifically speaks about tools with parallelism with other languages/setups that aren't applicable to our builds).

kenzierocks:
Okay, sorry I even spoke.

Minecrell:
We had the same problem already a while ago in Sponge, but for some reason only with OpenJDK6, that's why it is also running on the old infrastructure already: https://github.com/SpongePowered/Sponge/blob/master/.travis.yml#L4

gabizou:
> We had the same problem already a while ago in Sponge, but for some reason only with OpenJDK6, that's why it is also running on the old infrastructure already: https://github.com/SpongePowered/Sponge/blob/master/.travis.yml#L4

Ok, but in this case, javadocs too is failing not only on jdk6, but open7 and oracle jdk7. The only JDK that doesn't fail *always* is Oracle's JDK8, which occasionally fails for the same odd reason that the other three fail. 

Personally, I prefer faster builds, but if it's at the cost of randomly failing without rhyme or reason, I'd rather have stability with automated build hooks.

gabizou:
> Okay, sorry I even spoke.

It's not that you spoke, it's just the demeanor of simply posting a link without further looking into the reasoning of the PR. Personally, I find it rude to simply drop a link to an article to a PR that has to have taken some more time to consider making than it took for you to find that link. I wouldn't have replied tearing you down if you provided some simple context as to *why* you posted the link.

You could've said something along the lines of:

> I don't know if it's relevant to the issue, but I found this article after a minute googling about Travis builds suddenly getting killed. Link to the article: http://foo.bar.baz/

gabizou:
> We had the same problem already a while ago in Sponge

Which btw is now affecting Sponge as well.

kenzierocks:
I just remembered it, that's all.

Minecrell:
It seems like it is simply allocating not enough memory for the Javadoc tool, and after fixing that the same for checkstyle. It _seems_ to work fine on the new container infrastructure when allocating some more memory like this: https://github.com/SpongePowered/SpongeAPI/compare/master...Minecrell:Tmp

Travis build: [Build Status](https://travis-ci.org/Minecrell/SpongeAPI)

PR #506, opened 2015-03-18
Statistics API improvement suggestions

ST-DDT:
This PR is a collection of improvements i would like add to the Statistics API.
This suggestions are open for discussion, so feel free to comment on or criticize them.

Here is a list of changes with an explanation each.

**Change 1**
* `GameRegistry.getAchievements()`

Currently there is no way to get all Achievements, so i added this method to provide simple access to all Achievements similar to all other entries in the GameRegistry.

`Achievement` does not have a `getId()` method so we cannot get it with that from the registry.
Should that method be added to `Achievement` as well?

**Change 2**
- `Optional<StatisticFormat> Statistic.getStatisticFormat()` 
  - -> `StatisticFormat Statistic.getStatisticFormat()`
- `StatisticFormat StatisticGroup.getDefaultStatisticFormat()`
  - -> `Optional<StatisticFormat> StatisticGroup.getStatisticFormat()`

I flipped the placement of `Optional` to allow some more checks and allow using `Statistic`s without the `StatisticGroup`. Please let me explain.
What does the `StatisticGroup.getDefaultStatisticFormat()` tell us about the group?
It does not tell us anything (for sure) because every `Statistic` in that group is able to overwrite the group's default.
The new `StatisticGroup.getStatisticFormat()` allows the plugins to check whether the group has a fixed format for all elements in it.
If there is no fixed format for all elements then the group is a  general group.

In addition to that the `StatisticGroup` is not present in the minecraft code, so each `Statistic` in our API is never actually bound to a `StatisticGroup`. As a result of this from my point of view it should be possible to display/format the `Statistic`'s value without defaulting to the groups format.
(In fact each minecraft statistic stores its own reference to the format)
You can still check whether the `Statistic` uses its own format or the format of the group.

To sum it up:
* Default Statistic Format does not tell us anything, as long as it can be overwritten
* Fixed Formats for Groups allows distinguishing, between specialized groups and general groups
* Independent statistics should not be made dependent on a group that only contains elements that does not fit anywhere else
* All previous possibilities are preserved

**Change 3**
* `StatisticFormats` javadocs

Plugin authors may want to overwrite the default formatting to make it more readable in a specific context, so they should know the assumed input for the formatter. This is needed for plugin authors  to create their own instances, without looking at any server impl.

It also should guarantee, that if minecraft ever changes their statistic units (for example from ticks to seconds), that the format does still the same.
This is important for own/custom `Statistic`s to format the values properly and allows the plugin author to assign the right formatter to his statistic.

**Change 4**
* `Statistics` javadocs

Due to the addition of the `GENERAL` group that contains all elements that are not grouped/do not fit in any other group, the old javadocs are no longer true, so i changed them to describe the contents more precisely (I hope).

**Change 5**
* `StatisticGroups.DISTANCE_MOVED`

This is a logical group, that contains all statistics measuring the movement of the player, such as `WALK_DISTANCE`. There are quiet a lot elements in this group and all are related to each other so they should be in their own group. Especially if you want to show them together.

* `StatisticGroups.HIDDEN`

This is a logical group, that is meant to contain all `Statistic`s that should never be shown to the user, because the are used for internal counting for achievements or are used for plugins to keep track of a secret value, that may be needed for `Scoreboard`s. This is a reference meant for the check whether a statistic (or its group) is hidden.
````java
boolean hidden = statistic.getGroup() == HIDDEN; // => Do not show to the user.
````

**Change 6**
* `Statistics.DROP` -> `Statistics.ITEMS_DROPPED`

The old name was misleading so I made the name more expressive.

* `Statistics.ITEM_ENCHANTED` -> `Statistics.ITEMS_ENCHANTED`

IMO the plural name fits better in there.

* `Statistics.PLAY_ONE_MINUTE` -> `Statistics.TIME_PLAYED`

The old name was misleading because that value is not measured in minutes at all.
(Instead it is currently measured in ticks), so i changed the name to make it less surprising.

There are probably some other `Statistic`s in there that should be renamed as well. Suggestions?

**Change 7**
* `List<Achievement> Achievement.getChildren()` -> `Collection<Achievement> Achievement.getChildren()`

There is no order in the child/dependent achievements so we should not encourage index access to the child achievements (list). 
In addition to that it is more likely to be a set, because it will never contain duplicates (or at least it should not). But I don't want to limit the server implementations here so i used `Collection` here.
This is basically the same reason why we used `Collection` everywhere in the `GameRegistry`

**End of Changes**

I hope this is persuasive enough. If not please leave a comment, with a notice what i should explain some more.

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -64,6 +64,7 @@ import org.spongepowered.api.stats.StatisticBuilder;
 import org.spongepowered.api.stats.StatisticFormat;
 import org.spongepowered.api.stats.StatisticGroup;
 import org.spongepowered.api.stats.TeamStatistic;
+import org.spongepowered.api.stats.achievement.Achievement;
 import org.spongepowered.api.status.Favicon;
 import org.spongepowered.api.text.format.TextColor;
 import org.spongepowered.api.util.rotation.Rotation;
@@ -503,6 +504,13 @@ public interface GameRegistry {
      * @return An immutable collection containing all available formats
      */
     Collection<StatisticFormat> getStatisticFormats();
+
+    /**
+     * Gets a collection of all available {@link Achievement}s.
+     *
+     * @return An immutable collection containing all available achievements
+     */
+    Collection<Achievement> getAchievements();
 
     /**
      * Gets the {@link DimensionType} with the provided name.
--- a/src/main/java/org/spongepowered/api/stats/Statistic.java
+++ b/src/main/java/org/spongepowered/api/stats/Statistic.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.api.stats;
 
-import com.google.common.base.Optional;
 import org.spongepowered.api.text.translation.Translatable;
 
 /**
@@ -40,13 +39,11 @@ public interface Statistic extends Translatable {
     String getId();
 
     /**
-     * Gets the {@link StatisticFormat} of this statistic. If this is not
-     * present that this statistic's format is deferred to its group's default
-     * format.
+     * Gets the {@link StatisticFormat} of this statistic.
      *
-     * @return The format of this statistic, if available
+     * @return The format of this statistic
      */
-    Optional<StatisticFormat> getStatisticFormat();
+    StatisticFormat getStatisticFormat();
 
     /**
      * Gets the {@link StatisticGroup} this {@link Statistic} belongs to.
--- a/src/main/java/org/spongepowered/api/stats/StatisticGroups.java
+++ b/src/main/java/org/spongepowered/api/stats/StatisticGroups.java
@@ -30,7 +30,14 @@ package org.spongepowered.api.stats;
  */
 public class StatisticGroups {
 
+    /**
+     * Group for statistics that do not match in any group.
+     */
     public static final StatisticGroup GENERAL = null;
+    /**
+     * Statistic counting the distances moved in a specific way.
+     */
+    public static final StatisticGroup DISTANCE_MOVED = null;
 
     /**
      * Statistic counting the number of killed entities of a specific type.
@@ -79,6 +86,13 @@ public class StatisticGroups {
      * team.
      */
     public static final StatisticGroup KILLED_BY_TEAM = null;
+
+    /**
+     * Statistic for internal and plugin statistics, that are not meant to be
+     * displayed. Plugins should avoid displaying statistics in this group to
+     * the user.
+     */
+    public static final StatisticGroup HIDDEN = null;
 
     private StatisticGroups() {
     }
--- a/src/main/java/org/spongepowered/api/stats/StatisticFormats.java
+++ b/src/main/java/org/spongepowered/api/stats/StatisticFormats.java
@@ -36,7 +36,7 @@ public final class StatisticFormats {
     public static StatisticFormat COUNT = null;
     /**
      * A statistic measured in centimeters, meters, or kilometers depending on
-     * the magnitude.
+     * the magnitude. The formatter assumes the input are centimeters.
      */
     public static StatisticFormat DISTANCE = null;
     /**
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/api/GameRegistry.java [hunk 1]
```diff
@@ -505,6 +506,13 @@
     Collection<StatisticFormat> getStatisticFormats();
 
     /**
+     * Gets a collection of all available {@link Achievement}s.
+     *
+     * @return An immutable collection containing all available achievements
+     */
+    Collection<Achievement> getAchievements();
+
```
Deamon5550: Missing a get by name method.

src/main/java/org/spongepowered/api/stats/Statistic.java
```diff
@@ -40,13 +39,11 @@
...
-     * Gets the {@link StatisticFormat} of this statistic. If this is not
-     * present that this statistic's format is deferred to its group's default
-     * format.
+     * Gets the {@link StatisticFormat} of this statistic.
      *
-     * @return The format of this statistic, if available
+     * @return The format of this statistic
      */
-    Optional<StatisticFormat> getStatisticFormat();
+    StatisticFormat getStatisticFormat();
```
Deamon5550: This is completely backwards, the child should override the parent, not the parent the child.

Deamon5550:
Closed in 7444b8de26ebc476fc56cf4454bcab399f7022ab

src/main/java/org/spongepowered/api/stats/Statistic.java
ST-DDT: The parent does not overwrite the child. It just states the format for all of its children. Without exceptions.

From my point the child should not overwrite what the parent says/states if it says/states anything.

src/main/java/org/spongepowered/api/stats/StatisticGroups.java
```diff
@@ -30,7 +30,14 @@
...
 public class StatisticGroups {
 
+    /**
+     * Group for statistics that do not match in any group.
+     */
     public static final StatisticGroup GENERAL = null;
+    /**
+     * Statistic counting the distances moved in a specific way.
+     */
+    public static final StatisticGroup DISTANCE_MOVED = null;
```
ST-DDT: Any comment on this one?

src/main/java/org/spongepowered/api/GameRegistry.java
ST-DDT: There hasnt been a getId method on Achievements. And i wasnt sure whether it should be exposed so i skipped that. See description.

src/main/java/org/spongepowered/api/stats/StatisticGroups.java
Deamon5550: Its a completely arbitrary group which serves no purpose..., other than making the fact that it and the other 'general' statistics should be grouped some strange implementation detail.

you keep spouting that the general group is some arbitrary bin that I've cooked up and just chucked every other statistic but its really not...
stats

ST-DDT: I have seen the picture before (the last time you? quoted it)
I know that you are referencing the general view in minecraft, but this entire group has a mayor subgroup that is worth putting in a own subgroup. (Because it may be split in a extra view in minecraft itself)

I considered an improvement. But its a helper group so it is optional.

src/main/java/org/spongepowered/api/stats/Statistic.java
ST-DDT: Sorry for keeping on discussion, but i really want to talk this through.
Lets try it the other way round.

- What does the old method tell us about the group.
  - IMO: Nothing, am i wrong here? My version states whether the group has a format and if it has one it can be used for all contained Statistic.
- What does the old method tell us about the statistic with the group.
  - Whether it has the same format as the group. Well my version does that too.
- Does the old method prevent adding statistics to the wrong group?
  - No, you can add any `Statistic` `TIMES_KICKED` to `StatisticGroup.ITEM_USED`. (IMO the API should prevent that). My version does prevent that.

These are my concerns regarding the current system. Could you please explain it (1+3) to me then i won't bother you anymore. Sorry for the extra work this causes you.

PR #511, opened 2015-03-22
Make Operations less tied to Vanilla

AlphaModder:
Me and @Aaron1011 decided that the current attributes operations system was too tied to vanilla, because of it's hardcoded behavior. So now it's not hardcoded, to allow custom operations to how exactly they modify a base. Yeah.


gabizou:
Is this ready for review or a WIP?

AlphaModder:
@gabizou I don't plan on adding anything more.

src/main/java/org/spongepowered/api/attribute/AttributeCalculator.java
```diff
@@ -0,0 +1,53 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.attribute;
+
+import java.util.Collection;
+
+
+public interface AttributeCalculator {
```
ST-DDT: Missing javadocs

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -631,6 +632,13 @@
     AttributeModifierBuilder getAttributeModifierBuilder();
 
     /**
+     * Gets a new {@link AttributeCalculator}.
+     * 
+     * @return A new {@link AttributeCalculator}
+     */
+    AttributeCalculator getAttributeCalculator();
```
ST-DDT: Does this really returns a new instance?
I guess it should return the registered/default `AttributeCalculator`.

Because the current interface looks really stateless.

Or is there a reason you always need a new instance.
In which cases? One per entity? One per calculation?

AlphaModder: @ST-DDT Good point.

modwizcode:
Don't see a problem with merging this in, but I would like to understand what it's doing better. I don't understand what makes the API with this change any less specific than without it. 

More importantly, I don't know how to use this PR's API with operations to perform the same tasks as the current implementation.

AlphaModder:
@modwizcode Vanilla's operations are executed in a certain order in a certain way. 
There are three types of these:
* Additive
* Additive Percentage
* Multiplicative Percentage
Additives are the first processed. They are very straightforward, adding their values to the base.
Next, additive percentage operators are processed, but they work differently. They calculate how much they'd add to the modified base and add it to a sum which  is added to the modified base after all of these have been processed.
Finally, multiplicative percentage operators are processed, where they calculate how much to add to the modified base and immediately add it, resulting in a compound effect.
This PR allows custom operations to specify if they should be totaled or added immediately, and where in this chain they should go.

modwizcode:
Ah yes I see now, it's impossible currently to have the functionality where each modifier uses the current value then adds their modification after all modifiers have been calculated, because the calculation doesn't maintain any state.

I'm in support of merging this.
ping @SpongePowered/developers for final comments.

AlphaModder:
@modwizcode Great!

me4502:
I support this.

ST-DDT:
Is it possible to add a getter and setter for `AttributeCalculator` to `AttributeHolder`?
Otherwise you can caculate the value for yourself but you cannot use the value. Or did i missunderstand the purpose of this Pr?

AlphaModder:
@ST-DDT There is only one AttributeCalculator. It should be used behind-the-scenes in implementations.

ST-DDT:
> It should be used behind-the-scenes in implementations.

So why do we expose it in the first place?

AlphaModder:
@ST-DDT To calculate the value of modifiers without assigning them to an AttributeHolder.

modwizcode:
Nobody seems to have a real opposition to this, so I'm thinking I'll merge it now.

AlphaModder:
@modwizcode That sounds great!

Issue #479, opened 2015-03-02
Access to WorldType and usage of it in Server.createWorld()


dpatron:
Currently you don't have access to WorldType (AMPLIFIED, FLAT, LARGE_BIOMES, DEFAULT etc.) and can't set it in createWorld(). I think it is needed in order to grant a >basic< customized worldcreation (beside the usage of WorldGenerator).

Path in Minecraft: net.minecraft.world.WorldType

kenzierocks:
You're looking for dimension types I think.

dpatron:
DimensionTypes include End, Nether and Normal - not the WorldType.

boformer:
That's right. Additionally, there is also no way to set the config string of the flat generator (x layers of stone, y layers of grass, villages...).

Is this something that can be set independently from the the world generator, or is every value bound to a specific vanilla generator?

I know that Bukkit had a WorldType enum. You even had to set it when using a custom generator, which was kind of senseless.

uxsvent:
in 1.8 vanilla there is a limited generator that lets you choose biome and othe ore gen stuff. is there a way to tap into that code?

Zidane:
@bloodmc You typically don't do API but you are focused on this currently. Thoughts?

Zidane:
Closed as this is handled in Deamon's branch.

PR #515, opened 2015-03-24
Expand World creation process + Add WorldProperties

Deamon5550:
This PR expands the process for creating worlds.

Notable new classes:
- `WorldBuilder` This is a builder which can produce instances of `WorldCreationSettings` or `World`s directly.
- `WorldCreationSettings` Holds information which is used to define a new world for creation.
- `WorldProperties` defines numerous properties about the world (essentially a representation of the data with the `level.dat` file).
- `GeneratorType` This is a representation of what you may also think of as a world type, it holds a generator and some settings. Examples are flatland or amplified.

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -25,6 +25,8 @@
 
 package org.spongepowered.api;
 
+import org.spongepowered.api.service.persistence.data.DataContainer;
+import org.spongepowered.api.world.gen.WorldGenerator;
 import com.google.common.base.Optional;
 import org.spongepowered.api.attribute.Attribute;
 import org.spongepowered.api.attribute.AttributeBuilder;
@@ -91,8 +93,13 @@ import org.spongepowered.api.text.selector.SelectorType;
 import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.api.util.rotation.Rotation;
 import org.spongepowered.api.world.DimensionType;
+import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.WorldBuilder;
+import org.spongepowered.api.world.WorldCreationSettings;
+import org.spongepowered.api.world.GeneratorType;
 import org.spongepowered.api.world.biome.BiomeType;
 import org.spongepowered.api.world.difficulty.Difficulty;
+import org.spongepowered.api.world.storage.WorldProperties;
 
 import java.awt.image.BufferedImage;
 import java.io.File;
@@ -103,6 +110,7 @@ import java.net.URL;
 import java.util.Collection;
 import java.util.Locale;
 import java.util.UUID;
+import java.util.concurrent.Callable;
 
 /**
  * Provides an easy way to retrieve types from a {@link Game}.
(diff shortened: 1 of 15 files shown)


bloodmc:
Looks great.

@dumptruckman Can you review this PR and make sure it provides everything you need for MultiVerse?

Deamon5550:
Oh this also appears to close #479

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -825,4 +834,62 @@
...
+     * <p>The {@link DataContainer} with settings for the world generator is
+     * optional and an empty DataContainer will be created if it is not
+     * provided.</p>
+     * 
+     * @param name The name of the generator type
+     * @param generator The generator for the type to use
+     * @param settings Any settings that the generator uses, or null if no
+     *            settings are used
+     */
+    void registerGeneratorType(String name, ChunkGenerator generator, @Nullable DataContainer settings);
```
dumptruckman: Should this return GeneratorType?

dumptruckman: Also, how will the ChunkGenerator in this turn into a new ChunkGenerator from GeneratorType.createGenerator()? Should this be Callable<ChunkGenerator> or something?

src/main/java/org/spongepowered/api/world/gen/BiomeGenerator.java
```diff
@@ -45,6 +45,6 @@
      * @param width The width of the area (X-axis size)
      * @param length The length of the area (Z-axis size)
      */
-    void getBiomesForArea(World world, MutableBiomeArea buffer, int x, int z, int width, int length);
+    void generateBiomes(World world, MutableBiomeArea buffer, int x, int z, int width, int length);
```
dumptruckman: Why the name change here?

kenzierocks: Probably to match the javadoc.

src/main/java/org/spongepowered/api/GameRegistry.java
Deamon5550: Good points, I think my thinking at the time was that a generator is stateless, but thats not quite right and doesn't match the method in GeneratorType.

Deamon5550:
In order to support the separation of creating a world and loading it I'd like to create a representation of a World which is not loaded (similar to how User works). Which I called WorldData (needs a better name) and is now extended by World with several of the data related methods moved over.

Thoughts? @bloodmc @dumptruckman

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,352 @@
...
+     * @return The type
+     */
+    GeneratorType getType();
+
+    /**
+     * Sets the type of this world.
+     * 
+     * @param type The type
+     */
+    void setType(GeneratorType type);
```
ST-DDT: How can i set specific options for this generator like the different material layers?

Deamon5550: `WorldBuilder.settings(DataContainer settings)`

See also: `GeneratorType.getGeneratorSettings()` a variant of which is what the builder will be expecting.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -107,51 +116,70 @@
...
+     * container.
      *
-     * <p>If a world with the given name is already loaded then it is returned
-     * instead.</p>
-     *
-     * @param worldName The new world name
-     * @param generator The generator to generate the world with
-     * @return The new world
+     * @param worldName The name to lookup
+     * @return the world, if found
```
dumptruckman: Needs clarification here I think.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -107,51 +116,70 @@
...
+    /**
+     * Creates a new world from the given {@link WorldCreationSettings}. For the
+     * creation of the WorldCreationSettings please see
+     * {@link GameRegistry#getWorldBuilder()}.
+     * 
+     * <p>Although the world is created it is not loaded at this time. Please
+     * see {@link #loadWorld(String)} for loading the world.</p>
+     * 
+     * @param settings The settings for creation
+     * @return The world
```
dumptruckman: Clarification needed.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,397 @@
...
+     * @param time The time of day
+     */
+    void setWorldTime(long time);
+
+    /**
+     * Gets the Unix time stamp of when this world was last loaded.
+     * 
+     * @return The time this world was last loaded
+     */
+    long getLastTimePlayed();
```
bloodmc: This is a client-only method used to check when the player last played in world. This should be removed.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,397 @@
...
+     */
+    boolean isRaining();
+
+    /**
+     * Sets whether this world is currently experiencing rain/snow/cloud-cover
+     * (depending on the biome of a specific location).
+     * 
+     * @param state Is raining
+     */
+    void isRaining(boolean state);
```
bloodmc: This should be setRaining

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,397 @@
...
+     */
+    int getRainTime();
+
+    /**
+     * Sets the number of ticks until the weather is next toggled to a new
+     * random value.
+     * 
+     * @param time The time until the weather changes
+     */
+    void getRainTime(int time);
```
bloodmc: This should be setRainTime

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
progwml6: @bloodmc put a client annotation on stuff like this there is really no reason for us to not have limited client support in the api

src/main/java/org/spongepowered/api/world/WorldBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+     * strongholds. If not specified this will default to true.
+     * 
+     * @param enabled Are map features enabled
+     * @return The builder, for chaining
+     */
+    WorldBuilder usesMapFeatures(boolean enabled);
+
+    /**
+     * Sets whether hardcore mode is enabled. On servers this will cause players
+     * to be banned upon death, on client's the world will be deleted! If not
```
amaranth: clients

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,399 @@
...
+     */
+    boolean areCommandsAllowed();
+
+    /**
+     * Sets whether commands are allowed within this world. May not be respected
+     * when not in single player.
+     * 
+     * @param state Whether commands are allowed
+     */
+    void areCommandsAllowed(boolean state);
```
kashike: setCommandsAllowed

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -0,0 +1,399 @@
...
+     * @param difficulty The difficulty
+     */
+    void setDifficulty(Difficulty difficulty);
+
+    /**
+     * Gets the {@link WorldBorder} of this world.
+     * 
+     * @return The world border
+     */
+    WorldBorder getWorldBorder();
```
kashike: Where's the setter for this?

ST-DDT: WorldBorder is mutable and could be live.
Maybe this should be mentioned in the javadocs.
IMO we should change it to be similar with the rest of the API.

Deamon5550: WorldBorder is mutable, why are you wanting to set it? Setting the world border is not something you've ever been able to do to a world, nor should you.

Issue #70, opened 2014-09-10
Vector2 coordinate values


Soren025:
Wouldn't it make more sense to use X and Z instead of X and Y for the coordinate values for the interfaces 'Vector2d' 'Vector2i' and 'Vector2f'?

I say this considering that the primary use of a 2 dimensional vector would most likely be used along the Block/Chunk X/Z axis.

PR #522, opened 2015-03-30
Add API for custom world generators

rutgerkok:
This is the API part of a custom world generation API. I was asked by @bloodmc to create a PR, so here it is.

*updated on April 2, for latest API changes*

Plugins can now register their own implementations of the WorldGeneratorModifier interface. It is intended that, when the world is set to use that plugin as world generator, the plugin will be passed a WorldGenerator instance to this WorldGeneratorModifier. This WorldGeneratorModifier can then replace (parts of/the whole) world generator with their own implementations.

When a plugin wants to modify the terrain generation of an *existing* world, it can call `world.getWorldGenerator()`, make the desired changes and then call `world.setWorldGenerator()`. A plugin can also ask the world generator to generate something in a custom buffer. This is for example useful if a plugin wants to create a biome map: it just calls `world.getWorldGenerator().getBiomeGenerator()`.

I'm also working on an implementation, see https://github.com/SpongePowered/Sponge/pull/214.

##### GameRegistry
GameRegistry got a new method to let a plugin register a WorldGeneratorModifier. Also included are some methods to retrieve those registered modfiers.

##### World
The get/setSeed methods have been moved to WorldGenerator, as the world seed is purely a world generator property. The JavaDocs of getWorldGenerator have been clarified.

##### BiomeType
Previously, there were two methods: getPopulators and insertPopulator. There was no indication whether the iterable returned by getPopulators was mutable. The existance of a insertPopulator method suggests it isn't, and that insertPopulator should be used instead to add populators (but how do you then remove populators?).

I have just made the list mutable, so that people can do `BiomeType.JUNGLE.getPopulators().add(...)`, but also `BiomeType.JUNGLE.getPopulators().clear()`. The insertPopulator method is now redundant, and has been removed.

The JavaDocs of getPopulators have also been clarified.

##### BiomeGenerator
Lots of parameters have been removed. The size and location of the area to generate can be retrieved from the buffer, so no need to specify it in the method. If the need ever arises to only generate biomes in a part of the buffer, a method can be added to the buffer to view a subset of it, similar to `java.util.List.subList`. I feel that adding the dimensions in extra parameters will only be a source of bugs. It's not hard to imagine a plugin that ignores those parameters and ruins the whole buffer.

The world parameter is also removed. Many methods on world can cause new chunks to be generated, which would cause calls to the biome generator, which would cause an infinite loop. All useful information of the world has been provided when the plugin was asked to modify the world generator, so the plugin can keep track itself of the world seed, name, settings, etc.

##### GeneratorPopulator
Same story here as for BiomeGenerator.

##### Populator
The world parameter has been removed from the method: people can just call `chunk.getWorld()` to get the world.

A huge wall of text has been added to explain terrain population, hopefully people will now write populators that don't have all kinds of rare bugs.

##### PopulatorObject
No changes here, only a bit of extra JavaDocs.

##### PopulatorObjects
As per the JavaDocs of Populator, it is not at all easy to write a well-behaving populator. This class helps you to turn a PopulatorObject into a well-behaving Populator. Currently there's only one method, but more can be added in the future.

##### WorldGenerator
A WorldGenerator represents everything that's needed to generate a world. So it stores the seed and populators alike.

Compared to how the interface looks before this PR, a getter/setter for the base GeneratorPopulator has been added. A method called `getGeneratorPopulators()` has been added. I also made the same changes to getPopulators()/insertPopulator() as I did for BiomeType.

PR #515 renames the WorldGenerator to ChunkGenerator. I have kept the name the same, as I feel that this class should contain everything related to world generation, not just chunks, but also the biome generator, world seed and whether map features are enabled.

In other words: the design of this class is still in flux, if the class indeed gets renamed to ChunkGenerator some methods will need to be moved.

Now that #515 is merged, I will look into finding a solution to merge both PRs nicely.

##### WorldGeneratorModifier
A plugin needs to provide an implementation of this interface when it wants to modify world generator. When the server admin (or the player in singleplayer) has selected this WorldGeneratorModifier, its `modifyWorldGenerator` method is called, giving the plugin a chance to modify the world generator before the first chunk is generated.

Sure, a plugin can just manually modify the world generator of an existing world using `world.getWorldGenerator` and `world.setWorldGenerator`, but this will cause a mess with each plugin having its own way to (de)activate their modifications to a world. `world.setWorldGenerator` is intended for plugins wanting to modify some aspects of the world (like the seed), not as the main way to provide custom world generators.


src/main/java/org/spongepowered/api/plugin/PluginManager.java
```diff
@@ -77,4 +79,17 @@
...
+     * plugin does not implement the {@link WorldGeneratorProvider} interface.
+     *
+     * @param plugin The plugin to get the world generator from.
+     * @param worldName Name of the world.
+     * @param settings A string with (usually user-supplied) settings for the
+     *        world generator plugin, it's up to the world generator plugin to
+     *        do anything with this.
+     * @param worldGenerator The world generator to modify.
+     */
+    void modifyWorldGenerator(PluginContainer plugin, String worldName, String settings, WorldGenerator worldGenerator);
```
Zidane: This should go in GameRegistry likely.

src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -157,22 +157,12 @@
     Dimension getDimension();
 
     /**
-     * Gets the random seed for this world.
-     *
-     * @return The seed
-     */
-    long getWorldSeed();
```
Zidane: Wouldn't it be best to keep world seed here namely for the fact that you could override the seed by World if you want?

src/main/java/org/spongepowered/api/world/gen/PopulatorObject.java
```diff
@@ -28,7 +28,10 @@
 import org.spongepowered.api.world.World;
 
 /**
- * Represents a structure which can be placed into the world by a populator.
+ * Represents a relatively small object (must fit in a single chunk) that can be
```
Zidane: A Nether Fortress would be considered a PopulatorObject no? If so, that can easily go across chunk boundaries.

bloodmc:
Thanks @rutgerkok , Looks great!

@DDoS @Deamon5550 Please review.

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.world.gen;
+
+import org.spongepowered.api.plugin.Plugin;
+
+/**
+ * If a plugin wishes to provide a world generator, the plugin class (the class
```
Zidane: Great PR so far, you've put a lot of thought into this.

The only issue I have (besides a few questions above) is this interface. Maybe it would be best to fire an event to allow modification?

src/main/java/org/spongepowered/api/world/gen/PopulatorObject.java
Deamon5550: While it is true that the pieces that make up structures are typically less than a chunk in size it is by no means a requirement, in theory there is no limit to the size of a populator object. Especially since the objects placed by populators will often cross the boundary between two chunks which is why they are only placed in after the chunks in a 1 chunk radius around the target chunk have been generated.

src/main/java/org/spongepowered/api/plugin/PluginManager.java
Deamon5550: The settings should probably be changed to be a DataContainer as well, it would be easier for plugins to use as it doesn't require knowledge of the setting's format.

src/main/java/org/spongepowered/api/world/gen/WorldGenerator.java
```diff
@@ -35,16 +32,17 @@
...
-     * {@link BlockBuffer}.
-     *
-     * @param world The world
-     * @param buffer The buffer to generate the region into
-     * @param position The chunk position
+     * Gets the {@link GeneratorPopulator} of this world generator.
+     * @return The {@link GeneratorPopulator}.
+     * @see #setGeneratorPopulator(GeneratorPopulator)
+     */
+    GeneratorPopulator getGeneratorPopulator();
```
Deamon5550: Using a GeneratorPopulator for the base terrain generation is good, but there is a secondary need for them which is performing operations similar to that of the standard Populators but during generation. Things similar to the generation of caves would need to be applied as GeneratorPopulators. So having a 'primary' GeneratorPopulator is good but an addition structure similar to that of Populators should also be added.

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * annotated with {@link Plugin}) must implement this interface.
+ *
+ */
+public interface WorldGeneratorProvider {
```
DDoS: Minor thing: since the goal of this interface is to modify world generators, not provide new ones, wouldn't it be more appropriate to call it a WorldGeneratorModifier?

src/main/java/org/spongepowered/api/world/gen/BiomeGenerator.java
```diff
@@ -34,17 +33,16 @@
...
-     * @param width The width of the area (X-axis size)
-     * @param length The length of the area (Z-axis size)
+     * <p>The biome generator can be called multiple times for the same
+     * position. The biome generator should, for any position/world seed
+     * combination, always return the same biome.</p>
+     *
+     * @param buffer The buffer to generate the biomes into.
      */
-    void getBiomesForArea(World world, MutableBiomeArea buffer, int x, int z, int width, int length);
+    void getBiomesForArea(MutableBiomeArea buffer);
```
DDoS: This method needs to be renamed to generateBiomes for consistency.

DDoS:
BlockVolume and BiomeArea should have bound getters added. To prevent conflicts when both interfaces are implemented, BlockVolume should have getBlockMin() and getBlockMax(), and BiomeArea, getBiomeMin() and getBiomeMax(). For convenience, there should also be getBlockSize() and getBiomeSize() which return the equivalent of Max - Min + 1.

An OutOfBoundsException is thrown if an attempt is made to access data outside the Min and Max bounds.

The coordinate system for these access interfaces should be the most global one. If one represents a region of a world from p1 to p2, then blockMin = min(p1, p2) and blockMax = max(p1, p2). Same for biomes but with y discarded.

This is more desirable than always having min = 0 and max = size -1 because to properly generate you need the location of the buffer in the world.

    Vector3i min = buffer.getBlockMin();
    Vector3i max = buffer.getBlockMax();
    for (int z = min.getZ(); z <= max.getZ(); z++) {
        for (int y = min.getY(); y <= max.getY(); y++) {
            for (int x = min.getX(); x <= max.getX(); x++) {
                buffer.setBlock(x, y, z, someGenFunc(x, y, z));
           }
       }
    }

is more natural than

    Vector3i size = buffer.getBlockSize();
    for (int z = 0; z < size.getZ(); z++) {
        for (int y = 0; y < size.getY(); y++) {
            for (int x = 0; x < size.getX(); x++) {
                int wx = x + pos.getX();
                int wy = y + pos.getY();
                int wz = z + pos.getZ();
                buffer.setBlock(wx, wy, wz, someGenFunc(wx, wy, wz));
           }
       }
    }

This might be a bit confusing at first for some cases such a chunks, which also implement these interfaces, since chunk.getBlock(0, 0, 0) would only by valid for chunk (0, 0), but it's worth it for consistency and simpler code.

src/main/java/org/spongepowered/api/world/World.java
rutgerkok: Yeah, each world has its own WorldGenerator. In this PR the WorldGenerator class holds everything that's needed for world generation, like the chunk generator, the populators and their settings. The way I see it, the seed is just another setting, not more important than which biome generator is active.

Ideally, the world generator would be completely independent of the world. It would just be one big bag of classes and settings. `oneWorld.setWorldGenerator(anotherWorld.getWorldGenerator())` would then make `oneWorld` produce exactly, block for block, the same terrain as `anotherWorld`. In practice, Minecraft's ChunkProviderGenerate keeps a reference to the world object, so it cannot be assigned to another world. (I've made the implementation throw an IllegalArgumentException if you try.)

src/main/java/org/spongepowered/api/plugin/PluginManager.java
rutgerkok: This method was intended for inter-plugin communication. Let's say the command

`/mv create testworld normal -g AwesomeGenerator`

was called, then Multiverse would find the PluginContainer of the plugin AwesomeGenerator, use the above method to let it modify the WorldGenerator of the soon-to-be world, and then Multiverse would call server.createWorld with the modified WorldGenerator.

However, I think a better solution would be to add a method to #515's WorldBuilder class. Something like this?

```java
/**
 * Just before the actual world is created, each plugin listed here is notified so that it can
 * modify the world generator of the new world, for example to inject a custom biome
 * generator or populator.
 */
WorldBuilder plugins(PluginContainer... plugins)
```

I will make sure the plugin receives a DataContainer instead of a String.

src/main/java/org/spongepowered/api/world/gen/PopulatorObject.java
boformer: how about "(usually fits in a single chunk)"?

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
rutgerkok: I'm afraid that if an event is fired, every generator plugin installed will try to claim every world. I guess that most plugins will offer a setting to enable/disable the plugin for a world, but then the server admin has to find this setting for every world generator plugin.

This implementation allows the server admin to go to the WorldConfig and say: this world is assigned to AwesomeGenerator, that world to IceWorldGenerator, etc.

This implementation also makes it easy for Sponge to see which plugins are able to generate worlds (just an instanceof check), so that a nice generator selector can be shown in the singleplayer world creation menu.

Not that this implementation is ideal: it restricts each plugin to one generator. Maybe a similar system to command registration can be used for world generators?

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
rutgerkok: Yeah, I will change that. This name was from back when plugins would be responsible for providing an implementation of WorldGenerator, but that turned out to be a bad idea.

src/main/java/org/spongepowered/api/world/gen/BiomeGenerator.java
rutgerkok: Will do so.

boformer:
It would be better to have a ``WorldGeneratorRegistry`` (or use ``GameRegistry``) where plugins can register ``WorldGeneratorProvider``s.

It should not be a requirement that the plugin class is the provider class. Plugins should also be able to register multiple world generator providers.

src/main/java/org/spongepowered/api/world/gen/WorldGenerator.java
rutgerkok: I originally had a List similar to the populators, but that didn't work due to implementation difficulties: how to add a GeneratorPopulator to the standard ChunkProviderGenerate? How to add it to an implemention of IChunkProvider created by some random mod? The first case can probably be handled using a mixin, but the second would require something more creative.

For normal populators we don't have this problem: we can just listen to an event Forge (and any well-behaving mod) fires.

src/main/java/org/spongepowered/api/world/gen/PopulatorObject.java
rutgerkok: This is worded poorly. I meant that the object may not be larger than 16 x 256 x 16 blocks, not that the object cannot cross chunk boundaries.

Placing larger objects requires some careful planning to avoid hitting unloaded chunks, which is why Minecraft's structure code is so complex. Structures are not part of this PR, but I can imagine a Structure interface that holds many smaller PopulatorObject.

kenzierocks:
You seem to be focusing on making sure implementing is not difficult, but try to make the API look good, not make the implementation easy to do. Implementation shouldn't guide API.

rutgerkok:
This is already mostly how I assumed the buffers would work, and it's also how I implemented the first few buffers.

One exception: `getMaxBound()` is exclusive (so getting a block at that position results in an IndexOutOfBoundsException). This is how I iterate over the blocks in [my example plugin](https://github.com/rutgerkok/SpongeGeneratorTest/blob/a27864b9a256db8b443edabdea40e9ed45755c94/src/main/java/nl/rutgerkok/generatortest/GeneratorPopulatorTest.java#L39): (note that the x, y and z are the real world coords)

```java
        Vector3i min = buffer.getMinBound();
        Vector3i max = buffer.getMaxBound();
        for (int x = min.getX(); x < max.getX(); x++) {
            for (int y = min.getY(); y < max.getY() && y <= MAX_Y; y++) {
                for (int z = min.getZ(); z < max.getZ(); z++) {
                    setBlock(buffer, x, y, z, seed);
                }
            }
        }
```

I can change the getMaxBound() to be inclusive if you want. Either way, the JavaDocs need  to be improved.

You other suggentsion still stands: there are currently no different `getMinBound()` and `getMaxBound()` methods for biome and block areas, so implementing both buffers is currently impossible: one returns a Vector3i, the other Vector2i.

In the Vector2i returned by the biome area, the **y** coord corresponds to the **z** coord in the world, which leads to confusing code like [this](https://github.com/rutgerkok/Sponge/blob/58ce2b7268042330499f7099570ead17528a67a3/src/main/java/org/spongepowered/mod/world/gen/SpongeBiomeGenerator.java#L63): `int zStart = min.getY();`

Possible fixes:

* Do what you said, ignore that `buffer.getBlockSize().getZ() == buffer.getBiomeSize().getY()`.
* Use Vector3i even for biome areas. However, `buffer.setBiome(new Vector3i(3, 3, 3))` would then update the whole column, ignoring the y value.
* Don't use vectors at all for biome areas. (buffer.getMinX(), buffer.getMinZ(), buffer.setBiome(x, z, biome), etc.)
* Add a new interface called `Column` with an x and z pos.
* Other ideas?

DDoS:
@rutgerkok 

I don't really have any issues against an exclusive max, but inclusive seems more natural to me conceptually (although for code it isn't as much). Does anyone have any reason to favour one over the other?

For the biomes coord issue, the simplest solution is really the best here:

    buffer.getBlockSize().getZ() == buffer.getBiomeSize().getY()

The problem with using Vector3i for biomes is that it might give the wrong idea that biomes are 3D.

There's also no reason why the biome area should be the same size as the block volume, although it will almost always be the case. Hence having two independent sizes is somewhat better.

It has already been decided that biomes will use Vector2i, instead of Vector3i with the y coordinate ignored. See: https://github.com/SpongePowered/SpongeAPI/pull/480#discussion_r25638806

rutgerkok:
@kenzierocks 

I started by creating an API that would be ideal, then I made compromises to make the implementation not too difficult to maintain. The Sponge team will have to update the code to each and every Minecraft release....

That said, I don't always choose the easiest way. I already have this beast over [here](https://github.com/rutgerkok/Sponge/blob/58ce2b7268042330499f7099570ead17528a67a3/src/main/java/org/spongepowered/mod/world/gen/SpongeGeneratorPopulator.java#L71) to make any IChunkProvider able to generate chunks of any size.

The structure problem requires much more thought, it really won't fit in this PR. For structures, you need to do predictions about terrain that doesn't exist yet to decide on the layout of a structure. In Minecraft, we have seen many approaches to that problem: the desert temple ignores the terrain altogether and spawns at a fixed height, the mineshafts also ignore the terrain and spawn at a random height, the villages always spawn at the surface no matter how hilly the terrain is, etc. Adding a nice structure API can always be done at a later date, it's just too much work for me right now.

The `oneWorld.setWorldGenerator(anotherWorld.getWorldGenerator())` problem is there because the IChunkProvider class has a close relationship to one world. Maybe ChunkProviderGenerate can be updated to fix this, but this is impossible to do for all modded IChunkProviders in existance.

And so on.. The API will have limitations, we can't do everything, at least not now. The `List<GeneratorPopulator>` issue will probably be fixed, but the Sponge team will need to maintain more code because of this.

@boformer 

Yeah, your idea about the registry could work. I will try to implement something like that.

rutgerkok:
@DDoS 

I will change the max coords to be inclusive, unless anyone objects. I will also rename the bounds methods to be like you suggested.

DDoS:
@rutgerkok 

Ok, just to be sure there's no confusion, they should be added to the BiomeArea and BlockVolume classes, not the buffers.

MattBDev:
Looks great :+1: 
I hope you work on addressing the Sponge teams critiques soon so I can start using it

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
```diff
@@ -0,0 +1,56 @@
...
+     * {@link WorldGenerator#setBiomeGenerator(BiomeGenerator)}. To change
+     * terrain population, modify the populator list returned by
+     * {@link WorldGenerator#getPopulators()}.</p>
+     *
+     * @param worldName The name of the world.
+     * @param settings A string with settings, can be used by the plugin to
+     *        modify the world generator.
+     * @param worldGenerator The world generator, should be modified.
+     */
+    void modifyWorldGenerator(String worldName, String settings, WorldGenerator worldGenerator);
```
ZephireNZ: For `settings`, perhaps have an overload that uses a `DataContainer` rather than a `String`? This could mean that a node in a config could easily be passed through to the provider, while still allowing a simple `String`.

Same for the method under `PluginContainer`.

src/main/java/org/spongepowered/api/world/gen/WorldGeneratorProvider.java
dumptruckman: I think a plugin should be able to provide multiple generators. Just make it so they can register it to a name and then you can look up the generator by name.

kenzierocks: Can we have the name be like blocks + items? `pluginid:name`. It'll prevent conflicts.

rutgerkok:
I updated the API, I hope I didn't miss anything.

The base generator populator is not part of the getGeneratorPopulators() list. Although the two implement the same interface, they are completely different. You should not have two generator populators that generate the base terrain, but two generator populators that create caves work fine. So for the base terrain there's no List, only for the other generator populators. Maybe it would be a good idea to change the base generator populator over to a new interface called BaseChunkGenerator? In this way, it's impossible to accidentally add the generator populator using the wrong method.

@DDos I have added two methods to the Area classes (getBlockMin/getBlockMax), but the methods of the VolumeBuffer and AreaBuffer classes are still there (getMinBound/getMaxBound). What should I do there? Leave both? I have changed getMaxBound to be inclusive for consistency with getBlockMax and getBiomeMax.

DDoS:
@rutgerkok Do we even need the Area/VolumeBuffer interfaces? You could probably add the bound methods from the access interfaces to the immutable buffers separately, since they can't implement these interfaces. The mutable buffer gets these by extending the access interfaces. The mutable and immutable buffers don't need a common super interface, so you can remove Area/VolumeBuffer.

Zidane:
Looking fantastic so far @rutgerkok . Thanks for making the requested changes.

src/main/java/org/spongepowered/api/world/gen/WorldGenerator.java
Zidane: As said elsewhere, the impl bends to the API's will, always. This needs to be a list.

Deamon5550: In the MC source our GeneratorPopulators are equivalent to MapGenBase in function if that makes more sense.

Deamon5550: Also perhaps not replacing this with a List, but having an additional list. So there is one GeneratorPopulator which is the 'primary' and it creates the base terrain. The members of the list then modify this base.

src/main/java/org/spongepowered/api/world/extent/BiomeArea.java
```diff
@@ -30,10 +30,31 @@
 
 /**
  * An area containing biomes.
+ *
+ * <p>Some methods accept a pair of two ints, representing the x and z location
+ * of a biome. Some other methods accept or return a {@link Vector2i}. The
+ * y position of this vector is equal to the z position of a biome in the
+ * world.</p>
  */
```
Adubbz: I can certainly understand the reasoning behind this, although my concern would be that developers may become confused as to which y values represent z values and vice-versa.

src/main/java/org/spongepowered/api/world/extent/BiomeArea.java
```diff
@@ -30,10 +30,31 @@
...
     /**
+     * Gets the biome location with the lowest x and y that is still a valid
+     * position for {@link #getBiome(Vector2i)}.
+     *
+     * @return The lowest biome location
+     */
+    Vector2i getBiomeMin();
+
+    /**
+     * Gets the biome location with the higest x and y that is still a valid
```
Adubbz: Some minor spelling nitpicking, highest not higest :P

src/main/java/org/spongepowered/api/world/extent/BiomeArea.java
gratimax: That's a concern outside the scope of this PR, see the resolution of #70 (decided to use a vector2 in the end).

boformer:
@rutgerkok Great that you moved the registration to the ``GameRegistry`` :)

src/main/java/org/spongepowered/api/world/extent/BiomeArea.java
rutgerkok: Thanks, fixed. :smiley:

rutgerkok:
@DDoS 

> So remove Area/VolumeBuffer, add the proper bounds methods to BiomeBuffer and VolumeBuffer. I don't think that the fact that the mutable buffers implement the access interfaces will be a problem.

Done.

----

I will now try to rebase this PR on https://github.com/SpongePowered/Sponge/pull/218 (both the API and implementation).

rutgerkok:
Rebased the API and made it work better with the API on master. Notable changes:

* All generator settings have been moved out of WorldGenerator, they are now in WorldProperties/WorldCreationSettings/WorldBuilder. The (generator) populator lists remain, as well as the biome generator.
* You can now specify the WorldGeneratorModifiers to use in WorldBuilder.
* You can now change the WorldGeneratorModifiers in WorldProperties. world.setWorldGenerator will be removed soon, but it's still here until I have implemented the methods on WorldBuilder and WorldProperties.
* Removed `GeneratorType registerGeneratorType`. This method forced plugins to create an implementation of WorldGenerator (making it almost impossible to evolve this interface, and risking plugins implementing the interface incorrectly). World generator modifiers are the way forward for plugins that want to modify terrain generation.

I hope to update the implementation tomorrow.

DDoS:
The SpongeEventFactory tests have failed for StateEvent. This needs to be fixed.

Other than that, I'm approve this PR.

PR #409, opened 2015-01-23
[WIP] Dynamic Commands (New PR for my old commands PR.)

AlphaModder:
Apparently github does not allow you to change the head fork after creating a PR, so i closed it and remade this one.

This slightly modifies the command interfaces to allow for more variety in command systems, and adds a more featured default command system The features it will introduce to it are:

* Allows use of /plugin:command in a less hacky way.
* Allows plugins to register conflicting commands, and resolves those conflicts when the command is sent.
* Removes the callback feature from command registration as it is no longer neccessary due to the above change.
* Allows use of /mc or /minecraft:command to run vanilla commands [NYI]
* Creates an API that may be used to set certain aliases to point to certain plugins in the command system.

It is based on EvilSeph's old blog post [here](http://evilseph.com/post/75608917830/bukkit-supporting-minecraft-commands).

Due to configuration files not yet having a standard format yet, I will only create an API that allows the command configuration described in the blog post to be implemented, but this PR does not include the actual implementation of the configuration file. 

I still need help with: a good way to wrap vanilla/sponge commands into the existing system. maybe some kind of VanillaPluginContainer/SpongePluginContainer as wrappers so the command system can use them?

PR #524, opened 2015-03-31
[WIP] Dynamic Commands

gabizou:
This PR Adds the capability of dynamic commands.

 A Continuation PR of #409

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -75,6 +75,7 @@ import org.spongepowered.api.stats.StatisticGroup;
 import org.spongepowered.api.stats.TeamStatistic;
 import org.spongepowered.api.stats.achievement.Achievement;
 import org.spongepowered.api.stats.achievement.AchievementBuilder;
+import org.spongepowered.api.service.command.sponge.CommandResultBuilder;
 import org.spongepowered.api.status.Favicon;
 import org.spongepowered.api.text.chat.ChatType;
 import org.spongepowered.api.text.format.TextColor;
@@ -1021,5 +1022,12 @@ public interface GameRegistry {
      *         none could be found
      */
     Optional<ResourcePack> getById(String id);
+
+    /**
+     * Gets a new {@link CommandResultBuilder}.
+     * 
+     * @return A new {@link CommandResultBuilder}
+     */
+    CommandResultBuilder getCommandResultBuilder();
 
 }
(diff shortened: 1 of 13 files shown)


gabizou:
@AlphaModder while I'm getting the branch to compile, can you please explain the whole reasoning of Dynamic Commands (i.e. What are they for? What do they do? Why do we need them?)?

.gitignore
```diff
@@ -64,6 +64,8 @@ Thumbs.db
 .idea
 .project
 .settings
+.metadata
```
Zidane: Doesn't belong in this PR.

AlphaModder:
The dynamic commands idea is that it allows commands to be dynamically resolved based on their alias. Using this, the "plugin:command" notation does not need to be a duplicate of every command registered, as the code can check the prefix when the command is executed and decide which plugin to run from. This also opens up the possibility of using minecraft commands "minecraft:", and completely solves the issue of command conflicts, as a single alias could be registered by more than one plugin and each command would still be individually addressable. Finally, it also allows plugins to take control of this system, and explicitly set the plugin that will be used when a non-prefixed command is used by a certain kind of CommandSender. A way to modify this behavior through a configuration file could be in either Sponge or a sponge plugin. As for the CommandResult thing, it was just something sk89q asked me to do in this... 

tbh, with all of @boformers changes to the code, and just the general API differences, I think I should rewrite this from the ground up. It'd probably be about the same, functionality and api-structure wise, but it was my first PR, and I've learned a bit since then.

The whole thing is based on [EvilSeph's blog post](http://evilseph.com/post/75608917830/bukkit-supporting-minecraft-commands).

boformer:
Please think about proper names for the classes and packages (not sponge...).

I also still think that ``CommandCallable.call`` should always return a ``CommandResult`` (no generics).

Maybe we can split the PR:
1. addition of ``CommandResult``
2. Command prefixes

AlphaModder:
@boformer Yeah, if I do rewrite it I'll make CommandResult a seperate PR.

AlphaModder:
@gabizou I assume the close is so I can open a rewritten version?

gabizou:
Yes

gabizou:
@boformer if you can get the CommandResult, and @AlphaModder get the dynamic command system with aliases, we can focus on the exact features. As of right now, the branch is too convoluted as to what it needs.

boformer:
@AlphaModder is there a reason for the getAffectedBlocks/Entities/Items/Query method of CommandResult in your PR?
It seems like CommandBlocks only require the wasProcessed and getSuccessCount method.

AlphaModder:
@boformer Those are the statistics recorded by vanilla commands.

PR #529, opened 2015-04-02
Add getter and setter for Chat receivers

DotDashh:
This PR will add the following methods:

List<Player> getReceivers()
void setReceivers(List<Player>)

The reason it should be added is to simplify who/what receives messages.


src/main/java/org/spongepowered/api/event/entity/player/PlayerChatEvent.java
```diff
@@ -32,5 +32,18 @@
...
+     * @return The list of {@link Player}s
+     */
+    List<Player> getReceivers();
+    
+    /*
+     * Sets the list of players who receive the message.
+     *
+     * @param receivers The list of {@link Player}s
+     */
+    void setReceivers(List<Player> receivers);
```
Zidane: If the list returned above is mutable, no need for a setter of them. You can grab the list and manipulate it.

Also, is ordering important to you? Otherwise, I'd use a Set.

Zidane:
To be re-created.

PR #531, opened 2015-04-02
Maps API

### LINKED ISSUE #295: Maps
The Sponge API is in need of maps. A potential map API would most likely look very similar to [bukkit](http://jd.bukkit.org/rb/apidocs/org/bukkit/map/package-summary.html)'s map API, although any improvements are welcome.

Some things to consider:
- How coupled is this to NBT/persistence? See the [map format](http://minecraft.gamepedia.com/Map_Item_Format). We may need to wait on persistence and then write map data as a persistence class.

modwizcode:
Well I realized recently that I'd be unable to complete this PR in a timely manner so I'll mention the state of this PR.

Currently it's a feature complete base API for maps. Based on the old non-2.0 data API. Much of the existing code for colors and canvas interaction is correct however the structure is also incomplete and missing major features like cursors/decorations/icons (different named for the same thing). [This gist](https://gist.github.com/modwizcode/12a08d5a9399285853e0) represents my plans for the API that I would've worked on if I had time. 

I leave this PR here simply as a reminder of the feature and because it has some of the base structure already completed.

If anyone has questions regarding map behavior I can be pinged here or in a comment on the gist.

I found time to write the full API. I won't be doing implementation however.

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -44,6 +44,8 @@ import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.item.inventory.ItemStackBuilder;
 import org.spongepowered.api.item.merchant.TradeOfferBuilder;
 import org.spongepowered.api.item.recipe.RecipeRegistry;
+import org.spongepowered.api.map.MapView;
+import org.spongepowered.api.map.MapViewRegistry;
 import org.spongepowered.api.potion.PotionEffectBuilder;
 import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.scoreboard.ScoreboardBuilder;
@@ -534,4 +536,12 @@ public interface GameRegistry {
      * @return The {@link Translation} with the given ID or Optional.absent() if not found
      */
     Optional<Translation> getTranslationById(String id);
+
+    /**
+     * Returns the {@link MapViewRegistry} to use for manipulating
+     * {@link MapView}s.
+     *
+     * @return The map view registry
+     */
+    MapViewRegistry getMapViewRegistry();
 }
--- a/src/main/java/org/spongepowered/api/map/MapCanvas.java
+++ b/src/main/java/org/spongepowered/api/map/MapCanvas.java
@@ -1,0 +1,114 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.map;
+
+import com.google.common.base.Function;
+import org.spongepowered.api.map.color.MapColor;
+import org.spongepowered.api.map.font.MapFont;
+import org.spongepowered.api.text.Text;
+
+import java.awt.Image;
+import java.awt.image.BufferedImage;
+
+/**
+ * A representation of the canvas (pixel surface) of a map.
+ */
+public interface MapCanvas {
+
+    /**
+     * Returns the width of the map's canvas.
+     *
+     * @return The map canvas's width
+     */
+    int getWidth();
+
+    /**
+     * Returns the height of the map's canvas.
+     *
+     * @return The map canvas's height
+     */
+    int getHeight();
+
+    /**
+     * Sets the specified pixel.
+     *
+     * <p>The coordinates start with 0,0 as the top left corner moving right
+     * and down.</p>
+     *
+     * @param x The x coordinate of the pixel
+     * @param y The y coordinate of the pixel
+     * @param color The color to set the pixel to
+     * @throws IndexOutOfBoundsException If the coordinates are greater than
+     *         the width and height of the map's canvas
+     */
+    void setPixel(int x, int y, MapColor color);
+
+    /**
+     * Returns the {@link MapColor} at the specified pixel.
+     *
+     * The coordinates start with 0,0 as the top left corner moving right and
+     * down.
+     *
+     * @param x The x coordinate of the pixel to get
+     * @param y The y coordinate of the pixel to get
+     * @return The color of the pixel at the specified coordinates
+     * @throws IndexOutOfBoundsException If the coordinates are greater than
+     *         the width and height of the map's canvas
+     */
+    MapColor getPixel(int x, int y);
+
+    /**
+     * Create an {@link Image} out of the map allowing it to be viewed outside
+     * the game in conventional image viewers and shared on the web. Simply
+     * converts indexes in the map's canvas to the RGB values they represent.
+     *
+     * @return The rendered image
+     */
+    Image toImage();
+
+    /**
+     * Draws an image to the map, clipping it if it goes out of bounds.
+     *
+     * @param x The top left x coordinate on the map
+     * @param y The top left y coordinate on the map
+     * @param image The image to draw
+     */
+    void drawImage(int x, int y, Image image);
+
+    /**
+     * Draws text to the map, handling colors but not formatting. Clips text
+     * that goes out of bounds at the character before it's out of bounds.
+     *
+     * @param x The top left x coordinate of the drawn text
+     * @param y The top left y coordinate of the drawn text
+     * @param text The text to draw
+     * @param font The font to use when drawing the text
+     *
+     * @throws IllegalArgumentException If the font is missing a character used
+     *         by the Text provided
+     */
+    void drawText(int x, int y, Text text, MapFont font);
+
+}
--- a/src/main/java/org/spongepowered/api/map/MapRenderer.java
+++ b/src/main/java/org/spongepowered/api/map/MapRenderer.java
@@ -22,26 +22,39 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.data.manipulator.mutable.item;
+package org.spongepowered.api.map;
 
-import org.spongepowered.api.data.manipulator.DataManipulator;
-import org.spongepowered.api.data.manipulator.immutable.item.ImmutableMapItemData;
-import org.spongepowered.api.data.value.mutable.Value;
-import org.spongepowered.api.item.ItemTypes;
-import org.spongepowered.api.item.inventory.ItemStack;
-import org.spongepowered.api.map.MapView;
+import com.google.common.base.Optional;
+import org.spongepowered.api.entity.player.Player;
 
 /**
- * Represents an {@link DataManipulator} hosting the specific map
- * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
+ * Plugin classes should implement this to modify the image shown on a map.
  */
-public interface MapItemData extends DataManipulator<MapItemData, ImmutableMapItemData> {
+public interface MapRenderer {
 
     /**
-     * Gets the {@link Value} for the {@link MapView} attached to this
-     * {@link ItemTypes#FILLED_MAP} {@link ItemStack}.
+     * Returns if this map renders different views for each player viewing it.
      *
-     * @return The value for the attached map view
+     * <p>Most maps representing shared content are the same for all players
+     * that hold them, but sometimes a plugin may wish to render a unique view
+     * for each player.</p>
+     *
+     * @return True if the rendering is specific to each player, false otherwise
      */
-    Value<MapView> attachedMapView();
+    boolean isContextual();
+
+    /**
+     * Gives access to the {@link MapCanvas} for rendering onto a map, this also
+     * provides a viewer and a reference to the original map data to allow
+     * updating cursors and accessing relevant information to the render.
+     *
+     * <p>The viewer is guaranteed to be present if {@link #isContextual()}
+     * returns true.</p>
+     *
+     * @param viewer The player viewing this render, {@link Optional#absent()}
+     *        otherwise
+     * @param view The map that's being rendered
+     * @param canvas The canvas that the renderer will draw onto
+     */
+    void render(Optional<Player> viewer, MapView view, MapCanvas canvas);
 }
--- a/src/main/java/org/spongepowered/api/map/MapView.java
+++ b/src/main/java/org/spongepowered/api/map/MapView.java
@@ -1,0 +1,212 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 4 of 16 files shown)


src/main/java/org/spongepowered/api/item/data/map/MapData.java
```diff
@@ -0,0 +1,139 @@
...
+     * Sets the center of this map to the world x and y coordinates in the
+     * supplied {@link Vector2i}.
+     *
+     * @param center The new center of this map
+     */
+    void setCenter(Vector2i center);
+
+    /**
+     * A virtual map is only backed by it's {@link MapRenderer}s and will not
+     * have any relation
```
gabizou: Not a complete sentence. Also missing an extra space between comment and `@return`.

src/main/java/org/spongepowered/api/item/data/map/MapRenderer.java
```diff
@@ -0,0 +1,42 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.item.data.map;
+
+import org.spongepowered.api.entity.player.Player;
+
+/**
+ * Plugin classes should implement this interface to create dynamic maps.
```
gabizou: This doesn't quite describe what the class does though. Maybe elaborate?

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -1199,4 +1204,45 @@
      */
     GeneratorType registerGeneratorType(String name, Callable<WorldGenerator> generator, DataContainer settings);
 
+    /*
+     * Gets a complete list of {@link MapColor}s including the non-base colors.
+     *
+     * @return The list of all {@link MapColor}s available to be used
+     */
+    List<MapColor> getCompleteMapColors();
```
ZephireNZ: Why can't this simply be `getMapColors()`? Unless you were planning on adding another method with only the base colours.

modwizcode: I will actually add the other method now that you mentioned it.

Aaron1011:
Is support for map cursors going to be added?

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Gets the {@link MapData} that this canvas is attached to.
+     *
+     * @return The {@link MapData} for this canvas
+     */
+    MapData getMapData();
+
+    /**
+     * Gets the {@link MapColor} at the specified pixel position. If the position
+     * is invalid (out of range), the value will be absent. Pixels start at (0,0)
```
simon816: If it's out of range, it should throw `IndexOutOfBoundsException`, no need for Optional in this case.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
```diff
@@ -0,0 +1,89 @@
...
+     * in the top-left corner and go to (width-1, height-1).
+     *
+     * @param x The x coordinate of the pixel to get
+     * @param y The y coordinate of the pixel to get
+     * @return The {@link MapColor} of the pixel if coordinates are valid
+     */
+    Optional<MapColor> getColorAt(int x, int y);
```
simon816: Should have `Vector2i` overloads too

src/main/java/org/spongepowered/api/item/data/map/MapColor.java
```diff
@@ -0,0 +1,87 @@
...
+
+package org.spongepowered.api.item.data.map;
+
+
+import java.awt.Color;
+
+/**
+ * Represents a color on a map.
+ */
+public interface MapColor {
```
simon816: Should have `@CatalogedBy(MapColors.class)`

simon816:
May be useful to add a method on `BlockState` (or some other block interface) to get the `MapColor` associated with the block.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Provides access to the raw map colors, only use this if your plugin has
+     * a special need to avoid built-in color matching methods. The pixels are
+     * arranged in single-dimensional (row * width) + xOffset format. Each pixel
+     * represents a byte index into the color palette array from
+     * {@link org.spongepowered.api.GameRegistry#getCompleteMapColors() GameRegistry#getCompleteMapColors()};
+     *
+     * @return The mutable color array of this canvas
+     */
+    byte[] getRawPixels();
```
simon816: By using `byte` isn't it limiting the maximum possible colors to 256?

src/main/java/org/spongepowered/api/item/data/map/MapColor.java
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param shade The shade modifier
+     * @return The a new {@link MapColor} with the selected shade modifier
+     */
+    MapColor createShaded(int shade);
+
+    /**
+     * Gets the raw map palette index for advanced manipulation.
+     *
+     * @return The index suitable for raw manipulation of map colors.
```
simon816: remove the period

src/main/java/org/spongepowered/api/item/data/map/MapColor.java
modwizcode: Even if the MapColors class is not a full enumeration?

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
modwizcode: It's a raw vanilla accessor/mutator, so it's using the raw type from vanilla. I don't have a great way to mark things that may change unexpectedly. But I think this is a rare exception to the no exposure of vanilla where it's better to have an option with full access.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
modwizcode: I don't think anything uses an `IndexOutOfBoundsException` (or an exception) in the API. Why have this be the exception?

modwizcode:
I am trying to figure out the best way to handle cursors. Assume I will add them in a later PR as they aren't as important as the core API.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
modwizcode: Added

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
simon816: There's a few places where it's used, e.g see 14035b44da5ee3f8063f6e089a434495c31f7a1c and [search](https://github.com/SpongePowered/SpongeAPI/search?utf8=%E2%9C%93&q=IndexOutOfBoundsException&type=Code)

ST-DDT: There should also be a method `Vector2i getSize()` for this purpose.
Otherwise you dont know which pixels you can access. Because mods may be able to change the LOD for maps as well.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
```diff
@@ -0,0 +1,112 @@
...
+     * Uses the color matching from {@link org.spongepowered.api.GameRegistry#matchColor(Color)} to
+     * draw an {@link Image} to the map.
+     *
+     * @param x The x coordinate of the top-left corner
+     * @param y The y coordinate of the top-right corner
+     * @param width The width of the image that is drawn
+     * @param height The height of the image that is drawn
+     * @param image The {@link Image} to draw
+     */
+    void drawImage(int x, int y, int width, int height, Image image);
```
ST-DDT: What happens if zhe given image is too big to be displayed? Is it risized? Is there a way to customize it?

Deamon5550: Probably better (and easier) to just clamp it to the size.

modwizcode: It will match the behavior of a Java Graphics

src/main/java/org/spongepowered/api/item/data/map/MapRenderer.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.spongepowered.api.item.data.map;
+
+import org.spongepowered.api.entity.player.Player;
+
+/**
+ * Used to dynamically draw a map. This is implemented by plugins to create maps
+ * that can update while they're held.
+ */
+public interface MapRenderer {
```
ST-DDT: There should be a method to obtain a renderer for default maps in the GameRegistry to allow plugins the usage of the landscape as a background image.
`MapRenderer newMapRenderer(World, Vectori center)`

gratimax: I assumed that a map renderer is only for plugins (as the javadocs say) and that a MapData without any renderers attached to it would have the default behavior. Perhaps this needs some clarification.

modwizcode: Yes. The plugins could use a non-virtual map with their chosen center location and then use map renderers to add overlays, I will try to clarify this in the docs.

src/main/java/org/spongepowered/api/item/data/map/MapCanvas.java
modwizcode: LOD? Maps are exactly hardcoded to 128x128 pixels and it's been like that forever, additionally, mods cannot change this unless they also are transformer mods from what I've seen. getWidth and getHeight are your getSize.

ieiesting:
etrdgtdfgfgdfgdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdfdf

me4502:
This is looking good. I remember @sk89q once told me that Bukkit had a few issues with the maps API, causing slow updating and other issues. I don't exactly recall what those issues are, but it may be best to talk to him before merging this.

gabizou:
@modwizcode as #542 has been merged in, please update your PR ;)

gabizou:
@modwizcode can you rebase onto master since the Data API changes?

modwizcode:
I am planning to, but I'm behind on school work, I should have more time
once school wraps up. Until then I'm not sure I'll have time to do much
more than talk about it.

modwizcode:
I've had a lot of school work to catch up on recently but as the school year is wrapping to a close, I would like some help on progressing with this. First of all knowing what to put in SpongeCommon would be helpful, second of all I would like some help with how the data API is implemented in both interface and implementation. I took a look but the structure is really complex and I don't understand how to best take advantage of the structure.

src/main/java/org/spongepowered/api/item/data/map/MapColor.java
modwizcode: @SpongePowered/developers comments on this? There are more complicated colors mapped in the implementation. Does `@CatalogedBy` really fit?

src/main/java/org/spongepowered/api/item/data/map/MapData.java
```diff
@@ -0,0 +1,142 @@
...
+     * Sets the world associated with this map's display.
+     *
+     * @param world The new world for this map
+     */
+    void setWorld(World world);
+
+
+    /**
+     * Gets the string identifier of the map that will be used by Minecraft to
+     * represent this map. No guarantee in relation to stored map files on the
```
modwizcode: Type should be "guarantee"

modwizcode:
Update, working on the rebase into master to deal with data manipulators. Map api is a very unique type of data API given all the components it has so the structure will be more specific.

src/main/java/org/spongepowered/api/item/data/map/MapColor.java
me4502: I feel it needs a catalogue.

modwizcode:
@gabizou In order to use the Data API maps have to be created through builders. This is primarily because plugins need to choose between creating a new real map (Minecraft will render the world on it), creating a virtual map (Minecraft will not update the map) and loading existing maps. 

Since maps are stored separately on disk, the ability to change the data for maps previously created, without a data holder is required. 

I'm wondering if it's worth trying to shoehorn maps into the data API when it's starting to seem like maps are a very separate API. Since maps are different from all existing data manipulators and types does it even make sense to put maps in as a data API? Or should I simply provide a single value that contains a separate MapData instance which is part of a separate API specifically for maps.

ST-DDT:
> Or should I simply provide a single value that contains a separate MapData instance which is part of a separate API specifically for maps.

IMO this is the best solution as we already use it for `Villager`s and their careers.
* `ItemStack` <-> `MapData` <-> `MapSpecification`
* `Villager` <-> `CareerData` <-> `Career`

`MapSpecification` (currently known as `MapData`) should be renamed to avoid confusion with the Data API. And for obvious reasons it should not be called `Map`.

If we use `MapData` directly we would modify more than one `DataHolder` at once and this would be an anomaly in our Data API. AFAIK it also would also be the only "data" object in our API that is persisted independently of any `DataHolder`.

ryantheleach:
So MapSpecification would either contain data on which mapID, zoom data, and location in the world for a RealMapSpecification, or an ID and Type/class for a VirtualMapSpecification?

> Since maps are stored separately on disk, the ability to change the data for maps previously created, without a data holder is required.

What is the need? In addition couldn't you create a MapStore which is a dataHolder which contains references to many MapData's ? It would return a DataManipulator that maps the map id's to (real)MapSpecifications.

sk89q:
@me4502 I wrote Bukkit's map API, and my only complaint at the time was that they removed some features when they put it into Bukkit. That said, it didn't update any slower than regular Minecraft.

modwizcode:
There would be a single data class. Virtual maps have all the same info except isVirtual would be true. All the NBT is the same (and is still required if I am correct). @ryantheleach I'm not understanding the second half of your comment. The ability to modified stored maps with no DataHolder reference is important because plugins should be able to update existing maps that might be present all over the world, plugins shouldn't have to store the location of the maps.

@ST-DDT is correct, I almost forgot to mention the fact that maps effect multiple items/itemframes at once when they are updated and that's completely bizarre when taking the rest of the data API for comparison.

gabizou:
@modwizcode It would be best to just have a `MapItemData` that contains a `MapSpecification` so that the map specification can be changed out at will.

modwizcode:
Seems good to me.

AlphaModder:
Is this PR still going? It's been over a month since it's been updated...

modwizcode:
Waiting for values. I just pulled this up to look at yesterday before
becoming busy I'll try to work it off the values branch eventually.

AlphaModder:
@modwizcode Ah, alright. Thanks!

modwizcode:
Closing due to incompleteness and irrelevance to current api.

modwizcode:
Just a note, the API defined here is valid so reopening. the interaction with data is not.

modwizcode:
I will attempt to get this merged tomorrow, implementation will not be completed by me as I have no time to perform it.

modwizcode:
I have documented some information on my goals for the api and the details for the implementation along with research found about how maps work. This will help whoever continues the api pr and implements the result. [This gist],(https://gist.github.com/modwizcode/12a08d5a9399285853e0), has the information.

gabizou:
@modwizcode mind updating the PR description with some actual description of where it's at, what is needed etc.?

modwizcode:
@gabizou well I just defined the structure of the API I'd like to see which is somewhat similar but basically completely different than what's here.

modwizcode:
Well I was able to work on this so here you go. I won't have any time to implement or fix problems in the API I just pushed though.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,114 @@
...
+     * The coordinates start with 0,0 as the top left corner moving right and
+     * down.
+     *
+     * @param x The x coordinate of the pixel to get
+     * @param y The y coordinate of the pixel to get
+     * @return The color of the pixel at the specified coordinates
+     * @throws IndexOutOfBoundsException If the coordinates are greater than
+     *         the width and height of the map's canvas
+     */
+    MapColor getPixel(int x, int y);
```
evilmidget38: What's the default color if nothing has been drawn at a specific point(AKA background color)?

modwizcode: transparent, aka, map color.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Sets the specified pixel.
+     *
+     * <p>The coordinates start with 0,0 as the top left corner moving right
+     * and down.</p>
+     *
+     * @param x The x coordinate of the pixel
+     * @param y The y coordinate of the pixel
+     * @param color The color to set the pixel to
+     * @throws IndexOutOfBoundsException If the coordinates are greater than
```
evilmidget38: If the coordinates are greater than the width and height?  That would mean that getWidth and getHeight are valid coordinates, and with the coordinates starting at (0,0) that would mean that the actual width and height are one greater than the values returned by getWidth and getHeight.

src/main/java/org/spongepowered/api/map/MapCanvas.java
evilmidget38: Should be documented.

src/main/java/org/spongepowered/api/map/MapRenderer.java
```diff
@@ -0,0 +1,60 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.map;
+
+import com.google.common.base.Optional;
+import org.spongepowered.api.entity.player.Player;
+
+/**
+ * Plugin classes should implement this to modify the image shown on a map.
```
simon816: Perhaps a description of what this interface actually is? Then state the above in a `<p></p>`

src/main/java/org/spongepowered/api/map/MapRenderer.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * <p>The viewer is guaranteed to be present if {@link #isContextual()}
+     * returns true.</p>
+     *
+     * @param viewer The player viewing this render, {@link Optional#absent()}
+     *        otherwise
+     * @param view The map that's being rendered
+     * @param canvas The canvas that the renderer will draw onto
+     */
+    void render(Optional<Player> viewer, MapView view, MapCanvas canvas);
```
simon816: `Optional` isn't used for parameters, this should be `@Nullable Player viewer`

src/main/java/org/spongepowered/api/map/MapView.java
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * Sends a forced update of the map to a player. This will send only the
+     * region of the canvas between min x,y and max x,y to the player. Allowing
+     * custom trickle loading, or forced immediate updating of the map.
+     *
+     * @param player The player to send the update to
+     * @param min The upper left corner of the region to send
+     * @param max The bottom right corner of the region to send
+     */
```
simon816: Throws `PositionOutOfBoundsException`?

src/main/java/org/spongepowered/api/map/MapRenderer.java
modwizcode: This is a callback not a parameters? It's provided by the API and fits the
contract for optional perfectly.

Zidane:
Since this needs to be re-done and @modwizcode cannot be around on a regular basis...I'm closing this.

PR #532, opened 2015-04-02
Add SimpleCommandBase. Fixes #523

### LINKED ISSUE #523: Add simple implementation of CommandCallable
It would be nice to have a simple abstract implementation of ``CommandCallable`` (like ``SimpleDispatcher`` implementation for ``Dispatcher``).

The abstract class provides the basic functionality that is sufficient in many cases. The only method it should not implement is the ``call`` method. A command class extending the abstract class would look like this:

```java
public class SendCommand extends AbstractCommand {
    public SendCommand() {
        super(
                "simplemail.write", // permission
                "Send a mail", // description
                Texts.of("Send a server mail to a player."), // help
                "<player> <msg>"); // usage
    }

    @Override
    public boolean call(CommandSource source, String arguments, List<String> parents) throws CommandException {
        // Command stuff...
        return true;
    }
}
```

This is the proposed abstract command class:

```java
public abstract class AbstractCommand implements CommandCallable {
    private final String permission;
    private final String shortDescription;
    private final Text help;
    private final String usage;
    
    public AbstractCommand(String permission, String shortDescription, Text help, String usage) {
        this.permission = permission;
        this.shortDescription = shortDescription;
        this.help = help;
        this.usage = usage;
    }

    @Override
    public boolean testPermission(CommandSource source) {
        return source.hasPermission(this.permission);
    }

    @Override
    public String getShortDescription(CommandSource source) {
        return this.shortDescription;
    }

    @Override
    public Text getHelp(CommandSource source) {
        return this.help;
    }

    @Override
    public String getUsage(CommandSource source) {
        return this.usage;
    }
    
    @Override
    public List<String> getSuggestions(CommandSource source, String arguments) throws CommandException {
        return Collections.emptyList();
    }

}
```

boformer:
This PR adds a simple abstract implementation of ``CommandCallable`` (like ``SimpleDispatcher`` implementation for ``Dispatcher``).

Plugin developers can extend the class to create simple commands, without implementing all methods of ``CommandCallable``:

```java
public class SendCommand extends SimpleCommandBase {
    public SendCommand() {
        super(
                "simplemail.write", // permission
                "Send a mail", // description
                Texts.of("Send a server mail to a player."), // help
                "<player> <msg>"); // usage
    }

    @Override
    public boolean call(CommandSource source, String arguments, List<String> parents) throws CommandException {
        // command logic...
    }
}
```

I did not name this ``Command`` or ``AbstractCommand`` because I want to reserve those names for an annotation based Command API.

#523

### DIFF
--- a/src/main/java/org/spongepowered/api/util/command/SimpleCommandBase.java
+++ b/src/main/java/org/spongepowered/api/util/command/SimpleCommandBase.java
@@ -1,3 +1,27 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
 package org.spongepowered.api.util.command;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -61,5 +85,5 @@ public abstract class SimpleCommandBase implements CommandCallable {
     public List<String> getSuggestions(CommandSource source, String arguments) throws CommandException {
         return Collections.<String>emptyList();
     }
-    
+
 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/util/command/SimpleCommandBase.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Creates a new command.
+     *
+     * @param permission The permission needed to execute the command
+     * @param shortDescription A command description
+     * @param help A formatted help message
+     * @param usage A description of the command arguments
+     */
+    public SimpleCommandBase(String permission, String shortDescription, Text help, String usage) {
+        checkNotNull(permission);
```
Deamon5550: You can simplify these lines by just doing `this.permission = checkNotNull(permission);`

gabizou: You can use `checkNotNull` on assignment since they return the reference if it isn't null.

Minecrell: Or you could use `this.permission = checkNotNull(permission, "permission");` to provide the user a meaningful error message additionally

boformer:
I will make the permission optional (no permission = all users permitted) and add a second constructor.

src/main/java/org/spongepowered/api/util/command/SimpleCommandBase.java
```diff
@@ -0,0 +1,104 @@
...
+    public SimpleCommandBase(@Nullable String permission, String shortDescription, String usage) {
+        this.permission = permission;
+        this.shortDescription = checkNotNull(shortDescription, "shortDescription");
+        this.help = Texts.of(shortDescription);
+        this.usage = checkNotNull(usage, "usage");
+    }
+
+    @Override
+    public boolean testPermission(CommandSource source) {
+        return source.hasPermission(this.permission);
```
ST-DDT: There should be a null check for permissions here.
Because `Subject.testPermission(String)` does not accept null permissions as input.

boformer:
@ST-DDT fixed that.

zml2008:
I am opposed to having a big constructor like this because it a) prevents us from adding more arguments to commands in the future without adding a ton of constructors and b) worsens usability -- you have an unlabeled array of strings. (I'm also opposed to the CommandCallable megainterface -- for the same reasons relating to limited potential for expanding the API)

I've been doing some work on a new commands API, and I'm looking at using a builder pattern for setting command metadata -- take a look at https://github.com/PEXPlugins/PermissionsEx/blob/master/src/main/java/ninja/leaping/permissionsex/command/ParentCommands.java#L35

boformer:
@zml2008 Which arguments do you want to add to the command system? 

I also think that it **improves** usability by drastically reducing the amount of code for simple commands.

I would also prefer an annotation-based command-system, but that is something that can be built on top of the existing system. 
Your system can also be built on top of the existing system.

Zidane:
Now handled by zml's Commands Overhaul PR. Closing this in favor of it.

PR #539, opened 2015-04-04
[WIP] Revised Dynamic Commands

AlphaModder:
This PR modifies the default command system so that it allows commands to be dynamically resolved based on their alias. Using this, the "plugin:command" notation does not need to be a duplicate of every command registered, as the code can check the prefix when the command is executed and decide which plugin to run from.

This also opens up the possibility of using minecraft or forge-mod commands "minecraft:" or "mod:" (at least in the forge implementation), and completely solves the issue of command conflicts, as a single alias could be registered by more than one plugin and each command would still be individually addressable.

Finally, it also allows plugins to take control of this system, and explicitly set the plugin that will be used when a non-prefixed command is used by a certain kind of CommandSource. A way to modify this behavior through a configuration file could be in either Sponge or a sponge plugin. This would be extremely useful for sponge servers with "Essentials"-type plugins that still want to play vanilla adventure maps. 

The whole thing is based on [EvilSeph's blog post](http://evilseph.com/post/75608917830/bukkit-supporting-minecraft-commands).


src/main/java/org/spongepowered/api/plugin/Plugin.java
```diff
@@ -55,6 +55,15 @@
     String name();
 
     /**
+     * Gets the prefixes that can be prepended to a command to indicate this
+     * plugin's version of said command. This must have at least one item.
+     * 
+     * @return The prefix prepended to a command to indicate this plugin's
+     *         version of said command.
+     */
+    String[] commandPrefixes();
```
Deamon5550: can't we just use the plugin's id for this? Keeps things consistent and automatically prevents overlaps since Ids should already be unique.

AlphaModder: @Deamon5550 Plugins may want to have multiple, minecraft almost certainly would. (mc:give, minecraft:give), I don't see why not to open this option up to plugins. Though I will change the fallback from .getName() to .getId()

Deamon5550: Then I would suggest not making the Id a fallback, but rather the primary prefix. Once again for consistency.

Deamon5550:
Lot of missing javadocs, make sure you get them.

src/main/java/org/spongepowered/api/plugin/Plugin.java
AlphaModder: @Deamon5550 Alright, I'll do that.

AlphaModder: @Deamon5550 I guess I'll mention that PluginContainer impls should map getRegistrarName to getId

src/main/java/org/spongepowered/api/plugin/Plugin.java
```diff
@@ -82,5 +82,13 @@
      */
     String dependencies() default "";
 
+    /**
+     * Gets the prefixes that can be prepended to a command to indicate this
+     * plugin's version of said command. This must have at least one item.
```
simon816: There doesn't have to be at least one item if you state that the default prefix is to use the plugin ID, which is most likely what plugin devs would use as the primary prefix anyway.

src/main/java/org/spongepowered/api/service/command/CommandService.java
```diff
@@ -44,72 +45,90 @@
...
      *
-     * @param plugin A plugin instance
+     * @param registrar The object registering the command. Can be a class
+     *        annotated by {@link Plugin}, or anything implementing
+     *        {@link CommandRegistrar}.
      * @param callable The command
-     * @param aliases A list of aliases
-     * @return The registered command mapping, unless no aliases could be registered
-     * @throws IllegalArgumentException Thrown if {@code plugin} is not a plugin instance
+     * @param alias An array of aliases
```
simon816: No reason to change "list" to "array" here. In fact, 'list' is more correct because it's `List<String>` not `String[]`

src/main/java/org/spongepowered/api/service/command/CommandService.java
```diff
@@ -44,72 +45,90 @@
...
-     * @return The previous mapping associated with the alias, if one was found
+     * @return The registered command mapping, unless no aliases could be
+     *         registered
+     * @throws IllegalArgumentException Thrown if new conflicting aliases are
+     *         added in the callback
+     * @throws IllegalArgumentException Thrown if {@code plugin} is not a plugin
+     *         instance
+     * @throws IllegalArgumentException Thrown if a command name is invalid.
      */
-    Optional<CommandMapping> remove(String alias);
```
simon816: Why can't aliases be removed?

src/main/java/org/spongepowered/api/service/command/CommandService.java
```diff
@@ -120,19 +139,20 @@
     Optional<CommandMapping> removeMapping(CommandMapping mapping);
 
     /**
-     * Get a set of plugin containers that have commands registered.
+     * Get a set of {@link CommandRegistrar}s that have registered command(s)
+     * with this CommandService.
      *
-     * @return A set of plugin containers
+     * @return a set of CommandRegistrars
```
simon816: Capital 'A' is correct. Undo this change

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -283,9 +283,19 @@ public synchronized boolean removeMappings(Collection<?> c) {
...
     @Override
-    public synchronized Optional<CommandMapping> get(String alias) {
+    public synchronized Optional<CommandMapping> resolveMapping(String alias, CommandSource source) {
         return Optional.fromNullable(this.commands.get(alias.toLowerCase()));
     }
+    
+    @Override
+    public synchronized Set<CommandMapping> getAll(String alias) {
+        Set<CommandMapping> set = new HashSet<CommandMapping>();
+        Optional<CommandMapping> mapping = Optional.fromNullable(this.commands.get(alias.toLowerCase()));
```
simon816: No reason to wrap this in an Optional to just unwrap it again. Just check for whether it's null or not.

src/main/java/org/spongepowered/api/service/command/sponge/SpongeCommandService.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @Override
+    public boolean testPermission(CommandSource source) {
+        return true;
+    }
+
+    @Override
+    public String getShortDescription(CommandSource source) {
+        return "";
```
simon816: These things should be specified in the constructor parameters ideally. Similar to [`SimpleDispatcher`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java#L77).

simon816:
What are the differences between `SimpleCommandService`/`SimpleDispatcher` and `SpongeCommandService`/`SpongeDispatcher`? There is some overlap in some places and plugin devs may not know which one to pick when creating commands.

boformer:
The ``CommandRegistrar`` interface is useless.

Sponge uses a wrapper to register Minecraft commands, and plugin containers can be created manually if necessary. Look at the implementation:
https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/plugin/SpongePluginManager.java#L65

Revert the registrar stuff.

AlphaModder:
@boformer The CommandRegistrar is needed for the prefix system. It could also be used for forge mod support or similar on other implementations.

src/main/java/org/spongepowered/api/service/command/sponge/SpongeCommandService.java
AlphaModder: @simon816 Alright.

AlphaModder:
@simon816 
SimpleDispatcher should be used if a plugin wants to make a dispatcher for it's subcommands, obviously you would not want /plugin mc:give to work. 
SpongeDispatcher/CommandService is what all implementations of Sponge should use for the default command system.
SimpleCommandService can be extended for when plugins want to make a custom command service.

src/main/java/org/spongepowered/api/service/command/CommandService.java
AlphaModder: @simon816 Because that would not be at all useful in some systems. SimpleCommandService supports it, but it would not make sense in SpongeCommandService, as you might remove multiple commands from multiple plugins at the same time.

boformer:
I'm still not happy with the Sponge- prefix.

The name should describe the behavior of the class.

AlphaModder:
@boformer I've got a few names, dunno which to use:
DynamicCommandService / DynamicDispatcher
ContextualCommandService / ContextualDispatcher
RootCommandService / RootDispatcher

boformer:
ContextualCommandService sounds good to me

AlphaModder:
Aaaand the build failed for no reason. Fantastic.

boformer:
This issue is related to your CommandRegistrar:
 https://github.com/SpongePowered/SpongeAPI/issues/426

It explains why commands should be registered by plugin objects.

src/main/java/org/spongepowered/api/service/command/sponge/CommandRegistrar.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.service.command.CommandService;
+
+import java.util.Set;
+
+/**
+ * An object that can register commands with a {@link CommandService}. When
+ * implementing {@link PluginContainer}, getRegistrarName should be the same as
+ * getId, and getFriendlyName the same as getName.
```
ST-DDT: The method naming is strange.
Why don't you simply use `getId()` and `getName()`?

AlphaModder:
@ST-DDT That only applies to the forge implementation, for one. Granite mods would not work with it. Second, it is still required for prefixes.

Zidane:
Now handled by zml's Commands Overhaul PR. Closing this in favor of it.

AlphaModder:
@Zidane zml's PR does not address this issue.

PR #541, opened 2015-04-04
Refactor GameRegistry

gabizou:
GameRegistry as we know it is bloated with all the `getXX(): Optional<?>` methods when all of our "types" aren't conforming to a common standard, some "types" have a `getName()` and others have `getId()`. This PR attempts to unify the types and clean up GameRegistry along with it. 

Adds a handy `CatalogTypes` for indexing known `CatalogType` classes for easy auto completion in Eclipse.

### DIFF
--- a/src/main/java/org/spongepowered/api/CatalogTypes.java
+++ b/src/main/java/org/spongepowered/api/CatalogTypes.java
@@ -1,0 +1,129 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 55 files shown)


Deamon5550:
+1

src/main/java/org/spongepowered/api/CatalogTypes.java
```diff
@@ -0,0 +1,129 @@
...
+import org.spongepowered.api.world.biome.BiomeType;
+import org.spongepowered.api.world.difficulty.Difficulty;
+import org.spongepowered.api.world.weather.Weather;
+
+/**
+ * Enumeration of all known {@link CatalogType}s for autocompletion when using the
+ * {@link GameRegistry#getType(Class, String)} and {@link
+ * GameRegistry#getAllOf(Class)}.
+ */
+public final class CatalogTypes {
```
gratimax: You missed one :stuck_out_tongue:. CatalogTypes is a catalog of catalog types, too!

ZephireNZ:
Nice work! How woud the implementations likely work? Check if the provided class implements `CatalogType`, and if so, check for the `CataloguedBy` annotation, and provide that class?

Minecrell:
IMO we should get the builders out of there. The game registry should be really only related to working with the catalog types and not also for creating a builder.

gabizou:
> IMO we should get the builders out of there. The game registry should be really only related to working with the catalog types and not also for creating a builder.

I was contemplating on whether this PR should handle refactoring the remaining builders and extra getters by mimicking the Text builders. Maybe we should have a Builder of builders?

src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -50,7 +51,7 @@
      *
      * @return The id
      */
-    String getId();
+    String getName();
```
ST-DDT: What is the id of the block then? The numerical id?
Because there is only a lookup method in the game registry for getById and not for getByName anymore.
(The javadocs may be misleading now)

src/main/java/org/spongepowered/api/entity/living/animal/HorseStyle.java
```diff
@@ -33,13 +34,6 @@
...
 @CatalogedBy(HorseStyles.class)
-public interface HorseStyle extends DataSerializable {
-
-    /**
-     * Gets the name of this style.
-     *
-     * @return The name of this style
-     */
-    String getName();
+public interface HorseStyle extends CatalogType {
```
ST-DDT: Is the removal of `extends DataSerializable` intentional?
Because the import has not been removed.

This also has happened in some other classes.

src/main/java/org/spongepowered/api/block/BlockType.java
kenzierocks: The ID is `block_iron`, `block_gold`, `dirt`, etc. It's pretty much a name.

src/main/java/org/spongepowered/api/entity/living/animal/HorseStyle.java
gabizou: Yes. Types shouldn't be DataSerializable. The DataHolder should be the one to decide how the type is serialized.

src/main/java/org/spongepowered/api/block/BlockType.java
Deamon5550: CatalogType has the getId method, this is simple overriding the getName method to replace the javadoc.

gabizou:
@Zidane I do believe this is at a point of "pull or not pull", getting the #542 PR ready almost relies on this being merged (I say almost because I don't want to have to add `Optional<Foo> getFoo(String)` for each of the new types I've added in the PR).

PR #543, opened 2015-04-04
Replace WorldBorder in WorldProperties with individual methods

Deamon5550:
Fixes the issue of requiring a WorldBorder object within the WorldProperties. Since WorldProperties can exist without the associated world being loaded this places the addition requirement on WorldBorder implementations that they can exist without the associated world which is not always correct.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/WorldBorder.java
+++ b/src/main/java/org/spongepowered/api/world/WorldBorder.java
@@ -185,7 +185,7 @@ public interface WorldBorder {
      *
      * @return The damage amount
      */
-    int getDamageAmount();
+    double getDamageAmount();
 
     /**
      * Set the damage done to a player per block per tick when outside the
@@ -193,5 +193,5 @@ public interface WorldBorder {
      *
      * @param damage The damage amount
      */
-    void setDamageAmount(int damage);
+    void setDamageAmount(double damage);
 }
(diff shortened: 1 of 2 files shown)


Zidane:
@bloodmc thoughts?

modwizcode:
Why don't we have two implementations of worldborder? Mixin and Sponge, they will transparently be handled correctly. Similar to Mumfrey's suggestion for MapColor, we could just have two implementations and WorldProperties will handle the magic of making sure the right border is setup. 

The implementation would be similar to this change because of the need to store the values, but it would make the separation between WorldProperties and WorldBorder much cleaner.

clienthax:
shouldnt damage be a double?

Deamon5550:
@clienthax I just took those methods from our world border interface, but indeed you are correct they should both be doubles. Fixed now.

@modwizcode The problem with having two instances in my opinion is that its just asking for someone to hang on to a copy of one of them and have it become invalid as the world state changes.

modwizcode:
@Deamon5550 I feel like the more sane solution is to tell people not to cache it and to just grab an instance when they need to make their changes. That's just my 2 cents though.

ST-DDT:
AFAIK caching these border objects could as well lead to a resource leak if the reference is not disposed automatically.

Zidane:
@bloodmc is changing how this works a bit in the impl and so the API will be changing too.

Zidane:
Ignore my comment, I forgot what this PR was meant for.

ryantheleach:
Make worldproperties/world a data holder and worldBorder a datatype.

You can now serialize worldborders to configs, offer them to players for player specific world borders, set them on a loaded world, or set their value on a worldProperties.

As a part of this the other world settings could become datatypes as well, allowing for easy serialization / loading from templates.

PR #544, opened 2015-04-05
Expand Populator System

Deamon5550:
In order to finish fleshing out the WorldGenerator the last piece is per-biome populators. So these would be populators which are registered into individual biomes and run on a per biome basis. Additional to the populators each biome needs a special case of the GeneratorPopulator except that it only handles a single collumn within the BlockBuffer in order to replace the blocks within that column to the correct counterparts for the biome. This has been tentatively named a ColumnPopulator (name suggestions wanted).

The order of operations for world generation:
---------------------------------------------
The generation process for chunks is divided into two phases, generation and population. The generation phase is in charge of creating the base terrain shape and generating large terrain features. All operations during the generation phase act upon a BlockBuffer rather than a live chunk object.

Conversely the population phase operates against a live chunk object and has the guarantee that all immediately surrounding chunks have at least passed the generation phase. The population phase is typically used for the placement of small features and objects placed my cross chunk boundaries.

**The generation phase:**
 1. Create the chunk buffer
 2. Call the base GeneratorPopulator to create the base terrain shape
 3. Pass each 1x1 collumn of the chunk to the BiomePopulator
 4. Call each of the GeneratorPopulators registered to the biome
 5. Call each of the GeneratorPopulators registered to the generator
 6. Build the chunk object from the chunk buffer

**The population phase:**
 1. Validate surrounding chunks
 2. Pass the chunk to each of the populators registered to the biome (based off the biome of an arbitrary point in the chunk (16, 0, 16 in vanilla))
 3. Pass the chunk to each of the populators registered to the generator

Additionally
--------------------------------------------
This PR also adds a number of the vanilla `Populator`s, accessible through the `PopulatorFactory` (via the game registry). which may be modified and added to existing worlds or custom world generators in order to gain easy access to the vanilla population procedures.

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -62,6 +62,9 @@ import org.spongepowered.api.util.rotation.Rotation;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.WorldBuilder;
 import org.spongepowered.api.world.WorldCreationSettings;
+import org.spongepowered.api.world.gen.GeneratorPopulator;
+import org.spongepowered.api.world.gen.Populator;
+import org.spongepowered.api.world.gen.PopulatorFactory;
 import org.spongepowered.api.world.gen.WorldGeneratorModifier;
 
 import java.awt.image.BufferedImage;
@@ -452,5 +455,13 @@ public interface GameRegistry {
      * @param modifier The modifier to register
      */
     void registerWorldGeneratorModifier(PluginContainer plugin, String genId, WorldGeneratorModifier modifier);
+    
+    /**
+     * Gets the {@link PopulatorFactory} for creating {@link Populator}s and
+     * {@link GeneratorPopulator}s.
+     * 
+     * @return The populator factory
+     */
+    PopulatorFactory getPopulatorFactory();
 
 }
--- a/src/main/java/org/spongepowered/api/data/manipulators/MobSpawnerData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulators/MobSpawnerData.java
@@ -70,7 +70,7 @@ public interface MobSpawnerData extends DataManipulator<MobSpawnerData> {
     short getMinimumSpawnDelay();
 
     /**
-     * Gets the minimum delay between batches of monsters.
+     * Sets the minimum delay between batches of monsters.
      * <p>
      * Each time the timer is reset the new delay is chosen randomly from
      * between the minimum (inclusive) and maximum (exclusive) delays.
@@ -92,7 +92,7 @@ public interface MobSpawnerData extends DataManipulator<MobSpawnerData> {
     short getMaximumSpawnDelay();
 
     /**
-     * Gets the maximum delay between batches of monsters.
+     * Sets the maximum delay between batches of monsters.
      * <p>
      * Each time the timer is reset the new delay is chosen randomly from
      * between the minimum (inclusive) and maximum (exclusive) delays.
--- a/src/main/java/org/spongepowered/api/util/WeightedRandomItemStack.java
+++ b/src/main/java/org/spongepowered/api/util/WeightedRandomItemStack.java
@@ -23,21 +23,45 @@
  * THE SOFTWARE.
  */
 
-package org.spongepowered.api.world;
+package org.spongepowered.api.util;
+
+import org.spongepowered.api.item.inventory.ItemStack;
 
 /**
- * An enumeration of default {@link GeneratorType}s.
+ * Represents an item stack with a range of possible quantities and a numerical weight used for random selection
+ * from a collection of weighted types.
  */
-public final class GeneratorTypes {
+public interface WeightedRandomItemStack {
 
-    public static final GeneratorType DEBUG = null;
-    public static final GeneratorType DEFAULT = null;
-    public static final GeneratorType FLAT = null;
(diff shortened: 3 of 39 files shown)


DDoS:
I'm not sure I understand what generation step 3 is for. Could you clarify?

src/main/java/org/spongepowered/api/world/gen/ColumnPopulator.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Operates on a single 1x1 column of the given {@link MutableBlockBuffer}.
+     * The given x,z position is the location of the column relative to the
+     * buffer.
+     *
+     * @param buffer The buffer to apply the changes to. The buffer can be of
+     *            any size.
+     * @param biome The biome of the targeted column
+     * @param x The X position of the column, relative to the buffer
+     * @param z The Z position of the column, relative to the buffer
```
DDoS: There's no need to mention relativeness here, all coordinates given should be valid in the buffer. You should be able to call

    buffer.setBlock(x, z, ...);

Instead of

    Vector3i blockMin = buffer.getBlockMin();
    buffer.setBlock(blockMin.getX() + x, blockMin.getZ() + z, ...);

Deamon5550:
@DDoS The replacing of blocks from the base terrain with the biome specific blocks (for example sand and sandstone for deserts) needs to be done for each column of the chunk. So each 1x256x1 column is sent to the correct ColumnPopulator for the biome of that column.

DDoS:
@Deamon5550 I see. I had something similar for Spout's Minecraft clone plugin, but I just had that implemented as a regular generate populator. Is it really necessary to add a stage for column population? Instead it might be more useful to provide a generic implementation of a replacer for surface blocks. This is what I had: https://github.com/SpoutDev/Vanilla/blob/master/src/main/java/org/spout/vanilla/world/generator/normal/populator/GroundCoverPopulator.java

Deamon5550:
@DDoS Yeah we could do something similar where biomes provide an array of ground cover layers. The reason I'm not jumping to make the change is it feels like it would have around the same level of complexity without as much customization for plugins as that which is provided by the ColumnPopulator. But it could go either way I guess.

Thoughts on this @SpongePowered/developers / @Adubbz?

DDoS:
@Deamon5550 If you need something more flexible, you use a custom generator populator instead of the default ground cover one.

Deamon5550:
@DDoS While that's true its a bit awkward as a solution as it requires you to support all biomes potentially just to make one of them more custom. Additionally it restricts the number of plugins which can have custom biomes to one, whereas with ColumnPopulators a number of plugins can have custom biomes so long as there aren't overlaps.

Adubbz:
From my standpoint, ground cover layers are something we make heavy across many of our biomes, and I would say it's a similar case for other biome mods as well. If a 'standard' for such a feature isn't provided, i'd expect it to be something biome plugins would all be implementing themselves.

DDoS:
@Deamon5550 Right, so the cover generator only applies to columns that have a biome with a defined ground cover stack, the rest is untouched. A secondary custom generator can than pass to handle the custom cover, including the untouched layers.

Deamon5550:
@DDoS @Adubbz Ah alright so we can have a (Possibly empty) List of ground cover layers, and an Optional ColumnPopulator?

Se the new generation phase would look like this:

**The generation phase:**
 1. Create the chunk buffer
 2. Call the base GeneratorPopulator to create the base terrain shape
 3. Pass the List of ground cover layers to the GoundCoverPopulator (if the list is not empty)
 4. Pass each 1x1 collumn of the chunk to the ColumnPopulator (if it exists)
 5. Call each of the GeneratorPopulators registered to the biome
 6. Call each of the GeneratorPopulators registered to the generator
 7. Build the chunk object from the chunk buffer

This way your more typical biome can just set some ground cover layers. But a plugin wanting finer control and define a custom ColumnPopulator for a biome.

Sounds good?

DDoS:
I was trying to get rid of the column populator entirely:

1. Create the chunk buffer
2. Call the base GeneratorPopulator to create the base terrain shape
3. Pass the buffer to the built-in ground cover populator
  i. For each column, check if the biome has cover layers
  ii. If so apply them
  iii. Else skip the column. These will be custom and done next
4. Call each of the GeneratorPopulators registered to the generator
  i. Custom ground cover generator populator is ran
  ii. Checks if biome uses the custom cover for the column
  iii. If so apply them
  iv. Else skip the column 
  v. This might repeat for other custom cover populators

I'm also curious on how you plan to run generator populators for each biome since a buffer can have a different biome for each column.

Deamon5550:
Ah alright yeah I guess we can drop the ColumnPopulator

>I'm also curious on how you plan to run generator populators for each biome since a buffer can have a different biome for each column.

Yeah you can't really. In vanilla it just runs the populators as would apply to the biome at (x*16+16, 0, z*16+16). Since populators apply to whole chunks there isn't really much we can do about this.

DDoS:
@Deamon5550 Yeah, Spout did the same, although we didn't have per biome populator generators (only per generator). Per biome was only for regular populators.

Deamon5550:
Yeah Vanilla has only regular populators per biome as well, but there are some things better done in the generation phase, before the other GeneratorPopulators (caves and such) are applied.

Adubbz:
@Deamon5550 Sounds good to me. Although, rather than an empty list, by default it really should contain a layer of grass and a layer of dirt. Unless of course you guys have moved that to an earlier phase.

Issue #540, opened 2015-04-04
Get location's chunk.


nikosgram13:
I'm working to a chunk protection plugin and i need to get the location's chunk.

Example:
```java
location.getChunk(); //return Optional<Chunk>
```

or from a block:
```java
block.getChunk(); //return Optional<Chunk>
```

DDoS:
Right now you can do the following:

```java
Chunk getChunk(Location location) {
    Extent extent = location.getExtent();
    if (extent instance of Chunk) {
        return (Chunk) extent;     
    } else if (extent instance of World) {
        return ((World) extent).getChunk(location.getBlockPosition().div(16));
    }
}
```

Of course this would be really annoying to have to copy-paste as a utility method. I suggest adding this to the Location class.

gratimax:
It's completely possible (even if unreasonable) for an implementation to provide other types of extent, like "Region". I would think that only way to handle this would be to consider more carefully what location means with regards to different types of extents, and either generifying Extent or adding some methods to World.

DDoS:
I'm unsure how you'd want to handle concepts that are foreign to the API. We're using a World made of Chunks made of Blocks approach, to match Minecraft. If you want to support a concept such as "Region", you need to delegate it getChunk since it's foreign. But adding getChunk to Extent isn't desirable as Extent my not have anything to do with Chunks.

I'm not sure of how you'd want to cope with other world structures...

simon816:
Some relevant discussion was made on https://github.com/SpongePowered/SpongeAPI/pull/322 and https://github.com/SpongePowered/SpongeAPI/issues/453

DDoS:
Oddly enough, I think not supporting getters to subclasses of Extent would be the best course of action, now that I think about it. Else there's no end in sight and I don't like the idea of only supporting the API concepts.

ryantheleach:
Why would you need to protect chunks as opposed to 16x16x256 area's? (keeping in mind that chunks are variable in size, and possibly 3d) it seems to me that using chunks for anything but low level loading / unloading / generation is a mistake.

PR #545, opened 2015-04-06
Add getChunk() to Location

DDoS:
Proposed solution to https://github.com/SpongePowered/SpongeAPI/issues/540


DDoS:
See related issue.

Issue #326, opened 2014-12-27
Command Stats / command block support


Kobata:
Much of the command block 'scripting' ability comes from the command stats concept.

There are five of these in vanilla, `SuccessCount`, `AffectedBlocks`, `AffectedEntities`, `AffectedItems`, and `QueryResult`. Of these, `SuccessCount` and `AffectedEntities` have special support inside the command processor.

When running a command, vanilla expands the main target selector, and then runs the command *once for each targeted entity.* During this process it keeps track of how many times the command succeeded (defined as 'did not throw an exception' because Mojang), this count becomes `SuccessCount` and also the return value of the method as command blocks store the count separately for comparator support. The expanded selector's full count becomes the value of `AffectedEntities`

The rest of the stats (and occasionally `AffectedEntities` for commands that do not use the standard selector expansion, see `/spreadplayers`) are set by each individual command by calling a function on the command sender.

The basic support for the non-special stats is fairly easy, but getting the automatic `SuccessCount`/`AffectedEntities` support requires a bunch of changes to the default command implementation in `SimpleCommandService`/`SimpleDispatcher`, and likely some changes to the command interfaces.

modwizcode:
@Kobata are you working on a PR for this, or can I start working on this?

zml2008:
I feel like this is something that should be handled internally to the command executor. Not all plugins need a success count or want to be repeatedly called when multiple matching entities are provided (because some plugins handle multiple targets *properly*). (and relatedly, this is handled for any minecraft ICommands registered in our wrapper executor)

Kobata:
@zml2008 Don't touch this issue if you don't understand what it is about, and have not actually addressed that issue.

Issue #551, opened 2015-04-06
Fire CommandEvent on SimpleCommandService.call()


boformer:
The ``SimpleCommandService`` does not fire a command event when the ``call`` method is used.

I don't know if this is intended. When a plugin uses the ``call`` method, there would be no way for other plugins to cancel the command execution.

zml2008:
It might make sense, rather than the MC ServerCommandManager calling a CommandEvent which is picked up by the SimpleCommandService, to have the ServerCommandManager directly use the call method, and have the SimpleCommandService call the CommandEvent.

Another issue is integrating the Forge CommandEvent with ours. This may be more complicated because the forge event requires an ICommandSender and an ICommand, which we don't really have if we don't use the MC CommandHandler.

PR #552, opened 2015-04-06
[WIP] Add CommandResult

boformer:
Continuation of https://github.com/SpongePowered/SpongeAPI/pull/409 and https://github.com/SpongePowered/SpongeAPI/pull/524

This issue explains what command stats are used for: https://github.com/SpongePowered/SpongeAPI/issues/326. I don't know why it was closed.

This PR replaces the ``boolean`` that is returned when a command is called with a ``CommandResult`` object, which stores the command stats ``SuccessCount``, ``AffectedBlocks``, ``AffectedEntities``, ``AffectedItems`` and ``QueryResult``. I also added a ``Map`` for custom data.

The ``CommandResult`` is passed with the ``CommandEvent`` after the execution of the command.

The implementation is based on @AlphaModder's pull request, but uses classes instead of interfaces to make it easier to use the system (similar to the Text API). 

The only thing that changes is that plugin developers now have to use a builder to create a ``CommandResult``:

```java
MyCommand implements CommandCallable {
    CommandResult call(CommandSource source, String arguments, List<String> parents) {
        ...
        return CommandResults.processed(); 
        // equal to CommandResults.builder().setProcessed(true).build();
    }
    ...
}
```

I'm not sure if this is the best way of implementing it. I also thought about a system that uses separate classes for each result data type (similar to the Data API proposal):

```java
CommandResult {
    boolean wasProcessed();
    <T> Optional<T> getData(Class<T extends CommandResultData> clazz);
}
```

### DIFF
--- a/src/main/java/org/spongepowered/api/event/message/CommandEvent.java
+++ b/src/main/java/org/spongepowered/api/event/message/CommandEvent.java
@@ -27,7 +27,11 @@ package org.spongepowered.api.event.message;
 
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.GameEvent;
+import org.spongepowered.api.util.command.CommandResult;
 import org.spongepowered.api.util.command.CommandSource;
+import com.google.common.base.Optional;
+
+import javax.annotation.Nullable;
 
 /**
  * Fired when a command has been used and needs to be processed.
@@ -60,5 +64,20 @@ public interface CommandEvent extends GameEvent, Cancellable {
      * @return The arguments
      */
     String getArguments();
+
+    /**
+     * The result of the command. This is only available after the execution 
+     * of the command.
+     * 
+     * @return The result of the command, if present
+     */
+    Optional<CommandResult> getResult();
+
+    /**
+     * Sets the result of the command.
+     * 
+     * @param result The result of the command, or null
+     */
+    void setResult(@Nullable CommandResult result);
 
 }
--- a/src/main/java/org/spongepowered/api/util/command/CommandResult.java
+++ b/src/main/java/org/spongepowered/api/util/command/CommandResult.java
@@ -55,7 +55,7 @@ public class CommandResult {
      * @param queryResult The query result
      * @param resultInfo A map storing additional information
      */
-    protected CommandResult(boolean processed, @Nullable Integer successCount, @Nullable Integer affectedBlocks, @Nullable Integer affectedEntities,
+    CommandResult(boolean processed, @Nullable Integer successCount, @Nullable Integer affectedBlocks, @Nullable Integer affectedEntities,
             @Nullable Integer affectedItems, @Nullable Integer queryResult, @Nullable Map<String, Object> resultInfo) {
         this.processed = processed;
         this.successCount = Optional.fromNullable(successCount);
--- a/src/main/java/org/spongepowered/api/util/command/CommandResult.java
+++ b/src/main/java/org/spongepowered/api/util/command/CommandResult.java
@@ -115,7 +115,8 @@ public class CommandResult {
     }
 
     /**
-     * Gets the query result of the command.
+     * Gets the query result of the command, e.g. the time of the day,
+     * an amount of money or a player's amount of XP.
      *
      * @return The query result of the command, if one exists
      */
--- a/src/main/java/org/spongepowered/api/util/command/CommandResult.java
+++ b/src/main/java/org/spongepowered/api/util/command/CommandResult.java
@@ -1,0 +1,135 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.command;
+
+import com.google.common.base.Optional;
+
+import java.util.Collections;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents the result of a command in Sponge.
+ */
+public class CommandResult {
+
+    private final boolean processed;
+    private final Optional<Integer> successCount;
+    private final Optional<Integer> affectedBlocks;
+    private final Optional<Integer> affectedEntities;
+    private final Optional<Integer> affectedItems;
+    private final Optional<Integer> queryResult;
+    private final Map<String, Object> resultInfo;
+
+    /**
+     * Constructs a new command result.
+     * 
+     * @param processed True if the command was processed
+     * @param successCount The success count
+     * @param affectedBlocks The number of affected blocks
+     * @param affectedEntities The number of affected entities
+     * @param affectedItems The number of affected items
+     * @param queryResult The query result
+     * @param resultInfo A map storing additional information
+     */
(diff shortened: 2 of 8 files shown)


Minecrell:
The code style is wrong in some files

src/main/java/org/spongepowered/api/util/command/CommandResult.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @return The number of items affected by the command, if such a count
+     *         exists
+     */
+    public Optional<Integer> getAffectedItems() {
+        return this.affectedItems;
+    }
+
+    /**
+     * Gets the query result of the command.
```
Deamon5550: Need better docs here for what the query result is.

src/main/java/org/spongepowered/api/util/command/CommandResult.java
```diff
@@ -0,0 +1,135 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents the result of a command in Sponge.
+ */
+public class CommandResult {
+
+    private final boolean processed;
+    private final Optional<Integer> successCount;
```
Deamon5550: Any reason for optionals and not simply defaulting to zero?

src/main/java/org/spongepowered/api/util/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * A builder for {@link CommandResult}s.
+ */
+public class CommandResultBuilder {
+
+    private boolean processed;
+    private Integer successCount;
```
Deamon5550: Same question as above about not having defaults

src/main/java/org/spongepowered/api/util/command/CommandResult.java
AlphaModder: @Deamon5550 There is a difference between a success count of 0 (no successes) and absent (not applicable).

gabizou: Wouldn't having a not-applicable successCount imply that the Command failed to succeed?

AlphaModder: @gabizou Good for commands where it wouldn't make sense. /motd, for instance. Doesn't make sense for it to succeed or not.

gabizou: But an motd command being executed would succeed once.

AlphaModder: @gabizou Yes, you could say it always succeeds, but how is that useful information? There'd be no use in knowing the success count, so why store it?

Xemiru: @AlphaModder Command blocks?

src/main/java/org/spongepowered/api/event/message/CommandEvent.java
```diff
@@ -61,4 +65,19 @@
...
+     * @return The result of the command, if present
+     */
+    Optional<CommandResult> getResult();
+
+    /**
+     * Sets the result of the command.
+     * 
+     * @param result The result of the command, or null
+     */
+    void setResult(@Nullable CommandResult result);
```
boformer: Right now there is a problem with this setter (tests fail).

Should I change it to ``setResult(Optional<CommandResult> result)``?

ST-DDT: No Afaik the optional is only for return types.

boformer: That's true, but the event system does not work if the setter does not use the same type.

simon816: You haven't added the key 'result' to the [`SpongeEventFactory`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java#L909)

zml2008:
I'm not sure about even having the processed flag in here. If we go with what I proposed in #551, the SimpleCommandService call method can handle all the logic of calling the event and printing not found messages if necessary.

I definitely think adding a map for custom attributes is too much at this point in time. If somebody can give an actual (meaning something they're trying to do in a plugin), specific use case for it that is a lot more complicated without, then we can always add it in later.

Also, you should expand on your javadocs to explain what each value does in terms of the game (like: entities affected is entities matched by arguments, success count is entities successfuly affected (why? who knows)

I'll put in a few line comments on other little things.

src/main/java/org/spongepowered/api/util/command/CommandResult.java
```diff
@@ -0,0 +1,135 @@
...
+     * 
+     * @param processed True if the command was processed
+     * @param successCount The success count
+     * @param affectedBlocks The number of affected blocks
+     * @param affectedEntities The number of affected entities
+     * @param affectedItems The number of affected items
+     * @param queryResult The query result
+     * @param resultInfo A map storing additional information
+     */
+    protected CommandResult(boolean processed, @Nullable Integer successCount, @Nullable Integer affectedBlocks, @Nullable Integer affectedEntities,
```
zml2008: This should be private or package-local (no spec) or this class should be marked final. We only want people instantiating it through the builder

src/main/java/org/spongepowered/api/util/command/CommandResults.java
```diff
@@ -0,0 +1,65 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.command;
+
+/**
+ * Utility class to create {@link CommandResult}s.
+ */
+public class CommandResults {
```
zml2008: these methods should be in CommandResult

src/main/java/org/spongepowered/api/util/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+import com.google.common.collect.Maps;
+
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * A builder for {@link CommandResult}s.
+ */
+public class CommandResultBuilder {
```
zml2008: This could probably also work as an inner class of CommandResult. That's what people usually do for not-huge builders.

src/main/java/org/spongepowered/api/util/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+        this.resultInfo = resultInfo;
+        return this;
+    }
+
+    /**
+     * Builds the {@link CommandResult}.
+     *
+     * @return A CommandResult with the specified settings
+     */
+    CommandResult build() {
```
Minecrell: public?

src/main/java/org/spongepowered/api/util/command/CommandResults.java
Minecrell: Not really, for consistency we should use this here too. The Text API consists also of classes and there are also methods in Texts instead of directly in Text

src/main/java/org/spongepowered/api/util/command/CommandResults.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Returns a {@link CommandResultBuilder}.
+     * 
+     * @return A new command result builder
+     */
+    public static CommandResultBuilder builder() 
+    {
+        return new CommandResultBuilder();
```
Minecrell: Code style

src/main/java/org/spongepowered/api/util/command/CommandResults.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Returns a new {@link CommandResult} indicating that a command was processed.
+     * 
+     * @return The command result 
+     */
+    public static CommandResult processed() 
+    {
+        return new CommandResultBuilder().processed(true).build();
```
Minecrell: If this never changes you can also make this a constant and then return it (see Text API for some examples)

src/main/java/org/spongepowered/api/util/command/CommandResults.java
boformer: I think the Text API is a special case because there are many different types and helper methods.

It just bloats up the API if we have 3 classes different classes for commandresults.

Minecrell: The decision for using Texts instead of Text was not because there are a lot of helper methods or different types, but rather only because we have the catalog lists always in extra classes instead of directly in the interface. It would be confusing if you use an extra class for some things but no extra classes for other things

Minecrell: This class should be final by the way

src/main/java/org/spongepowered/api/util/command/CommandResultBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Sets the query result of the command, e.g. the time of the day,
+     * an amount of money or a player's amount of XP.
+     *
+     * @param queryResult The query result of the command
+     * @return This builder, for chaining
+     */
+    public  CommandResultBuilder queryResult(@Nullable Integer queryResult) {
```
Minecrell: Unnecessary space

src/main/java/org/spongepowered/api/util/command/CommandResults.java
zml2008: Builders in general should be treated like extended constructors -- meaning the methods to create a builder should be in the class they are appropriate for. Builders should only be in a separate class if the builder can return multiple types (like multiple implementations of an interface). The reason Texts is in a separate class is because there are utility methods that create Texts from different types of objects, so it's less confusing to have everything in one place.

Zidane:
This is now a part of zml's Commands Overhaul PR. Closing.

PR #557, opened 2015-04-08
Unify Gradle setup for API and implementation

Minecrell:
We're currently using a quite huge amount of shared code in the Gradle setup of our projects. Because of this, changes always need to be done in each of them. Because SpongeAPI is referenced in the implementations splitting SpongeAPI's Gradle setup into separate (reusable) files simplifies an unified Gradle setup for SpongeAPI, Sponge and Granite.

The license header changes is only to provide a shared license header across the projects, therefore the project name is included directly from Gradle now instead of having it hard-coded in the header configuration file.


Minecrell:
@progwml6

gabizou:
*coughs* **sends @Mumfrey $5** *coughs*

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -22,7 +22,6 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-
```
simon816: There should be a blank line between the header and package declaration.
Any way to enforce that (e.x in the `HEADER.txt`)?

Minecrell: This is what the Gradle license plugin does. There is currently no easy way around this without some really ugly hacks

Zidane: I don't mind a new line after the header useful or needed at best. We won't be doing that.

Minecrell: The Google Java style requires an empty line between the license header and the package statement, but I don't think it's that important: https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3-source-file-structure

progwml6:
looks good to me @AbrarSyed please review

Minecrell:
@progwml6 Why do we move this to 2.1? This does not depend on the PR in the implementation at all and delaying this will just extremely delay the other Sponge PRs because I always need to wait for the implementation to update for the API changes and then have to fix tons of conflicts which takes a lot of time

PR #561, opened 2015-04-10
Commands refactor

zml2008:
This PR pretty much entirely replaces the commands system.

The new system is a bit more structured -- it contains facilities for argument parsing.

An example of the version of this system included in PEX is at https://github.com/PEXPlugins/PermissionsEx/blob/master/src/main/java/ninja/leaping/permissionsex/command/ParentCommands.java#L37
(there are a few differences in this version, mostly relating to how command names are added, the fact that Text is used for stuff rather than PEX's abstracted wrappers, and to add some more flexibility).

I have included #552 on top of my changes since a lot of command argument stuff has changed, to help @boformer in updating their PR.

Example usage
---------------------
This registers a simple command. It would be put in the plugin's PreInitializationEvent listener.

```java
        this.game.getCommandDispatcher().register(this, CommandSpec.builder()
                .setArguments(seq(string(t("item")), optional(integer(t("count")), 1)))
                .setExecutor(new CommandExecutor() {
                    @Override
                    public CommandResult execute(CommandSource src, CommandContext args) throws CommandException {
                        final Optional<String> item = args.getOne("item");
                        final Optional<Integer> count = args.getOne("count");
                        src.sendMessage(Texts.of("You've gotten ", count.get(), " ", item.get(), count.get() != 1 ? "s" : ""));
                        return CommandResult.builder().affectedItems(count.get()).build();
                    }
                })
                .build(), "testcmd");
```

Want to work with all the provided arguments as a list?
```java
        this.game.getCommandDispatcher().register(this, CommandSpec.builder()
                .setArguments(allOf(string(t("args"))))
                .setExecutor(new CommandExecutor() {
                    @Override
                    public CommandResult execute(CommandSource src, CommandContext args) throws CommandException {
                        final Collection<String> argList = args.getAll("args");
                        src.sendMessage(Texts.of("Arguments provided", argList.toString()));
                        return CommandResult.empty();
                    }
                }).build(), "rawargscmd");
```

Major Components
-------------------------

**CommandSpec:** Immutable object that contains the execution data for the command, including permission, expected arguments, executor, description, and extended description.

**Dispatcher:** Handles the logic of registering, calling, and tabcompleting commands (the logic to CommandSpec's data)

**CommandExecutor:** Single-method interface that takes the sender and parsed arguments and performs the command's action

**CommandElement:** The parent class for any type of command argument. These are often composed to create a complex argument chain. CommandElements are defined in GenericArguments and GameArguments (in o.s.a.util.commands and o.s.a.service.command respectively)


AlphaModder:
This does not address #539. Was it supposed to?

AlphaModder:
@zml2008 zml, I'm trying to implement changes from #539 into this, but I'm having trouble understanding some parts. Is there any time on IRC or other places I could contact you?

src/main/java/org/spongepowered/api/service/command/GameArguments.java
```diff
@@ -0,0 +1,247 @@
...
+
+import java.util.List;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+public class GameArguments {
+    private GameArguments() {}
+
+    private static Text _(String input) {
```
Minecrell: Is this really necessary? This is an extremely bad method name and you could get almost an as short name if you just add an static import for Texts.of()

src/main/java/org/spongepowered/api/util/command/CommandSpec.java
```diff
@@ -0,0 +1,351 @@
...
+     * statement, and the extended description
+     * @param source The source to get the extended description for
+     * @return the extended description
+     */
+    public Text getExtendedDescription(CommandSource source) {
+        Preconditions.checkNotNull(source, "source");
+        TextBuilder builder = Texts.builder();
+        Optional<Text> desc = getDescription();
+        if (desc.isPresent()) {
+            builder.append(desc.get(), Texts.of("\n"));
```
Minecrell: Create the `\n` Text once and save it in a constant so you don't waste objects

src/main/java/org/spongepowered/api/util/command/dispatcher/TranslationPlaceholder.java
```diff
@@ -0,0 +1,70 @@
...
+import org.spongepowered.api.text.Texts;
+import org.spongepowered.api.text.translation.Translation;
+
+/**
+ * This class contains methods intended for use as placeholders until true localization is implemented for SpongeAPI strings.
+ */
+class TranslationPlaceholder {
+    private TranslationPlaceholder() {}
+
+    public static Text _(String untranslated, Object... args) {
```
Minecrell: Using a single underscore character as an identifier for method names is not only making code hard to read if you don't know what this special `_` is doing but it is also discouraged and generates a compiler warning in Java 8: https://travis-ci.org/SpongePowered/SpongeAPI/jobs/58021029#L202-L294

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -305,94 +318,137 @@ public boolean containsMapping(CommandMapping mapping) {
...
+            @Override
+            public boolean apply(String input) {
+                return SimpleDispatcher.this.commands.get(input).getPrimaryAlias().equals(input); // Restrict to primary aliases in usage
+            }
+        });
+
+        for (Iterator<String> it = filteredCommands.iterator(); it.hasNext();) {
+            build.append(Texts.of(it.next()));
+            if (it.hasNext()) {
+                build.append(Texts.of("|"));
```
Minecrell: Can't you use the PIPE_TEXT here you defined somewhere else?

src/main/java/org/spongepowered/api/util/command/TranslationPlaceholder.java
```diff
@@ -0,0 +1,70 @@
...
+package org.spongepowered.api.util.command;
+
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.Texts;
+import org.spongepowered.api.text.translation.Translation;
+
+/**
+ * This class contains methods intended for use as placeholders until true localization is implemented for SpongeAPI strings.
+ */
+class TranslationPlaceholder {
```
ST-DDT: This class looks like a nice utility class thar should be public so everyone can use it.

ST-DDT: This way we dont need it 3 or more times (for every package). And we can improve maintainability.

src/main/java/org/spongepowered/api/util/command/CommandSpec.java
```diff
@@ -0,0 +1,351 @@
...
+
+        private Builder() {}
+
+        /**
+         * Set the permission that will be checked before using this command. May be null.
+         *
+         * @param permission The permission to check
+         * @return this
+         */
+        public Builder setPermission(String permission) {
```
ST-DDT: The method argument should be annotated with `@Nullable`.

src/main/java/org/spongepowered/api/util/command/args/CommandElement.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Attempt to extract a value for this element from the given arguments.
+     *
+     * @param args the arguments
+     * @return The extracted value
+     * @throws ArgumentParseException if unable to extract a value
+     */
+    protected abstract Object parseValue(CommandArgs args) throws ArgumentParseException;
```
ST-DDT: Parsing of the value should contain the CommandSource, so you can use aliases.
Usecases:
Specify target player with "me"
Specify target location with "here"
Specify relative target location with "+10 100 +10"

ST-DDT:
This PR adds some nice features, but IMO this forces commands to much in a specific design/layout.

How can i achieve the tab completion with your suggested system.
Example:
ArenaPvP-options: allowFriendlyFire...
ArenaPvE-options difficultyPerRound....
/myarenas setoptions <Arena> <Option> <Value...>
TabComplete for Option is based on Arena selection
TabComplete for Value is based on Option selection

nikosgram13:
I believe, the best method to register and execute a command is the @sk89q's command-framework from WorldEdit.

https://github.com/sk89q/WorldEdit/tree/master/worldedit-core/src/main/java/com/sk89q/minecraft/util/commands

ST-DDT:
@nikosgram13 The issue with that system is that annotations do not support `Text`s -> no color support.

And i think everyone prefers a different style, thats why i like the current/old system, because you can build everything on top of it.

src/main/java/org/spongepowered/api/service/command/GameArguments.java
zml2008: it's the traditional name used for gettext

Minecrell: See below:
> Using a single underscore character as an identifier for method names is not only making code hard to read if you don't know what this special `_` method is doing but it is also discouraged and generates a compiler warning in Java 8: https://travis-ci.org/SpongePowered/SpongeAPI/jobs/58021029#L202-L294

src/main/java/org/spongepowered/api/util/command/TranslationPlaceholder.java
zml2008: it's not designed to exist long-term -- it's a placeholder until a proper translation system is present in Sponge

src/main/java/org/spongepowered/api/util/command/CommandSpec.java
zml2008: no it shouldn't. you shouldn't be setting the permission to null.

simon816: > Set the permission that will be checked before using this command. May be null.

zml2008:
@ST-DDT tab completion gives each successive argument values from the previous arguments. so you can get stuff from the CommandContext if necessary.

I kinda don't want to expose the CommandSource in the CommandElement -- could run into issues with people doing too much with command sources. All the methods in CommandElement are supposed to have no side effects and be repeatable, meaning anything people do with CommandSources must not change game state at all.

src/main/java/org/spongepowered/api/util/command/CommandSpec.java
zml2008: ok, I'll fix the javadoc then

src/main/java/org/spongepowered/api/service/command/GameArguments.java
zml2008: they can't exactly make a valid method name invalid -- that'd break existing source... but I guess I can use t() or something

ST-DDT: use `Texts.of()` directly with static imports this is only one character more or add String constructors to your exceptions that do the conversion themselfes.

boformer: I agree with @ST-DDT

src/main/java/org/spongepowered/api/util/command/CommandSpec.java
```diff
@@ -0,0 +1,351 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+/**
+ * Specification for how command arguments should be parsed.
+ */
+public final class CommandSpec {
```
boformer: This should be an interface/not final (like ``CommandCallable``).

Some plugins may want to implement a custom ``checkPermission`` or ``getDescription`` method.

src/main/java/org/spongepowered/api/util/command/dispatcher/TranslationPlaceholder.java
```diff
@@ -0,0 +1,70 @@
...
+package org.spongepowered.api.util.command.dispatcher;
+
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.Texts;
+import org.spongepowered.api.text.translation.Translation;
+
+/**
+ * This class contains methods intended for use as placeholders until true localization is implemented for SpongeAPI strings.
+ */
+class TranslationPlaceholder {
```
boformer: I would remove this class. We don't need placeholders, and this is not in the scope of this PR.

Just add ``TODO`` comments in the code where translations are needed.

zml2008: it's not that simple -- translations allow templating, and it'd be a pain to have to rewrite all the messages a bunch. Therefore, placeholder classes. They can be deleted anyway...

boformer:
Please add a detailed PR description. I don't understand how this system works.

**Right now I think that the system we currently have is better than this one. Please don't merge it in its current state!**

Argument Parsing is a nice feature, but I think that it should be optional. There should still be a way to get a simple string array (or a single string) of the arguments.

I don't understand why Sponge needs such a complex command system. It should be minimal and flexible, so that other command systems can be built on top of it. The system proposed here is very static (final classes instead of interfaces). I think it should be a third-party library.

Is there anything that makes it impossible to implement this system on top of the existing command system?

Zidane:
@boformer 

I disagree on the point of providing a String of arguments, there should be standardized parsing in play.

zml2008:
The point of this PR is for every plugin's commands to be handled similarly -- so that every plugin benefits from the same level of usability and users don't have to learn 384854 arguments systems for each plugin and which plugins support quoted arguments and escapes any flags and whatever.
It is still possible to go around the API and get a raw string (we do it for MC/mod commands) -- it's just not recommended because it increases confusion for users.

Zidane:
Absolutely agree with @zml2008

sk89q:
Realistically there will be a lot of people (I included) that will not be providing standardized access to the commands because switching over commands is only going to be needless work that introduces new bugs.

I wouldn't be surprised if the situation was the same as with Bukkit, except that now people have to figure out something complicated to bypass.

src/main/java/org/spongepowered/api/service/command/GameArguments.java
zml2008: ... you do not understand translations

AlphaModder:
@nikosgram13 That is what we used to use, before this.

zml2008:
I see where you're coming from sk -- I'm going to be overriding the commands system in PEX. I've been thinking about making CommandSpec implement CommandCallable -- mostly I've been worried that that doesn't encourage users to do what we want enough.

@AlphaModder we've never used sk's full command API -- what we've used is some super basic stuff that's very low-level. WE uses sk's full commands API, which is imo also a good commands API.

ST-DDT:
@sk89q I agree with you.

> The point of this PR is for every plugin's commands to be handled similarly -- so that every plugin benefits from the same level of usability and users don't have to learn 384854 arguments systems for each plugin and which plugins support quoted arguments and escapes any flags and whatever.

An basic implementation of the `CommandCallable` would do the same job. And if tutorials mostly deal with that system beginners or writers of small plugins will use it anyway. But if expert programmers like sk89q will not use it and instead they break holes in the system everyone else will follow their example. Because the Command system then as a bad image in general although it might be useful in some places..

I for myself i already know that i will break the suggested system or at least try to apply some wrappers all around it, mainly of these 3 reasons:
- I need multilanguage support
- I want to provide a help to the player if he types something wrong.
  - "Could not find Material `STONE_BRIC` did you mean STONE_BRICK?"
- I need named command params for some cases.

I would be happy if you would reintroduce the `CommandCallable` as base interface.

> mostly I've been worried that that doesn't encourage users to do what we want enough.

Rules/Limitations alone have rarely encouraged anyone to do what you want. Ask any parents.
You can only remember the authors via the javadocs and by leading by example.

zml2008:
@ST-DDT except that all those features you talked about are supported in this commands implementation.

I'm not sure where your logic of "if the big plugins stop using the commands API everyone else will too" came from -- even after 5 years of Worldedit, not a lot of plugins use WE's commands API (not like Bukkit offered much anyway).

multilanguage: everything is passed as a Text, which is translatable (I will tweak the Translation interfaces in Text soon to make this simpler)
suggestions for incorrect items: those go in the appropriate CommandElement for the type. In your case, this would be GameArguments.CatalogedTypeCommandElement. (you could even do something like levenschtein difference filtering for tab complete & suggestions)
named command params: you mean like long flags? because those are supported -- see GenericArguments#flags(). And argument lookup is done by name too.

I encourage you to take a look at what the API offers before you start complaining.

ST-DDT:
> I'm not sure where your logic of "if the big plugins stop using the commands API everyone else will too" came from -- even after 5 years of Worldedit, not a lot of plugins use WE's commands API (not like Bukkit offered much anyway).

Because it is not integrated in Bukkit and most tutorials do not use it.
Plugin dependencies are rarely used by beginners.

> multilanguage: everything is passed as a Text, which is translatable (I will tweak the Translation interfaces in Text soon to make this simpler)

throw args.createError(t("Unknown long flag %s specified", args));

-> catch(TextException e) throw new TextException(translate(e.getText())

> suggestions for incorrect items: those go in the appropriate CommandElement for the type. In your case, this would be GameArguments.CatalogedTypeCommandElement. (you could even do something like levenschtein difference filtering for tab complete & suggestions)

That feature is pretty nice, but if i only support a subset (example helmets), i end up implementing it all myself. I never complained about the parsing part. I don't like being forced to define all parsing stuff when declaring the method. Cherry picking your parsers for some arguments is actually a nice benefit of this PR.

> named command params: you mean like long flags? because those are supported -- see GenericArguments#flags(). And argument lookup is done by name too.

Almost, but here is another issue. I want to use the location resolver to not require quotes and `--`.
/cmd foo --target="[world] <x> <y> <z>"
if it just could be
/cmd foo target=[world] <x> <y> <z>
I know I could my own `CommandElement`...

Wait is there anything i have not created wrappers or own implementations for?

zml2008:
I'm going to make Translation support gettext-style serverside translation handling in the near future -- it'll be a small change.

zml2008:
Ok, translations are PR'd in #563, and I've re-added CommandCallable

src/main/java/org/spongepowered/api/util/command/CommandCallable.java
```diff
@@ -24,18 +24,19 @@
...
 import java.util.List;
 
 /**
- * A command that can be executed.
+ * A low-level interface for commands that can be executed. For almost all use cases, higher-level tools should be used instead, like {@link
+ * CommandSpec}
  *
  * <p>Implementations are not required to implement a sane
- * {@link java.lang.Object#equals(Object)} but may choose to do so.</p>
+ * {@link java.lang.Object#equals(Object)} but really should.</p>
```
ST-DDT: What about the `hashCode()` method?

Minecrell:
Although it was proposed above already with WorldEdit's command API, what are the reasons against an annotation-based command API on top of `CommandCallable`? We've an existing annotation-based event system already, having an annotation-based command API would provide a consistent way for plugin authors to listen for events and commands.

The event system could have been designed it the same way using a builder, but we designed it this way because the code required for all the builders can easily get bloated up if you have a lot of events. This is actually even more likely for a lot of commands or complex commands because they have much more properties to configure.

From looking at your example usages, it seems like this could be actually even easier usable with annotations without this whole load of static imports required to create a command in a clean way.

```java
@Command(name = {"testcmd", "test"}, permission = "testplugin.test")
public CommandResult excecuteTestCmd(CommandSource source,
        @Argument(t = "item") String item,
        @Argument(t = "count", def = "0") Integer count) throws CommandException {

    source.sendMessage(Texts.of("You've gotten ", count, " ", item, count != 1 ? "s" : ""));
    return CommandResult.builder().affectedItems(count).build();
}

@Command(name = {"rawargscmd", "rawargs"}, permission = "testplugin.raw")
public CommandResult executeRawArgsCmd(Player player, // Limit command to players
        @Argument(t = "args", all = true) List<String> args) {

    source.sendMessage(Texts.of("Arguments provided", argList.toString()));
    return CommandResult.empty();
}
```

ST-DDT:
@Minecrell This looks also nice, and you can also utilize some parsers from this PR for that.
The only drawback i see with this one, that description/help/usage could explode your annotation.

How is command completion handled in this approach?
If we add this the `@Argument` annotation should also contain a field for a (custom) parser class.

Minecrell:
@ST-DDT The base for most command systems would be the same, mostly the differences are just in the way to define commands. As for (custom) command completion that's a pretty good question, but basic command completion is (as already right now) provided by the argument types (e.g. completing players).

ST-DDT:
@Minecrell You are right. I see 3 different concepts for (custom) command completion here:
- Parsers with provide a autocomplete function, (custom parsers could do the custom stuff)
- A new field `completer` in the `@Command` annotation with a class or methodname that provides the functionality
- Automatically use the `<command>Complete(...)` method if available

boformer:
@Minecrell This is a Bukkit library for annotation-based commands: https://github.com/Ranzdo/MethodCommand/wiki

Your proposal is very similar.

A system like that can easily be built on top of zml's system.

Minecrell:
@boformer If we want to provide additional utilities for creating commands in SpongeAPI itself, I see nothing speaking against an annotation-based approach for the reasons mentioned above. You could as well say, an annotation-based event system could be built on top of an event system with a design similar to this one.

At the end, it may be worth to provide both ways especially with keeping some plugins wanting to provide scripting support in mind.

As for the MethodCommand lib, I have used it long ago when I was starting with Bukkit plugin development and excluding some smaller things it was really nice to work with, although it was stupid you always had to bundle this huge command lib even if you just wanted to make a simple command.

boformer:
@Minecrell I would love to see an annotation command system :+1:

src/main/java/org/spongepowered/api/util/command/CommandCallable.java
Mumfrey: That doesn't need to be specified explicitly because *Java's* contract for `equals` requires overriding `hashCode` if you override it.

kashike:
[Intake](https://github.com/sk89q/Intake) is a wonderful annotation-based command framework, by @sk89q. It is similar to the command framework in WorldEdit, and is quite nice to work with.

ST-DDT:
IMO we should focus on the current PR and should not discuss to much about systems for the next PR.

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -0,0 +1,839 @@
...
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+
+/**
+ * Class containing factory methods for common command elements.
+ */
+public class GenericArguments {
```
simon816: Should this be final?

Mumfrey: It doesn't need to be, it has a private ctor.

ST-DDT: IMO it should be final in that case.
We did the same for a lot of helper/util classes.

src/main/java/org/spongepowered/api/util/command/spec/CommandSpec.java
```diff
@@ -0,0 +1,392 @@
...
+        if (this.permission != null ? !this.permission.equals(that.permission) : that.permission != null) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = this.args.hashCode();
```
Minecrell: You can use `Objects.hashCode(Object...)` from Guava here to simplify this (we use this in a lot of classes already)

src/main/java/org/spongepowered/api/util/command/spec/CommandSpec.java
```diff
@@ -0,0 +1,392 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof CommandSpec)) {
+            return false;
+        }
+
+        CommandSpec that = (CommandSpec) o;
+
+        if (!this.args.equals(that.args)) {
```
Minecrell: Same for Guava here as below, `Objects.equal()` will handle the null check for you:
```java
return Objects.equal(this.args, that.args)
        && Objects.equal(this.argumentParser, that.argumentParser)
        && ...;
```

src/main/java/org/spongepowered/api/util/command/spec/CommandSpec.java
```diff
@@ -0,0 +1,392 @@
...
+        result = 31 * result + (this.description != null ? this.description.hashCode() : 0);
+        result = 31 * result + (this.extendedDescription != null ? this.extendedDescription.hashCode() : 0);
+        result = 31 * result + (this.permission != null ? this.permission.hashCode() : 0);
+        result = 31 * result + this.argumentParser.hashCode();
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "CommandSpec{"
```
Minecrell: Same as above for `toString()`: https://code.google.com/p/guava-libraries/wiki/CommonObjectUtilitiesExplained#toString

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -0,0 +1,798 @@
...
+            final String input = args.next();
+            try {
+                return Integer.parseInt(input);
+            } catch (NumberFormatException ex) {
+                throw args.createError(t("Expected an integer, but input '%s' was not", input));
+            }
+        }
+    }
+
+    private static final Map<String, Boolean> BOOLEAN_CHOICES = ImmutableMap.<String, Boolean>builder()
```
Minecrell: I guess `1` (for true) and `0` for false would be also good ideas to add

src/main/java/org/spongepowered/api/service/command/GameArguments.java
```diff
@@ -0,0 +1,282 @@
...
+
+    private static class Vector3dCommandElement extends CommandElement {
+
+        protected Vector3dCommandElement(Text key) {
+            super(key);
+        }
+
+        @Override
+        protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            return null;
```
Minecrell: Do you plan to implement this before the merge? If not, add a TODO comment for the missing stuff

src/main/java/org/spongepowered/api/util/command/spec/CommandSpec.java
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    /**
+     * Get a short, one-line description used with this command if any is present
+     *
+     * @return the short description.
+     */
+    @Override
+    public Optional<Text> getShortDescription(CommandSource source) {
+        return Optional.fromNullable(this.description);
```
Minecrell: You can just change the field to an `Optional` directly, no need to wrap this every time

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
Minecrell: Should be `final` for consistency with other classes like this one

Minecrell:
Can you also fix these minor formatting problems so we can keep master clean:
```
service/command/GameArguments.java:242: warning - Tag @link: reference not found: T
util/command/args/GenericArguments.java:643: warning - Tag @link: reference not found: T
```
```
util/command/args/CommandFlagsTest.java:28: warning: Using the '.*' form of import should be avoided - org.spongepowered.api.util.command.args.GenericArguments.*.
util/command/args/GenericArgumentsTest.java:29: warning: Using the '.*' form of import should be avoided - org.spongepowered.api.util.command.args.GenericArguments.*.
```
(Optimizing the imports using your IDE should fix these)

simon816:
There's a few places where you have double spaces where there shouldn't be:
[ArgumentParseException:89](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-59b2adcaec3b60e1caa6ac1bdf10dc9fR89)
[CommandElement:76](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-066448d553966b8de836bfc87ca20bb6R76)
[CommandFlags:259](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-94175cb0b523ffd500e763411ba49c18R259)
[GenericArguments:127](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R127)
[GenericArguments:158](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R158)
[GenericArguments:216](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R216)
[GenericArguments:284](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R284)
[GenericArguments:302](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R302)
[GenericArguments:455](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R455)
[GenericArguments:469](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R469)
[GenericArguments:507](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R507)
[GenericArguments:666](https://github.com/SpongePowered/SpongeAPI/pull/561/files#diff-bd123d476b4046fba0c04ed1536a5725R666)

PR #562, opened 2015-04-11
Dynamic Commands 3

AlphaModder:
Continuation of #539 for zml's new API. I'll write a description in more detail when it isn't 2 AM...


src/main/java/org/spongepowered/api/service/command/AliasContext.java
```diff
@@ -0,0 +1,11 @@
+package org.spongepowered.api.service.command;
+
+import com.google.common.base.Optional;
+import org.spongepowered.api.util.command.CommandSource;
+
+
```
simon816: Missing javadocs

src/main/java/org/spongepowered/api/util/command/dispatcher/TranslationPlaceholder.java
```diff
@@ -31,7 +31,7 @@
 /**
  * This class contains methods intended for use as placeholders until true localization is implemented for SpongeAPI strings.
  */
-class TranslationPlaceholder {
+public class TranslationPlaceholder {
```
zml2008: please don't -- just copy the class if you need it in a different package. This is a temporary thing, needs to be replaced by a real translation system eventually.

AlphaModder: @zml2008 ok.

AlphaModder:
No reason for this anymore.

PR #577, opened 2015-04-18
Revamp selectors

### LINKED ISSUE #516: Update Selectors for Scoreboard merge
Selectors have a `TODO` on them to update with Scoreboard API's merge. This needs to be done once #299 is complete.

kenzierocks:
Required revamp to make selectors implementable without lots of hacks. Requirement of https://github.com/SpongePowered/SpongeCommon/pull/14.

#### More details
There was a problem with having `Vector3` as an `ArgumentType`. Because it was an `ArgumentType`  but it's `x/y/z` parts were also `ArgumentTypes` it was unclear which types would be put into the internal `ArgumentType->Argument` map. This resolves the ambiguity by adding an `ArgumentHolder` interface, which is extended by both `Vector3` and `Limit`. This interface is also extended by `ArgumentType`, as any `ArgumentType` "holds" itself. This also allows for `Limit` to have a more strict upper bound on its maximum and minimum to allow a `Limit<Vector3>`.

Additionally, `ArgumentType` was generic and not a part of the `CatalogType` system. This still holds, so `ArgumentTypes` now allows retrieval of an `ArgumentType` via `valueOf(String)` and the (current) set via `values()`.

#### Summary
- `Vector3` and `Limit` are now `ArgumentHolders`
- `ArgumentType` also is a holder, with itself as the only element.
- `ArgumentTypes` has `valueOf` and `values`
- An `ArgumentHolder` can be passed to create a `Set` of arguments (although I think the implementation currently doesn't support using `Limit` as the holder).
- Scoreboard is in Selectors (closes #516)
- Probably a few other things

### DIFF
--- a/src/main/java/org/spongepowered/api/text/selector/ArgumentHolder.java
+++ b/src/main/java/org/spongepowered/api/text/selector/ArgumentHolder.java
@@ -1,0 +1,117 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.selector;
+
+import java.util.Set;
+
+/**
+ * Represents a holder of {@link ArgumentHolder}s. This interface is extended
+ * by both Vector3 and Limit. It is also extended by {@link ArgumentType}, as
+ * any {@link ArgumentType} "holds" itself. This also allows for Limit to
+ * have  more strict upper bound on its maximum and minimum to allow a
+ * {@code Limit<Vector3>} without allowing {@code Limit<Object>}.
+ *
+ * @param <T> The type for the value of this argument holder
+ * @see ArgumentType
+ * @see ArgumentTypes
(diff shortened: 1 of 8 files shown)


Minecrell:
@kenzierocks Can you add a short description how this helps with implementing and how this affects API usage (if it does)?

kenzierocks:
Done.

src/main/java/org/spongepowered/api/text/selector/ArgumentHolder.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Returns the size of the Set from {@link #getTypes()}.
+     * 
+     * @return The size of the Set from {@link #getTypes()}
+     */
+    int getCount();
+
+    /**
+     * Returns a set containing all the {@link ArgumentHolder}s this {@link ArgumentHolder} holds.
+     * @return
```
simon816: Missing description on this `@return` tag

kenzierocks: Patched.

src/main/java/org/spongepowered/api/text/selector/Arguments.java
```diff
@@ -34,6 +36,8 @@ private Arguments() {
 
     /**
      * Creates a new {@link Argument} using the specified type and value.
+     * 
+     * <p>If the type is invertible, the Argument will not be inverted.</p>
```
Minecrell: `the {@link Argument}` or just `the argument`

kenzierocks:
Updated to master.

src/main/java/org/spongepowered/api/text/selector/Arguments.java
kenzierocks: Fixed.

src/main/java/org/spongepowered/api/text/selector/ArgumentHolder.java
```diff
@@ -0,0 +1,111 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.selector;
+
+import java.util.Set;
+
+/**
+ * Represents a holder of {@link ArgumentHolder}s.
```
zml2008: Do you mean ArgumentType here?

kenzierocks: No, actually. It's a little strange but it works the best.

zml2008: ok, might be good to expand on this javadoc a bit to help people understand how this fits in -- you could probably use more or less your PR description here

kenzierocks: Ok, tell me what you think.

zml2008:
Ok, seems good. Could you squash the commits?

kenzierocks:
Done.

kenzierocks:
Added a method to Selector to try and just use a CommandSource to resolve, unsure what happens when it fails. If this is wanted I'll flesh out the javadocs.

kenzierocks:
Not ready for pull, overlooked gap in getting inverted arguments reliably from the selector.

kenzierocks:
Added `has(ArgumentType)` and `isInverted(ArgumentType.Invertible)` for convenience, patched above gap.

Zidane:
@kenzierocks 

Status update?

kenzierocks:
Still implementing Extent support, I think this is almost ready, but in all honesty I would wait for the implementation part to be complete and see if I need any more changes.

kenzierocks:
Added the ability to select in multiple extents. Should I note the vanilla behavior of not using the given `Extent`s unless there are certain `Argument`s present?

kenzierocks:
Added javadocs, and also added the ability to override vanilla behavior.

Issue #528, opened 2015-04-01
Access offline player data


boformer:
Right now I'm working on a server mail plugin. Of course, players can also send mails to offline players.

I want to check if the player name entered by the player is a registered Minecraft profile name and if the player has played before on the server.

These kind of methods are important for many plugins (e.g. /lastonline <offline-player> or /ban <offline-player>).

I searched the SpongeAPI code for a method to get the ``GameProfile``/``User`` of an offline player, but did not find one.

Proposed methods:
* ``Optional<User> getUser(UUID uuid, boolean playedBefore)``
* ``Optional<User> getUser(String name, boolean playedBefore)``
* ``List<User> getUsers()`` --> Returns list of all users who played before.

The ``getUser`` method should first check if the user exists in the world save files. If it does not exist (user has not played before) and ``playedBefore`` is set to false, the server should contact the Mojang servers to check if the username/UUID exists.

MattBDev:
I do believe this is needed but `getUser` with a string parameter (I am assuming it is accepting a username) wouldn't be able to check the world save files for the player based on their name since mojang keeps no record of their last joined name. Bukkit added their own nbt tag to each player data file but I am not sure if Sponge would be willing to do this.
I also feel that there should be methods without `playedBefore`.

modwizcode:
I disagree with having `playedBefore` anywhere, that's a very strange thing to have. We should maintain a cache for getUser(String username) which will be update by Mojang if no cache and no player online and also updated when the player joins the server. People will be encouraged to use UUID's but we shouldn't make plugins have their own cache.

boformer:
@modwizcode I agree with you that a ``username <-> UUID`` cache is the way to go.

But how would this cache work when a player changes the username?

Example: 
```
modwizcode plays on server.
server saves username in cache.
modwizcode logs out
modwizcode changes name to FakeNotch
Player calls /lastseen modwizcode, which calls getUser("modwizcode")
Player calls /lastseen FakeNotch  , which calls ``getUser("FakeNotch")``
```
What will the method return for ``"modwizcode"``, what for ``"FakeNotch"``?

Maybe we should add 2 different methods for name lookups:
* ``getUserFromCache(String name)`` --> Only searches cached player data 
(``Optional.absent()`` for ``"FakeNotch"``)
* ``getUserFromMojang(String name)`` --> Only contacts Mojang's servers

We can also combine these methods with a boolean parameter.

To make it perfect, there should be a way to get both usernames:

[...]

PR #580, opened 2015-04-19
Add access to GameProfile cache, a GameProfile resolver and Server.getUser(UUID)

boformer:
This PR adds a service interface that can be used to lookup player UUIDs by name and player names by UUIDS. The service contacts the Mojang auth servers if no cached profile was found in the local cache.
It also adds a ``UserStorage`` interface to get, create or delete a ``User`` object.

This resolves https://github.com/SpongePowered/SpongeAPI/issues/528

Access to cached offline user data is important to:

* complete or validate names of offline players
* lookup the name of an offline player by an UUID stored by a plugin
* ban offline players

Things to discuss:

**Access to the [Name Changes API](http://wiki.vg/Mojang_API#UUID_-.3E_Name_history)?** 
Mojang provides an API which lists up all names and change dates of a player.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
+++ b/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
@@ -1,0 +1,136 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.profile;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import org.spongepowered.api.GameProfile;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * A service which contacts the authentication servers to get a
+ * {@link GameProfile} by a given UUID or name.
+ *
+ * <p>The service may cache the data of a request for faster lookups. Note that
+ * the cached data may not always be up to date.</p>
+ *
+ * <p>The returned {@link ListenableFuture} throws an {@link ExecutionException}
+ * caused by a {@link ProfileNotFoundException} if the profile does not exist or
+ * an {@link IOException} if a network error occurred.</p>
+ */
+public interface GameProfileResolver {
+
+    /**
+     * Looks up a {@link GameProfile} by its unique ID.
+     *
+     * <p>This method checks the local profile cache before contacting the
+     * profile servers. Use {@link #get(UUID, boolean)} to disable the cache
+     * lookup.</p>
+     *
+     * @param uniqueId The unique ID
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(UUID uniqueId);
+
+    /**
+     * Looks up a {@link GameProfile} by its unique ID.
+     *
+     * @param uniqueId The unique ID
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(UUID uniqueId, boolean useCache);
+
+    /**
+     * Looks up a {@link GameProfile} by its user name (case-insensitive).
+     *
+     * <p>This method checks the local profile cache before contacting the
+     * profile servers. Use {@link #get(String, boolean)} to disable the cache
+     * lookup.</p>
+     *
+     * @param name The username
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(String name);
+
+    /**
+     * Looks up a {@link GameProfile} by its user name (case-insensitive).
+     *
+     * @param name The username
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(String name, boolean useCache);
+
+    /**
+     * Gets a collection of {@link GameProfile}s by their user names
+     * (case-insensitive).
+     *
+     * @param names The user names
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<Collection<GameProfile>> getAllByName(Iterable<String> names, boolean useCache);
+
+    /**
+     * Gets a collection of {@link GameProfile}s by their unique IDs.
+     *
+     * @param uniqueIds The UUIDs
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<Collection<GameProfile>> getAllById(Iterable<UUID> uniqueIds, boolean useCache);
+
+    /**
+     * Gets a collection of all cached {@link GameProfile}s.
+     *
+     * @return A {@link Collection} of {@link GameProfile}s
+     */
+    Collection<GameProfile> getAllCached();
+
+    /**
+     * Returns a collection of matching cached {@link GameProfile}s whose last
+     * known user names start with the given string (case-insensitive).
+     *
+     * <p>This collection may also contain profiles of players who never played
+     * on the server!</p>
+     *
+     * <p>Use
+     * {@link org.spongepowered.api.service.user.UserStorage#match(String)} for
+     * a collection that only contains {@link GameProfile}s with attached
+     * {@link org.spongepowered.api.entity.player.User} data.</p>
+     *
+     * <p>This method only searches the local cache, so the data may not be up
+     * to date.</p>
+     *
+     * @param lastKnownName The user name
+     * @return The result of the request
+     */
+    Collection<GameProfile> match(String lastKnownName);
+
+}
--- a/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
+++ b/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
@@ -111,7 +111,7 @@ public interface GameProfileResolver {
      *
      * @return A {@link Collection} of {@link GameProfile}s
      */
-    Collection<GameProfile> getAllCached();
+    Collection<GameProfile> getAllCachedProfiles();
 
     /**
      * Returns a collection of matching cached {@link GameProfile}s whose last
--- a/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
+++ b/src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
@@ -111,7 +111,7 @@ public interface GameProfileResolver {
      *
      * @return A {@link Collection} of {@link GameProfile}s
      */
-    Collection<GameProfile> getAllCachedProfiles();
+    Collection<GameProfile> getCachedProfiles();
 
     /**
      * Returns a collection of matching cached {@link GameProfile}s whose last
--- a/src/main/java/org/spongepowered/api/service/profile/ProfileNotFoundException.java
+++ b/src/main/java/org/spongepowered/api/service/profile/ProfileNotFoundException.java
@@ -22,22 +22,51 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api;
-
-import org.spongepowered.api.util.Identifiable;
+package org.spongepowered.api.service.profile;
 
 /**
- * Represents a name and an associated unique identifier.
- *
- * <p>Use the {@link org.spongepowered.api.service.user.UserStorage} service to
- * obtain the stored data associated with a profile.</p>
+ * Thrown by a profile lookup service if a profile does not exist.
  */
-public interface GameProfile extends Identifiable {
+public class ProfileNotFoundException extends Exception {
+
+    private static final long serialVersionUID = -6165011303382043479L;
 
     /**
-     * Gets the player's last known username.
-     *
-     * @return The player's last known username
+     * Constructs a new {@link ProfileNotFoundException}.
      */
-    String getName();
+    public ProfileNotFoundException() {
+        super();
+    }
+
+    /**
+     * Constructs a new {@link ProfileNotFoundException} with the specified
+     * message.
+     *
+     * @param message The exception message
+     */
+    public ProfileNotFoundException(String message) {
+        super(message);
+    }
+
+    /**
+     * Constructs a new {@link ProfileNotFoundException} with the specified
+     * cause and a null message.
+     *
+     * @param cause The cause of this exception
+     */
+    public ProfileNotFoundException(Throwable cause) {
+        super(cause);
+    }
+
+    /**
+     * Constructs a new {@link ProfileNotFoundException} with the specified
+     * message and cause.
+     *
+     * @param message The exception message
+     * @param cause The cause of this exception
+     */
+    public ProfileNotFoundException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
 }
--- a/src/main/java/org/spongepowered/api/service/user/UserStorage.java
+++ b/src/main/java/org/spongepowered/api/service/user/UserStorage.java
@@ -1,0 +1,127 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.user;
+
+import com.google.common.base.Optional;
+import org.spongepowered.api.GameProfile;
+import org.spongepowered.api.entity.player.Player;
+import org.spongepowered.api.entity.player.User;
+import org.spongepowered.api.service.profile.GameProfileResolver;
+
+import java.util.Collection;
+import java.util.UUID;
+
+/**
+ * Stores the persistent {@link User} data of a {@link Player}.
+ */
+public interface UserStorage {
+
+    /**
+     * Gets the data of a {@link User} by their unique id.
+     *
+     * @param uniqueId The UUID of the user
+     * @return {@link User} or Optional.absent() if not found
+     */
+    Optional<User> get(UUID uniqueId);
+
+    /**
+     * Gets the data of a {@link User} by their last known user name
+     * (case-insensitive).
+     *
+     * <p>To get the current name of a player, use the
+     * {@link GameProfileResolver} service.</p>
+     *
+     * @param lastKnownName The user name
+     * @return {@link User} or Optional.absent() if not found
+     */
+    Optional<User> get(String lastKnownName);
+
+    /**
+     * Gets the data of a {@link User} by their {@link GameProfile}.
+     *
+     * @param profile The profile
+     * @return {@link User} or Optional.absent() if not found
+     */
+    Optional<User> get(GameProfile profile);
+
+    /**
+     * Gets or creates a persistent {@link User} associated with the given
+     * {@link GameProfile}.
+     *
+     * <p>To obtain a {@link GameProfile}, use the {@link GameProfileResolver}.
+     * </p>
+     *
+     * @param profile The profile
+     * @return The user object
+     */
+    User getOrCreate(GameProfile profile);
+
+    /**
+     * Gets the collection of all {@link GameProfile}s with stored {@link User}
+     * data.
+     *
+     * <p>Note that this method is resource-intensive depending on the amount of
+     * stored data.</p>
+     *
+     * <p>Use {@link #get(GameProfile)} to get the {@link User} data
+     * corresponding to a {@link GameProfile}.</p>
+     *
+     * @return A {@link Collection} of {@link GameProfile}s
+     */
+    Collection<GameProfile> getAll();
+
+    /**
+     * Deletes the data associated with a {@link User}.
+     *
+     * <p>This may not work if the user is logged in.</p>
+     *
+     * @param profile The profile of the user to delete
+     * @return true if the deletion was successful
+     */
+    boolean delete(GameProfile profile);
+
+    /**
+     * Deletes the data associated with a {@link User}.
+     *
+     * <p>This may not work if the user is logged in.</p>
+     *
+     * @param user The user to delete
+     * @return true if the deletion was successful
+     */
+    boolean delete(User user);
+
+    /**
+     * Returns a collection of matching {@link GameProfile}s with stored
+     * {@link User} data whose last known user names start with the given string
+     * (case-insensitive).
+     *
+     * <p>Use {@link #get(GameProfile)} to get the {@link User} data
+     * corresponding to a {@link GameProfile}.</p>
+     *
+     * @param lastKnownName The user name
+     * @return The result of the request
+     */
+    Collection<GameProfile> match(String lastKnownName);
+}


Minecrell:
Cache isn't really a good name for getting `User` instances. The difference between `GameProfile` and `User` is not only persistence or being cached, they represent two completely different use cases:

- `GameProfile`: The identification of a player (UUID + name), as well as optionally a few properties from Mojang's server (like skin and cape). This is generally not persistent, but Minecraft is maintaining a cache for the game profiles for faster profile lookup (`usercache.json`). This includes not only players that have played on the server, but also name->UUID lookups caused through commands.

- `User`: Persistent data of a player, this contains his location in worlds, as well as inventory contents and all the other persistent data of a `Player` - even though not everything is accessible through SpongeAPI. For that to exist, the player has to join the server once. This gets saved together with the world data, in NBT.

A cache is typically used for performance, where entries (can) expire and where it doesn't matter much if certain entries get lost, because they can be queried again. While you could see `User` as cached data being loaded from the disk, a "cache" would be more appropriate for `GameProfile`s, which will be eventually queried from Mojang servers.

Loading `User` data should never trigger any queries to Mojang's server, but rather rely only on the saved persistent user data. If you would want to get a player's data with his newer name (e.g. because he hasn't joined the server since he changed yet) you would need to query name->UUID using a `GameProfile` first and then load the `User` using the returned UUID.

boformer:
@Minecrell What I primarily want is to access the local username<-->uuid storage (only for players who played on the server before).

What do you think of this proposal?

* UserStorage (maybe part of ``Server``)
  * get ``User`` by UUID
* GameProfileCache
  * get stored ``GameProfile`` by UUID
. * get stored ``GameProfile`` by name
. * get list of all profiles
. * list of matching profiles starting with string
* GameProfileResolver
  * get ``GameProfile`` from auth server by UUID
. * get ``GameProfile`` from auth server by name

Minecrell:
@boformer Basically yes, although I think `getUser` should be actually at the same place as `getPlayer` so it would be probably a good idea to move it.

boformer:
I added a new commit which adds the profile cache and resolver and a few new methods in the server interface.

src/main/java/org/spongepowered/api/service/userdata/UserDataCache.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return A list of cached users
+     */
+    List<User> match(String name);
+
+    /**
+     * Returns a list of all cached users.
+     *
+     * @return A list of cached users
+     */
+    List<User> getAll();
```
simon816: I would imagine this method could be expensive, especially for a lot of previous player data. Perhaps it could be `List<UUID>` instead?

src/main/java/org/spongepowered/api/service/profile/GameProfileCache.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Returns a collection of all cached profiles.
+     *
+     * <p>This collection may include profiles of players who never joined the
+     * server.</p>
+     *
+     * @return A collection of {@link GameProfile}s
+     */
+    ListenableFuture<Collection<GameProfile>> getAll();
```
Minecrell: Why is this a future? The cached game profiles should be available without additional query time

src/main/java/org/spongepowered/api/service/profile/GameProfileSource.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public interface GameProfileSource {
+
+    /**
+     * Looks up a {@link GameProfile} by its unique ID.
+     *
+     * @param uniqueId The unique ID
+     * @return The result of the request
+     */
+    ListenableFuture<Optional<GameProfile>> get(UUID uniqueId);
```
Minecrell: I think we can generally say, fetch the game profile or fail (with an exception), so the Optional is not needed

src/main/java/org/spongepowered/api/service/profile/GameProfileCache.java
boformer: It's based on this suggestion: https://github.com/SpongePowered/SpongeAPI/issues/528#issuecomment-94164268

In theory, the cache could also be a database, so both service interfaces extend ``GameProfileSource``, which also utilizes ``Future``s.

boformer:
Some more things to discuss:

**Access to the [Name Changes API](http://wiki.vg/Mojang_API#UUID_-.3E_Name_history)?** 
Mojang provides an API which lists up all names and change dates of a player.

**Server.deleteUser(..) method?** 

**A chained cache --> resolver interface?** 
First check the cache, contact Mojang servers if no data is cached or is expired?

ST-DDT:
> get list of all profiles (boolean playedBefore to exclude lookups)

What kind of lookup? I guess you are not refering to get all available GameProfiles from Mojang.

boformer:
No, from the local profile cache.
The cache may also store profiles of players who never played on the server (a plugin just queried the mojang server, and the result is stored in cache).

For the command autocomplete feature, these profiles should not be included.

The completer could also check if a ``User`` object exists for a profile.

To workflow to get a user by name is just becoming very complicated:
* check if profile is in cache
  * no? Query mojang server 
    * no? --> Profile does not exist
* get User object for profile
  * absent? --> player never played on server
* check if user hasPlayedBefore()
...

boformer:
I thought about this again:

* It would make sense to merge the resolver and the cache. Caching should be handled internally (expiration date)
* We need a function that returns a collection of all Users with a name that starts with a string.

boformer:
* The function should not cause a http request to a Mojang server, but use the last known username.

:arrow_right:  We have to store the last known username of a ``User``.

MattBDev:
Any updates on the status of this PR?

boformer:
@MattBDev I think it is fleshed out now. It basically adds 2 new service interfaces:

``UserStorage`` - To get a ``User`` by UUID or last known name
``GameProfileResolver`` - Search the profile cache or contact Mojang servers to get the current name by UUID (or vice versa)

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * <p>The service may cache the data of a request for faster lookups. Note that
+ * the cached data may not always be up to date.</p>
+ *
+ * <p>The returned {@code ListenableFuture} throws an {@link ExecutionException}
+ * caused by a {@link ProfileNotFoundException} if the profile does not exist or
+ * an {@link IOException} if a network error occured.</p>
+ */
+public interface GameProfileResolver
+{
```
Minecrell: Code formatting is wrong here

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -25,8 +25,10 @@
 package org.spongepowered.api;
 
 import com.google.common.base.Optional;
+
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.player.Player;
+import org.spongepowered.api.entity.player.User;
```
simon816: Unused import

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Collection;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
+import org.spongepowered.api.GameProfile;
+
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * A service which contacts the authentification servers to get a
```
simon816: Spelling. "authentification" => "authentication"

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param profile The profile
+     * @return The user object
+     */
+    User getOrCreate(GameProfile profile);
+
+    /**
+     * Deletes the data associated with a {@link User}.
+     *
+     * @param profile The profile
+     * @return The user object
```
simon816: this is a `void` method, no `@return`. I suspect this was copy + paste from above method.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @return A {@link Collection} of users.
+     */
+    Collection<User> getAll();
+
+    /**
+     * Returns a collection of matching {@code User}s whose last known user
+     * names start with the given string (case-insensitive).
+     *
+     * @param name
```
simon816: `@param lastKnownName`, also need description of parameter

simon816:
Would this PR be an appropriate place to move the method [`createGameProfile`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/GameRegistry.java#L366) out of `GameRegistry` as I don't think it belongs there?

MattBDev:
@simon816 I think that method may have a use that these new classes won't work (I can't think of any myself though) for although I could be wrong.

boformer:
@simon816 The method in GameRegistry has it's own use cases:

* Import ``User`` data from another server (e.g. a Bukkit server)
* Add fake profiles to a ``StatusPingEvent``

src/main/java/org/spongepowered/api/Server.java
me4502: All of the modifications in this file are unrelated to this PR.

MattBDev: Changes on this file should be removed before pulling.

MattBDev:
I think if all the changes on the Server.java file are removed, this can be pulled.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * A service which contacts the authentication servers to get a
+ * {@link GameProfile} by a given UUID or name.
+ *
+ * <p>The service may cache the data of a request for faster lookups. Note that
+ * the cached data may not always be up to date.</p>
+ *
+ * <p>The returned {@code ListenableFuture} throws an {@link ExecutionException}
```
simon816: `{@link ListenableFuture}`

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Gets the collection of all stored {@link User}s.
+     *
+     * @return A {@link Collection} of users.
+     */
+    Collection<User> getAll();
+
+    /**
+     * Returns a collection of matching {@code User}s whose last known user
```
simon816: `{@link User}`

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,103 @@
...
+     * Deletes the data associated with a {@link User}.
+     *
+     * @param profile The profile
+     */
+    void delete(GameProfile profile);
+
+    /**
+     * Gets the collection of all stored {@link User}s.
+     *
+     * @return A {@link Collection} of users.
```
simon816: Remove the period

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param uniqueId The UUID of the user.
+     * @return {@link User} or Optional.absent() if not found
+     */
+    Optional<User> get(String lastKnownName);
+
+    /**
+     * Gets the data of a {@link User} by their profile.
+     *
+     * @param uniqueId The UUID of the user.
```
simon816: Remove the period

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,103 @@
...
+    Optional<User> get(UUID uniqueId);
+
+    /**
+     * Gets the data of a {@link User} by their last known user name
+     * (case-insensitive).
+     *
+     * <p>To get the current name of a player, use the
+     * {@link GameProfileResolver} service.</p>
+     *
+     * @param uniqueId The UUID of the user.
```
simon816: Remove the period

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Stores the persistent {@link User} data of a {@link Player}.
+ */
+public interface UserStorage {
+
+    /**
+     * Gets the data of a {@link User} by their unique id.
+     *
+     * @param uniqueId The UUID of the user.
```
simon816: Remove the period

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param name The username
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(String name, boolean useCache);
+
+    /**
+     * Gets a collection of {@link GameProfile}s by their user names
+     * (case-insensitive).
+     *
+     * @param names The user names
```
simon816: Need `@param useCache`

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param uniqueIds The UUIDs
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<Collection<GameProfile>> getAllById(Iterable<UUID> uniqueIds, boolean useCache);
+
+    /**
+     * Looks up a {@link GameProfile} by its user name (case-insensitive).
+     *
+     * @param name The username
```
simon816: Need `@param useCache`

MattBDev:
@boformer are you able to update the PR with the changes @simon816 suggested

boformer:
@MattBDev Done.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,85 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.profile;
+
+import java.io.IOException;
```
simon816: Wrong import order, `java` imports should be below others. (See [guidelines](https://docs.spongepowered.org/en/contributing/guidelines.html))

boformer:
@simon816 I finally found out how to control this with eclipse

MattBDev:
:exclamation: *waits impatiently for sponge team to review this*

RobertHerhold:
@MattBDev 
Just chill out for a minute. These guys have full-time jobs, and they are *very* busy at times. They will review this when they get to it.

Zidane:
@SpongePowered/developers 

Final call before this is merged. I need some feedback.

me4502:
looks good to me

Minecrell:
How would you resolve a `GameProfile` **only** from the cache (if available), without querying Mojang's servers? Sometimes it might be not that important so you just want it if available.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,102 @@
...
+     * @return The user object
+     */
+    User getOrCreate(GameProfile profile);
+
+    /**
+     * Deletes the data associated with a {@link User}.
+     *
+     * @param profile The profile
+     */
+    void delete(GameProfile profile);
```
Minecrell: What would happen here if the user is currently online?

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,102 @@
...
+    Collection<User> getAll();
+
+    /**
+     * Returns a collection of matching {@link User}s whose last known user
+     * names start with the given string (case-insensitive).
+     *
+     * @param lastKnownName The user name
+     * @return The result of the request
+     */
+    Collection<User> match(String lastKnownName);
```
Minecrell: Couldn't we provide the same also for `GameProfile`s? There will be most likely more `GameProfile`s cached than `User`s, so you could provide more profiles than users (e.g. when used for tab completion).

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param profile The profile
+     */
+    void delete(GameProfile profile);
+
+    /**
+     * Gets the collection of all stored {@link User}s.
+     *
+     * @return A {@link Collection} of users
+     */
+    Collection<User> getAll();
```
Minecrell: Considering that this can be quite inefficient for bigger uses (which may have 5,000+ players persisted) this should have _at least_ some note that this can be quite inefficient depending on the player count on the server.

simon816: I think it should return a collection of UUIDs as it is much cheaper to load many UUID objects.

me4502: I feel GameProfile would be more useful.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
MattBDev: I think it would make sense to delete inventory and other data if they are online

simon816: It would cause sync issues with the client unless everything is updated on the client. Location, inventory, hunger, health etc. Maybe this should fail (and return a boolean false) if the user is online.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: The problem is that cached gameprofiles expire (at least in vanilla). Also, not all cached gameprofiles are players who played on the server.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: Or throw an exception?

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: Ok, I will change it to GameProfile.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: @Minecrell Do you think this method should also return GameProfiles for performance reasons? At least that would be sufficient for tab completion.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
MattBDev: An exception would work.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
Minecrell: If you change the other method this should also return `GameProfile`, yes.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
ryantheleach: The API has largely avoided checked exceptions where possible, why add one here, now?

src/main/java/org/spongepowered/api/service/user/UserStorage.java
ryantheleach: If it's that likely to be inefficient, maybe make some sort of a lazy collection, or provide search methods or some way to page it?

Minecrell:
Can you remove [`GameProfile.getUser()`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/GameProfile.java#L43-L48)? With the additions in here these methods should be preferred over the current one.

boformer:
@minecrell yes, I will. Also thought about that today.

boformer:
I have removed the getUser method from GameProfile and changed the return types of the resource-intensive methods to Collection<GameProfile>. I also changed the return type of the delete method to boolean to indicate the success of the deletion.

I think this is ready.

Minecrell:
I'll repeat this question from above:
> How would you resolve a `GameProfile` **only** from the cache (if available), without querying Mojang's servers? Sometimes it might be not that important so you just want it if available.

boformer:
@Minecrell If you want to lookup a player who played on the server before, you can use getUser(UUID/Name).getProfile(), though that's not very elegant.

Do you think it is necessary to have direct access to the profile cache (which also contains players who did not play on the server)?

I could add a third service for that: ``GameProfileCache``, similar to ``GameProfileResolver`` (maybe without Futures).

Or replace the ``boolean useCache`` parameter with a ``ResolveMethod`` enum parameter: CACHE, REMOTE, CACHE_REMOTE.

What do you prefer?

Minecrell:
I would basically like a way to access the cache _without_ any delay (-> without a future). Sometimes you don't want (or can't) halt the execution of your code to wait for the exectution of a future. IMO there should also be a way to access only the cached profiles with direct access.

I know that you made it that way because of https://github.com/SpongePowered/SpongeAPI/pull/580#discussion_r28693330, however there will be most likely a local memory cache anyway simply because that's much faster than querying a remote server everytime. And if not, this cache could simply never return anything.

Minecrell:
@boformer Actually never mind, I realized there is really not a big use case for it.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,84 @@
...
+public interface GameProfileResolver {
+
+    /**
+     * Looks up a {@link GameProfile} by its unique ID.
+     *
+     * @param uniqueId The unique ID
+     * @param useCache true to perform a cache lookup first
+     * @return The result of the request
+     */
+    ListenableFuture<GameProfile> get(UUID uniqueId, boolean useCache);
```
Minecrell: What do you think about separating this in like `get()` and `query()` instead of the parameter,  it makes reading the code more clear: `resolver.query(uuid)` vs `resolver.get(uuid, false)` (you always have to look what exactly was this boolean again)

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Deletes the data associated with a {@link User}.
+     *
+     * <p>This might not work if the user is logged in.</p>
+     *
+     * @param profile The profile
+     * @return true if the deletion was successful
+     */
+    boolean delete(GameProfile profile);
```
Minecrell: Couldn't this take a user? This would ensure the given parameter is really a user that exists.

src/main/java/org/spongepowered/api/service/user/UserStorage.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * <p>Note that this method is resource-intensive depending on the amount of
+     * stored data.</p>
+     *
+     * <p>Use {@link #get(GameProfile)} to get the {@link User} data
+     * corresponding to a {@link GameProfile}.</p>
+     *
+     * @return A {@link Collection} of {@link GameProfile}s
+     */
+    Collection<GameProfile> getAll();
```
Minecrell: While these both methods return a `GameProfile` already, I think they would also additionally make sense with the data from the cache. For example for a command adding player heads to the inventory you don't really care if the user has played on the server already, you want too provide as many options in the tab complete as possible.

Minecrell:
You have a checkstyle warnings in here https://travis-ci.org/SpongePowered/SpongeAPI/jobs/62137985#L257

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
boformer: Or just ``get(UUID)`` (Shortcut for ``get(UUID, true)``)and ``get(UUID, boolean)``

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: A ``User`` object could also be deleted (if you call the method twice).

I will just add both methods...

src/main/java/org/spongepowered/api/service/user/UserStorage.java
boformer: For that use case, we should add similar methods to the profile resolver service:

``getAllCachedProfiles()`` and ``matchCachedProfile(String)``

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
ST-DDT: I prefer get and query.
Because the difference between them is very clear due to the method naming.

boformer: I'm not sure about this. Mojang recommends to cache profiles.
I don't want to provide a shortcut for "contact servers directly".

That's why i prefer 2 methods with the same name. Only developers who really want to force a server lookup will use the boolean function.

progwml6: The cache bypass is not the best thing to let people near

MattBDev: Agreed. It's best to have it default to true because performing too many direct requests to Mojang servers can cause the the server to be blacklisted from accessing the API.
Most developers will most likely prefer to use `get(UUID uniqueId)` which will be a lot safer and avoid this risk.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,135 @@
...
+     */
+    ListenableFuture<Collection<GameProfile>> getAllById(Iterable<UUID> uniqueIds, boolean useCache);
+
+    /**
+     * Returns a collection of matching cached {@link GameProfile}s whose last
+     * known user names start with the given string (case-insensitive).
+     *
+     * <p>This collection may also contain profiles of players who never played
+     * on the server!</p>
+     *
```
MattBDev: If this only fetches from the local cache then how can it contain profiles of players who have never played?

boformer: Whenever you do a lookup with get(UUID/name), the profile is saved in the local cache.

You can also lookup profiles of players who never played on the server.

MattBDev: Thanks for this clarification.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,135 @@
...
+     * @return The result of the request
+     */
+    Collection<GameProfile> match(String lastKnownName);
+
+    /**
+     * Gets a collection of all cached {@link GameProfile}s.
+     *
+     * @return A {@link Collection} of {@link GameProfile}s
+     */
+    Collection<GameProfile> getAllCachedProfiles();
```
Minecrell: Why not just `getAll`?

Minecrell:
https://travis-ci.org/SpongePowered/SpongeAPI/jobs/62295585#L264

boformer:
I squashed the commits.

MattBDev:
:heart_eyes: Looks awesome

boformer:
I fixed the travis errors and renamed the method to ``getAllCached()``.

src/main/java/org/spongepowered/api/service/profile/GameProfileResolver.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return The result of the request
+     */
+    ListenableFuture<Collection<GameProfile>> getAllById(Iterable<UUID> uniqueIds, boolean useCache);
+
+    /**
+     * Gets a collection of all cached {@link GameProfile}s.
+     *
+     * @return A {@link Collection} of {@link GameProfile}s
+     */
+    Collection<GameProfile> getAllCached();
```
Minecrell: I think this should be either `getCachedProfiles()` or `getAll()` but `getAllCached()` sounds a bit weird

boformer:
I renamed the method to ``getAllCachedProfiles`` to make clear what it does. ``getAll`` is misleading.

Minecrell:
@boformer `getAllCachedProfiles` is essentially the same as `getCachedProfiles`, the all is kinda redundant in the name.

boformer:
You are right.

src/main/java/org/spongepowered/api/service/profile/ProfileNotFoundException.java
```diff
@@ -0,0 +1,72 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.profile;
+
+/**
+ * Thrown by a profile lookup service if a profile does not exist.
+ */
+public class ProfileNotFoundException extends Exception {
```
gabizou: Couldn't this extend some other exception than just `Exception`?

gabizou: Never mind. It's fine to just extend `Exception` in this case.

simon816:
Why is this a service? I don't think it should be possible for plugins to replace the implementation of the  `GameProfileResolver` and `UserStorage`.

MattBDev:
I didn't realize that this was being added as a service. Extra plugins just to get a User shouldn't be needed. This should be only implemented by sponge

boformer:
@simon816 First of all, almost any system is a service, even the command and event system.

Maybe it would even make sense to have a custom GameProfileResolver with a custom caching solution etc.

modwizcode:
That may just represent a more fundamental problem with the way services
are being used.

On Thu, May 14, 2015 at 4:23 PM, boformer <notifications@github.com> wrote:

> @simon816 <https://github.com/simon816> First of all, almost any system
> is a service, even the command and event system.
>
> Maybe it would even make sense to have a custom GameProfileResolver with a
> custom caching solution etc.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpongePowered/SpongeAPI/pull/580#issuecomment-102175101>
> .
>



-- 

Starbuck Johnson, Java Developer
www.modwiz.com | starbuck@modwiz.com

boformer:
@modwizcode Services are a part of the Sponge philosophy. There are certain core services that can't be overridden. I don't see a problem with that.

modwizcode:
Alright I guess I misunderstood the goal of services.

On Thu, May 14, 2015 at 4:41 PM, boformer <notifications@github.com> wrote:

> @modwizcode <https://github.com/modwizcode> Services are a part of the
> Sponge philosophy. There are certain core services that can't be
> overridden. I don't see a problem with that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpongePowered/SpongeAPI/pull/580#issuecomment-102178139>
> .
>



-- 

Starbuck Johnson, Java Developer
www.modwiz.com | starbuck@modwiz.com

ryantheleach:
If it were possible for UserStorage to be overridden it would be awesome.

Store all the users out to a database instead, and you already have cross server inventory compatibility.

simon816:
If plugins were able to implement this, they would need to provide an implementation of `User`, which of course would be a problem as the implementation will be expecting it to be an instance of their own user (say `SpongeUser`), ultimately leading to cast exceptions.

MattBDev:
I think if anyone was to implement this in their plugin, they would do so only if they knew what they were doing. I retract my previous statement of not adding this as a service.

Issue #546, opened 2015-04-06
More useful Extent hierarchy


maxov:
Issues like #319, #453, and #540 point to me that the extent system is rather ill-defined and could serve from some splitting-up in terms of behavior.

My idea of locations, extents, and worlds has solidified because of issues like the ones I have pointed out. There's several interesting properties that are served by extents, what follows is some thoughts I have on those properties, not necessarily in any order:
- All extents I can think of have a 3D cartesian coordinate system. In some cases, manhattan distance is used instead of euclidean distance to calculate certain things, like light level and AI.
- The "base" extent is World. Every extent currently available in the Sponge API is either grounded by or is a world (which makes me wonder if a schematic extent or ways to construct block buffers would be interesting...). Chunks belong in worlds, "Regions" belongs in worlds, and any kind of "selection" (think WorldEdit) also belongs in a world.
- Some sub-extents like chunks define coordinate transformations for convenience. So, for instance, blocks can be located based on some corner of the chunk. Other extents like selections do not.
- Some (if not most) sub-extents have a range defined to them. Outside this range, it doesn't make sense to manipulate blocks in terms of that extent. Note that worlds do not _conceptually_ have a range and are infinite (but practically worlds become useless after some large number of blocks, I think 32 million as a hard-coded limit).
- Some sub-extents (mainly chunks and technically locations, if you consider them to be 1x1x1 extents) are defined as "grids" upon their base extent. This means that all sub-extents of one type (Chunk as an example) cannot intersect with each other and so form a perfect grid across the containing world. Note that this requires these kinds of sub-extents to have a same-size range or bound defined to them.
- Some extents take up a whole column of y-values. Perhaps it is easier to visualize this when comparing a "column extent" to a cuboid selection; the selection is bounded by two x values, two y values, and two z values. Extents like chunks and "regions" are only bounded by two x values and two z values, and take up every possible y value (256 now, but some Spout experiments were able to simulate higher values).

Some interesting consequences of a system that could implement this:
- Locations (or perhaps BlockLoc) become 1x1x1 gridded, coordinate-shifted "nano"-extents, much like chunks. It still makes sense to call all the Extent methods on them, the coordinates just have to be (0, 0, 0). I'm not advocating removing Location altogether, but it can extend Extent (heh) as well as add a coordinate transformation. 
[...]

PR #581, opened 2015-04-20
Add ChunkLayout class

DDoS:
Provides information about the chunk coordinate space for better cross-implementation support. The current API assume all implementations use the same chunk layout or leaves it undefined.

See https://github.com/SpongePowered/Sponge/issues/228
Implemented in https://github.com/SpongePowered/Sponge/pull/247

### DIFF
--- a/src/main/java/org/spongepowered/api/util/Direction.java
+++ b/src/main/java/org/spongepowered/api/util/Direction.java
@@ -68,8 +68,8 @@ public enum Direction {
     private final int flags;
     private Direction opposite;
 
-    private Direction(Vector3d vector3d, int flags) {
-        if (vector3d.length() == 0) {
+    Direction(Vector3d vector3d, int flags) {
+        if (vector3d.lengthSquared() == 0) {
             // Prevent normalization of the zero direction
             this.direction = vector3d;
         } else {
--- a/src/main/java/org/spongepowered/api/world/storage/ChunkLayout.java
+++ b/src/main/java/org/spongepowered/api/world/storage/ChunkLayout.java
@@ -1,0 +1,201 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/api/world/storage/ChunkLayout.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Returns the size of the chunks in blocks. The axes are matched to the corresponding vector component.
+     *
+     * @return The size of chunks
+     */
+    Vector3i getChunkSize();
+
+    /**
+     * Returns the maximum coordinates for chunks for each axis. The axes are matched to the corresponding vector component. Due to the limited
+     * precision of 32bit integer, there always is a practice limit.
```
kenzierocks: Maybe 'practical' not 'practice'?

src/main/java/org/spongepowered/api/util/Direction.java
```diff
@@ -68,8 +68,8 @@
     private final int flags;
     private Direction opposite;
 
-    private Direction(Vector3d vector3d, int flags) {
-        if (vector3d.length() == 0) {
+    Direction(Vector3d vector3d, int flags) {
+        if (vector3d.lengthSquared() == 0) {
```
kenzierocks: Not sure this change is in the scope of the PR

src/main/java/org/spongepowered/api/world/storage/ChunkLayout.java
DDoS: Oops.

src/main/java/org/spongepowered/api/util/Direction.java
DDoS: Doesn't really matter, it just saves a sqrt operation.

AlphaModder:
This is extremely awesome. Especially because this could mean sponge could work with a Cubic Chunks system.

gratimax:
Linking #546 since there are remarkable similarities in goals that these accomplish...

PR #590, opened 2015-04-25
Add irc protocol links to README

Phroa:
Anybody think I should use `ircs://` links? My reasoning in not doing so is I know I had to go and download an openssl dll and shove it in some arcane folder to even enable ssl at all in mIRC.

Not sure if anybody cares, or if that has even been patched in some form by now.

For those saying it's a stupid formatting PR, you're right. Blame @gabizou.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ A Minecraft plugin API (not including an implementation). It is licensed under t
 * [Source]
 * [Issues]
 * [Wiki]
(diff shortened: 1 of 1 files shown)


FerusGrim:
@Zidane I just noticed your tag to work with @Phroa on the PR.

The problem with this, is that mark(down/up) has some kind of qualm about having two links on the same line. This is, possibly, only an issue on lists.

If you look  these changes don't actually appear visually.

PR #602, opened 2015-04-28
[WIP] A better, easier to use Data API

gratimax:
Will elaborate a better description later.

Simply put, the existing Data API is not the best because it is verbose. I want to decrease verbosity, yet keep type safety.

This is done with a generic object-property model that mirrors the current entity-component model, except the guarantees are flexible. In the signature of DataObject, the type parameter reflects how restrictive it is, in terms of marker interfaces.

Here's some code from me playing around earlier:

```java
Block myblock;

// Does not compile
myblock.set(Props.HEALTH, 5.0);

Entity myentity;

myentity.get(PositionData.class).set(X_POSITION, 5.0);

// Does not compile
myentity.get(PositionData.class).set(HEALTH, 9.0);

myentity.set(HEALTH, 5.0);
```

A lot of everything is still stubs. The basic idea is in place, but the values are all missing and I've yet to rework DataManipulators and the like.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/DataObject.java
+++ b/src/main/java/org/spongepowered/api/data/DataObject.java
@@ -45,20 +45,18 @@ public interface DataObject<D> {
     /**
      * Sets some value.
      *
-     * @param prop
+     * @param prop The property
      * @param value The value to set
      * @param <E> The type of value
      * @param <V> The subtype of data that this DataObject supports
-     * @return The value
+     * @return This object
      */
     <E, V extends D> DataObject<D> set(Prop<E, V> prop, E value);
 
     /**
      * Gets a data object with further restrictions on data.
      *
-     * <p>
-     *     As an example, use this to get a data "manipulator" from a more generic data object.
-     * </p>
+     * <p>As an example, use this to get a data "manipulator" from a more generic data object.</p>
      *
      * @param clazz The class of data to restrict to
      * @param <V> The type of data to restrict to
--- a/src/main/java/org/spongepowered/api/data/DataObject.java
+++ b/src/main/java/org/spongepowered/api/data/DataObject.java
@@ -1,0 +1,69 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data;
+
+/**
+ * Represents some container of arbitrary data values.
+ *
+ * @param <D> The type of data that this DataObject is restricted to
+ */
+public interface DataObject<D> {
+
+    /**
+     * Gets some value that is known to be present.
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/api/data/DataObject.java
```diff
@@ -0,0 +1,69 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data;
+
+/**
+ * Represents some container of arbitrary data values.
+ *
+ * @param <D> The type of data that this DataObject is restricted to
+ */
+public interface DataObject<D> {
```
gabizou: You removed `DataSerializable` from TileEntity, but didn't replicate it back here, is there a reason?

gabizou:
Mostly because there is lacking documentation on `EntityData`, `GameData`, and `TileEntityData`, I'm not sure I'm going to like having a hierarchy of data types.

src/main/java/org/spongepowered/api/data/DataObject.java
gratimax: I've yet to decide what to do with it. I'm sure you know that I dislike it as an interface, but as I probably won't think of anything better I'll add it back

gratimax:
@gabizou The beauty of the hierarchy is that it is very easily customizable. If there's a specific case where the types don't work, we can remove it. When I add entity data it'll become more clear, it'll be able to keep the type safety of individual data manipulators.

gabizou:
> If there's a specific case where the types don't work, we can remove it

Right, but the problem with that is eventually, we break backwards compatibility when certain data types no longer are applicable to a specific entity. The biggest advantage to having no contractual restrictions on the "types" of data possibly applicable is that you can represent any `DataObject` by another `DataObject`, or more importantly in the case of `ItemStack`s representing entire constructs of an entity, so what then? `ItemStack` not only accepts `ItemData`, but also `EntityData` due to the nature of Minecraft already present in 1.8.

src/main/java/org/spongepowered/api/data/DataObject.java
gratimax: Also technically it's not gone yet since DataHolder extends DataSerializable (ahhh the hierarchy...).

Mumfrey:
> Simply put, the existing Data API sucks because it is verbose.

This is the most ridiculous statement I've ever read.

> Here's some examples of me playing around earlier:
> playing around
> playing around

Your example is close to line-noise. If this is your goal and you're going to replace verbosity with complete illegibility which literally *encourages* bad code, I don't see how you're going to justify pulling this.

src/main/java/org/spongepowered/api/data/DataObject.java
gabizou: Also another thing, why doesn't this just replace `DataHolder`?

gratimax:
@Mumfrey 
I had a few minutes to elaborate why I am making this PR and work on it a bit today, and that's it. That's why the justification sucks and the examples also suck (and I know they suck). I would really appreciate less inflammatory comments as I work on this PR. 

@gabizou 
Zidane also expressed discontent with the `GameData`/`EntityData`, etc. hierarchy, and I can remove it. But I want to wait until I add the more component-like properties of entities first.

src/main/java/org/spongepowered/api/data/DataObject.java
gratimax: In another line note, "Keeping it for now before I overhaul DataHolder"

Zidane:
@gratimax 

I would encourage you to not start an OP off in the future with that stance. It begins a PR with hostility and distracts away from your goal of improvement.

src/main/java/org/spongepowered/api/data/props/Props.java
```diff
@@ -0,0 +1,336 @@
...
+import org.spongepowered.api.data.types.StoneType;
+import org.spongepowered.api.data.types.TreeType;
+import org.spongepowered.api.data.types.WallType;
+import org.spongepowered.api.util.Axis;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.api.util.rotation.Rotation;
+
+import java.util.List;
+
+public class Props {
```
Zidane: Could make Prop(Property) a CatalogType

gratimax:
@Zidane Had I more time I would put in a nicer description, but I quite literally jotted down the first thought s in my head.

Mumfrey:
> @Mumfrey
> I had a few minutes to elaborate why I am making this PR and work on it a bit today, and that's it. That's why the justification sucks and the examples also suck (and I know they suck). I would really appreciate less inflammatory comments as I work on this PR.

I don't really buy that. If you want to present a PR for consideration then it behoves you to write up a cogent summary of your objectives, and the scope of the PR. Since you've chosen to label this as "better" (which is a subjective statement posing as an objective one), it seems only fair that your description should back up that claim.

You say I should be less inflamatory, but I can only guess how @gabizou must feel when you make remarks such as 

> Simply put, the existing Data API sucks

Which is a purely subjective view of yours which doesn't really bear scrutiny and you don't even bother to justify. Personally, if someone made comments like this about something which I'd spent the best part of *two months* designing, and then decided they could make some half-baked rebuttal with no corroborating evidence and make derisive comments about my own work at the same time I'd be absolutely devastated.

It's not warranted, and I think you should understand that your attitude going into this is going to put people's noses out of joint. If you can't understand why it would be insulting to someone else to come up with this kind of remark, I suggest you sit down and think about how *you* would feel were the roles reversed.

gratimax:
This was obviously a mistake.

Zidane:
@gratimax 

Everyone is for improving the API's quality. We don't have anything against you wanting to improve the data API (I encouraged you to do so by asking you to make a PR/issue; you've made valid points and @gabizou agrees as well), but starting off the post in such a manner was really insulting.

I see your revised OP and that would have been perfect from the get go and everyone would have been just fine.

PR #603, opened 2015-04-28
Add methods to support key-value options within the permissions inheritance structure

zml2008:
Options are commonly used for things like a user's prefix and suffix, which are attached to a subject but are often useful to have inherited from defined parents.

These are used as arbitrary metadata for various features, like a prefix, or player-set messages.

Options are implemented as a subinterface of the existing permissions interfaces. This way, plugins can simply check instance and cast from an existing permissions service provider. However, this means that no plugin can provide only options (imo having two separate inheritance systems available through two services would just be a bad idea & create confusion).

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
+++ b/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
@@ -31,10 +31,12 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import org.spongepowered.api.service.permission.context.Context;
+import org.spongepowered.api.service.permission.option.OptionSubjectData;
 import org.spongepowered.api.util.Tristate;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -47,9 +49,10 @@ import javax.annotation.Nullable;
  *
  * <p>This class is thread-safe.
  */
-public class MemorySubjectData implements SubjectData {
+public class MemorySubjectData implements OptionSubjectData {
 
     private final PermissionService service;
+    private final ConcurrentMap<Set<Context>, Map<String, String>> options = Maps.newConcurrentMap();
     private final ConcurrentMap<Set<Context>, NodeTree> permissions = Maps.newConcurrentMap();
     private final ConcurrentMap<Set<Context>, List<Map.Entry<String, String>>> parents = Maps.newConcurrentMap();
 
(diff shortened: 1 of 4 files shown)


boformer:
Is this really neccessary? Something like that is also possible without Options, with a separate plugin configuration.

In the end this will cause permission plugins to become a megaplugins which store permission data, group data and **metadata**.

Example how it works without Options: Chat Prefix Plugin

*Prefix Plugin config:*
```yaml
// Ranks can be applied by giving permission "chatprefix.rank.<name>"
ranks {
    admin { prefix = "[Admin]" }
    donor { prefix = "[Donor]" }
}
```
*Permission Plugin config:*
```yaml
groups {
    admins {
        permissions { chatprefix.rank.admin = true }
    }
    donors {
        permissions { chatprefix.rank.donor = true }
    }
}
players {
    Notch {
        groups = [admins, donors]
    }
    zml2008 {
        groups = [donors]
    }
}
```

zml2008:
Then you bring in contexts, and expect every plugin that wants to do data to store it in a different place, really messing up separation of concerns when writing plugins. A permissions plugin should focus on being very good as storing player data (permissions are basically the same as options).

I think it would be difficult (@gabizou correct me if I am wrong) to have a permissions plugin be the provider for any plugin that wants to provide its own Subjects that get data from the permissions plugin. If that weren't possible, it would mean that each plugin would have a separate data store, and would have to implement a lot of the features of a permissions plugin anyway, which ends up being fairly complex (and could lead to performance issues in the future).

gratimax:
Objects currently in the data API represent more ingame things like item data, entity data, and the like. I don't know too much about the Permissions implementation, but I think that the only thing we can say about a Subject for sure is whatever properties are given to it by the Permissions API (parents, permissions, identifier), and not any data that can be actively persisted.

ryantheleach:
@boformer The main advantage of storing stuff in the permissionsAPI is the inheritance system.

bPermissions, and zPermissions (and probably more not familiar with the new pex or group manager) both had this ability in the plugins and had to rely upon Vault to provide plugin independent access.

The only solutions I see is either,
1. Each plugin having it's own storage, where each data group is assigned via a permission.
2. A new service being declared that plugins can implement in order to supply the information.
3. Having a DataAPI dataManipulator attached to subjects that all plugins know about...

Compared to automatically having the inheritance system everyone is somewhat used to, getting contexts for free, and having a cross plugin method of accessing everything out of the box.

Especially when it comes to permissions that are essentially x.x.blocktype or x.x.1 x.x.2 x.x.3 x.x.25 Instead of using regex or other obtuse methods we can just support options.

zml2008:
@SpongePowered/developers any comments/issues blocking merge?

Mumfrey:
This is one of my favorite features in pex, gets my vote!

gabizou:
I'm fine with the merge.

Zidane:
No arguments here. You are free to merge this WITHOUT THE MERGE BUTTON @zml2008

zml2008:
kk :)

PR #605, opened 2015-04-29
[WIP] Value API

gratimax:
This is a tracker for iterative work done on the data API for the purpose of allowing type-bounded and keyed data manipulation. 

I make no guarantees until I deem it "finished'.


Deamon5550:
Just going to put this here so it is not forgotten.

One of the primary goals of the data api, if not THE primary goal, is the seperation of data from the holding objects. This allows firstly much of the duplicated code to be reused (such as the command data being used by command blocks, and command block minecarts), and for the data to be held by other objects (a chest item holding its contents, or a dungeon populator holding a representation of the mob spawners it'll be placing down). This seperation is something which is absolutely critical in my oppinion.

src/main/java/org/spongepowered/api/data/BoundedProp.java
```diff
@@ -22,11 +22,13 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.data.manipulators.blocks;
+package org.spongepowered.api.data;
 
-import org.spongepowered.api.data.manipulators.SingleValueData;
-import org.spongepowered.api.data.types.StoneType;
+// TODO Does a numeric property (int, double) imply boundedness for all possible cases?
+public interface BoundedProp<E, V> extends Prop<E, V> {
```
gabizou: Possibly make `E extends Comparable<E>` so that the bounds can be *properly* calculated by both the developer using this and the implementation?

gratimax: In IRC I said that we may want to use some comparisons which aren't the "natural" implementation ,so it will most likely require a getComparator method instead.

gratimax:
An open-ended question to anyone looking in:
```java
public static final Prop<Double, DamagingData> DAMAGE = null;
 
// TODO could do capitals, but is bad style
public static Prop<Double, DamagingData> damageForEntity(EntityType type) {
    return null;
}
 
entity.get(Props.DAMAGE);
 
entity.get(Props.DAMAGE_FOR_ENTITY(EntityTypes.SKELETON);
// vs
entity.get(Props.damageForEntity(EntityTypes.Skeleton));

// or maybe another solution like another Prop type? idk
```

ST-DDT:
@gratimax I have rarely seen capital method names and i kind of don't like them.
Method names should always be camel case although they might return the same result over and over again. Ex: `Singleton.getInstance()`.

(Public) static final Fields however should be uppercase. Ex: `Singleton.INSTANCE`

So IMO this one is the right one:
````java
entity.get(Props.damageForEntity(EntityTypes.SKELETON));
````
I'm used to this style from Hibernate as well (`Projections` and `Transformers`).

I also prefer this one over the following one:
````java
entity.get(Props.DAMAGE_FOR_ENTITY, EntityTypes.SKELETON); 
````
(Get with additional param), as it opens the API for possible inconsistencies.

**Comment by gratimax** [comment]
@ST-DDT Totally not doing get with additional param, because there's inflexibility there and it is no longer typesafe.

File: src/main/java/org/spongepowered/api/data/DataObject.java [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Gets a data object with further restrictions on data.
+     *
+     * <p>As an example, use this to get a data "manipulator" from a more generic data object.</p>
+     *
+     * @param clazz The class of data to restrict to
+     * @param <V> The type of data to restrict to
+     * @return The nested data object
+     */
+    <V extends D> DataObject<V> get(Class<V> clazz);
```
Zidane: This still returns data object?

gratimax: Yes, as said by the javadocs.

gratimax:
I've moved branch names so it's now #685.

Issue #615, opened 2015-05-01
Allow listening for events without annotations


Minecrell:
Currently, the event system requires the use of annotations. This is quite inconvenient, especially for plugins implementing a kind of dynamic event registration, for example if all needed events are not known at compile time. This affects especially scripting plugins which would always need to add a separate event handler method for every event accessible through their scripting API.

As an example, [Pore](https://github.com/LapisBlue/Pore) uses dynamic event registration for calling Bukkit's events already, however currently it has to [generate classes at runtime](https://github.com/LapisBlue/Pore/blob/master/src/main/java/blue/lapis/pore/event/PoreListenerGenerator.java#L61-L134) to make that work. There should be an easy way to register events using a simple interface additionally instead of just using annotations, for example something like this:

```java
interface EventHandler<T> {
    void handleEvent(T event);
}

interface EventManager {
    // ...
    <T extends Event> void register(Object plugin, Class<T> eventClass, EventHandler<T> handler);
}

game.getEventManager().register(plugin, ServerStartedEvent.class, new EventHandler<ServerStartedEvent> {

    @Override
    public void handleEvent(ServerStartedEvent event) {
        // Do something
    }
});
```

The only thing missing here would be the `Order` of the event and `ignoreCancelled`, but that could be just added as additional overload for the registration method.

Zidane:
@Minecrell 

Issue is we would have a chicken and egg problem...unless you would want lifecycle events to remain annotation-based.

Minecrell:
@Zidane 
You mean like if I would want to register something like `PreInitializationEvent` without annotations?

Zidane:
Yes

Minecrell:
@Zidane 
This is not supposed to _replace_ the current annotation-based event system. Plugins still need to add a `@Plugin` class so they can just add the annotation-based handler there. This is only to help when you don't know exactly which events you will want to listen to at runtime, and obviously you do know that with the state events.

Zidane:
@Minecrell 

Right but some may ask for whichever way to pull off not needing annotated events at all.

Bammerbom:
I also need this for my Plugin.

ST-DDT:
@Minecrell 
The `EventManager` should probably have some more wildcards.
`<T extends Event> void register(Object plugin, Class<T> eventClass, EventHandler<? super T> handler)`
or
`<T extends Event> void register(Object plugin, Class<? extends T> eventClass, EventHandler<T> handler)`

[...]

PR #617, opened 2015-05-01
Improve event system to allow registering events without annotations

Minecrell:
This will make a few changes in the event system to add/fix the following:
- Allow registering events without annotations (see #615)
- Fix a few ambiguous Javadocs
- Add a few minor methods

### DIFF
--- a/src/main/java/org/spongepowered/api/event/EventHandler.java
+++ b/src/main/java/org/spongepowered/api/event/EventHandler.java
@@ -24,32 +24,18 @@
  */
 package org.spongepowered.api.event;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
-
-import org.spongepowered.api.service.event.EventManager;
-
-import java.lang.annotation.Retention;
-import java.lang.annotation.Target;
-
-@Retention(RUNTIME)
-@Target(METHOD)
-public @interface Subscribe {
+/**
+ * Represents a handler accepting events of a specified type.
+ *
+ * @param <T> The type of the event
+ */
+public interface EventHandler<T extends Event> {
 
     /**
-     * The order this handler should be called in relation to other handlers in
-     * the {@link EventManager}.
+     * Called when a event registered to this handler is called.
      *
-     * @return The order the handler should be called in
+     * @param event The called event
      */
-    Order order() default Order.DEFAULT;
-
-    /**
-     * Whether this handler should ignore cancelled events. If enabled the
-     * handler will not be called for any cancelled events.
-     *
-     * @return If the handler should ignore cancelled events
-     */
-    boolean ignoreCancelled() default true;
+    void handle(T event);
 
 }
--- a/src/main/java/org/spongepowered/api/service/event/EventManager.java
+++ b/src/main/java/org/spongepowered/api/service/event/EventManager.java
@@ -25,6 +25,9 @@
 package org.spongepowered.api.service.event;
 
 import org.spongepowered.api.event.Event;
+import org.spongepowered.api.event.EventHandler;
+import org.spongepowered.api.event.Order;
+import org.spongepowered.api.event.Subscribe;
 
 /**
  * Manages the registration of event handlers and the dispatching of events.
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/service/event/EventManager.java
```diff
@@ -34,13 +35,23 @@
     /**
      * Registers an object to receive {@link Event}s.
      *
-     * @param plugin A plugin instance
+     * @param plugin The plugin instance
      * @param obj The object
      * @throws IllegalArgumentException Thrown if {@code plugin} is not a plugin instance
      */
     void register(Object plugin, Object obj);
```
ST-DDT: I guess we should mention here that this method searches for annotations, or how the input is used at all.
For example what happens if we pass in an EventHandler? (For example if we forgot the class param)

Maybe we should rename the method to make it more clear.
`registerAnnotatedHandlers(...)`

We also should return the actually registered listeners from this method
`Collection/Set<EventHandler> registerAnnotatedHandlers(Object plugin, Object obj)`;

src/main/java/org/spongepowered/api/service/event/EventManager.java
```diff
@@ -48,10 +59,17 @@
     void unregister(Object obj);
 
     /**
+     * Un-registers all event handlers of a plugin.
+     *
+     * @param plugin The plugin instance
+     */
+    void unregisterPlugin(Object plugin);
```
Zidane: Maybe just unregister?

src/main/java/org/spongepowered/api/event/EventHandler.java
```diff
@@ -0,0 +1,41 @@
...
+ * @param <T> The type of the event
+ */
+public interface EventHandler<T extends Event> {
+
+    /**
+     * Called when a event registered to this handler is called.
+     *
+     * @param event The called event
+     */
+    void handleEvent(T event);
```
Zidane: handle here. No need to repeat event.

src/main/java/org/spongepowered/api/service/event/EventManager.java
Zidane: We aren't renaming this method. We also aren't returning the handlers as it would have to be immutable and that would really do no good for you.

The better way would be to let you ask the event manager (by class) for an immutable list of eventhandlers (or just the amount of handlers registered for that event, which is what anyone really wants that for anyhow).

src/main/java/org/spongepowered/api/service/event/EventManager.java
```diff
@@ -34,13 +35,23 @@
...
 
     /**
+     * Registers an event handler for a specific event class.
+     *
+     * @param plugin The plugin instance
+     * @param eventClass The event to listen to
+     * @param handler The handler to receive the events
+     * @param <T> The type of the event
+     */
+    <T extends Event> void register(Object plugin, Class<T> eventClass, EventHandler<? super T> handler);
```
ST-DDT: How do i specify the priority and ignoreCancelled flag?
Maybe we should pass in a `@Subscribe` (fake) instance or something similar (`EventHandlerOptions`).
That way we could make this method the main method and the method above (annotations) will (internally) use this method to register all found handlers.
We could also remove the above (annotations) method from this interface and move it to a fully implemented helper/util class. If you are interested i could write down some more details.

src/main/java/org/spongepowered/api/service/event/EventManager.java
ST-DDT: > We aren't renaming this method

(Why not?) Are you adding some clarifications to the javadocs?

> We also aren't returning the handlers as it would have to be immutable and that would really do no good for you.

It would be quite useful to be able to realign the priority of the event handlers to better match another plugin's handler's priority. Example execute the Login plugin's handler after the daily join gift handler.

Zidane: (Why not?) Are you adding some clarifications to the javadocs? Or is it meant to be multi-functional.

Because this is an established method and everyone knows about it at this point.

    It would be quite useful to be able to realign the priority of the event handlers to better match another plugin's handler's priority. Example execute the Login plugin's handler after the daily join gift handler.

Way to easy for abuse. Plugins shouldn't have any control over the event pipeline.

src/main/java/org/spongepowered/api/service/event/EventManager.java
Minecrell: Conflicts with the other `unregister` method, and if given a plugin instance we wouldn't be able to tell if it's supposed to unregister the plugin class only or all plugin listeners.

src/main/java/org/spongepowered/api/service/event/EventManager.java
ST-DDT: >> It would be quite useful to be able to realign the priority of the event handlers to better match another plugin's handler's priority. Example execute the Login plugin's handler after the daily join gift handler.

> Way to easy for abuse.

There is no more abuse in there than you can already do with the new `register(...EventHandler)`

src/main/java/org/spongepowered/api/service/event/EventManager.java
Minecrell: We are not removing the method to register annotated event handlers. This is just additional, the primary supported way to listen for events is still the annotation-based one.

ST-DDT: Thanks for your explanation.

How do i specify the `priority` and `ignoreCancelled` flag?

Minecrell: Added the order now, you can just check if the event is cancelled on your own.

Issue #625, opened 2015-05-02
Provide JSON dependency for plugins.


Minecrell:
Plugins will quite often need to work with JSON, especially when working with web services. We should provide a JSON dependency in SpongeAPI so plugins don't need to shade the same dependency over and over again. I think there are basically 2 options:

- **[GSON](https://github.com/google/gson):** This is included in Minecraft already so there would be no additional dependency to shade - there is basically no change for our implementation when we would provide this for plugin developers.
- **[Jackson](https://github.com/FasterXML/jackson)**
- **...?**

RobertHerhold:
I vote Jackson. I've been using it forever and I love it.

gratimax:
Personal preference shouldn't be a factor, I think that practicality here is more important. While I have used both and like both, Gson is shaded with Minecraft and does not need to be distributed further.

Minecrell:
I agree with @gratimax here, we should try to keep our distribution as small as possible, if plugin authors want something special they can still package these dependencies together with their plugins.

zml2008:
imo jackson is preferable because it provides more flexibility in how json is parsed and represented. Also take a look at configurate-json and configurate-gson.

progwml6:
gson 2.2.4 is already in the classpath for both implementations use it

gratimax:
I feel like you've closed this prematurely. Sure, Gson is on the classpath, but we haven't answered the question of if it should be an API dependency...

jamierocks:
This comes with the assumption that every Minecraft implementation is bundled with GSON. :|

Minecrell:
@jamierocks If not, they can just include it when implementing our API.

jamierocks:
I was merely pointing out not to include GSON, because Minecraft has it. You should pick the one developers want...

Minecrell:
@jamierocks This is pretty much personal preference. If plugin authors want to use a different library they are free to do so, but we provide GSON because we have it already available. We bundle quite a lot of dependencies already, there is no real need for 20MB+ JARs just to bundle every dependency a developer could ever need.

PR #630, opened 2015-05-03
Add GSON and commons-lang3 to dependencies

Minecrell:
This adds 2 new dependencies to SpongeAPI (primarily for plugin authors to use). The distribution of our both implementations will not be affected by this. Both dependencies are already included in Minecraft so this just exposes them to plugin developers.

- **GSON:** As a JSON dependency, see #625. While some may prefer another library to work with JSON (e.g. Jackson), GSON is already included in Minecraft and so we don't need to bundle anything additional. Plugin authors can still bundle an alternative library to work with JSON.

- **commons-lang3:** I'm not completely sure if we should provide this. It contains quite a lot of utilities which make stuff easier to use, and given its size (~600 KB) it would be stupid if plugin developers would need to bundle it (without a dependency in SpongeAPI there is no guarantee it's available in all implementations). It's included in Minecraft already so this would not affect many implementations.

Additionally updated `slf4j-api` to 1.7.12 and Guice to the 1.4 release. Didn't cause any problems in the implementation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,12 +15,12 @@ apply from: 'gradle/java.gradle'
 
 // Project dependencies
 dependencies {
-    compile 'org.slf4j:slf4j-api:1.7.7'
+    compile 'org.slf4j:slf4j-api:1.7.12'
     compile 'com.google.guava:guava:17.0'
     compile 'com.google.code.gson:gson:2.2.4'
     compile 'org.apache.commons:commons-lang3:3.3.2'
     compile 'com.google.code.findbugs:jsr305:1.3.9'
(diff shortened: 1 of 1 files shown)


Minecrell:
@Zidane This is basically ready for merge if there are no objections - there were not a lot of changes so there is not much reason to keep it open that long.

Zidane:
You have a go @Minecrell

jamierocks:
What about updating guava?

Minecrell:
@jamierocks We use the versions provided by Minecraft, the updated dependencies are provided by us.

Issue #641, opened 2015-05-06
Redundant or inconsistant get methods in PlayerEvents


SonarBeserk:
getEntity()/GetHuman()/getLiving()/getPlayer()/getUser() all return player. As per documentation getHuman() should seemingly specifically return a human object and getLiving() a living object. getEntity()/getPlayer()/getUser() however seems outright redundant.

gratimax:
I don't quite see what the issue is. Sure a lot of these methods seem redundant, but they're useful in terms of several event base types, like HumanEvent, PlayerEvent, UserEvent, and EntityEvent. Technically a PlayerEvent's getPlayer/getLiving/getEntity/getHuman/getUser all return the same thing, but it means that the PlayerEvent has a specific set of methods related to what it can actually return. 

If we took out these methods, we would have to cast to player from PlayerEvent like this:
```java
PlayerEvent myEvent;
myEvent.getEntity() // -> Entity
(Player) myEvent.getEntity() // -> Player 
```
So we basically trade a concise getPlayer method for a less concise getEntity method that _also_ requires an unchecked cast, making the compiler issue warnings or require a `@SuppressedWarnings` annotation.

No, generics are not a solution (`EntityEvent<E extends Entity>`) because they are not available at runtime, meaning that the event buses wouldn't be able to differentiate EntityEvents based on whatever subtype is expressed by the generics.

All in all, I don't see a problem. And even if there was a problem, trading more concise methods for less concise ones causes more problems than it solves.

SonarBeserk:
Proper code wouldnt be making an unchecked cast imo. Should be type checking.

PR #643, opened 2015-05-06
Standardize use of getEntity() in UserEvent and subclasses

SonarBeserk:
Intended as a potential fix for #641. Guaranteed to be breaking but some form of change like this is needed for sanity purposes. Please do give any feedback possible!

A pr has been created to fix the mixins in the current sponge implementation at https://github.com/SpongePowered/Sponge/pull/264

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -951,6 +951,7 @@ public final class SpongeEventFactory {
         values.put("cause", Optional.fromNullable(cause));
         values.put("block", block);
         values.put("entity", player);
+        values.put("user", player);
         values.put("replacementBlock", replacementBlock);
         values.put("blockFaceDirection", direction);
         values.put("exp", exp);
@@ -969,6 +970,7 @@ public final class SpongeEventFactory {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("entity", player);
+        values.put("user", player);
         values.put("fishHook", fishHook);
         return createEvent(PlayerCastFishingLineEvent.class, values);
     }
@@ -986,6 +988,7 @@ public final class SpongeEventFactory {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("entity", player);
+        values.put("user", player);
         values.put("fishHook", fishHook);
         values.put("caughtEntity", Optional.fromNullable(caughtEntity));
         return createEvent(PlayerHookedEntityEvent.class, values);
@@ -1007,6 +1010,7 @@ public final class SpongeEventFactory {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("entity", player);
+        values.put("user", player);
         values.put("fishHook", fishHook);
         values.put("caughtEntity", Optional.fromNullable(caughtEntity));
         values.put("caughtItem", Optional.fromNullable(caughtItem));
@@ -1032,6 +1036,7 @@ public final class SpongeEventFactory {
         values.put("cause", Optional.fromNullable(cause));
         values.put("block", block);
         values.put("entity", player);
+        values.put("user", player);
         values.put("replacementBlock", replacementBlock);
         values.put("blockFaceDirection", direction);
         return createEvent(PlayerChangeBlockEvent.class, values);
@@ -1050,6 +1055,7 @@ public final class SpongeEventFactory {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("entity", player);
+        values.put("user", player);
         values.put("newGameMode", newGameMode);
         values.put("oldGameMode", oldGameMode);
         return createEvent(PlayerChangeGameModeEvent.class, values);
@@ -1068,6 +1074,7 @@ public final class SpongeEventFactory {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
         values.put("entity", player);
+        values.put("user", player);
         values.put("fromWorld", fromWorld);
         values.put("toWorld", toWorld);
         return createEvent(PlayerChangeWorldEvent.class, values);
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/event/entity/player/UserEvent.java
```diff
@@ -37,6 +37,6 @@
      *
      * @return The user
      */
-    User getUser();
+    Entity getEntity();
```
Deamon5550: `User` does not extend `Entity` so this one should stay.

src/main/java/org/spongepowered/api/event/entity/living/LivingEvent.java
```diff
@@ -37,8 +38,5 @@
      *
      * @return The {@link Living} involved
      */
-    Living getLiving();
-
-    @Override
-    Living getEntity();
+    Entity getEntity();
```
Deamon5550: Should return `Living` not `Entity`.

SonarBeserk: Java doesnt allow use of anything but entity here, thus why it was used. Much would have preferred it to be living.

src/main/java/org/spongepowered/api/event/entity/living/human/HumanEvent.java
```diff
@@ -37,11 +37,5 @@
      *
      * @return The {@link Human} involved
      */
-    Human getHuman();
-
-    @Override
-    Human getLiving();
-
-    @Override
```
Deamon5550: Do not remove the override annotation

src/main/java/org/spongepowered/api/event/entity/living/LivingEvent.java
Deamon5550: Uh, no you're incorrect try it again, also you need to keep the override annotation.

zml2008:
This is good for cleanup, removes some redundancy in the methods. If you've tested this, I don't have a problem with it.

src/main/java/org/spongepowered/api/event/entity/player/UserEvent.java
zml2008: yeah, this needs to be fixed

SonarBeserk:
After pulling in changed files and updating mixins, the results is a more clear set of methods and player method calls are working absolutely fine.

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -1113,13 +1092,11 @@ public static PlayerChangeWorldEvent createPlayerChangeWorld(Game game, Player p
     public static PlayerChatEvent createPlayerChat(Game game, Player player, CommandSource source, Text message) {
         Map<String, Object> values = Maps.newHashMap();
         values.put("game", game);
+        values.put("user", player);
         values.put("entity", player);
+        values.put("user", player);
```
spaceemotion: uh, is this intended? You already have the user added two lines above :)

src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
```diff
@@ -951,11 +951,8 @@ public static PlayerBreakBlockEvent createPlayerBreakBlock(Game game, Cause caus
         values.put("cause", Optional.fromNullable(cause));
         values.put("block", block);
         values.put("entity", player);
-        values.put("replacementBlock", replacementBlock);
-        values.put("player", player);
         values.put("user", player);
-        values.put("human", player);
-        values.put("living", player);
+        values.put("replacementBlock", replacementBlock);
```
gabizou: Why are you moving this line?

SonarBeserk: Thats an issue of my needing to add and remove the value for user. Seemingly I readded them at the wrong spot. Arg

src/main/java/org/spongepowered/api/event/entity/player/PlayerEvent.java
```diff
@@ -38,18 +38,6 @@
...
-    Player getHuman();
-
-    @Override
-    Player getLiving();
-
     @Override
     Player getEntity();
-
-    @Override
-    Player getUser();
```
gabizou: What about `UserEvent`? A `Player` is a subclass of `User`, so there should be an override for that.

SonarBeserk: The thought was that knowing a player is a type of user, you are getting the user as well.

gabizou: Yes but `getUser()` returns `User`, not `Player`. The whole point of having the overrides is so that you can return a specific type of `User`, in this case, `Player`.

SonarBeserk: Hmm so you would prefer to keep an overridable getUser() over using the getUser() inherited?

gabizou: Since `getUser()` is *required* for a `UserEvent`, I'd rather we have an override for `getUser()` returning `Player` in `PlayerEvent`.

src/main/java/org/spongepowered/api/event/entity/player/PlayerEvent.java
```diff
@@ -38,18 +39,9 @@
...
-    @Override
-    Player getLiving();
-
     @Override
     Player getEntity();
 
     @Override
-    Player getUser();
-
+    User getUser();
```
gabizou: Keep the `Player getUser()` override.

SonarBeserk: User Event returns a user, thus the subclasses should as well. Plus if we return a direct player the method is basically pointless and redundant.

gabizou: `Player` extends `User`. Why would we not override to return `Player` in a `PlayerEvent`? There is no other "object" that represents the `User` that is a `Player` since the `Player` IS the `User`.

SonarBeserk: Thus there is no reason to return anything but user. It still functions the same way, and keeps in line with super classes.

gratimax: There is no reason to _not_ return player. It gives you more specifics and it is overall more useful. You are already changing getEntity, so you need to change getUser as well, if for nothing else other than consistency.

PR #658, opened 2015-05-08
Add an implementable economy service.

Xemiru:
The forums have been /flooded/ with so many economy service providers. To allow server owners to use what they choose instead of the one chosen by most developers, an `EconomyService` has been added for plugins to implement. Developers can instead reference to this service to interact with server economy.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/economy/Account.java
+++ b/src/main/java/org/spongepowered/api/service/economy/Account.java
@@ -29,6 +29,7 @@ import org.spongepowered.api.entity.player.Player;
 import java.util.UUID;
 
 public interface Account {
+
     /**
      * The string that prefixes all custom account names.
      */
@@ -40,33 +41,27 @@ public interface Account {
     public static final String SERVER_ACCOUNT_NAME = "_ESCSV";
 
     /**
-     * Retrieves the string representative of the owner of
-     * this {@link Account}.
+     * Retrieves the string representative of the owner of this {@link Account}.
      * 
-     * <p>The String returned can represent 3 types of
-     * accounts; namely a Player's, the Server's, or a
-     * customly-registered account.</p>
+     * <p>The String returned can represent 3 types of accounts; namely a
+     * Player's, the Server's, or a customly-registered account.</p>
      * 
-     * <p>If the String returned matches the value denoted
-     * by {@link Account#SERVER_ACCOUNT_NAME}, then the
-     * owner of this Account is the server itself.</p>
+     * <p>If the String returned matches the value denoted by
+     * {@link Account#SERVER_ACCOUNT_NAME}, then the owner of this Account is
+     * the server itself.</p>
      * 
-     * <p>If the String returned begins with the value
-     * denoted by {@link Account#CUSTOM_DATA_PREFIX},
-     * then the owner of this Account is a custom owner,
-     * as set by a plugin.</p>
+     * <p>If the String returned begins with the value denoted by
+     * {@link Account#CUSTOM_DATA_PREFIX}, then the owner of this Account is a
+     * custom owner, as set by a plugin.</p>
      * 
-     * <p>If the String returned does not match either of
-     * the above, then it is expected to be the {@link UUID}
-     * of a {@link Player}, in the form given by
+     * <p>If the String returned does not match either of the above, then it is
+     * expected to be the {@link UUID} of a {@link Player}, in the form given by
      * {@link UUID#toString()}.</p>
      * 
-     * <p>Convenience methods should exist within
-     * implementation of the service; it is not necessary to
-     * check the String on your own.</p>
+     * <p>Convenience methods should exist within implementation of the service;
+     * it is not necessary to check the String on your own.</p>
      * 
-     * @return a String representative of the owner of this
-     *         Account
+     * @return a String representative of the owner of this Account
      * 
      * @see UUID
      * @see Player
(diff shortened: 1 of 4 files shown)


gratimax:
I remember that @Sleaker was toying with an economy service and trying to integrate it with the permissions system using `Context`. Pinging him here so perhaps he can suggest things, or create another PR?

Azatik:
It is something like a Vault?

src/main/java/org/spongepowered/api/service/economy/Account.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param amount a double representing the amount to
+     *        withdraw
+     * 
+     * @return true if this Account's balance turned
+     *         negative
+     * 
+     * @throws IllegalArgumentException if the amount to
+     *         remove is negative
+     */
+    public boolean withdraw(Account destination, double amount);
```
ST-DDT: What is the difference between pay and withdraw?
From the javadocs i only get 'transfer the given amount of money to the target account'.

src/main/java/org/spongepowered/api/service/economy/EconomyService.java
```diff
@@ -0,0 +1,131 @@
...
+     * @return the Account belonging to the server
+     * 
+     * @see Account
+     * @see EconomyService#getAccount(String);
+     */
+    public Account getServerAccount();
+    
+    /**
+     * Returns an immutable list of all {@link Account}s
+     * belonging to a {@link Player}.
```
ST-DDT: Online players only or any existing Accounts for players?

src/main/java/org/spongepowered/api/service/economy/EconomyService.java
```diff
@@ -0,0 +1,131 @@
...
+     * reference.</p>
+     * 
+     * @param owner the owner string to search with
+     * 
+     * @return the Account owned by the given owner, or
+     *         Optional.absent() otherwise
+     * 
+     * @see Account#getOwner()
+     */
+    public Optional<Account> getAccount(String owner);
```
ST-DDT: What is returned if the requested account does not exist yet?
Is it created automatically? If not how do i create an account?

src/main/java/org/spongepowered/api/service/economy/Account.java
Xemiru: account1.pay(account2, 30) -- Account 1 loses 30 while Account 2 receives 30.
account1.withdraw(bank1, 30) -- Account 1 receives 30 while Bank 1 loses 30.

They're essentially convenience methods. They won't check for negative results until the end; so developers should be checking with getBalance().

src/main/java/org/spongepowered/api/service/economy/EconomyService.java
```diff
@@ -0,0 +1,164 @@
...
+     * only. See {@link Account#getOwner()} for
+     * reference.</p>
+     * 
+     * @param owner the owner string to assign the new
+     *        Account to
+     * 
+     * @return whether or not the Account was created
+     *         successfully
+     */
+    public boolean createNewAccount(String owner);
```
simon816: Should return the account. If it may fail, return `Optional<Account>`
Also, methods in interface don't need the `public` access modifier declared explicitly.

Xemiru:
omf 420 lines changed

---
@Azatik yes to some extent; really the difference would be that this isn't a plugin; developers don't need to have a Vault-like plugin as a dependency and instead would reference to this service for interaction with economy.

src/main/java/org/spongepowered/api/service/economy/EconomyService.java
Xemiru: Updated. Includes offline players.

src/main/java/org/spongepowered/api/service/economy/Account.java
raGan262: ```account1.withdraw(bank1, 30) -- Account 1 receives 30 while Bank 1 loses 30.```

What you said is not compatible with documentation.

```Removes the given sum of currency from this Account and sends it to the destination Account.```

File: src/main/java/org/spongepowered/api/service/economy/EconomyService.java [hunk 5]
```diff
@@ -0,0 +1,168 @@
...
+     */
+    Account getServerAccount();
+    
+    /**
+     * Returns an immutable list of all {@link Account}s
+     * belonging to a {@link Player}.
+     * 
+     * <p>This method should try to return all Accounts
+     * registered belonging to Players; including offline
+     * Players.</p>
```
simon816: Just say `belonging to a {@link User}` above, which includes online and offline players.

Sleaker:
Alright lets go over some basic rules for what needs to happen on the API as an implementation to satisfy the majority of people:

1. Amounts need to be either BigDecimal or BigInteger. Doubles are inappropriate here.
2. Accounts need to be tied to Contexts like the Permission API.
3. Elicit and pay are incorrect terms for the API. They should be changed to deposit and withdraw.  These are the terms used when people think about banks and what has been provided in past APIs.
4. There must be a Currency interface and all API methods dealing with amounts should have the currency passed into them.
5. A direct transfer method must be provided. ie. transfer(account1, account2, currency, amount)
6. Documentation must not say that certain methods will throw Exceptions.  API has no constraints on what implementations throw.  If something is undefined, such as withdrawing/depositing negative amounts. Then the API should simply state that operations like this are 'undefined' and different implementations may handle it differently.
7. Return values should be considered for all failure operations rather than exceptions. Such as a class TransactionResult.
8. Accounts need to have more than just a single owner. Owners may not be appropriate. Rather there should be a way to lookup if a player may deposit or withdraw from an account. Some accounts may not have any owners.
9. The basic top-level interface should allow for easily grabbing a player's 'wallet'  ie the account that is directly linked to the player via getWallet(User user).
10. The API should indicate that it is almost never appropriate to use getAllAccounts or this method should not be provided. ie What happens on a server with 100,000 saved accounts in this instance?
11. The API should enforce search features into the account data store so getAllAccounts is not necessary ever.


This isn't exhaustive on what needs to change, the biggest things missing right now are Currencies, and Contexts. Lets start with those and go from there.  This was my draft: https://gist.github.com/Sleaker/27521de9ca4b418b0fe9 - it is missing a few things still, and it doesn't have the Numbers replaced properly. It also didn't take into account Contexts yet, but it's a good start on where we should be heading.

Xemiru:
started making mods based off of your gist that someone linked me to so i made some changes that weren't based off that list just now

BigInteger/BigDouble -- would probably go with bigint since it can also represent the smaller values and still allow for an integer-based currency.

- [x] elicit & pay -> withdraw/deposit
- [x] Currencies -- using Java's Currency class; a default Currency should be set within the 
- [x] direct transfer method
- [x] documentation no longer has @throw tags or warningsEconomyService and one in the Account that returns that default if it doesnt have its own one set
- [ ] transactionresult made, but not completed

I'd still go with the owner system just so we have a "head owner;" I can't really imagine a headless account system. Permission to withdraw or deposit from the account i'd say is left to the implementation of the service.

Top-level has `getPlayerAccount(UUID)`.

- [x] removed getAll methods
- [ ] add account existance check methods

Sleaker:
With no way to poll permissions on accounts there's no way to know if a transaction would result in a success or failure before you attempt it.  This is not desirable as you want to be at least mostly certain that you can perform a specific transaction before it even goes through.

Account ownership is necessary as, in the past, there are massive numbers of plugins that utilize holding values when performing things like auctions or bounties. In the past, the setup had to be done by the admin and sometimes had to be done with a real player account. Forcing accounts to be owned keeps this problem. It's also not a necessity at all when you provide other methods to obtain an account such as through search features with an account id.

src/main/java/org/spongepowered/api/service/economy/Account.java
```diff
@@ -0,0 +1,201 @@
...
+     * Sets the {@link Currency} followed by this {@link Account}.
+     * 
+     * @param currency the Currency instance to follow
+     * 
+     * @see Currency
+     */
+    void setCurrency(Currency currency);
+
+    /**
+     * Returns a double representative of the balance stored within this
```
me4502: Change double to Number

ryantheleach:
I'd like to see a way to chain transactions, if any fail they are all reversed, with the ability to attach custom transactions (think the result of a transaction being successful) as well as some way to transfer items as a transaction.

Sleaker:
@ryantheleach - This is not possible as what you described requires implementation specific cases.  You can't return back an Object off of a transaction that allows another operation to be performed as you don't know what state the transaction ended up being in.  Also I can't see a proper use case for this.  Chained statements are generally done when you want to setup an Object for use without needing to reference the original object over and over. Not when you want to handle multiple disparate transactions that have differing return values.

Sleaker:
@Xemiru - the current way Accounts were setup didn't make much sense. It also wasn't leveraging the Context system similar to Permissions.  I've started reworking what I believe will work the best to provide a way to allow Accounts to be Contextualized, along with necessary Currency support, and the ideas of Wallets.

https://github.com/Sleaker/SpongeAPI/commit/ea25637fe96879e3fed5cdf6558c236a8d89fcfb

Xemiru:
Guess this PR should be closed then; I don't think I should continue trying to improve on this.

MattBDev:
@Xemiru You can close it yourself if you feel that you don't want to continue. You did a great job with the beginnings of this :+1:

Xemiru:
ohicancloseitmyself
oops

PR #663, opened 2015-05-09
Add ClientConnectEvent, for when a player initially connects to the server

simon816:
Fired before `PlayerJoinEvent`, other players will not know the player attempted to join if the event is cancelled.

No longer modelling around `FMLNetworkEvent.ServerConnectionFromClientEvent`, `PlayerConnectEvent` will occur before FML initializes it's handshake.

Renamed to `ClientConnectEvent`

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -32,6 +32,7 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Maps;
 import org.spongepowered.api.Game;
+import org.spongepowered.api.GameProfile;
 import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.block.tileentity.Sign;
 import org.spongepowered.api.block.tileentity.carrier.BrewingStand;
@@ -45,7 +46,6 @@ import org.spongepowered.api.entity.EntityInteractionType;
 import org.spongepowered.api.entity.Tamer;
 import org.spongepowered.api.entity.living.Ageable;
 import org.spongepowered.api.entity.player.Player;
-import org.spongepowered.api.entity.player.User;
 import org.spongepowered.api.entity.player.gamemode.GameMode;
 import org.spongepowered.api.entity.projectile.FishHook;
 import org.spongepowered.api.entity.projectile.Projectile;
@@ -109,11 +109,9 @@ import org.spongepowered.api.event.entity.player.PlayerInteractBlockEvent;
 import org.spongepowered.api.event.entity.player.PlayerInteractEntityEvent;
 import org.spongepowered.api.event.entity.player.PlayerInteractEvent;
 import org.spongepowered.api.event.entity.player.PlayerJoinEvent;
-import org.spongepowered.api.event.entity.player.PlayerLoginEvent;
 import org.spongepowered.api.event.entity.player.PlayerMoveEvent;
 import org.spongepowered.api.event.entity.player.PlayerPickUpItemEvent;
 import org.spongepowered.api.event.entity.player.PlayerPlaceBlockEvent;
-import org.spongepowered.api.event.entity.player.PlayerPreLoginEvent;
 import org.spongepowered.api.event.entity.player.PlayerQuitEvent;
 import org.spongepowered.api.event.entity.player.PlayerRespawnEvent;
 import org.spongepowered.api.event.entity.player.PlayerUpdateEvent;
@@ -123,6 +121,8 @@ import org.spongepowered.api.event.entity.player.fishing.PlayerRetractFishingLin
 import org.spongepowered.api.event.message.CommandEvent;
 import org.spongepowered.api.event.message.CommandSuggestionsEvent;
 import org.spongepowered.api.event.message.MessageEvent;
+import org.spongepowered.api.event.network.GameClientAuthEvent;
+import org.spongepowered.api.event.network.GameClientConnectEvent;
 import org.spongepowered.api.event.rcon.RconLoginEvent;
 import org.spongepowered.api.event.rcon.RconQuitEvent;
 import org.spongepowered.api.event.server.StatusPingEvent;
@@ -176,7 +176,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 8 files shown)


Minecrell:
There should be some way to set a disconnect message when this event get cancelled.

src/main/java/org/spongepowered/api/event/net/PlayerConnectEvent.java
```diff
@@ -0,0 +1,60 @@
...
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.GameEvent;
+import org.spongepowered.api.event.entity.player.PlayerJoinEvent;
+import org.spongepowered.api.text.Text;
+
+/**
+ * Fired when a player attempts to connect to the server, after the success of
+ * the handshake but before the player actually joins the game.
+ *
+ * Cancelling the event will prevent the player from joining and show
```
Minecrell: <p>

simon816:
I have added a method to get the player connection (`PlayerConnection getConnection();`) but the `PlayerConnection` object has a `Player getPlayer()` but the player object may not be ready at that moment in time.
Either I change it to `Optional<Player> getPlayer()`, throw an exception (`IllegalStateException`) or simply remove the back-link method (because you can still do `player.getConnection()` on a player object).
Calling @SpongePowered/developers for thoughts.

src/main/java/org/spongepowered/api/event/net/PlayerConnectEvent.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Fired when a player attempts to connect to the server, after the success of
+ * the handshake but before the player actually joins the game.
+ *
+ * <p>Cancelling the event will prevent the player from joining and show
+ * {@link #getDisconnectMessage} to the player.</p>
+ *
+ * @see PlayerJoinEvent
+ */
+public interface PlayerConnectEvent extends GameEvent, Cancellable {
```
Deamon5550: Any reason you're not extending `PlayerConnectionEvent` ?

simon816: I suppose not, though I don't really understand it's javadoc
> Represents an event that occurs when a {@link PlayerConnection} is being changed.

Deamon5550: Should just be updated to allow it to represent more generic events involving the player's connection.

me4502: If a player is connecting, the connection is being changed anyway.

simon816:
After attempting to implement this, it's rather difficult to cancel the event the way I wanted. When cancelling, I hoped that it would be as though the player never connected but other players still receive a 'player left the game' message. It's possible to use reflection to set the disconnected flag to true but that feels like a bad idea.

What I'd like to know it what should the behaviour be like?
An event like this can be useful to dynamically kick players when joining (e.g. they don't meet the requirements of what a plugin wants).
Perhaps this could be moved earlier in the login process (some point during the handshake).

SonarBeserk:
Is there any reason to not just implement it to work similarly to PlayerLoginEvent like bukkit had? Where you can set the result and messages to use?

simon816:
I have changed the javadoc of the event to state that it occurs earlier in the login process.
> This event fires during the login process when the login state is `READY_TO_ACCEPT`

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -274,7 +274,7 @@
 
     /**
      * Shuts down the server, and kicks all players with the default kick
-     * message (the translation key {@code disconnect.closed}).
+     * message (the literal text "Server closed").
```
Minecrell: Why not just remove this part and just say "with the default kick message"? I see no reason why a plugin or even Sponge itself shouldn't allow modifying the default kick message and it (should) actually get translated depending on the client.
Not completely sure why this is in this PR (seems unrelated), but I guess it's not a big problem

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public interface ClientConnectEvent extends PlayerConnectionEvent, ConnectEvent {
+
+    /**
+     * Gets the disconnect message that will show to the client if the event is
+     * cancelled.
+     *
+     * @return The disconnect message
+     */
+    Text getDisconnectMessage();
```
Minecrell: Maybe this should be changed according to https://github.com/SpongePowered/SpongeAPI/commit/6a85888593c34a83850e35befbec32c4469807e4?

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,74 @@
...
+import org.spongepowered.api.GameProfile;
+import org.spongepowered.api.event.entity.player.PlayerJoinEvent;
+import org.spongepowered.api.text.Text;
+
+/**
+ * Fired when a client (player) attempts to connect to the server.
+ *
+ * <p>This event fires during the login process when the login state is
+ * {@code READY_TO_ACCEPT} - after the token has been verified but before the
+ * connection state switches to 'play'. See
```
Minecrell: If this fires before the connection state switches to `play`, I guess plugin messages sent through the `PlayerConnection` interface could cause problems, as plugin messages are part of the `play` protocol.

src/main/java/org/spongepowered/api/Server.java
simon816: I realized that what I thought was the kick message was actually not, so I just slipped this change in here. Perhaps it should not be defined in the docs.

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
simon816: Yeah I suppose that's true, I'll remove the `PlayerConnection` and replace it with `RemoteConnection`

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
simon816: Unfortunately not, because a `CommandSource` is not available. `MessageEvent` is actually for the other way around.
> Describes events when a CommandSource sends a Text message.

Minecrell: Ah, right...

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
Minecrell: Maybe you could add `ClientConnection` with a `GameProfile getProfile()` directly instead of providing it in the event here, then you can also make PlayerConnection extend ClientConnection.

simon816: I _could_. Depends if any other @SpongePowered/developers agree. I don't think that would really be a useful interface other than for this one event.

src/main/java/org/spongepowered/api/Server.java
Zidane: I'd not define it in the docs.

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
Zidane: I'd say having the RemoteConnection is fine. No need for the interface.

Zidane:
@simon816 

Thinking back, we should do...

MessageEvent (just has get/setNewMessage and getOriginalMessage) -> SinkMessageEvent (dunno name...anyone?) that has the CommandSource and the sink (get/setNewSink and getOriginalSink which would be read-only).

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
simon816: Hmm thinking about it, the `CommandSource` could in fact be the `Server`

simon816:
I have redefined the message events in a second commit on this PR (https://github.com/simon816/SpongeAPI/commit/fcb790aebb3ae9f59d6527aa7f530422336e85b3). Not sure if the changes make it any more useful but feedback/thoughts?

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    void setDisconnectMessage(Text message);
+
+    /**
+     * Gets the currently set reason why the client will be disconnected if the
+     * event is cancelled.
+     *
+     * @return The reason for disconnecting
+     */
+    Cause getDisconnectCause();
```
zml2008: This and the disconnect message should both be `Optional`

zml2008:
It'd be good to have the prelogin event for PEX -- we use that event to cache player data (including fetching from a database, calculating inheritance, etc), which has the potential to lock up the main server thread for a while (but is also needed immediately upon login, for things like a permissions-based whitelist).

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,97 @@
...
+package org.spongepowered.api.event.network;
+
+import org.spongepowered.api.GameProfile;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.entity.player.PlayerJoinEvent;
+import org.spongepowered.api.network.RemoteConnection;
+import org.spongepowered.api.text.Text;
+
+/**
+ * Fired when a client (player) attempts to connect to the server.
```
kashike: This isn't player-specific, all mentions of Player should be removed.

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,97 @@
...
+     * @return The remote connection
+     */
+    RemoteConnection getConnection();
+
+    /**
+     * Gets the profile of the client attempting to connect.
+     *
+     * @return The player's profile
+     */
+    GameProfile getProfile();
```
kashike: This should be `Optional<GameProfile>`.

src/main/java/org/spongepowered/api/event/network/ClientConnectEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ * <p>The server may have cancelled the event if the player is banned or
+ * whitelisted. Be sure to set {@code ignoreCancelled=false} in the
+ * {@code @Subscribe} annotation to receive the event in this case.</p>
+ *
+ * <p>Cancelling the event will prevent the player from joining and show
+ * {@link #getDisconnectMessage} to the player.</p>
+ *
+ * @see PlayerJoinEvent
+ */
+public interface ClientConnectEvent extends ConnectEvent {
```
kashike: GameClientConnectEvent

mrgoodrich:
Quick question, how long until this will be usable in plugin development?  And how do I tell this for the future?  As I can see, this was merged, but when I update my maven dependencies and try to add it to my plugin, it's not recognized.  Sorry for the noob question.. :frowning:

twizmwazin:
@mrgoodrich Make sure you are using version 2.1-SNAPSHOT

mrgoodrich:
@twizmwazin Are you sure it's been added yet?  I'm using version 2.1-SNAPSHOT and even tried making a new project to test, and neither of them recognized the ClientConnectEvent.

simon816:
It was renamed to `GameClientConnectEvent` (FQN `org.spongepowered.api.event.network.GameClientConnectEvent`)

mrgoodrich:
Yeah all I'm finding in the network folder with the version 2.1-SNAPSHOT is package-info, PlayerConnectionEvent, PlayerConnectionRegisterChannelEvent, and PlayerConnectionUnregisterChannelEvent.  Is it possible that this hasn't been added yet?  Or am I screwing something up? @simon816

simon816:
The repo is updated, although it looks like the front page (http://repo.spongepowered.org/maven/org/spongepowered/spongeapi/) isn't.

I created a build.gradle in lightning speed, ran `gradle eclipse --refresh-dependencies` and the classes were there
```groovy
apply plugin: 'java'
apply plugin: 'eclipse'

repositories {
    maven {
        name = 'sponge'
        url = 'http://repo.spongepowered.org/maven'
    }
}

dependencies {
    compile 'org.spongepowered:spongeapi:2.1-SNAPSHOT'
}
```

mrgoodrich:
I'm not sure what was happening, but I switched over to Gradle, and now it's working!  Thanks @twizmwazin and @simon816 !

PR #683, opened 2015-05-16
Added RayTracingUtil - Fixes #365

### LINKED ISSUE #365: Get Coordinates of the target Player.
Get Coordinates of the target Player. There are similar in Bukkit:
Block block = player.getTargetBlock(null, 100);

ST-DDT:
Some time ago a I read the suggestion of a ray tracing / lineOfSight method for world in this issue: https://github.com/SpongePowered/SpongeAPI/issues/365

This PR adds very abstract methods to the API, that work almost independently from the rest of the API.

I first tried to implement all 4 methods suggested in that issue, but after a while i dropped that because you can implement all those methods in just a single one using some kind of callback.

**Contents**
- `RayTracingUtil`
  - `void trace(Location loc, Vector3d dir, BlockConsumer con)`
  - `void trace(Extent world, Vector3d loc, Vector3d dir, BlockConsumer con)`
  - `void trace(Vector3d loc, Vector3d dir, PositionConsumer con)` // Main method
- `PositionConsumer` (callback)
  - `boolean apply(double posX, double posY, double posZ, Vector3d rayDirection)` // Stop on false
- `BlockConsumer` (callback)
  - `boolean apply(Location location, Direction faceDirection)` // Stop on false
- `PostionToBlockConsumer`
  - adapter: `BlockConsumer` -> `PositionConsumer`

**Example**

The following code shows the first 100 blocks the ray tracing went through

````java
RayTracingUtil.trace(randomLocation, randomDirection, new TestPositionConsumer());

class TestPositionConsumer implements PositionConsumer {

    private static final DecimalFormat df = new DecimalFormat("0.000");
    private static final int startTtl = 100; 
    // For testing I used startTtl = 1kk with randomized start and direction vectors
    private int ttl = startTtl;

    @Override
    public boolean apply(double posX, double posY, double posZ, Vector3d direction) {
        System.out.println("Pos: " + df.format(posX) + " , " + df.format(posY) + " , " + df.format(posZ));
        if (posX % 1 != 0 && posY % 1 != 0 && posZ % 1 != 0 && this.ttl != startTtl) {
            // Should never happen, no coordinate hits the boundaries of a block
            throw new IllegalArgumentException("Bad Position: " + df.format(posX) + " - " + df.format(posY) + " - " + df.format(posZ));
        }
        return this.ttl-- > 0; // Abort ray trace if false
    }
}
````

Which results in

````text
Pos:  0.214 , -0.359 ,  0.307 // Start location
Pos:  0.000 , -0.041 ,  0.065
Pos: -0.028 ,  0.000 ,  0.033
Pos: -0.057 ,  0.043 ,  0.000
Pos: -0.703 ,  1.000 , -0.731
Pos: -0.941 ,  1.352 , -1.000
Pos: -1.000 ,  1.441 , -1.067
Pos: -1.378 ,  2.000 , -1.495
Pos: -1.824 ,  2.662 , -2.000
Pos: -2.000 ,  2.923 , -2.199
Pos: -2.052 ,  3.000 , -2.258
Pos: -2.708 ,  3.971 , -3.000
[..]
````

**TBD**

There are missing some default abort conditions, I started writing some, but ended up with to much, so i would like to ask which one i should add.

### CODE REVIEW

**Comment by Deamon5550** [comment]
As I stated in the original issue, the only instance under which I would agree that adding a ray tracing utility is not unnecessary bloat is if the provided utility can offer time, memory, or complexity increases over what is possible to be made within a plugin. This is not the case with your proposal.

At a minimum our provided utility would need to offer options for taking the bounding boxes of blocks into account.

**Comment by DDoS** [comment]
This should also let you choose between hitting all the blocks or one per column.

File: src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+            return distanceLeft / absDirection;
+        } else {
+            return (1 - distanceLeft) / absDirection;
+        }
+    }
+
+    // Sanitize the location (remove double fragments)
+    private static double sanitize(double d) {
+        final double dd = Math.abs(d % 1);
+        if ((dd != 0 && dd < 0.0000001) || (dd != 1 && dd > 0.9999999)) {
```
DDoS: You can use GenericMath.DBL_EPSILON here.

ST-DDT:
@Deamon5550 Since there is no method in the API to get the bounding box of a block, i cannot check for it. But if someone adds some kind of functionality to the APi this can easily be used in the ray tracing using the PositionConsumer / a modified BlockConsumer.

@DDoS Sorry i don't understand that.
It is already possible to get all the blocks in the line of sight or to only get the last one in that line.

Deamon5550:
This doesn't have to me implemented in the API, in fact it almost certainly shouldn't be in order to take advantage of things like bounding boxes. If you could get bounding boxes through the API then there would be even less use for having this utility.

ST-DDT:
So you want me to add a function call that wraps a `PositionConsumer` in a boundig box checker similar to [`RayTracingUtil.POSITION_LIMITER`](https://github.com/SpongePowered/SpongeAPI/pull/683/files#diff-0ca62276b0f471ab453e7a8e60225110R40)

DDoS:
@ST-DDT 

<pre>
^


</pre>

ST-DDT:
>@ST-DDT
>````
>^
>
>````

@DDoS I still don't get it.

src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java
ST-DDT: After some testing i need more than `1 * GenericMath.DBL_EPSILON`, `2 *  GenericMath.DBL_EPSILON` works perfectly fine.
Shall i use my own constant for that or shall i let him calculate it every run?
````java
if ((dd != 0 && dd < GenericMath.DBL_EPSILON * 2) || (dd != 1 && dd > 1 - GenericMath.DBL_EPSILON * 2)) {
````

Comment (DDoS) [inline src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java]: The field is constant, so it will probably be pre-computed automatically. It does depend on the compiler/JVM.

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java]: With oracle jdk8 it is still the same after compilation. 
I guess Java Hot Spot Compiler does it at runtime. (Or at least i hope it does so)

**Comment by ST-DDT** [comment]
@DDoS Thanks for your detailed explanations. I now get what you want.

But what happens if the abort condition would match one of the ignored blocks?

Here is an example column filter:
````java
/**
 * Fitlers all calls to only a single entry per
 */
public class ColumnFilter implements PositionConsumer {

    private final PositionConsumer consumer;
    private final Axis axis;
    private int last = Integer.MAX_VALUE;

    public ColumnFilter(PositionConsumer consumer, Vector3d direction) {
        super();
        this.consumer = checkNotNull(consumer, "consumer");
        this.axis = Axis.getClosest(checkNotNull(direction, "direction"));
    }

    public ColumnFilter(PositionConsumer consumer, Axis axis) {
        super();
        this.consumer = checkNotNull(consumer, "consumer");
        this.axis = checkNotNull(axis, "axis");
    }

    @Override
    public boolean apply(double posX, double posY, double posZ, Vector3d rayDirection) {
        boolean forward = true;
        switch (this.axis) {
            case X:
                forward = checkAxis(posX);
                break;
            case Y:
                forward = checkAxis(posY);
                break;
            case Z:
                forward = checkAxis(posZ);
                break;
            default:
                throw new IllegalArgumentException("Unsupported axis " + this.axis);
        }
        return !forward || this.consumer.apply(posX, posY, posZ, rayDirection);

    }

    private boolean checkAxis(double doublePos) {
        int intPos = (int) doublePos;
        if (intPos == this.last) {
            return false;
        }
        this.last = intPos;
        return true;
    }

}
````

for `BlockConsumers` this is a lot easier to implement ( because you only have to check for the faceDirection..

**Comment by ST-DDT** [comment]
@DDoS I improved the VisitedBlockConsumer so that you can insert a filter condition:
[VisitedBlockConsumer](https://github.com/SpongePowered/SpongeAPI/pull/683/files#diff-17fea08351a0ca2f5ec9079602e0c244R55)
[BlockSearchConditions](https://github.com/SpongePowered/SpongeAPI/pull/683/files#diff-25c865936a01750c25f180e52bb8af5aR148)

**Comment by ST-DDT** [comment]
@Deamon5550 I added some methods/fields to allow the server to check for solid and interact boxes.
[`RayTracingUtil`](https://github.com/SpongePowered/SpongeAPI/pull/683/files#diff-0ca62276b0f471ab453e7a8e60225110R40)

**Comment by modwizcode** [comment]
@Deamon5550 I disagree with the need for a built-in raytracing utility to be better than a self implemented one, most plugin developers I'm guessing will not know how to implement raytracing and it's a substantially useful feature.

**Comment by modwizcode** [comment]
@ST-DDT Can you post the results along with start position AND direction?

File: src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java [hunk 2]
```diff
@@ -0,0 +1,160 @@
...
+        if (min > moves[1]) {
+            min = moves[1];
+        }
+        if (min > moves[2]) {
+            min = moves[2];
+        }
+        return min;
+    }
+
+    // times how often you have to apply the given direction to pass any borders
```
modwizcode: This needs more explanation.

src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java
```diff
@@ -0,0 +1,160 @@
...
+            distances[1] = distance(posY, dy);
+            distances[2] = distance(posZ, dz);
+            final double min = min(distances);
+            posX = sanitize(posX + min * dx);
+            posY = sanitize(posY + min * dy);
+            posZ = sanitize(posZ + min * dz);
+        }
+    }
+
+    // The minimum distance
```
modwizcode: This also needs more detailed information on what it's doing.

src/main/java/org/spongepowered/api/util/raytrace/RayTracingUtil.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return min;
+    }
+
+    // times how often you have to apply the given direction to pass any borders
+    private static double distance(double position, double direction) {
+        if (direction == 0) {
+            return Double.MAX_VALUE;
+        }
+        final double distanceLeft = (position % 1 + 1) % 1;
```
modwizcode: Because this method deals with directions, distanceLeft should be renamed or commented to diversify it from the direction left. Perhaps rename it to "distanceRemaining"

ST-DDT:
@modwizcode 
````text
Start: -0.381 , -0.910 , 0.519
Direction: -0.548 , -0.692 , 0.691
Pos: -0.381 , -0.910 , 0.519
Pos: -0.452 , -1.000 , 0.609
Pos: -0.762 , -1.392 , 1.000
Pos: -1.000 , -1.692 , 1.300
Pos: -1.244 , -2.000 , 1.607
Pos: -1.556 , -2.394 , 2.000
Pos: -2.000 , -2.955 , 2.560
Pos: -2.035 , -3.000 , 2.605
Pos: -2.349 , -3.396 , 3.000
Pos: -2.827 , -4.000 , 3.602
Pos: -3.000 , -4.219 , 3.821
````

DDoS:
The BlockSearchConditions file needs documentation.

ST-DDT:
@modwizcode I added some more documentation, does this meet your requirements now?

@DDoS 

> Also, the abort condition should ignore the ignored blocks. If it's not part of the trace, you shouldn't be aborting on it.

AFAIK the `AbortBlockConsumer` does not forward the call to the consumer, if the abort condition is met. `false && method()` will never execute method.

modwizcode:
Correct ```false && method()``` will short-circuit the method() call because of the false.

Zidane:
Closing this in favor of https://github.com/SpongePowered/SpongeAPI/pull/687

Thanks @ST-DDT for the initial work, we'll finish up on the other PR for the merge.

PR #685, opened 2015-05-17
Value API Additions to Data API

gratimax:
Values API PR | [Data 2.0 Implementation PR](https://github.com/SpongePowered/SpongeCommon/pull/94)

PLEASE NOTE THAT @gabizou IS HANDLING THIS PR
=======

Having heard complaints and grievances over the usage of the original Data API, many changes were made, but for the sake of discussion, here
are a few of the complaints:

- What used to be a single setter method for something so simple is now overly complex with retrieving `DataManipulator`s and offering them back
- Unless I look at implementation, I can't find out what `DataQuery` I need to use to get the *author* from a `DataContainer` of a serialized `AuthorData`
- I can't simply request a single value from any `DataHolder` without going through the appropriate `DataManipulator`

And these points were all valid, as the first iteration of Data API was simply aimed at migrating everything data access-wise to using `DataManipulator`s and not the previously used hierarchical interfaces declaring specific optional accessor methods.

So, without further ado, I present Data API 2.0!

##### What's *new*?

The actual **new** things:

 - [`BaseValue`] object that wraps around an element type, and a [`Key`]
 - [`Key`]s present themselves as identifiable keys to retrieve a particular type of data from any [`ValueContainer`]
 - [`ValueContainer`]s are quite literally containers of [`BaseValue`]s
 - [`CompositeValueStore`] is a mutable version of a [`ValueContainer`] that allows getting and setting [`BaseValue`]s while the mutations are all transactional
 - [`ImmutableValueStore`] is an immutable variant of the [`ValueContainer`] 
 - [`ImmutableDataManipulator`] is an immutable variant and self translating of a [`DataManipulator`]
 
The changes:

 - [`DataHolder`]s are now extending [`CompositeValueStore`] as they are able to be queried for not only [`DataManipulator`]s but also [`Value`]s
 - [`DataManipulator`]s are able to call [`#asImmutable()`] to return an [`ImmutableDataManipulator`] variant of itself.
 - [`DataManipulator`]s no longer hve accessor methods to the data of which they contain, they have [`Value`] getters and can be mutated with [`#set(Key,Object)`]
 - [`Location`]s and [`Extent`]s no longer contain property specific accessor methods, delegating to the property system of the Data API
 
##### Holy smokes batman! That makes no real sense to me as to what ACTUALLY CHANGED!

Well, you didn't really give me time to explain, but for the sake of discussion, I'll give you an example:

In the previous Data API, we would have to do the following to change the health of an entity:

``` java

final Entity entity;
final Optional<HealthData> optional = entity.getData(HealthData.class);
if (optional.isPresent) {
  entity.offer(optional.get().setHealth(10D));
}
// continue doing other things


```

And what it looks like you can do now is:

``` java

final Entity entity;
entity.offer(Keys.HEALTH, 10D);


```

Really simple, right? Right! But what does this actually mean? First, we saw that we have a similar `#offer` method, but what is this `Keys` thing comming from?? Unfortunately, I'm not going to state what `Keys` is before I explain what a [`Key`] actually does:

```
A Key is similiar to a UUID in that it is uniquely identifying a piece of data. The 
difference is that a *key* contains a nice method to retrieve the suggested DataQuery
that the piece of data should reside when serialized to a DataContainer.

```

From that description, a [`Key`] could have potential for being a `CatalogType`, but we don't want to actually require querying for a `Key` through the `GameRegistry`. Therefor, this is where [`Keys`] comes into play: It is an enumeration of vanilla supported `Key`s that Sponge API knows about!

##### Ok, so what about `Value`?

Ok ok ok... I'm guessing you understood that you could use a `Key` to retrieve a specific value of information, or alternatively set a specific value of information onto a [`DataHolder`]. Well, it's nice being able to do that, but what about sharing this information with other people? Say, for an `Event` or simply just a `DataTransactionResult`? This lies in where `Value` really comes from.

A [`Value`] is a pairing of an actual value of data with a `Key`. Some `Value`s are more complex or require special treatment, such as the [`CollectionValue`]. While it's nice being able to `get` and `set` an entire `Collection`, sometimes it just might be easier to `add` or `remove` from said collection. With this comprisal of simple and complex `Value`s, we can surmise that effectively all information previously handled by `DataManipulator`s can be handled by these [`Value`]s.

##### And what about immutability? I read that somewhere above in the description.

Man... you really are jumping around, aren't you?

In Computer Science, there's a concept of *immutable objects*. To [summarize], an immutable object can not be changed/modified after it is created. *Mutable* objects CAN be modified after they are created. 

In this PR, we bring the concept of [`ImmutableValue`]s and [`ImmutableDataManipulator`]s to the API. The point of an `ImmutableValue` is the same as a [`Value`], with the primary exception that an `ImmutableValue` provides no mutators, but rather method calls for retrieving a *new* `ImmutableValue` with the provided information. The same thing can be said about [`ImmutableDataManipulator`]s. However, to improve usability even further, I've added method calls to retrieve an *immutable* instance from a *mutable* instance, and vice versa. 

The difference between retrieving an `ImmutableValue` from a `Value` and retrieving an [`ImmutableDataManipulator`] from a [`DataManipulator`] is that the manipulators keep type safety between each other. In other words, they refer to the types of each other.

The primary advantage of having the *immutable* concept is that with `Event`s, it is now possible to simply pass around an `ImmutableDataManipulator` or an `ImmutableValue` without worrying about plugin developers mutating your value/manipulator! 
The other advantage to this is that [`DataTransactionResult`]s now solely work around `ImmutableValue`s. Being able to therefor save, restore, and [undo] is one of the benefits of the additions through values.

##### Well, what else did you change?

Overall, I'd have to show you the new methods added to [`CompositeValueStore`] and [`ValueContainer`]:

Let's start with `ValueContainer`
```java

public interface ValueContainer<C extends ValueContainer<C>> {

  <E> Optional<E> get(Key<? extends BaseValue<E>> key);
  
  <E> E getOrNull(Key<? extends BaseValue<E>> key);
  
  <E> E getOrElse(Key<? extends BaseValue<E>> key, E defaultValue);
  
  <E, V extends BaseValue<E>> Optional<V> getValue(Key<V> key);
  
  boolean supports(Key<?> key);
  
  boolean supports(BaseValue<?> baseValue);
  
  C copy();
  
  ImmutableSet<Key<?>> getKeys();
  
  ImmutableSet<ImmutableValue<?>> getValues();

}

```

For starters, we see the aforementioned `get(Key)` method that allows us to retrieve a familiar [Optional] of the underlying value type based on the `Key`. This is where we can perform
```java

final Optional<Double> healthOptional = entity.get(Keys.HEALTH);
```

And we're not only positive about type safety, but we're also positive that if the `Optional#isPresent()` returns `true`, the `Entity` must have *health*!

But what about the other two methods? What does `getOrNull(Key)` give us? In short, it allows for a more traditional (read legacy) form of data handling such as the following
```java
final Text displayName = entity.getOrNull(Keys.DISPLAY_NAME);
if (displayName != null) {
  // do things with display names
}
```
And similarly for `getOrElse(Key)`
```java
final Text displayName = entity.getOrElse(Keys.DISPLAY_NAME, Texts.of("I have no default name!"));
```
Now the actually interesting method that *actually* uses Values is `getValue(Key)`. With `getValue`, we can retrieve the Value prescribed for the `Key`, and not only that, but manipulate the underlying value as needed, without having to ever leave the safety of the `Value` itself. For example:
```java
if (entity.supports(Keys.POTION_EFFECTS) {
  final ListValue<PotionEffect> potionValue = entity.getValue(Keys.POTION_EFFECTS).get();
  entity.set(potionValue.removeAll(new Predicate<PotionEffect>() {
    @Override
    public boolean apply(PotionEffect potionEffect) {
      return potionEffect.getType() == PotionEffectTypes.SPEED;
    }
  }).add(myCustomPotionEffect));
}
```

Note that never once did we have to assign a `List<PotionEffect>` or call any external methods short of those found in [`ListValue`], [`CollectionValue`], and [`Value`]. 

Now, let us continue on with [`CompositeValueStore`]:

A `CompositeValueStore` is meant to be a *mutable* `ValueContainer` with methods declared to modify, add, or replace `Value`s pre-existing on said `CompositeValueStore`. Here, let's take a look at the interafce really quickly:
```java
public interface CompositeValueStore<S extends CompositeValueStore<S, H>, H extends ValueContainer<?>> extends ValueContainer<S> {
  <T extends H> Optional<T> get(Class<T> containerClass);
  
  <T extends H> Optional<T> getOrCreate(Class<T> containerClass);
  
  boolean supports(Class<? extends H> containerclass);
  
  <E> DataTransactionResult transform(Key<? extends BaseValue<E>> key, Function<E, E> transformingFunction);
  
  <E> DataTransactionResult offer(Key<? extends BaseValue<E>> key, E value);
  
  DataTransactionResult offer(BaseValue<?> value);
  
  DataTransactionResult offer(H valueContainer);
  
  DataTransactionResult offer(H valueContainer, MergeFunction function);
  
  DataTransactionResult offer(Iterable<H> valueContainers);
  
  DataTransactionResult offer(Iterable<H> valueContainers, MergeFunction function);
  
  DataTransactionResult remove(Class<? extends H> containerclass);
  
  DataTransactionResult remove(BaseValue<?> value);
  
  DataTransactionResult remove(Key<?> key);
  
  DataTransactionResult undo(DataTransactionResult result);
  
  DataTransactionResult copyFrom(S that);
  
  DataTransactionResult copyFrom(S that, MergeFunction function);
  
  Collection<H> getContainers();
```

##### Holy cow! Why so many methods? What is H? Why is CompositeValueStore self referencing?? WHAT IS THIS MAGIC?!

*Love*...

No, seriously though. The methods declared in `CompositeValueStore` are required for achieving full mutability in almost all use cases. The point of the self referencing generic `S extends CompositeValueStore<S>` is so that whatever extends `CompositeValueStore` will return itself when calling [`#copy()`]. The second generic, `H extends ValueContainer<?>` is for further defining the *type* of `ValueContainer` the various getter and setter methods will allow. This is useful for defining [`DataHolder`] to define that it can only accept [`DataManipulator`]s and anything extending it thereof. 

Now, there's something you may have noticed that is new, *there's no more DataPriority*! What replaced it is [`MergeFunction`].
A `MergeFunction` only handles `ValueContainer`s, and more  specifically, in the case of `offer(Iterable<H>,MergeFunction`, wildcarded `ValueContainer`s. The point of it is that when you have a `DataManipulator` such as `CommandData`, there's **a lot** of data to be replaced, and potentially, data to be merged. The previously mentioned `DataPriority` aimed to simplify merging, but it *over simplified*. What `MergeFunction` aims to do is allow for any, or some specific `DataManipulator`s to be handled in a merge process, such that complex custom merges can take place within the `MergeFunction` itself.

The rest of the methods are self explanatory at this point.

##### Ok... so now what? Any more changes?

Minor changes were made here and there, to summarize them:
- [`LocationCompositeValueStore`] is a delegate to all Data API related methods that `Location` proxies back onto
- [`Cycleable`] allows for catalogtype cycling of values, so, you can cycle `Direction`s, or `Axis`, etc. This isn't meant for cycling between possible values such as `DyeColor`s
- [`PropertyStore`] is for external retrieval of properties such that they can be dictated by external sources that SpongeAPI implementations do not provide
- And other minor data related changes. 

##### Wow... that's some changelog. What precisely do you plan on having in the future for Data API?

All I plan on doing with Data API is adding onto the API by introducing new `Key`s, `DataManipulator`s, and a few new `CatalogType`s or so. Until someone finds a problem that can be solved in a much cleaner fashion, I don't predict making any large changes to Data API. 

A final demonstration of a plugin working with the new Data API:
```java
package org.spongepowered.cookbook.plugin;

import org.spongepowered.api.data.key.Keys;
import org.spongepowered.api.data.manipulator.mutable.entity.HealthData;
import org.spongepowered.api.entity.EntityTypes;
import org.spongepowered.api.event.Subscribe;
import org.spongepowered.api.event.entity.EntitySpawnEvent;
import org.spongepowered.api.event.entity.player.PlayerJoinEvent;
import org.spongepowered.api.text.Texts;
import org.spongepowered.api.text.format.TextColors;

@Plugin(id = "DataTest", name = "DataTest", version = "0.1")
public class DataTest {

    @Subscribe
    public void onJoin(PlayerJoinEvent event) {
        final HealthData healthData = event.getEntity().getHealthData();
        event.getEntity().offer(healthData.set(Keys.MAX_HEALTH, 400D).set(Keys.HEALTH, 300D));
        event.getEntity().sendMessage(event.getEntity().get(Keys.DISPLAY_NAME).get());
    }

    @Subscribe
    public void onSpawn(EntitySpawnEvent event) {
        if (event.getEntity().getType().equals(EntityTypes.DROPPED_ITEM)) {
            event.getEntity().offer(Keys.DISPLAY_NAME, Texts.of(TextColors.GOLD, "Sponge", TextColors.WHITE, "Powered"));
            event.getEntity().offer(Keys.SHOWS_DISPLAY_NAME, true);
        }
    }
}
```

As an added note: If you're STILL wanting to use an NBT API, there are Nbt utilities freely available and you can readily translate them to `DataContainer`s through the provided [`DataTranslator`] interface (you extend it and implement it however you need to).

[`BaseValue`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/BaseValue.java
[`Key`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/key/Key.java
[`ValueContainer`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/ValueContainer.java
[`CompositeValueStore`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/mutable/CompositeValueStore.java#L47
[`ImmutableValueStore`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/immutable/ImmutableValueStore.java
[`DataManipulator`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/manipulator/DataManipulator.java
[`ImmutableDataManipulator`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/manipulator/ImmutableDataManipulator.java
[`DataHolder`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/DataHolder.java
[`#asImmutable()`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/manipulator/DataManipulator.java#L174
[`#set(Key,Object)`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/manipulator/DataManipulator.java#L117
[`Location`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/world/Location.java
[`Extent`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/world/extent/Extent.java
[`Value`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/mutable/Value.java
[UUID]:http://docs.oracle.com/javase/7/docs/api/java/util/UUID.html
[DataQuery]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/DataQuery.java
[`CollectionValue`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/mutable/CollectionValue.java
[summarize]:https://en.wikipedia.org/wiki/Immutable_object
[undo]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/mutable/CompositeValueStore.java#L288
[`DataTransactionResult`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/DataTransactionResult.java
[Optional]:http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Optional.html
[`ListValue`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/mutable/ListValue.java
[`#copy()`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/ValueContainer.java#L126
[`MergeFunction`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/merge/MergeFunction.java
[`DataTranslator`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/translator/DataTranslator.java
[`LocationCompositeValueStore`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/world/extent/LocationCompositeValueStore.java
[`Cycleable`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/util/Cycleable.java
[`PropertyStore`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/property/PropertyStore.java
[`ImmutableValue`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/value/immutable/ImmutableValue.java
[`Keys`]:https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/data/key/Keys.java


ST-DDT:
Should this be some kind of view like an editable+live column selection in a database?
First you have the entire creature, then only the health part and in the end you obtain the current health data? 
And in order to get/set/change the health you can use all three entries with the same result?
Wasn't the main point of the data API to avoid internal references all over the implementation?

gratimax:
@ST-DDT Yes, there are internal references to DataObjects within DataProjections. However, this is as far as they go, meaning that the actual use of internally-linked objects is extremely limited. For most use cases, data is set by-value. DataObject.copyOf can be used as the current manipulator implementation is used for storing a reference to a certain amount of values that is separate from the originator object. The restrict method exists as a convenience, but it is possible to remove it if it turns out to promote bad practice.

src/main/java/org/spongepowered/api/data/value/Value.java
```diff
@@ -0,0 +1,42 @@
...
+    E get();
+
+    S set(E value);
+
+    @Override
+    Value<E, S> replace(E value);
+
+    Key<E> getKey();
+
+    ValueStore<S> getStore();
```
gabizou: Shouldn't this be just returning type `S`?

src/main/java/org/spongepowered/api/data/value/ValueStore.java
```diff
@@ -0,0 +1,67 @@
...
+
+    <E> Value<E, S> tryBind(Key<E> key);
+
+    boolean supports(Key<?> key);
+
+    boolean canMutate();
+
+    S copy();
+
+    S copyTo(ValueStore<?> that);
```
gabizou: Either have copyTo or copyFrom. Don't have both because it's just silly.

src/main/java/org/spongepowered/api/data/value/ValueStore.java
```diff
@@ -0,0 +1,67 @@
...
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.merge.MergeStrategy;
+
+import java.util.Set;
+
+public interface ValueStore<S extends ValueStore<S>> {
+
+    <E> Optional<E> get(Key<E> key);
+
+    <E> E tryGet(Key<E> key) throws UnsupportedOperationException;
```
gabizou: Could instead getOrNull. but again, as some people have complained before, we're just replicating the functionality of `Optional` with these methods.

src/main/java/org/spongepowered/api/data/value/ValueStore.java
gratimax: They function slightly differently

src/main/java/org/spongepowered/api/data/value/Value.java
gratimax: Whoops!

src/main/java/org/spongepowered/api/data/value/ValueStore.java
gratimax: It's not really supposed to do the same as getOrNull. This is for the case where we've already checked if a value is supported and are 99.99% sure that it's there, like so:

```java
if (entity.supports(Values.HEALTH)) {
   entity.tryGet(Values.HEALTH);
}
```

If an UnsupportedOperationException is thrown with a good enough error message then we know that either something is gravely wrong or that the user has used tryGet in the wrong way.

src/main/java/org/spongepowered/api/data/value/ValueStore.java
Deamon5550: You're diagrams don't really make things clearer, by your diagrams `Foo.copyTo(Bar)` could still be replaced with `Bar.copyFrom(Foo)`.

gratimax: Well yeah, but the reason there's two methods here is for chaining. `foo.copyTo(bar)` returns `foo` and `bar.copyFrom(foo)` returns `bar`. Depending on the usage a developer would either want to chain from the origin instance or the destination instance. This does not hold true for immutable value stores however, the copied store is always returned.

src/main/java/org/spongepowered/api/data/value/Value.java
```diff
@@ -0,0 +1,42 @@
...
+public interface Value<E, S extends ValueStore<S>> extends BaseValue<E> {
+
+    E get();
+
+    S set(E value);
+
+    @Override
+    Value<E, S> replace(E value);
+
+    Key<E> getKey();
```
gabizou: If `Key` is just an empty marker interface with a method `getClass(): Class<E>`, why not just use `Class<E>`?

gratimax: Values are intrinsically connected to the key you can access them with. Keys are not "empty marker interfaces", they actually represent a way to lookup data within a value store.

gabizou:
I have quite a few changes I will push when I arrive in the UK, I'll make note of these comments though.

gabizou:
@Zidane I also started writing the documentation on the implementation for ValueProcessors and the newly improved DataProcessors.

Zidane:
@gabizou 

https://forums.spongepowered.org/t/checking-if-new-player/7661/7?u=zidane

This post got me thinking. Some data shouldn't be allowed to be "created" by the API.

For example, say I listened to PlayerJoinEvent at Order.FIRST and did getOrCreate(JoinData.class)...now any other plugin who listens to this event afterwards will incorrectly think that the player has played before when, in reality, they haven't.

So two ways to handle this I think:

A. Remove the concept of data being present = some status (such as having joined before). Make all data containers which means JoinData gets hasPlayedBefore and InvisiblityData gets isInvisible.
B. Introduce concept of data not being creatable (as we did with some data not being removable). This would violate the contract of getOrCreate...

I don't like either solution but if I made a final call right this moment, I would go with A .

gabizou:
@Zidane You do realize `getOrCreate()` doesn't attach the information to the `DataHolder` to begin with, so it very well is possible to check that a player is in fact joining for the first time by calling `get(PlayerData.class).isPresent()` because again, `getOrCreate` NEVER creates the `DataManipulator` and associates the "default" data presented back onto the `DataHolder`. I honestly don't see how this is an issue.

Zidane:
@gabizou 

Unsure how I forgot that create doesn't add it back on the DM. Nvm on this then...except for the fact that it can be confusing. I can honestly say that I wouldn't expect to know that if the data isn't present it means they haven't played before. Needs to be a way to make this more clear or have JoinData be always present and add the boolean for playedBefore.

gabizou:
> Needs to be a way to make this more clear or have JoinData be always present and add the boolean for playedBefore.

Or, we add a method to the event to simplify the check.

src/main/java/org/spongepowered/api/data/DataQuery.java
```diff
@@ -75,6 +77,16 @@ public DataQuery(List<String> parts) {
     }
 
     /**
+     * Gets an empty {@link DataQuery}. This query performs exactly the same
+     * when used in all aspects as
```
simon816: as... what?

src/main/java/org/spongepowered/api/data/DataTransactionBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    public static DataTransactionBuilder builder() {
+        return new DataTransactionBuilder();
+    }
+
+    /**
+     * Creates a {@link DataTransactionResult} with no data successfully added,
+     * removed, or rejected, and with the
+     * {@link org.spongepowered.api.data.DataTransactionResult.Type} of
+     * {@link org.spongepowered.api.data.DataTransactionResult.Type#SUCCESS}
```
simon816: Fully qualifying makes this more difficult to read.

gabizou: > Fully qualifying makes this more difficult to read.

Yes, but absolutely necessary since one of the JDK's doesn't locate the linked enum value correctly.

JBYoshi:
@gratimax So data manipulators are now just a set of related values?

gabizou:
> @gratimax So data manipulators are now just a set of related values?

Short answer, yes. Long answer, still yes, a `DataManipulator` doesn't change from what it functioned previously, an object with a few fields that can be retrieved and manipulated. The change now is that a `DataManipulator` can represent these fields as `Value`s. Their function to collect a bunch of `DataManipulator`s for later use with chucking them back at various `DataHolder`s and or use them for `Populator`s.

ST-DDT:
@gabizou @gratimax May I create a PR for https://github.com/SpongePowered/SpongeAPI/issues/710?
I ask because I don't want to collide with your (far more important) work.
I plan to update the javadocs of 
* CatalogType
* CatalogedBy

AFAIK you did not modify any of those classes yet.

As an alternative: I could also post the new javadocs here so you can integrate them in your PR directly.

gabizou:
> @gabizou @gratimax May I create a PR for #710?

I don't see why anything stops you from making a PR, but I can safely say I don't have time this week (or more) to be reviewing any PR's except for Data. If there are javadocs I should add in, do comment for them to be added.

gabizou:
Just pinging everyone watching as the PR is marked as ready.

@Zidane Implementation has been tested with a plugin specifically written to interact with the new Data API. The plugin code should be in the API PR description.

PR #687, opened 2015-05-17
[WIP] BlockRay, BlockRayFilter

gratimax:
This is intended as an alternative to #683.

I thought that while #683 was a good start to a ray tracing utility, I wanted to try to keep as much functionality yet keeping the added surface area as simple as possible.

This PR adds two new classes:

- BlockRay, which will eventually contain the ray tracing algorithm
- BlockRayFilter, which is essentially a predicate that takes a Vector3d, but takes doubles in order to prevent object churn

Here's some example usage:

```java
Player myPlayer;

BlockRayFilter max30 = BlockRayFilter.ONLY_AIR.and(BlockRayFilter.maxDistance(30));

Optional<Location> end = BlockRay.fromEntityRotation(myPlayer, max30).getEndingLocation();
````

Here's a visual for showing the difference between intersecting locations and the discretization of a ray.

I've plotted the function `f(t) = <t, 2t, t>`. Here are the intersection locations of this function. This is basically just every block that this function passes through.

intersection location

And here is the discretization of this path. One can view this as the lowest y-value for each column the function passes through.

discretization

### DIFF
--- a/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
+++ b/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
@@ -1,0 +1,655 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.blockray;
+
+import com.flowpowered.math.GenericMath;
+import com.flowpowered.math.imaginary.Quaterniond;
+import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.block.BlockTypes;
+import org.spongepowered.api.data.manipulator.entity.EyeLocationData;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.extent.Extent;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * A block ray which traces a line and returns all block boundaries intersected in order,
+ * starting from the start location. This class implements the {@link Iterator} interface
+ * with the exception of {@link Iterator#remove()}.
+ *
+ * Filters determine at what location a {@link BlockRay} should stop. A filter
+ * is called at the boundary of each new location that a {@link BlockRay} passes through in order
+ * to determine whether the ray cast should continue or terminate at that location.
+ *
+ * <p>Any one instance of a {@link Predicate} should only be run on one path.
+ * It is not specified that {@link Predicate}s have to be stateless, pure functions.
+ * They are allowed to keep state along an individual path, based on the assertion that a
+ * single instance is only called on one path.</p>
+ *
+ * <p>Filters are most useful for limiting the target block a player is looking
+ * at based on some metric, like transparency, block model, or even distance. The standard
+ * Bukkit-like behavior for finding the target block can be achieved with using
+ * {@link BlockRay#ONLY_AIR_FILTER}, optionally combined with
+ * {@link BlockRay#maxDistanceFilter(Vector3d, double)} to limit the target block to be within some
+ * distance.</p>
+ *
+ * To get the block targeted by an entity, use the following:
+ * <pre>
+ * {@code final Optional<BlockRayHit> block = BlockRay.from(entity).filter(BlockRay.ONLY_AIR_FILTER).end();}
+ * </pre>
+ *
+ * @see BlockRayHit
+ */
+public class BlockRay implements Iterator<BlockRayHit> {
+
+    /**
+     * A block type filter that only permits air as a transparent block.
+     *
+     * <p>This is provided for convenience, as the default behavior in previous systems was to pass
+     * through air blocks only until a non-air block was hit.</p>
+     */
+    public static final Predicate<BlockRayHit> ONLY_AIR_FILTER = blockTypeFilter(BlockTypes.AIR);
+
+    /**
+     * A filter that accepts all blocks. A {@link BlockRay} combined with no other filter than this
+     * one could run endlessly.
+     */
+    public static final Predicate<BlockRayHit> ALL_FILTER = new Predicate<BlockRayHit>() {
+
+        @Override
+        public boolean apply(BlockRayHit lastHit) {
+            return true;
+        }
+
+    };
+
+    private static final Vector3d X_POSITIVE = Vector3d.UNIT_X;
+    private static final Vector3d X_NEGATIVE = X_POSITIVE.negate();
+    private static final Vector3d Y_POSITIVE = Vector3d.UNIT_Y;
+    private static final Vector3d Y_NEGATIVE = Y_POSITIVE.negate();
+    private static final Vector3d Z_POSITIVE = Vector3d.UNIT_Z;
+    private static final Vector3d Z_NEGATIVE = Z_POSITIVE.negate();
+    private static final int DEFAULT_BLOCK_LIMIT = 1000;
+    // Ending test predicate
+    private final Predicate<BlockRayHit> filter;
+    // Extent to iterate in
+    private final Extent extent;
+    // Starting position
+    private final Vector3d position;
+    // Direction of the ray
+    private final Vector3d direction;
+    // The directions the faces are passed through
+    private final Vector3d xNormal, yNormal, zNormal;
+    // The directions the edges and corners are passed through, lazily computed
+    private Vector3d xyzNormal, xyNormal, xzNormal, yzNormal;
+    // The plane increments for the direction
+    private final int xPlaneIncrement, yPlaneIncrement, zPlaneIncrement;
+    // The current coordinates
+    private double xCurrent, yCurrent, zCurrent;
+    // The current passed face
+    private Vector3d normalCurrent;
+    // The next plane values
+    private int xPlaneNext, yPlaneNext, zPlaneNext;
+    // The solutions for the nearest plane intersections
+    private double xPlaneT, yPlaneT, zPlaneT;
+    // Limits to help prevent infinite iteration
+    private int blockLimit = DEFAULT_BLOCK_LIMIT, blockCount;
+    // Last block hit
+    private BlockRayHit hit;
+    // If hasNext() is called, we need to move ahead to check the next hit
+    private boolean ahead;
+
+    private BlockRay(Predicate<BlockRayHit> filter, Extent extent, Vector3d position, Vector3d direction) {
+        this.filter = filter;
+
+        this.extent = extent;
+        this.position = position;
+        this.direction = direction;
+
+        // Figure out the direction of the ray for each axis
+        if (this.direction.getX() >= 0) {
+            this.xPlaneIncrement = 1;
+            this.xNormal = X_NEGATIVE;
+        } else {
+            this.xPlaneIncrement = -1;
+            this.xNormal = X_POSITIVE;
+        }
+        if (this.direction.getY() >= 0) {
+            this.yPlaneIncrement = 1;
+            this.yNormal = Y_NEGATIVE;
+        } else {
+            this.yPlaneIncrement = -1;
+            this.yNormal = Y_POSITIVE;
+        }
+        if (this.direction.getZ() >= 0) {
+            this.zPlaneIncrement = 1;
+            this.zNormal = Z_NEGATIVE;
+        } else {
+            this.zPlaneIncrement = -1;
+            this.zNormal = Z_POSITIVE;
+        }
+
+        reset();
+    }
+
+    /**
+     * Sets the maximum number of blocks to intersect before stopping.
+     * This is a safeguard to prevent infinite iteration.
+     * Default value is 1000. Use a negative value to disable this.
+     *
+     * @param blockLimit The block limit
+     */
+    public void setBlockLimit(int blockLimit) {
+        this.blockLimit = blockLimit;
+    }
+
+    /**
+     * Resets the iterator; it will iterate from the starting location again.
+     */
+    public final void reset() {
+        // Start at the position
+        this.xCurrent = this.position.getX();
+        this.yCurrent = this.position.getY();
+        this.zCurrent = this.position.getZ();
+
+        // First planes are for the block that contains the coordinates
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by TheAwesomeGem** [comment]
This is quite helpful.

File: src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java [hunk 1]
```diff
@@ -0,0 +1,126 @@
...
+        @Override
+        public void start(double x, double y, double z, Vector3d direction, Extent extent) {}
+
+        @Override
+        public boolean shouldContinue(double x, double y, double z) {
+            return false;
+        }
+    };
+
+    public static BlockRayFilter blockType(final BlockType type) {
```
ST-DDT: Maybe you should use a `Predicate` here.

ST-DDT:
This is a nice PR that defenitly has some benefits over my implementation.

src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java
gratimax: No idea what you mean... this is supposed to be a constructor

src/main/java/org/spongepowered/api/util/Collect.java
```diff
@@ -0,0 +1,53 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+import java.util.Iterator;
+
+/**
+ *
+ */
+public final class Collect {
```
gratimax: I'm still thinking over the proper implementation of `BlockRay` (working with iterables and iterators in this manner is... difficult), so this utility may not be needed at all.

src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java
ST-DDT: Sorry i should have added some more explanation. I meant you could use a `Predicate` as argument. Example:
````java
public static BlockRayFilter blockTypes(final Predicate<BlockType> types) {
	return new DiscreteBlockRayFilter() {
		[..]
		public boolean shouldContinue(int x, int y, int z, Direction blockFace) {
			return types.apply(extent.getBlockType(x, y, z));
		}
	};
}
````
With example usage:
````java
public static final BlockRayFilter ONLY_AIR = blockTypes(Predicates.equals(BlockType.AIR));
public static final BlockRayFilter NONE_SOLID = blockTypes(BlockTypePredicates.NONE_SOLID);
````

And maybe a similar method for `Predicate<Location>` and `Predicate<BlockState>`.

Comment (gratimax) [inline src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java]: The issue I have with `Predicate<Location>` and `Predicate<BlockState>` is that they may create A LOT of unnecessary location objects. I may just change the DiscreteBlockRayFilter type to be `boolean shouldContinue(Extent extent, int x, int y, int z, Direction blockFace)`. That way you can get fast access to what you need without much worry

**Comment by ST-DDT** [comment]
A method to create a `BlockRayFilter` whether the coordinates are still within valid world bounds (0<=y<=255) would be nice. Unless you are going to include that check in your ray algorithm.
But if you abstract the implementation away then this would be quite difficult to guarantee.

**Comment by gratimax** [comment]
@ST-DDT Well, I haven't finished the javadocs yet, but yes the ray cast algorithm will include the height check by default unless one explicitly turns it off (which is not recommended for.. obvious reasons)

File: src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java (hunk 1 above)
Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java]: This is only true for bad implementations.
(If you are and/or-ing the `RayTraceFilter`s, but if you instead and/or the `Predicate`s you can use the created object over and over again, and since you have actually added it to your filter chain, it is guaranteed to be used.)

**Comment by ST-DDT** [comment]
@gratimax Maybe we should create some kind of `RayTracingAlgorithm` interface that defines all default methods of a RayTracing, and the real implementations could add more methods example to turn of height checks etc. That way we have a clean base, but everyone can implement his own ray tracing ( if he don't likes the default one ).

**Comment by gratimax** [comment]
@ST-DDT 

1. The implementation of custom ray casting algorithms by a plugin developer goes against the entire point of this PR, and encouraging it with a base interface seems like a big mistake. The point is that for 99% of use cases the given algorithm should function, and anyone who wants a custom algorithm would need something significantly different.

2. I don't see at all how forcing people to extend BlockRay would be in any way easier or better than having a simple flag for height/distance detection on constructors. Yes, the default settings are default for a reason, but there are valid use cases for wanting to turn them off if the developer knows what they are doing.

3. You purport that a base interface is in the name of allowing people to extend the algorithm, but it's already a class that can be extended on its own with absolutely no problem.. if I make sure I design it right it would be easy enough to implement a different ray casting algorithm if one so pleases. For reasons I have outlined above implementing a custom algorithm should be unnecessary for the absolute majority of use cases.

4. One of the problems I found with your PR was the amount of noise it added in terms of extra interfaces and classes. I already feel that I am at the limit of the scope of the PR with 3 new types, and I do not want to add any more, especially when the reasons for them seem to not be well-founded.

File: src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java (hunk 1 above)
Comment (gratimax) [inline src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java]: Location is unnecessary when it is a proxy to all methods within Extent, simple as that. It's unnecessary churn that can become very large (think thousands or tens of thousands of new objects) for raycasts that seem very simple.

**Comment by ST-DDT** [comment]
@gratimax As for point 4 i already thought about that and a possible fix for that  but i was unsure whether i should create a new PR for that for comparison reasons. (Probably not)
The new version would then be closer to your suggestion.

**Comment by DDoS** [comment]
Has there been any progress on this? I intend on picking it up and completing it for use in the Skylands generator.

File: src/main/java/org/spongepowered/api/util/blockray/BlockRayHit.java [hunk 3]
```diff
@@ -0,0 +1,141 @@
...
+     * Constructs a new block ray hit from the extent that contains it, the coordinates
+     * and the face that was entered.
+     *
+     * @param extent The extent of the block
+     * @param x The x coordinate of the block
+     * @param y The y coordinate of the block
+     * @param z The x coordinate of the block
+     * @param face The entered face
+     */
+    public BlockRayHit(Extent extent, double x, double y, double z, Direction face) {
```
gratimax: I might imagine that the dimension of the ray might matter as well, in case you want to do some more granular sub-block bounding-box calculations, like within stairs or half slabs

src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java
```diff
@@ -0,0 +1,260 @@
...
+     *
+     * <p>This method provides the instance of the last block hit, which contains information
+     * about the exact intersection coordinates, the block face hit and the affected block.
+     * If no blocks have been hit yet, this will be the starting position with face of {@link Direction#NONE}.</p>
+     *
+     * @param lastHit The last block hit
+     * @return True to continue ray casting, false to stop
+     * @see BlockRayHit
+     */
+    public abstract boolean shouldContinue(BlockRayHit lastHit);
```
gratimax: The biggest problem I had with doing something like this was that this creates a lot of objects for block rays that span large distances. Even though several objects within block ray hits are lazily-loaded, they still create multiple objects instead, instead of plain xyz and direction as it was previously.

DDoS: But it make for a nicer interface since I can use iterators. Also only one object is created per iteration, as long as you use the getters for the coordinates instead of the vectors.

src/main/java/org/spongepowered/api/util/blockray/BlockRayHit.java
DDoS: What do you mean by "dimension of the ray"?

gratimax: I meant direction. direction of the ray, whoops

DDoS: Ok, I can add that. I'm also changing face to Vector3d because Direction doesn't provide values for up and down diagonals, which occur when rays enter corners (example, v = (1, 1, 1) and p = (0, 0, 0)).

In general I'm a bit unsure of how to handle sides and corners, I was thinking of averaging and normalizing the normal vectors of the faces that form it.

src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java
gratimax: Then I would make BlockRayFilter implement Predicate<BlockRayHit>, or even replace it all together and have a BlockRayFilters class of common filters (since we have Predicates.and and such from guava)

DDoS: Don't you want to keep the `start()` method?

MikaalSky:
Since I know nothing about this, could I use this to look for the nearest block of a certain type to a certain entitys Location?

gratimax:
@MikaalSky no, block rays are traced in lines (as I thought seemed rather obvious from the PR description, if it was unclear please tell me). The best use case with regards to entities is finding what block an entity is looking towards. 

Checking the nearest block of a certain type does not match. You would probably accomplish that by iterating over the nearest blocks outwards and checking each one. I also wouldn't consider that question well-defined, because there can be multiple nearest blocks of a certain type to an entity.

src/main/java/org/spongepowered/api/util/blockray/BlockRayHit.java
gratimax: Maybe that would be better solved by adding the relevant directions to either Direction or another enum (standard diagonal vectors of which there are twelve in addition to directions).

On the second point, I haven't taken a good look at the method for which the block ray is calculated, but I think that it would need to check if certain coordinates when a ray enters a block are integers. If they are, then the normal averaging and normalization would occur.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,420 @@
...
+/**
+ * A block ray which traces a line and returns all block boundaries intersected in order,
+ * starting from the start location. A {@link BlockRayFilter} is used as a predicate
+ * to know when the trace should be ended. This class implements the {@link Iterator}
+ * interface with the exception of {@link Iterator#remove()}.
+ *
+ * @see BlockRayHit
+ * @see BlockRayFilter
+ */
+public class BlockRay implements Iterator<BlockRayHit> {
```
gratimax: How do you plan to expose the discretization of a block ray? A method on the BlockRay class or one on the builder class?

DDoS: A boolean flag, which you can also toggle in the builder, similarly to `blockLimit`.

src/main/java/org/spongepowered/api/util/blockray/BlockRayFilter.java
gratimax: As I see it `start()` serves (and served) two purposes:

1. Give the starting location of the trace. I don't really see how useful this could be, because it is essentially the same as the first block within the trace. If there is a compelling use case I guess it's possible to keep it
2. Give the direction of the trace, since the direction never changes. In another comment to this PR I suggested putting the direction in BlockRayHit which would solve this issue and make this argument unneeded. 
3. It also served the purpose of giving the extent where the block filter was taking place by passing in a location where `shouldContinue` only passed in x, y, and z coordinates. With BlockRayHit that is no longer useful, since BlockRayHit exposes the extent and such.

src/main/java/org/spongepowered/api/util/blockray/BlockRayHit.java
```diff
@@ -0,0 +1,141 @@
...
+import com.flowpowered.math.GenericMath;
+import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.api.world.extent.Extent;
+
+/**
+ * Represents a block hit by a ray. Stores more information than a regular location.
+ */
+public class BlockRayHit {
```
gratimax: I think there's enough information here to build a Location object, so a `toLocation()` method would be useful

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,420 @@
...
+        final Vector3d direction = Quaterniond.fromAxesAnglesDeg(rotation.getY(), 360 - rotation.getX(), rotation.getZ()).getDirection();
+        // TODO: this needs to use the eye location
+        return from(entity.getLocation()).direction(direction);
+    }
+
+    /**
+     * A builder for block ray, which also implements {@link Iterable} which makes it
+     * useful for 'advanced for loops'. Use {@link #from(Location)} to get an instance.
+     */
+    public static class BlockRayBuilder implements Iterable<BlockRayHit> {
```
gratimax: In a lot of cases the plugin developer only wants to know the final ending location of the ray, so iterating it through seems a little long. I would suggest a `getEndingLocation()` method or `getEndingBlockRayHit`, but in these same cases location objects are more useful.

gratimax:
I really want to keep the major use case of finding the target block a player is looking at easy. That probably would need a `getEndingLocationFromEntity` method or something like that, or perhaps even adding a method to Entity to get the target block, that refers to the block ray methods (and takes in a filter, stuff like that).

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
DDoS: I'll add a method

    end()

that iterates to the end and returns the last hit.

DDoS:
BlockRayHit block = BlockRay.from(entity).end();

Is this simple enough?

gratimax:
@DDoS yeah, looks good. Probably would just need a bit of explanation in the documentation.

DDoS:
Unless anyone has anything to add, the implementation is complete.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,650 @@
...
+
+    /**
+     * Initializes a block ray builder for the entity's eye.
+     * This sets both the starting point and direction.
+     *
+     * @param entity The entity
+     * @return A new block ray builder
+     */
+    public static BlockRayBuilder from(Entity entity) {
+        final Vector3d rotation = entity.getRotation();
```
DDoS: Missed a null check...

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,650 @@
...
+     * @param type The type of blocks to allow
+     * @return The filter instance
+     */
+    public static Predicate<BlockRayHit> blockTypeFilter(final BlockType type) {
+
+        return new Predicate<BlockRayHit>() {
+
+            @Override
+            public boolean apply(BlockRayHit lastHit) {
+                return lastHit.getLocation().getType().equals(type);
```
DDoS: I could probably reduce object creation quite a bit here.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,650 @@
...
+
+        private final Vector3i target;
+
+        private TargetBlockFilter(Vector3d target) {
+            this.target = target.floor().toInt();
+        }
+
+        @Override
+        public boolean apply(BlockRayHit lastHit) {
+            return !lastHit.getBlockPosition().equals(this.target);
```
DDoS: I could save one object from being created here.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,650 @@
...
+        }
+    }
+
+    /**
+     * A filter that only allows blocks of a certain block type.
+     *
+     * @param type The type of blocks to allow
+     * @return The filter instance
+     */
+    public static Predicate<BlockRayHit> blockTypeFilter(final BlockType type) {
```
gratimax: I feel like methods like these, related to block ray filters belong on another class, since they aren't strictly related to dealing with block rays themselves.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -0,0 +1,650 @@
...
+         */
+        public BlockRay build() {
+            Preconditions.checkState(this.direction != null, "Either end point or direction needs to be set");
+            final BlockRay blockRay = new BlockRay(this.filter, this.extent, this.position, this.direction);
+            blockRay.setBlockLimit(this.blockLimit);
+            return blockRay;
+        }
+
+        @Override
+        public Iterator<BlockRayHit> iterator() {
```
gratimax: How about adding an `end()` method here as well that builds a block ray and iterates fully?

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
DDoS: There's only a few of them, I don't think it's worth it. Adding a "filter" suffix makes it clear enough.

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
DDoS: I could do that, just feels a bit out of place.

DDoS:
Squashed everything, ready for merge. @gratimax good?

PR #688, opened 2015-05-18
Add AuthorData which represents the author of a book

Grinch:

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/item/AuthorData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/item/AuthorData.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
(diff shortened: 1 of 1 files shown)


ST-DDT:
After a first check i have seen PagedData.
But havent seen Title and SubtitleData or GenerationData lately. Do they (still) exist? Or could you add them as well?

Grinch:
@ST-DDT TitleData is now there (literally just pushed it). Not sure what you mean for Subtitle data. Could you elaborate?

ST-DDT:
Oh i mixed that up with the title screen.
But books have a generation. Original, CopyOfOriginal, CopyOfCopy. Is there something in the API yet? Afaik it had been at some point. I guess before zhe unified data api

Grinch:
@ST-DDT After listening to what @gabizou has done, DisplayNameData handles titles for books and CloneableData handles generation data for books.

gabizou:
Added in 6b7623

TheAwesomeGem:
Looks great!

PR #692, opened 2015-05-20
Add SkinData for human entities and add HUMAN to EntityTypes

simon816:
Adds a `SingleValueData` called `SkinData` that stores the UUID of a human's skin


modwizcode:
@simon816 Why have this? It cannot be manipulated. Shouldn't the skin UUID have the same value as the user's UUID?

simon816:
For human entities (https://github.com/SpongePowered/SpongeCommon/pull/19) the skin can be changed. On players, `offer(SkinData)` will always fail.

PR #694, opened 2015-05-21
Add message sinks to control message targeting

zml2008:
The goal of this PR is to provide a proposal for a system that targets and formats messages sent to the game. This will provide chat channel functionality as well -- plugins will be able to provide custom sink implementations.

This is an initial sketch of the API. I hope to have richer API for combining message sinks and am considering making MessageSink an abstract class to make it more useful on its own (an actual send() method comes to mind here).

Implementation is upcoming.


src/main/java/org/spongepowered/api/util/command/CommandSource.java
```diff
@@ -58,4 +58,19 @@
...
+    MessageSink getMessageSink();
+
+    /**
+     * Set the message sink that messages sent by this source should be broadcast to.
+     *
+     * @param sink The message sink to broadcast messages to
+     */
+    void setMessageSink(MessageSink sink);
+
+
```
mbax: Double space

src/main/java/org/spongepowered/api/util/command/MessageSinks.java
```diff
@@ -0,0 +1,39 @@
+package org.spongepowered.api.util.command;
+
+/**
+ * Useful building blocks for message sinks.
+ */
+public class MessageSinks {
+    private static final MessageSinkFactory factory = null;
+
+
```
mbax: Double space

modwizcode:
This looks pretty good. I like the sinks, is it assumed that plugins will provide MessageSinks as well?

mbax:
This is... rather fancy. 

Looks like right now the easiest way to get a sink containing an arbitrary group of players is to iterate them to get their sinks and then use the factory method, which is only even close to short via Java 8.
```java
MessageSink sinky = MessageSinks.combined(players.stream().map(CommandSource::getMessageSink).toArray(MessageSink::new));
```

MessageSinks are meant to be overrided by the plugins, right? That would allow for customizing various components of the message depending on who is receiving it.

zml2008:
@mbax That is something that will be in the factory -- a method to create a sink for a given set of sources. The line you posted wouldn't even work -- the CommandSource.getMessageSink returns the default message sink that the command source should be broadcasting to -- in most cases all players -- so you would just broadcast to all players.

Sinks combined using the 'combine' method will have message transformations applied in order and broadcast the message to all receivers returned.

And yes, chat plugins are able to implement their own message sinks

boformer:
Maybe add a way to set the **default message sink** that is used for broadcasts and login messages?

src/main/java/org/spongepowered/api/util/command/MessageSink.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return The transformed text. May be input.
+     */
+    Text transformMessage(Text text);
+
+    /**
+     * Return all command sources that will receive messages sent through to this sink.
+     *
+     * @return An iterable of all possible receivers of messages
+     */
+    Iterable<CommandSource> getReceivers();
```
boformer: Receivers or Recipients? I prefer the second term.

src/main/java/org/spongepowered/api/util/command/MessageSink.java
```diff
@@ -0,0 +1,48 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.command;
+
+import org.spongepowered.api.text.Text;
+
+/**
+ * Represents a function that takes a message and distributes it to the appropriate targets.
```
boformer: Well, right now it does not really distribute the message. It only transforms the message and returns a list of recipients.

src/main/java/org/spongepowered/api/util/command/MessageSink.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public interface MessageSink {
+    /**
+     * Handle transforming the input message appropriately.
+     *
+     * @param text The message to send
+     * @return The transformed text. May be input.
+     */
+    Text transformMessage(Text text);
```
boformer: Is this method intended to produce chat-ready messages from "raw input"?

Like "Hello" --> "[Staff] zml2008: Hello".

In that case, a way to obtain the name of the message sender is missing.

mbax:
What's the difference between the MessageEvent's ability to change the message sent, and the transform message in the sink? Feels redundant. 

Still trying to properly visualize how one plugin might define different messages going to different players all in one MessageEvent and then another plugin edit those messages.

src/main/java/org/spongepowered/api/util/command/MessageSink.java
zml2008: This method is intended to produce sink-specific output for a given input -- so in a channels system, this would be the method that prepends the name of the channel. No information is really known about the type of message, and that is intentional to allow any message to pass through channels

modwizcode: So if I needed to prepend a rank?

zml2008: Then you would modify the chat event's message. Not every message that goes through a message sink is at all associated with a player. A MessageSink is simply a targeting system for generic messages.

src/main/java/org/spongepowered/api/util/command/MessageSink.java
ryantheleach: I think the idea is that any message sent, will have a sink on the message event.

The API has no method to process the message when the event resolves as that is implementation specific and shouldn't be exposed or else plugins could circumvent the message event and send messages straight to sinks?

I may be wrong, but it seems like doing this excludes custom sinks from processing the data sent to them, maybe an abstract class is needed for custom sinks?

src/main/java/org/spongepowered/api/util/command/MessageSinkFactory.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    MessageSink toPermission(String permission);
+
+    /**
+     * A message sink that targets all subjects currently active.
+     *
+     * @return The sink
+     */
+    MessageSink toAll();
+
```
ryantheleach: toWorld? or would you just assign a permission with context of world for each world, and use toPermission?

Or make it yourself?

zml2008:
The MessageSink setMessage method is applied to any message passed through without any knowledge of context, while the event setNewMessage methods are used with knowledge of context, to change the message being passed.

Currently sending different messages to different users is not supported. I suppose I could change the transformMessage to also take the CommandSource the transformation is occurring for? Not sure.

src/main/java/org/spongepowered/api/text/sink/MessageSinkFactory.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    MessageSink combined(MessageSink... sinks);
+
+    /**
+     * Get a message sink that targets the given sources.
+     *
+     * @param sources The sources to have as recipients
+     * @return The sink
+     */
+    MessageSink to(Set<CommandSource> sources);
```
ST-DDT: Is the set copied to an immutable Set internally?
If yes then we could change the param to a collection instead. This would allow more generic use.

zml2008: The point of using a set is that it explicitly denies duplicate entries -- we don't want duplicate entries.

ST-DDT:
Could you please give me an example usecase for the `transformMessage` method. Except writing `Hello ${player}` for broadcasts in a single call.

zml2008:
@ST-DDT 
The transformMessage method is not supposed to have much knowledge about what's happening. It's used to annotate the message with information about the given sink being applied. An example would look like:
```java
@Override
public Text transformMessage(Text input) {
    return Texts.of(Texts.of(TextColors.GREEN, "[", channelName, "] "), input);
}
```

ST-DDT:
Okay then how do i change the order of the tags/names applied.
As far as i can tell the only thing this think allows me to do is `[World][Sender] Hello World`. But nothing more. `[Sender][World] Hello World` is not possible. Afaict this is only usefull for a small subset of events that could be covered by a simple for loop as well.

mbax:
> The MessageSink setMessage method is applied to any message passed through without any knowledge of context

In what scenario would I want to use MessageSink to change a message, then?

zml2008:
You'd use MessageSink to change a message for things like word filtering, or to apply a channel prefix -- basically any modification that doesn't require context about the message being passed.

boformer:
The name "sink" bugs me because right now there is no direct way to pass a Message to a sink.  Wouldn't it be better to add a sendMessage() function to MessageSink? I would also remove the transform and maybe even getReceivers method. Player and other CommandSources could implement the MessageSink interface, which would only contain the sendMesssage method.

zml2008:
There is a sendMessage method -> https://github.com/SpongePowered/SpongeAPI/pull/694/files#diff-0555bf28824e34820b7d484fd1310cf8R39

Sinks are limited how they are because that restricts implementations enough so that MessageSinks can be operated on without controlling the implementation as long as plugin devs don't do very dumb things (so we have to push them a bit towards the right way to do things).

mbax:
> You'd use MessageSink to change a message for things like word filtering, or to apply a channel prefix -- basically any modification that doesn't require context about the message being passed.

But why not just use the event's setter for that?

zml2008:
Because you can just set the sink on the player, and your filtering will apply on any message sent from the player going to the appropriate targets, rather than having to listen to each event -- it's more generic

mbax:
So it's preprocessing messages prior to them showing up in events? That's the part I was missing.

zml2008:
It's post-event

mbax:
Post event means that the message could now be any type of Text with any content. Good luck, little filter method, with figuring out what to do with yourself!

zml2008:
The point of the filter method is to perform filtering related to the destination of the message, not the contents of the message or the source of the message.

mbax:
What about scenarios where the destination and source are important? I feel like it could benefit from more information being available at time of customization. :3

zml2008:
Well you can always set a message sink on the event with whatever parameters you want

ST-DDT:
Wasnt the whole point of MessageSink to avoid the event stuff?

zml2008:
@SpongePowered/developers unless people bring up any outstanding concerns in this or impl, I plan to merge this tomorrow

@ST-DDT the point is to provide a consistent interface to message targeting, not to entirely replace events. events will always provide the most fine-grained control.

ryantheleach:
Any chance of some example code on how they would be used for per player formatting for example before merging?

Like imagine coloring the players name, red if they are on an enemy team, green if allied. this is specific for the relationship between the message sender and the viewer.

PR #697, opened 2015-05-24
Refactor the scheduler service. Cleanup javadocs, unify async and sync schedulers, tasks created using a builder.

simon816:
~~As put by gratimax
"rewrote some javadocs to make them less implementation-specific, changed a few types from `Collection<Item>` to `Set<Item>`, and unwrapped a few `Optional` types within `Task`."~~

This does more than cleaning up the scheduler now.

* Reduce duplication and the amount of text in the javadocs. Some details were implementation specific which is not be mentioned.
* Change `Collection<Task>` to `Set<Task>` as there should be no duplicate tasks.
* Removed `Optional`s within `Task`
* Removes separation of sync and async schedulers. There is now a single scheduler for all tasks
* Tasks are created using a `TaskBuilder` which follows the builder pattern except you 'submit' the task at the end rather than build. This submits the task to the scheduler to then run according to it's parameters.
* Ability to schedule synchronous tasks in wall clock time parameters, as well as ticks


src/main/java/org/spongepowered/api/service/scheduler/SchedulerQuery.java
```diff
@@ -26,70 +26,54 @@
...
+     * Gets a set of task IDs for the given task name. There may be more than
+     * one task with the same name.
      *
-     * @param name  The name of the task to search
-     * @return The Optional&lt;UUID&gt; result from the search by name.
+     * @param name The name of the task to search
+     * @return A set of task IDs, empty if no tasks have that name
      */
-    Optional<UUID> getUuidOfTaskByName(String name);
+    Set<UUID> getUuidOfTaskByName(String name);
```
Minecrell: Is this method really needed? The method below (task by name) should provide the same

simon816: I'll remove it.

FerusGrim:
Would you mind adding a little description explaining what the changes are, and the benefits of them (for those of us who can't tell as easily, looking at raw files)?

gratimax:
He rewrote some javadocs to make them less implementation-specific, changed a few types from `Collection<Item>` to `Set<Item>`, and unwrapped a few `Optional` types within `Task`.

simon816:
As suggested by @Zidane here: https://github.com/SpongePowered/SpongeCommon/pull/47#issuecomment-105766781, the scheduler API could adopt a builder pattern for creating tasks.
```java
final Task task = this.game.getTaskBuilder()
    .async(true)
    .execute(new Runnable() {})
    .delay(10, TimeUnit.MILISECONDS)
    .repeatEvery(10, TimeUnit.MILISECONDS)
    .side(Platform.Side.SERVER) // Once Sponge is on the client, this can determine if to run on client or not
    .build()

this.game.getScheduler().submit(this, task);
```
I will go ahead making these changes, feedback is welcome if there are any suggestions or doubts about it.

Minecrell:
The new way to schedule tasks looks awesome, here are a few suggestions that would be good in my opinion:

- `TaskBuilder` creates instances of unscheduled tasks - generally the tasks are only schedulable once so it doesn't make much sense to make them buildable without directly scheduling them. If we would move the `TaskBuilder` method to the `Scheduler` the builder could have a method like `submit(Plugin)` to build it and schedule it directly (which would also solve the problem of tasks not having any owner before they're scheduled).

- The scheduler currently only supports scheduling runnables directly. If you want to query a special value asynchronously (e.g. get something from a webserver), it would be probably also useful to be able to schedule a [`Callable`](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Callable.html) that would provide a `ListenableFuture` when scheduled. Of course, repeating this wouldn't make any sense so that should be only possible to schedule for execution once in this special case.

src/main/java/org/spongepowered/api/service/scheduler/Scheduler.java
```diff
@@ -26,60 +26,21 @@
...
-import java.util.Collection;
+import java.util.Set;
 import java.util.UUID;
 
 /**
- * The base scheduler that schedules tasks.
+ * Represents a scheduler for running {@link Task}s.
  */
-
 public interface Scheduler {
```
Minecrell: This should be actually called `SchedulerService` for consistency with the other services

src/main/java/org/spongepowered/api/service/scheduler/TaskBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+     * @param unit The unit the delay is in
+     * @return This builder, for chaining
+     * @throws IllegalArgumentException If the delay is below 0
+     */
+    TaskBuilder delay(long delay, TimeUnit unit);
+
+    /**
+     * Sets the delay in ticks before the task runs. This only has an affect
+     * when the builder is in sync mode, if {@link #async()} has been called,
+     * this method does nothing.
```
Minecrell: It's not a good idea to make this simply do nothing in one special case. It would be really more consistent especially with `Task.getDelay()` to make this use the default unit for the selected type of task (in our implementations most likely ms for async, ticks for sync).

src/main/java/org/spongepowered/api/service/scheduler/Task.java
```diff
@@ -48,21 +47,26 @@
     PluginContainer getOwner();
 
     /**
-     * Gets the delay that the task was scheduled to run after.
+     * Gets the delay that the task was scheduled to run after. A delay of 0
+     * represents that the task will start immediately.
      *
-     * @return The delay (offset) in the scale of the time unit applied if asynchronous, otherwise raw synchronous ticks
+     * @return The delay (offset) in the scale of the time unit applied if
```
Minecrell: Should probably use the same defaults as the builder, maybe provide `getDelay(TimeUnit)` additionally?

gratimax:
I would prefer being able to create a TaskBuilder more naturally and severing any connection to a scheduler instance until it is explicitly set, so it would look like:
```java
new TaskBuilder()
    .async(true)
    .execute(new Runnable() {})
    .delay(10, TimeUnit.MILISECONDS)
    .interval(10, TimeUnit.MILISECONDS)
    .scheduler(game.getScheduler()) // <- only here do we first make a reference to the game
    .submit(plugin);
```
In my opinion, this has the advantage of being more usable for people ultimately using the task scheduler API. It's sort of a minor change and would require a class of some sort. It seems to me that a Task can be expressed the same way across implementations, but the way it is actually run differs. If this presumption is wrong, then the current method of a builder interface is fine.

Also, is it really a TaskBuilder if it doesn't have a `build()` method? Just some food for thought.

src/main/java/org/spongepowered/api/service/scheduler/TaskBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+     * of the tick cycle and in it's own thread. Therefore the task is <b>not
+     * thread safe</b> with game data and care must be taken. It is strongly
+     * advised to <b>not</b> manipulate game data in asynchronous tasks.</p>
+     *
+     * <p>It is not possible to schedule a task in unit ticks when running
+     * asynchronously, that is why the returned builder is specialised for just
+     * building async tasks. The delay and interval are converted by the
+     * implementation to the equivalent wall clock time.</p>
+     *
+     * @return A specialised asynchronous task builder
```
Zidane: The API doesn't care about the fact the impl may have this as an AsyncTaskBuilder at this point. Docs need to be fixed :).

simon816:
I omitted having a `side(Platform.Type)` method in the builder because plugins can easily check the side and determine whether they should build the task or not.

@Minecrell I think adding support for `Callable` is a bit too specific. If a plugin wants to do some callback type of task they can do that within the `run()` method and handle it themselves.

@SpongePowered/developers Thoughts on @gratimax's suggestion?

Zidane:
Tasks are only for the scheduler, don't see why the instance would be needed to set.

@simon816 

The side check is a convenience for when you want to just run it on one side without needing to do checks. We can add it later once we think we'll be on the client.

me4502:
How far from merging is this PR?

RobertHerhold:
Status update?

modwizcode:
In favor of merging.

Zidane:
Merged in 227cf04e4ac4f5393f5dd0687f713bc00683e317

Huge thanks @simon816

PR #705, opened 2015-05-29
Token API

gabizou:
Tokens API Proposal
=============

SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/54)

Over a month ago, *Data API* was introduced to SpongeAPI. Initially, people didn't quite understand all of the reasoning behind it, however, now, 
people are beginning to use it! This is a good thing! Fortunately, the API solved the issue of relying on interface hierarchy to retrieve various 
bits of information with the following example:

###### Getting health from an entity (previously)

```java
if (entity instanceof Living) {
    final double health = ((Living) entity).getHealth();
    final double maxHealth = ((Living) entity).getMaxHealth();
    ((Living) entity).setHealth(maxHealth);
}
```

###### Getting health with Data API (current)

```java

final Optional<HealthData> healthOptional = entity.getData(HealthData.class);
if (healthOptional.isPresent()) {
    entity.offer(healthOptional.get().setValue(healthOptional.get().getMaxHealth());
}

```

What we've gained from *Data API* is the following:
- `Entity`, `ItemStack`, `Location`, and `TileEntity` all are oblivious as to what possible *data* they have, all they know is that internally, 
they store some various `DataManipulator`s, much like an *EntityComponentSystem* setup would have.
- We no longer focus on what *interface* an `Entity` is, we have a simple way of retrieving a *DataManipulator* which focuses on containing the 
various values we care about
- Serializing and deserializing these `DataHolder`s becomes easy since we now have a way to get a `DataContainer` from these `DataHolder`s

On the other hand, what we've lost is the ability to write the following:

```java

entity.setHealth(entity.getMaxHealth());

```

What if however, we had a way to identify the values we want to set, without worrying whether the `DataHolder` actually contains the 
`DataManipulator` and just blindly set a value without a care in the world?

That's part of what I'm presenting today: `Token API`

##### What is *Token API*?

It's a simple representation of a piece of data that a `DataHolder` may contain, that can be retrieved and replaced with a new value. The `Token` 
is simply a "key" of sorts to an underlying "value". To put it simply, all `DataHolder`s are realistically 
[`Tuples`](http://docs.oracle.com/javaee/6/api/javax/persistence/Tuple.html), even when the `value` is far more complex, such as `MobSpawnerData`.

The full changelog of this PR is:

 - **Rename `DataManipulator` to `Component`** : Let's be honest here, `DataManipulator`s were really just `Component`s the entire time, just not 
 called that way to avoid scrutiny.
 - All `Component`s with values have a `Token` to retrieve that particular value from a `DataHolder`
 - `Token`s provide their `DataQuery` of which they will be placed within a `DataContainer` when `Component#toContainer()` is used
 - Simpler methods to manipulate specific values on `DataHolder`s with a `Token`
 - *Marker* `Component`s that previously would only exist if a "state" was true, it was functionally difficult to understand whether that marker 
 `component` was compatible or just not available for the `DataHolder`. These marker `Component`s are now `BooleanComponent`s such that if the 
 boolean value can be set to `true` or `false`, they will always exist if the `DataHolder` supports it
 - `Token`s have a link to their parent `Component` class
 - `DataHolder` has gained a few data manipulation methods using `Token`s:
    - `DataHolder#get(GetterToken<T>)::Optional<T>` : Retrieves the value if available
    - `DataHolder#getOrDefault(GetterToken<V>,V)::V` : Retrieves the value, or returns default if not available
    - `DataHolder#getOrNull(GetterToken<V>)::V` : Retrieves the value, or null if not available
    - `DataHolder#set(SetterToken<V>,V)::DataHolder` : Sets the given value onto the DataHolder, returns the DataHolder
    - `DataHolder#transform(Token<V>,Function<V, V>)::DataHolder` : Applies the transformation function on the given value and sets it, returns the
     DataHolder
    - `DataHolder#predicateFor(Token<V>)::Optional<Predicate<V>>` : Retrieves the [`Predicate`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Predicate.html)
     for the given `Token`
 

##### So... what's the point?

Well, fortunately, `Token` will simplify performing simple operations that wish to edit and transform a specific value belonging to a `DataHolder` 
without meddling with the `Component` that otherwise contains that specific value.

##### Wait, but isn't this allowing multiple ways to get the same data?

Yes and no. When you use a `Token` from a `DataHolder`, you're getting the specific value, nothing more, nothing less. You're not able to apply 
multiple `Token`s to multiple `DataHolder`s without knowing the type of the `Token`, that functionality resides in the `Component` aspect of the API.

##### How does the first example look now with *TokenAPI*?

I'm glad you asked, it will now look like this:

```java 

if (entity.supports(Tokens.HEALTH)) {
    entity.set(Tokens.HEALTH, entity.get(Tokens.MAX_HEALTH).get());
}

```

To be honest, that seems about as simple as the normal interface would have us use to perform the same operation. Without the use of the 
`Component`, we simply are not worrying about retrieving the `HealthData` beforehand, we're merely applying a new value directly.

##### What else does this *TokenAPI* allow?

Interestingly, when you decouple the notion of a value being manipulated outside of a `Component`, you're able to perform interesting operations, 
such as the following:

```java

entity.transform(Tokens.HEALTH, new Function<Double, Double>() {

    @Override
    Double apply(Double input) {
        return entity.get(Tokens.MAX_HEALTH).get() * 0.9;
    }
});

```
##### How does this look like for a plugin?

Really simple!

```java

import org.spongepowered.api.data.DataContainer;
import org.spongepowered.api.data.component.entity.HealthComponent;
import org.spongepowered.api.data.token.TokenFunctions;
import org.spongepowered.api.data.token.Tokens;
import org.spongepowered.api.event.Subscribe;
import org.spongepowered.api.event.entity.player.PlayerJoinEvent;

@Plugin(id = "tokensTest", name = "TokensTest", version = "0.1")
public class TokensTest {

    @Subscribe
    public void onPlayerJoin(PlayerJoinEvent event) {
        event.getEntity().transform(Tokens.HEALTH, TokenFunctions.add(5D));
    }

    // This is if you were to want to manipulate or retrieve data from a DataContainer
    // with some DataQueries from Tokens.
    public void setHealthOnContainer(Player player) {
        final DataContainer container = player.toContainer();
        final double maxHealth = container.getDouble(Tokens.MAX_HEALTH.getQuery()).get();
        container.set(Tokens.HEALTH.getQuery(), maxHealth);
    }
}
```

#### But wait, isn't this the same as [Values PR](https://github.com/SpongePowered/SpongeAPI/pull/685)?
Short answer, yes.
Long answer: The way I see this is as a step towards *Value API*, setting the *Data API* in a readier state to evolve into *Value API*. While *Data API* and *Token API* strive to achieve a lot of functionality, there are some limitations with regards to how they interact with the game, and *Value API* aims to fully re-engineer *Data API* into something much more powerful, however, due to time constraints, I've presented *Token API* as a compromise for the duration that *Value API* remains in development.


modwizcode:
I like this, especially `BooleanComponent` which is something I really dislike about the current Data API. I'm a little confused about the proper time to use the `Component` vs. the `Token` based interaction. 

What's making this so different from the full Value API? I see the major change for both being the same?

ST-DDT:
Why did you add 
* `DataHolder#getOrDefault(GetterToken<V>,V)::V`
* `DataHolder#getOrNull(GetterToken<V>)::V`

since you can achieve the same with 
* `DataHolder#get(GetterToken<T>)::Optional<T>` and `Optional<T>#or(T)::T`

IMO its just adds additional method declaration noise and more methods that have to be kept uptodate.

gabizou:
> Why did you add
> 
> DataHolder#getOrDefault(GetterToken<V>,V)::V
> DataHolder#getOrNull(GetterToken<V>)::V
> since you can achieve the same with
> 
> DataHolder#get(GetterToken<T>)::Optional<T> and Optional<T>#or(T)::T
> IMO its just adds additional method declaration noise and more methods that have to be kept uptodate.

Compromises for those who wish to not have to deal with `Optional`. Some people just want to deal with nulls, and some people want defaults.

boformer:
I agree with ST. You can just use ``optional.orNull()`` or ``optional.or(...)``.

gabizou:
> I agree with ST. You can just use optional.orNull() or optional.or(...).

There's no need to re-iterate an opinion that I've already responded to.

jamierocks:
Is this not just https://github.com/SpongePowered/SpongeAPI/pull/685?

simon816:
@jamierocks Read the bit at the bottom of the PR description.

jamierocks:
:| /me needs to read properly

gabizou:
Closing after agreement with @gratimax to continue with #685.

PR #708, opened 2015-05-31
Add methods to work with XML-format text messages.

zml2008:
This PR proposes the integration of @TomyLobo's XML format for chat styling into the Text API through the addition of new methods in Texts to go to and from XML.

The XML format vastly improves usability of plugins wishing to have user-configurable formatted text by providing a less verbose method for specifying formatting, with similarities to HTML. The tags handled are as follows:

### HTML-compatible

- `b`: bold text
- `i`: italic text
- `u`: underlined text
- `<a href="url">`: create a link
- `<span>`: Does nothing to the text inside, useful to apply text actions though

### Custom
- `<o>` or `<obfuscated>`: obfuscated text
- `<color n="red">` or `<c n="red">` or `<color name="red">`: specify the color of text
- `<s>` or `<strikethrough>`: Strikethrough
- `<tr key="mc-key">`: Translation key. Each child element is an argument

### Text Actions
Any tag can take one of the attributes onClick, onHover, or onShiftClick. Each attribute has a value of the form type('value'), for example `<span onClick="run_command('/say hi')">Greet the server!</span>`

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Texts.java
+++ b/src/main/java/org/spongepowered/api/text/Texts.java
@@ -435,6 +435,29 @@ public final class Texts {
     public static String toJson(Text text, Locale locale) {
         return factory.toJson(text, locale);
     }
+
+    /**
+     * Parses the provided XML-formatted message to a Text object.
+     *
+     * @param input The input string to parse
+     * @return The parsed {@link Text}
+     */
+    public static Text parseXml(String input) {
+        return factory.parseXml(input);
+    }
+
+    /**
+     * Returns an xml representation of the {@link Text} for usage in configurations.
+     * TODO: Does Locale make sense here?
+     *
+     * @param input The input to use
+     * @param locale The locale to dump to
+     * @return The text as xml
+     */
+    public static String toXml(Text input, Locale locale) {
+        return factory.toXml(input, locale);
+    }
+
     /**
      * Returns the default legacy formatting character.
      *
--- a/src/main/java/org/spongepowered/api/text/Texts.java
+++ b/src/main/java/org/spongepowered/api/text/Texts.java
@@ -370,31 +370,6 @@ public final class Texts {
     }
 
     /**
-     * Parses the specified JSON text and returns the parsed result.
-     *
-     * @param json The valid JSON text
-     * @return The parsed text
-     * @throws IllegalArgumentException If the JSON is invalid
-     */
-    public static Text parseJson(String json) throws IllegalArgumentException {
-        return factory.parseJson(json);
-    }
-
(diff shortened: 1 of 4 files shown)


boformer:
Color could also be an attribute:

```
<span color="gold">
```

src/main/java/org/spongepowered/api/text/Texts.java
```diff
@@ -435,6 +435,29 @@ public static String toPlain(Text text, Locale locale) {
...
+     * @param input The input string to parse
+     * @return The parsed {@link Text}
+     */
+    public static Text parseXml(String input) {
+        return factory.parseXml(input);
+    }
+
+    /**
+     * Returns an xml representation of the {@link Text} for usage in configurations.
+     * TODO: Does Locale make sense here?
```
simon816: Looking at some of the other methods, `Locale` is an overload parameter
(e.g `toJson(Text)` and `toJson(Text, Locale)` so perhaps provide a `toXml(Text)` method too

TomyLobo:
@boformer good point, but that doesn't mirror the structure as well as the onClick and onHover attributes do (they're an attribute to every json element, iirc), which is why I left it out.

But that was more laziness than a design choice.
I'm all for it, if someone other than me implements it *g*

zml2008:
So proposal for a more generic API:

```java

interface TextSerializer {
    String serialize(Text text);
    String serialize(Text text, Locale locale);
    Text deserialize(String serialized);
}

class Texts {
    // [...]
    public TextSerializer xml() {...}
    public TextSerializer json() { ... }
    // [...]
}
```
This deals with the issue of this format getting too much in-api preferential treatment by providing a common interface for other formats. Additionally, the common interface can be implemented for plugin-provided formats. Thoughts?

TomyLobo:
Would that be ``Texts.xml().deserialize(...)``? In that case you forgot a "static" :)

zml2008:
yep, that's how late-night github notes go :/ but it's actually working in my real impl :)

TomyLobo:
You could go with ``Texts.XML.[de]serialize(...)``

ryantheleach:
Looking back at this is there any reason why links are given there own tag where onClick could have been used? what would <a href="asdf" onClick="runCommand('asfd')">test</a> mean?

zml2008:
That would perform `run_command('asfd')` -- text actions are applied after the tag-specific modifiers are.

TomyLobo:
Now if you define what happens there, you need to support it.
Make it explicitly undefined behaviour.
Then it's on whoever exploits that behaviour to check their code when someone rearranges things :)

Issue #707, opened 2015-05-31
EntityDamageEvent Changes


gabizou:
So, I've had some time to think about the long often misrepresented `EntityDamageEvent` that existed for quite some time with previous API's, and I think I'm on the verge of finding an appropriate solution for the following problem:

Skeleton shoots bow launching an arrow hitting a Player.

This is pretty plain and simple to understand what happens, but we need to look at it much deeper to fully understand what the final damage comes out to:

I'll give an order of how the damage is calculated:
1) Skeleton is shooting the arrow, by default it will apply say 3 damage to the Arrow
2) The bow, if enchanted applies a bonus amount of damage to the arrow, say 1.5 additional damage
The final raw damage of the arrow is now 4.5.

Now here comes the fun part:
The Player is hit with the arrow
1) The damage is modified based on the game difficulty, in this case, it's on `HARD` difficulty, so the damage is multiplied by 1.5 (damage is now 6.75)
2) Armor calculations are factored in, Say that the player has full iron armor, and the iron armor reduces the damage by 2 (damage is now at 4.75)
3) Potion effects add to resistances, so damage is reduced by multiplying damage by 20 and dividing by 25 (damage is now at 3.8)
4) Armor Enchantments reduce the damage by 20% for example (the damage is now at 3.04)
5) The Absorption potion effect is now factored in to find out how much actual health is removed from the entity, say we have an additional 2 health from absorption, (damage is now at 1.04)

The end result is a damage value of 1.04 damage dealt directly to the entity, where 4.5 damage originated.

The reason why I walked you through all of this is the following: I've been contemplating that we should have a LivingPreAttackEntityEvent of which we can trace the modifiers for the damage being dealt to the entity (in the case of the skeleton shooting, there's the modifier of the bow and the modifier base damage of the skeleton). After that event, we'd need a proper `WeightedCollection` to store the `DamageModifier`s and their `Function`s being applied to the incoming damage for an `EntityDamageEvent`. Considering the limitless possibilities of being able to identify the `DamageModifier`s and the `Cause` for each modifier, I do believe this would make an excellent change with Causes PR in the near future.

I'd like to hear some thoughts from the developer community.

modwizcode:
Only one word describes this proposal, Awesome! I would like to hear more about why we are storing `Function`s applied. Also what is a `WeightedCollection`?

AlphaModder:
[...]

Issue #556, opened 2015-04-08
setBlock for players and FakePlayers?


thomas15v:
Shouldn't we have a these methods in BlockVolume?
```java
/* Returns true if the block has been set */
boolean setBlock(Vector3i position, BlockState block, Player player);

boolean setBlock(Vector3i position, BlockState block, String fakePlayer /* Or PluginName */);

/* For Bulk Actions, Both arrays should have the same size */
boolean setBlock(Vector3i[] position, BlockState[] block, Player player);

boolean setBlock(Vector3i[] position, BlockState[] block, String fakePlayer);
```
These methods would than give of an event, so other plugins can deny the placement. With this plugins easily could modify blocks, without having to worry if other plugins agree with it.

DDoS:
No, these shouldn't be in block volume as a block volume could be a buffer. There is no guarantee that a block volume contains the player placing the block. This should go in extent.

It seems to me like this could be made more generic. I don't really like this. Too specific.

thomas15v:
It doesn't need to do the player blockplace animation. It just has to place the block for the player. So that it gives a correct event. Maybe a OfflinePlayer should fit better.

DDoS:
To place a block, a player needs to be in the same volume as that block. Hence the block volume must also be an entity universe. This is a contradiction.

But, Extent extends both block volume and entity universe. This interface could accept these methods.

DDoS:
These methods would be out of place in block volume: specifc methods for specific cases don't belong in a generic interface for generic usage.

thomas15v:
Oh indeed this would fit better in the Extent class. I didn't really looked in to in what class it would fit.

simon816:
So essentially what you're asking here is to include a `Cause` argument?
Something like
```java
boolean setBlock(Vector3i position, BlockState block, Cause customCause);
```

@SpongePowered/developers add this to Extent/Location?

gabizou:
To be honest, I'd only want to bother seeing these sort of "fake" blocks being sent through to `Viewer`s, consider that it's their local world etc, that part is easy.

JBYoshi:
@thomas15v Is the parameter:
- The player that should see the block change (as in a fake block), or
- The player the block should appear to be placed by (as in a cause to the "place block" event)?

DarkArc:
@JBYoshi Do you mean "The block should appear to be placed by (as in a cause to the "place block" event)?"

[...]

PR #712, opened 2015-06-03
[WIP] Cause Enhancements

gabizou:
So far, [`Cause`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/cause/Cause.java) as it stands is somewhat confusing with the inclusion of [`Reason`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/cause/reason/Reason.java) such that it isn't clear as to what should be considered a `Cause` and what should be considered a `Reason`.

This PR aims to achieve the following:
- `Cause` becomes a container object of all related objects associated with the cause for a `CauseTracked` event
- "Reasons" or additional causes aiding the direct cause for an event are included in `Cause`, the closer the association to the direct cause, the lower the index that object will be
- Include `EntityDamageEvent` and `DamageSource`s while keeping in line with the primary `Cause` goal. Any additional associated objects aiding the direct `DamageSource` are included in the `Cause` associated with the `EntityDamageEvent`
- Introduce `EntityPreDamageEvent` where the damage calculations for modifying the raw damage are included, this is primarily used for modifying the outgoing damage BEFORE the `EntityDamageEvent` is fired, this is equivalent to having an event to modify the damage a `Player` will output when using an `ItemStack` of `ItemTypes.DIAMOND_SWORD` such that all modifiers are included (such as the enchantments on the sword, the potion effects, and any other `AttributeModifier`s or alternative `DamageModifier`s relating to the final raw damage output from the `Player`
- For many `CauseTrack`ed events, the `Cause` now is always available, but a `Cause.empty()` may be included

#### So what does this really mean for `CauseTracked` events?

What this means is that a vanilla interaction throwing a `CauseTracked` event will have as much information as possible within the `Cause`. With the following addition to `Cause`:

```java
<T> Optional<T> getFirst(Class<T>);

<T> Optional<T> getLast(Class<T>);
```

a plugin is easily able to determine whether a `BlockState` was the cause for an event, a `Player`, or if a `TileEntity` was part of the mix. 

The one issue with `Cause` previously was that it was always optional. Passing in a null to the [`SpongeEventFactory`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java#L341) is that the `Cause` is always optional, even when it should not be.  In the PR, I change this to where `Cause` can be "empty". Some might think: *Wait! Isn't that making `Cause` just become a similar object to `Optional`?* To which the answer is yes. The one issue with `Optional` however is that  `Optional` can only hold a single value. There is no way to hold multiple values, except if the `Optional` is declared like so: `Optional<Object[]>`, which eliminates any utility methods provided by `getFirst(Class<T>)` and `getAllOf(Class<T>)` from `Cause`.

#### Ok, so where does this come into play with actual events? I want to know the cause for a `BlockChangeEvent`!

Well, see, this is where the uniqueness comes into play. Since we always have a `Cause`, we can always call `getFirst(Player.class)` which will return a present `Player` if and only if a `Player` was involved in the `BlockChangeEvent`. How the objects are ordered in the `Cause` is simple: The direct cause of the event comes first, any additional "helpers" that aid the *root cause* are indexed afterwards.

I'll give an example here of how a `BlockChangeEvent` could have the order of objects in the cause, for the case of a `Player` placing a `ItemTypes.SAPPLING`.

First, the `Player` is the root cause, because the player is right clicking with the `ItemStack`.
Second, the `ItemStack` follows because the `ItemStack` is a part of the "reason" why the `BlockChangeEvent` is being called.

#### Now, what I haven't seen is how this affects something like `EntityDamageEvent`, care to elaborate?

Of course! With an `EntityDamageEvent`, I'll have to pull the same description from #707 

> Skeleton shoots bow launching an arrow hitting a Player.
>
>This is pretty plain and simple to understand what happens, but we need to look at it much deeper to fully understand what the final damage comes out to:

>I'll give an order of how the damage is calculated:
1) Skeleton is shooting the arrow, by default it will apply say 3 damage to the Arrow
2) The bow, if enchanted applies a bonus amount of damage to the arrow, say 1.5 additional damage
The final raw damage of the arrow is now 4.5.

>Now here comes the fun part:
The Player is hit with the arrow
1) The damage is modified based on the game difficulty, in this case, it's on `HARD` difficulty, so the damage is multiplied by 1.5 (damage is now 6.75)
2) Armor calculations are factored in, Say that the player has full iron armor, and the iron armor reduces the damage by 2 (damage is now at 4.75)
3) Potion effects add to resistances, so damage is reduced by multiplying damage by 20 and dividing by 25 (damage is now at 3.8)
4) Armor Enchantments reduce the damage by 20% for example (the damage is now at 3.04)
5) The Absorption potion effect is now factored in to find out how much actual health is removed from the entity, say we have an additional 2 health from absorption, (damage is now at 1.04)

>The end result is a damage value of 1.04 damage dealt directly to the entity, where 4.5 damage originated.

>The reason why I walked you through all of this is the following: I've been contemplating that we should have a LivingPreAttackEntityEvent of which we can trace the modifiers for the damage being dealt to the entity (in the case of the skeleton shooting, there's the modifier of the bow and the modifier base damage of the skeleton). After that event, we'd need a proper `WeightedCollection` to store the `DamageModifier`s and their `Function`s being applied to the incoming damage for an `EntityDamageEvent`. Considering the limitless possibilities of being able to identify the `DamageModifier`s and the `Cause` for each modifier, I do believe this would make an excellent change with Causes PR in the near future.

What HAS been achieved however, is slightly different. Instead of just including the associated `Entity` as a *root* to the `Cause`, a `DamageSource` is now the *root* cause. A `DamageSource` is essentially a wrapper around any object with a descriptive `DamageType` with various *attributes* of the `DamageSource` including whether the damage is *absolute*, *ignores armor*, *magical*, *explosive*, or even *scaled with difficulty*. The included `DamageSource`s in this PR facilitate understanding the *type* of damage being dealt, and the parties involved in the `DamageSource`.

#### I'm a little confused about `DamageSource`, can you give me an example?

Of course! Say you have an `EntityPreDamageEvent`:

The `DamageSource` in this case is actually a `ProjectileDamageSource`, where the `Projectile` is an `Arrow`, and the `ProjectileSource` is a `Skeleton`. Given that the `ProjectileDamageSource` is the *root* cause to the `EntityPreDamageEvent`, we can then use this information to know that the `Cause` should have a related `ItemStack` used as a *bow* to shoot the *Arrow*. With the *bow*, we know that it can have no `Enchantment`s or it can have some `Enchantment`s. Of course, with these objects in the `Cause`, it makes it plain and clear to use them:

```java
@Subscribe
public void cancelAnvils(EntityPreDamageEvent event) {
    if (event.getCause().isEmpty()) {
        return;
    }
    final Cause cause = event.getCause();
    final Optional<BlockDamageSource> damageSource = cause.getFirst(BlockDamageSource.class);
    if (damageSource.isPresent()) {
        if (damageSource.get().getBlockState().getType() == BlockTypes.ANVIL) {
            event.setCancelled(true);
        }
    }
}

@Subscribe
public void skeletonArrows(EntityPreDamageEvent event) {
    if (event.getCause().isEmpty()) {
        return;
    }
    final Cause cause = event.getCause();
    final Optional<ProjectileDamageSource> projectileDamageSource = cause.getFirst(ProjectileDamageSource.class);
    if (projectileDamageSource.isPresent()) {
        final ProjectileSource source = projectileDamageSource.get().getShooter();
        if (source instanceof Skeleton) {
            final SkeletonType skeletonType = ((Skeleton) source).getData(SkeletonData.class).get().getValue();
            if (skeletonType == SkeletonTypes.WITHER) {
                event.setDamageFunction(new MyDamageModifier(Cause.of(source, this)), new Function<Double, Double>() {
                    @Nullable
                    @Override
                    public Double apply(Double input) {
                        return input * 3 / 2;
                    }
                });
            } else if (skeletonType == SkeletonTypes.NORMAL) {
                event.setDamageFunction(new MyDamageModifier(Cause.of(source, this)), new Function<Double, Double>() {
                    @Nullable
                    @Override
                    public Double apply(Double input) {
                        return input * 3 / 4;
                    }
                });
            }
        }
    }
    for (final Tuple<DamageModifier, Function<? super Double, Double>> damageModifier : event.getModifiers()) {
        if (damageModifier.getFirst().getCause().getFirst(ItemStack.class).isPresent()) {
            // We know that we have an ItemStack that was used in the DamageModifier cause, we can suspect that
            // the ItemStack has some sort of enchantments, or not
            final ItemStack itemStack = damageModifier.getFirst().getCause().getFirst(ItemStack.class).get();
            final Optional<EnchantmentData> dataOptional = itemStack.getData(EnchantmentData.class);
            if (dataOptional.isPresent() && dataOptional.get().get(Enchantments.POWER).isPresent()) {
                event.setDamageFunction(damageModifier.getFirst(), new Function<Double, Double>() {
                    @Nullable
                    @Override
                    public Double apply(@Nullable Double input) {
                        return input * 1.25 * dataOptional.get().get(Enchantments.POWER).get().doubleValue();
                    }
                });
            }
        }
    }
}

public final class MyDamageModifier implements DamageModifier {
    private Cause cause;

    public MyDamageModifier(Cause cause) {
        this.cause = checkNotNull(cause);
    }

    @Override
    public Cause getCause() {
        return this.cause;
    }

    @Override
    public String getId() {
        return "com.gabizou.MyDamageModifier";
    }

    @Override
    public String getName() {
        return "MyDamageModifier";
    }
}
``` 

#### Holy cow batman! Why so complicated?!

Unfortunately, there is A LOT that goes into an `EntityDamageEvent`, even more so if you want to be able to manipulate the damage *coming in*, essentially the `EntityPreDamageEvent`. To fully unlock the power available with this, the possibilities to associate new `DamageModifier`s with a damage event are nigh limitless. Since we can expose the underlying functions, as they placed in order from vanilla mechanics, we can successfully interpret that an `EntityPreDamageEvent` is augmented not only by the `Skeleton`, but also that the `ItemStack` used by the `Skeleton` affects the total damage applied onto the `Arrow`. With that in mind, we can also assume that we can interpret the following `EntityDamageEvent` with similar processing for `DamageModifier`s when an `Entity` is being *damaged*, dictating the *final* damage actually being *applied* to the `Entity`, and not just interpret the *raw* incoming damage.

Of course, there are so many factors that can go into the `EntityDamageEvent` that I can't possibly explain it all in a simple code snippet, but I can explain in layman terms here.

With the `EntityDamageEvent` being fired only while the `Entity` is being damaged, we can gather all `DamageModifier`s that would further process the raw incoming damage, such as `ItemStack`s that are considered to be *armor*, `PotionEffect`s that are absorbing damage, heck, we can even apply our own custom `DamageModifier`s if we had a party plugin that had a "reducing damage" effect based on the number of players joined in the party! All of this is as extensible as possible.

#### Ok... well, it's clear that you've given a lot of thought about `EntityDamageEvent`, but what about other `CauseTracked` events?

Similar to the `EntityDamageEvent` and `EntityPreDamageEvent`, we can safely say that most `CauseTracked` events are going to receive similar treatment with regards to adding additional *sources*/*causes*/*reasons* whatever you want to call them. So far, I've only been able to seriously work on the damage events, however, my thought process is that with `EntitySpawnEvent`, it is very easy to apply the same idea of having the following:

`EntitySpawnEvent`:
 - `SpawnCause` : An abstract *root cause* that aids in helping further understand the `Cause` for why an `Entity` is being spawned
 - `SpawnType` : Another `CatalogType` that aims to further simplify the *type* of `SpawnCause`, more specifically, when an `Entity` is spawned by a `MobSpawner`, the `SpawnCause` would be an instance of a `MobSpawnerSpawnCause` with the associated `MobSpawnerData` that defined the `Entity` to be spawned

`EntityTeleportEvent`:
- `TeleportType` : Yet again, a helper `CatalogType` that aims to make it understandable why an `Entity` is being teleported
- `TeleportCause` : Another *root cause* for the `Entity` to be teleporting, can range from `EntityTeleportCause` (an `Enderman` teleporting due to rain), `TeleporterTeleportCause` (a portal performing the teleport as per vanilla mechanics)

The list will grow as the PR is worked on and time goes on.

#### So, will this make it easy for me to know if a `BlockChangeEvent` was caused by a `Player` using some bonemeal on a sapling?

In a short answer: yes. 
The long answer: The `Cause` would indeed have a `Player` object, and the secondary `cause` would be the `ItemStack` that "aided" in the cause for the `BlockChangeEvent` to take place. What is better is that with the knowledge of the `Player`, the `ItemStack`, and any other objects placed into the `Cause`, it becomes very simple for you to decide on changing the `BlockChangeEvent` for whatever purpose.

#### How about throwing custom events? I don't really understand what I'm supposed to do here if I want to allow plugins to hook into my custom `EntityTeleportEvent`.

To be honest, that is why I simplified `Cause` to replicate `Optional`. When you are wanting to teleport an `Entity` with a `Cause`, you can include the *root* cause, say the `Entity` called a `Command`, and the `Command` was entered through a `Sign`, you now have *two* objects to include in your `Cause`. However, let's say you wanted to expand and provide some custom `TeleportCause`s, you could extend `TeleportCause` and provide the extended `Cause` to be the `Sign`, and likewise you could include your own custom marker object so that in the event you have a listener for `EntityTeleportEvent`, you can safely ignore the event if your custom marker object is included.

#### How about X event?

Well, as I said before, the PR is still a work in progress, so I haven't finished designing the `Cause` for every `CauseTracked` event.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -28,6 +28,8 @@ import com.flowpowered.math.vector.Vector3d;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.manipulator.mutable.TargetedLocationData;
+import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.event.cause.entity.damage.source.DamageSource;
 import org.spongepowered.api.util.Identifiable;
 import org.spongepowered.api.util.RelativePositions;
 import org.spongepowered.api.world.Location;
@@ -277,5 +279,17 @@ public interface Entity extends Identifiable, DataHolder, DataSerializable {
      * within one game tick.
      */
     void remove();
+
+    /**
+     * Damages this {@link Entity} with the given {@link Cause}. It is
+     * imperative that a {@link DamageSource} is included
+     * with the cause for maximum compatibility with plugins and the game
+     * itself.
+     *
+     * @param damage The damage to deal
+     * @param cause The cause of the damage
+     * @return True if damaging the entity was successful
+     */
+    boolean damage(double damage, Cause cause);
 
 }
--- a/src/main/java/org/spongepowered/api/event/cause/Cause.java
+++ b/src/main/java/org/spongepowered/api/event/cause/Cause.java
@@ -26,10 +26,16 @@ package org.spongepowered.api.event.cause;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
 import org.spongepowered.api.entity.Entity;
-import org.spongepowered.api.event.cause.reason.Reason;
-import org.spongepowered.api.world.Location;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.event.cause.entity.damage.source.DamageSource;
+import org.spongepowered.api.event.cause.entity.spawn.SpawnCause;
+
+import java.util.Arrays;
+import java.util.List;
 
 import javax.annotation.Nullable;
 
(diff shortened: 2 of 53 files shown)


src/main/java/org/spongepowered/api/event/entity/util/damage/DamageTypes.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+public final class DamageTypes {
+
+    public static final DamageType CONTACT = null;
+    public static final DamageType EXPLOSIVE = null;
+    public static final DamageType FIRE = null;
+    public static final DamageType MAGIC = null;
+    public static final DamageType PROJECTILE = null;
+    public static final DamageType PLUGIN = null;
```
simon816: I prefer the name `CUSTOM`

ZephireNZ: Agreed - if a mod uses a new damage type, it would like use this one, and mods != plugins.

gabizou: If a mod uses a new damage type, it wouldn't be `PLUGIN`, it would have it's own `id` and everything.

ryantheleach:
Would/Could this tie into https://github.com/SpongePowered/SpongeAPI/issues/556 ? Setting a block with a cause on behalf of a player?

Faithcaio:
BlockMoveEvent, EntityExplosionEvent and EntityDeathEvent(breaking Vehicles/Hanging) are not Cancellable.
I think they should be.

ryantheleach:
How would you prevent an entity from dying when it already has 0 health? Wouldn't you rather listen for the entity damage event that would kill the entity and prevent damage?

src/main/java/org/spongepowered/api/util/Tuple.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package org.spongepowered.api.util;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.base.Objects;
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * A tuple of objects. This can be considered a {@link Pair}
```
gratimax: Then what's wrong with ImmutablePair?

src/main/java/org/spongepowered/api/event/entity/EntityDamageEvent.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Gets a list of simple {@link Tuple}s of {@link DamageModifier} keyed to
+     * their representative {@link Function}s. All {@link DamageModifier}s are
+     * applicable to the entity based on the {@link DamageSource} and any
+     * possible invulnerabilities due to the {@link DamageSource}.
+     *
+     * @return A list of damage modifiers to functions
+     */
+    List<Tuple<DamageModifier, Function<? super Double, Double>>> getModifiers();
```
gratimax: Why do you need a tuple here? Why not a Map? Also, I don't see what use there is for `? super Double` considering that there are very few supertypes of double (probably just Number, Comparable, Serializable, and Object).

src/main/java/org/spongepowered/api/event/cause/Cause.java
```diff
@@ -48,52 +52,211 @@
  * some blocks, but tracing this event would be too complicated and thus
  * may not be attempted.</p>
  */
-public class Cause {
+@SuppressWarnings("unchecked")
+public abstract class Cause {
```
gratimax: You keep the "cause" nomenclature for stuff like `getAllCauses`, yet the class name is Cause. Maybe it deserves a different name like CauseSet?

Also, I'm not really sure what the difference is between an absent cause and a present one that has no "elements".

src/main/java/org/spongepowered/api/data/manipulator/entity/HealthData.java
```diff
@@ -33,14 +33,6 @@
 public interface HealthData extends DataManipulator<HealthData> {
 
     /**
-     * Subtracts from the health by the given amount.
-     *
-     * @param amount The damage amount
-     * @return This instance, for chaining
-     */
-    HealthData damage(double amount);
```
gratimax: Seems like an unrelated change that is out of scope.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -246,4 +248,16 @@
...
+     * Damages this {@link Entity} with the given {@link Cause}. It is
+     * imperative that a {@link DamageSource} is included
+     * with the cause for maximum compatibility with plugins and the game
+     * itself.
+     *
+     * @param damage The damage to deal
+     * @param cause The cause of the damage
+     * @return True if damaging the entity was successful
+     */
+    boolean damage(double damage, Cause cause);
```
gratimax: I am unsure why this belongs in now `Entity` instead of `HealthData`, putting it in a manipulator seems perfect. If not `HealthData` then `DamageableData`.

src/main/java/org/spongepowered/api/event/cause/Cause.java
```diff
@@ -48,52 +52,211 @@
...
+     */
+    public abstract ImmutableList<Object> getAllCauses();
+
+    /**
+     * Returns {@code true} if {@code object} is a {@code Cause} instance, and either
+     * the contained references are {@linkplain Object#equals equal} to each other or both
+     * are absent.
+     */
+    @Override
+    public abstract boolean equals(@Nullable Object object);
```
gratimax: I'm not really sure what these overrides are adding, the relevant documentation is already in Object.equals

src/main/java/org/spongepowered/api/event/cause/Cause.java
```diff
@@ -48,52 +52,211 @@
...
+
+    /**
+     * Returns a hash code for this instance.
      */
-    public Optional<Reason> getReason() {
-        return this.reason;
+    @Override
+    public abstract int hashCode();
```
gratimax: Same here as in `equals`

Faithcaio:
I couldn't find an EntityDamageEvent. 
There is one for Living but what I need is an event where entities without Health die/are removed.
(VehicleBreak and HangingBreakEvents in Bukkit)

src/main/java/org/spongepowered/api/event/cause/Cause.java
gabizou: A present one always has at least one cause.

src/main/java/org/spongepowered/api/event/cause/Cause.java
gabizou: Ensures that the methods are properly overridden instead of ignored like most classes/interfaces in the API.

gabizou:
> I couldn't find an EntityDamageEvent. 
> There is one for Living but what I need is an event where entities without Health die/are removed.
> (VehicleBreak and HangingBreakEvents in Bukkit)

https://github.com/SpongePowered/SpongeAPI/pull/712/files#diff-775b410055a6c5940497d25e4bcdff8dR43

src/main/java/org/spongepowered/api/event/entity/EntityDamageEvent.java
gabizou: > Why do you need a tuple here, and not some sort of Map? Also, I don't see what use there is for ? super Double considering that there are very few supertypes of double (probably just Number, Comparable, Serializable, and Object).

Because the order of the functions applied need to be kept. No map implementation retains ordering and allows manipulating the order of elements.

Faithcaio:
So this one will be fired for non Living Entities too?

src/main/java/org/spongepowered/api/util/Tuple.java
gabizou: That name works as well.

gratimax: No, I mean the ImmutablePair class in Guava.

src/main/java/org/spongepowered/api/entity/Entity.java
gabizou: I personally don't find it to be useful to store an extra variable in the manipulator for throwing an event since this is a function call to throw a new damage event.

src/main/java/org/spongepowered/api/event/cause/Cause.java
gratimax: Then what's the benefit in having an additional subclass and adding additional complexity instead of simply wrapping List and adding an `isPresent` method? It seems to me that functionality can be completely covered by a simpler overall Cause implementation.

src/main/java/org/spongepowered/api/event/cause/Cause.java
gratimax: But there's only two subclasses...

gabizou:
> So this one will be fired for non Living Entities too?

That is the point of any `EntityEvent`, it is thrown for any `Entity`. The `LivingChangeHealthEvent` is specifically for any `Living` entities, etc.

src/main/java/org/spongepowered/api/event/cause/Cause.java
gabizou: So? Optional only has two sub classes, why does it matter here?

src/main/java/org/spongepowered/api/event/cause/Cause.java
gabizou: I'd rather that if you have an empty cause, it will never have to perform if checks for emptiness. I don't think it's necessary to rely on a `List` when `Cause` is final once created, there's no array instance for empty `Cause`s, there's no worry of anything. It's like `Optional` but for causes.

src/main/java/org/spongepowered/api/util/Tuple.java
gabizou: > No, I mean the ImmutablePair class in Guava.

Why can't I find it?

simon816: Looks like it's actually from commons-lang https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/tuple/ImmutablePair.html

gabizou: Which is what I'm avoiding because of the comment I made three comments above this one.

gabizou:
After a [lengthy discussion](https://gist.github.com/gabizou/1b497a142615ec1136fc) with @bloodmc and @modwizcode, I've come to the conclusion of the following agreements for this PR:

- [ ] Causes will have two types of lists: 
 - [ ] A list of Events for the event chain
 - [ ] A list of the Cause objects
 - [ ] Additional methods for querying said events in the Cause object chain:
   - [ ] getFirstEvent(Class<E extends Class<Event>>) returning the Event in the event chain
    - [ ] getLastEvent(Class<E extends Class<Event>>) returning the E event last in the event chain
    - [ ] getAllEventsOf(Class<E extends Class<Event>>) returning a List<E> of all compatible events in the event chain 
- [ ] Each event that can have a custom direct cause would have extensive documentation of such causes so that querying for them in the `Cause` object is simple.
- [ ] Custom event causes (such as `SpawnCause` or `DamageReason`) that include extra information as necessary related to the cause itself will remain as they are: included in the `Cause` list of objects.
- [ ] Custom Cause interfaces would be superinterfaced for a simple `getEventCauseUniqueId():UUID` 
- [ ] Events included in the event chain of `Cause` would become immutable by virtue of calling any modification methods resulting in an `UsupportedOperationException`. This will be possible through the event factory and an annotation provided by @Aaron1011.

gabizou:
This has been merged into the [refactor/event-names](https://github.com/SpongePowered/SpongeAPI/tree/refactor/event-names/) branch.

Issue #710, opened 2015-06-01
Possible ways to compare Cataloged Types?


saladoc:
Please note that this is *not* a best practice question. This is to clarify if plugins will horribly break if the wrong comparison method is used / people have a choice to use what they deem more readable.

Is it guaranteed that the identity check (`someBlock.getType() == BlockTypes.STONE`) will return the same as the equality check (`someBlock.getType().equals(BlockTypes.STONE)`) for cataloged types? 

When I asked this question on IRC a couple of weeks ago specifically aimed at `BlockType`s, the answer was 'yes'. Now I'm asking to clarify that this shall be true for all cataloged types over all implementations, so that the docs may accurately represent that stance.

Xemiru:
This purely depends on whether or not `someBlock.getType()` returns `Blocktypes.STONE` and not a new instance that it implements on its own (which is a bad idea unless you're writing for mods). The answer is yes since it points to the same reference assuming the method _does_ return the former.

Because this is in the API, then it should be true across all implementations assuming all implementations follow the above rule to return the reference given by `BlockTypes`. Alternatively, if using this isn't a trusted way (I honestly still find it unreliable to use; it breaks if something changes), then one can also just use `someBlock.getType().getName()` and compare that to `BlockTypes.MATERIAL.getName()`.

saladoc:
A little reminder that this question still stands. @gratimax already pointed out that no matter the answer, using `equals()` is to be preferred as a best practice. The docs will reflect that. Still open questions are:

1. Best practices aside, is it possible to use `==` for all `CatalogedType`s introduced by Sponge itself or will it break plugins?
2. Is the `id` supplied to the `GameRegistry#getType(Class, String)` method case sensitive? 
3. How can custom `CatalogedType`s be added to the game registry?

ST-DDT:
May be a little late:

> Best practices aside, is it possible to use `==` for all `CatalogedType`s introduced by Sponge itself or will it break plugins?

Yes, this is possible for most `CatalogedType`s. Although i cannot guarantee it for all (future) catalog types/implementations. But `BlockType`s and `EntityType`s will work usually.

> Is the `id` supplied to the `GameRegistry#getType(Class, String)` method case sensitive?

Yes, but i'm pretty sure that this will change in the future. https://github.com/SpongePowered/SpongeAPI/pull/717

> How can custom CatalogedTypes be added to the game registry?

This is currently not possible API wise. Why would you like to do this?
[...]

PR #717, opened 2015-06-07
Implement provider based GameRegistry

modwizcode:
Currently the `GameRegistry` class is a monster, especially implementation wise. All of the different builder instances, factories and despite the class's name, only a few registries are all combined into a massive single class. 

This pull request aims to change that massively convoluted system with a series of modules, called `Provider`s that provide (hence the name) a subset of what `GameRegistry` did. The system isn't meant to be used by plugins, so all Providers are expected to be available upon request.

I've implemented the changes in a branch of SpongeCommon and was able to successfully start up a server after making numerous changes to replace hard-coded registry calls.

I know this is a drastic change, in some ways more drastic than the Data API simply because of what it encompasses, but it's needed to prevent future cruft from the over packed class.


ST-DDT:
Why do you have a Factory and a Provider for everything?
IMO The factory is an implementation detail, how to implement the provider.

You have also tried to split everything based on type builder, factory, statistic, career.
But it would be very convenient, if i could get a `StatisticBuilder` from the `StatisticProvider`.
And a list of all `Career`s from the `CareerProvider`.

Why do you still need the `GameRegistry` after this?

There are also some places where you replaced the text "factory" with "provider" in classes that explicitly call themselves a factory.
https://github.com/SpongePowered/SpongeAPI/pull/717/files#diff-b5be1bf622309d8523a3e70772fabaf8L67

modwizcode:
That was fixes in a later commit. I also don't have any class or interface
called `Factory`

ST-DDT:
> I also don't have any class or interface called `Factory`

[`src/main/java/org/spongepowered/api/util/factory/Factory.java`
](https://github.com/SpongePowered/SpongeAPI/pull/717/files#diff-eb0747589682ef08cc03f5838e7c493bR30) or am I missing something here?

modwizcode:
Sorry that's a commit error I'll have to fix that. Those were deleted.

modwizcode:
GameRegistry is still a good class for things that actually need a generic registry classes. There are a few things that don't fit into the provider model which will remain in `GameRegistry`

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -25,57 +25,13 @@
 package org.spongepowered.api;
 
 import com.google.common.base.Optional;
-import org.spongepowered.api.attribute.AttributeBuilder;
-import org.spongepowered.api.attribute.AttributeCalculator;
-import org.spongepowered.api.attribute.AttributeModifierBuilder;
-import org.spongepowered.api.block.BlockType;
-import org.spongepowered.api.data.DataManipulatorRegistry;
-import org.spongepowered.api.data.ImmutableDataRegistry;
 import org.spongepowered.api.data.type.Career;
```
ST-DDT: Some imports survived although they are no longer required.

src/main/java/org/spongepowered/api/resourcepack/ResourcePackFactory.java
```diff
@@ -28,7 +28,7 @@
 import java.net.URL;
 
 /**
- * A factory for creating {@link ResourcePack}s.
+ * A provider for creating {@link ResourcePack}s.
```
kashike: I have a feeling your find & replace borked this, along with other event-related classes.

modwizcode: Just a bit. I'll go through and revise javadoc after the core content and
concept are accepted or rejected.

modwizcode:
I have made the fixes locally and will push later tonight.

gratimax:
You probably want to rename TextFactory to TextProvider as well. Nice work so far :+1:

modwizcode:
Thanks @gratimax, what are the changes to TextFactory you're interested in? It wasn't already in GameRegistry iirc so I didn't change it. I was going to change some existing classes to meet the Provider system, but it seemed too messy as most providers were in one package and a few are scattered throughout the API, which just seemed like an unwise decision.

gratimax:
I pretty much am only interested in the rename which is in the scope of this PR. It's not really a factory and the name "provider" just fits better. I don't think we have much use of giving the instance to plugin authors, but maybe @Minecrell has an opinion here.

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -75,6 +77,15 @@
     GameRegistry getRegistry();
 
     /**
+     * Gets an instance of the provided {@link Provider} class.
+     *
+     * @param providerClass The class of the type of {@link Provider}
+     * @param <T> The type of {@link Provider}
+     * @return An instance of the requested provider type
+     */
+    <T extends Provider> T getProvider(Class<T> providerClass);
```
simon816: Why is this here __and__ in `GameRegistry`? Should only be in one place.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -92,190 +48,13 @@
...
-     *
-     * <p>The presented {@link CatalogType}s may not exist in default catalogs
-     * due to various reasons including but not restricted to: mods, plugins,
-     * game changes.</p>
-     *
-     * @param typeClass The class of {@link CatalogType}
-     * @param <T> The type of {@link CatalogType}
-     * @return A collection of all known types of the requested catalog type
-     */
-    <T extends CatalogType> Collection<T> getAllOf(Class<T> typeClass);
```
simon816: I think this should stay because it will just make for longer method call chains and `getProvider(CatalogTypeProvider.class).getAllOf` may not be immediately obvious

src/main/java/org/spongepowered/api/Game.java
modwizcode: Because technically a provider is contained in the registry, but as it was
already convoluted to get the game registry and then retrieve the instance.

By using providers it makes more sense to get them from the Game instance.
However it still made some sense to leave them with the registry itself. I
think I'll remove it. I meant to ask about opinions for the inclusion there
but I forgot.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -92,190 +48,13 @@
...
-     * <p>Some types may not be available for various reasons including but not
-     * restricted to: mods adding custom types, plugins providing custom types,
-     * game version changes.</p>
-     *
-     * @param typeClass The class of the type of {@link CatalogType}
-     * @param id The string id of the catalog type
-     * @param <T> The type of catalog type
-     * @return The found catalog type, if available
-     */
-    <T extends CatalogType> Optional<T> getType(Class<T> typeClass, String id);
```
simon816: Same again for getType as `getProvider(CatalogTypeProvider.class).getType` is less obvious

src/main/java/org/spongepowered/api/GameRegistry.java
modwizcode: But it woudn't make any sense to put there. It's not that much longer and I think a lot of our API isn't immediately obvious which is why we use the javadocs and docs site to convey the needed information. This could easily be mentioned in the CatalogedBy annotation or other parts of the docs.

modwizcode:
@SpongePowered/developers I'd like to get this merged, so if people could take a look and mention changes to make I'd appreciate it. The faster this is merged the faster I can update SpongeCommon to use the new system.

Deamon5550:
I"m very much against this change. You haven't really 'solved' the problem you've just moved things around and hidden it behind another layer. Perhaps if your providers were discrete to an individual part of the api (such as if you had a `StatisticsProvider` that had all the getters, builders, factories etc. relevant to statistics), but right now things like the [BuilderProvider](https://github.com/SpongePowered/SpongeAPI/pull/717/files#diff-d963bec62e2dae4434cb83b6b68f173aR47) really aren't helping the issue of "the GameRegistry class is a monster".

Additionally your `getProvider` method needs to return an optional.

modwizcode:
@Deamon5550 It actually doesn't need to be optional, did game registry ever return itself as optional? No, providers aren't modified or set by plugins, Sponge should always return a non-optional provider. I had them optional briefly before realizing that made no sense for an internally controlled system. A null provider is a fatal error.

I appreciate the feedback, but I don't see how I haven't "solved" the problem. Maybe it's not perfect, but it's much better than GameRegistry. Of course I've just moved things around, the GameRegistry is the one way for plugins to get instances, those methods still need to exist.

 Nothing is preventing me from moving related builders, factories and the like into their own providers, I just split it up the way the GameRegistry was organized.

Deamon5550:
My biggest issue is just that your providers are all over the place. If I use statistics as an example again I get the simple statistics from the `CatalogTypeProvider`, I get some of the other statistics from `StatisticProvider `, and I get the builders for them from the `BuildersProvider`. Three separate places all while just working with statistics. If you want to make the move to providers then I think your goal for the system should be to get things into their own self contained providers so that if I'm working with statistics, all I need is the `StatisticsProvider`.

modwizcode:
Yeah that makes a lot of sense. I'm not exactly sure what to do with `CatalogTypeProvider` since that's used so often. It doesn't really belong in StatisticsProvider for instance. I agree builders should be fit into the relevant other providers if there are other more specific providers related to that builder.

Deamon5550:
Add back a simple `Statistic get(String name);` method to the `StatisticsProvider` and still have the `CatalogTypeProvider` for more abstract fetching of data (and stuff with no provider). Although the act of having to retrieve the CatalogTypeProvider when dealing with catalog types (which you have to do fairly regularily) is rather annoying, I wonder if that is one thing which is better left out of the providers and left in GameRegistry.

modwizcode:
It's almost the same length, but much more clear to get the provider and
get the type as it to get the registry from the game and get the catalog
type.

On Mon, Jun 22, 2015 at 2:34 PM, Nick Conde <notifications@github.com>
wrote:

> Add back a simple Statistic get(String name); method to the
> StatisticsProvider and still have the CatalogTypeProvider for more
> abstract fetching of data (and stuff with no provider). Although the act of
> having to retrieve the CatalogTypeProvider when dealing with catalog types
> (which you have to do fairly regularily) is rather annoying, I wonder if
> that is one thing which is better left out of the providers and left in
> GameRegistry.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpongePowered/SpongeAPI/pull/717#issuecomment-114231173>
> .
>



-- 

Starbuck Johnson, Java Developer
www.modwiz.com | starbuck@modwiz.com

Deamon5550:
In that case I think that you should remove the GameRegistry entirely, `getProvider` is in Game, `Collection<String> getDefaultGameRules()` should be in a provider, and `registerWorldGeneratorModifier(WorldGeneratorModifier modifier)` can move to Game.

modwizcode:
But `registerWorldGeneratorModifier(WorldGeneratorModifier modifier)` doesn't seem like it really belongs in `Game`

Deamon5550:
Well otherwise its literally the only thing in GameRegistry so find a home for it because it certainly doesn't belong in GameRegistry anymore.

ST-DDT:
I guess we can move it out of GameRegistry to any location until we find a good place for it.

How about `Server`? There already are some `World` methods in there.

ST-DDT:
Two suggestions from https://github.com/SpongePowered/SpongeAPI/issues/710
* Make get `CatalogedType`s by id case insensitive (and explicitly state this in the docs)
* Add the possibility to add custom `CatalogedType`s

Zidane:
This needs to be redone (same as Maps) and @modwizcode isn't available so closing.

modwizcode:
?

PR #725, opened 2015-06-14
Add a Catalogue for BlockStates.

me4502:
This PR creates a catalogue system for all blockstates in Minecraft.


Deamon5550:
So nowhere in those 7800 lines did you stop to think that this was a really stupid idea?

me4502:
@Deamon5550 I didn't write it out by hand, I just generated a list.

I was writing it up so that we had something to discuss off.

me4502:
(For the curious, http://pastebin.com/32DcvV7D)

MikaalSky:
So, tell me why this is stupid?
Length..?

me4502:
Length is the most obvious reason. Most IDEs would fail to generate autocomplete definitions.

MikaalSky:
Got it.

ST-DDT:
However having such a list (split into smaller pieces or provided as a map) could be useful.

PR #729, opened 2015-06-18
Fixes all CheckStyle warnings

gabizou:
There were a lot of various OCD fixes to make the API build entirely clean again.


src/main/java/org/spongepowered/api/util/weighted/WeightedEntity.java
```diff
@@ -120,4 +120,6 @@ public boolean equals(Object obj) {
         return this.weight == object.weight;
     }
 
+
+
```
simon816: Unnecessary extra ines

Issue #573, opened 2015-04-18
 Allow generating a single tree at a single location


zml2008:
Would also be good to be able to pass a custom Extent/some sort of proxy object to be able to record block changes performed. Bukkit had this as a BlockChangeDelegate.

kenzierocks:
Tree as in sapling tree, or tree as in TreeMap?

Deamon5550:
@kenzierocks sapling tree

zml2008:
![Imgur](http://i.imgur.com/hO7v56O.png)
This kind of tree.

Deamon5550:
This will be added with #734, you'll be able to fetch the PopulatorObject representing the tree and placed it where ever you want. For example `BiomeTreeTypes.OAK.getPopulatorObject().placeAd(...)`.

PR #734, opened 2015-06-24
World Generation Phase 1 - Populators

### LINKED ISSUE #582: Problems with the global biome populators
So as it stands biomes have a series of populators, and these populators are defined per-biome. The biomes are defined globally which means that if you modify the populators of a particular biome it will have effect across all worlds. Now this isn't necessarily a bad thing however if someone is wanting to make a world generator which uses the default biomes but perhaps slightly modified. For example a skylands generator world which only wants to modify the y-levels that ores appear at would have negative side effects on the generation of the overworld as well.

**Some potential solutions:**
- Define the populators of biomes in the WorldGenerator rather than the BiomeType. The obvious solution.
- Allow a WorldGenerator to override the populators of specific biomes. A slightly messier solution as the biome's populators would now be defined in two places both of which would need to be checked.
- Allow the registration of new biomes. With this rather than editing the standard biomes the skylands generator would create its own new biomes with the modifications. Allowing the registration of new biomes is not a highly invasive change and is doable without client modifications. However there is an  implementation restriction of 256 biomes (and there are 63 vanilla ones) which should be kept in mind before we start requiring this as a solution to wanting to modify biomes.

Allowing the registration of new biomes is something we're going to want to allow anyway eventually. However in my opinion we'd be better off going with both the first and the third option in that case.

Deamon5550:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/253) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/405)
This contains API changes for the first phase of populators.

Closes #582, #573

Changes:
- Separation of biome generation parameters out of `BiomeType` and into `BiomeGenerationSettings`
- Move `BiomeGenerationSettings` to per-generator basis, getting the settings from a generator that does not have then specifically set will construct a set of settings from the defaults
- Modified populators to make use of new `VariableAmount`, `SeededValue`, and `WeightedTable` utilities to allow greater flexibility
- Removed JungleBush populator and added it as a tree type
- Combined RandomFire and RandomLiquids populators into a more generic RandomBlock poppulator
- Removed the PopulatorObjects utility and re-added it as a RandomObject populator
- Added `SeededVariableAmount` to allow specifing amounts that vary depending on some seed input (this is used by the GroundCoverLayers to seed the layer depth from the stone noise generators)
- Added MushroomType to represent the two large mushroom types
- Added getters to BiomeTreeType and MushroomType to get the PopulatorObject which represents them (this can also be used by plugins to easily place these objects where ever they want)
- Minor fixes to VariableAmount
- Expansion of the `WeightedCollection` into the more expansive LootTable/WeightedTable/ChanceTable for use in a variety of APIs including the upcoming loot tables in 1.9
- Minor changes and usability fixes to most populators

Known Issues / TODO:
- [x] Weighted package need hashCode / equals / toString
- [x] Weighted package need more usability methods
- [x] Weighted package needs docs
- [x] Need to examine where more populator objects can be leveraged
- [x] remove temporary downgrade of configurate (fix for eclipse)
- [x] ~~Update skylands for generator changes (especially ground cover layers and variable amounts)~~


src/main/java/org/spongepowered/api/world/gen/populator/Ore.java
```diff
@@ -77,35 +79,35 @@
...
-     * Sets the minimum height that the ore can generate at.
+     * Sets the height that the ore can generate at.
      * 
-     * @param min The new minimum height
+     * @param height The new height
      */
-    void setMinHeight(int min);
-
+    void setHeight(VariableAmount height);
+
```
simon816: remove these spaces

Zidane:
@Deamon5550 

What is the status of the impl of this?

build.gradle
```diff
@@ -37,9 +37,9 @@ dependencies {
     compile 'com.google.inject:guice:4.0'
 
     // Configuration
-    compile 'ninja.leaping.configurate:configurate-hocon:3.0'
-    compile 'ninja.leaping.configurate:configurate-gson:3.0'
-    compile 'ninja.leaping.configurate:configurate-yaml:3.0'
+    compile 'ninja.leaping.configurate:configurate-hocon:2.0'
```
Deamon5550: this is a temporary fix for eclipse, will remove before merging

Deamon5550:
Impl is somewhat stable, few known issues listed here and on impl PRs but testing can begin. Any errors or places where generation deviates from vanilla behaviour should be reported on the SpongeCommon PR.

src/main/java/org/spongepowered/api/event/block/NotifyNeighborBlockEvent.java
```diff
@@ -26,6 +26,7 @@
 package org.spongepowered.api.event.block;
 
 import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.block.BlockType;
```
Deamon5550: probably a merge conflict artifact, remove

src/main/java/org/spongepowered/api/util/Functional.java
```diff
@@ -83,7 +83,8 @@ private Functional() {
     }
 
     /**
-     * Get the value of an {@link Optional} as either a zero- or one-element immutable set.
+     * Get the value of an {@link Optional} as either a zero- or one-element
```
Deamon5550: uneeded formatting changes

src/main/java/org/spongepowered/api/world/biome/BiomeGenerationSettingsBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.spongepowered.api.world.gen.GenerationPopulator;
+import org.spongepowered.api.world.gen.Populator;
+
+import java.util.Collection;
+
+/**
+ * A builder for creating {@link BiomeGenerationSettings}.
+ */
+public interface BiomeGenerationSettingsBuilder {
```
Deamon5550: Remove this builder, no longer needed

src/test/java/org/spongepowered/api/util/weighted/TestWeightedTable.java
```diff
@@ -0,0 +1,44 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.weighted;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.List;
+
+public class TestWeightedTable {
```
Deamon5550: Need a lot more unit tests...

Deamon5550:
@SpongePowered/developers If there are no more issues this will be merged tonight.

src/test/java/org/spongepowered/api/util/weighted/TestWeightedTable.java
gabizou: Did this get a lot more unit tests (aside from demonstration in implementation)?

ryantheleach:
https://github.com/sk89q/WorldEdit/blob/a4455e875e9f2588af3b2ac0d0a6513343770aac/worldedit-forge/src/main/java/com/sk89q/worldedit/forge/ForgeWorld.java#L263

How WE - Forge regens chunks.

Ugly... but I guess it works.

Deamon5550:
@ryantheleach I would only need to do something like that if the generation has been replaced wholesale by a mod. Otherwise I can leverage the fact that our api can populate an arbitrary area to do the generation without creating an entire world for it.

Also this is totally the wrong issue for this, there is an open issue for regenerating chunks.

ryantheleach:
@Deamon5550 Not sure how I managed to get my tabs crossed, was supposed to be posted on that ticket.

Issue #74, opened 2014-09-10
Doxygen for JavaDocs


AedDev:
I considered that Doxygen would be nice for Documentation. I could host it and keep it up2date.

PR #735, opened 2015-06-26
Add methods to ignore players when skipping the night

caseif:
This PR adds the API methods `isSleepingIgnored()Z` and `setSleepingIgnored(Z)V` to the `Player` interface. When this toggle is set, the `Player` will not be considered when determining whether the night should be skipped due to all players being asleep.

These methods are implemented by [SpongeCommon #74](https://github.com/SpongePowered/SpongeCommon/pull/74).

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/player/Player.java
@@ -135,13 +135,13 @@ public interface Player extends Human, User, LocatedSource, RemoteSource, Viewer
     GameModeData getGameModeData();
 
     /**
-     * Returns whether this {@link Player} will be ignored when checking whether
-     * to skip the night due to players sleeping. The time in a world will be
+     * Gets whether this {@link Player} will be ignored when checking whether to
+     * skip the night due to players sleeping. The time in a world will be
      * advanced to day if all players in it either are sleeping or have this
      * tag.
      *
      * @return Whether this {@link Player} will be ignored when checking whether
-     *         to skip the night
+     *     to skip the night
      */
     boolean isSleepingIgnored();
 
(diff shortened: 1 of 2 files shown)


simon816:
This should be on `Player`, because humans (like other creatures) don't affect the world sleep. The sleep is only for players.

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -134,4 +134,26 @@
      */
     GameModeData getGameModeData();
 
+    /**
+     * Returns whether this {@link Player} will be ignored when checking whether
```
modwizcode: Gets

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -134,4 +134,26 @@
...
+     */
+    boolean isSleepingIgnored();
+
+    /**
+     * Sets whether this {@link Player} will be ignored when checking whether
+     * to skip the night due to players sleeping. The time in a world will be
+     * advanced to day if all players in it either are sleeping or have this
+     * tag.
+     *
+     * @return Whether this {@link Player} will be ignored when checking whether
```
modwizcode: This is a parameter not a return

src/main/java/org/spongepowered/api/entity/player/Player.java
```diff
@@ -134,4 +134,26 @@
...
     GameModeData getGameModeData();
 
+    /**
+     * Returns whether this {@link Player} will be ignored when checking whether
+     * to skip the night due to players sleeping. The time in a world will be
+     * advanced to day if all players in it either are sleeping or have this
+     * tag.
+     *
+     * @return Whether this {@link Player} will be ignored when checking whether
+     *         to skip the night
```
gabizou: indentation should be 4, not matching the start of the sentence.

like so:

```java
    /* ...
     * @return Whether this {@link Player} will be ignored when checking whether
     *     to skip the night
     */

```

modwizcode:
Seems good.

PR #740, opened 2015-06-28
Fix return type for Ban.User.getUser()

twizmwazin:
Currently, the method returns Ban.User rather than User, making the method fairly useless. This corrects the issue, returning the correct User type.


simon816:
This is actually mentioned in the minor issues list https://github.com/SpongePowered/SpongeAPI/issues/221#issuecomment-100543585
mind you, that list is kind of hard to know what has/has not been fixed.

twizmwazin:
@simon816 I saw when looking to see if there was a proper issue for it.

simon816:
This will be fixed on the next sweep of minor issues (#221)

PR #741, opened 2015-06-28
Adds world UUID to Location toContainer

viveleroi:
A UUID is naturally the preferred value for future retrieval. I'll leave the decision as to whether to the world name needs to remain here to others.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Location.java
+++ b/src/main/java/org/spongepowered/api/world/Location.java
@@ -691,7 +691,9 @@ public final class Location implements DataHolder {
     public DataContainer toContainer() {
         DataContainer container = new MemoryDataContainer();
         if (getExtent() instanceof World) {
-            container.set(of("WorldName"), ((World) getExtent()).getName());
+            World world = (World) getExtent();
+            container.set(of("WorldName"), world.getName());
+            container.set(of("WorldUUID"), world.getUniqueId());
         }
         container.set(of("BlockType"), this.getExtent().getBlockType(getBlockPosition()).getId());
         container.set(of("x"), this.getX());


DDoS:
Since we're talking about the toContainer method, should we be adding the block type? The block type at a location can change, but this won't be reflected in the container, and there is no guarantee that a location even points to a valid block. This method could throw a PositionOutOfBounds exception. The same applies to the manipulator. Location is also used for entites, where this data isn't relevant.

gabizou:
I'll bundle this in with #685 such that `Location#toContainer` only stores the `World#getName()` and `World#getUniqueId()` if the `Extent` is indeed a `World`, and of course stores the `Vector3d`. Ultimately, all a `Location` should care about is the information it absolutely "knows" about, instead of the `BlockState` information or `BlockSnapshot`.

DDoS:
I think we need some identifier for Extents World can use the name and/or UUID, Chunk can use the this identifier (from its world) plus some its coordinates. I'll make a separate issue for this.

gabizou:
> I think we need some identifier for Extents World can use the name and/or UUID, Chunk can use the this identifier (from its world) plus some its coordinates. I'll make a separate issue for this.

Agreed, what I think we can do is have the `Extent` provide some sort of `DataView` filled with the identifying information, such as a `Chunk` providing the `World`, `ChunkCoordinate`s, etc.

Zidane:
I don't see the value of the WorldName namely cause there is nothing stopping that from changing.

WorldName is derived from foldername on the server and I could easily change "world" to "Midgar" and change server.properties to "Midgar" for default world. Server would load up the map without a second thought except all saved locations in plugins that made the mistake of using the container's name will be wrong now.

Put simply, UUID is enough.

gabizou:
> Put simply, UUID is enough.

Ok, considering it's been changed [here](https://github.com/SpongePowered/SpongeAPI/blob/feature/value-api/src/main/java/org/spongepowered/api/world/Location.java#L671) in #685, would we close this? Or perhaps expand on this with a sort of `Extent#writeDataToken(DataView)` filling all necessary information of the extent, in this case the `World#getUniqueId()`, or for `Chunk`, the `Vector3i` position.

Zidane:
@viveleroi 

https://github.com/SpongePowered/SpongeAPI/commit/29d1df81e69f0ea9fb0cb2877408bed0f3a3f114

I manually took care of this here. Thanks for bringing this to our attention.

PR #744, opened 2015-06-30
Expand the Command API. Closes #473

### LINKED ISSUE #473: CommandSource for /execute and signs
Right now we have `Player`, `ConsoleSource`, `CommandBlockSource` and `RemoteSource`, but there is no `CommandSource` for commands called with the execution of the `/execute` command.

There should be a special `CommandSource` for these senders with a method for getting the sender of the command (another `CommandSource`) as well as the entity the command is being executed for.

JBYoshi:
This PR adds three new `CommandSource` interfaces:
- `EntitySource`, for `CommandSource`s bound to an entity
- `ProxySource`, for /execute
- `SignSource`, for commands run via a sign

It also makes a backwards-incompatible change - making `CommandCallable` return `CommandResult` instead of `Optional<CommandResult>`. Returning `Optional.absent()` is replaced by throwing a `CommandNotFoundException`.

Finally, it adds `PaginationBuilder.page(int)`, which is used by my SpongeCommon PR: SpongePowered/SpongeCommon#80.


modwizcode:
I don't think we should be adding an exception and replacing Optional. As someone once said, exceptions are for exceptional conditions, people typing in the wrong command is not exceptional, just incorrect.

simon816:
I don't agree on having a specific source for signs. Maybe something like `TileEntitySource` that can cover signs/command blocks/modded blocks.

modwizcode:
@simon816 Signs are a unique source, after working with signs and their commands for awhile I find that it might be valid to have them uniquely separated. Especially considering that the permissions on signs might eventually be tied to the data of the `Subject` who placed them.

JBYoshi:
Should I rename SignSource to TileEntitySource?

JBYoshi:
The reason I chose to change CommandCallable  to not be Optional was actually tgat returning absent means “command not found”, and if control passes to the command... then the  command exists, so the command *is* found. In other words, commands should not return absent.

An alternative would be to make the command service not a CommandCallable, allowing it to return an optional, while all the other ones stay non-optional.

zml2008:
I really don't like the idea of EntitySource -- it makes for inconsistencies in the API, since Player is an entity while other CommandSources would hold references to entities. I think it would make much more sense to have entity classes directly implement CommandSource (ProxySource still makes sense though)

SignSource shouldn't refer to a Player -- especially since the docs are a bit ambiguous on which player it's supposed to be -- the one who placed the sign or the one who interacted with the sign. In the first case, the player might not be online or might not even be set. In the second case, the player could be a SpongeActor possibly. However, for the second case, the sign would be a proxy, and should be a ProxySource.

I'd also like to see the use case for the page() method in pagination builder -- in what cases would you want to start midway through a pagination?

I'm not sure about `Optional<CommandResult>`. I think part of the point of allowing a CommandCallable to handle subcommands is so a Dispatcher can be used to have a subcommands system. However, this still only makes sense for Dispatchers and the CommandService. The question for this is does removing the Optional<> from the CommandCallable interface result in a bigger usability improvement than that usability loss, and if it's worth the plugin break.

JBYoshi:
The main reason I created EntitySource was to abstract out the getEntity() method from SignSource and ProxySource. I made Player and MinecartCommandBlock implement it actually to create consistency. (I thought about making the entities not implement CommandSource at all to force plugin makers to use it, but I didn't want to break compatibility with that.)

I was wanting to make SignSource extend ProxySource. Doing that would remove the need for EntitySource. However, it doesn't work because of return types. I'm intending that the “cause” for a ProxySource is the one who chose and has permission to do the command. In the case of SignSource, it's the sign, but for ProxySource, it's the CommandSender who ran /execute. Maybe the sign itself could be made into a CommandSource? 

The page() method is actually for a custom implementation of /help that works with Sponge commands. (I haven't made the PR yet because I messed up a merge commit and will probably modify it.) To keep compatibility, I'm retaining the page argument. It can also be used for other plugins with paged commands that are porting to Sponge.

I was debating about the plugin break from removing the Optional. I found two disadvantages to having it as Optional:
- If you have both the root dispatcher and a command with subcommands return absent when a command is not found, then the user won't know the difference.
- I was writing a command to test my changes, and I assumed that returning absent meant a failure (I come from Bukkit). In that case, the docs for that should be improved.
- Returning absent is simply an error, and since Sponge uses exceptions for command errors, an exception should suit. This also makes the API more consistent, as an error comes from only one place.
The downside for that is, of course, the plugin break, plus the slight performance hit with exceptions. However, as modwizcode mentioned, the performance hit could be avoided by making CommandResult also include an optional error message instead of throwing an exception.

zml2008:
Ok EntitySource is not happening. It looks like you've made the various entities/tileentities implement CommandSource in the PR, just keep things like that.

For SignSource, it seems to be acting mostly as a proxy for the original clicking player (apart from permissions -- we might want to do that differently?)

as mentioned in the other PR, page() is not going to happen either

Removing the Optional<> is fine, the exception is insignificant. Just make sure the string send matches what the vanilla message is -- yours currently doesn't.

JBYoshi:
> Ok EntitySource is not happening. It looks like you've made the various entities/tileentities implement CommandSource in the PR, just keep things like that.

The only problem with that is SignSource - see below.

> For SignSource, it seems to be acting mostly as a proxy for the original clicking player (apart from permissions -- we might want to do that differently?)

So... does that basically mean that TileEntitySign needs to implement CommandSource?

> Removing the Optional<> is fine, the exception is insignificant. Just make sure the string send matches what the vanilla message is -- yours currently doesn't.

Only one question: The root dispatcher can send the translation commands.generic.notFound, but it might be confusing to send the same thing for subcommands. For example, with a command /foo < bar|baz >, saying /fooo bar and /foo barr would both send the same error with no indication of what went wrong. How should I do that?

zml2008:
well, TileEntitySign wouldn't have to implement CommandSource -- only the proxy object would. I think this could go either way -- the execute as command itself doesn't implement the CommandSource instance, entities do. The question here is whether we think of it as the sign being the source of the command, or a mystical proxy less directly attached to the world -- if we think of it as concretely the sign, then the Sign should be a CommandSource (indicating the proxyness), otherwise just have the anonymous class implement CommandSource.

wrt the error message, that was an advantage of the Optional system -- the direct parent always handled (well had the opportunity to handle) the 'not found' case -- so if you are doing child commands just throw a CommandException with the missing child command exception. You could catch the ChildCommandNotFoundException (or whatever you called it) in an intermediate wrapper and apply a more appropriate error message possibly?

JBYoshi:
I just thought of an idea. ProxySource would have three methods:
- getInitiator() returns the CommandSource that caused the command to be executed at that specific point in time. This would be the player who typed /execute or the player who clicked the sign.
- getProvider() returns the Permissible who holds the permissions for the command. This would be the player who typed /execute or the sign itself.
- getEntity() returns the Entity that the command is being run as. This would be the entity selected by /execute or the player clicking the sign.

Thoughts?

@zml2008 I like the idea for catching the CommandNotFoundException and throwing a better one. I'll get to work on that.

JBYoshi:
Anything about the Optional debate?

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -119,7 +122,7 @@ public SimpleDispatcher(Disambiguator disambiguatorFunc) {
      * @param callable The command
      * @param alias An array of aliases
      * @return The registered command mapping, unless no aliases could be registered
-     */
+    */
```
ST-DDT: The white-space removal should be reverted here.

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    CommandSource getInitiator();
+
+    /**
+     * Gets the {@link Subject} that holds the permissions necessary to perform
+     * the command.
+     *
+     * @return The permission holder
+     */
+    Subject getProvider();
```
ST-DDT: Why not call this `getSubject()` or `getPermissionHolder()` since there are a lot of things that could be provided other than permissions.

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
JBYoshi: Whoops

src/main/java/org/spongepowered/api/util/command/dispatcher/SimpleDispatcher.java
```diff
@@ -24,11 +24,12 @@
  */
 package org.spongepowered.api.util.command.dispatcher;
 
+import org.spongepowered.api.util.command.CommandNotFoundException;
```
ST-DDT: Wrong import order

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
```diff
@@ -0,0 +1,59 @@
...
+ * different entity.
+ */
+public interface ProxySource extends LocatedSource {
+
+    /**
+     * Gets the entity operated on by this proxy.
+     *
+     * @return The entity
+     */
+    Entity getEntity();
```
zml2008: Not sure I like this -- I think it really only makes sense to have this refer to the source being proxied -- so the player being proxied through the sign, whatever source being proxied through `/execute`, and so on.

JBYoshi: @zml2008 I'm good on removing `getSubject()`, as the `Subject` methods of `ProxySource` implementations delegate to them. I'm also OK with removing `getInitiator()`, though plugins may still want to use it.

JBYoshi:
@zml2008 How's it look?

Issue #601, opened 2015-04-27
Permissions registry-type object for plugins?


Xemiru:
A way to get all the permissions of a plugin would be nice.

* A `PermissionData` class to store all information about a permission node's purpose. Should include information about the permission, the name (e.g. "/call"), the description of what it allows (not just "Allows access to the /call command."), a default setting, etc.

* A `Permissible` interface that can give the list of registered permissions of the plugin?
* A `PermissionRegistry` class holding `PermissionData` for plugins?
* Letting the `PluginContainer` hold all `PermissionData` for its plugin?
* boformer: Allow it as a parameter in the @Plugin annotation?

Use Cases:
* https://github.com/Xemiru/Permissions-Writer-Utility/issues/3
* Permissions plugins' tab completion

Pinging @zml2008 as he mentioned being able to do something like this.

boformer:
It is part of the metadata of a plugin, so it should be a part of the @Plugin annotation.

I would prefer a simple builder for the permission data.

This has a really low priority for me.

Xemiru:
Very low priority, but a feature nonetheless.

MattBDev:
I really hope this can be more of a priority because for a plugins that involve protected blocks or land, they need to be able to register a permissions from outside of the command service. I see this as being critical for plugins like LWC, GriefPrevention, and PlotMe which all have the option to allow admins to override protections.

zml2008:
uh permissions aren't registered at all currently why are you still talking about registering permissions?

MattBDev:
It is something that is needed so developers of plugins similar to the ones I mentioned above can start working on their own plugins.

zml2008:
Why do permissions need to be registered? There's nothing in sponge that requires registration of permissions?

MattBDev:
How else are they supposed to be created? Since there is no plugin metadata file, there is no place to say what permissions a plugin has.

zml2008:
permissions are a string. you pass the string to the hasPermission method. you don't need a plugin.yml to get strings?

MattBDev:
So essentially any permission plugin will consider every string a permission regardless if there isn't a plugin on the server that has that permission?

zml2008:
Well if a plugin checks for a permission then ofc it is a permission. simple as that

boformer:
Yep. You can do hasPermission(randomString) and it would return true if the player has it. The same worked in Bukkit as well ;)

ST-DDT:
[...]

PR #747, opened 2015-07-03
Add PermissionDescription API

ST-DDT:
Adds https://github.com/SpongePowered/SpongeAPI/issues/601 (low priority)
Implemented in https://github.com/SpongePowered/SpongeCommon/pull/84

This adds `PermissionDescription` to the `PermissionService`.

`PermissionDescription`s should help server owners manage the registered permissions more easyly.

Some facts about `PermissionDescription`s:
* Registration is optional
* Registration may not be supported by all impls
* Registration has **NO** impact on the default value in general
  - Permission Plugins may offer util commands/options to import the defaults (per plugin)
  - The `role-templates` can be assigned to groups (manually) to allow changing the defaults
* Contains of
  - Owning plugin
  - PermissionId (`String`)
    - May include placeholders `myPlugin.command.give.<ItemType>`
  - Description `Text`
    - May include a link to a more detailed description page
  - `Role-emplate` assignment
     - `User`, `Staff`, `Admin`, custom values
     - This can be used to group permissions and allow server admins to add them to groups more easily.

The discussion and a short explanation why i have chosen this properties can be found [here](https://github.com/SpongePowered/SpongeAPI/issues/601#issuecomment-116326563)

**This PR does NOT contain**
* Annotations to automatically create and register permissions
  - Just too messy if it should contain useful info/descriptions
* File specifications for permission data files
  - I thought about that, but i think this should be done in a more general way.
    - This should be done similar to json/config->object parsing, which maps to the keys to method names. If you like this i can write some example code. (NOT part of this PR)
* Commands to get the registered descriptions
  - This is up to the permission plugins, since the storing of the descriptions has great influence on the possibilities, however a basic (`/sponge` sub-) command could be added nevertheless.

**TBD**
* Explicit `PermissionReference`s (maybe as `String`)
  - `CHILD`/ `PARENT` - Basic inheritance
  - `REFERENCE` - "See also"
  - `BASE`/`SPECIFICATION` - Base permission `myPlugin.command.give.<ItemType>` of `myPlugin.command.give.DIRT` or `myPlugin.command.give.*`
  - `EXAMPLE` - Opposite of `BASE`/`SPECIFICATION`

Anything i missed?

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
+++ b/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
@@ -114,7 +114,8 @@ public interface PermissionDescription {
     /**
      * Gets an immutable {@link Set} of {@link Subject}s that have set a none
      * default value for this permission (true and false). This may include
-     * permissions that are directly related to this permission.
+     * permissions that are directly related to this permission. This may not
+     * include subjects that inherit the none default value.
      *
      * <p>If you want to know to which role-templates this permission is
      * assigned use {@link PermissionService#SUBJECTS_ROLE_TEMPLATE}.</p>
--- a/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
+++ b/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
@@ -73,14 +73,38 @@ public interface PermissionDescription {
     static final String RANK_ADMIN = "admin";
     /**
      * A permission with this rank should normally not granted to anybody. This
-     * usually applies to debug permissions.
+     * usually applies to debug permissions or specializations of permissions
+     * that would already be granted.
+     *
+     * <p>Examples can be show debug messages or myPlugin.give.type.DIRT where
+     * myPlugin.give.type is already granted.</p>
      */
     static final String RANK_NOBODY = "nobody";
 
     /**
-     * Gets the permission id this description belongs too. May include
-     * placeholders {@code "<SomeParameter>"} such as in
-     * {@code "pluginId.give.item.<ItemType>"}.
+     * Gets the permission id this description belongs too.
+     *
+     * <p>The permission id must be of the specified format as specified using
+     * EBNF:
+     * <ul>
+     * <li>CHARACTER  = "A" - "Z" | "a" - "z" | "0" - "9" | "_" | "-"</li>
+     * <li>NAME       = CHARACTER , { CHARACTER }</li>
+     * <li>TEMPLATE   = "&lt" , NAME , "&gt"</li>
+     * <li>PART       = NAME | TEMPLATE</li>
+     * <li>PERMISSION = NAME , { "." , PART }</li>
+     * </ul>
+     * </p>
+     *
+     * <p>Examples:
+     * <ul>
+     * <li>"myPlugin.give.type" - Grants all ItemTypes</li>
+     * <li>"myPlugin.give.type.&ltItemType&gt" - A template should not be granted to anybody</li>
+     * <li>"myPlugin.give.type.DIAMOND" - Only grants DIAMOND</li>
+     * </ul>
+     * </p>
+     *
+     * <p><b>Note:</b> Permission ids are case insensitive! Permission ids should
+     * start with the owning plugin's id.</p>
      *
      * @return The permission id
      */
@@ -118,9 +142,8 @@ public interface PermissionDescription {
     interface Builder {
 
         /**
-         * Sets the permission id for the description this builder creates. May
-         * include placeholders {@code "<SomeParameter>"} such as in
-         * {@code "pluginId.give.item.<ItemType>"}.
+         * Sets the permission id for the description this builder creates. See
+         * {@link PermissionDescription#getId()} for format specifications.
          *
          * @param permissionId The permission id
          * @return This builder for chaining
--- a/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
+++ b/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
@@ -1,0 +1,170 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.service.permission;
+
+import com.google.common.base.Optional;
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.text.Text;
+
+import javax.annotation.Nullable;
+
+/**
+ * A description object for permissions. The description is meant to provide
+ * human readable descriptions and meta data for a permission. Instances should
+ * be registered at the {@link PermissionService}, but the registration does not
+ * have any impact on the results of the permission checks of the service.
+ */
+public interface PermissionDescription {
+
+    /**
+     * A permission with this rank can be granted to anybody including the
+     * untrusted guests. For most servers there is no difference between
+     * {@link #RANK_GUEST} and {@link #RANK_USER}.
+     *
+     * <p>Examples can be chat, MotD receive.</p>
+     */
+    static final String RANK_GUEST = "guest";
+    /**
+     * A permission with this rank can be granted to anybody that should have
+     * default user access. For most servers there is no difference between
+     * {@link #RANK_GUEST} and {@link #RANK_USER}.
+     *
+     * <p>Examples can be claim city plot, access to games and public warps.</p>
+     */
+    static final String RANK_USER = "user";
+    /**
+     * A permission with this rank should only be granted to known server staff.
+     * The permission's impact on the server is quite low.
+     *
+     * <p>Examples can be give items, teleport other players, access to support
+     * commands, globally mute other players.</p>
+     */
+    static final String RANK_STAFF = "staff";
+    /**
+     * A permission with this rank should only be granted to fully trusted
+     * server administrators. The permission's impact on the server may be high.
+     *
+     * <p>Examples can be create/delete worlds, grant permissions, restart
+     * server.</p>
+     */
+    static final String RANK_ADMIN = "admin";
+    /**
+     * A permission with this rank should normally not granted to anybody. This
+     * usually applies to debug permissions.
+     */
+    static final String RANK_NOBODY = "nobody";
+
+    /**
+     * Gets the permission id this description belongs too. May include
+     * placeholders {@code "<SomeParameter>"} such as in
+     * {@code "pluginId.give.item.<ItemType>"}.
+     *
+     * @return The permission id
+     */
+    String getId();
+
+    /**
+     * Gets a short description of the linked permission. May include a link to
+     * a more detailed description on the plugin's web page.
+     *
+     * @return A short description of the linked permission
+     */
+    Text getDescription();
+
+    /**
+     * Gets the human readable <b>suggested</b> rank the players should have
+     * before they should have access to this permission.
+     *
+     * <p><b>Note:</b> This is a suggestion for server owners and does not have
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -0,0 +1,170 @@
...
+
+        /**
+         * Sets the suggested rank players should have before being able to
+         * access this permission. Setting this is optional although
+         * recommended.
+         *
+         * @param rank The suggested rank to have
+         * @return This builder for chaining
+         */
+        Builder suggestedRank(@Nullable String rank);
```
zml2008: So the way I brought up of doing this was setting permissions on `rank-template` type subjects (or maybe `role`, just a name), which could then be added as parents to the server-specific groups. Adding this method just makes another thing for permissions plugins to handle when thinking about what permissions to assign where.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -0,0 +1,170 @@
...
+         *         overwriting descriptions
+         */
+        PermissionDescription register() throws IllegalStateException;
+
+        /**
+         * Resets this builder to its default settings.
+         *
+         * @return This builder for chaining
+         */
+        Builder reset();
```
zml2008: I don't think this is necessary. Builders are used only once, so why would a reset be needed?

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * A builder for permission descriptions.
+     */
+    interface Builder {
+
+        /**
+         * Sets the permission id for the description this builder creates. May
+         * include placeholders {@code "<SomeParameter>"} such as in
+         * {@code "pluginId.give.item.<ItemType>"}.
```
zml2008: Not sure about this. If it is going to be implemented like, there should be a stricter/more formal grammar. Take a look at EBNF for inspiration.

Also would be good to spend some thought on what uses these parameters will have in-plugin. I think plugins would have to end up stripping the parameters from permissions when presenting them to a user (autocomplete for example). Might be good to provide them as help or as some sort of fill-in-the-blanks though.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
ST-DDT: If you want to assign them to `role`/`rank-template` you can do something similar already.

The suggested rank is basically your `rank-template` but IMO wrapping this in another object does not make sense since you try to register them from the plugin side not the permission service side. The problem with wrapping it in a `Subject` is that you somehow have to leak references to them. Either via Pseudo-Enum or via a getter. This can be achieved internally quite easy i guess. But this new `Subject` type is a new code you have to implement whereas the current suggestion is just for show or optional. Also you would not want to assign a permission like `myPlugin.give.<ItemType>` to a subject although a hint to whom this permission should be granted can be quite useful.



But for the most important part. This value is used for server owners to help them choose whether they would like to grant a permission to a player or not. The defaults are just for convenience to help plugin creators and permission services to assign the permissions to specific groups and reduce the total number of ranks/group names.

We sure can change this to

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
ST-DDT: The reset method is available in quite a lot of builders so i thought i add it here as well.

I thought of something like this for importing from permission definition files:

````java
Optional<Builder> optBuilder = permissionService.newDescriptionBuilder(plugin);
if (!optBuilder.isPresent) {
    return;
}
Builder builder = optBuilder.get();
Config config = YAMLConfig.loadFromResource(plugin, "permissions.yml");
for (String key : config.getKeys()) {
    builder.reset();
    builder.id(key);
    builder.description(config.getText(key + ".description"));
    if (config.hasEntry(key + ".suggestedRank")) {
        builder.suggestedRank(config.getString(key + ".suggestedRank"));
    }
    builder.register();
}
````

File: src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java (hunk 3 above)
Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java]: My current intention was that the server owner can read it as well as some methods to provide some basic syntax plugins can rely on. 

Let my explain this a bit.
The entire information included in a list of permissions like
* `pluginId.give.item.DIRT`
* `pluginId.give.item.GRAVEL`
* `pluginId.give.item.STONE`
* `pluginId.give.item.GRASS`

is also included in the single permission template `pluginId.give.item.<ItemType>`.
This also better for typos since if you type `pluginId.give.item.GASS` it can still match it to the template. Although the ItemType is not available on that server right now.
I thought of something like this:
````
PermissionDescriptionNode node = ROOTNODE;
String[] parts = permissionId.split(".", LITTERAL);
boolean matches = true;
for (String part : parts) {
    if (node.hasChild(part)) {
        node = node.getChild(part);
    } else if (node.hasTemplateChildren()) {
        node = node.getTemplateChildren();
    } else {
        matches = false;
        break;
    }
}
if (matches) {
    if (node.isMultiNode()) {
         showAmbigousMessage(player, node.getChoices(current));
    } else {
        showMessage(player, node.getDescription(), node.getChoices(parent), node.getChoices(children));
    }
} else {
    showNotFound(player, node.getChoices(parent + current + children));
}
````

`/perm description <permissionId>`

NodeTree:
- `myPlugin`
  - `<Command>`
  - `give`
    - `<ItemType>`
      - `fullStack`
    - `<Enchantment>`

Example run on `myPlugin.give.<ItemType>`:
* `node = ROOT`
* `node = node.getChild("myPlugin") // "myPlugin"`
* `node = node.getChild("give") // "myPlugin.give"`
* `node = node.getChild("<ItemType>") // "myPlugin.give.<ItemType>"`
* `showMessage(player, [current node], [parent node], [child nodes])`
* Outputs:
  - `Allows the usage of the specified <ItemType> in the /give command.`
  - `For a full list of possible ItemTypes see LINK`
  - `Suggested for: Staff`
  - `Provided by: MyPlugin`
  - `Parent permission: /perm description myPlugin.give`
  - `Available sub permissions: /perm description myPlugin.give.<ItemType>.fullStack`

Example run on `myPlugin.give.GASS`:
* `node = ROOT`
* `node = node.getChild("myPlugin") // "myPlugin"`
* `node = node.getChild("give") // "myPlugin.give"`
* `node ("myPlugin.give.GASS") does not exist`
* `node = node.getTemplateChildren() // "myPlugin.give" , ["<ItemType>", "<Enchantment>"]`
* `showAmbigousMessage(player, [current nodes])`
* Outputs:
  - `The specified permission is ambigous. Please choose:`
  - `/perm description myPlugin.give.<ItemType>`
  - `/perm description myPlugin.give.<Enchantment>`

Example run on `myPlugin.give.GASS.*`:
* `node = ROOT`
* `node = node.getChild("myPlugin") // "myPlugin"`
* `node = node.getChild("give") // "myPlugin.give"`
* `node "myPlugin.give.GASS" does not exist`
* `node = node.getTemplateChildren() // "myPlugin.give" , ["<ItemType>", "<Enchantment>"]`
* `nodes ("myPlugin.give" , ["<ItemType>", "<Enchantment>"] , "*") do not exist`
* `matches = false`
* `showNotFound(player, [parent, current & sub nodes])`
* Outputs:
  - `The specified permission was not found. Please choose:`
  - `/perm description myPlugin.give` // Parent
  - `/perm description myPlugin.give.<ItemType>` // Current (multi) node
  - `/perm description myPlugin.give.<Enchantment>`
  - `/perm description myPlugin.give.<ItemType>.fullStack` // Sub nodes of this (multi) node

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java]: So here is my grammar specification:

-----------------------------------------

The permission id must be of the specified format as specified using EBNF:
* CHARACTER  = 'A' - 'Z' | 'a' - 'z' | '0' - '9' | '_' | '-'
* NAME       = CHARACTER , { CHARACTER }
* TEMPLATE   = "<" , NAME , ">"
* PART       = NAME | TEMPLATE | '*'
* PERMISSION = PART , { "." , PART }

Examples:
* `myPlugin.give.*`
* `myPlugin.give.<ItemType>`
* `myPlugin.give.DIAMOND`

**Note:** Permissions ids are case insensitive!
Permission ids should start with the owning plugin's id.

-----------------------------------------

As Javadoc
````java
/**
 * Gets the permission id this description belongs too.
 * 
 * <p>The permission id must be of the specified format as specified using
 * EBNF:
 * <ul>
 * <li>CHARACTER  = "A" - "Z" | "a" - "z" | "0" - "9" | "_" | "-"</li>
 * <li>NAME       = CHARACTER , { CHARACTER }</li>
 * <li>TEMPLATE   = "&lt" , NAME , "&gt"</li>
 * <li>PART       = NAME | TEMPLATE | "*"</li>
 * <li>PERMISSION = PART , { "." , PART }</li>
 * </ul>
 * </p>
 * 
 * <p>Examples:
 * <ul>
 * <li>"myPlugin.give.*</li>
 * <li>"myPlugin.give.&ltItemType&gt"</li>
 * <li>"myPlugin.give.DIAMOND"</li>
 * </ul>
 * </p>
 * 
 * <p><b>Note:</b> Permission ids are case insensitive! Permission ids should
 * start with the owning plugin's id.</p>
 *
 * @return The permission id
 */
String getId();
````
Anything you would like to change?

File: src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java (hunk 1 above)
Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java]: Pro Subject:
* Allows Context aware permission defaults for each role (ex: "User")
 * How to skip some contexts from the suggestion?
* Easier to implement for permission service providers

Pro String
* More fine grained server owner controll whether he wants to add the role "User in BedWarsArena" to an existing group.
 * How to use the context here?
* Easier to implement plugin wise

File: src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java (hunk 3 above)
Comment (zml2008) [inline src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java]: the .* is not a thing in sponge permissions, a.b automatically gives a.b.c and a.b.d, so you'll want to get rid of that. I think the first, more ebnfy spec is clearer overall. In general this is really good though

File: src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java (hunk 1 above)
Comment (zml2008) [inline src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java]: >  But this new Subject type is a new code you have to implement 

no extra implementation is required. Code example:
```java
permService.getSubjects("rank-templates").get("moderator").getTransientData().setPermission(SubjectData.GLOBAL_CONTEXTS, "some-perm", Tristate.TRUE);
```

then a user can either view that rank template at any time (`/pex rank-template moderator info`), or add it as a parent of an actual group (`/pex group mod parent add rank-template moderator`)
> Reminder: Wasn't there a the consent that the registration does not have any impact on the defaults?

subjects don't have an impact on the defaults, they have to be explicitly applied.

> Less/Fixed roles

not true

> More fine grained server owner controll whether he wants to add the role "User@BedWarsArena" to an existing group.

again, not true. a server admin has the choice of whether they add a specific rank template or simply look at it for values.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
zml2008: just run permissionService.newDescriptionBuilder(plugin); each time?

ST-DDT: That would be possible as well, but with the Optional this is quite annoying.

`Builder`s without `reset()`
* org.spongepowered.api.effect.particle.ParticleEffectBuilder
* org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder
* org.spongepowered.api.item.recipe.ShapedRecipeBuilder
* org.spongepowered.api.item.recipe.ShapelessRecipeBuilder
* org.spongepowered.api.util.command.CommandResult.Builder
* org.spongepowered.api.util.command.args.CommandFlags.Builder
* org.spongepowered.api.util.command.spec.CommandSpec.Builder
* ...

`Builder`s with `reset()`
* org.spongepowered.api.attribute.AttributeBuilder
* org.spongepowered.api.attribute.AttributeModifierBuilder
* org.spongepowered.api.data.ImmutableDataBuilder<H, E>
* org.spongepowered.api.item.FireworkEffectBuilder
* org.spongepowered.api.item.inventory.ItemStackBuilder
* org.spongepowered.api.item.merchant.TradeOfferBuilder
* org.spongepowered.api.potion.PotionEffectBuilder
* org.spongepowered.api.world.gen.populator.*.Builder
* ...

There should be a common decision whether our builders should have a `reset()` method or not.

Pro `reset()`
* Do not have to leak an entire service or the `GameRegistry` to a method where only a `Builder` is needed.
* Lesser object creation and thus less garbage collection.

Cons:
* ?

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
ST-DDT: So your suggestion is that we remove the suggestedRank from the description.
And add a new `Subject` type (constant) to the permission service. 
(As part of this PR?)

If the suggestedRank is dropped this is basically a `id` -> `description` mapping (since the owner does not count really) maybe the entire interface can be removed and moved to a simple `offerDescription(String,Text)` method inside the `PermissionService`. This would also remove the need of the builder.
Or are there some other features you would like to add to the descriptions?
Anyone else who wants to share his opinion?

Shall there be a shortcut in the builder/description to get a `rank-template` `Subject`?
Maybe these both concerns should be kept separated.

zml2008: Having this utility method to add a permission either as a default or as a rank template probably makes sense... if only we had interface default methods...

ST-DDT: > if only we had interface default methods...

True words.

ST-DDT:
@zml2008 I reworked the role suggestion to our role template example.
Did you have something like this in mind?
[`Description#getAssignedSubjects(String identifier) : Set<Subject>`](https://github.com/SpongePowered/SpongeAPI/pull/747/files#diff-30abf20e9e6a3cc1d34dd5bf68a981a1R115)
[`Builder#assign(String role, boolean value) : Builder`](https://github.com/SpongePowered/SpongeAPI/pull/747/files#diff-30abf20e9e6a3cc1d34dd5bf68a981a1R160)

Issue #750, opened 2015-07-05
World Copy From Existing Source


FerusGrim:
Currently, copying worlds requires manually modifying NBT files to change the World UUID/Name (speculation), otherwise you'll run into quite a bit of `ReportedException` and `ConcurrentModificationError`s (eg; it appears the server can't tell the difference, and attempts to make changes to both maps as if they were a single instance).

(NOTE, the below log is _definitely_ truncated)

```java
[12:40:42] [Server thread/ERROR]: Encountered an unexpected exception
net.minecraft.util.ReportedException: Exception ticking world
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:770) ~[MinecraftServer.class:?]
	at net.minecraft.server.dedicated.DedicatedServer.updateTimeLightAndEntities(DedicatedServer.java:385) ~[DedicatedServer.class:?]
	at net.minecraft.server.MinecraftServer.tick(MinecraftServer.java:669) ~[MinecraftServer.class:?]
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:540) [MinecraftServer.class:?]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_45]
Caused by: java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901) ~[?:1.8.0_45]
	at java.util.ArrayList$Itr.next(ArrayList.java:851) ~[?:1.8.0_45]
	at net.minecraft.world.gen.ChunkProviderServer.unloadAllChunks(ChunkProviderServer.java:98) ~[ChunkProviderServer.class:?]
	at net.minecraft.server.management.PlayerManager.updatePlayerInstances(PlayerManager.java:94) ~[PlayerManager.class:?]
	at net.minecraft.world.WorldServer.tick(WorldServer.java:230) ~[WorldServer.class:?]
	at net.minecraft.server.MinecraftServer.updateTimeLightAndEntities(MinecraftServer.java:764) ~[MinecraftServer.class:?]
	... 4 more
[12:40:42] [Server thread/ERROR]: This crash report has been saved to: E:\Programming\Projects\Sponge\run\server\.\crash-reports\crash-2015-07-05_12.40.42-server.txt
[12:40:42] [pool-2-thread-1/INFO]: World was prepared and loaded successfully. If you're a player, we'll teleport you, now.
[...]

PR #751, opened 2015-07-06
[WIP] Add API's for additional world management operations

modwizcode:
This pull request introduces an API where worlds that exist can be copied and loaded as a new world with a different name. 

This pretty much needs an API because of the uuid added for Sponge's worlds. Fixing this would require plugins to copy the world, then load up the NBT and change the name, dimension id and UUID in the file manually, which is prone to breaking and also likely impossible, because plugins shouldn't have access to a way to get the next open dimension ID.

Solves https://github.com/SpongePowered/SpongeAPI/issues/750

Feature list:
- [X] Support the ability to copy an existing world and load it as a duplicate with the same name.
- [X] ~~Support the ability to get the world directory from the API to allow plugins to store data with the worlds.~~
- [X]  Add ability to delete existing worlds from the disk (after being unloaded first)
- [ ] Add the ability to fully rename an existing world, including folder or other stored name.

Todo:
- [X] Determine if world copying should only be allowed after unloading a world **(It should)**

### DIFF
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -207,12 +207,27 @@ public interface Server extends ChannelRegistrar {
 
     /**
      * Creates a copy of the provided world under the new name given and returns
-     * the new copy if the copy was possible
+     * the new copy if the copy was possible. 
      *
      * @param originalWorld The original world instance to copy
      * @param copyName The name that should be used in the duplicated entry instead of the originalWorld's name.
      */
-    Optional<World> copyWorld(World originalWorld, String copyName);
+    Optional<WorldProperties> copyWorld(WorldProperties originalWorld, String copyName);
+
+    /**
+     * Allows the copy to be made while the world is still loaded by forcing a save
+     * first, this could cause changes to be missed if made after that save. By
+     * using this method the developer is assumed to properly respect that large worlds
+     * will likely take a long enough period to copy that a reasonable number of changes are
+     * missed.
+     *
+     * A {@link World} instance is used here because unloaded worlds should not be copied
+     * using this method, and instead should be copied using {@link #copyWorld(WorldProperties,String)}.
+     *
+     * @param originalWorld The original world instance to copy
+     * @param copyName The name for the copy that is made
+     */
+    Optional<WorldProperties> copyLoadedWorld(World originalWorld, String copyName);
 
     /**
      * Persists the given {@link WorldProperties} to the world storage for it,
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -232,7 +232,17 @@ public interface Server extends ChannelRegistrar {
     Optional<WorldProperties> copyLoadedWorld(World originalWorld, String copyName);
 
     /**
+     * Renames an unloaded world.
+     *
+     * @param worldProperties The world instance to rename
+     * @param newName The name that should be used as a replacement for the current world name
+     * @return An {@link Optional} containing the new {@link WorldProperties} if the rename was successful
+     */
+    Optional<WorldProperties> renameWorld(WorldProperties worldProperties, String newName);
+
+    /**
      * Deletes the provided world's files from the disk.
+     *
      * @param worldProperties The world instance to delete
      * @return True if the deletion was successful.
      */
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -206,6 +206,15 @@ public interface Server extends ChannelRegistrar {
     Optional<WorldProperties> createWorld(WorldCreationSettings settings);
 
     /**
+     * Creates a copy of the provided world under the new name given and returns
+     * the new copy if the copy was possible
+     *
+     * @param originalWorld The original world instance to copy
+     * @param copyName The name that should be used in the duplicated entry instead of the originalWorld's name.
+     */
+    Optional<World> copyWorld(World originalWorld, String copyName);
+
+    /**
      * Persists the given {@link WorldProperties} to the world storage for it,
      * updating any modified values.
      *
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -210,7 +210,8 @@ public interface Server extends ChannelRegistrar {
      * the new copy if the copy was possible. 
      *
      * @param originalWorld The original world instance to copy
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,15 @@
     Optional<WorldProperties> createWorld(WorldCreationSettings settings);
 
     /**
+     * Creates a copy of the provided world under the new name given and returns
+     * the new copy if the copy was possible
+     *
+     * @param originalWorld The original world instance to copy
+     * @param copyName The name that should be used in the duplicated entry instead of the originalWorld's name.
+     */
+    Optional<World> copyWorld(World originalWorld, String copyName);
```
Deamon5550: Might be good to change this to `WorldProperties` rather than `World` so that you don't have to load a world to copy it.

JBYoshi:
> Todo: Determine if world copying should only be allowed after unloading a world

I was doing something like this to account for a bug in Multiverse's clone code (that was before I had a GitHub account, so I just copied and changed it). It's easier to implement if the world does need to be unloaded, as you can simply copy the directory, but it's annoying to users who are in the world.

modwizcode:
I found that the copy works even if the world isn't unloaded but I don't
know how safe that is

ryantheleach:
Do you at least lock saving first?

modwizcode:
No. That will definitely be changing, the implementation is in the linked
SpongeCommon PR if you wish to look. That code worked for me, but I didn't
have modification happening during the copy in my testing so I'm pretty
sure that will break.

The only proper way I know to lock saving would be to have the world
unloaded which makes more sense to me anyway.

JBYoshi:
Will there be a way to clone the world without loading it immediately afterwards? (To have a chance to modify the WorldProperties before the new world is loaded, e.g. making the spawn area not be loaded when the world is.)

modwizcode:
Yes the copy will be an optional `WorldProperties`, the optional `World`
was before I realized that it could be split.

blablubbabc:
I have no actual clue, but how fast is the copying of (large) worlds usually? Will this be a problem if done on the server's main thread? If yes, would it make sense to run the file io in a separate thread and let the caller provider some sort of callback, which gets called once the world has been finished copied and loaded?

modwizcode:
An asynchronous copy might be a good idea. @SpongePowered/Developers
comments?

simon816:
Yes I think an asynchronous call would be a good idea. It would return a `ListenableFuture`.
I think you'd need to be careful in the implementation that you can't attempt to load the world before the thread has finished execution. Possibly introduce a lock to ensure nothing can touch the world while it's copying.

JBYoshi:
I do like an asynchronous load and copy. I did something like this for a Bukkit minigame plugin. When the world loaded (`WorldInitEvent`), I would set `keepSpawnLoaded` to `false`, use a repeating task to load each chunk, and set `keepSpawnLoaded` to `true`. `ListenableFuture` seems good, though it would be nice to have a way to listen for load progress (e.g. 42%).

ryantheleach:
Should we allow access to the execution context, or provide some ListenableFutures that already have the execution context defined? I could see one that executes the callback at the next possible synchronous moment using the scheduler if created on the main thread, or an asynchronous context being created if created by an asynchronous task. 

It's probably overcomplicating things for people who know what they are doing, but doing magic for beginners.

JBYoshi:
One problem with asynchronous chunk loading: the `ChunkLoadEvent` would have to be fired - and probably waited for - on the main thread.

JBYoshi:
Also, maybe events for world copying, deleting, and renaming ought to be good (so plugins can copy, delete, and rename their data).

modwizcode:
Same as API, the code for the import and change of worlds that are cold (haven't been loaded at any point, fresh out of a compressed archive), will be moved into a separate, easier to complete PR.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,49 @@
     Optional<WorldProperties> createWorld(WorldCreationSettings settings);
 
     /**
+     * Creates a copy of the provided world under the new name given and returns
+     * the new copy if the copy was possible. 
+     *
+     * @param originalWorld The original world instance to copy
+     * @param copyName The name that should be used in the duplicated entry instead of the originalWorld's name
```
gabizou: 80 line character limit for javadocs please.

intronate67:
OMG SOLID GOLD
Thats what this is

intronate67:
Any updates with this?

modwizcode:
Will be trying to wrap up by the end of this week. If not done by then I'll
need to push it to someone else as I'll be too busy with school and work
for awhile.

modwizcode:
I will be unable to handle proper implementation of this, I should be able to finish API tomorrow.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,49 @@
...
+     * missed.
+     *
+     * A {@link World} instance is used here because unloaded worlds should not be copied
+     * using this method, and instead should be copied using {@link #copyWorld(WorldProperties,String)}.
+     *
+     * @param originalWorld The original world instance to copy
+     * @param copyName The name for the copy that is made
+     * @return An {@link Optional} containing the properties of the new world instance, if the copy was successful
+     */
+    Optional<WorldProperties> copyLoadedWorld(World originalWorld, String copyName);
```
bloodmc: Why have a second method for this? Loaded worlds can be handled silently in implementation. Just state in copyWorld if the world being copied is already loaded the following will happen

1. A force save will occur
2. world saving, for the world being copied, is temporarily disabled
3. Copy starts
4. After copy is complete, world saving is re-enabled.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,49 @@
...
+
+    /**
+     * Renames an unloaded world.
+     *
+     * @param worldProperties The world instance to rename
```
bloodmc: WorldProperties is not the world instance. It represents storage (level.dat, level_sponge.dat)

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,49 @@
...
+     * @param newName The name that should be used as a replacement for the current world name
+     * @return An {@link Optional} containing the new {@link WorldProperties} if the rename was successful
+     */
+    Optional<WorldProperties> renameWorld(WorldProperties worldProperties, String newName);
+
+    /**
+     * Deletes the provided world's files from the disk.
+     *
+     * @param worldProperties The world instance to delete
```
bloodmc: Same as above.

PR #755, opened 2015-07-10
Additional API for World Management

FerusGrim:
Add methods to delete an unloaded World, as well as a method to retrieve the folder for a World.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
+++ b/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
@@ -39,6 +39,7 @@ import org.spongepowered.api.world.WorldBorder;
 import org.spongepowered.api.world.difficulty.Difficulty;
 import org.spongepowered.api.world.gen.WorldGeneratorModifier;
 
+import java.io.File;
 import java.util.Collection;
 import java.util.Map;
 import java.util.UUID;
@@ -103,6 +104,13 @@ public interface WorldProperties extends DataSerializable {
      * @return The name
      */
     String getWorldName();
+
+    /**
+     * Gets the directory the files for this world are located.
+     *
+     * @return The directory the files are located in
+     */
+    File getWorldDirectory();
 
     /**
      * Gets the {@link UUID} of the world.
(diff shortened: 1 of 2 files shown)


modwizcode:
Seems about right, thanks.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -105,6 +106,13 @@
     String getWorldName();
 
     /**
+     * Gets the directory the files for this world are located.
+     *
+     * @return The directory the files are located in
+     */
+    File getWorldDirectory();
+
+    /**
```
jamierocks: what if someone comes up with an elaborate way to store worlds in a database or something that would be very inefficient, but not file-storage?

modwizcode: That would break all our current world code too and would have to be addressed then.

jamierocks: ok.

simon816: @modwizcode yeah but that's only in Sponge's implementation. So far the API has not defined how worlds should be stored so some implementation 'MinecraftAsSQL' may not want to store it in a file system. Just a thought.

FerusGrim: @simon816 Getting the literal location of the world files is important. Whether that be a directory, or some other convoluted madness.

Who's taking on the task of creating this service?

Just an additional two-cents, however - If we're going to go this route of allowing the potential for world-storage that isn't flat-file, that means whatever service defines the world location would also need a standardized way for storing additional data provided by plugins.

jamierocks: The point wasn't that someone is working on this, merely that someone could.

Or couldn't now SpongeAPI assumes it has a directory.

ST-DDT: If we asume the world has a folder. Can we add methods for regions files as well?
Because both ways have been rejected last time this was discussed. For the above reasons.

> Just an additional two-cents, however - If we're going to go this route of allowing the potential for world-storage that isn't flat-file, that means whatever service defines the world location would also need a standardized way for storing additional data provided by plugins.

The plugin could store it in its own folder or database. It is not required to save it in the worlds folder. I really dont like plugins that randomly polute the world folder.

PR #756, opened 2015-07-10
Add API for renaming Worlds

FerusGrim:
(also remove unused java.io,File import)

### DIFF
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -233,6 +233,7 @@ public interface Server extends ChannelRegistrar {
 
     /**
      * Deletes the provided world's files from the disk.
+     *
      * @param worldProperties The world instance to delete
      * @return True if the deletion was successful.
      */
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -207,12 +207,27 @@ public interface Server extends ChannelRegistrar {
 
     /**
      * Creates a copy of the provided world under the new name given and returns
-     * the new copy if the copy was possible
+     * the new copy if the copy was possible. 
      *
      * @param originalWorld The original world instance to copy
      * @param copyName The name that should be used in the duplicated entry instead of the originalWorld's name.
      */
-    Optional<World> copyWorld(World originalWorld, String copyName);
+    Optional<WorldProperties> copyWorld(WorldProperties originalWorld, String copyName);
+
+    /**
+     * Allows the copy to be made while the world is still loaded by forcing a save
+     * first, this could cause changes to be missed if made after that save. By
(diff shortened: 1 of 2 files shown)


modwizcode:
Can you amend this to fix the deleteWorld javadoc?

modwizcode:
Please update to follow _blood's comments from IRC

ST-DDT:
Can we somehow change the PermissionContext of a world. So that if we have renamed the world the permission service still detects this world as the original one?

FerusGrim:
@ST-DDT Correct me if I'm wrong, but from what I can tell this would be done by the implementation.

ST-DDT:
Currently the world context is directly connected to the worlds name. But i would like to set it myself using API methods. It does not really have anything in common with renaming a world besides that both are associated with worlds and changing the identifier.

Just for the reference
https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/permission/context/Contextual.java

ryantheleach:
Really the name should just be a friendly name for users to use. If worlds have a UUID shouldn't the permissions service associate with the worlds UUID, and only use the name for display purposes?

modwizcode:
Rename will preserve UUID

ST-DDT:
@ryantheleach Please rememer the permission service must store the value to the config, and it should be human readable at that point.
I'm asking for this feature, because i would like to create some temporary worlds (for games) that are connected to a specific permission context, without having to somehow inject it to all joined players.

modwizcode:
Because this issue could take awhile to work out, I'm going to split the offline import functionality into a separate PR that can be merged quicker.

modwizcode:
Merged into PR commit https://github.com/SpongePowered/SpongeAPI/commit/142f2eaf011fb4e0972b8bcea9e6922232632a22.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -206,6 +206,49 @@
...
+     */
+    Optional<WorldProperties> renameWorld(WorldProperties worldProperties, String newName);
+
+    /**
+     * Deletes the provided world's files from the disk.
+     *
+     * @param worldProperties The world instance to delete
+     * @return True if the deletion was successful.
+     */
+    boolean deleteWorld(WorldProperties worldProperties);
```
spaceemotion: Will this only delete the world's `level.dat` file or the whole folder that contains the region files as well?

JBYoshi: The whole folder.

gabizou:
@modwizcode The merge commit has only one method from this PR, is this intended? Why were the others left out?

modwizcode:
there was only one? it's a merge into my PR branch?

JBYoshi:
They were added to modwizcode's PR in other commits.

PR #759, opened 2015-07-11
Chunk Manager

AIOOB:
I realise this is a short PR, but I am adding a way to get the ChunkManager as currently there is no other way.


src/main/java/org/spongepowered/api/Game.java
```diff
@@ -104,4 +105,11 @@
      */
     TeleportHelper getTeleportHelper();
 
+    /**
+     * Gets the {@link ChunkManager} used to force load chunks in the world.
```
me4502: I probably wouldn't specify the usage of the ChunkManager in the javadocs here.

bloodmc:
I'll be pushing ChunkManager implementation along with a getter in API soon. This should not be in Game but rather Server.

PR #762, opened 2015-07-13
The world view API

DDoS:
This adds methods for manipulating BiomeAreas and BlockVolume, as well as Extents. This include changing bounds, transforming views and copying data. More features will be added later (in future PRs) for iteration and filling (with support for multithreaded operations).

Implementation: https://github.com/SpongePowered/SpongeCommon/pull/102

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/ImmutableBiomeArea.java
+++ b/src/main/java/org/spongepowered/api/world/extent/ImmutableBiomeArea.java
@@ -24,11 +24,52 @@
  */
 package org.spongepowered.api.world.extent;
 
+import com.flowpowered.math.vector.Vector2i;
+import org.spongepowered.api.util.DiscreteTransform2;
+import org.spongepowered.api.util.PositionOutOfBoundsException;
+
 /**
  * An area containing biomes that can be accessed but not modified.
+ * The data will never change.
  *
  * @see BiomeArea
  */
-public interface ImmutableBiomeArea extends BiomeArea {
+public interface ImmutableBiomeArea extends UnmodifiableBiomeArea {
+
+    /**
+     * Returns a new area that is the same or smaller than the current area.
+     * This does not copy the biomes, it only provides a new view of the
+     * storage.
+     *
+     * @param newMin The new minimum coordinates in this area
+     * @param newMax The new maximum coordinates in this area
+     * @return The new area with the new bounds
+     * @throws PositionOutOfBoundsException If the new minimum and maximum
+     *     are outside the current area
+     */
+    @Override
+    ImmutableBiomeArea getBiomeView(Vector2i newMin, Vector2i newMax);
+
+    /**
+     * Returns a new area that is viewed through some transformation.
+     * This does not copy the biomes, it only provides a new view of the
+     * storage.
+     *
+     * @param transform The transformation to be applied
+     * @return The new area with the transform
+     */
+    @Override
+    ImmutableBiomeArea getBiomeView(DiscreteTransform2 transform);
+
+    /**
+     * Returns a new area that is translated so that
+     * {@link BiomeArea#getBiomeMin()} returns {@link Vector2i#ZERO}.
+     * This does not copy the biomes, it only provides a new view of the
+     * storage.
+     *
+     * @return The new area with its minimum at zero
+     */
+    @Override
+    ImmutableBiomeArea getRelativeBiomeView();
 
 }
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -49,6 +50,38 @@
...
+
+    /**
+     * Returns a new extent that is translated so that
+     * {@link Extent#getBlockMin()} returns {@link Vector3i#ZERO}.
+     * This does not copy anything, it only provides a new view of the
+     * storage.
+     *
+     * @return The new extent with its block minimum at zero
+     */
+    Extent getRelativeExtentView();
```
Zidane: Could you give a use case for methods like this? It could just be me tired but not seeing the point of them.

src/main/java/org/spongepowered/api/world/extent/ImmutableBiomeArea.java
```diff
@@ -24,11 +24,50 @@
...
+     * This does not copy the biomes, it only provides a new view of the
+     * storage.
+     *
+     * @param newMin The new minimum coordinates in this area
+     * @param newMax The new maximum coordinates in this area
+     * @return The new area with the new bounds
+     * @throws PositionOutOfBoundsException If the new minimum and maximum
+     *     are outside the current area
+     */
+    @Override ImmutableBiomeArea getBiomeView(Vector2i newMin, Vector2i newMax);
```
Zidane: Styling issue. Override shouldn't be on same line.

src/main/java/org/spongepowered/api/world/extent/StorageType.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.spongepowered.api.world.extent;
+
+/**
+ * Storage types that can be used for storing blocks and biomes.
+ * These are used when obtaining new copies. Storage obtained
+ * from the implementation should be assumed to be {@link #STANDARD}
+ * unless specified otherwise.
+ */
+public enum StorageType {
```
Zidane: Unsure about the name choice here. When I think "StorageType", I think the mechanism of how it is storing (FlatFile, MySQL, etc) not essentially when its safe to do so.

Naming hasn't ever been...a strong attribute of mine so I have no alternatives for you.

kenzierocks: StorageMethod? It's literally in the javadocs of `STANDARD` :P

DDoS: I was thinking about porting over Spout's compressed atomic storage, so I might be adding `COMPRESSED` and `COMPRESSED_THREAD_SAFE` to the types. This is why I opted for a more generic name.

src/main/java/org/spongepowered/api/world/extent/ImmutableBiomeArea.java
DDoS: The formatter was placing them on the same line... Maybe we should look into fixing it.

Zidane: Unfortunately, that isn't possible.

We elected for same line so it would fix all the "@Shadow"s in the impl being on their own line. We can't be selective in the formatting file.

DDoS: ???

Do you want them on the same line or not? This is what the formatter gave me.

Zidane: No they shouldn't be, as I said lol. Only "@Shadow" is on same line.

DDoS: Oh, OK...

src/main/java/org/spongepowered/api/world/extent/Extent.java
DDoS: It lets you get blocks in chunks with coordinates in range [0, 16). This is my resolution for an issue @gabizou (IIRC) opened recently. Right now all coordinates are absolute, so you need to get blocks from a chunk by adding an offset to the minimum coordinate. The other methods are meant for manipulation and world editing. This is the first step towards that API.

src/main/java/org/spongepowered/api/world/extent/StorageType.java
DDoS: @kenzierocks `Method` is to "active" to work here.

DDoS:
It's complete, including the implementation.

DDoS:
Anyone against me merging this tonight? Including the implementation.

PR #766, opened 2015-07-18
Add checks for interactable objects.

Xemiru:
_enhancement, low priority_

Previously, when registering an interaction event one must always consider the item in hand, as well as the block the player is targetting. In order for normal gameplay to not be interrupted, a blacklist of objects must be checked to make sure the player isn't intending to do something other than the plugin's function. This usually includes blocks like doors or chests, which do things within the world when you right-click them.

These two methods will eliminate having to have a blacklist of blocks and items to check with, and replace the former with a simple call checking whether or not the item/block has a reaction to being right-clicked.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -113,7 +113,16 @@ public interface BlockType extends CatalogType, Translatable {
      * @return Is affected by gravity
      */
     boolean isAffectedByGravity();
-
+    
+    /**
+     * Gets whether or not this block will cause a reaction when
+     * right-clicked.
+     * 
+     * <p>This includes blocks that open GUIs and switch their own states
+     * when right-clicked, e.g. opening/closing a door or opening a chest.</p>
+     */
+    boolean isInteractable();
+    
     /**
      * Gets if a block should be counted for statistics gathering.
      *
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -27,6 +27,7 @@ package org.spongepowered.api.block;
 import com.google.common.base.Optional;
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.data.DataHolder;
+import org.spongepowered.api.entity.living.Human;
 import org.spongepowered.api.item.ItemBlock;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
@@ -115,13 +116,14 @@ public interface BlockType extends CatalogType, Translatable {
     boolean isAffectedByGravity();
     
     /**
-     * Gets whether or not this block will cause a reaction when
-     * right-clicked.
+     * Gets whether or not this block will cause a reaction when right-clicked
+     * by the given {@link Human} entity.
      * 
-     * <p>This includes blocks that open GUIs and switch their own states
-     * when right-clicked, e.g. opening/closing a door or opening a chest.</p>
+     * <p>This includes blocks that open GUIs and switch their own states when
+     * right-clicked, e.g. opening/closing a door or opening a chest. Mod
+     * objects may make this value vary based on the given Human.</p>
      */
-    boolean isInteractable();
+    boolean isInteractable(Human cause);
     
     /**
      * Gets if a block should be counted for statistics gathering.
--- a/src/main/java/org/spongepowered/api/item/ItemType.java
+++ b/src/main/java/org/spongepowered/api/item/ItemType.java
@@ -27,6 +27,7 @@ package org.spongepowered.api.item;
 import com.google.common.base.Optional;
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.data.Property;
+import org.spongepowered.api.entity.living.Human;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
@@ -57,14 +58,15 @@ public interface ItemType extends CatalogType, Translatable {
     int getMaxStackQuantity();
     
     /**
-     * Gets whether or not this item will cause a reaction when
-     * right-clicked.
+     * Gets whether or not this item will cause a reaction when right-clicked by
+     * the given {@link Human} entity.
      * 
-     * <p>This includes items that open GUIs or switch their own states
-     * when right-clicked in hand, e.g. opening a Written Book to view its
-     * contents or drinking a Potion.</p>
+     * <p>This includes items that open GUIs or switch their own states when
+     * right-clicked in hand, e.g. opening a Written Book to view its contents
+     * or drinking a Potion. Mod objects may make this value vary based on the
+     * given Human.</p>
      */
-    boolean isInteractable();
+    boolean isInteractable(Human cause);
 
     /**
      * Gets the default {@link Property} of this {@link ItemType}.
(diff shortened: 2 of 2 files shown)


AlphaModder:
Perhaps add a `Player` (`Human`?) argument to the method for mod items which could only be used by some players, or items that only interact when clicking a certain kind of block with them, etc.

simon816:
The example scenario reminds me of the idea of event callbacks, #232.
There would be some sort of `ItemUseCallback` along with `BlockInteractCallback` and you can enable/disable/replace the default behaviour.
Not sure if callbacks are still planned to be added but I just felt a similarity.

Xemiru:
@AlphaModder Its valid to add, but when I think about it, technically that does still make it 'interactable' to some extent; typically when an item is unavailable for use in a mod the mod will explicitly tell you in chat. It has function, function that we shouldn't sit on when we try to add a new interaction event.

@simon816 It does show some similarity, but it still does require some form of blacklist check, which is what this PR is trying to get rid of.

Should these two methods be moved to `ItemStack` and `BlockState` instead? Or at least have another way of being added; I feel like plugins may want access to this as well in order to be able to say that they've changed a certain block to be interactable when it normally isn't in the base game. Not only that, but moving this to those two classes will make it easier for specific individual objects that have an environment (e.g. a multiblock structure) to check with.

ryantheleach:
>  requires some form of blacklist check

How does it still require a blacklist check?

Assuming the example is you want to prevent the player from interacting with any block in a region (and not necessarily cancel their item action)

Wouldn't you just let the event succeed but cancel the block interact callback if it exists, and leave the item interact callback alone?

AlphaModder:
@xemiru Well, sending a message could be a function. But there are plenty of items (wrenches come to mind) that do nothing except on a select few blocks.

Xemiru:
@ryantheleach Sorry, the way callbacks were planned to work slipped my mind writing that comment. Though, I'd think you'd cancel the item interact callback instead of the block interact one to prioritize the block's action before the item. As far as I'm concerned the right order for it would be entity -> block -> item.

On the matter of implementation, for Vanilla, we'd be able to assume item/block functions won't change to suddenly not exist (not the best, but its not likely that an item/block's function'll suddenly disappear).

For Forge however, I may have jumped the gun. Looking at forge code it doesn't seem like we'd easily be able to find out whether or not an item/block has a function, considering all mod items/blocks implement methods that're called when the player performs a given action with them. Not sure if reflection'll do the trick.

Do note that I'm not a guru on either part, someone else would be able to explain how the system works better than I can. If I've made any errors in what I've come up with above, feel free to correct me.

If there aren't any ideas on implementation then I suppose I should close this due to limitations.

ryantheleach:
I'm far from being a guru, sorry if I gave that impression. I'm legitimately wondering if it's possible.

If there is no way to tell it might just be a case of having three possible values, true, false, unknown. If there was a way for compatibility modules to hook into it and provide answers that might be enough.

But... It's far from a clean solution so pretty much anything would be preferable to that.

JBYoshi:
Here's an idea:

- A `PlayerInteractEvent` is fired immediately. Plugins use this to handle custom item handling, then cancel the event to stop Minecraft handling.
- Before control passes to the `Block`, a `PlayerInteractBlockEvent` is fired. Plugins use this to handle custom block interactions. To prevent plugins from doing their handling twice, `PlayerInteractBlockEvent` would NOT extend `PlayerInteractEvent`.

Xemiru:
@JBYoshi Thing is though, `PlayerInteractBlockEvent` will occur regardless of whether or not the block is 'interactable;' in the context of the events, 'interact' is equivalent to 'right-click' (sometimes left-click, too). In our case, its whether or not right-clicking the target object has a function. I don't think there's a specific event that says a block's function has been activated.

JBYoshi:
I was meaning a new event, I just called it `PlayerInteractBlockEvent`. I didn't know it already existed.

Xemiru:
@JBYoshi The problem still remains that we can't check whether or not an object actually has a function without Minecraft/Forge explicitly telling us. If it can't do that, then that event would be exactly the same as `PlayerInteractBlockEvent.`

JBYoshi:
That's not what I mean. They would be two separate events, called at two different places. Here's some code from `ItemInWorldManager.activateBlockOrUseItem()`, with the points at which the events would be called:
```java
if (useBlock) {
    if (event.useBlock != net.minecraftforge.fml.common.eventhandler.Event.Result.DENY) {
        // call PlayerInteractBlockEvent here
        result = iblockstate.getBlock().onBlockActivated(worldIn, pos, iblockstate, player, side, offsetX, offsetY, offsetZ);
    } else {
        thisPlayerMP.playerNetServerHandler.sendPacket(new S23PacketBlockChange(theWorld, pos));
        result = event.useItem != net.minecraftforge.fml.common.eventhandler.Event.Result.ALLOW;
    }
}
if (stack != null && !result && event.useItem != net.minecraftforge.fml.common.eventhandler.Event.Result.DENY) {
    int meta = stack.getMetadata();
    int size = stack.stackSize;
    // call PlayerInteractEvent here
    result = stack.onItemUse(player, worldIn, pos, side, offsetX, offsetY, offsetZ);
    if (isCreative()) {
        stack.setItemDamage(meta);
        stack.stackSize = size;
    }
    if (stack.stackSize <= 0) net.minecraftforge.event.ForgeEventFactory.onPlayerDestroyItem(thisPlayerMP, stack);
}
```

Xemiru:
I get that they would be called at separate times; it'd be pointless to make the event otherwise. What I'm trying to say is that we can't cancel the event at all because of how Minecraft/Forge does its implementations of block/item functions. We can only activate it as far as I'm concerned, unless reflection can check if a method is implemented or not. It'll be able to tell when an object is activated, yes, but it can't cancel it.

Though, after a bit of research it seems like Reflection may be able to do the trick. (pinging @ryantheleach for interest)
``` java
public boolean check(Class<? extends Block> blockClass) {
    if(blockClass.getMethod("onBlockClicked", World.class, Integer.class, Integer.class, Integer.class, EntityPlayer.class).getDeclaringClass() == Block.class) {
        return false; // method is being implemented by block
    }
    
    return true; method is being implemented by another class, presumably the mods'
}
```

The method `onBlockClicked` is not an abstract method; a block with no function shouldn't be trying to implement it. Though, some mods may try to implement a blank function for whatever reason. With how MC/Forge does its stuffs though, this could be one of the few ways to do it (untested).

JBYoshi:
I don't really get what you mean.

> we can't cancel the event at all because of how Minecraft/Forge does its implementations of block/item functions.

- What do you mean by “cancel”?
- What is it about the block/item code that prevents Sponge from canceling the event?

ryantheleach:
@Xemiru Using reflection constantly wouldn't be great... However we could use it to lazily initialize a field that describes whether the block is interact-able, seems like a fair solution to be honest.

ryantheleach:
> then cancel the event to stop Minecraft handling.

Acting on an event that is cancelled isn't a great way of doing it. Cancelled should represent that no action was taken by minecraft or plugins.

In order to replace the interaction that would have happened by default, the callback system should be used, allows for other plugins to see that the vanilla interaction was cancelled but a modded interaction was allowed.

Zidane:
I'm not seeing the point of this PR and you can't know at all with mods in place. This is what events really are for.

JBYoshi:
@Zidane I believe this is about distinguishing between interacting with a *block* and interacting with an *item*.

Example:
Let's say a plugin allows players to use a compass to open a menu. If the player right clicks with the compass in hand while looking at a block, the plugin would need to determine whether it should let the block handler execute or the compass handler execute. Vanilla standards (and, likely, user expectations) suggest that the block handler is invoked if and only if the player is *not sneaking* (this can be checked using the Data API) and if the block handler exists (which cannot be determined at the moment using only Sponge). I suggested an idea above that should help with mod compatibility.

Zidane:
@JBYoshi 

The PR is about adding two methods to BlockType/ItemType that, with mods, cannot be implemented whatsoever.

As for your proposals above, the new event framework changes how interact is done completely.

PR #770, opened 2015-07-21
Remove getScheduler() from Game

RobertHerhold:
I don't think there should be access to the `SchedulerService` directly through `Game`. As it is a service, it should be accessible through the `ServiceManager`, and it only bloats `Game` by making it accessible from there.

See SpongePowered/SpongeCommon#97 for the implementation update

### DIFF
--- a/src/main/java/org/spongepowered/api/Game.java
+++ b/src/main/java/org/spongepowered/api/Game.java
@@ -84,13 +84,6 @@ public interface Game {
     ServiceManager getServiceManager();
 
     /**
-     * Gets the scheduler used to schedule tasks.
-     *
-     * @return The scheduler
-     */
-    SchedulerService getScheduler();
-
-    /**
      * Get the command dispatcher used for registering and dispatching
      * registered commands.
      *


modwizcode:
The scheduler is a fairly popular service that will be used a lot, similar to the command service which is also accessible from `Game`. I don't think this is beneficial to remove, it will only bloat plugin code more.

RobertHerhold:
Regardless of its usefulness/popularity, I don't think it should be accessible in multiple places. If we keep putting common methods in Game, it will end up like the GameRegistry

gratimax:
Adding these methods into `Game` is not just for easier access. They assert that these services will _always_ be available (because the service manager does not do that by default) for anyone running Sponge. 

If you wanted to be consistent with this PR, you should have removed CommandService as mentioned, as well as EventManager, CommandService, and PluginManager (and actually make them services in Game in SpongeCommon, which for some reason hasn't been done yet...). I think that gives some good scope for what this change actually means.

Zidane:
I standby @gratimax 's decision here. If this is to change then you do it for all of them or none of them. Inconsistency is an API killer (and one we suffer from in some areas).

RobertHerhold:
@Zidane 

I would agree with you that it should be consistent (as I acknowledged in an edit to my second post). My real question is why it shouldn't be changed, accepting gratimax's previous explanation that it provides the assurance of the Scheduler (or whatever other service) always being there.

Deamon5550:
Why is it such an issue for you that everything should **only** be available from one location? Its the same case as `Living.getHealthData()` the method exists because A) its always available and B) using `entity.getData(HealthData.class).get()` or `game.getServiceManger().provide(Scheduler.class).get()` everytime is a PITA.

 I would think of the ServiceManager as more of a provider for things that you don't know are there (plugin provided services) rather than the default services. Not to say that you can't get the Scheduler from the ServiceManager if you so choose, just that an easier alternative exists.

PR #771, opened 2015-07-21
Added Sneaking to Entities

joshgarde:
Sneaking! Yay!

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -245,5 +245,19 @@ public interface Entity extends Identifiable, DataHolder, DataSerializable {
      * within one game tick.
      */
     void remove();
+    
+    /**
+     * Returns whether this entity is currently sneaking
+     * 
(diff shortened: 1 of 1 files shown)


ryantheleach:
There is no guarantee that Entities can sneak. So why add these methods? It's already covered in the DataAPI.

joshgarde:
Because there really isn't any way to distinguish between an entity that can sneak and one that can't sneak within MC's code. In MC, an entity - Regardless of type - can sneak.

ryantheleach:
Until when? What happens if they remove sneaking from entities that don't have sneaking animations? If you forcibly make a bat sneak does anything change?

Anyway, it can still be handled by Data manipulators just fine as is, even if *at the moment* all entities can sneak.

joshgarde:
It was worth a try. I suppose future-proofing is favorable over my personal convenience.
As for effects on entities without sneaking animations, I haven't done the testing so I can't say for sure.

ryantheleach:
Yeah the DataAPI at the moment is a little verbose but it's getting better when the Value API is merged.

PR #772, opened 2015-07-21
Support for the resource pack implementation

JBYoshi:
This PR:
- ~~Adds `PlayerResourcePackStatusEvent` to the event factory.~~ This was taken care of with the events refactor.
- Replaces `URL` with `URI`. This allows Minecraft's `level://` resource packs (used for `resources.zip` insode a world) to be represented using Sponge (as `java.net.URL` does not support unknown protocols, e.g. `level://`).
- Renames `ResourcePackStatus.PACK_FILE_FORMAT_NOT_RECOGNIZED` to `FAILED`.
- Adds `Server.getDefaultResourcePack()`.

Implemented by SpongePowered/SpongeCommon#100.

### DIFF
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -30,6 +30,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.network.ChannelRegistrar;
+import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.service.world.ChunkLoadService;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.sink.MessageSink;
@@ -366,4 +367,12 @@ public interface Server extends ChannelRegistrar {
      * @return The current ticks per second
      */
     double getTicksPerSecond();
+
+    /**
+     * Gets the default resource pack. The default resource pack is sent to
+     * players when they join the server.
+     * 
+     * @return The default resource pack
+     */
+    Optional<ResourcePack> getDefaultResourcePack();
 }
--- a/src/main/java/org/spongepowered/api/event/entity/living/player/ResourcePackStatusEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/living/player/ResourcePackStatusEvent.java
@@ -69,12 +69,12 @@ public interface ResourcePackStatusEvent extends GameEvent {
         DECLINED(false),
 
         /**
-         * The pack was not a .zip file.
+         * The client failed to download the resource pack.
          */
-        PACK_FILE_FORMAT_NOT_RECOGNIZED(false),
+        FAILED(false),
 
         /**
-         * The pack URL was successfully loaded. This does not mean that pack
+         * The pack URI was successfully loaded. This does not mean that pack
          * was loaded, as the vanilla client sends this even when encountering a
          * 404 or similar.
          */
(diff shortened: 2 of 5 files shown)


Mumfrey:
You should be using `java.net.URI` rather than `java.net.URL` to be honest.

JBYoshi:
Well, in that case, it looks like I just need to change URL to URI everywhere in the resource packs area and that should take care of the `URLResourcePack`/`WorldResourcePack` stuff.

JBYoshi:
(Although I will need to special case in the implementation.)

ST-DDT:
Do you have a test plugin to verify that your implementation works and how the API can be used?

JBYoshi:
Not at the moment.

JBYoshi:
Here's a test plugin that you can use:
https://gist.github.com/JBYoshi/e1e21f2f243d13a0110e

Type /pack to send yourself a resource pack. When the client sends a `PlayerResourcePackStatusEvent`, the URI and status are printed to the chat.

kashike:
I might be missing something, but it doesn't appear there are any methods on `Player` (or a parent of) to get resource pack information, which would mean using the event would be the only way.
It would be helpful to have methods to get information about the resource pack the player sent a status for.

AlphaModder:
@kashike I was the one who wrote the API for this, and that was by design. Since the plugin can monitor the event and maintain its own information, and since clients could change resources packs after sending the information undetectably, it didn't seem needed.

kashike:
@AlphaModder I don't agree with requiring plugins to store that information.

In my opinion that is like saying I should store health information when a health-related event is posted, because I can monitor and maintain my own information, and because the data can change. 

```java
if (player.getResourcePackStatus() == ResourcePackStatus.FAILED) {
    // meow
}
```
```java
if (MyPlugin.get().getPlayerResourcePackStatusOf(player) == ResourcePackStatus.FAILED) {
    // meow
}
```

AlphaModder:
The thing with health is that you know how and when it changes. The client can change resources packs and not say so, unlike an entity and it's health. There is no way for the server to know. Still, if you think that we should make a method for this I don't have a problem with it. Just explaining why I didn't add one in the first place.

JBYoshi:
@kashike I don't see why you would need to store the status of the resource pack download. Could you give me some example use cases?

kashike:
@JBYoshi An example use case would be to check the status and check the hash. I'm really not sure how to explain an example usage right now.
A plugin author would need to either:
a) use an api method
b) storeand maintain an accurate value their self

JBYoshi:
Usually, in the cases that I can think of, the state wouldn't need to be stored outside of the `PlayerResourcePackStatusEvent`. For example, waiting until a player gets the resource pack to start a minigame, or sending the player out of a minigame if they deny the resource pack.

Mumfrey:
I'm with @AlphaModder on this one, we shouldn't be representing non-stateful data as stateful, if plugins want to persist the "last known state" themselves that's fine, but the API shouldn't give the false illusion that the current state is known.

Zidane:
I agree with Mumfrey's assesment.

Considering there is no way to ask the client what their ResourcePack actually is and the client will send what its ResourcePackStatus is when it determines it wants to, you can't trust anything about it. You can't cache it, you can't give a contract that gives the impression that "this is the ResourcePack this player currently has".

Having the event for when the client decides to update its status is as much support as Sponge will be able to give.

bloodmc:
@Zidane @Mumfrey any other issues with this PR? @JBYoshi has implementation ready

JBYoshi:
Would there be any need for plug-ins to get the default resource pack from the server? I could add that quite easily.

Issue #768, opened 2015-07-19
Common interface for text formatting types


kashike:
There is no common "`Formatting`" interface for __all__ text formatting types, nor is it nice to use `CatalogType` to reference any `Base` type.

[`TextColor`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextColor.java) extends [`CatalogType`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/CatalogType.java)
[`TextColor.Base`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextColor.java#L54) extends [`TextColor`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextColor.java) and [`BaseFormatting`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/BaseFormatting.java)

[`TextStyle`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextStyle.java) extends/implements __nothing__
[`TextStyle.Base`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextStyle.java#L511) extends [`TextStyle`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/TextStyle.java), [`BaseFormatting`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/format/BaseFormatting.java), and [`CatalogType`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/CatalogType.java)

ST-DDT:
So if I get you right you want to add the `Formatting` interface (extending nothing) and `TextColor`, `TextStyle` and `BaseFormatting` extend `Formatting`.

But what do you want to use that interface for?

But you are right with `TextColor` extending `CatalogType` being wrong, since the is no id or name for a random color. Only `TextColor.Base` should extend it.

kashike:
You summed it up, @ST-DDT. `Formatting` was just an example name. `TextFormatting` sounds much better,

zml2008:
[...]

PR #775, opened 2015-07-21
More formal CatalogType definition

ST-DDT:
Solves https://github.com/SpongePowered/SpongeAPI/issues/710

Currently there is only a very unspecific definition what an `CatalogType` is.
This PR try to solves that by adding a very precise definition on what it is and what is connected to that.

**`CatalogType`**

> Represents a type of a catalog that can be used to identify types without using an Enum. 
>
> Implementing classes must meet the following requirements for all its members including those which are not listed in the classes listed in the `CatalogedBy` annotation: 
* Every pair (a, b) that matches ... 
  * `a.equals(b)` must also match `a == b`
  * `a.getId().equalsIgnoreCase(b.getId())` must also match `a == b`

**`CatalogedBy`**

> Represents a class that is intended to represent a type of enum, without using Enum. The class marked as CatalogedBy must have a registrar class that can be queried for all types and subtypes of the catalog. 
>
>All Classes mentioned in the CatalogedBy annotation must meet the following requirements:
* The values referenced by catalog classes must remain the same and valid at all times.
* The variables in catalog classes may link to null if the given value is no longer valid and no appropriate alternative can be found. If no alternative could be found and if there probably won't be a new one in the future then the field should be marked as deprecated and should be removed after a grace period or with the next big release of minecraft, SpongeAPI or the containing artifact.
* It is possible for two or more different variables to link to the same value. This includes both simple "well known" aliases and features that been merged together or that were originally very similar and one being removed. 
* It is also possible that one or more values are not (yet) listed in the catalog classes (Especially plugin provided ones).

This is a very formal definition so if anybody like to reword/rephrase anything feel free to comment.
I tried to cover all cases that needed to be covered and have been discussed in #710, but we may have missed something.

The new definition walks along with the issue exposed in https://github.com/SpongePowered/SpongeAPI/issues/768: 
* `TextColor` shouldn't be a `CatalogType` only `TextColor.Base` should.

It also seems worth discussing the definition of `GameRegistry.getType(Class, String)` in regards to plugin/mod provided type instances and type classes.
* IMO plugin/mod provided instances and type classes should be listed/returned there as well. Due to the impact this might have I haven't added this here yet.
  * `<T extends CatalogType> GameRegistry.registerType(Class<T>, T... type) : void` - 
Which would be sealed after server startup/init phase

### DIFF
--- a/src/main/java/org/spongepowered/api/CatalogType.java
+++ b/src/main/java/org/spongepowered/api/CatalogType.java
@@ -25,24 +25,40 @@
 package org.spongepowered.api;
 
 /**
- * Represents a type of a catalog that can be used to identify types
- * without using an {@link Enum}.
+ * Represents a type of a catalog that can be used to identify types without
+ * using an {@link Enum}.
+ *
+ * <p>Implementing classes must meet the requirement for all its members that if
+ * any of <code>`a.equals(b)`</code>, <code>`a == b`</code>, or
+ * <code>`a.getId().equalsIgnoreCase(b.getId())`</code> are true then all must
+ * be true. This must also apply to members which are not listed in
+ * the classes listed in the
+ * {@link org.spongepowered.api.util.annotation.CatalogedBy CatalogedBy}
+ * annotation but nevertheless implement this interface.</p>
  */
 public interface CatalogType {
 
     /**
-     * Gets the unique identifier of this {@link CatalogType}. The identifier
-     * can be formatted however needed.
+     * Gets the unique identifier of this {@link CatalogType}. The identifier is
+     * case insensitive, thus there cannot be another instance with a different
+     * character case. The id of this instance must remain the same for the
+     * entire duration of its existence. The identifier can be formatted however
+     * needed.
+     *
+     * <p>A typical id will look like <code>`modId:name`</code> or
+     * <code>`minecraft:name`</code>. However the prefix may be omitted for
+     * default/vanilla minecraft types.</p>
      *
      * @return The unique identifier of this catalog type
      */
     String getId();
 
     /**
-     * Gets the unique human-readable name of this individual
-     * {@link CatalogType}.
+     * Gets the human-readable name of this individual {@link CatalogType}. This
+     * name is not guaranteed to be unique. This value should not be used for
+     * serialization.
      *
-     * @return The uniquely identifiable name of this catalog type
+     * @return The human-readable name of this catalog type
      */
     String getName();
 
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -87,9 +87,10 @@ public interface GameRegistry {
      * game version changes.</p>
      *
      * @param typeClass The class of the type of {@link CatalogType}
-     * @param id The string id of the catalog type
+     * @param id The case insensitive string id of the catalog type
      * @param <T> The type of catalog type
      * @return The found catalog type, if available
+     * @see CatalogType
      */
     <T extends CatalogType> Optional<T> getType(Class<T> typeClass, String id);
 
--- a/src/main/java/org/spongepowered/api/util/annotation/CatalogedBy.java
+++ b/src/main/java/org/spongepowered/api/util/annotation/CatalogedBy.java
@@ -37,6 +37,26 @@ import javax.annotation.Nonnull;
  * using {@link Enum}. The class marked as {@link CatalogedBy} must have a
  * registrar class that can be queried for all types and subtypes of the
  * catalog.
+ *
+ * <p>All Classes mentioned in the CatalogedBy annotation must meet the
+ * following requirements:</p>
+ * <ul>
+ * <li>The values referenced by catalog classes must remain the same and valid
+ * at all times.</li>
+ * <li>The variables in catalog classes may link to null if the given value is
+ * no longer valid and no appropriate alternative can be found. If no
+ * alternative could be found and if there probably won't be a new one in the
+ * future then the field should be marked as deprecated and should be removed
+ * after a grace period or with the next big release of minecraft, SpongeAPI or
+ * the containing artifact.</li>
+ * <li>It is possible for two or more different variables to link to the same
+ * value. This includes both simple "well known" aliases and features that been
+ * merged together or that were originally very similar and one being removed.
+ * </li>
+ * <li>It is also possible that one or more values are not (yet) listed in the
+ * catalog classes (Especially plugin provided ones).</li>
+ * </ul>
+ * </p>
  */
 @Nonnull
 @Target(ElementType.TYPE)


Deamon5550:
> Instances are immutable

I don't think this is valid, several types could have / already have mutable properties and restricting this is pointless.

ST-DDT:
@Deamon5550 Can you give men an example?
But of course i can change that.

Deamon5550:
@ST-DDT [BiomeType](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/world/biome/BiomeType.java)

ST-DDT:
Oh thanks.
I really hoped I could avoid the mutability.

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * the same and valid.</li>
+ * <li>The value of a.getName() identifies the reference uniquely and must
+ * remain the same and valid.</li>
+ * <li>All methods that allow the combination of a CatalogType with other
+ * objects may not return a CatalogType.</li>
+ * </li>
+ * </ul>
+ * </p>
+ *
+ * <p>All Classes mentioned in the CatalogedBy annotation must meet the
```
Deamon5550: Should be moved to the CatalogedBy class javadoc

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * following requirements:</p>
+ * <ul>
+ * <li>The values referenced by catalog classes must remain the same and valid
+ * at all times.</li>
+ * <li>The variables in catalog classes may link to null if the given value is
+ * no longer valid and no appropriate alternative can be found. Thus two or more
+ * different variables can link to the same value. It is also possible that one
+ * or more values are not listed in the catalog classes.</li>
+ * <li>The variables in catalog classes may not be renamed or removed after
```
Deamon5550: In my opinion this isn't a very good requirement, a better one would be something requiring that removal of keys which are no longer valid must follow a sane deprecation policy.

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * annotation classes:
+ * <ul>
+ * <li>Every pair (a, b) that matches ...
+ * <ul>
+ * <li>a.equals(b) must also match a == b</li>
+ * <li>a.getId().equalsIgnoreCase(b.getId()) must also match a == b</li>
+ * <li>a.getName().equalsIgnoreCase(b.getName()) must also match a == b</li>
+ * </ul>
+ * </li>
+ * <li>The {@link GameRegistry#getType(Class, String)} method must serve all
```
Deamon5550: this should be in here

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * values independently from the case of the real id value.
+ * <ul>
+ * <li>It may but does not need to return the values for "well known aliases"
```
Deamon5550: this shouldn't be in here, or anywhere

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * </li>
+ * </ul>
+ * </li>
+ * <li>The value of a.getId() identifies the reference uniquely and must remain
```
Deamon5550: These two are just rehashing the javadocs on the methods, better to not repeat yourself

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
...
+ * the same and valid.</li>
+ * <li>The value of a.getName() identifies the reference uniquely and must
+ * remain the same and valid.</li>
+ * <li>All methods that allow the combination of a CatalogType with other
```
Deamon5550: This doesn't belong here, and I'm not sure its actually a valid requirement

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,8 +25,52 @@
 package org.spongepowered.api;
 
 /**
- * Represents a type of a catalog that can be used to identify types
- * without using an {@link Enum}.
+ * Represents a type of a catalog that can be used to identify types without
+ * using an {@link Enum}.
+ *
+ * <p>Implementing classes must meet the following requirements for all its
```
Deamon5550: Should be rephrased to 'Implementing classes will meet...'

ST-DDT:
I'll try to fix that this weekend.

ST-DDT:
@Deamon5550 Better?

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,14 +25,36 @@
...
+ * <li>Every pair (a, b) that matches ...
+ * <ul>
+ * <li>a.equals(b) must also match a == b</li>
+ * <li>a.getId().equalsIgnoreCase(b.getId()) must also match a == b</li>
+ * <li>a.getName().equalsIgnoreCase(b.getName()) must also match a == b</li>
+ * <li>a.getName().equalsIgnoreCase(b.getId()) must also match a == b</li>
+ * <li>a.getId().equalsIgnoreCase(b.getName()) must also match a == b</li>
+ * </ul>
+ * </li>
+ * <li>The {@link GameRegistry#getType(Class, String)} method must serve all
```
Deamon5550: This doesn't belong here

Deamon5550:
I'm having a real problem with the requirements on the name. I feel like its really restricting us into always implementing `getName` to return the same as `getId`. Currently in the implementation in almost every case this is exactly what we do and I feel like this is breaking the use of `getName` which is to be a human-friendly identifier for the object. If we add these restrictions to the name then we may as well remove the method all together as it will serve no purpose. We will always be forced into implementing it to be identical to `getId` as no other identifier is guaranteed to be unique for an object.

ST-DDT:
> I'm having a real problem with the requirements on the name.

@Deamon5550 Which requirement on name?

> * `a.getName().equalsIgnoreCase(b.getId())` must also match `a == b`
* `a.getId().equalsIgnoreCase(b.getName())` must also match `a == b`

Or are you referring to something else?

>  I feel like its really restricting us into always implementing getName to return the same as getId.

Why? That not specified at all! It just states that a name shouldn't match the id of another instance and vice versa. IMO if this is not guaranteed then something is wrong with the `CatalogType` impl.

> As @Aaron1011 just pointed out to be the best action may be to simply make the objects translatable. Then we could just have a getName which is a shortcut for getting the english translation of the object.

If it would be that easy... Unfortunately there are no translations for most of them. (See one of my other PRs that add most remaining missing `Translatable`s #837 )

Deamon5550:
We're restricted into implementing `getName` to return `getId` because there is no other identifier that is unique. Take BlockTypes for example, the only unique identifier we can use is the registry key, the unlocalized name is not unique and also not human-friendly (its the translation key). The name needs to be human-friendly, and the only human friendly identifier we have is the english translation or one that we make ourselves, and that will not always be unique.

ST-DDT:
@Deamon5550 I removed the name must be unique constraints.

Deamon5550:
Looks good to me @SpongePowered/developers

We should also update the impl to not return `getId` from `getName` as we currently do in the majority of cases.

Deamon5550:
@ST-DDT Please squash this to a single commit so it is ready to be merged

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,24 +25,35 @@
...
+ * <li>a.getId().equalsIgnoreCase(b.getId()) must also match a == b</li> </ul>
+ * </li> </ul> </p>
  */
 public interface CatalogType {
 
     /**
-     * Gets the unique identifier of this {@link CatalogType}. The identifier
-     * can be formatted however needed.
+     * Gets the unique identifier of this {@link CatalogType}. Its uniqueness is
+     * guaranteed to the include every character case. The id of this instance
```
Deamon5550: Perhaps something like 'Its uniqueness is guaranteed regardless of character case'

ST-DDT:
@JBYoshi Fixed
@Deamon5550 Done

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,24 +25,41 @@
...
- * without using an {@link Enum}.
+ * Represents a type of a catalog that can be used to identify types without
+ * using an {@link Enum}.
+ *
+ * <p>Implementing classes must meet the following requirements for all its
+ * members, including those which are not listed in the classes listed in the
+ * {@link org.spongepowered.api.util.annotation.CatalogedBy CatalogedBy}
+ * annotation:
+ * <ul>
+ * <li>Every pair (a, b) that matches:
```
Deamon5550: I feel like its a little easier to simply say 'if any of `a.equals(b)`, `a == b`, or `a.getId().equalsIgnoreCase(b.getId())` are true then all must be true' rather than the nested lists. So something like:

```
Implementing classes must meet the requirement for all its members that if any of `a.equals(b)`, `a == b`, or 
`a.getId().equalsIgnoreCase(b.getId())` are true then all must be true. This must also be also applies to members 
which are not listed in the classes listed in the {@link org.spongepowered.api.util.annotation.CatalogedBy 
CatalogedBy} annotation but nevertheless implement this interface.
```

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,24 +25,41 @@
...
  */
 public interface CatalogType {
 
     /**
-     * Gets the unique identifier of this {@link CatalogType}. The identifier
-     * can be formatted however needed.
+     * Gets the unique identifier of this {@link CatalogType}. The identifier is
+     * case insensitive, thus there cannot be another instance with a different
+     * character case. The id of this instance must remain the same as long as
+     * the server is running. The identifier can be formatted however needed.
```
Deamon5550: Change this to 'must remain the same for the entire duration of its existence'. 'As long as the server is running' means something else entirely.

Deamon5550: and perhaps specify that a typical format for the identifier is `modid:name` or `minecraft:name`

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -25,24 +25,40 @@
...
 /**
- * Represents a type of a catalog that can be used to identify types
- * without using an {@link Enum}.
+ * Represents a type of a catalog that can be used to identify types without
+ * using an {@link Enum}.
+ *
+ * <p>Implementing classes must meet the requirement for all its members that if
+ * any of <code>`a.equals(b)`</code>, <code>`a == b`</code>, or
+ * <code>`a.getId().equalsIgnoreCase(b.getId())`</code> are true then all must
+ * be true. This must also be also applies to members which are not listed in
```
Deamon5550: `This must also be also applies to` -> `This must also apply to`

ST-DDT: Ups, I didn't read your suggestion careful enough.

Deamon5550: Oh ha...

Deamon5550:
Merged in 8cbfafec442c45

PR #776, opened 2015-07-22
[WIP] API for getting the inhabitted time and regional difficulty of a chunk.

AlphaModder:
Exactly what it says on the tin, not sure what I can add. For those who are not aware of regional difficulty, here is the [minecraftwiki article](http://minecraft.gamepedia.com/Difficulty#Regional_difficulty).

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Chunk.java
+++ b/src/main/java/org/spongepowered/api/world/Chunk.java
@@ -119,5 +119,4 @@ public interface Chunk extends Extent {
      * @return The regional difficulty percentage for this chunk
      */
     float getRegionalDifficultyPercentage();
-
 }
--- a/src/main/java/org/spongepowered/api/world/Chunk.java
+++ b/src/main/java/org/spongepowered/api/world/Chunk.java
@@ -84,4 +84,40 @@ public interface Chunk extends Extent {
      * @return Whether or not the chunk unloaded
      */
     boolean unloadChunk();
+
+    /**
+     * Gets the number of ticks players have been present in this chunk, used
+     * for calculation of the regional difficulty factor. In vanilla, it is
+     * increased by the number of players in the chunk every tick, and is capped
+     * at 3,600,000 ticks (50 hours).
+     */
+    int getInhabittedTime();
+
+    /**
+     * 
+     * Gets the regional difficulty factor for this chunk. In vanilla, it is
+     * dependent on the playtime of the world, inhabited time of the chunk, the
+     * phase of the moon, and the current difficulty setting. This number ranges
+     * from 0.75-1.5 on easy, 1.5-4.0 on normal, and 2.25-6.75 on hard.
+     * 
+     * This value is used for display only in vanilla.
+     * 
+     * @return The regional difficulty factor for this chunk
+     */
+    float getRegionalDifficultyFactor();
+
+    /**
+     * Gets the regional difficulty percentage for this chunk. It is calculated
+     * by taking the regional difficulty factor and using the following rules:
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/Chunk.java
```diff
@@ -84,4 +84,39 @@
...
+     * Gets the regional difficulty factor for this chunk. In vanilla, it is
+     * dependent on the playtime of the world, inhabited time of the chunk, the
+     * phase of the moon, and the current difficulty setting. This number ranges
+     * from 0.75-1.5 on easy, 1.5-4.0 on normal, and 2.25-6.75 on hard.
+     * 
+     * This value is used for display only in vanilla.
+     * 
+     * @return The regional difficulty factor for this chunk
+     */
+    float getRegionalDifficultyFactor();
```
ST-DDT: Why did you choose floats?
Other implementations might use doubles.

It would be cool to get the min and max values as well.

Why didn't you use a DataManipulator?

JBYoshi: > Why didn't you use a DataManipulator?

`Chunk`s are not `DataHolder`s.

AlphaModder: @ST-DDT Good point. I'll switch it.

gabizou:
So far, this seems like a leak from implementation, the "difficulty factor" and the fact that a `Chunk` is tracking an "inhabited time". What purpose do these two serve? What functionality does this provide? What is the "metric" of "difficulty" this is used for? What does this have to do with `Chunk`?

From reading the wiki article, it's use is for spawning algorithms, which honestly I have more understanding of a `Spawner` representation than just the `Chunk` containing this information. In essence, this would make most sense where it is used, not where it's collected from.

AlphaModder:
@gabizou The idea is that plugins would have access to the difficulty so they could modify their behavior inside a chunk if they wanted, and I added the factor because it seems to be the client's preferred way of displaying the difficulty to the player. You may be right about the inhabited time, however.

gabizou:
> @gabizou The idea is that plugins would have access to the difficulty so they could modify their behavior inside a chunk if they wanted, and I added the factor because it seems to be the client's preferred way of displaying the difficulty to the player. You may be right about the inhabited time, however.

Again, this is something I wouldn't feel being right to reside in a `Chunk` but quite possibly a `Spawner` object system that Minecraft already uses, since the `Spawner` itself is the object that decides to process over this information, not the `Chunk`.

AlphaModder:
@gabizou I only put it in chunk because it is dependent (in vanilla) on the chunk you ask. But putting it in a `Spawner` or something seems sorta strange. It seems to me there's nothing that says the value can only be used for spawning mobs, regardless of what vanilla does with it.

Deamon5550:
Ignoring the issue of these values being super arbitrary. A better place for this could be in `Extent` (or more likely an extent subclass) with added parameters for the location you are querying for.

JBYoshi:
@Deamon5550 Don't forget `Location`.

gabizou:
> @Deamon5550 Don't forget Location.

The problem with `Location` is that it is for a single point in the world where a `Player` or `Entity` may take up a volume of `Location`s that may have varying "difficulty" so to speak. Of course though, exposing through an `Extent` or subset thereof would be optimal, it feels as though `Location` already is blown up with regards to the methods it contains.

gabizou:
@AlphaModder I didn't realize how long ago this has been since it's been updated. Any chance of completion or update?

Issue #639, opened 2015-05-05
Add Text.format() [similar to String.format()]


boformer:
The issue is described here:

https://github.com/SpongePowered/Sponge/issues/261#issuecomment-98931861

gratimax:
I like this idea. I'll try to take a look and write a simple Text#format/Texts.format implementation when I get time to do so.

It may be possible to unify Text.Translation and text formatting somehow, since they both technically use the same system to replace labels. The difference is that a prospective Text formatting system replaces string templates with text and the current translation system replaces them with strings.

RobertHerhold:
Also, a `Text#replace()` would be greatly appreciated (also similar to `String#replace()`)

TomyLobo:
For your use case, a "Text.replace feature wouldn't be useful, since then you'd have an O(warp_count*string_length) complexity as opposed to just O(string_length).

TomyLobo:
@boformer You could xmlescape the arguments and then just use String.format.

boformer:
@TomyLobo I want a complete, integrated solution:

Input: Text object with color, styles, text actions, translatables
Output: Similar Text object with placeholders replaced.

TomyLobo:
Well we could have a Texts.xml().format(format, args...) function, of course.
But you want actual text objects, which is an order of magnitude harder to do.
How about xml/json formatting now and Text object formatting later, when it's done?

gratimax:
I've summarized my current thoughts on text formatting here: https://gist.github.com/gratimax/abaa32a49b5b82c08c13.

It's a short description of a format that I worked out. It is more opinionated in order to have a better templating experience, but I think that is a good thing. I've yet to work out if one can insert arbitrary JSON in `&{}` statements or if you can template within them.

TomyLobo:
Oh, didn't know you meant to make a 5th way of specifying colors (§, json, Text objects, xml, your inline json thing).
I thought by Text.format you meant something like String.format...
So what does this text representation offer that none of the other 4 can offer?

ST-DDT:
The main issue i have with this suggestion is that it introduces a new Text syntax that is used nowhere else and in the worst case scenario you have to convert your existing Text object to a special (legacy) string to insert it into your Text.
````
Text displayName = player.getDisplayName();
Texts.format(myMessage, displayName.toFormatString());
````

gratimax:
@TomyLobo @ST-DDT 

[...]

PR #779, opened 2015-07-26
Add Texts.format(...) method

ST-DDT:
Adds https://github.com/SpongePowered/SpongeAPI/issues/639
Implemented in https://github.com/SpongePowered/SpongeCommon/pull/108

Currently it is quite hard to use message bundles and insert formated `Text`s in there.
This PR solves this by adding the following features
* `Text.Placeholder`
  * A placeholder with a key, that can be replaced with any other Object wrapped in a `Text`. If it is not replaced it will display a default message.
* `TextRepresentable`
  * An interface for all those objects that have a `Text` representation (Ex: `ItemStack`s)
* `ItemStack` is `Translatable`
* `Texts.format(Text, Map<String,?>)`
  * Creates a new `Text` with all possible `Placeholder`s replaced. Uses Strings `player`.
* `Texts.format(Text, Object...)`
  * Creates a new `Text` with all possible `Placeholder`s replaced. Uses indexes `0`.

**Tasks**
- [x] Create Placeholder
- [x] Implement `Texts.format(...)`
- [x] Optimize `Texts.of(Object...)`
  - `TextRepresentable` (`ItemStack`s + `Achievement`s + custom)
  - Made `TextColor.NONE.toString()` and `TextStyle.NONE.toString()` more readable
- [x] Create Test-Plugin

**Usage: Create Template**
````java
// "&4Player {killer} have killed {killed} with {weapon=with his bare hands}"
TextBuilder builder = Texts.builder();
builder.color(TextColors.DARK_AQUA);
builder.append(Text.of("Player "));
builder.append(Text.placeholderOf("killer"));
builder.append(Text.of(" have killed "));
builder.append(Text.placeholderOf("killed"));
builder.append(Text.of(" with "));
builder.append(Text.placeholderOf("weapon", "with his bare hands")); // with fallback
Text text = builder.build();
````
The created `Text` instance is reusable.

**Usage: Key-Value**
````java
Text someTemplateText = ...;
player.sendMessage(someTemplateText);
// Displays :"&4Player {killer} have killed {killed} with his bare hands"
Map<String,Object> replacements = Maps.new();
replacements.put("killer", killer.getName());
replacements.put("killed", Texts.of(TextColors.GRAY, killed));
ItemStack weapon = killer.getItemInHand();
replacements.put("weapon", weapon);
player.sendMessage(Texts.format(someTemplateText, replacements)); 
// Displays: "&4Player PvPMonster have killed &7Innocent&4 with Stick of Doom"
````

**Usage: Index**
````java
Text someTemplateText = ...; // With {0} as placeholder
player.sendMessage(someTemplateText);
// Displays :"&4Player {0} have killed {1} with his bare hands"
player.sendMessage(Texts.format(someTemplateText, killer.getName(), Texts.of(TextColors.GRAY, killed), killer.getItemInHand())); 
// Displays: "&4Player PvPMonster have killed &7Innocent&4 with Stick of Doom"
````

**Test Plugin**
[TextFormat.jar](https://dl.dropboxusercontent.com/u/16999313/Sponge/TextFormat.jar) (Plugin)
[TextFormat.zip](https://dl.dropboxusercontent.com/u/16999313/Sponge/TextFormat.zip) (Source)
(works already)
Try the following commands to see the resutls:
* `/textformat`
* `/textformat PvPMonster`
* `/textformat PvPMonster Innocent`
* `/textformat PvPMonster Innocent &6Stick of Doom`
* `/textformat <YourUsername> Innocent` // Uses the item in hand
* `/textjson` - Proves the json (de-)serialization works
* `/textxml` - Proves the xml(de-)serialization works

**Things not part of this PR**
* `Text#replace(String, Text)`
  * There are too few possibilities where this method can be effectively be used.
  * There are too much variants of this method that would be needed
    * replaceFirst- only the first one
    * replaceAll - every one
    * replaceEagerFirst - only the first one although it might be splitted into multiple `Text`s (children)
    * replaceEagerAll - every one including this which might be splitted into multiple `Text`s (children)
  * Easy to implement yourself with optimisations based on your needs otherwise it might be slow
    * Example: Replaces all words in a `Text` with `warp actions` if such a warp exists
      * API-Provided: O(warps * text-instances * text-length) - for warp replace all warpname with warp action
      * Custom impl: O(text-instances * text-length) - for each word in the text search for a warp in a map
* A new enhenced plain-text `TextRepresentation` as suggested by @gratimax in this [comment](https://github.com/SpongePowered/SpongeAPI/issues/639#issuecomment-109789419)
  * This feature can be implemented nevertheless it just does not fit in the topic of this PR

### CODE REVIEW

File: src/main/java/org/spongepowered/api/text/Texts.java [hunk 1]
```diff
@@ -172,6 +206,70 @@ public static Text of(Object... objects) {
...
+    }
+
+    private static Text formatNoChecks(Text template, Map<String, ?> replacements) {
+        // Is this a placeholder that should be replaced?
+        if (template instanceof Placeholder) {
+            Object replacement = replacements.get(((Placeholder) template).getKey());
+            // Only replace
+            if (replacement != null) {
+                // Copy color and style from placeholder
+                return Texts.of(template.getColor(), template.getStyle(), replacement);
```
zml2008: also click, hover, etc.

And if the replacement is a Text, you might want to have it just return the replacement itself

ST-DDT: @zml2008
> also click, hover, etc.

Click and Hover are not supported by Texts.of, 
but i will add that too. I did not add that before because IMO the textactions are always provided by the replacements itself.
A clickaction like OpenLink does not make sense with a variable text. And a hover text usually shows a hint/description for the object it describes.

> And if the replacement is a Text, you might want to have it just return the replacement itself

No, because in that case the color and style... are lost. This would be inconsistent with the rest of the method.

JBYoshi: > A clickaction like OpenLink does not make sense with a variable text. And a hover text usually shows a hint/description for the object it describes.

What if a plugin adds something like an imgur link (I know of some servers that do) or some other variable link?

ST-DDT: >> A clickaction like OpenLink does not make sense with a variable text. And a hover text usually shows a hint/description for the object it describes.

> What if a plugin adds something like an imgur link (I know of some servers that do) or some other variable link?

The text with the open url action is usually provided as the param and not part of the template. 
 And if its part of the template then you will usually only replace a sub text.
If you need more explanation i can write an example code, but i cannot do that right now.

JBYoshi: >>>A clickaction like OpenLink does not make sense with a variable text. And a hover text usually shows a hint/description for the object it describes.

>> What if a plugin adds something like an imgur link (I know of some servers that do) or some other variable link?

>The text with the open url action is usually provided as the param and not part of the template.
And if its part of the template then you will usually only replace a sub text.
If you need more explanation i can write an example code, but i cannot do that right now.

I get it now. Thanks for clarifying.

ST-DDT:
WHat do you thing of my last commit? https://github.com/ST-DDT/SpongeAPI/commit/bbfb986aded2820981fcf15ce32892ea1061d933

ST-DDT:
Should be ready now

RobertHerhold:
@ST-DDT 
I realize this is nearing the time to merge, and what I am about to suggest should be in a different PR, but have you considered adding a `Text#split` or `Text#substring`?

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -24,9 +24,12 @@
  */
 package org.spongepowered.api.item.inventory;
 
+import org.spongepowered.api.text.translation.Translatable;
```
simon816: Import not in the right place (needs to be a few lines down)

src/main/java/org/spongepowered/api/text/TextRepresentable.java
```diff
@@ -0,0 +1,49 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.api.text;
+
+import org.spongepowered.api.text.action.TextAction;
+
```
simon816: Unnecessary new line

src/main/java/org/spongepowered/api/text/format/TextColors.java
```diff
@@ -59,6 +59,11 @@ public Color getColor() {
         public String getId() {
             return "NONE";
         }
+
+        public String toString() {
```
simon816: Missing `@Override`

ST-DDT:
> I realize this is nearing the time to merge, and what I am about to suggest may need to be in a different PR, but have you considered adding a Text#split or Text#substring?

I haven't seen a usecase for those two methods.
In addition to that I see the problem that except for `Text.Literal` the real content and its length will be unkown.

However if you have a real usecase we could start discussing that topic here or in a new issue.

PR #780, opened 2015-07-26
Selectors in command arguments

JBYoshi:
As it stands currently, Sponge commands using `GenericArguments` cannot use selectors without going out of their way. This pull request adds the ability to use selectors in `GenericArguments.player()`. It also adds `GenericArguments.entity()` and `GenericArguments.entityOrSource()`, which are made much easier due to selectors.

Requires SpongePowered/SpongeCommon#110.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandSource.java
+++ b/src/main/java/org/spongepowered/api/command/CommandSource.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.command;
 
+import org.spongepowered.api.command.source.ProxySourceTarget;
 import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.text.channel.MessageReceiver;
 import org.spongepowered.api.text.translation.locale.Locales;
@@ -36,7 +37,7 @@ import java.util.Locale;
  * <p>Examples of potential implementations include players, the server console,
  * Rcon clients, web-based clients, command blocks, and so on.</p>
  */
-public interface CommandSource extends MessageReceiver, Subject {
+public interface CommandSource extends MessageReceiver, Subject, ProxySourceTarget {
 
     /**
      * Gets the name identifying this command source.
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -37,6 +37,8 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.CommandMessageFormatting;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.command.source.LocatedSource;
+import org.spongepowered.api.command.source.ProxySource;
+import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.util.GuavaCollectors;
@@ -49,11 +51,14 @@ import org.spongepowered.api.world.storage.WorldProperties;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.Set;
+import java.util.UUID;
 
 import javax.annotation.Nullable;
 
@@ -903,7 +908,8 @@ public final class GenericArguments {
         }
     }
 
-    private static class PlayerCommandElement extends PatternMatchingCommandElement {
+    private static class PlayerCommandElement extends SelectorCommandElement {
+        
         private final boolean returnSource;
 
         protected PlayerCommandElement(Text key, boolean returnSource) {
@@ -911,16 +917,16 @@ public final class GenericArguments {
             this.returnSource = returnSource;
         }
 
+        @SuppressWarnings("unchecked")
         @Override
         protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
-            // TODO: Make player name resolution better -- support selectors, etc
             if (!args.hasNext() && this.returnSource) {
                 return tryReturnSource(source, args);
             }
 
             Object state = args.getState();
             try {
-                return super.parseValue(source, args);
+                return Iterables.filter((Iterable<Entity>)super.parseValue(source, args), e -> e instanceof Player);
             } catch (ArgumentParseException ex) {
                 if (this.returnSource) {
                     args.setState(state);
@@ -948,6 +954,9 @@ public final class GenericArguments {
         }
 
         private Player tryReturnSource(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            if (source instanceof ProxySource && ((ProxySource) source).getTarget() instanceof Player) {
+                return (Player) ((ProxySource) source).getTarget();
+            }
             if (source instanceof Player) {
                 return ((Player) source);
             } else {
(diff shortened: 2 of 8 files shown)


src/main/java/org/spongepowered/api/util/command/args/SelectorCommandElement.java
```diff
@@ -0,0 +1,109 @@
...
+        final Optional<String> nextArg = args.nextIfPresent();
+        args.setState(state);
+        
+        if (nextArg.isPresent()) {
+            if (nextArg.get().startsWith("@")) {
+                if (nextArg.get().contains("]")) {
+                    return ImmutableList.of();
+                }
+                if (nextArg.get().contains("[")) {
+                    // TODO tab complete selector arguments?
```
kenzierocks: I may be able to add this if you don't feel like it, shouldn't be too hard.

JBYoshi: The only real problem I had was finding out which arguments are relevant to the selector, e.g. `@a` doesn't support `type`, but `@e` and `@r` do.

kenzierocks: Probably just hardcoding, also `@p` supports it too so it may be best just to complete it for all? Not sure if we really need to separate out the different types.

JBYoshi: Now that I think about it, I'd agree with not separating them, so long as plugins can't add their own selectors and arguments.

JBYoshi: Seems like they can, so I'll need to figure out how to do value completion.

ryantheleach:
The only issue I see is that the format of selectors has been hardcoded into the API with this commit.

@kenzierocks put in quite a lot of effort in order to abstract out the format of selectors, and provide API for parsing & building them.

Could this possibly be moved to the implementation? It would also allow for CommandBook variable targets to possibly be implemented as selectors making plugins who are selector aware be able to handle them.

JBYoshi:
@ryantheleach Sounds good.

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -1247,4 +1254,90 @@ public Text getUsage(CommandSource src) {
...
+                Optional<Entity> ret = world.getEntity(uuid);
+                if (ret.isPresent()) {
+                    return ret.get();
+                }
+            }
+            throw new IllegalArgumentException("Input value " + choice + " was not an entity");
+        }
+    
+        private Entity tryReturnSource(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            // TODO get entity from ProxySource?
```
JBYoshi: @zml2008 I thought this was added in my commands PR. What happened to it?

zml2008: Not all ProxySources have entities attached. You have to check if the real invoker (proxied source) is an entity.

JBYoshi: All the vanilla `ProxySource`s have an entity.
- `/execute` is run as the entity it selects. This is different from the `CommandSource` causing the command to be executed.
- `SignSource` is run as the player clicking it.

ST-DDT: @JBYoshi
How about `/su` or `/execute console`.
There is no entity connected to that. RCON and Console both have no Entity.
However we could make the method return an Optional Entity as the entity is present most of the time.

JBYoshi: How about we add a new interface, `ProxySourceTarget`, that is extended by `Entity` and `CommandSource`? That should take care of it.

JBYoshi: Now that I think about it, `ProxySource` can't always be a `LocatedSource` if the target isn't an `Entity`. I'll be making a subinterface of `ProxySource` to fix that.

src/main/java/org/spongepowered/api/util/command/source/LocatedProxySource.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.command.source;
+
+/**
+ * Located proxy sources are {@link ProxySource}s that are run at a position in
+ * the world.
+ */
+public interface LocatedProxySource extends ProxySource, LocatedSource {
```
zml2008: or you could just check instanceof LocatedSource? might as well take advantage of multiple interface inheritance.

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
```diff
@@ -24,20 +24,25 @@
...
      * @return The proxied source
      */
     CommandSource getOriginalSource();
+
+    /**
+     * Get the target of this source.
+     *
+     * @return The target
+     */
+    ProxySourceTarget getTarget();
```
zml2008: This should be covered fine by getOriginalSource

JBYoshi: Actually, it's not. If PlayerA types `/execute PlayerB ~ ~ ~ /command`, `getOriginalSource()` returns PlayerA, while `getTarget()` returns PlayerB.

src/main/java/org/spongepowered/api/util/command/source/LocatedProxySource.java
JBYoshi: Sounds good.

Zidane:
@JBYoshi 

Can you update your PR so we can evaluate this for merge?

JBYoshi:
Sounds good. I should have time to rebase tonight.

JBYoshi:
@Zidane Done.

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -31,11 +31,13 @@
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Optional;
+import com.google.common.base.Predicates;
```
Phroa: Should migrate to Java's

src/main/java/org/spongepowered/api/util/command/args/SelectorCommandElement.java
```diff
@@ -0,0 +1,78 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.command.args;
+
+import com.google.common.base.Optional;
```
Phroa: Should use Java's

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
```diff
@@ -28,14 +28,21 @@
...
  * Proxy sources are {@link CommandSource}s that are run by one source as a
- * different source.
+ * different object.
  */
-public interface ProxySource extends LocatedSource {
+public interface ProxySource extends CommandSource {
 
     /**
-     * Get the {@link CommandSource} this source is proxying. This
+     * Get the {@link CommandSource} this source is created by.
```
Phroa: These should be `Gets` for consistency

src/main/java/org/spongepowered/api/util/command/args/SelectorCommandElement.java
JBYoshi: I haven't updated to Java 8 yet.

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
JBYoshi: I didn't change that, but I'll fix it.

Phroa: :heart:

JBYoshi:
@zml2008 Status?

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -895,16 +901,16 @@ protected PlayerCommandElement(Text key, Game game, boolean returnSource) {
...
         protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
-            // TODO: Make player name resolution better -- support selectors, etc
             if (!args.hasNext() && this.returnSource) {
                 return tryReturnSource(source, args);
             }
 
             Object state = args.getState();
             try {
-                return super.parseValue(source, args);
+                return Iterables.filter((Iterable<Entity>)super.parseValue(source, args), Predicates.instanceOf(Player.class));
```
gabizou: Couldn't you use streams or any of the new collections API?

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
```diff
@@ -28,14 +28,21 @@
 
 /**
  * Proxy sources are {@link CommandSource}s that are run by one source as a
- * different source.
+ * different object.
  */
-public interface ProxySource extends LocatedSource {
+public interface ProxySource extends CommandSource {
```
gabizou: Why the change? (Out of curiosity).

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
JBYoshi: I wrote this before the java 8 update. I'll change it later.

src/main/java/org/spongepowered/api/util/command/source/ProxySource.java
JBYoshi: It's for those cases where the target is not in a location. There was a conversation in an "outdated diff" about that.

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
JBYoshi: Changed this to lambdas, but I couldn't use streams because I might be given an `Iterable` (which doesn't support streams) as opposed to a `Collection`.

JBYoshi:
Status?

gabizou:
Merged in https://github.com/SpongePowered/SpongeAPI/commit/221b0b7b5166aaef1f56cdad2a04e01820afe044

JBYoshi:
Status?

PR #782, opened 2015-07-27
Added CommandReceiveEvent

me4502:
This event exists to fill the feature-set of Bukkit's CommandPreProcessEvent and subsequent events. The word Receive has been used instead of PreProcess, because I personally don't like words that would generally be hyphenated to be inside a class name.

The reasoning for including such an event is not, 'because bukkit did it', but instead to allow plugins such as CraftBook with the variables feature, to replace variables in the command with the values. I had many discussions with Bukkit staff back in the day about making this event 'generic' to all CommandSources, but they declined on the as they did not want to keep the event at all.

I propose that with Sponge, we include this event and allow it to filter any commands from any CommandSources.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/message/CommandReceiveEvent.java
+++ b/src/main/java/org/spongepowered/api/event/message/CommandReceiveEvent.java
@@ -1,0 +1,52 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.message;
+
+/**
+ * Fired when a command has been received by the server, but before it has been passed into any handlers.
+ */
+public interface CommandReceiveEvent extends CommandEvent {
+
+    /**
+     * Set the command string, without any prefix.
+     *
+     * <p>For example, if the command being set was {@code /example bob 3 -f}, then
+     * the parameter would be {@code example}.</p>
+     *
+     * @param command The command
+     */
+    void setCommand(String command);
+
+    /**
+     * Set the arguments string.
+     *
+     * <p>For example, if the command being set was {@code /example bob 3 -f}, then
+     * the parameter would be {@code bob 3 -f}.</p>
+     *
+     * @param arguments The arguments
+     */
+    void setArguments(String arguments);
+
+}


ryantheleach:
How would this interact with the selector API? I assume it would be thrown before the selectors are resolved, allowing modification of selectors? as opposed to after the selectors resolve which might be done per command and have custom resolution?

me4502:
The point of this event is to have access to the raw command before any processing is done, so this would be thrown before selectors. However I also see use cases for after selectors, such as modifying the command per-player in an @a command

Mumfrey:
Can I throw something at the wall and see if it sticks, this is just a bit of a brain dump so sorry if it's a bit incoherent:

Rather than adding a concept of "pre" processing chat, why not handle commands in a dispatch-loop-esque manner. What I mean is add a new concept to this kind of event, let's say an interface `Dispatchable` which works similar to `Cancellable` except that the `dispatch()` method effectively cancels the event *but then re-raises the event with new arguments*. This behavour would obviously be restricted only to handlers registered with the highest (lowest?) priority.

For example:

```java
public interface Dispatchable<E extends Event> extends Event, Cancellable {

    public abstract void dispatch(E event);

    public abstract Optional<E> previous();

}
```

Dispatching against an event effectively cancels the event, dispatching the new event (which is enforced as being the same type or a derived type by virtue of the generic parameter `E` which always refers to the concrete type of the derived event, eg.

```java
public interface CommandEvent extends GameEvent, Dispatchable<CommandEvent> {
```

When re-entering the dispatch loop, the previous event is preserved as the `previous` member inside the newly re-dispatched event, and always accessible via the `previous()` accessor, this allows consumers to walk back up the event chain in order to determine the original event arguments if necessary.

Another option which would achieve more or less the same result, without mutating the event handling style too much would be to allow **all** command event handlers to mutate the incoming text, but have the relevant `setCommand()` accessor merely push the new value into a queue internal to the event. Thus original values are again accessible (maybe via an `Iterable`), and `getCommand()` merely returns the head of the queue to consumers that just want access to the "current" value.

Anyway, just thinking aloud. It could be a stupid idea but it would be nice to be able to do pre-processing to arbitrary depth, and to make the previous values available as part of that contract. If you have any alternative ideas of how to do that then ignore my ramblings above. I just feel like the idea of "*an event that happens before an event*" isn't a good pattern, we should be looking for more general solutions to the "*I need to do something before something else*" problem, *callbacks* are a good example of a good pattern in this instance. In fact something along those general lines encoding a general-purpose event pre-processing chain for certain events may be more desirable.

`</ramble>`

ST-DDT:
I dont like the idea of the `Dispatchable` because you (pre) process the event over and over again.
In the worst case until a stackoverflow occurs.

Its probably better to modify the command string during command execution as you suggested.
However i would not preserve the entire history.
The input and the current value is just enough.
Why and when would you use the entire history?

Mumfrey:
> I dont like the idea of the Dispatchable because you (pre) process the event over and over again.
> In the worst case until a stackoverflow occurs.

I dont think you understand the concept of "dispatch loop", I have no idea how this supposed stack overflow would occur.

> Its probably better to modify the command string during command execution as you suggested.

Possibly, although it's less flexible

> However i would not preserve the entire history.

Why? You have provided no justification for this whatsoever.

> The input and the current value is just enough.

"Enough" for what?

> Why and when would you use the entire history?

Allows later dispatches to rewind previous transformations by an arbitrary number of steps.

JBYoshi:
@Mumfrey 
>> Why and when would you use the entire history?
>
> Allows later dispatches to rewind previous transformations by any number of steps.

Why would plugins want to do this?

ST-DDT:
> When re-entering the dispatch loop, the previous event is preserved as the previous member inside the newly re-dispatched event

As i understand this you post a new event to the handlers which performs all listeners again. Or do you just continue processing the event handler chain with the new event?
Changing the events in the middle of processing seems weird.

>> However i would not preserve the entire history.

> Why? You have provided no justification for this whatsoever.


>> Why and when would you use the entire history?

> Allows later dispatches to rewind previous transformations by an arbitrary number of steps.
Well because if I modify the command that has been send by the client and i would like to replace some templates or variables I want them to be replaced and no other plugin should mess around with the command because there have been something in the message before that is no longer there. The previous handler removed that for a reason and we shouldn't act like we know it better. 
If we know it better, then we should either change that handler in question or ask the plugin owner to do so or flag the plugin as conflicting.
But correcting a plugin's actions outside of its code just wrong.

Zidane:
I actually like @Mumfrey 's idea because it solves a lot more than just this issue.

In Bukkit/Forge there are patterns of Pre/Post events. They are defined as:

Pre - Usually what most know as your typical event. Can be sometimes cancelled.
Post - Usually called if its Pre brother isn't cancelled but sometimes known as a typical event...(if necessary to call AFTER game logic occurs).

Its all very subjective; in a perfect world, most events will have the ability for code to run before event creation and then post firing to all handlers. We've got 50% of the problem now, IE callbacks which allow you to get a read-only version of the event and do whatever you want. Really good use for things like GUI updates. Now the issue is pre-events and what to do there. I agree with Mumfrey that adding another event isn't the answer.

I think the dispatchable idea is a step in the right direction.

Mumfrey:
> I think the dispatchable idea is a step in the right direction.

It's just an example, I really wanted to open up a discussion on a better way to handle this idea of "pre" processing an event. I think adding a "pre-x" event is fundamentally not the way to achieve this though, and we should have a more general way of encoding this kind of concept which can then just be applied to events as appropriate.

me4502:
I do quite like the dispatch system, however in this case - the event would still be needed. Currently there isn't a way to obtain an entered command at this stage.

zml2008:
The CommandEvent is already executed at the same time this event would be -- after the arguments are split into base command and arguments but before any lookup has occurred. I don't think this makes sense as a separate event -- maybe as an addition to the current CommandEvent.

Additionally, implementing these changes in SimpleCommandService#process is fairly trivial, so it would be nice to see that in any PR making changes.

Zidane:
This needs to be updated for the new events framework.

PR #785, opened 2015-07-30
Request for SprintAPI

mosemister:
This is my first pull request so please excuse any issues I have done.

I have been working with Sponge for a while but originally I came from Bukkit. Bukkit has these features embedded into their API however SpongeAPI currently does not. I am currently checking if the player is sprinting by checking the distance between the new location and old location of a PlayerMoveEvent. This is fine until players simulate the speed of sprinting (impact of exploitation).

I believe these events and this data request will suit the need for a SprintAPI in Sponge.


src/main/java/org/spongepowered/api/event/entity/living/human/HumanToggleSprintEvent.java
```diff
@@ -0,0 +1,36 @@
...
+package org.spongepowered.api.event.entity.living.human;
+
+import org.spongepowered.api.entity.living.Human;
+import org.spongepowered.api.event.Cancellable;
+
+/**
+ * Called when a {@link Human} changes to/from sprint.
+ */
+public interface HumanToggleSprintEvent extends HumanEvent, Cancellable {
+	boolean isNowSprinting();
```
ST-DDT: Missing javadocs and empty lines

ST-DDT:
There is a huge rewrite of the data API here https://github.com/SpongePowered/SpongeAPI/pull/685
So you might want to integrate that there. But this PR is a good starting point for a discussion.

gabizou:
@mosemister If you're fine with it, I can easily add this to #685.

mosemister:
@gabizou Go for it

PR #787, opened 2015-08-02
Better support for Chunk grabbing

mosemister:
currently the only ways of getting a Chunk in Sponge is to either know the coords of a chunk (X, Z) and have the World already there. 
World.getChunk(X, 0, Z);
World.loadChunk(X, 0, Z, gen);

or hope that a location has a Chunk as its Extent instead of a World. But currently 100% of the times i have tried, it has always been a World. 

So what im purposing is to add more methods of getting Chunk. And allow plugin developers to create their own algorithms to get the chunk they need. 

World.getChunks();

This would grab all chunks in that world. This would mean plugin developers could create their own algorithms to get the chunk they want.

World.getChunk(Location);

This grabs the chunk that has that block. In this case the Location Extent does not matter, its the World that matters.

Entity.getChunk();

Sponge already has Entity.getWorld() so there are 2 ways of getting a world from a entity ((World)Entity.getLocation().getExtent()) so it makes sense to have the ability for Entity to have getChunk(). 

This would fix the following #540 #545


DDoS:
First off, you have to remove the commits related to the sprint API that aren't even in the changes. You squash all these commits into one. Also that commit name is way too vague.

`getChunks()` should return an Iterable, not a Collection. A Collection suggest that chunks can be added, removed and manipulated as a group. In reality they're can't. An Iterator only has the modification operation `remove()`, but it is optional so it can just throw UnsupportedOperationException.

`getChunk(Location)` should be `getChunkForBlock(Vector3i)`and `getChunkForBlock(int, int, int)`. The current method has poor disambiguation with `getChunk(Vector3i)` although they use completely different coordinate systems.

Finally, I need a corresponding PR implementation for this.

kenzierocks:
You don't need to choose the PR, just use `git push --force` and `git reset`

mosemister:
somethings happened, IDK how but it nows says Unknown repository

PR #793, opened 2015-08-07
Add API for generating error reports.

zml2008:
This service makes it easier for plugins to generate error reports that provide useful information for plugin developers.

[Example error report](https://gist.github.com/anonymous/ae434617c550c176f219)

The implementation wraps Minecraft's CrashReport system, with the changes of:
- Easily allowing sending the error report to a specific player
- Outputting the error report in Markdown format
- Posting the error report to an anonymous gist
- Using Text to allow formatting certain sections

Things to think about:
- [ ] How to pass through fatal exceptions? I currently use MC's ReportedException in the impl, but the SimpleCommandService wraps these in another layer of error reporting. Ways to deal with this are exposing the ReportedException to the API somehow, allowing custom error handlers to pass them through, moving the SimpleCommandService into the implementation and capturing there, or making the ErrorReportService simply reuse the existing error report while allowing the command service to tack on its additional data
- [ ] Should each Reportable be provided the entire error report, or its own section? 
- [ ] Additional pastebin services, authentication support? (would be configurable)

### DIFF
--- a/src/main/java/org/spongepowered/api/event/service/error/ErrorReportEvent.java
+++ b/src/main/java/org/spongepowered/api/event/service/error/ErrorReportEvent.java
@@ -1,0 +1,44 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.service.error;
+
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.service.error.ErrorReport;
+
+/**
+ * This event is thrown when an error report is generated, after the source of the error report has dispatched it but before it is sent to any
+ * dispatch targets.
+ * When this event is dispatched, the game may be in an inconsistent state. This means that some things that can generally be assumed to be
+ * available *MAY NOT BE*. Keep this in mind, and practice defensive programming to the extreme.
+ */
+public interface ErrorReportEvent extends Event {
+
+    /**
+     * Get the error report being dispatched.
+     *
+     * @return the error report
+     */
+    ErrorReport getErrorReport();
+}
--- a/src/main/java/org/spongepowered/api/service/error/ErrorReport.java
+++ b/src/main/java/org/spongepowered/api/service/error/ErrorReport.java
@@ -1,0 +1,164 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.error;
+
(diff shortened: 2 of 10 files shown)


kashike:
Is the `sponge.debug.hover-stacktrace` permission meant to be changed? `sponge.debug.error-report`

zml2008:
yeah, that would probably make sense to do

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Get the text of this error report.
+     *
+     * @return This error report's text
+     */
+    String toText();
+
+    /**
+     * Takes the text content of this error report and posts it to a pastebin service (implementation-determined).
+     * This allows users to more easily send it to the appropriate target.
```
progwml6: should probably be paste upload service or something like that not a name of a specific companies option

src/main/java/org/spongepowered/api/event/service/error/ErrorReportEvent.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * This event is thrown when an error report is generated, after the source of the error report has dispatched it but before it is sent to any
+ * dispatch targets.
+ * When this event is dispatched, the game may be in an inconsistent state. This means that some things that can generally be assumed to be
+ * available *MAY NOT BE*. Keep this in mind, and practice defensive programming to the extreme.
+ */
+public interface ErrorReportEvent extends Event {
+
+    /**
+     * Get the error report being dispatched.
```
Phroa: The style is `Gets`

src/main/java/org/spongepowered/api/event/service/error/ErrorReportEvent.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return the error report
```
Phroa: Capital `t` plz

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.error;
+
+import com.google.common.base.Optional;
```
Phroa: Should be Java's

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.net.URL;
+
+/**
+ * An error report. This object is mutable.
+ */
+public interface ErrorReport {
+
+    /**
+     * Get the {@link Throwable} that was responsible for creating this error report, if any is present.
```
Phroa: Should be `Gets`

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Set whether or not this is user error.
+     *
+     * @param isUserError Whether or not this is a user error
+     * @return this
+     */
+    ErrorReport setUserError(boolean isUserError);
+
+    /**
+     * Get a description of the base cause of this error.
```
Phroa: and here

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @return The description of this error's base cause
+     */
+    String getCauseDescription();
+
+    /**
+     * Get a section with the given title, creating and appending to the end of the error report as necessary.
```
Phroa: and here

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param reportable The object to append information from
+     * @return this
+     */
+    ErrorReport addReportable(Reportable reportable);
+
+    /**
+     * Dispatch this error as a fatal error. Fatal means that the server will be stopped and a full crash report will be generated.
+     *
+     * An exception will be used to pass the result of this error up the call stack. This exception must be passed on.
```
Phroa: This should probably have `<p>` tags wrapping it

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    void dispatchToCommandSource(CommandSource source);
+
+    /**
+     * Dispatch this as a non-fatal error that is sent to the console.
+     */
+    void dispatchToConsole();
+
+    /**
+     * Get the text of this error report.
```
Phroa: gets

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @return This error report's text
+     */
+    String toText();
+
+    /**
+     * Takes the text content of this error report and posts it to a pastebin service (implementation-determined).
+     * This allows users to more easily send it to the appropriate target.
+     *
+     * @return a url containing the content of this error report
```
Phroa: capital `a` plz

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+        /**
+         * Add an entry to this error report with the given information.
+         *
+         * @param content The content to add to this error report
+         * @return this
+         */
+        Section addEntry(Text content);
+
+        /**
+         * The entry to add.
```
Phroa: Huh?

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+     * This allows users to more easily send it to the appropriate target.
+     *
+     * @return a url containing the content of this error report
+     */
+    ListenableFuture<URL> toPastebin();
+
+    interface Section {
+
+        /**
+         * Add an entry to this error report with the given information.
```
Phroa: I feel like these should be `Adds` in keeping with style, but I'm actually not sure...

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,164 @@
...
+         * @param key The name of the segment
+         * @param value The block of text to include, with lines separated by \n
+         * @param language The language name to use for syntax highlighting in services where supported
+         * @return this
+         */
+        Section addCodeEntry(Text key, String value, String language);
+
+        /**
+         * Set the exception to be contained within this section.
+         * @param t The exception to be used
```
Phroa: Blank line above

gabizou:
@zml2008 how soon can this be finished/merged?

20zinnm:
Would it make sense to allow people to pass a WebService object or something that has `FutureCallable<URL> getUrl(ErrorReport report)` in order to allow individual people to implement custom paste services? Just abstract the Paste Service part.

phroa:
Along with @20zinnm's stuff, what if there was a config value to enable plugins to dispatch exceptions from that plugin to an author-run webhook?

Square has this super nifty program http://squash.io/ which absorbs stacktraces from (mainly) Android apps and Rails installations and ties them back to the lines of code and who wrote them (through obfuscation mappings even).
An API to ship stacktraces back to developers (or to a developer portal on Ore) easily could be quite neat; especially if it means there aren't twenty homebrewed data-leaking implementations that server owners have no means to disable.

20zinnm:
Oh yeah, @phroa brings up a good point. No plugin should be able to send data to a third party, regardless of intention (pastebin or otherwise), without the server operator's consent.

phroa:
Impossible to prevent sanely, but really scummy imo.

If there was a centralized, developer-and-user-friendly API for server data transfer, developers wouldn't need to make their own and users could centrally disable the functions. Straying a bit from @zml2008's PR here :P

zml2008:
Closing in favor of #1215

Issue #788, opened 2015-08-04
Generify Location's Extent


DDoS:
Location should have type `Location<T extends Extent>`. This would allow `T getExtent()` to return a concrete type, such as World. This would fix the common case of `(World) object.getLocation().getExtent()`. Currently it returns Extent which isn't all that useful.

zml2008:
This seems pretty useful -- probably wouldn't have to increase verbosity much in most places too.

DDoS:
I think the verbosity saved from casts and `instanceof` checks would compensate for the slightly longer type name.

DDoS:
[Here](https://github.com/SpongePowered/SpongeAPI/commit/02473cb31fac)'s what the API looks like with generic locations.

@SpongePowered/developers Yes or no? Should we just replace Extent by World in Location, since that seems to be the most used type by far?

kashike:
This seems pretty helpful, I think it is worth the verbosity?

zml2008:
yeah, I think generifying Location is probably the best way to go. Would be great for stuff to use Extent wherever possible too, rather than World, so that we can do things like what WE does with treegen on bukkit -- sticking an Extent in between that allows capturing changes occurring.

DDoS:
Well currently Extent is way too big of an interface to actually be useful, but that's another issue.

DDoS:
Closing this, follow in the PR.

PR #799, opened 2015-08-09
Generify Location on the type of extent

DDoS:
See issue #788.

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -503,7 +503,7 @@ public interface GameRegistry {
      * @return A new invalid identity transform
      * @see #createTransform(Extent)
      */
-    Transform createTransform();
+    <E extends Extent> Transform<E> createTransform();
 
     /**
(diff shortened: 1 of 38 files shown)


DDoS:
These changes are very sensitive to conflicts, so can I get a fast review and approval/rejection?

zml2008:
I agree with this change

Zidane:
Me as well.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -503,7 +503,7 @@
      * @return A new invalid identity transform
      * @see #createTransform(Extent)
      */
-    Transform createTransform();
+    <E extends Extent> Transform<E> createTransform();
```
gratimax: Commenting a bit too late, but this could just be `Transform<? extends Extent>`

gabizou: We'd want the Transform to be generified, so you could do `gameRegistry.<World>createTransform()`

PR #815, opened 2015-08-18
gamemode() method in GameModeData

jamierocks:


gabizou:
No need since GameModeData extends VariantData.

jamierocks:
Would the same, therefor apply to CareerData?
https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/CareerData.java

gabizou:
Yes.

PR #829, opened 2015-09-01
[WIP] Add ScheduledExecutorService

Kiskae:
@zml2008 has asked if I was willing to contribute my implementation of the standard ExecutorService (Original post: [Sponge forums](https://forums.spongepowered.org/t/executorservice-backed-by-sponges-schedulerservice/8352)) to the project, this implementation delegates the execution of its tasks to the existing SchedulerService.

While most of the implementation is based on the original version ([https://gist.github.com/Kiskae/5c44ea146857ca173170](https://gist.github.com/Kiskae/5c44ea146857ca173170)), I went through the contract of the ScheduledExecutorService again and changed the implementation to be more compliant:
* The `schedule(AtFixedRate|WithFixedDelay)` methods would not execute more than once due to the implementation of FutureTask preventing multiple executions, the private class `RepeatableFutureTask` solves this issue
* The contract of the  `schedule(AtFixedRate|WithFixedDelay)` methods specify that they should stop executing when they encounter an exception. Sponge tasks do not stop when they encounter an exception so the `RepeatableFutureTask` will cancel its parent task if an exception occurs.
  * ~~*TODO* the SchedulerService implementation prints to the log when an exception occurs, should this behaviour be replicated in the `RepeatableFutureTask`. Otherwise it can fail completely silently.~~
* `ScheduledFuture.getDelay(TimeUnit)` specifies that it must return the amount of time UNTIL the next execution, the previous implementation simply returned Sponge's `Task.getDelay()`, which is the original execution delay. The implementation now tries to approximate the delay until the next execution.
  * ~~*TODO* Decide whether the following should be implemented: This could be made more accurate (and less hacky) by moving a part of the SpongeFuture implementation to SpongeCommon, where the `timestamp` value is available.~~

Other changes made to better integrate with Sponge:
* Added `syncExecutor(Object)` and `asyncExecutor(Object)` to the `SchedulerService`, they take a plugin instance and return the ExecutorService
* Removed the ability to customize the task names, given that it is an abstraction over the SchedulerService, the default naming based on creation order makes the most sense anyway.

#### What will this PR allow for?
Through this pull request it is possible to use many of the concurrency frameworks that are available on the JVM, since many use the ExecutorService as basis for the execution of whatever they do.
Known frameworks that support execution through an arbitrary executorservice are:
* [RxJava](http://reactivex.io/RxJava/javadoc/rx/schedulers/Schedulers.html): A custom Scheduler can be used - Schedulers.from(game.getScheduler().syncExecutor(this))
* [Java 8 CompletableFuture]( http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html): All the *Async methods have an overload that accepts an Executor to run on.
* [Scala Futures](http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.concurrent.Future): All futures take an implicit ExecutionContext - ExecutionContext.fromExecutorService(game.getScheduler().syncExecutor(this))

Example of usage:
```java
final Server s = game.getServer();
ExecutorService executor = game.getScheduler().syncExecutor(this);
ObserableFuture<UUID> futurePlayerId = someSource();
futurePlayerId.thenApplyAsync((playerUUID) -> {
    //Any actions performed here are thread-safe!
    final Player p = server.getPlayer(playerUUID).get();
    //Modify the world at the players location
    return player;
}, executor).thenApply((player) -> {
    //Beware, this also runs on the server thread, use another *Async call to move to another thread
      player.kick();
    return player.getName();
}).thenApplyAsync((playerName) -> {
    //This is no longer thread-safe, but will not risk blocking the main thread
    computeNameCoolnessVerySlow(playerName);
});
```

### DIFF
--- a/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
+++ b/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
@@ -90,7 +90,7 @@ public interface SchedulerService {
      * @return A new executor service that can be used to execute
      *         synchronous tasks
      */
-    SpongeExecutorService syncExecutor(Object plugin);
+    SpongeExecutorService createSyncExecutor(Object plugin);
 
     /**
      * Creates a new ExecutorService that can be used to schedule asynchronous
@@ -99,6 +99,7 @@ public interface SchedulerService {
      * @param plugin The plugin that will own the created tasks
      * @return A new executor service that can be used to execute
      *         asynchronous tasks
+     * @see TaskBuilder#async()
      */
-    SpongeExecutorService asyncExecutor(Object plugin);
+    SpongeExecutorService createAsyncExecutor(Object plugin);
 }
--- a/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
+++ b/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
@@ -81,4 +81,28 @@ public interface SchedulerService {
      * @return A set of scheduled tasks
      */
     Set<Task> getScheduledTasks(Object plugin);
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule synchronous
+     * tasks through the standard Java concurrency interfaces.
+     *
+     * @param plugin The plugin that will own the created tasks
+     * @return A new executor service that can be used to execute
+     *         synchronous tasks
+     */
+    default SpongeExecutorService syncExecutor(Object plugin) {
+        return new SpongeExecutorService(() -> createTaskBuilder(), plugin);
+    }
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule asynchronous
+     * tasks through the standard Java concurrency interfaces.
+     *
+     * @param plugin The plugin that will own the created tasks
+     * @return A new executor service that can be used to execute
+     *         asynchronous tasks
+     */
+    default SpongeExecutorService asyncExecutor(Object plugin) {
+        return new SpongeExecutorService(() -> createTaskBuilder().async(), plugin);
+    }
 }
--- a/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
+++ b/src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
@@ -90,9 +90,7 @@ public interface SchedulerService {
      * @return A new executor service that can be used to execute
      *         synchronous tasks
      */
-    default SpongeExecutorService syncExecutor(Object plugin) {
-        return new SpongeExecutorService(() -> createTaskBuilder(), plugin);
-    }
+    SpongeExecutorService syncExecutor(Object plugin);
 
     /**
      * Creates a new ExecutorService that can be used to schedule asynchronous
@@ -102,7 +100,5 @@ public interface SchedulerService {
      * @return A new executor service that can be used to execute
      *         asynchronous tasks
      */
-    default SpongeExecutorService asyncExecutor(Object plugin) {
-        return new SpongeExecutorService(() -> createTaskBuilder().async(), plugin);
-    }
+    SpongeExecutorService asyncExecutor(Object plugin);
 }
--- a/src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
+++ b/src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
@@ -40,13 +40,17 @@ import java.util.concurrent.TimeUnit;
  */
 public interface SpongeExecutorService extends ScheduledExecutorService {
 
-    @Override SpongeFuture<?> schedule(Runnable command, long delay, TimeUnit unit);
+    @Override
+    SpongeFuture<?> schedule(Runnable command, long delay, TimeUnit unit);
 
-    @Override <V> SpongeFuture<V> schedule(Callable<V> callable, long delay, TimeUnit unit);
+    @Override
+    <V> SpongeFuture<V> schedule(Callable<V> callable, long delay, TimeUnit unit);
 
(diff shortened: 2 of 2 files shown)


Kiskae:
Regarding the current implementation of `ScheduledFuture.getDelay(TimeUnit)`, the approximation works but it will slowly get out of sync with the actual delay for repeating tasks due to the actual delay between scheduled time and execution time. However given that getDelay depends on the current time to calculate the delay I doubt any large libraries rely on the accuracy of this value for ordering.

If we wanted an accurate implementation then it would be possible to move the `SpongeFuture` implementation to SpongeCommon and make the ExecutorService abstract, so the creation of the SpongeFutures can be delegated to a subclass that knows the correct time.

zml2008:
Thanks a ton for PRing this! 

A few things:
- I think it would make sense to also expose this executor through Guice as you mentioned in your PM -- an ExecutorService per plugin -- so binding ExecutorServices to the appropriate implementaion.
  Because this is an implementation item, I'll hook it up when merging the PR so you don't have to deal with another branch. However, to separate sync vs async services, it would probably make sense to add an @Async annotation that would be used to mark the async scheduler (and sync is unannotated or @Sync -- I'm leaning towards unannotated but thoughts?)
- Would implementing shutdown()/awaitTermination() by cancelling all tasks started through the service make sense? This might be useful in some cases --if a plugin wants to manage tasks differently or something.

- For the delay calculation, it's probably fine to move that to common -- really you could move the whole SpongeExecutorService to common and just return ScheduledExecutorService from the API methods.

Kiskae:
@zml2008 

* Injection would be trivially implemented as a provider:
```java
@Provides @Sync
SpongeExecutorService provideSyncExecutor(SchedulerService ss, PluginContainer plugin) {
    return ss.syncExecutor(plugin.getInstance())
}
```
* I don't believe it makes sense when you combine it with the Guice injection support, since the contract specifies that no more tasks can be scheduled after `shutdown()` has been called. There is also the issue of identifying the tasks that were scheduled through the executor, which would require either maintaining a list and using `ListenableFuture`s to clear out finished futures.
* With the delay change, I presume the preferrable solution would be to reduce the class on the API side to just the interfaces, and move all of the implementation to SpongeCommon?

As a small side-note, while figuring out the actual contract of getDelay by looking at the reference implementation, I noted that all of java.concurrent uses `System.nanoTime` rather than `System.currentTimeMillis`, since nanoTime is guaranteed to be monotonic. It might be a good idea to replace currentTimeMillis with nanoTime.

Kiskae:
Implementation pushed to SpongePowered/SpongeCommon#151

zml2008:
ok great. I see your point about shutdown() not making sense to implement -- would probably be good to stick a comment about that in the implementation for clarity, since it's not necessarily immediately obvious.

Using System.nanoTime instead is probably a good idea too -- since currentTimeMillis may go backwards or skip with leap seconds or system time corrections.

Kiskae:
@zml2008 I've implemented those two points on the SpongeCommon PR.

We only need to figure out whether the executor should log exceptions like SchedulerBase does, since it would otherwise silently cancel a repeating task?

src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
```diff
@@ -81,4 +81,24 @@
...
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule synchronous
+     * tasks through the standard Java concurrency interfaces.
+     *
+     * @param plugin The plugin that will own the created tasks
+     * @return A new executor service that can be used to execute
+     *         synchronous tasks
+     */
+    SpongeExecutorService syncExecutor(Object plugin);
```
simon816: should be `createSyncExecutor`

src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
```diff
@@ -81,4 +81,24 @@
...
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule asynchronous
+     * tasks through the standard Java concurrency interfaces.
+     *
+     * @param plugin The plugin that will own the created tasks
+     * @return A new executor service that can be used to execute
+     *         asynchronous tasks
+     */
+    SpongeExecutorService asyncExecutor(Object plugin);
```
simon816: should be `createAsyncExecutor`

zml2008:
imo the best to do here it so match the documented behavior of ScheduledExecutorService, setting the thrown exception as the result of the future. Since the future is run through the executor service, I'd expect whatever is scheduling the future to be the one handling the exception, so not sure if it really makes sense to pass the exception through to the scheduler, instead just cancel the task.

Kiskae:
After considering it I agree, the exceptions are captured by the Future pattern and the `Executor.submit` method delegates exceptions to the underlying thread, which is handled by Sponge's Scheduler.

@zml2008 The current PRs should be functionally complete, the only thing that has been discussed and is still missing are the guice providers.

zml2008:
awesome! I'll let this sit for another day or two just in case anyone has other comments

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,59 @@
...
+ * Sponge's {@link SchedulerService}.
+ * <p>
+ *     This class can be used to allow any libraries that support the
+ *     standard concurrency interface to schedule their asynchronous
+ *     tasks through Sponge.
+ * </p>
+ */
+public interface SpongeExecutorService extends ScheduledExecutorService {
+
+    @Override SpongeFuture<?> schedule(Runnable command, long delay, TimeUnit unit);
```
simon816: The `@Override` annotations should be on a separate line

gratimax:
Forgot to close this, lol. Nice work @Kiskae

PR #834, opened 2015-09-04
Usage improvements

TheRaspPie:
Api PR | <a href="https://github.com/SpongePowered/SpongeCommon/pull/152">Implementation PR</a>

This pull request should improve the overall experience from the view of a plugin developer.
It should be merged before API Revision 3.0.

In summary, this pull request contains the tiny improvements that I think are needed to get into API Revision 3.0.

Please do not close this PR until Revision 3.0 as I may commit a few other changes in the meantime.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -430,6 +430,11 @@ public interface Entity extends Identifiable, DataHolder, DataSerializable {
      */
     boolean damage(double damage, Cause cause);
 
+    /**
+     * Creates a {@link EntitySnapshot} containing the {@link EntityType} and data of this entity.
+     *
+     * @return The snapshot
+     */
     EntitySnapshot createSnapshot();
 
 }
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -59,9 +59,9 @@ import java.util.UUID;
 public interface Entity extends Identifiable, DataHolder, DataSerializable {
 
     /**
-     * Get the type of entity.
+     * Gets the type of this entity.
      *
-     * @return The type of entity
+     * @return The type
      */
     EntityType getType();
 
@@ -72,14 +72,17 @@ public interface Entity extends Identifiable, DataHolder, DataSerializable {
      */
     World getWorld();
 
+<<<<<<< HEAD
     /**
      * Creates a {@link EntitySnapshot} containing the {@link EntityType} and data of this entity.
      * @return The snapshot
      */
     EntitySnapshot createSnapshot();
 
+=======
+>>>>>>> Added convenience methods to Entity for getting and setting positions and rotations.
     /**
-     * Get the location of this entity.
+     * Gets the location of this entity.
      *
      * @return The location
      */
@@ -103,7 +106,31 @@ public interface Entity extends Identifiable, DataHolder, DataSerializable {
     boolean setLocationSafely(Location<World> location);
 
     /**
-     * Gets the rotation.
+     * Gets the position of this entity in its current world.
+     *
+     * @return The position
+     */
+    Vector3d getPosition();
+
+    /**
+     * Sets the position of this entity in its current world. This is equivalent to a teleport,
+     * and also moves this entity's passengers.
+     *
+     * @param position The position to set
+     */
+    void setPosition(Vector3d position);
+
+    /**
+     * Sets the position of this entity in its current world using a safe one from {@link TeleportHelper#getSafeLocation(Location)}.
+     * This is equivalent to a teleport and also moves this entity's passengers.
+     *
+     * @param position The position to set
+     * @return True if position was set successfully, false if position couldn't be set as no safe position was found
+     */
+    boolean setPositionSafely(Vector3d position);
+
+    /**
+     * Gets the rotation of this entity.
      *
      * <p>The format of the rotation is represented by:</p>
      *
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -298,4 +421,6 @@
      */
     boolean damage(double damage, Cause cause);
 
+    EntitySnapshot createSnapshot();
```
bloodmc: javadocs?

Zidane: Already in master @bloodmc

TheRaspPie: Yes, but I didn't merge it right the first time.

TheRaspPie:
http://stackoverflow.com/questions/8622497/why-can-constructors-within-an-enum-in-java-be-declared-only-private

simon816:
@TheRaspPie could you update this on the latest `master`?

src/main/java/org/spongepowered/api/util/Axis.java
```diff
@@ -37,7 +37,7 @@
 
     private final Vector3d direction;
 
-    private Axis(final Vector3d vector3d) {
+    Axis(final Vector3d vector3d) {
```
JBYoshi: Why this change?

src/main/java/org/spongepowered/api/util/AxisDirection.java
```diff
@@ -36,7 +36,7 @@
 
     private final int signum;
 
-    private AxisDirection(final int signum) {
+    AxisDirection(final int signum) {
```
JBYoshi: ^

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -260,6 +341,48 @@
...
+     *
+     * <p>If the target world is unloaded and not enabled then the transfer
+     * will fail.</p>
+     *
+     * @param uuid The UUID of the target world to transfer to
+     * @param position The position in the target world
+     * @param rotation The rotation
+     * @return True if the teleport was successful
+     */
+    boolean transferToWorld(UUID uuid, Vector3d position, Vector3d rotation);
```
Deamon5550: This should be a WorldProperties object rather than a UUID to reference the unloaded world.

TheRaspPie: `UUID` is actually still used in the already existing `transferToWorld` methods.

Deamon5550: Well that is another issue then, we should be referring to worlds which may be loaded or unloaded by their `WorldProperties`. This ensures that we are actually referring to a valid world which makes the handling simpler.

TheRaspPie: Ok, I can add this, but I'm not even sure if this PR will be merged. Maybe I could get some input from other developers if I should keep working on this.
@Zidane

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -122,12 +140,74 @@
...
+     *
+     * <p>The format of the rotation is represented by:</p>
+     *
+     * <ul><code>x -> pitch</code>, <code>y -> yaw</code>, <code>z -> roll
+     * </code></ul>
+     *
+     * @param position The position to set
+     * @param rotation The rotation to set
+     */
+    void setPositionAndRotation(Vector3d position, Vector3d rotation);
```
Deamon5550: I wonder if instead of making these combined methods we could just make the setters fluent.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -73,13 +73,7 @@
     World getWorld();
 
     /**
-     * Creates a {@link EntitySnapshot} containing the {@link EntityType} and data of this entity.
```
gabizou: Why are you moving this around?

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -73,13 +73,7 @@
...
-     * @return The snapshot
-     */
-    EntitySnapshot createSnapshot();
-
-    /**
-     * Get the location of this entity.
+     * Gets the location of this entity.
```
gabizou: OCD fix.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -169,9 +250,9 @@
     void setLocationAndRotation(Location<World> location, Vector3d rotation, EnumSet<RelativePositions> relativePositions);
 
     /**
-     * Sets the location using a safe one from
-     * {@link TeleportHelper#getSafeLocation(Location)} and the rotation of this
-     * entity. {@link RelativePositions} listed inside the EnumSet are
+     * Sets the location of this entity using a safe one from
```
gabizou: OCD.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -281,7 +404,7 @@
     boolean isLoaded();
 
     /**
-     * Mark this entity for removal in the very near future, preferably
+     * Marks this entity for removal in the very near future, preferably
```
gabizou: Again, OCD.

src/main/java/org/spongepowered/api/util/Axis.java
gabizou: Keep these sort of OCD fixes out of PR's, make sure they're noted in the OCD #221 issue.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -44,41 +42,20 @@
...
      *
-     * @param world The world
-     */
-    void setWorld(World world);
-
-    /**
-     * Gets the source {@link Explosive} of the explosion.
-     *
-     * @return The source explosive or null if there is no source
+     * @return The source explosive, if present
```
gabizou: If this is a usability, you should note that you're making `Explosion` immutable.

src/main/java/org/spongepowered/api/world/explosion/ExplosionBuilder.java
```diff
@@ -39,15 +39,15 @@
      * Sets the {@link World} the explosion will occur in.
      *
      * @param world The world
-     * @return The builder, for chaining
+     * @return This builder, for chaining
```
gabizou: again, ocd.

TheRaspPie:
Alright, thank you guys for your feedback! I will update the PR until the next week or so, depending on whether I'll find time while learning for course tests.

gabizou:
@TheRaspPie As a general rule of thumb, if you want to make or suggest minor changes, make it a suggestion in the OCD issue; however, if you want to change functionality, make that functionality a PR. As it is, don't join various OCD fixes with functionality changes in a singular PR. You want to have multiple PR's for multiple functionality changes (unless it's an overarching change such as Data API).

TheRaspPie:
@gabizou I got that point with the OCD fixes from your many comments. I just thought if I'm already editing this file, I can fix the issues right now.

I find it very annoying to have to create a pull request for every little functionality change that I propose.
I think the way to go there (if it's only a small change) is through an issue. I would simply let staff with write permission make the commit.

gabizou:
@TheRaspPie I can see this being merge worthy if you undo the OCD fixes and re-arranging of various methods that have already been pointed out. Otherwise, it's a waste of not only your time, but our time to review a PR like this. Calling it annoying because you find it inconvenient to make individual PR's for each functionality change is the same as the team saying it's annoying that you're not making it simple for us to merge functionality changes.

On the issue of OCD fixes, it's a bunch of noise. Going through the git history and finding several hundred OCD fixes from hundreds of commits is a lot of noise. This is the primary reason we don't accept OCD "fixes" in any PR unless it's time to do so (which incidentally, we're prepping to do so before the end of the month).

Own the issue of functionality changes, it's equally noisy to discuss changes about two different parts of the API when one is ready for merging and the other is still being discussed. Further holding back such changes is a waste of time if they were simply two different PR's.

Lastly, requesting that this isn't merged until 3.0 release is just silly. Making more changes down the road prevents the existing possibly acceptable changes being merged, and further requires more discussion.

If there's any flaw in my reasoning, please let me know. Otherwise, I see no reason to argue semantics of how we handle PR's.

gabizou:
Yeah, there's a lot of different things going on in this PR that I'd rather discuss in separate PR's. Having just spent some time debating with the position/location/rotation methods and how they **should** be handled is just going to block any of the other changes from taking place.

TheRaspPie:
I think this PR was just intermingled with the things I wanted to change. One reason for this was that I was just a beginner Git user and it was tiresome for me to create multiple PRs.

PR #808, opened 2015-08-15
Improve Text API

ST-DDT:
This PR improves the changes made in #779.
Implementation: https://github.com/SpongePowered/SpongeCommon/pull/129

## Additions

* `Placeholder`s
  * Placeholders can be used to calculate their replacement using the given context
* `Text.format(Text, Function<String,?>)`
  * Allows lazy fetching/calulcation of replacements used in conjunction with `Placeholder`s
    * Example: Directly fetch player prefixes from a permission plugin without being bound to just `prefix` or similar. `{world}{faction}{foobar}{prefix}{player}: {message}`
  * Also avaiable as `Text.format(Text, Object...)` uses indexes as keys for the map
  * Also avaiable as `Text.format(Text, Map<String,?>)` uses the function impl
* Similarize the behaviour of all `joinX(...)`methods
* `Text#joinOf(Text separator, Object... objects)`
* `Text#joinOf(Text separator, Iterable<?> contents)`
* `Text#formatBuilder()`
  * Utility method to use the format call with named parameters in one row and without having to explicitly create a `Map` for it.
* `Functions` (Optional)
  * Utility class for creating and wrapping `Function`s
* `BiFunctions` (Optional)
  * Utility class for creating and wrapping `BiFunction`s
* `Dictionary` (Optional)
  *  Utility class for multi language support with fully configurable messages.

## Changes

* Better `TextRepresentable` specifications
* `Text.of(Object...)` 
  * added support for `Iterable`s
  * added support for arrays
  * Simplified handling of `TextRepresentable`s

## Test Plugin

[TextFormat.jar](https://dl.dropboxusercontent.com/u/16999313/Sponge/TextFormat.jar) (Plugin)
[TextFormat.zip](https://dl.dropboxusercontent.com/u/16999313/Sponge/TextFormat.zip) (Source)

Try the following commands to see the resutls:
* `/textformat`
* `/textformat PvPMonster`
* `/textformat PvPMonster Innocent`
* `/textformat PvPMonster Innocent &6Stick of Doom`
* `/textformat <YourUsername> Innocent` // Uses the item in hand
* `/texttype <CatalogType/*>`
* `/textjson` - Proves the json (de-)serialization works
* `/textxml` - Proves the xml(de-)serialization works

## Usage

Consider the following situation. You would like to show the following message to your arena participants on a regular basis.
````
Current ranking:
1) MonsterKiller - 42 points (324m away)
2) MunsterHunter - 21 points (123m away)
[...]

PR #835, opened 2015-09-05
[WIP] Text templates

gratimax:
See https://gist.github.com/gratimax/9aa7ecc33467e0ddfd0b. More info coming soon.

Hopefully this example can showcase some of the intended power here:

```java
import static org.spongepowered.api.text.template.TextElements.*;

static final TextTemplate playerMessage = TextTemplate.of(
  TextColors.BLUE, "Hello, ", TextColors.LIGHT_GRAY,
  playerDisplayName().pos(0), TextColors.BLUE,
  "See your team members: ", 
  TextColors.LIGHT_GRAY, 
    fallback(
      iterable(playerDisplayName(), Texts.of("|")), 
      Texts.of("No players")).pos(1), 
  TextColors.BLUE, "!");

for (Player player: players) {
  player.sendMessage(playerMessage.build(
    TextArgs.of(player, players)));
}
```


src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -289,6 +285,12 @@ public final String getContent() {
         }
 
         @Override
+        public boolean isEmpty() {
+            return content.equals("") && children.isEmpty() && !clickAction.isPresent() && !hoverAction.isPresent() &&
```
ST-DDT: You could reduce this to 
````java
return super.isEmpty() && this.content.isEmpty();
````

**Comment by TheRaspPie** [comment]
Will this remove `Texts.format`? Because if yes, we couldn't do a per-player message with `MessageSink.transformMessage` anymore.
If no, then we have two formatting APIs.

**Comment by gratimax** [comment]
@TheRaspPie 

Yes, it will remove `Texts.format`. I don't really get what you mean about per-player messaging, here it is in action:
```java
class MyMessageSink extends MessageSink {
   private final TextTemplate template = TextTemplate.of("A message was just sent to you: ", TextElement.identity().pos(0));

   public Text transformMessage(CommandSource target, Text text) {
        return template.build(TextArgs.of(text));
    }
}
```

I don't really get how having to pass around text templates is "inflexible". If anything, it makes using the text API clean since there is a really clean separation of concerns.

Flibio:
@gratimax 

Any idea when this PR will be accepted?

gratimax:
@Flibio Sorry, the PR is still not done (there's quite a lot to do yet), so I cannot guarantee any ETA. Sorry for making you wait :frowning:. Is there a specific use case you have that could perhaps be done sooner, so that you won't wait on this PR?

Zidane:
@gratimax 

MessageEvent needs template support to do formatting without touching source names. Really need to get this sped along into master.

gratimax:
@Zidane I understand that. I suggested in the comment that the specific use case of MessageEvent's touching source names could be done sooner so that no one has to wait on this PR. My suggestion would be to use functions representing templates, which is ugly but works provisionally. I appreciate that people have a use for this, but I can't guarantee any time that this PR can be merged since there is still a lot to do, as I have said several times before.

ST-DDT:
You could use Texts.format for that as well.

Zidane:
@gratimax 

I'd rather it be done right and, barring additional work. this PR is pretty nice. All I'm saying is it needs to become a priority.

Flibio:
@gratimax 

I'm just waiting for the ability to get the raw message from the `MessageSinkEvent`. I need the plain message, without the player name included.

bloodmc:
@gratimax Any idea when this will be finished?

Wolfizen:
I think the best feature of this PR is separating placeholders from text.

ryantheleach:
@Wolfizen +1 It also solves the issue of having multiple matching "variables" in shoddy message templating (but that could have been solved by using a message templating language that has starting and ending tags / brackets

e.g. +g and +global matching (g)lobal making Administratorlobal

src/main/java/org/spongepowered/api/text/template/TextArgs.java
```diff
@@ -0,0 +1,113 @@
...
+    private TextArgs(ImmutableList.Builder<Object> positionedArgsBuilder,
+                     ImmutableMap.Builder<String, Object> namedArgsBuilder) {
+        this.positionedArgsBuilder = positionedArgsBuilder;
+        this.namedArgsBuilder = namedArgsBuilder;
+    }
+
+    public TextArgs add(Object arg) {
+        if (positionedArgsBuilder == null) {
+            throw new IllegalArgumentException("Positioned arguments were already accessed, to add" +
+                "more arguments use TextArgs.copy()");
```
Phroa: Should have a space after add

src/main/java/org/spongepowered/api/text/template/TextArgs.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        positionedArgsBuilder.add(arg);
+        return this;
+    }
+
+    public TextArgs add(Object... args) {
+        if (positionedArgsBuilder == null) {
+            throw new IllegalArgumentException("Positioned arguments were already accessed, to add" +
+                "more arguments use TextArgs.copy()");
```
Phroa: Another space after add

src/main/java/org/spongepowered/api/text/template/TextArgs.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        positionedArgsBuilder.add(args);
+        return this;
+    }
+
+    public TextArgs with(String name, Object arg) {
+        if (positionedArgsBuilder == null) {
+            throw new IllegalArgumentException("Named arguments were already accessed, to add" +
+                "more arguments use TextArgs.copy()");
```
Phroa: Another space after add

src/main/java/org/spongepowered/api/text/template/TextElements.java
```diff
@@ -0,0 +1,227 @@
...
+                return text;
+            }
+        };
+    }
+
+    public static <T> TextElement<T> function(final Function<T, Text> function) {
+        return new TextElement<T>() {
+            @Override
+            public Text create(T value) {
+                return Preconditions.checkNotNull(function.apply(value));
```
Phroa: I think `Preconditions` is meant to be used via static imports, right?

simon816: Yes

ST-DDT:
Sorry for repeating myself, but are there any thoughts about recreating the TextRepresentable feature? (Show Uuid for player and start messaging him on click, show the ItemStack/Achievement details in chat. With all plugins doing it the same way on a server without any knowledge how that impl might display them. And especially not limited to API provided ones.)

Have there been considerations on how to handle the (de-)serialization of the TextTemplates?

gratimax:
The TextRepresentable feature here is no longer useful, since you can simply use the relevant TextElement. This has many advantages, such as allowing for multiple representations of any object and one clear place to see all the representations. I have added a few elements that mimic TextRepresentable, but the other ones haven't been added yet.

I have thought about deserialization. It is certainly possible, but it is out of scope, and I think way too opinionated to fit into an API.

ST-DDT:
> I have thought about deserialization. It is certainly possible, but it is out of scope, and way too opinionated to fit into an API. With the TextElement/TextTemplate framework, however, it is easy for a plugin to provide its own deserialization.

IMO its a must have. Its probably one of the first things I will write for my plugins. And I would be glad if the API would do that for me (even if I have to write it for the API). The entire use of being able to load messages from json/file is directly connected to the ability to replace variables in messages loaded from there.

> Previously with placeholders, this was not possible, since they were mixed in with normal text objects.

Thats true, but has been fixed in the latest PR.

gratimax:
Just because text configuration is something that _you_ are going to write for your plugins doesn't mean that it should be a utility that is forced upon everyone using the API. "Serializing" templates goes straight to dealing with configuration, which is 100% out of scope with both this PR and the API in general. Let plugin developers decide how configuration will be stored.

boformer:
I disagree. The deserialization of templates should be standardized. That's very important for i18n.

simon816:
Could this be merged into #808 or at least a decision reached between the two because nothing seems to be happening at the moment

ST-DDT:
@simon816
Which features should be ported/merged over to my PR? AFAIK its possible to do everything proposed by this PR with mine already.
I will have an in detail look on the features of this PR this weekend to make sure I did not miss anything.

PS: My PR is almost up to date with master. I just have to update my testplugin and check some javadocs this weekend.

simon816:
The fact that this PR is still open suggests that this is PR is still to be considered.
From what @Minecrell [said on #808](https://github.com/SpongePowered/SpongeAPI/pull/808#issuecomment-153811391)
> It's waiting on completion of #835 so we can decide which one to use.

I think @Minecrell / @gratimax should check over similarities/differences as I don't have an opinion on this topic.

src/main/java/org/spongepowered/api/text/template/TextArgs.java
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Adds the given arguments to end of the positioned arguments.
+     *
+     * @param args The new arguments
+     * @return A new TextArgs object
+     */
+    public TextArgs add(Object... args) {
+        return new TextArgs(
+            ImmutableList.builder().addAll(positionedArgs).add(args).build(),
```
TheRaspPie: Shouldn't this be `addAll(positionedArgs).addAll(args)`?

gratimax:
Closing this because I don't have enough time to finish. Sorry about that!

PR #836, opened 2015-09-05
SpongeExecutorService binding annotations

PieGuy128:
[Common Side][1]

Continuing from the idea being discussed in #829, I've created two annotations that can be applied to a SpongeExecutorService parameter/field for easy injections.  These annotations allow the plugin developer to distinguish between a synchronous and an asynchronous executor.

[1]: https://github.com/SpongePowered/SpongeCommon/pull/154


Kiskae:
Another thing that was discussed was to have a default inject of the synchronous executor.
In my opinion it would be best to only allow injection with the annotations present, so the choice of executor is explicitly made.
Since a lot of concurrency systems that build upon the ExecutorService have a 'viral' concept of what a task should be executed upon, this will force the plugin developer to consciously choose where to execute.

PieGuy128:
Commits are all borked - not really sure why :/
Will either make a new PR or reopen when fixed

PR #840, opened 2015-09-11
Binding annotations for SpongeExecutorServices

PieGuy128:
*This is a continuation of #836, whose commit logs have been borked for some odd reason*

[Common side][1]

This PR adds two annotations that can be applied to a SpongeExecutorService field/parameter for quick and easy injections.  These annotations allow for a plugin developer to explicitly define the type of ExecutorService he/she wants injected (sync. vs. async.)

[1]: https://github.com/SpongePowered/SpongeCommon/pull/163

### DIFF
--- a/src/main/java/org/spongepowered/api/service/scheduler/Asynchronous.java
+++ b/src/main/java/org/spongepowered/api/service/scheduler/Asynchronous.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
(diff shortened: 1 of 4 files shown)


Zidane:
@PieGuy128 

Can you rebase this for master as well as possibly extend their names to @AsynchronousExecutor and likewise? Just seeing @Asynchronous could mean many things.

PieGuy128:
I see that the ExecutorService has yet to be committed on master (here and on Common).  Are there plans to re-commit it?

Cheers.

PR #849, opened 2015-09-27
Fix possible NumberFormatExceptions in Coerce.

JBYoshi:
When parsing `byte`s, `short`s, and `long`s, `Coerce` seems to assume that parsing failures via Java's built in primitive classes will return `null`, while they actually throw `NumberFormatException`s.

The reason only these methods needed to be fixed is that the other methods use Guava methods that return `null` in the case of a parsing failure; however, Guava doesn't provide methods for parsing `byte`s, `short`s, and `long`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/util/Coerce.java
+++ b/src/main/java/org/spongepowered/api/util/Coerce.java
@@ -402,8 +402,11 @@ public final class Coerce {
             return ((Number) obj).shortValue();
         }
 
-        Short parsed = Short.parseShort(Coerce.sanitiseNumber(obj));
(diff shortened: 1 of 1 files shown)


Zidane:
Nice catch @JBYoshi 

Will pull this shortly.

JBYoshi:
I noticed this because Eclipse was giving me a few warnings where "parsed cannot be null at this location".

JBYoshi:
@Zidane Status?

Zidane:
@JBYoshi 

We are wrapping up a big branch on Sponge for drops so I can go about pulling again. Apologies for the delay.

PR #850, opened 2015-09-27
Rename OwnableData to RepresentedPlayerData

saladoc:
This PR is required by SpongePowered/SpongeCommon#178

rename `OwnableData` to `RepresentedPlayerData` following the discussion on #857

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -25,6 +25,7 @@
 package org.spongepowered.api.data.key;
 
 import com.flowpowered.math.vector.Vector3d;
+
 import org.spongepowered.api.GameProfile;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.block.BlockType;
@@ -174,7 +175,7 @@ public final class Keys {
     public static final Key<MutableBoundedValue<Short>> SPAWNER_SPAWN_RANGE = null;
     public static final Key<MobSpawnerData.NextEntityToSpawnValue> SPAWNER_NEXT_ENTITY_TO_SPAWN = null;
     public static final Key<WeightedEntityCollectionValue> SPAWNER_ENTITIES = null;
-    public static final Key<Value<GameProfile>> OWNED_BY_PROFILE = null;
+    public static final Key<Value<GameProfile>> REPRESENTED_PLAYER = null;
     public static final Key<ListValue<PotionEffect>> POTION_EFFECTS = null;
     public static final Key<Value<ItemStackSnapshot>> REPRESENTED_ITEM = null;
     public static final Key<Value<Rotation>> ROTATION = null;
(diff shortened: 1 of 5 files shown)


gabizou:
In what way are ownerless ownables possible?

JBYoshi:
Ownerless ownables could be implemented by using a fake profile. They would be used if a plugin wanted to make an entity that appeared tamed but did not have an owner. However, if ownerless ownables are possible, I'm not sure if there's any reason to use `UUID`s to represent owners.

saladoc:
I've been talking with gabizou and he told me that ownerless ownables are represented by the absence of `OwnableData`, which makes sense. I've updated the PR accordingly.

saladoc:
Reason to use `UUID`s over `GameProfile`s:

If `GameProfile` is used, a lookup has to be performed whenever data is read from a `Wolf`, an `Ocelot` or a `Skull`. If `UUID` is used, a lookup has to be performed whenever data is set on a `Skull`. 
If a Plugin needs a `GameProfile` it may always perform the lookup itself. but for pure identification, the `UUID` is sufficient.

JBYoshi:
When I said "ownerless ownables", I was actually thinking of being able to let a wolf be tamed without being assigned to any specific player. Thoughts?

gabizou:
> When I said "ownerless ownables", I was actually thinking of being able to let a wolf be tamed without being assigned to any specific player. Thoughts?

Minecraft performs validation such that if there is no player `UUID` that the wolf is tamed to, then the wolf is no longer tamed. It is only "marked" if the `UUID` is a) valid, and b) will require leaving the wolf to be left to remain in the world forever to be loaded without being despawned.

Also, another thing is: This shouldn't be targeting only Wolves, this should target any EntityTameable.

ryantheleach:
So if ownable is to apply to wolves and ocelots as well, when making changes to the owner should the tameable data change as well? Or should one be removed in favor of the other?

gabizou:
> So if ownable is to apply to wolves and ocelots as well, when making changes to the owner should the tameable data change as well? Or should one be removed in favor of the other?

What?

saladoc:
The only thing of the owner that is stored in an `EntityTameable` is a stringified `UUID`. Nothing more. This is basically the main reason for this change. 
The implementation (see link in initial comment) supports wolves and ocelots (by hooking into `EntityTameable` and player skulls.

saladoc:
While we're at it: [`TameableData`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/TameableData.java) is somewhat made redundant by the definition of `OwnableData`. Should I remove the former in this PR as well?

gabizou:
One issue with this: `Skull`s use a `GameProfile` such that you'd be able to not only assign a player's name, but also the `UUID` such that if we require lookups of the `GameProfile` by `UUID`, it would be slightly different than what is already stored on the `Skull`. Because as of right now, you can have the `GameProfile` linking to an older skin or a different skin instead of the current skin of the `Player`. Thoughts?

gabizou:
@saladoc Is this ready for review and testing?

PR #851, opened 2015-09-27
Updates to Dispatcher and Disambiguator.

JBYoshi:
Add `SimpleDispatcher.get(String, @Nullable CommandSource)` to `Dispatcher`.
Make the `CommandSource` in `Disambiguator` `@Nullable`.

Implementation: SpongePowered/SpongeCommon#179
Addresses SpongePowered/Sponge#320.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
+++ b/src/main/java/org/spongepowered/api/service/command/SimpleCommandService.java
@@ -66,6 +66,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 
 /**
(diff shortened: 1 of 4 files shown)


AlphaModder:
Judging by the issue this is based on, would it not be better for /help to perhaps pass in the CommandSource calling it? Rather than just allowing null.

JBYoshi:
@AlphaModder The implementation PR updates `/help` to use the calling `CommandSource`. Using `@Nullable` allows for `SimpleDispatcher.get(String alias)` to return `get(alias, null)`.

AlphaModder:
@JBYoshi Why? What's it supposed to return when there's a conflict?

JBYoshi:
@AlphaModder It depends on the disambiguator. Some, such as `SpongeCommandDisambiguator` or `SimpleDispatcher.FIRST_DISAMBIGUATOR` don't need a `CommandSource`. Any that *do* require a `CommandSource` but are given `null` can simply return `Optional.absent()` (that's the equivalent of what happens without this PR) or delegate to another `Disambiguator`.

AlphaModder:
@JBYoshi Alright, sounds good.

JBYoshi:
@zml2008 Status?

JBYoshi:
Status?

PR #852, opened 2015-09-27
ExperienceHolderData API changes

lucdon:
changes to make EXPERIENCE_FROM_START_OF_LEVEL immutable see SpongePowered/SpongeCommon#177

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/ExperienceHolderData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/ExperienceHolderData.java
@@ -48,7 +48,7 @@ public interface ExperienceHolderData extends DataManipulator<ExperienceHolderDa
     BoundedValue<Integer> experienceSinceLevel();
 
     /**
-     * Gets the {@link BoundedValue} for the amount of experience
+     * Gets the {@link ImmutableBoundedValue} for the amount of experience
      * required between the current {@link #level()} and the next level.
      * This can be presumed to be the supposed "maximum" for the
      * {@link #experienceSinceLevel()} amount.
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/ExperienceHolderData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/ExperienceHolderData.java
@@ -27,6 +27,8 @@ package org.spongepowered.api.data.manipulator.mutable.entity;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableExperienceHolderData;
 import org.spongepowered.api.data.value.BoundedValue;
+import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
 import org.spongepowered.api.entity.living.Human;
 
 /**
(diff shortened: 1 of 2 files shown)


gabizou:
Looking at the issue a bit more, there's other things that need changing to properly represent what the data manipulator was supposed to have. That aside, this change isn't needed as `BoundedValue` is vague enough such that it may not be settable in this case (because implementation wise, it is an `ImmutableBoundedValue`).

lucdon:
the change in keys.class is needed else it gives an error in the value processor of that key. at least if the value processor has to be immutable(in keys.class ExperienceFromStartOfLevel was described as mutable: 

````
    public static final Key<MutableBoundedValue<Integer>> EXPERIENCE_FROM_START_OF_LEVEL = null;
````

the other change i will remove

**Comment by gabizou** [comment]
So, here's what I was going to change, but I haven't had a chance to do so yet:

1) Rename the Keys as following:
```java
public static final Key<MutableBoundedValue<Integer>> EXPERIENCE_LEVEL = null;
public static final Key<MutableBoundedValue<Integer>> EXPERIENCE_SINCE_LEVEL = null;
public static final Key<MutableBoundedValue<Integer>> TOTAL_EXPERIENCE = null;
public static final Key<ImmutableBoundedValue<Integer>> EXPERIENCE_BETWEEN_LEVELS = null;
```
2) Narrow the return value types for `ExperienceHolderData`:
```java
public interface ExperienceHolderData extends DataManipulator<ExperienceHolderData, ImmutableExperienceHolderData> {

    /**
     * Gets the {@link BoundedValue}  for the amount of experience
     * gained since the beginning of the current {@link #level()}. Normally,
     * the higher the level, the more "experience" required to gain another
     * level.
     *
     * @return The bounded value of experience since the beginning of the
     *     current level
     */
    MutableBoundedValue<Integer> experienceSinceLevel();

    /**
     * Gets the {@link BoundedValue} for the amount of experience
     * required between the current {@link #level()} and the next level.
     * This can be presumed to be the supposed "maximum" for the
     * {@link #experienceSinceLevel()} amount.
     *
     * @return The immutable bounded required experience between levels
     */
    ImmutableBoundedValue<Integer> getExperienceBetweenLevels();

    /**
     * Gets the {@link BoundedValue} for the current "level" of
     * experience deserved according to the {@link #totalExperience()} and
     * a function from implementation defining how much experience required
     * per level.
     *
     * @return The current level according to the amount of total experience
     */
    MutableBoundedValue<Integer> level();

    /**
     * Gets the total amount of experience stored.
     *
     * @return The value of total amount of experience
     */
    MutableBoundedValue<Integer> totalExperience();
```
And for the following reasons:
We can still perfectly represent the amount of experience for the current level the holder has, we can represent the current experience level, and we can represent the "total" amount of experience from 0 -> current, and finally, we can represent the amount of experience for leveling from the current level to the next level.

I don't mind you taking those changes in the API branch (including the immutable version), I'll accept the PR. Otherwise, let me know if you want me to make those changes to `master` after which you can update your implementation.

lucdon:
@gabizou i will update the api and the implementation to those changes tonight

lucdon:
ready for merge i guess

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/ExperienceHolderData.java
```diff
@@ -44,33 +46,33 @@
...
-     * Gets the {@link BoundedValue} for the amount of experience
+     * Gets the {@link ImmutableBoundedValue} for the amount of experience
      * required between the current {@link #level()} and the next level.
      * This can be presumed to be the supposed "maximum" for the
      * {@link #experienceSinceLevel()} amount.
      *
      * @return The immutable bounded required experience between levels
      */
-    BoundedValue<Integer> getExperienceBetweenLevels();
+    ImmutableBoundedValue<Integer> getExperienceBetweenLevels();
```
kashike: should be `experienceBetweenLevels`

lucdon:
@kashike i think they made it getExperienceBetweenLevels() instead of experienceBetweenLevels() because it is immutable and because of that you cant set it. so to emphasize that it is not mutable(and not settable) it is getExperienceBetweenLevels()  

@gabizou  correct me if im wrong

gabizou:
Merged in 682090e9d6297471b31b2495cdf215526915f1ca

PR #871, opened 2015-10-11
Add ExecutorService backed by SchedulerService

Kiskae:
This PR was previously commited as PR #829 and was merged into the old java8 branch, but it was not included in the final java8 update.
So this PR updates that PR so it can be added to the master branch.

SpongeCommon PR: https://github.com/SpongePowered/SpongeCommon/pull/200

This PR will also satisfy the requirements for #840


src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
```diff
@@ -81,4 +81,26 @@
      * @return A set of scheduled tasks
      */
     Set<Task> getScheduledTasks(Object plugin);
+
+
```
Phroa: Double blank line.

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.RunnableScheduledFuture;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A delegating ExecutionService that schedules all its tasks on
+ * Sponge's {@link SchedulerService}.
+ * <p>
```
Phroa: Style is a blank line before a `<p>` and no indentation for their contents (e.g. `<p> content and tags are inline </p>`)

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    SpongeFuture<?> scheduleAtFixedRate(Runnable command, long initialDelay, long period, TimeUnit unit);
+
+    @Override
+    SpongeFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit);
+
+    interface SpongeFuture<V> extends RunnableScheduledFuture<V> {
+
+        /**
```
Phroa: This method needs a summary sentence

Kiskae:
@Phroa handled, thanks for the comments

Phroa:
:heart:

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,66 @@
...
+         * Retrieves the {@link SchedulerService} {@link Task} instance that
+         * is responsible for the execution of this future.
+         *
+         * @return The backing task
+         */
+        Task getTask();
+    }
+}
```
simon816: Missing newline

Kiskae: @simon816 fixed

Kiskae:
API and implementation have been rebased onto `SchedulerService` changes

gabizou:
@Kiskae can you fix the errors causing Travis to fail?

Kiskae:
@gabizou Rebased and errors have been fixed

Kiskae:
I've included the binding annotations written by @Pieguy128 from PR https://github.com/SpongePowered/SpongeAPI/pull/840 since it heavily depends on this PR and it only required minimal changes to the PR.

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    <V> SpongeFuture<V> schedule(Callable<V> callable, long delay, TimeUnit unit);
+
+    @Override
+    SpongeFuture<?> scheduleAtFixedRate(Runnable command, long initialDelay, long period, TimeUnit unit);
+
+    @Override
+    SpongeFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit);
+
+    interface SpongeFuture<V> extends RunnableScheduledFuture<V> {
```
Xfel: Is there a specific reason for this class extending RunnableScheduledFuture? I don't see the use case where I would actually want to call the run method on that task, so I'm wondering why you would expose this.

Kiskae: Mostly to keep the interface as flexible as possible, if the users wants to execute the future before the scheduler can do it then they should be able to.

It also mirrors some of the access shown with the `Task` class, which allows access to the internal `Consumer` and can be executed much like `run()` does in `RunnableFuture`

gratimax:
This looks fantastic, but has been sitting unchanged for a while. @SpongePowered/developers any more concerns?

Kiskae:
@gratimax I actually had one last idea that could be added based on https://github.com/SpongePowered/SpongeAPI/issues/892
Basically document that the object passed to `create(A)syncExecutor` can be either a plugin object or plugin container and adjusting accordingly in the implementation

Kiskae:
Although that probably should be done as a seperate consistent change throughout the project where plugin objects are used to avoid creating special cases

zml2008:
@Kiskae That's already supported -- it's implemented in the plugin manager.

Xfel:
However, using this feature, you could omit the getInstance() calls on the PluginContainers and pass them directly - which would save us two map lookups.

Kiskae:
@zml2008 But its not specified in the javadocs: https://github.com/SpongePowered/SpongeAPI/blob/20bad1be5c4b80598d820b4c937c6390d0b0384b/src/main/java/org/spongepowered/api/plugin/PluginManager.java#L40 If that method can accept PluginContainers as well then it should probably be documented so implementers can rely on it (Since the javadocs only specify 'instance', which refers to '{@link Plugin} instances' according to the javadocs).

src/main/java/org/spongepowered/api/service/scheduler/AsynchronousExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+ * Provides a convenient injection for {@link SchedulerService#createAsyncExecutor(Object)}.
+ *
+ * <p>Use this annotation on a {@link SpongeExecutorService}.
+ * Remember that {@link Inject} is also necessary.</p>
+ */
+@BindingAnnotation
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER, ElementType.FIELD})
+public @interface AsynchronousExecutor {
+}
```
simon816: Missing newline

src/main/java/org/spongepowered/api/service/scheduler/SynchronousExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+ * Provides a convenient injection for {@link SchedulerService#createSyncExecutor(Object)}.
+ *
+ * <p>Use this annotation on a {@link SpongeExecutorService}.
+ * Remember that {@link Inject} is also necessary.</p>
+ */
+@BindingAnnotation
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER, ElementType.FIELD})
+public @interface SynchronousExecutor {
+}
```
simon816: Missing newline

src/main/java/org/spongepowered/api/service/scheduler/AsynchronousExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.inject.Inject;
+
+/**
+ * Provides a convenient injection for {@link SchedulerService#createAsyncExecutor(Object)}.
```
simon816: Maximum line length for javadocs is 80 characters.

src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
```diff
@@ -89,4 +89,25 @@
      * @return The preferred tick interval
      */
     int getPreferredTickInterval();
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule synchronous
```
simon816: use `{@link ExecutorService}`

src/main/java/org/spongepowered/api/service/scheduler/SchedulerService.java
```diff
@@ -89,4 +89,25 @@
...
+     * tasks through the standard Java concurrency interfaces.
+     *
+     * @param plugin The plugin that will own the created tasks
+     * @return A new executor service that can be used to execute
+     *         synchronous tasks
+     */
+    SpongeExecutorService createSyncExecutor(Object plugin);
+
+    /**
+     * Creates a new ExecutorService that can be used to schedule asynchronous
```
simon816: use `{@link ExecutorService}`

src/main/java/org/spongepowered/api/service/scheduler/SpongeExecutorService.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.spongepowered.api.service.scheduler;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.RunnableScheduledFuture;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A delegating ExecutionService that schedules all its tasks on
```
simon816: ExecutionService? Probably mean `{@link ExecutorService}`

src/main/java/org/spongepowered/api/service/scheduler/SynchronousExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.inject.Inject;
+
+/**
+ * Provides a convenient injection for {@link SchedulerService#createSyncExecutor(Object)}.
```
simon816: Maximum line length for javadocs is 80 characters.

Kiskae:
@simon816 Javadocs have been adjusted :+1:

simon816:
@SpongePowered/developers can this be given the go-ahead? Seems good to me

gabizou:
Correct me if I'm wrong but, this service seems to be reliant that it's not going to be replaced by plugins, is this correct?

If so, please name it something aside from service as core services will be renamed very soon to note their difference from replaceable services.

simon816:
@gabizou I assume it's following the naming scheme from `java.util.concurrent.ScheduledExecutorService`

Kiskae:
@gabizou as @simon816 says, its not a Service in the sense that it is a part of Sponge' ServiceManager. I'm following the naming scheme of the java API regarding the various executors.

PR #873, opened 2015-10-13
XpOrbData

clienthax:


src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -214,7 +214,7 @@
     public static final Key<MutableBoundedValue<Double>> ATTACK_DAMAGE = null;
     public static final Key<MapValue<EntityType, Double>> DAMAGE_ENTITY_MAP = null;
     public static final Key<Value<Boolean>> ELDER_GUARDIAN = null;
-    public static final Key<MutableBoundedValue<Integer>> CONTAINED_EXPERIENCE = null;
+    public static final Key<Value<Integer>> CONTAINED_EXPERIENCE = null;
```
gabizou: You can't have negative experience.

PR #890, opened 2015-10-28
Fix arbitrary discrete transform builder

gratimax:
@DDoS 

Verifies that the given matrix is a discrete transform matrix before creating a discrete transform. This is because if the sum of the values of a row is an integer for each row in the matrix, then the matrix is a discrete transform.

Informal justification:

matrix

We assume that the values in the column vector are arbitrary integers. For the products in the resulting vector to be integers, (a+b) and (c+d) must also be integers. This result can be generalized to 3x3 and 4x4 matrices.


DDoS:
Counter example:

    | 0.5 0.5 | | 1 |   | 0.5 |
    | 0.5 0.5 | | 0 | = | 0.5 |

I realize that the current discrete tranform API isn't that great, it should let you specify a rotation as a basis instead of the weird 90 angle multiplier stuff and let you pass arbitrary matrices with integer components.

gratimax:
Whoops, I forgot how to multiply matrices for a second there :smile:. Fixed the logic, now it checks that all values in the matrix are integers.

PR #893, opened 2015-10-30
Fix BoundedValues

ZephireNZ:
This makes sure that all values containing numbers are represented as BoundedValues, rather than as plain values.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -117,7 +117,6 @@ import org.spongepowered.api.data.type.StoneType;
 import org.spongepowered.api.data.type.TreeType;
 import org.spongepowered.api.data.type.WallType;
 import org.spongepowered.api.data.type.WireAttachmentType;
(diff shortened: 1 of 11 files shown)


gratimax:
Some of these make sense, but I don't understand why things like `SUCCESS_COUNT` have to bounded, since they are technically infinite

ZephireNZ:
Success count can't be less than zero - that's bounded.

Even if that wasn't the case, by making it a bounded value we can specify
that it has infinite bounds.

JBYoshi:
@gratimax In most of those cases, there is a minimum bound (typically 0). If the value has no maximum, just set it to `Integer.MAX_VALUE`.

gratimax:
Being a bounded value was supposed to imply a very specific use case -- those values which are bounded _on both sides_ in the game. Yes, Integer.MAX_VALUE is technically an upper bound, but it's not one that is imposed by game conditions.

If those conditions don't qualify, then the value type is simply a number type. We don't have a type for that because I don't think it's very useful. If it does turn out to be useful, then perhaps we do need a value type for comparables without bound, i.e. ComparableValue.

PR #894, opened 2015-10-30
Add event for steering entities.

Xemiru:
**SteerEntityEvent will be thrown when a player attempts to move while mounted on an entity.** This can be used in a plethora of fun shenanigans (i'm making a player-controlled tank after this), especially in things like minigames.

dat click

***What are the use cases?***
Aside from just being another thing you can play with in Minecraft, this can be used for a bunch of fun things that can come from someone's insane mind or minigames. Controllable vehicles, specifically.

***Why isn't it cancellable?***
Steering is controlled completely on the client-side; the client sends a packet ([C0CPacketInput](http://wiki.vg/Protocol#Steer_Vehicle)) whenever the player attempts to move while mounted on a vehicle. The client doesn't listen as to whether or not the mounted entity is actually controllable, and will send a packet regardless of what entity it sits on.

***Does this throw when a vehicle mounted by a player moves?***
No, that would be the job of a movement event. If it did, this would be cancellable. Steering movement does not directly control the entity the player is mounted on; instead the movement is sent to the server to consider applying the movement onto the entity being ridden.

***How is this to be implemented?***
I don't exactly know how you guys do your implementation stuffs yet, but the way I've done my testing is by listening for the [C0CPacketInput / Steer Vehicle](http://wiki.vg/Protocol#Steer_Vehicle) packet and retrieving the values stored within.

***Any extras?***
A nice addition to this would be to allow `EntityType`s to identify themselves as controllable by steering or not. Easily not important, though.

*Wanna do some testing yourself? [Here's the code I used!](https://gist.github.com/Xemiru/c0fdb413e0f81a912f01)*
**NOTE:** The given code used for testing utilizes classes from the forge source code and the netty library. Both dependencies MUST be present within your project to be able to build correctly. Usage instructions and what the testing code does is within the provided gist.

The code also accesses Minecraft code directly; this means it must either be compiled with ForgeGradle for a plugin jar, or ran in a development environment.


src/main/java/org/spongepowered/api/event/entity/SteerEntityEvent.java
```diff
@@ -0,0 +1,77 @@
...
+import org.spongepowered.api.event.cause.CauseTracked;
+
+/**
+ * Raised when an {@link Entity} is being steered by a mounted {@link Player}.
+ * 
+ * <p>This event does not raise for when an entity mounted by another entity
+ * moves; it is only thrown when a player attempts to steer a mounted
+ * entity.</p>
+ */
+public interface SteerEntityEvent extends TargetEntityEvent, CauseTracked {
```
simon816: Pretty sure it should be called `PlayerSteerEntityEvent` to match `Source|Action|Target`

Xemiru: Following the naming scheme of most of the classes within the `api.event.entity` package.

also i have no idea where 'Source|Action|Target` is

src/main/java/org/spongepowered/api/event/entity/SteerEntityEvent.java
```diff
@@ -0,0 +1,77 @@
...
+     * <p>Positive values represent movement in the direction the mounted player
+     * is looking at (holding Move Forward), negative values represent movement
+     * in the opposite direction. A value of 0 represents no steering movement
+     * forwards or backwards.</p>
+     * 
+     * <p>A notchian client will, in most cases usually send a value of 0.98
+     * going forwards or backwards; 0.3 (rounded up from 0.29400003) while
+     * sneaking.</p>
+     */
+    public float getForwardSpeed();
```
kashike: Redundant `public` modifier (all methods).

src/main/java/org/spongepowered/api/event/entity/SteerEntityEvent.java
JBYoshi: @Xemiru Events are usually supposed to be in the form of `[Source][Action][Target]Event`. Parts may be omitted if they are not applicable (e.g. a generic `PlayerInteractEvent`).

ryantheleach:
I know it's still future stuff, but has anyone looked at how 1.9 is structured? Apparently there is better cheat detection and boats (a common buggy vehicle) got a complete redo.

Xemiru:
@ryantheleach boat speed got nerfed down to hell.
probably fixed in future updates, but rip speeding across the ocean
and the bugs for boats are probably because they're so fast anyway. i forgot how they control but its either a new packet for the oars or listening for whenever a player tries lclick/rclick on a boat

>better cheat detection
what

src/main/java/org/spongepowered/api/event/entity/SteerEntityEvent.java
gabizou: Usually, sources have been extracted out to the `Cause` object such that they are not explicitly requiring extra sure based events when th rebase event would've handled just fine.

Xemiru: ... is this name fine or not

Zidane: Should be ControlEntityEvent for when something controls what the Entity is doing.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    boolean isJumping();
+    
+    /**
+     * Sets whether or not the mounted player was jumping during this control
+     * event.
+     * 
+     * @param flag whether or not the player was jumping
+     */
+    boolean setJumping(boolean flag);
```
kashike: Missing `@return`.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,115 @@
...
+     * 
+     * @param flag whether or not the player was sneaking
+     */
+    void setSneaking(boolean flag);
+    
+    /**
+     * Returns whether or not the mounted player jumped during this control
+     * event.
+     */
+    boolean isJumping();
```
kashike: Missing `@return`.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,115 @@
...
+     * 
+     * @param speed the speed at which the strafing control movement is going
+     */
+    void setStrafeSpeed(float speed);
+    
+    /**
+     * Returns whether or not the mounted player was sneaking during this
+     * control event.
+     */
+    boolean isSneaking();
```
kashike: Missing `@return`.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,115 @@
...
+     * mounted {@link Player} is attempting to move at.
+     * 
+     * <p>Positive values represent strafing to the left of the mounted player,
+     * negative values represent strafing to the right. A value of 0 represents
+     * no steering movement sideways.</p>
+     * 
+     * <p>A notchian client will, in most cases usually send a value of 0.98
+     * going sideways; 0.3 (rounded up from 0.29400003) while sneaking.</p>
+     */
+    float getStrafeSpeed();
```
kashike: Missing `@return`.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,115 @@
...
+     * <p>Positive values represent movement in the direction the mounted player
+     * is looking at (holding Move Forward), negative values represent movement
+     * in the opposite direction. A value of 0 represents no steering movement
+     * forwards or backwards.</p>
+     * 
+     * <p>A notchian client will, in most cases usually send a value of 0.98
+     * going forwards or backwards; 0.3 (rounded up from 0.29400003) while
+     * sneaking.</p>
+     */
+    float getForwardSpeed();
```
kashike: Missing `@return`.

Cybermaxke:
You can also add the jump charge (for horses) to the event, the packet that sends this (C0BPacketEntityAction) is sent before the C0CPacketInput so it can be caught and then used later. The packey is only send when the jump state changes from true to false (releasing the spacebar) while mounting a horse.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,125 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.entity;
+
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.event.cause.CauseTracked;
+
+/**
+ * Raised when an {@link Entity} is being steered by a mounted {@link Player}.
```
simon816: I'd use the word 'Fired' instead of 'Raised' personally.

You say when steered by a player, is this not a more general entity-controlling-entity event?

Xemiru: Entity AI doesn't ever try to steer. This event is also based off of [this packet](http://wiki.vg/Protocol#Steer_Vehicle), which is only ever thrown when the player attempts to perform any movement while mounted on any entity.

gabizou:
@Xemiru can you run `gradle genEventImpl` so that the `SpongeEventFactory` is properly updated for this event?

Xemiru:
I forgot this existed.

I have no idea how trying to update this thing from a little over half a year ago is gonna go.

Ran `genEventImpl`, ready to cry if it looks messy.

Xemiru:
I forgot this existed.

I have no idea how updating this branch that's a little over half a year old is gonna go.

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    boolean isSneaking();
+
+    /**
+     * Sets whether or not the mounted player was sneaking during this control
+     * event.
+     * 
+     * @param flag whether or not the player was sneaking
+     */
+    void setSneaking(boolean flag);
```
kashike: `flag` -> `sneaking`

src/main/java/org/spongepowered/api/event/entity/ControlEntityEvent.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    boolean isJumping();
+
+    /**
+     * Sets whether or not the mounted player was jumping during this control
+     * event.
+     * 
+     * @param flag whether or not the player was jumping
+     */
+    void setJumping(boolean flag);
```
kashike: `flag` -> `jumping`

gabizou:
@Xemiru so, here's my thoughts: I can imagine this getting merged into bleeding as 1.9 does change how vehicles are moved (sadly this is very true), but I have the sneaking suspicion that @bloodmc would technically be able to add this event's implementation to master for usage with Minecraft 1.8.9 (though mod compatibility could be questionable for mob riding added by mods). I'd like @Zidane to possibly give this a look over if he's got a chance before I get around to it sometime next week after I hopefully finish a few higher priority things for SpongeCommon bleeding.

Xemiru:
I'm not even sure if this should really be an event now that I get to look at it again; its mostly just a data bag for something that's pretty invisible. The way I'd really use this event is to just store what data it pushes up and update it per player, so my repeating task can query that data and change the state of the custom vehicle accordingly.

And I'm not necessarily sure how 1.9's vehicle changes really affects this, as all this event is doing, should it be registered, is just listening for a packet and telling what it brought back. The event itself should have no direct control over the vehicle, otherwise this'd just be another version of a displacement event. The only control this event would have over the vehicle is by manipulating the values it lets you change so as to get a certain reaction from the server when it applies the data to the player and vehicle (e.g. setting all values to 0 and false, telling the server that the player didn't do anything while sitting on the entity even if they did prior to the changes and therefore having no effect).

Deamon5550:

Xemiru:
(can't reopen it myself)

DDoS:
Can't reopen it because "the repository that submitted this pull request has been deleted".

PR #896, opened 2015-10-31
Additional Logging API and Action Logger system

zml2008:
This API functions as a layer on top of slf4j, both providing increased
access to logging appenders and filters, as well as allowing the
logging of Text messages rather than just plain Strings.

These changes allow for loggers to have improved integration with
Sponge and for plugins to more easily intercept plugin log messages.

Additionally, the concept of an action logger is added. This is more
complicated than the existing API changes, and is designed for plugins
that perform actions that have a source, one or more targets, and the
need to notify some combination of those and the log/console of events
that occur.

A lot of this is still pretty WIP -- especially implementations of LogTargets, and most of the ActionLoggger system -- I'm not sure exactly how that will end up working out, but there will have to be some sort of additional configuration -- users should be able to customize the messages sent by the ActionLogger.

Implementationwise, there isn't much yet. It should be possible to integrate this into how Log4j2 works -- the biggest issue is mapping log4j Appenders to sponge LogTargets, second biggest the fact that LogTargets and Appender/Filters work slightly differently -- but should still be workable.

### DIFF
--- a/src/main/java/org/spongepowered/api/logging/LogTarget.java
+++ b/src/main/java/org/spongepowered/api/logging/LogTarget.java
@@ -1,0 +1,125 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.logging;
+
+import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
+
+import com.google.common.collect.ImmutableSet;
+import org.spongepowered.api.text.Texts;
+import org.spongepowered.api.text.sink.MessageSink;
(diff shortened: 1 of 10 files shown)


kashike:
I'd like to see Log4j2 exposed myself.

zml2008:
@SpongePowered/developers I don't have any opposition to exposing log4j2 -- as long as we're ok dealing with the (probably quite small) likelihood of mojang changing later on. Bigger issue is that switching could break a lot of plugins (unless we deprecated slf4j first).

progwml6:
@zml2008 that could lead to forcing other implementations of the API built on clones of the game to need to use log4j2 additionally we need slf4j anyway for some of the libraries we use. 

When we changed originally from log4j2 to slf4j we did it purposely to avoid having problems with different implementations using different logging implementations to avoid the mess that resulted when mojang switched to log4j2 where messages were getting lost between systems, or outputted to .log files incorrectly. slf4j can be hooked to any underlying log implementation easily, and i personally believe its a better api than log4j2's.

kashike:
@progwml6 slf4j is lacking things from log4j2 that @zml2008 is **trying** to add in this PR, but will result in non-standard implementations and lack of platform-independent support -- instead, exposing log4j2 makes it simpler.
slf4j doesn't know about `Appender`, `Filter`, etc. all of which are helpful.
As it is right now, there are a lot of things people do that disqualify different implementations (using Forge, depending on libraries that are provided by MC/Forge that are not part of the API, etc).

ST-DDT:
Can you give me a usecase that isn't covered by log4j config alone?

kashike:
@ST-DDT plugins can't touch the log4j configuration

ST-DDT:
> @ST-DDT plugins can't touch the log4j configuration

Why would you want to do that? (Give me a usecase.) Why should plugins be able to do so? IMO plugins should either document how to configure their logs in the main config or let them bundle a log4j config file, that can be attached to the main config by the server at runtime/will only affect the one plugin that provides it.

zml2008:
plugins may want to provide custom appenders, or custom filters, and nobody wants to work with XML.

ST-DDT:
Some guys won't want to use log4j at all...

Minecrell:
Minecraft uses a really old beta version of log4j2, which could possibly make upgrading log4j2 difficult without breaking plugins in the future. Using slf4j makes us completely independent from the implementation's logging framework and the version of log4j2. Without Minecraft using a proper stable version of log4j2 I don't think it is a good idea exposing it in the API at all.

As for these changes, how would you implement them? You would either need to replace the complete slf4j implementation, or wrap the slf4j loggers in a custom one that would add the new methods.

ST-DDT:
> As for these changes, how would you implement them? You would either need to replace the complete slf4j implementation, or wrap the slf4j loggers in a custom one that would add the new methods.

Maybe Mixins could solve it, although the impl will probably will get very messy if we start Mixin into third party libs.

Minecrell:
@ST-DDT Log4j2 has a [classloader exclusion](https://github.com/Mojang/LegacyLauncher/blob/master/src/main/java/net/minecraft/launchwrapper/LaunchClassLoader.java#L58) globally in LaunchWrapper, therefore they would never get the mixins applied in any way.

src/main/java/org/spongepowered/api/logging/LogTarget.java
```diff
@@ -0,0 +1,125 @@
...
+        return filter == null ? this : new FilteredLogTarget(filter, this);
+    }
+
+    /**
+     *
+     * @param others
+     * @return
+     */
+    default LogTarget with(LogTarget... others) {
+        return new MultiLogTarget(ImmutableSet.<LogTarget>builder().addAll(Arrays.asList(others)).add(this).build());
```
Deamon5550: This could be changed to `ImmutableSet.of(others)

Deamon5550:

PR #902, opened 2015-11-04
Make FireworkData more broad

ZephireNZ:
Currently FireworkData only applies to the firework rocket item and entity. This is because it references the flight modifier, which is only added once the rocket is fully crafted.

This PR changes it so that the modifier is an optional value - this way, we can not only use FireworkData for rockets, but for firework stars as well.

This also updates the API docs to make clear that a star can only have one effect attached to it - the first effect in the list.


gabizou:
I would honestly choose to make them separate instead. One `FireworkEffectData` and one `FireworkFlightData` or something.

ZephireNZ:
FireworkRocketData maybe?

ZephireNZ:
So one is a single list data, the other a multilist + bounded data

gabizou:
@ZephireNZ nope, one list data, and one just for the bounded data.

PR #903, opened 2015-11-05
Replace the use of java.awt.Color with our own Color

gabizou:
Currently, the API uses `java.awt.Color` due to it's useful capabilities, however, the issue with it is that it's not stable in that it can not be translated to and from various other types of colors that are and will be exposed in the API. This includes `DyeColor` and soon `MapColor`. The benefit of this is that we have a final class `Color` that provides not only some translations between `java.awt.Color`, but also adds some static methods for usability.

### DIFF
--- a/src/main/java/org/spongepowered/api/text/format/TextColors.java
+++ b/src/main/java/org/spongepowered/api/text/format/TextColors.java
@@ -25,8 +25,7 @@
 package org.spongepowered.api.text.format;
 
 import org.spongepowered.api.text.Text;
-
-import java.awt.Color;
+import org.spongepowered.api.util.Color;
 
 /**
  * TextColors is a list of text colors provided by Vanilla Minecraft.
@@ -43,7 +42,7 @@ public final class TextColors {
      */
     public static final TextColor NONE = new TextColor() {
 
-        private final Color color = new Color(0, 0, 0, 0);
+        private final Color color = Color.BLACK;
 
         @Override
         public String getName() {
--- a/src/main/java/org/spongepowered/api/util/Color.java
+++ b/src/main/java/org/spongepowered/api/util/Color.java
@@ -1,0 +1,245 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3f;
+import com.flowpowered.math.vector.Vector3i;
+import org.apache.commons.lang3.Validate;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.MemoryDataContainer;
+import org.spongepowered.api.data.Queries;
+import org.spongepowered.api.data.type.DyeColor;
+import org.spongepowered.api.util.persistence.DataBuilder;
+import org.spongepowered.api.util.persistence.InvalidDataException;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public final class Color implements DataSerializable {
+
+    private static final int MASK = 0xFF;
+
+    public static final Color BLACK = ofRgb(0x000000);
+
+    public static final Color GRAY = ofRgb(0x808080);
+
+    public static final Color WHITE = ofRgb(0xFFFFFF);
+
+    public static final Color BLUE = ofRgb(0x0000FF);
+
+    public static final Color GREEN = ofRgb(0x008000);
+
+    public static final Color LIME = ofRgb(0x00FF00);
+
+    public static final Color RED = ofRgb(0xFF0000);
+
+    public static final Color YELLOW = ofRgb(0xFFFF00);
+
+    public static final Color MAGENTA = ofRgb(0xFF00FF);
+
+    public static final Color PURPLE = ofRgb(0xAA00FF);
+
+    public static final Color DARK_CYAN = ofRgb(0x008B8B);
+
+    public static final Color DARK_GREEN = ofRgb(0x006400);
+
+    public static final Color DARK_MAGENTA = ofRgb(0x8B008B);
+
+    public static final Color CYAN = ofRgb(0x00FFFF);
+
+    public static final Color NAVY = ofRgb(0x000080);
+
+    public static final Color PINK = ofRgb(0xFF00AA);
+
+    public static Color ofRgb(int hex) {
+        return ofRgb((hex >> 0x10) & MASK, (hex >> 0x8) & MASK, hex & MASK);
+    }
+
+    public static Color ofRgb(int red, int green, int blue) {
+        return new Color(red, green, blue);
+    }
+
+    public static Color of(java.awt.Color color) {
+        return new Color(color.getRed(), color.getGreen(), color.getBlue());
+    }
+
+    public static Color of(Vector3i vector3i) {
+        return new Color(vector3i.getX(), vector3i.getY(), vector3i.getZ());
+    }
+
+    public static Color of(Vector3f vector3f) {
+        return new Color(Math.round(vector3f.getX()), Math.round(vector3f.getY()), Math.round(vector3f.getZ()));
+    }
+
+    public static Color of(Vector3d vector3d) {
+        return new Color((int) Math.round(vector3d.getX()), (int) Math.round(vector3d.getY()), (int) Math.round(vector3d.getZ()));
+    }
+
+    public static Color mixDyeColors(DyeColor... colors) {
+        Validate.noNullElements(colors, "No nulls allowed!");
+        final Color[] actualColors = new Color[colors.length];
+        for (int i = 0; i < colors.length; i++) {
+            actualColors[i] = colors[i].getColor();
+        }
+
+        return mixColors(actualColors);
+    }
+
+    public static Color mixColors(Color... colors) {
+        Validate.noNullElements(colors, "No null colors allowed!");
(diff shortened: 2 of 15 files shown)


Deamon5550:
:D

gabizou:
Obviously for now though, the PR is lackluster with javadocs and the default static colors provided. Now, it would be cool if people who are more aware of color theory to comment. @Deamon5550 maybe?

src/main/java/org/spongepowered/api/text/format/TextColors.java
```diff
@@ -43,7 +42,7 @@ private TextColors() {
      */
     public static final TextColor NONE = new TextColor() {
 
-        private final Color color = new Color(0, 0, 0, 0);
+        private final Color color = Color.WHITE;
```
simon816: Should be black

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,204 @@
...
+
+    public static final Color DARK_GREEN = ofHex(0x006400);
+
+    public static final Color DARK_MAGENTA = ofHex(0x8B008B);
+
+    public static final Color NAVY = ofHex(0x000080);
+
+    public static final Color UNKNOWN = ofHex(0xFF00AA);
+
+    public static Color ofHex(int hex) {
```
kenzierocks: I would not call this `ofHex` because the `int` doesn't need to be in hex format. I think `ofInt` might be better.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,204 @@
...
+        } else {
+            int red = colors[0].getRed();
+            int green = colors[0].getGreen();
+            int blue = colors[0].getBlue();
+            for (int i = 1; i < colors.length; i++) {
+                red += colors[i].getRed();
+                green += colors[i].getGreen();
+                blue += colors[i].getBlue();
+            }
+            double averageRed = red / colors.length;
```
kenzierocks: You've got a division bug here. `int / int = int` not `double`.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,204 @@
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        final Color other = (Color) obj;
+        return Objects.equals(this.red, other.red)
```
kenzierocks: Consider using `==` maybe?

src/main/java/org/spongepowered/api/util/Color.java
gabizou: The thing is that that's what the argument is being used as, a hex representation of a color.

kenzierocks: But this is also the convenient method for doing `int -> Color`. It doesn't make sense to do `Color.ofHex(mySavedColor)`.

gabizou: I guess then it should be named `ofRGB`? and a separate one for accepting `ofRGBA`?

kenzierocks: Yea, I'm not sold on any name, just don't like `ofHex`.

DDoS: Should be `ofPackedBytes(int)` or just `ofPacked(int)`.

src/main/java/org/spongepowered/api/util/Color.java
DDoS: No alpha mixing? This should be (in functional pseudo code) `colors.sum(color -> color.rgb * color.a) / colors.sum(color -> color.a)`

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,204 @@
...
+
+    public static Color ofHex(int hex) {
+        return ofRGB(hex >> 16 & MASK, hex >> 8 & MASK, hex & MASK);
+    }
+
+    public static Color ofRGB(int red, int green, int blue) {
+        return new Color(red, green, blue);
+    }
+
+    public static Color ofJavaColor(java.awt.Color color) {
```
DDoS: Should be `ofAwtColor()`

src/main/java/org/spongepowered/api/util/Color.java
gabizou: @DDoS I thought about alphas, but Minecraft doesn't use alphas.

src/test/java/org/spongepowered/api/util/ColorTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+import org.junit.Test;
+
+public class ColorTest {
```
DDoS: These tests seem to be missing all asserts... Also Java `assert` statements are disabled at runtime by default, this should use JUnit Asserts

gabizou: Again, it's a work in progress. as you can see, not everything is tested as it should.

DDoS: Oh, well it wasn't marked as WIP, sorry about that.

DDoS:
`ofVec(Vector3i/f/d)` would be a good addition.

src/main/java/org/spongepowered/api/util/Color.java
Minecrell: ... Or why not just `of(int)`? It's not like there could be really several methods accepting only one int, is there?

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,204 @@
...
+
+    public static final Color NAVY = ofHex(0x000080);
+
+    public static final Color UNKNOWN = ofHex(0xFF00AA);
+
+    public static Color ofHex(int hex) {
+        return ofRGB(hex >> 16 & MASK, hex >> 8 & MASK, hex & MASK);
+    }
+
+    public static Color ofRGB(int red, int green, int blue) {
```
Minecrell: Should be `ofRgb`

src/main/java/org/spongepowered/api/util/Color.java
gabizou: There could be, the question is the format of the packed `int` since there's various ways to represent a color.

Minecrell: Isn't the format you're accepting here pretty much the standard everywhere, e.g. also used in CSS etc.?

src/main/java/org/spongepowered/api/text/format/TextColors.java
ST-DDT: But the default color in texts is white.

src/main/java/org/spongepowered/api/util/Color.java
ST-DDT: I would prefer ofRGB as its the clearest solution what input is expected. It also may allow other int based methods such as ofRGBA ofHLS etc in case we ever need those...

src/main/java/org/spongepowered/api/util/Color.java
ST-DDT: I disagree here. RGB is an abbreviation and a well known term that shouldnt be mutated as it has negative impact on code readability/clearness

src/main/java/org/spongepowered/api/text/format/TextColors.java
simon816: Oh so `new Color(0, 0, 0, 0)` is white?

Minecrell: This is supposed to be simply "no color defined", it's not really related to any color that can be represented in Minecraft directly, it means the color will be inherited.

simon816: Ah an alpha of 0 makes it white, weird why this wasn't just java.awt.Color.WHITE to begin with

src/main/java/org/spongepowered/api/util/Color.java
Minecrell: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/org/spongepowered/api/util/Color.java
DDoS: `ofRGB` implies 3 separate color components. Packed components are almost always RGBA 8bit.

src/main/java/org/spongepowered/api/util/Color.java
Minecrell: I would just use `of(java.awt.Color)` here too, the parameter already implies that it is from a java color.

src/main/java/org/spongepowered/api/util/Color.java
liach: But SpongeAPI does not always follows google code style. [Example](https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation). The code of the api has an indentation of 4 spaces.

Minecrell: We use it with the additions/modifications explained in https://docs.spongepowered.org/en/contributing/guidelines.html#code-style

gabizou:
> ofVec(Vector3i/f/d) would be a good addition. The floating point variants should take in values in the range [0, 1], as is standard in CG.

I added one where it just blindly accepts the `Vector3d/f` as floored, but could you elaborate on how it should be calculating the range from 0,1? Just multiply by 255?

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,217 @@
...
+
+    public static Color of(Vector3d vector3d) {
+        return new Color(vector3d.getFloorX(), vector3d.getFloorY(), vector3d.getFloorZ());
+    }
+
+    public static Color mixDyeColors(DyeColor... colors) {
+        Validate.noNullElements(colors, "No nulls allowed!");
+        final Color[] actualColors = new Color[colors.length];
+        for (int i = 0; i < colors.length; i++) {
+            actualColors[i] = colors[i].getColor();
```
Minecrell: Why not just check for null using Guava Preconditions as usual here, no need to go through the array multiple times

gabizou: @Minecrell I could, but Validate will do it already for me, meh. I can change it.

gabizou:
@SpongePowered/developers Any objections to this being merged? (is there anything missing?)

src/main/java/org/spongepowered/api/util/Color.java
Minecrell: It just seems a bit inconsistent to use Validate and then again preconditions, but whatever. You can keep it if you want

DDoS: Floating point colours are always normalized, 0 is min, 1 is max. The byte range makes no sense here. Scale the values and round instead of flooring.

Minecrell:
@gabizou Javadocs? :P

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -0,0 +1,217 @@
...
+
+    public static final Color DARK_CYAN = ofRgb(0x008B8B);
+
+    public static final Color DARK_GREEN = ofRgb(0x006400);
+
+    public static final Color DARK_MAGENTA = ofRgb(0x8B008B);
+
+    public static final Color NAVY = ofRgb(0x000080);
+
+    public static final Color UNKNOWN = ofRgb(0xFF00AA);
```
Deamon5550: Ok if you don't want to call it `Hollywood Cerise` then how about `PINK` ?

src/test/java/org/spongepowered/api/util/ColorTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void testColorAwtTranslation() {
+        final Color color = Color.of(java.awt.Color.BLUE);
+        final java.awt.Color javaTranslated = color.asJavaColor();
+        assertTrue(java.awt.Color.BLUE.equals(javaTranslated));
+    }
+
+    @Test
+    public void testAverage() {
```
Deamon5550: These tests aren't actually testing anything

gabizou: Yeah... I don't know how else to actually test this.

PR #904, opened 2015-11-05
Add MinecartBlockData

ZephireNZ:
Implementation: SpongePowered/SpongeCommon#246

Currently there is no way in the API to represent a minecart with a block inside. This fixes that by adding `MinecartBlockData` - which contains a BlockState and offset.

Yep, that's about it!

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableMinecartBlockData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableMinecartBlockData.java
@@ -1,0 +1,50 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableMinecartBlockData.java
```diff
@@ -0,0 +1,26 @@
+package org.spongepowered.api.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.manipulator.mutable.entity.MinecartBlockData;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.entity.vehicle.minecart.Minecart;
+
+public interface ImmutableMinecartBlockData extends ImmutableDataManipulator<ImmutableMinecartBlockData, MinecartBlockData> {
```
ZephireNZ: Javadocs needed here

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/MinecartBlockData.java
```diff
@@ -0,0 +1,26 @@
+package org.spongepowered.api.data.manipulator.mutable.entity;
+
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableMinecartBlockData;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.vehicle.minecart.Minecart;
+
+public interface MinecartBlockData extends DataManipulator<MinecartBlockData, ImmutableMinecartBlockData> {
```
ZephireNZ: and here

PR #908, opened 2015-11-06
instrumentData and AxisData removal

lucdon:
InstrumentData isnt possible to implement
AxisData is only used in quartzBlocks only that is better to handle by just adding the axis to QuartzTypes(which is how minecraft handles EnumQuartzType)


AlphaModder:
@lucdon Why can't we use insturmentdata?

ryantheleach:
Because it's entirely dependent on the block below the noteblock, not the noteblock itself. I'm confused at why wood and quartz don't use axis data however.

gabizou:
@ryantheleach Because they are very specific in that they don't just accept `Axis` values, there's [`LogAxss`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/type/LogAxes.java#L31), and soon, `QuartzAxes`.

JBYoshi:
@gabizou Why don't logs and quartz just use the same `Axis` class?

PR #909, opened 2015-11-06
Move LockableData and ImmutableLockableData to the base manipulator packages.

JBYoshi:
They apply to minecart containers as well.


gabizou:
Why are we moving this to base?

JBYoshi:
Because they can be applicable to entities (minecarts with chests) as well as tile entities.

JBYoshi:
Wait a minute... `EntityMinecartContainer` has this code for its implementation of `ILockableContainer`:
```java
public boolean isLocked() {
    return false;
}
public void setLockCode(LockCode code) {}
public LockCode getLockCode() {
    return LockCode.EMPTY_CODE;
}
```
So it doesn't truly support lock data. Never mind.

PR #910, opened 2015-11-09
DelayableData

lucdon:
DelayableData for representing the Delay of a RedstoneRepeater


JBYoshi:
I don't see any issues with the API side.

PR #911, opened 2015-11-10
Add Fall Time key to falling block

clienthax:


JBYoshi:
Why would you need to use this?

clienthax:
@JBYoshi When spawning a falling block from a block that is air the logic in falling block will instantly destroy the entity,
if you set it to 1 and spawn it in air it will work as intended, this basically makes it possible to spawn the entity anywhere and not just where the original block was located

JBYoshi:
@clienthax It could be fixed by adding code to the implementation of `spawnEntity` to set the fall time to 1.

saladoc:
I think the time a block has been falling is also relevant for calculating damage dealt by and to a falling anvil.

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableFallingBlockData.java
```diff
@@ -76,4 +76,19 @@
      */
     ImmutableValue<Boolean> canDropAsItem();
 
+    /**
+     * Gets the time the block has been falling
+     * if spawning a entity in air this will need to be set to 1 or it will be instantly removed
```
phroa: These sentences need periods, and `if` should be capitalized

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableFallingBlockData.java
```diff
@@ -76,4 +76,19 @@
...
+     *
+     * @return The time the block has been falling
+     */
+    ImmutableValue<Integer> fallTime();
+
+    /**
+     * Gets whether this falling block will damage entities where it lands
```
phroa: Period

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/FallingBlockData.java
```diff
@@ -74,4 +74,19 @@
      */
     Value<Boolean> canDropAsItem();
 
+    /**
+     * Gets the time the block has been falling
+     * if spawning a entity in air this will need to be set to 1 or it will be instantly removed
```
phroa: Ditto on this file

PR #918, opened 2015-11-17
MapItemData API

hsyyid:
**What is MapItemData API?**

Quite simply, it's an API made for the data contained by Minecraft's filled map items. It's built to give plugin developers an easy way of modifying the data within these maps. 

**How it Works.**

I created a class called `MapItem` which gives access to the `colors` of the map, the `scale`, the `center`, the `world` of the map, the `width`, the `height`, as well as the `visible players` on that map. A plugin would then get this value from the `MapItemData`, which returns a `MapItem` that plugin developers can manipulate and offer back to the `ItemStack`.

**Notes**

- MapData is stored in the world NBT.
- Maps have methods for getting their MapData, but it requires the world object. (Need a good way of getting that in the processors)
- MapData has good methods for easily changing and updating values within it.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/item/MapItemData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/item/MapItemData.java
@@ -26,13 +26,21 @@ package org.spongepowered.api.data.manipulator.mutable.item;
 
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.item.ImmutableMapItemData;
+import org.spongepowered.api.data.value.mutable.Value;
 import org.spongepowered.api.item.ItemTypes;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.util.mapitem.MapItem;
 
 /**
- * Represents an {@link DataManipulator} hosting the specific map
- * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
- * // TODO for modwizcode
+ * Represents an {@link DataManipulator} hosting the specific map information of
+ * an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
  */
 public interface MapItemData extends DataManipulator<MapItemData, ImmutableMapItemData> {
+
+    /**
+     * Gets this map {@link MapItem}.
+     *
+     * @return The map.
+     */
+    Value<MapItem> map();
 }
--- a/src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
+++ b/src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
@@ -1,0 +1,157 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util.mapitem;
+
+import com.flowpowered.math.vector.Vector2i;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.awt.Color;
+import java.awt.Image;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Represents a map.
+ */
+public interface MapItem {
+
+    /**
+     * Gets this map's width.
+     *
+     * @return The width of the map
+     */
+    int getWidth();
+    
+    /**
+     * Gets this map's height.
+     *
+     * @return The height of the map
+     */
+    int getHeight();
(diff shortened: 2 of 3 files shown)


saladoc:
Any reason for having the colors be a `Byte` instead of some meaningful (Pseudo-)Enum?

hsyyid:
@saladoc None except for the fact that Minecraft has it stored as a byte array.

src/test/java/org/spongepowered/api/util/test/TestTextFactory.java
```diff
@@ -82,7 +82,7 @@ public TextRepresentation xml() {
 
     @Override
     public char getLegacyChar() {
-        return '§';
+        return '�';
```
simon816: Whatever you did here don't commit this change

ST-DDT: Changed charset

Deamon5550:
Or how about using our fancy new Colour class #903 

We shouldn't be exposing the colours as a meaningless byte that has to be mapped to an actual colour.

hsyyid:
@simon816 Yeah, I know. I'll fix the pr soon

hsyyid:
@Deamon5550 Okay. I'll do that soon.

gabizou:
@hsyyid is this a WIP or is it ready for review?

If it's ready for review, how do we actually work with Maps and provide map instances to map data?

src/main/java/org/spongepowered/api/data/manipulator/mutable/item/MapItemData.java
```diff
@@ -26,13 +26,56 @@
...
+
+    /**
+     * Gets the visible players by name {@link String} on this map.
+     *
+     * @return The list of players visible.
+     */
+    ListValue<String> visiblePlayers();
+    
+    /**
+     * Gets the colors {@link Byte} on this map.
```
gabizou: This is quite literally the worst way to phrase "here's how to make a map! admire my list of byte values!"

hsyyid:
@gabizou It's WIP. Not sure why the title was changed back.

simon816:
IMO the map information itself should not be represented as a data manipulator. Rather, it should be it's own class, `MapItem`
`MapItemData` would simply have a method `Value<MapItem> map()` to obtain this object.
I'm thinking something like how ban data is represented, there's [`BanData`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/BanData.java) and a [`Ban`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/ban/Ban.java) interface

hsyyid:
@simon816 I like this idea.

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+ * Represents a map.
+ */
+public interface MapItem {
+    
+    /**
+     * Gets this map's x center.
+     *
+     * @return The x center of the map.
+     */
+    Integer getXCenter();
```
simon816: This (and the method bellow) should just be `Vector2i getCenter()`

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+     * @return The z center of the map.
+     */
+    Integer getZCenter();
+
+    /**
+     * Gets this map's dimension.
+     *
+     * @return The dimension ID of the map.
+     */
+    Integer getDimensionId();
```
simon816: Dimension ID is not a thing in the API
Could just be a world UUID

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Gets this map's scale.
+     *
+     * @return The scale of the map.
+     */
+    Integer getScale();
```
simon816: Why use the integer wrapper? `int` is fine here

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Gets the visible players by name on this map.
+     *
+     * @return The list of players visible.
+     */
+    List<String> getVisiblePlayers();
```
simon816: Should be either Player/User objects or the UUIDs

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+    
+    /**
+     * Gets the colors on this map.
+     *
+     * @return The list of colors.
```
simon816: Don't put full-stops at the end of `@tag` lines

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    List<Color> getColors();
```
simon816: The map is 2-dimensional but this is a list (1-dimensional) I don't think this method really makes sense. Perhaps there should be `getColor(Vector2i)`

src/main/java/org/spongepowered/api/util/mapitem/MapItemBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.spongepowered.api.util.mapitem;
+
+import org.spongepowered.api.item.ItemTypes;
+import org.spongepowered.api.item.inventory.ItemStack;
+
+/**
+ * Represents a builder that creates MapItems.
+ */
+public interface MapItemBuilder {
```
JBYoshi: With the recent changes, this should be an inner class named `Builder` inside `MapItem`.

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Gets the visible players by name on this map.
+     *
+     * @return The list of players visible
+     */
+    List<String> getVisiblePlayers();
+    
+    /**
+     * Gets the color on this map at specific pixel.
+     *
```
simon816: Missing `@param pixel`

src/main/java/org/spongepowered/api/util/mapitem/MapItem.java
```diff
@@ -0,0 +1,111 @@
...
+     * @return The scale of the map
+     */
+    int getScale();
+
+    /**
+     * Gets the visible players by name on this map.
+     *
+     * @return The list of players visible
+     */
+    List<String> getVisiblePlayers();
```
simon816: Should be `List<UUID>`

modwizcode:
This seems like a good start to me, I don't think it's the full-featured API I was working on, so I might like to see some of those features. Mainly the full featured ability to control player cursors on the map, handle creating "virtual" maps not backed by any real storage and the ability to force replace a map by setting the id to the same as an existing map. 

Unless I'm mistaken the intention currently is that Sponge has a relatively stable API now, so I'd like to see some more work on this before I feel it's really a stable framework to build on. This leaves a lot of work in the plugins, which will be duplicated quite a bit and doesn't expose all the useful features of maps.

See https://github.com/SpongePowered/SpongeAPI/pull/531 for some concrete ideas.

hsyyid:
@modwizcode Thanks for the suggestions.

PR #920, opened 2015-11-17
[WIP] Phase 1 - AI Framework

Zidane:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/264) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/414)

It has been a long time coming but I wanted to be the one to tackle AI.

Phase 1 involves Goals and Tasks. A Goal is simply a collection of Tasks. All Agents in Minecraft have two Goals, NORMAL and TARGET. In each Goal, tasks are arranged by priority (lower = higher priority).

Task is where all the action is. It is the goal of our API to handle tweaking of Vanilla AI, Mod AI, but also adding custom ones (using AbstractAITask).

### DIFF
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -29,8 +29,12 @@ import org.spongepowered.api.data.ImmutableDataRegistry;
 import org.spongepowered.api.data.manipulator.DataManipulatorRegistry;
 import org.spongepowered.api.data.value.ValueFactory;
 import org.spongepowered.api.entity.EntityType;
+import org.spongepowered.api.entity.ai.task.AITaskType;
+import org.spongepowered.api.entity.ai.task.AbstractAITask;
+import org.spongepowered.api.entity.living.Agent;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.item.recipe.RecipeRegistry;
+import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.scoreboard.displayslot.DisplaySlot;
 import org.spongepowered.api.statistic.BlockStatistic;
@@ -281,6 +285,19 @@ public interface GameRegistry {
      * @param modifier The modifier to register
      */
     void registerWorldGeneratorModifier(WorldGeneratorModifier modifier);
+
+    /**
+     * Registers a new {@link AbstractAITask} with an {@link Agent} as the owner. The complete id
+     * will be in the format of <code>{@link PluginContainer#getId()}:id</code>.
+     *
+     * @param plugin The plugin who owns it
+     * @param id The id that represents the task type
+     * @param ownerClass The class of the owner
+     * @param aiClass The class of the task
+     * @return The type
+     */
+    <T extends Agent> AITaskType registerAITaskType(Object plugin, String id, String name, Class<T> ownerClass, Class<? extends
+            AbstractAITask<T>> aiClass);
 
     /**
      * Gets the {@link PopulatorFactory} for creating {@link Populator}s and
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -292,12 +292,10 @@ public interface GameRegistry {
      *
      * @param plugin The plugin who owns it
      * @param id The id that represents the task type
-     * @param ownerClass The class of the owner
      * @param aiClass The class of the task
      * @return The type
      */
-    <T extends Agent> AITaskType registerAITaskType(Object plugin, String id, String name, Class<T> ownerClass, Class<? extends
-            AbstractAITask<T>> aiClass);
+    AITaskType registerAITaskType(Object plugin, String id, String name, AbstractAITask<? extends Agent> aiClass);
 
     /**
      * Gets the {@link PopulatorFactory} for creating {@link Populator}s and
--- a/src/main/java/org/spongepowered/api/GameRegistry.java
+++ b/src/main/java/org/spongepowered/api/GameRegistry.java
@@ -295,7 +295,7 @@ public interface GameRegistry {
      * @param aiClass The class of the task
      * @return The type
      */
-    AITaskType registerAITaskType(Object plugin, String id, String name, AbstractAITask<? extends Agent> aiClass);
+    AITaskType registerAITaskType(Object plugin, String id, String name, Class<? extends AbstractAITask<? extends Agent>> aiClass);
 
     /**
      * Gets the {@link PopulatorFactory} for creating {@link Populator}s and
--- a/src/main/java/org/spongepowered/api/entity/ai/Goal.java
+++ b/src/main/java/org/spongepowered/api/entity/ai/Goal.java
@@ -32,10 +32,10 @@ import org.spongepowered.api.entity.living.monster.Monster;
 import java.util.List;
 
 /**
- * Represents a series of tasks meant to achieve a "goal".
+ * Represents a set of tasks that will be updated together by an {@link Agent}.
  *
- * In Minecraft, most {@link Agent}s have just one goal which is living or simulating, or
- * whichever term you want to call it. Other agents, such as {@link Monster}s, have a
+ * In Minecraft, most agents have just one goal which is living or simulating, or
+ * whichever term you want to call it. Other agents, such as {@link Monster}s, use a
  * "target" goal which serves to seek out potential victims and execute a series of attack
  * tasks.
  *
--- a/src/main/java/org/spongepowered/api/entity/ai/Goal.java
+++ b/src/main/java/org/spongepowered/api/entity/ai/Goal.java
@@ -1,0 +1,101 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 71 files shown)


liach:
Will more impl of ais coming soon?

Zidane:
@liach 

Yep, all will be interfaced out.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -296,6 +300,19 @@
...
+     * Registers a new {@link AbstractAITask} with an {@link Agent} as the owner. The complete id
+     * will be in the format of <code>{@link PluginContainer#getId()}:id</code>.
+     *
+     * @param plugin The plugin who owns it
+     * @param id The id that represents the task type
+     * @param ownerClass The class of the owner
+     * @param aiClass The class of the task
+     * @return The type
+     */
+    AITaskType registerAITaskType(Object plugin, String id, String name, Class<? extends Agent> ownerClass, Class<? extends AbstractAITask<? extends
```
simon816: Should `ownerClass` be the same as the class in the `AbstractAITask` generic?
If so then should use a generic on the method `<T extends Agent>` ... `Class<T> ownerClass, Class<? extends AbstractAITask<T>>`

Zidane: Right, you are correct. I meant to change that before commit.

AlphaModder:
So, we're definitely going with the "wrap vanilla AI" approach?

Zidane:
@AlphaModder 

We aren't wrapping their AI. We are exposing the attributes of it so you can tweak it + creating your own. Once I expose the PathFinder, you could write a lib that re-creates the Vanilla AI with your own specifications.

AlphaModder:
@Zidane I meant using the same system, not the specific AITasks. We're not implementing our own AI system.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -296,6 +300,19 @@
...
+     * will be in the format of <code>{@link PluginContainer#getId()}:id</code>.
+     *
+     * @param plugin The plugin who owns it
+     * @param id The id that represents the task type
+     * @param ownerClass The class of the owner
+     * @param aiClass The class of the task
+     * @return The type
+     */
+    <T extends Agent> AITaskType registerAITaskType(Object plugin, String id, String name, Class<T> ownerClass, Class<? extends
+            AbstractAITask<T>> aiClass);
```
liach: Why is there an id and an extra name? Name is redundant.

src/main/java/org/spongepowered/api/entity/ai/task/AITaskType.java
```diff
@@ -0,0 +1,49 @@
...
+
+@CatalogedBy(AITaskTypes.class)
+public interface AITaskType extends CatalogType {
+
+    /**
+     * Gets the {@link Agent} class that this type is owned by.
+     *
+     * @return The owner class
+     */
+    Optional<Class<? extends Agent>> getOwnerClass();
```
liach: Can we make this return `Agent.class` by default? Also `Optional` is not needed since we can return `Agent.class`.
```java
default Class<? extends Agent> getOwnerClass() {
    return Agent.class;
}
```

src/main/java/org/spongepowered/api/entity/ai/task/AITaskType.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Gets the {@link AITask} class that this type represents.
+     *
+     * @return The ai class
+     */
+    Class<? extends AITask<?>> getAIClass();
```
liach: Here I suggest to add an `Optional`, otherwise all `AITaskType`s needs an actual entity to exist.

Zidane: No, the point is for all AITaskTypes to have an AI class.

src/main/java/org/spongepowered/api/GameRegistry.java
Zidane: Wrong. For pretty printing, CatalogTypes have a Name.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return The type
+     */
+    AITaskType getType();
+
+    /**
+     * Gets the {@link Goal} that is executing this task.
+     *
+     * @return The goal
+     */
+    Goal<O> getGoal();
```
liach: `Optional` here.

src/main/java/org/spongepowered/api/entity/ai/task/AITaskType.java
Zidane: Wrong. Plugins don't implement AITaskType, SpongeCommon does this. Due to mods, the owner class is not known until its added to the Entity UNLESS the AITaskType is created by SpongeAPI through the register method. Since I force you to give me the class and I enforce that class to be the same as the generic or a child, it is safe.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Gets the {@link Agent} that owns this task.
+     *
+     * @return The owner
+     */
+    default O getOwner() {
+        return getGoal().getOwner();
+    }
```
liach:
```java
default Optional<O> getOwner() {
    if (this.getGoal().isPresent()) {
        return getGoal().get().getOwner();
    }
    return Optional.empty();
}
```

src/main/java/org/spongepowered/api/entity/ai/task/AITaskType.java
Zidane: After thinking about it, I think I'd be fine doing that...as Optional is odd in this case.

src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -47,6 +48,8 @@
  */
 public interface World extends Extent, WeatherUniverse, Viewer, Contextual {
 
+    Game getGame();
```
liach: Why to add this?

Zidane: Shouldn't be in here. @gabizou was supposed to remove it.

src/main/java/org/spongepowered/api/entity/ai/Goal.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.spongepowered.api.entity.ai.task.AITask;
+import org.spongepowered.api.entity.ai.task.AITaskType;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.entity.living.monster.Monster;
+
+import java.util.List;
+
+/**
+ * Represents a series of tasks meant to achieve a "goal".
```
liach: I think we'd describe it not only a series but also a category of tasks. All tasks must have a goal, which categorizes it.

src/main/java/org/spongepowered/api/entity/ai/task/AITaskType.java
liach: Missed that this is returning a `Class`. Bad.

src/main/java/org/spongepowered/api/entity/ai/Goal.java
Zidane: Going to argue semantics...? Really?

liach: In other words, it should be called "a set of `AITask`s that will be executed together on an `Agent` which could be replaced by other sets".

liach:
I wonder that if `Goal` is the set of `AITask`s, might it be possible to apply similar goals to different individuals of the same type of entities?

src/main/java/org/spongepowered/api/entity/ai/Goal.java
```diff
@@ -0,0 +1,101 @@
...
+     * @return The type
+     */
+    GoalType getType();
+
+    /**
+     * The {@link Agent} that owns this goal.
+     *
+     * @return The owner
+     */
+    O getOwner();
```
liach: I wonder whether we need this... Or just 
```java
Class<O> getOwnerClass();
```
In my opinion the advantage of a goal is that it could be easily copied and applied on to other entities.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return The goal
+     */
+    Goal<O> getGoal();
+
+    /**
+     * Gets the {@link Agent} that owns this task.
+     *
+     * @return The owner
+     */
+    default O getOwner() {
```
liach: [Same](https://github.com/SpongePowered/SpongeAPI/pull/920/files#r45221742)

src/main/java/org/spongepowered/api/entity/ai/Goal.java
Zidane: Goals are the direct instances sitting on an Agent currently. It isn't the intention, at this time, to let you create custom goals (as that involves actually rigging those into the update loop).

I'll think about it.

src/main/java/org/spongepowered/api/world/World.java
gabizou: I was supposed to remove it? I thought you wanted it there?

Minecrell:
Can you add a checkstyle suppression for AI, just like for TNT and PEBKAC https://github.com/SpongePowered/SpongeAPI/blob/master/checkstyle-suppressions.xml#L7-L8?

liach:
Anything else needed to change? I think this is perfect for phase 1.

src/main/java/org/spongepowered/api/entity/ai/task/AITaskBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A utility to assist in building {@link AITask}s.
+ *
+ * @param <O> The type of agent
+ * @param <A> The type of task
+ * @param <B> The type of builder, self-referencing
+ */
+public interface AITaskBuilder<O extends Agent, A extends AITask<O>, B extends AITaskBuilder<O, A, B>> extends ResettableBuilder<B> {
```
simon816: Shouldn't this builder be inside `AITask`?

Zidane: Yes, I've not merged this into master for a reason :P.

liach:
:+1:  Finally.

PR #921, opened 2015-11-17
PlayerSimulator changes

simon816:
I'm taking the conversation on https://github.com/SpongePowered/SpongeAPI/commit/d0bc27252b81e2288a2a19681585bac112437a74#commitcomment-13578083 into a PR so we can discuss actual API changes.

Hopefully the issues raised so far have been addressed.

One thing I'm not sure about is how to make `PlayerSimulator` a `Subject` in the implementation since it's not got a real UUID.
One thought is that plugins provide the simulator a subject e.g a player.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/PlayerSimulator.java
+++ b/src/main/java/org/spongepowered/api/world/PlayerSimulator.java
@@ -24,22 +24,95 @@
  */
 package org.spongepowered.api.world;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.profile.GameProfile;
+import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.util.Direction;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+import javax.annotation.Nullable;
 
 /**
  * Simulates the results of a player interacting with the world.
  */
-public interface PlayerSimulator {
+public interface PlayerSimulator extends Subject {
+
+    /**
+     * A factory for creating instances of {@link PlayerSimulator}.
+     */
+    class Factory {
+
+        private static final Function<GameProfile, PlayerSimulator> factory = null;
+
+        /**
+         * Creates a new {@link PlayerSimulator} with the given profile.
+         *
+         * @param profile The game profile
+         * @return A new simulator instance
+         */
+        public static PlayerSimulator createSimulator(GameProfile profile) {
+            return factory.apply(checkNotNull(profile, "profile"));
+        }
+    }
+
+    /**
+     * Gets the game profile that represents this simulator.
+     *
+     * @return The game profile
+     */
+    GameProfile getProfile();
+
+    /**
+     * Gets the world this simulator is currently in, if it's in a loaded world.
+     *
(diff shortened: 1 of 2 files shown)


simon816:
A few design questions:

Currently the methods take a Vector3i or x/y/z and there is a setWorld() and getWorld() method.
Would it be better to have a Location parameter? Problem being quite expensive if a large area was wanted to be simulated upon.

Does it make sense for `PlayerSimulator` to extent Subject` or would the subject be something that's set in a `setSubject(Subject)` method.

The question @JBYoshi asked above

simon816:
Does anyone have any opinion on the above questions?
If not then this PR is complete

ryantheleach:
Small niggling suggestion, if everything else uses a builder, why do player simulators use a factory?

Zidane:
@simon816 

I'll answer @JBYoshi 's and say that if ANY of these methods trigger an event then a Cause should be optionally provided.

simon816:
I've added a `setCause(Cause)` method to the interface. Could keep it like that or add a cause to the end of all methods (which would bloat the method signatures a bit)

src/main/java/org/spongepowered/api/world/PlayerSimulator.java
```diff
@@ -24,14 +24,82 @@
...
+    Cause getCause();
+
+    /**
+     * Sets the cause that will be passed to events fired from actions caused by
+     * this simulator. If null is passed, the default cause is this simulator as
+     * the root.
+     *
+     * @param cause The cause, or null to use the default cause.
+     */
+    void setCause(@Nullable Cause cause);
```
gabizou: Make this fluent please so I can do the following:
```java
if (playerSimulator.setCause(Cause.of(NamedCause.source(this), NamedCause.of("simulator", playerSimulator))).interactBlock(pos, Direction.UP)) {
 // do stuff
}
```

src/main/java/org/spongepowered/api/world/PlayerSimulator.java
```diff
@@ -52,18 +122,21 @@
     void interactBlock(int x, int y, int z, Direction side);
```
gabizou: This should be return `boolean` as well.

src/main/java/org/spongepowered/api/world/PlayerSimulator.java
```diff
@@ -39,7 +107,9 @@
      * @param position The position of the block
      * @param side The side of the block to interact with
      */
-    void interactBlock(Vector3i position, Direction side);
+    default void interactBlock(Vector3i position, Direction side) {
```
gabizou: Return boolean since it's `Cancellable`.

src/main/java/org/spongepowered/api/world/PlayerSimulator.java
```diff
@@ -52,18 +122,21 @@
...
-     * the player had done so.
+     * Simulates the interaction with this object using the given item as if the
+     * player had done so.
      *
      * @param position The position of the block
      * @param itemStack The item
      * @param side The side of the block to interact with
      */
-    void interactBlockWith(Vector3i position, ItemStack itemStack, Direction side);
+    default void interactBlockWith(Vector3i position, ItemStack itemStack, Direction side) {
```
gabizou: Likewise.

simon816:
@ryantheleach It could be either really but I went with factory because it essentially is just an instance 'factory' and there's no configuration methods, just supply a profile and get a simulator.

gabizou:
Closing as discussed in IRC, the idea of a player simulator is useless in the API.

PR #922, opened 2015-11-18
Add JukeboxData

clienthax:


JBYoshi:
I think this is supposed to be handled by `RepresentedItemData`. @gabizou Is this correct?

gabizou:
This doesn't need to exist since `RepresentedItemData` is all that is needed.

Issue #900, opened 2015-11-04
Run event in case of cancellation.


Azatik:
I'm writing a plugin to the graves.
If a player dies in a private area, it will not break the grave.
How to allow the player to break the grave in a private area?
![2015-11-04_20 59 19](https://cloud.githubusercontent.com/assets/10154956/10944110/4d599dc2-8339-11e5-803a-f3dfe1bba4da.png)

Zidane:
A question: What do you define as a "private area"?

Zidane:
Also, the Listener annotation has:

https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/Listener.java#L59

Simply specify it as false and you'll get cancelled events.

Azatik:
Private area - Anti Grief plugin

Azatik:
How to use "ignoreCancelled()"

kenzierocks:
`@Listener(ignoreCancelled=true)`. Read up on annotations :)

Azatik:
I will write in a different way.
I want execute an event when his cancels other plugin.

Zidane:
So, just to decipher this:

You only want to handle this event if it was cancelled via the AntiGrief plugin?

Azatik:
Allow to break blocks on the territory of the World Guard (any Anti Grief Plugin).

Zidane:
@Azatik 

Then what you want to do is adjust your Order to POST as well as ignoreCancelled = false

https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/event/Order.java#L111

Then, if your criteria is met, setCancelled(false)

Azatik:
I understood.I write the result of sixteen hours.

Azatik:
<h4>Report.</h4>

I created two events.
EventBreakBlock:
```
public class EventBreakBlock {   
    @Listener
    public void onEventBreakBlock(ChangeBlockEvent.Break event) {
        if (event.getCause().first(Player.class).isPresent()) {
            Player player = (Player) event.getCause().first(Player.class).get();
            Text msg = Texts.of(TextColors.DARK_RED, "Action prohibited.");
            player.sendMessage(msg);
            event.setCancelled(true);
            }
        }
    }
```

EventBreakBlockAllow:
```
public class EventBreakBlockAllow {
    @Listener(ignoreCancelled=false, order = POST)
[...]

PR #927, opened 2015-11-20
[WIP] Add Event Filters

Deamon5550:
[Impl](https://github.com/SpongePowered/SpongeCommon/pull/274)

#### Event Filtering
This branch adds annotation based event filtering to assist with reducing the amount of boilerplate necessary while writing event listeners.

##### Event type Filters
These filters are applied to the method and apply a range of filters based on the event object.
######  Include
This annotation allows you to listen to a supertype event (such as EntityEvent) and filter it to only receive a specified subset of the event type's subclasses.
```java
    @Listener
    @Include({ InteractBlockEvent.Secondary.class })
    public void include(InteractBlockEvent event) {
        // This listener will only fire if the event is one of the included types
    }
```
######  Exclude
The opposite of include, this allows you to listen to a supertype event and filter it to not receive a specified subset of the event type's subclasses.
```java
    @Listener
    @Exclude({ InteractBlockEvent.Secondary.class })
    public void exclude(InteractBlockEvent event) {
        // This listener will only fire if the event is not one of the excluded types
    }
```
######  IsCancelled
This will filter the event such that your listener will only be called if the event has been cancelled by a previous event listener. This is essentially the opposite of the `ignoreCancelled` flag on the Listener annotation.
```java
    @Listener(order = Order.LATE, ignoreCancelled = false)
    @IsCancelled
    public void cancel(InteractBlockEvent.Secondary event) {
        // This listener will only fire if the event was previously cancelled
    }
```
#####  Parameter Sources
These annotations are applied to additional parameters in the event handler method and specify a source for that parameter. If an object matching the source and type is not found in the cause then your listener is not called.

######  First
This gets the first object from the event `Cause` matching the parameter type.

A fairly typical event listener for a player event will go something like the following
```java
    @Listener
    public void onInteract(InteractBlockEvent.Secondary event) {
        Optional<Player> player = event.getCause().first(Player.class);
        if(player.isPresent()) {
            //do something
        }
    }
```
With the `@First` annotation this can be simplified to the following
```java
    @Listener
    public void first(InteractBlockEvent.Secondary event, @First Player player) {
        // This listener will fire if the event cause has a player
        // The player parameter will be the first player in the cause chain
    }
```
######  Last
Equivalent to `First` except it calls `Cause#last(Class)`.
######  All
The `All` annotation allows you to receive all cause objects of a certain type. It must be applied to an array type and is equivalent to `Cause#allOf(Class)`
```java
    @Listener
    public void all(InteractBlockEvent.Secondary event, @All Player[] players) {
        // This listener will fire if the event cause has at least one player
        // players will contain all players in the cause chain
    }
```
By default your listener will be skipped if the returned array would be empty, this can be disabled however by setting the ignoreEmpty flag to false.
```java
    @Listener
    public void all(InteractBlockEvent.Secondary event, @All(ignoreEmpty = false) BiomeTypes[] players) {
        // This listener will always fire, as even if there are no BiomeTypes
        // objects in the cause chain (and there won't be) its set to not ignore
        // empty and therefore will simply return an empty array
    }
```

#####  Parameter Filters
These filters apply additional conditions to a parameter. These require that the paramter also have a parameter source annotation (see above).

###### Has
This takes a data manipulator type and checks that the annotated data holder has a data manipulator of the specified type. If the object is not a `DataHolder` or does not have a matching manipulator then your listener is not called. Equivalent to `DataHolder#has(Class)`.

###### Supports
Similar to `Has` except the it is only checked that the DataHolder supports the given data manipulator type. Equivalent to `DataHolder#supports(Class)`.

A full example plugin can be found [here](https://gist.github.com/Deamon5550/6d4e4cdec9361ddace93)

TODO:
- [x] Add Root cause annotation
- [ ] remove `ignoreCancelled` from Listener

### DIFF
--- a/src/main/java/org/spongepowered/api/event/filter/cause/First.java
+++ b/src/main/java/org/spongepowered/api/event/filter/cause/First.java
@@ -22,39 +22,23 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.event;
+package org.spongepowered.api.event.filter.cause;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import org.spongepowered.api.event.cause.Cause;
 
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Used to annotate a method as an {@link EventListener}.
- *
- * <p>The method being targeted must be public and must be in a class that is
- * also public.</p>
+ * Sets the parameter to the first object in the cause chain of the
+ * parameter type.
+ * 
+ * @see Cause#first(Class)
  */
-@Retention(RUNTIME)
-@Target(METHOD)
-public @interface Listener {
-
-    /**
-     * The order this listener should be called in relation to other listeners in
-     * the {@link EventManager}.
-     *
-     * @return The order the listener should be called in
-     */
-    Order order() default Order.DEFAULT;
-
-    /**
-     * Whether this listener should be called before any other server mods, such
-     * as Forge mods. All Sponge event listeners are called after mods, unless
-     * they specify the {@link #beforeModifications()} flag to be true.
-     *
-     * @return If the listener should be fired before other server mods
-     */
-    boolean beforeModifications() default false;
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface First {
 
 }
--- a/src/main/java/org/spongepowered/api/event/filter/type/Include.java
+++ b/src/main/java/org/spongepowered/api/event/filter/type/Include.java
@@ -22,39 +22,30 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.event;
+package org.spongepowered.api.event.filter.type;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
-
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Used to annotate a method as an {@link EventListener}.
- *
- * <p>The method being targeted must be public and must be in a class that is
- * also public.</p>
+ * Filters out all event types which are not in the specified array of event
+ * classes. This allows listening for a supertype event and filtering to only
+ * receive events for a specific subset of the annotated event's subtypes.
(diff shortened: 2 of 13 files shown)


Zidane:
I'm really onboard with this concept because I can see it being used to cut out a few lines of code :P. I know I'd use it myself.

ST-DDT:
i agree although i would prefer if the event filter (subtype or cancelled state) to be applied to the method instead or alternativly to the param because its easier to read there

Minecrell:
Not sure if the annotations should be really as subclasses, most code styles (including ours) will always import this like `@EventTypeFilters.IsCancelled` instead of `@IsCancelled`

saladoc:
Maybe we should remove the `ignoreCancelled` value from `@Listener` and move that whole functionality to Event Filter annotations.

Deamon5550:
@ST-DDT Could do yeah, it would change the target type which would be good because it would prevent you from applying these filters to any other parameters.

@Minecrell Thats a good point I'll move them out.

@saladoc I thought about that but the two functions are a little different. Although while thinking about it perhaps I could add a tristate to it to specify the three cases NOT_CANCELLED, IGNORE_CANCELLED, and ONLY_CANCELLED.

src/main/java/org/spongepowered/api/event/filter/cause/First.java
```diff
@@ -0,0 +1,44 @@
...
+ * Sets the parameter to the first object in the cause chain of the
+ * parameter type.
+ * 
+ * @see Cause#first(Class)
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface First {
+
+}
```
ST-DDT: Missing newline at EOF

src/main/java/org/spongepowered/api/event/filter/type/Include.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Filters out all event types which are no in the specified array of event
+ * classes. This allows listening for a supertype event and filtering to only
+ * receive events for a specific subset of the annotated event's subtypes.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Include {
```
ST-DDT: You should specify which fitler is stronger. Include or Exclude or that they are exclusive.

ST-DDT:
Nice

Azatik:
The annotation `ignoreCancelled` has been removed. Is there an analogue?

JBYoshi:
Add `@IsCancelled(Tristate.FALSE)` before your method in addition to `@Listener`.

Azatik:
@JBYoshi, i wrote code:
```java
@Listener (order = BEFORE_POST)
@IsCancelled(Tristate.FALSE)
public void onChangeGrave(ChangeBlockEvent.Break event) {
    ...
}
```

The event will still be canceled.
Note: The old issue: https://github.com/SpongePowered/SpongeAPI/issues/900

simon816:
To replicate the old `ignoreCancelled` you now do `@IsCancelled(Tristate.UNDEFINED)`

Azatik:
@simon816, thanks. Everything works well.

Issue #822, opened 2015-08-24
Additional plugin metadata- author/description


varrix:
Are there any plans to expand the `@Plugin` annotation to incorporate an "author" and "description". This is something I would like to use for polling and sorting plugins to be displayed for more information in-game.

wizjany:
related https://github.com/SpongePowered/SpongeAPI/issues/624

Minecrell:
This is planned, however it depends on #624 being decided first.

Zidane:
Closing since Minecrell's PR covers this + more.

PR #928, opened 2015-11-21
Extend platform API

Minecrell:
This extends the Platform API by replacing specific methods like `getName()` or `getApiVersion()` with an exposed `PluginContainer` for the API and the implementation, which provides all of these methods and possibly more in the future (as soon we have added more [plugin metadata](https://github.com/SpongePowered/SpongeAPI/issues/822)).

This also matches the current design in SpongeVanilla and Forge where the components of the implementation (e.g. Minecraft, Forge, SpongeForge) are all represented as separate plugins, even though they aren't standalone plugins.

SpongeCommon PR: https://github.com/SpongePowered/SpongeCommon/pull/277
SpongeVanilla PR: https://github.com/SpongePowered/SpongeVanilla/pull/187
SpongeForge PR: https://github.com/SpongePowered/SpongeForge/pull/423

### DIFF
--- a/src/main/java/org/spongepowered/api/plugin/PluginContainer.java
+++ b/src/main/java/org/spongepowered/api/plugin/PluginContainer.java
@@ -24,6 +24,11 @@
  */
 package org.spongepowered.api.plugin;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
 /**
  * A wrapper around a class marked with an {@link Plugin} annotation to retrieve
  * information from the annotation for easier use.
@@ -52,10 +57,19 @@ public interface PluginContainer {
     String getVersion();
 
     /**
-     * Returns the created instance of {@link Plugin}.
+     * Returns the assigned logger to this {@link Plugin}.
      *
-     * @return The instance
+     * @return The assigned logger
      */
-    Object getInstance();
+    default Logger getLogger() {
+        return LoggerFactory.getLogger(getId());
+    }
+
+    /**
+     * Returns the created instance of {@link Plugin} if it is available.
+     *
+     * @return The instance if available
+     */
+    Optional<Object> getInstance();
 
 }
--- a/src/main/java/org/spongepowered/api/Platform.java
+++ b/src/main/java/org/spongepowered/api/Platform.java
@@ -24,10 +24,12 @@
  */
 package org.spongepowered.api;
 
+import org.spongepowered.api.plugin.PluginContainer;
+
 import java.util.Map;
 
 /**
- * Represents a possible platform, or implementation, a {@link Game} could be
+ * Represents a possible platform or implementation a {@link Game} could be
  * running on.
  */
 public interface Platform {
@@ -51,32 +53,21 @@ public interface Platform {
     Type getExecutionType();
 
     /**
-     * Retrieves the current platform name.
+     * Returns the current API plugin container.
      *
-     * @return The platform name
+     * @return The API plugin container
      */
-    String getName();
+    PluginContainer getApi();
 
     /**
-     * Retrieves the current platform version.
+     * Returns the current implementation plugin container.
      *
-     * <p>This version can be in any format, like a build number or
-     * <a href="http://semver.org/">semantic versioning</a>.</p>
-     *
-     * @return The platform version, as a string
+     * @return The implementation plugin container
      */
-    String getVersion();
+    PluginContainer getImplementation();
 
     /**
-     * Retrieves the current Sponge API version that this platform is
-     * implementing.
-     *
-     * @return The API version
-     */
-    String getApiVersion();
-
-    /**
-     * Gets current Minecraft version of this platform.
+     * Gets the current Minecraft version of this platform.
      *
      * @return The Minecraft version
      */
@@ -145,7 +136,7 @@ public interface Platform {
          * @return False if the platform is {@link #UNKNOWN}, true otherwise
          */
         public boolean isKnown() {
-            return !(this == UNKNOWN);
+            return this != UNKNOWN;
         }
 
     }


src/main/java/org/spongepowered/api/plugin/PluginContainer.java
```diff
@@ -52,6 +54,13 @@
     String getVersion();
 
     /**
+     * Returns the assigned logger to this {@link Plugin}.
+     *
+     * @return The assigned logger
+     */
+    Logger getLogger();
```
JBYoshi: This probably shouldn't be made available to all plugins.

Zidane: Explain.

Minecrell: At the end they will be able to get it anyway using `LoggerFactory.getLogger(container.getId())` so it doesn't really make a difference.

JBYoshi: @Zidane A plugin that wanted to troll server admins could grab the logger of another plugin and log a bunch of errors as if they were caused by another plugin. However, now that I think about it, they could just use `LoggerFactory` (see @Minecrell's comment).

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
```diff
@@ -52,10 +57,19 @@
...
+    }
+
+    /**
+     * Returns the created instance of {@link Plugin} if it is available.
      *
-     * @return The instance
+     * @return The instance if available
      */
-    Object getInstance();
+    Optional<Object> getInstance();
```
Zidane: Not sure I agree with this change...if you have a PluginContainer, you have an instance in it.

Remember, we don't unload plugins or the like.

Minecrell: Keep in mind we also represent, e.g. Forge mod containers as plugins. Basically in this case there are two options, because some of the Forge containers don't have an instance assigned (currently the method returns null in some cases!):

1. Return a dummy instance or something like `Optional.empty()` as object.
2. Change the method to use an `Optional`

simon816: Every user-created mod or plugin would have an instance. If it's a dummy mod (e.g. the "Minecraft" mod) then could just return the mod container or the `Minecraft` instance or something.

Minecrell: That would still fall under (1) because Forge can simply not promise us that we won't get nulls there.

Minecrell: I'll change it back and will simply return the plugin container if Forge gives us no instance

Minecrell: The `/sponge plugins` command currently uses this method to display the plugin class. Without optional, this would always return the mod container class which would be also stupid. Any other opinions on this change?

ST-DDT: IMO we should not fake the plugin instance. If the plugin has no instance or we cannot obtain it we should simply tell the developer so using `Optional.empty()` then he can decide whether he uses the container instance or maybe utilize another API to fetch the plugin instance.

````java
container.getInstance().orElse(container);
````
or
````java
container.getInstance().orElseGet(() -> magicAPI.getPluginInstance(container.getId()));
````

Zidane: Again, Forge will not give you a container without an instance in it if it could not create the mod object. Meaning calling ``` getPluginManager().getPlugin(someId) ``` will ALWAYS return either an empty Optional or one containing a container with a mod instance inside except for one exception (I think, needs testing).

Exception being the construct event. At that point, FML has the list of mod candidates and creates the containers. After construct, all the containers have instances (or are removed if creating the mod object fails). I'm not 100% sure on the latter.

Should I be correct about the above, this means that, for a very short window, it could come back at null on the instance. I'm of the strong opinion that the API shouldn't be made "silly" just because the Forge side doesn't get its act together for a short time.

Minecrell: Forge doesn't guarantee that there is a instance attached for all mod containers. It provides several "dummy" mod containers which don't ever get an attached mod instance, e.g. the MCP container, the Minecraft container and even the FML and Minecraft Forge container.

Additionally the mod containers are implemented by core mods manually, so they could theoretically return anything there. I just checked a few core mods and only some of them actually provide the instance. Currently we even check in the `/sponge plugins` command [if the provided instance is null](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/command/CommandSponge.java#L382-L383), which is - in my opinion - definitely not a good way considering as per specification of our API it should never return null.

Minecrell: @Zidane Can I keep the optional now or do you still want it changed?

Deamon5550: How about returning dummies if we need to for things like the MCP container, and throwing an exception if we for some reason have a mod from forge which has no instance attached, since I think that case does qualify as `exceptional` eh?

Minecrell: @Deamon5550 It would be probably a bad idea to throw an exception for a lot of coremods. Basically this adds several ways the method can fail, while changing it to optional changes it to one way only. Consider the things we do in `/sponge plugins` (display the main class), it would be rather stupid to check for a) dummy containers and b) exceptions etc.

kashike: In my opinion, Optional is the way to go.

PR #940, opened 2015-11-29
Supply Abstract Data implementations for Plugin usability

gabizou:
With the stability of abstract `DataManipulator` implementation in SpongeCommon, I feel it's safe to now supply these abstractions for plugin usage. Little should change at this point.


Zidane:
@gabizou 

I want to see the generics get proper names and explained before merge since plugin devs do need to know what is going on here.

Same for the classes themselves.

PR #941, opened 2015-11-30
Add the ability to change the "affects spawning" state of Humanoids

kashike:
Impl: https://github.com/SpongePowered/SpongeCommon/pull/297

This adds the ability to prevent humanoids from affecting spawning (activating mob spawners, and allowing mobs to spawn and despawn).

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -25,7 +25,9 @@
 package org.spongepowered.api.data.key;
 
 import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.data.manipulator.mutable.entity.AffectsSpawningData;
 import org.spongepowered.api.data.manipulator.mutable.item.HideData;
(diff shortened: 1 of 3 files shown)


gabizou:
This needs updating to master.

kashike:
@gabizou updated to master, but it is still against `EntityPlayer` and not `EntityPlayerMP` since it is a `List<EntityPlayer>` in `World`.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -311,6 +311,7 @@
     public static final Key<Value<Integer>> FALL_TIME = null;
     public static final Key<Value<Boolean>> FALLING_BLOCK_CAN_HURT_ENTITIES = null;
     public static final Key<MutableBoundedValue<Double>> HEALTH_SCALE = null;
+    public static final Key<Value<Boolean>> AFFECTS_SPAWNING = null;
```
gabizou: Document this key please.

kashike: done

PR #944, opened 2015-12-02
Extend TargetPlayerEvent in ClientConnectionEvent.Login

kashike:
Common: https://github.com/SpongePowered/SpongeCommon/pull/302

It seems like `TargetPlayerEvent` was accidentally left out.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -4225,9 +4225,10 @@ public class SpongeEventFactory {
      * @param toTransform The to transform
      * @param connection The connection
      * @param profile The profile
+     * @param targetEntity The target entity
      * @return A new login client connection event
      */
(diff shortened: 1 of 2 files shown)


simon816:
The player entity doesn't exist when logging in. Use `ClientConnectionEvent.Join` instead

kashike:
Yes it does. https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/server/MixinServerConfigurationManager.java#L178

ryantheleach:
Does it have all it's properties set correctly at that time though?
https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/server/MixinServerConfigurationManager.java#L199

Zidane:
@kashike 

Unsure I agree with this change. Granted you are correct that the Player object is created at Login but you are not correct in the sense that the object is in a workable state.

Minecraft's code is really annoying here because many of the properties of EntityPlayer aren't set until afterwards in a variety of methods. Should the API make the statement that a Player object is available (via ``` TargetPlayerEvent ```) and a method called against this object ends up crashing/leading to instability, that is Sponge's fault and not the plugin developers.

Basically, with the API contract we have to make sure any calls against the object will behave the same way whether here or later on in the lifecycle.

gabizou:
We should not expose the `Player` object in the Login phase because there's various things where the object's state is in transition and would completely collapse if used. It's our mistake to include the `Player` object in the `Cause`, so we'll be changing that shortly.

PR #945, opened 2015-12-03
Data & entity contexts

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/945) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/304)

Depends on:
- [x] Profile properties
 - [x] https://github.com/SpongePowered/SpongeAPI/pull/974
 - [x] https://github.com/SpongePowered/SpongeCommon/pull/344
 - [x] https://github.com/SpongePowered/SpongeForge/pull/455
 - [x] https://github.com/SpongePowered/SpongeVanilla/pull/194
- [ ] [Fire `ClientConnectionEvent.Prepare`](https://github.com/SpongePowered/SpongeCommon/pull/408)
- [x] `CustomNameVislbleData`
 - [x] https://github.com/SpongePowered/SpongeAPI/pull/1013
 - [x] https://github.com/SpongePowered/SpongeCommon/pull/403

This pull request is still a WIP - I'm opening it for a place to get feedback while I'm writing it.

`Entity#getDefaultContext()` - gets the default ("global") context
`Entity#getContext(CommandSource)` - gets the context as seen by the `CommandSource`. `CommandSource` is a placeholder for now.

I'd be able to do, for example
```java
deamon5550.getContext(kashike).offer(Keys.DISPLAY_NAME, Texts.of("Deamon"));
```
to set the name I see you as ("Deamon") when querying `deamon5550.getContext(kashike).get(Keys.DISPLAY_NAME);`, but for the default context (and contexts with no data set) I'd see your primary name ("Deamon5550") unless changed.

Display names are an example, this is a requirement of other things I'm working on.


kashike:
I'm wondering what to do with [these methods](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/DataHolder.java#L38-L68). I can move them to a new interface, or add them to `DataSerializable`. Adding the methods to `DataSerializable` would require all extending classes to implement them, unless they are defaulted (sane defaults?).

Deamon5550:
I'm wondering what feedback you're actually looking for given that you have provided no description of what you're trying to do and only a single class with no docs and no context (heh).

kashike:
@Deamon5550 as the title says, this is a WIP. I'm asking questions here as I go (see comment above yours). :roller_coaster:

Deamon5550:
Yes but you have given no information as to why you are trying to move those methods or what you are actually trying to achieve.

kashike:
@Deamon5550 updated description, should help to explain

kashike:
As for the moving of the methods, `DataHolder` either needs to become non-serializable (move those methods somewhere else (`DataSerializable`?), and remove `DataSerializable` from extends). Data contexts should not be serializable, and changing `DataHolder` would be the easiest (and less breaking) solution.

jamierocks:
This is quite neat actually :+1:  (or will be once complete)

gratimax:
Not sure how useful getDefaultContext is, seems like that should just be standard data API, i.e. setting and getting on the entity itself

gabizou:
> I'm wondering what to do with these methods. I can move them to a new interface, or add them to DataSerializable. Adding the methods to DataSerializable would require all extending classes to implement them, unless they are defaulted (sane defaults?).

Those stay exactly where they are, moving them has nothing to do with this PR to be honest.

> As for the moving of the methods, DataHolder either needs to become non-serializable (move those methods somewhere else (DataSerializable?), and remove DataSerializable from extends). Data contexts should not be serializable, and changing DataHolder would be the easiest (and less breaking) solution.

I already told you beforehand, you can simply make `Context` extend `CompositeValueStore` since it is the `Context` that has the focus of the "data" associated. You can have it where `Entity` may extend `Contextualized` so that something like `InvisibilityData` can be handled through that, while `Player`'s `Context` would support possibly more data. I see little reason to change any part of the Data API.

kashike:
You told me that `Context` should **not** be serializable, but `DataHolder` extends `DataSerializable` (as well as the two methods mentioned above). `DataManipulator`, as well as others, require a `DataHolder` to be passed.

gabizou:
Again, extending `CompositeValueStore` will prevent it being `DataSerializable`. Granted, there's a couple of the methods that do use `DataHolder`, but mostly because they rely on a "serializable" notion that the `DataHolder` is serializable. I'm not at all sure what these "others" uses that don't have to do with serialization deal with `DataHolder`.

src/main/java/org/spongepowered/api/data/context/DataContext.java
```diff
@@ -0,0 +1,35 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.context;
+
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.value.mutable.CompositeValueStore;
+
+/**
+ * A context of data.
```
gabizou: Very descriptive.

src/main/java/org/spongepowered/api/entity/context/Contextual.java
```diff
@@ -0,0 +1,47 @@
...
+ * A collection of {@link EntityContext}s.
+ */
+public interface Contextual {
+
+    /**
+     * Gets the default context, with no specific viewer.
+     *
+     * @return The default context
+     */
+    EntityContext getContext();
```
Deamon5550: Why have a default context at all, using the standard dataholder methods should be the default context.

src/main/java/org/spongepowered/api/entity/context/EntityContext.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.spongepowered.api.entity.context;
+
+import org.spongepowered.api.data.context.DataContext;
+import org.spongepowered.api.entity.Entity;
+
+/**
+ * A {@link ContextViewer}'s context of an {@link Entity}.
+ */
+public interface EntityContext extends DataContext {
```
gabizou: Why have an explicit EntityContext? What does this do? How does this differ from a `DataContext`? When do I use this compared to a normal `DataContext`?

src/main/java/org/spongepowered/api/entity/context/Contextual.java
gabizou: Yeah.. there doesn't need to be a default context at all. Normal Data API methods should be considered the "default context".

src/main/java/org/spongepowered/api/entity/context/Contextual.java
```diff
@@ -0,0 +1,47 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.context;
+
+/**
+ * A collection of {@link EntityContext}s.
```
gabizou: Is this really a collection? I don't see any `List<DataContext> getContexts()` anywhere, let alone removing any.

src/main/java/org/spongepowered/api/entity/context/ContextViewer.java
```diff
@@ -0,0 +1,32 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.context;
+
+/**
+ * A viewer of an {@link EntityContext}.
```
gabizou: Why so specific to an entity?

src/main/java/org/spongepowered/api/data/context/ContextViewer.java
```diff
@@ -0,0 +1,32 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.context;
+
+/**
+ * A viewer of a {@link DataContext}.
```
gabizou: Could you clarify what this means?

src/main/java/org/spongepowered/api/command/CommandSource.java
```diff
@@ -34,7 +35,7 @@
  * <p>Examples of potential implementations include players, the server console,
  * Rcon clients, web-based clients, command blocks, and so on.</p>
  */
-public interface CommandSource extends Subject {
+public interface CommandSource extends ContextViewer, Subject {
```
gabizou: This might be a little interesting to handle, I'm not 100% sure that this is going to be easily implementable with command block sources trying to execute a command according to another player? Or is this intended for other use cases?

kashike: It's intended for a few use cases, most of which are going to be in another PR.

ST-DDT:
This also adds some nice usecases for #808 or at least you can nicely use the features proposed by this PR with the features added by my PR.

src/main/java/org/spongepowered/api/data/context/Contextual.java
```diff
@@ -0,0 +1,40 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.context;
+
+/**
+ * An object that has per-{@link ContextViewer} views of data.
+ */
+public interface Contextual {
```
gabizou: This needs to have a different name as we already have [`Contextual`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/permission/context/Contextual.java).

kashike:
I can get more done on this once #974 is merged.

src/main/java/org/spongepowered/api/entity/living/player/tab/TabList.java
```diff
@@ -70,45 +72,46 @@
...
+     * Adds an entry to the list.
      *
-     * @param playerId the UUID of the player to remove
-     * @return The {@link PlayerTabInfo} that was associated with the UUID
+     * @param builder The entry to add
+     * @throws IllegalArgumentException if an entry already with the same unique
+     *     id exists on the list
      */
-    PlayerTabInfo removePlayer(UUID playerId);
+    void addEntry(TabListEntry.Builder builder) throws IllegalArgumentException;
```
ST-DDT: Why adding the Builder instead of the real instance?

kashike: Instance must be created impl-side.

ST-DDT: Isn't that what the `build()` method is for?
If not because you need more data than you should not expose the build method at all or is there a different usecase for it?

src/main/java/org/spongepowered/api/context/NamedContextual.java
```diff
@@ -0,0 +1,50 @@
...
+ * Represents a contextual object with a name.
+ */
+public interface NamedContextual {
+
+    /**
+     * Gets the name identifying this object.
+     *
+     * @return The name of this object
+     */
+    String getName();
```
ST-DDT: Is this supposed to be human readable?

kashike: It's `String` intentionally, if that's what you're asking. `CommandSource` inherits from this method.

src/main/java/org/spongepowered/api/entity/living/player/tab/TabList.java
kashike: I can remove the `build(TabList)` method from the API, if that's what you're asking. There is no `build()` method.

src/main/java/org/spongepowered/api/context/NamedContextual.java
ST-DDT: That was not my question. I'm not referring to `Text` or things alike if that was your thought.

Is the result required to be human readable?

Yes or No?

kashike:
Updated description with pull requests this depends on.

gabizou:
@kashike Can you rebase and or resolve the conflicts and demonstrate to me how this will work now that the dependent PR's have been merged?

kashike:
Superseded by https://github.com/SpongePowered/SpongeAPI/pull/1046

PR #947, opened 2015-12-04
Implement decimal (double) parsing. Resolves #937

### LINKED ISSUE #937: Add a way to parse a double/float using GenericArguments
Currently, the only Java base types that the command arguments parser will parse are `Boolean`, `String`, and `Integer`. It would be very useful for some commands where very precise numbers are needed, such as tweaking `/flyspeed` to the most comfortable rate.

CrystalMageX:
This adds the much needed useful parsing for non-integer numbers to Sponge.

For further details, see #937

### DIFF
--- a/src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
@@ -691,6 +691,34 @@ public final class GenericArguments {
         }
     }
 
+    /**
+     * Require an argument to be an double-precision floating point number.
+     * Gives values of type {@link Double}
+     *
+     * @param key The key to store the parsed argument under
+     * @return the element to match the input
+     */
+    public static CommandElement decimal(Text key) {
+        return new DecimalElement(key);
+    }
+
+    private static class DecimalElement extends KeyElement {
+
+        private IntegerElement(Text key) {
+            super(key);
+        }
+
+        @Override
+        public Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            final String input = args.next();
+            try {
+                return Double.parseDouble(input);
+            } catch (NumberFormatException ex) {
+                throw args.createError(t("Expected an decimal, but input '%s' was not", input));
+            }
+        }
+    }
+
     private static final Map<String, Boolean> BOOLEAN_CHOICES = ImmutableMap.<String, Boolean>builder()
             .put("true", true)
             .put("t", true)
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
```diff
@@ -691,6 +691,34 @@ public Object parseValue(CommandSource source, CommandArgs args) throws Argument
...
+            super(key);
+        }
+
+        @Override
+        public Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            final String input = args.next();
+            try {
+                return Double.parseDouble(input);
+            } catch (NumberFormatException ex) {
+                throw args.createError(t("Expected an decimal, but input '%s' was not", input));
```
kashike: an -> a

kashike:
Perhaps `doubleDecimal` would be a better, and more descriptive, name?

src/main/java/org/spongepowered/api/util/command/args/GenericArguments.java
CrystalMageX: Fixed in b40732a77de51f700b95ce2dc769730ab42dd251

CrystalMageX:
The word `double` in `doubleDecimal` is what I'm trying to avoid from the beginning. It might confuse people that there are two (two=2=double) decimals. `doubleDecimal` is *quite awkward* to say. Perhaps just `double` if you would like a more descriptive name?

zml2008:
Call it double (or like doubleNum since reserved keywords) -- that's what it's called in java and making up another naming scheme is dumb.

ryantheleach:
Especially since Decimal in Java generally refers to "accurate math" decimals such as http://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html

CrystalMageX:
We have these names to choose from:

- `decimal` (original)
- `decimal` (and use `BigDecimal` instead)
- `doubleDecimal` (suggested by @kashike)
- `double_` (keeps the word "double" intact)
- `doubleNum` (suggested by @zml2008)
- `doubl` (shortest one)

Which one should we choose?

kashike:
`doubleNum` seems like a good choice.

ST-DDT:
I agree `doubleNum` is a good choice. 

If there will ever be a usecase for `BgDecimal` then we can add it using `decimal` without conflicts

CrystalMageX:
I've changed it to `doubleNum`, by consensus of @kashike , @ST-DDT , and @zml2008 .

For the error description, however, it might be too technical for **players to know** what a `double` even is, so I've left it as "expected a *decimal*" for now.

zml2008:
imo should be number for double and integer/whole number for int (integer is a math term, not java-specific)

&lt;/bikeshed>

CrystalMageX:
How about **"whole number"** for **integer** parsing, and **"whole number or decimal"** for **double** parsing?

PR #950, opened 2015-12-07
Add OptionSubjectCollection

CrystalMageX:
We need an OptionSubjectCollection when a SubjectCollection needs to be ensured to have OptionSubjects.


CrystalMageX:
Closed - Might as well try to combine Subject, and OptionSubject.

PR #951, opened 2015-12-08
GameDictionary revamp

### LINKED ISSUE #806: Fix GameDictionary
There is a slight problem with the `GameDictionary` API. In Forge, an ore is represented by an `ItemStack`, while in SpongeAPI, only the `ItemType` is relevant. This can lead to some confusion; for example, an item stack containing an ink sack would pass for "gemLapis". They have the same `ItemType`, but not the same data (dye color). This should be fixed.

JBYoshi:
**SpongeAPI PR** | [SpongeCommon PR](https://github.com/SpongePowered/SpongeCommon/pull/314) | [SpongeForge PR](https://github.com/SpongePowered/SpongeForge/pull/439)

Originally, `GameDictionary` only provides access to `ItemType`s. This has the problem of false positives when performing comparisons using it, which can lead to **duplication bugs** (see #806). This PR changes it to store `GameDictionaryEntry`s instead. `GameDictionaryEntry`s can come from `ItemType` (which now extends it) or from a method in `ItemStackSnapshot`, and they provide methods to inspect their data and check whether an item stack matches them.

This change may break plugins.

Fixes #806.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/ItemType.java
+++ b/src/main/java/org/spongepowered/api/item/ItemType.java
@@ -24,10 +24,14 @@
  */
 package org.spongepowered.api.item;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.GameDictionary;
 import org.spongepowered.api.block.BlockType;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 
@@ -37,7 +41,7 @@ import java.util.Optional;
  * A type of item.
  */
 @CatalogedBy(ItemTypes.class)
-public interface ItemType extends CatalogType, Translatable {
+public interface ItemType extends CatalogType, Translatable, GameDictionary.Entry {
 
     /**
      * Gets the corresponding {@link BlockType} of this item if one exists.
@@ -80,5 +84,25 @@ public interface ItemType extends CatalogType, Translatable {
      * @return The item property, if available
      */
     <T extends Property<?, ?>> Optional<T> getDefaultProperty(Class<T> propertyClass);
+    
+    @Override
+    default ItemType getType() {
+        return this;
+    }
+    
+    @Override
+    default boolean matches(ItemStack stack) {
+        return checkNotNull(stack, "stack").getItem().equals(this);
+    }
+
+    @Override
+    default boolean isSpecific() {
+        return false;
+    }
+
+    @Override
+    default ItemStackSnapshot getTemplate() {
+        return ItemStack.of(this, 1).createSnapshot();
+    }
 
 }
--- a/src/main/java/org/spongepowered/api/GameDictionary.java
+++ b/src/main/java/org/spongepowered/api/GameDictionary.java
@@ -24,44 +24,90 @@
  */
 package org.spongepowered.api;
 
+import com.google.common.collect.SetMultimap;
 import org.spongepowered.api.item.ItemType;
-import org.spongepowered.api.item.ItemTypes;
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 
-import java.util.Map;
 import java.util.Set;
 
 /**
- * A GameDictionary is a store of {@link ItemTypes}.
+ * A GameDictionary is a store of {@link GameDictionary.Entry}s.
  *
- * <p>Note that the GameDictionary's keys are different from Minecraft item
- * ids. Minecraft item IDs are namespaces, e.g. minecraft:carrot while
- * ItemDictionary keys are not, by design(e.g. carrot). This is mainly to keep
- * supporting the existing Forge 'ore dictionary'.</p>
+ * <p>Note that the GameDictionary's keys are different from Minecraft item ids.
+ * Minecraft item IDs are namespaces, e.g. minecraft:carrot while ItemDictionary
+ * keys are not, by design(e.g. carrot). This is mainly to keep supporting the
+ * existing Forge 'ore dictionary'.</p>
  */
 public interface GameDictionary {
 
     /**
-     * Registers an ItemType in the dictionary with a String key.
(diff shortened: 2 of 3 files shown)


JBYoshi:
Ready for review.

src/main/java/org/spongepowered/api/GameDictionaryEntry.java
```diff
@@ -0,0 +1,75 @@
...
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+
+/**
+ * Represents an object that can be added to the {@link GameDictionary}.
+ * 
+ * @see {@link ItemType}
+ * @see {@link ItemStackSnapshot#createGameDictionaryEntry()}
+ */
+public interface GameDictionaryEntry {
```
gabizou: Would it make sense to make this an inner class of `GameDictionary`?

src/main/java/org/spongepowered/api/item/ItemType.java
```diff
@@ -80,5 +82,15 @@
      * @return The item property, if available
      */
     <T extends Property<?, ?>> Optional<T> getDefaultProperty(Class<T> propertyClass);
+    
+    @Override
+    default ItemType getType() {
```
Zidane: Not fond of this name as it means myType.getType()

Maybe, getGameDictionaryType()? or...

liach:
Will `GameDictionary` be applicable on other things (e.g. biome, entity) in the future? May it be applied to all `CatalogType`s?

JBYoshi:
@liach This is intended to be a Sponge wrapper around [Forge's ore dictionary](http://www.minecraftforge.net/wiki/How_to_use_the_ore_dictionary). So it's only available for items and item data.

liach:
Then why not put it in the `org.spongepowered.api.util` package? It might be useful for everything, since different mods are going to add duplicate things. We can choose not to implement it when we do not need to.

JBYoshi:
@liach Like I said, this is a wrapper around the Forge ore dictionary, which only supports items.

src/main/java/org/spongepowered/api/GameDictionaryEntry.java
JBYoshi: @gabizou Possibly. I wasn't sure since it has an inner class of its own (`GameDictionaryEntry.Specific`).

src/main/java/org/spongepowered/api/item/ItemType.java
JBYoshi: @Zidane I agree, but I don't know of any better name (other than `getItemType()`, which would also cause confusion). IMHO `getGameDictionaryType()` seems weird because it's returning a type of item, not a type of game dictionary.

liach:
Why cannot this idea extend to everything applicable to registries? Because "Forge had it we just do a forge compat"?

src/main/java/org/spongepowered/api/item/ItemType.java
gabizou: Maybe getDefinition()?

ST-DDT: Or `getBaseType()`.
Also I'm not sure whether the return type is right.
Maybe this should Optional.empty for BaseTypes.

Zidane: @gabizou 's name would be fine...

src/main/java/org/spongepowered/api/item/ItemType.java
```diff
@@ -37,7 +39,7 @@
  * A type of item.
  */
 @CatalogedBy(ItemTypes.class)
-public interface ItemType extends CatalogType, Translatable {
+public interface ItemType extends CatalogType, Translatable, GameDictionaryEntry {
```
liach: Should `ItemType` implement `GameDictionaryEntry` or just provide a method for it?

src/main/java/org/spongepowered/api/item/ItemType.java
JBYoshi: I thought of an idea that might fix this. See below.

src/main/java/org/spongepowered/api/item/ItemType.java
JBYoshi: Good point. Posting a comment below to discuss it.

JBYoshi:
Some of the comments seem to relate to `ItemType` extending `GameDictionaryEntry`. I'm thinking a simple solution to those comments would be to make a separate interface `OfItemType` (name subject to change) extending `GameDictionaryEntry` that wraps an `ItemType`. `ItemType` would then have a method `asGameDictionaryEntry()` that would return an instance of that interface, and calling `getType()` would convert it back to an `ItemType`.

Thoughts?

ST-DDT:
Why the extra interface?
IMO every GDEntry has an ItemType that should be obtainable.

> `ItemType` would then have a method `asGameDictionaryEntry()` that would return an instance of that interface,

I would rather use the path via `GameDictionary.from(ItemType)` because Vanilla types dont know anything about a GDEntry.

gabizou:
I would have to assume that `ItemType.getDefinition()` would instead return `GameDefinition<ItemType>` where `GameDefinition.getType()` would return the generic. Am I understanding correctly?

JBYoshi:
@gabizou What other values of the type parameter would be used?

@ST-DDT Originally, the purpose for the extra interface was to clarify that the `GameDictionaryEntry` only compares based on the item type. Now that I think about it, however, `getTemplate()` may be useful for any `GameDictionaryEntry`, allowing there to be only a single interface for entries (with the `instanceof` check being replaced by an `isSpecific()` method).

src/main/java/org/spongepowered/api/item/ItemType.java
```diff
@@ -80,5 +83,25 @@
...
     <T extends Property<?, ?>> Optional<T> getDefaultProperty(Class<T> propertyClass);
+    
+    @Override
+    default ItemType getType() {
+        return this;
+    }
+    
+    @Override
+    default boolean matches(ItemStack stack) {
+        return Preconditions.checkNotNull(stack, "stack").getItem().equals(this);
```
simon816: `Preconditions` should be imported statically

JBYoshi: Fixed.

JBYoshi:
Status?

JBYoshi:
It would be nice to get this in before the end of breaking changes.

PR #956, opened 2015-12-12
Added HideData, and corresponding keys. Adds #905

me4502:
Added the HideData manipulator, the corresponding keys, and the Immutable version of the manipulator.

Adds https://github.com/SpongePowered/SpongeAPI/issues/905

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -25,6 +25,8 @@
 package org.spongepowered.api.data.key;
 
 import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.data.manipulator.mutable.item.HideData;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -751,4 +753,52 @@ private Keys() {}
      */
     public static final Key<Value<Boolean>> IS_SPLASH_POTION = null;
 
+    /**
+     * Represents the {@link Key} for representing the "enchantments hidden"
+     * state of a {@link ItemStack}
+     *
+     * @see HideData#hideEnchantments()
+     */
+    public static final Key<Value<Boolean>> HIDE_ENCHANTMENTS = null;
```
kashike: Perhaps `HIDE_ITEMSTACK_*` would be more descriptive names?

PR #958, opened 2015-12-12
Add missing setters to Achievement + Statistic builders

ST-DDT:
[[API]](https://github.com/SpongePowered/SpongeAPI/pull/958) [[Common]](https://github.com/SpongePowered/SpongeCommon/pull/131)

Required changes to the builders to be properly implemented in [SpongeCommon#131](https://github.com/SpongePowered/SpongeCommon/pull/131)

---

There is also another issue with the `AchievementBuilder`

The [`Achievement`](https://github.com/SpongePowered/SpongeCommon/pull/131/files#diff-d2caa86494b4868deeba86211da08ccbR117) needs a `row` (`displayRow`) and a `column` (`displayColumn`) parameter however I'm not sure whether this is only an implementation detail or whether this is actually relevant to use them. I haven't found a usage of those fields/values, but this might be the case because those values are only needed for the client or the obfuscator trolled me.

Shall I add them here too or not?

### DIFF
--- a/src/main/java/org/spongepowered/api/statistic/Statistic.java
+++ b/src/main/java/org/spongepowered/api/statistic/Statistic.java
@@ -78,9 +78,9 @@ public interface Statistic extends CatalogType, Translatable {
     interface StatisticBuilder<T extends Statistic, B extends StatisticBuilder<T, B>> extends ResettableBuilder<T, B> {
 
         /**
-         * Sets the internal name for the {@link Statistic}.
+         * Sets the human readable name for the {@link Statistic}.
          *
-         * @param name The name of this achievement
+         * @param name The name for the statistic
          * @return This builder, for chaining
          */
         B name(String name);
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/statistic/Statistic.java
```diff
@@ -78,9 +78,17 @@ static Builder builder() {
...
 
         /**
-         * Sets the internal name for the {@link Statistic}.
+         * Sets the unique id for the {@link Statistic}.
          *
-         * @param name The name of this achievement
+         * @param id The unique for the statistic
+         * @return This builder, for chaining
+         */
+        Builder id(String id);
```
Zidane: Build method needs to take the id.

ST-DDT: Fixed

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -154,10 +165,11 @@ static Builder builder() {
         /**
          * Builds and registers an instance of an {@link Achievement}.
          *
+         * @param id The unique for the achievement
```
liach, line 168: `@param id The unique id for the achievement`

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -154,10 +165,11 @@ static Builder builder() {
...
          * @return A new instance of a achievement
```
liach, line 169: `@return A new instance of an achievement`

liach:
How is the status now?

ST-DDT:
Superseded by https://github.com/SpongePowered/SpongeAPI/commit/f2daea62080131bc3e1ee4bbb0ce2007ef215764

Issue #882, opened 2015-10-20
Add support for generating all missing chunks in a world within WorldBorder


bloodmc:
Currently, there is no way for a plugin to "pre-generate" their world in order to improve performance for players. I suggest we add an option similar to the WorldBorder plugin for Bukkit that allowed a plugin to fill all missing chunks within a WorldBorder. More info can be found here
http://dev.bukkit.org/bukkit-plugins/worldborder/pages/commands-and-permissions/#w-the-fill-command

@DDoS can you add support for this?

JBYoshi:
Might not be a good idea, though... unless you change the border size, you'll end up having to generate 14 *trillion* chunks.

Azatik:
Good Idea.

@JBYoshi Adequate administrator will not allow this.

liach:
How about the function of removing existing chunks far enough that will not be loaded when the player is at the border?

bloodmc:
@JBYoshi As @Azatik stated, the intended use is for admins that know what they are doing by setting a reasonable WorldBorder limit. A server that has no limit is asking for trouble.

DDoS:
I'll try and make this non-blocking with proper output to the console so the admin can shutdown the server normally, stoping the process, if he accidentally uses a border that is too large.

boformer:
Isn't this something that could be handled by a plugin?

All that is needed is a ``generateChunk(x,z)`` method.

bloodmc:
Many things in our API can be handled by a plugin but that doesn't mean it shouldn't be added.

PR #959, opened 2015-12-15
API for pre-generating chunks inside a world border

DDoS:
For #882, implemented in https://github.com/SpongePowered/SpongeCommon/pull/328

@bloodmc

### DIFF
--- a/src/main/java/org/spongepowered/api/world/WorldBorder.java
+++ b/src/main/java/org/spongepowered/api/world/WorldBorder.java
@@ -25,6 +25,13 @@
 package org.spongepowered.api.world;
 
 import com.flowpowered.math.vector.Vector3d;
+import org.slf4j.Logger;
+import org.spongepowered.api.Game;
+import org.spongepowered.api.scheduler.Scheduler;
+import org.spongepowered.api.scheduler.Task;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import javax.annotation.Nullable;
 
 /**
  * A world border is a square boundary, extending through the entire y-axis.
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/world/WorldBorder.java
```diff
@@ -194,4 +201,109 @@
...
+         *
+         * <p>Optional.</p>
+         *
+         * @param logger A logger for the generator
+         * @return This for chained calls
+         */
+        ChunkPreGenerate logger(@Nullable Logger logger);
+
+        /**
+         * <p>Sets the interval between generation runs.</p>
```
ST-DDT: IMO the first sentence does not need to encapsulated in `<p>`.

src/main/java/org/spongepowered/api/world/WorldBorder.java
```diff
@@ -194,4 +201,109 @@
      * @param damage The damage amount
      */
     void setDamageAmount(double damage);
+
+    /**
+     * Returns a new builder for creating a task to pre-generate the chunks
+     * inside the border. The current state of the border is "snapshoted" for
+     * the builder. It is not backed by this instance. If the border size is
+     * changing then the target size is used.
```
ST-DDT: the new or the old target size?

DDoS: The target is by definition the new size. The "target" as in "the size that is going to be reached".

DDoS:
Can I merge this and the implementation tonight?

PR #961, opened 2015-12-15
Add ability to get/set forced bed location. Resolves #955

### LINKED ISSUE #955: Add a way to force non-bed spawns with RespawnLocationData
Hello all, I've been working on a Graveyards plugin and have noticed that currently with [RespawnLocationData](https://jd.spongepowered.org/org/spongepowered/api/data/manipulator/mutable/entity/RespawnLocationData.html), there is no way to force a non-bed spawn location.  Whenever I offer a non-bed location it is rejected and the player is told that their home bed was missing or obstructed and they are respawned at the default spawn point in the world.  I've been told that forge had a force spawn position flag, is there anyway would could implement a similar feature in sponge?

Thanks!

simon816:
Changes `Vector3d` to `Tuple<Vector3d, Boolean>`
See #955


simon816:
@gabizou do you agree this should be represented as a `Tuple<Vector3d, Boolean>` or not?

simon816:
This PR is now rebased on the `bleeding` branch

Minecrell:
I'd personally open a new PR here, this is pretty much unreadable now because you don't see the actual file changes.

kashike:
+1 to the above, this is just bad to look at at this point

PR #978, opened 2015-12-21
Add dual setter for header and footer, in addition to single setters

kashike:
This adds a new dual (header and footer) setter for the tablist header and footer. Since the header and footer are sent together, it makes sense to allow setting both at once, as well as individually.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/tab/TabList.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/tab/TabList.java
@@ -65,6 +65,17 @@ public interface TabList {
     void setFooter(Text footer);
 
     /**
+     * Sets this list's header and footer.
+     *
+     * @param header The new header
+     * @param footer The new footer
+     */
+    default void setHeaderAndFooter(Text header, Text footer) {
+        this.setHeader(header);
+        this.setFooter(footer);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/entity/living/player/tab/TabList.java
```diff
@@ -65,6 +65,14 @@
     void setFooter(Text footer);
 
     /**
+     * Sets this list's header and footer.
+     *
+     * @param header The new header
+     * @param footer The new footer
+     */
+    void setHeaderAndFooter(Text header, Text footer);
```
gabizou: Can't this be a default method?

kashike: The purpose of this method is to prevent calling both individually. In the implementation, the packet is sent with both the header and footer - to send just a header or footer will involve sending an empty text for the opposite, however with this method both will be sent together.

gabizou: Still could be a default method.

kashike: Calling `setHeader` and `setFooter` in a default method like this would be exactly what this method is trying **to avoid**.

PR #980, opened 2015-12-21
Refactor message sinks and message events

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/issues/980) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/349) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/458) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/195)

This PR:
- changes `transformMessage` to return an `Optional<Text` instead of a nullable `Text`, and removes `final` from `sendMessage` to allow for more customisability
- changes `MessageEvent` to return an `Optional<Text>` for `getMessage`, and marks `setMessage` as nullable
 - this additionally cleans up the implementation, which is currently checking if an empty text is being sent - this is bad because sending an empty text is valid
- prevents sending null values to the recipient in `MessageSink`


Zidane:
I disagree on the MessageEvent change for original message...it shouldn't be fired if the original message does not exist (then what is the point of the event ??)

Minecrell:
@Zidane `DestructEntityEvent` is also a message event for example, but called without message because there is no message by default (plugins could set one though): https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/world/MixinWorld.java#L518-L535

Zidane:
@Minecrell 

Then send an empty Text object. Don't violate the contract of the event. We can easily check for empty text objects and not send them to the output.

ST-DDT:
Sending an empty line is different from sending nothing and in some cases you may want to send an empty line that could be enriched with the senders name in the message sink.

However I disagree with the `Optional`/`@Nullable` stuff in the `MessageSink`s. If the event has a null `Text` then the `MessageSink` shouldn't be called at all, because it shouldn't send a message in that case at all. Since setting the message to null in the event effectively disables/cancels the `MessageEvent` part and setting it to not null will enable the part.

````java
Optional<Text> text = event.getMessage();
MessageSink sink = event.getMessageSink();
text.ifPresent(sink::sendMessage);
````

**Comment by kashike** [comment]
@Zidane Sending an empty text is a valid thing to do. An **empty** text isn't the same as **no** message.

There are other cases where the original text can be null, most notably if a Forge mod/etc callsa Forge event with a null value. Leaving `MessageEvent`'s `getOriginalMessage` method as `@Nonnull` would not be accurate.

**Comment by kashike** [comment]
@ST-DDT I reverted the `@Nullable` changes in `MessageSink` - I forgot to earlier.

File: src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java [hunk 1]
```diff
@@ -40,21 +44,21 @@
      *
      * @return The original message sink to send to
      */
-    MessageSink getOriginalSink();
+    Optional<MessageSink> getOriginalSink();
```
Minecrell: Why do you also make the message sink optional? Unlike for Text there is `MessageSinks.toNone()` which should represent the same.

src/main/java/org/spongepowered/api/text/sink/MessageSink.java
```diff
@@ -24,34 +24,42 @@
...
+     * @param original The original text to send
      */
-    public final void sendMessage(Text text) {
+    public void sendMessage(Text original) {
+        checkNotNull(original, "original text");
         for (CommandSource recipient : getRecipients()) {
-            Text transformed = transformMessage(recipient, text);
-            recipient.sendMessage(transformed == null ? text : transformed);
+            Optional<Text> transformed = transformMessage(recipient, original);
+            if (transformed.isPresent()) {
```
ST-DDT: You can simplyfy this using
````java
transformed.ifPresent(recipient::sendMessage);
````

File: src/main/java/org/spongepowered/api/text/sink/MessageSink.java [hunk 3]
```diff
@@ -24,34 +24,42 @@
...
+
 import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
 
 /**
  * Represents a function that takes a message and transforms it for distribution to the given targets.
  */
 public abstract class MessageSink {
```
ST-DDT: This class could be turned into a functional interface since there are neither fields nor final methods anymore.
````java
event.setMessageSink(server::getOnlinePlayers);
````

File: src/main/java/org/spongepowered/api/text/sink/MessageSink.java (hunk 2 above)
Comment (kashike) [inline src/main/java/org/spongepowered/api/text/sink/MessageSink.java]: Thanks.

File: src/main/java/org/spongepowered/api/text/sink/MessageSink.java (hunk 3 above)
Comment (kashike) [inline src/main/java/org/spongepowered/api/text/sink/MessageSink.java]: Thanks.

File: src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java (hunk 1 above)
Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java]: I agree with @Minecrell here. `MessageSink`s shouldn't be `Optional`.

Comment (kashike) [inline src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java]: In the middle of changing impl in able to make it non-`Optional`.

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java]: ^^

Comment (kashike) [inline src/main/java/org/spongepowered/api/event/command/MessageSinkEvent.java]: @ST-DDT done

File: src/main/java/org/spongepowered/api/text/sink/CombinedSink.java [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+            text = sink.transformMessage(recipient, text).orElse(text);
+        }
+
+        return Optional.ofNullable(text);
+    }
+
+    @Override
+    public Iterable<CommandSource> getRecipients() {
+        return ImmutableSet.copyOf(StreamSupport.stream(this.contents.spliterator(), false)
+                .filter(sink -> sink != null)
```
Minecrell: Why would a sink here be null? Not sure if that should be even valid.

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
```diff
@@ -0,0 +1,61 @@
...
+                .map(MessageSink::getRecipients)
+                .flatMap(sources -> StreamSupport.stream(sources.spliterator(), false))
```
Minecrell: You can skip the map above and do this directly in your flatMap here:
`sink -> ...(sink.getRecipients...))`

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
```diff
@@ -0,0 +1,61 @@
...
+                .collect(Collectors.toList()));
```
ST-DDT: A set would probably be a better return value to avoid duplicates.

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        return Optional.ofNullable(text);
+    }
+
+    @Override
+    public Iterable<CommandSource> getRecipients() {
+        return ImmutableSet.copyOf(StreamSupport.stream(this.contents.spliterator(), false)
+                .flatMap(sink -> StreamSupport.stream(sink.getRecipients().spliterator(), false))
+                .collect(Collectors.toSet()));
```
ST-DDT: IMO we should use a `ImmutableSet` builder collector here.

Just push something like this in a util class like `Functional`, so everyone can use it:

````java
public static <T> Collector<T, ImmutableSet.Builder<T>, ImmutableSet<T>> immutableSetCollector() {
    return Collector.of(ImmutableSet::builder, ImmutableSet.Builder::add,
            (left, right) -> left.addAll(right.build()), ImmutableSet.Builder::build,
            Collector.Characteristics.UNORDERED);
}
````

And then you can use it like this:
````java
return StreamSupport.stream(this.it.spliterator(), false)
        .flatMap(sink -> StreamSupport.stream(sink.getRecipients().spliterator(), false))
        .collect(immutableSetCollector());
````

File: src/main/java/org/spongepowered/api/text/sink/FixedSink.java [hunk 8]
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+public class FixedSink implements MessageSink {
+
+    protected final Set<CommandSource> recipients;
+
+    public FixedSink(Set<CommandSource> provided) {
```
ST-DDT: You should allow using `Set<? extends CommandSource>` here

src/main/java/org/spongepowered/api/text/sink/PermissionSink.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public Iterable<CommandSource> getRecipients() {
+        PermissionService service = Sponge.getGame().getServiceManager().provideUnchecked(PermissionService.class);
+
+        return service.getKnownSubjects().values().stream()
+                .flatMap(input -> input.getAllWithPermission(this.permission).entrySet().stream()
+                        .filter(Map.Entry::getValue)
+                        .map(entry -> entry.getKey().getCommandSource().orElse(null))
+                        .filter(source -> source != null))
+                .collect(Collectors.toList());
```
ST-DDT: This should be a `Set` to avoid duplicate entries and the order is really not important here.

src/main/java/org/spongepowered/api/event/message/MessageSinkEvent.java
```diff
@@ -47,14 +49,14 @@
      *
      * @return The message sink the message in this event will be sent to
      */
-    MessageSink getSink();
+    Optional<MessageSink> getSink();
```
ST-DDT: Wasn't there a discussion to keep this non-`Optional`?

https://github.com/SpongePowered/SpongeAPI/pull/980#discussion_r48228484

Or was there a side discussion I don't not know about?

kashike: `getSink` is different from `getOriginalSink`.

ST-DDT: Well IMO the same arguments apply here as well.

And for the possible argument "returns Optional.empty if not changed" IMO that only adds additional noise when using sinks and has no benefits at all. it even reduces the clarity of the method since one could asume the message shouldnt be send all of the sink is absent.
There is always a Sink the message will be broadcasted with and the magic value null/Optional.empty just hides this fact.
However if you keep the null stuff then you should at least explain in the javadocs what the values are supposed to express.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
```diff
@@ -0,0 +1,51 @@
...
+    public TransformingSink(MessageSink delegate, BiFunction<CommandSource, Optional<Text>, Optional<Text>> function) {
+        super(delegate);
+        this.function = checkNotNull(function, "function");
+    }
+
+    @Override
+    public Optional<Text> transformMessage(CommandSource recipient, Text original) {
+        Optional<Text> text = this.delegate.transformMessage(recipient, original);
+
+        return this.function.apply(recipient, text);
```
ST-DDT: You should check for null returns from the function here.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
+import java.util.function.BiFunction;
+
+public class TransformingSink extends ForwardingSink {
```
ST-DDT: IMO this class is perfectly suited to be used in a new default method in `MessageSink`.

````java
default MessageSink  transformMessage(BiFunction);
````

File: src/main/java/org/spongepowered/api/text/sink/TransformingSink.java [hunk 13]
```diff
@@ -0,0 +1,51 @@
...
+    protected final BiFunction<CommandSource, Optional<Text>, Optional<Text>> function;
+
+    public TransformingSink(MessageSink delegate, BiFunction<CommandSource, Optional<Text>, Optional<Text>> function) {
+        super(delegate);
+        this.function = checkNotNull(function, "function");
+    }
+
+    @Override
+    public Optional<Text> transformMessage(CommandSource recipient, Text original) {
+        Optional<Text> text = this.delegate.transformMessage(recipient, original);
```
ST-DDT: Use an `orElse(original)` here (to match CombinedSink's logic) and change the function specs accordingly.

src/main/java/org/spongepowered/api/text/sink/MessageSinks.java
```diff
@@ -24,16 +24,29 @@
...
 public class MessageSinks {
 
-    private static final MessageSinkFactory factory = null;
+    public static final MessageSink TO_NONE = ImmutableSet::of;
+    public static final MessageSink TO_ALL = () -> {
+        Set<CommandSource> recipients = Sets.newHashSet(Sponge.getGame().getServer().getOnlinePlayers());
+        recipients.add(Sponge.getGame().getServer().getConsole());
+        return ImmutableSet.copyOf(recipients);
+    };
+    public static final MessageSink TO_ALL_PLAYERS = () -> ImmutableSet.copyOf(Sponge.getGame().getServer().getOnlinePlayers());
```
ST-DDT: `TO_CONSOLE` would be a nice addition here.

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        return Optional.ofNullable(text);
+    }
+
+    @Override
+    public Iterable<CommandSource> getRecipients() {
+        return ImmutableSet.copyOf(StreamSupport.stream(this.sinks.spliterator(), false)
+                .flatMap(sink -> StreamSupport.stream(sink.getRecipients().spliterator(), false))
+                .collect(Collectors.toSet()));
```
ST-DDT: IMO we should use a `ImmutableSet` builder collector here.

Just push something like this in a util class like `Functional`, so everyone can use it:

````java
public static <T> Collector<T, ImmutableSet.Builder<T>, ImmutableSet<T>> immutableSetCollector() {
    return Collector.of(ImmutableSet::builder, ImmutableSet.Builder::add,
            (left, right) -> left.addAll(right.build()), ImmutableSet.Builder::build,
            Collector.Characteristics.UNORDERED);
}
````

And then you can use it like this:
````java
return StreamSupport.stream(this.it.spliterator(), false)
        .flatMap(sink -> StreamSupport.stream(sink.getRecipients().spliterator(), false))
        .collect(immutableSetCollector());
````

Or would that be overkill here?

File: src/main/java/org/spongepowered/api/text/sink/MessageSinks.java [hunk 16]
```diff
@@ -24,16 +24,29 @@
...
  * Useful building blocks for message sinks.
  */
 public class MessageSinks {
 
-    private static final MessageSinkFactory factory = null;
+    public static final MessageSink TO_NONE = ImmutableSet::of;
+    public static final MessageSink TO_ALL = () -> {
+        Set<CommandSource> recipients = Sets.newHashSet(Sponge.getGame().getServer().getOnlinePlayers());
+        recipients.add(Sponge.getGame().getServer().getConsole());
+        return ImmutableSet.copyOf(recipients);
```
ST-DDT: If you return a `ImmutableSet` here anyway then you could just use the `ImmutableSet.Builder` here.

````java
final ImmutableSet.Builder<CommandSource> builder = ImmutableSet.builder();
        final Server server = Sponge.getGame().getServer();
        builder.addAll(server.getOnlinePlayers());
        builder.add(server.getConsole());
        return builder.build();
````
or
````java
final Server server = Sponge.getGame().getServer();
return ImmutableSet.builder()
        .addAll(server.getOnlinePlayers())
        .add(server.getConsole())
        .build();
````

File: src/main/java/org/spongepowered/api/text/sink/MessageSinks.java [hunk 17]
```diff
@@ -83,16 +69,29 @@ public static MessageSink toNone() {
...
     }
 
     /**
      * Get a message sink that targets the given sources.
      *
-     * @param sources The sources to have as recipients
+     * @param recipients The sources to have as recipients
+     * @return The sink
+     */
+    public static MessageSink fixed(Collection<? extends CommandSource> recipients) {
```
ST-DDT: An array version of this one would be nice.

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+public class CombinedSink implements MessageSink {
+
+    protected final Iterable<MessageSink> sinks;
```
ST-DDT: IMO you should change this one to a `Collection` or is there a special reason why you used a `Iterable` here?

(This would simplify the things in `getRecipients()` a bit.)

ST-DDT:
I updated my comments with new alternatives, that better utilize Java 8 features.

src/main/java/org/spongepowered/api/text/sink/CombinedSink.java
Cybermaxke: There is already a `Collector` provided for `ImmutableSet`s in the class `GuavaCollectors`, just use `GuavaCollectors.toImmutableSet`.

src/main/java/org/spongepowered/api/event/message/MessageSinkEvent.java
kashike: Using `Optional<MessageSink>` prevents any processing, whereas `MessageSinks.NONE` would still process for each recipient.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
kashike: It returns an `Optional<Text>`, I'm not sure a null check is really required here (it isn't anywhere else).
It accepts, and returns, `Optional<Text>` so that nulls are not passed around.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
kashike: It is it's own type purposefully, it's a forwarding sink. It takes the sink you provide in the ctor to get the recipients, as well as running the message through `transformMessage` on the parent before the `transformMessage` method on the transforming sink.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
kashike: This sink is different from `CombinedSink`, it doesn't work that way.

src/main/java/org/spongepowered/api/text/sink/MessageSinks.java
kashike: Agreed.

kashike:
@ST-DDT Give me a poke when you're in #spongedev again.

src/main/java/org/spongepowered/api/event/message/MessageSinkEvent.java
ST-DDT: > `MessageSinks.NONE` would still process for each recipient.

`MessageSinks.NONE` is defined as no recipients thus there is no processing.

src/main/java/org/spongepowered/api/text/sink/TransformingSink.java
ST-DDT: Sorry, I made myself not clear enough:

````java
interface MessageSink {

    default MessageSink transformMessage(BiFunction function) {
        return new TransformingSink(this, function);
    }

}
````

Usage:
````java
event.setSink(event.getSink().transformMessage((target, message) -> Text.of("@", target, " : " + message)));
````

File: src/main/java/org/spongepowered/api/event/message/MessageSinkEvent.java (hunk 10 above)
Comment (kashike) [inline src/main/java/org/spongepowered/api/event/message/MessageSinkEvent.java]: I'd still prefer to keep this `Optional`.

**Comment by kashike** [comment]
Big thanks to @ST-DDT for their help with this.

**Comment by Minecrell** [comment]
There are quite a lot of Javadocs missing here, would you mind adding them?

File: src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java [hunk 19]
```diff
@@ -0,0 +1,71 @@
...
+public abstract class AbstractMutableMessageChannel implements MutableMessageChannel {
+
+    protected final Set<MessageReceiver> members;
+
+    protected AbstractMutableMessageChannel() {
+        this(Collections.newSetFromMap(new WeakHashMap<>()));
+    }
+
+    protected AbstractMutableMessageChannel(Set<MessageReceiver> members) {
+        this.members = members;
```
ST-DDT: You should call the no arg constructor here and then `this.members.addAll(members)`. Just to be on the safe side.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Send a message to this channel, transforming and sending it to the members.
+     *
+     * @param sender The sender of the message
+     * @param original The original message to send
+     */
+    default void send(Object sender, Text original) {
+        checkNotNull(sender, "sender");
```
ST-DDT: Why are you requiring a none-null sender here?

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -0,0 +1,181 @@
...
+        return new FixedMessageChannel(recipients);
+    }
+
+    /**
+     * Broadcast a message to this channel, transforming and sending it to
+     * the members.
+     *
+     * @param original The original text to send
+     */
+    default void broadcast(Text original) {
```
ST-DDT: Should be `send` as well and should default to `send(null, original)`

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
kashike: I actually left this as-is so that people who know what they are doing can provide their own `Set`.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
kashike: I'm still thinking about this one, but my thoughts currently are: if no sender is available the `broadcast` method should be sent.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
kashike: See above comment

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Gets a collection of all members in this channel.
+     *
+     * @return A collection of all members of this channel
+     */
+    Collection<MessageReceiver> getMembers();
+
+    /**
+     * Gets a mutable delegate of this channel.
```
ST-DDT: Should be something more like this:

> Gets or creates a mutable delegate of this channel. if this is already a mutable message channel than this instance is returned. Otherwise the current member list will be copied and used in the newly created message channel which delegates its #transformMessage(...) method to this instance. After that any changes to the member list of this instance won't reflect on the created instance.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -22,63 +22,51 @@
...
     /**
-     * A message sink that targets all subjects currently active.
+     * Sends the formatted text message(s) to source when possible. If text formatting
+     * is not supported in the implementation it will be displayed as plain text.
      *
-     * @return The sink
+     * @param messages The message(s)
      */
-    MessageSink toAll();
+    void sendMessages(Text... messages);
```
ST-DDT: Should default to the `sendMessage(Text)` version to simplify implementation.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -22,63 +22,51 @@
...
     /**
-     * A message sink that targets all players currently active.
+     * Sends the formatted text message(s) to source when possible. If text formatting
+     * is not supported in the implementation it will be displayed as plain text.
      *
-     * @return The sink
+     * @param messages The messages
      */
-    MessageSink toAllPlayers();
+    void sendMessages(Iterable<Text> messages);
```
ST-DDT: Same here.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -22,63 +22,51 @@
...
 
     /**
-     * An empty message sink.
+     * Return the message channel that messages from this source should be sent to.
      *
-     * @return The sink
+     * @return This source's active message channel
      */
-    MessageSink toNone();
+    MessageChannel getMessageChannel();
```
ST-DDT: Wasn't there a discussion to move this in a separate interface `MessageSender` or similar?

src/main/java/org/spongepowered/api/text/channel/MutableMessageChannel.java
```diff
@@ -22,42 +22,39 @@
...
-package org.spongepowered.api.text.sink;
-
-import org.spongepowered.api.text.Text;
-import org.spongepowered.api.command.CommandSource;
+package org.spongepowered.api.text.channel;
 
 /**
- * Represents a function that takes a message and transforms it for distribution to the given targets.
+ * Represents a channel that takes a message and transforms it for distribution
+ * to the members.
```
ST-DDT: > to a __mutable__ list of members

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
kashike: I don't want to put a bunch of that in here, seeing as that behaviour can change in custom implementations.

src/main/java/org/spongepowered/api/text/channel/impl/DelegateMutableMessageChannel.java
```diff
@@ -0,0 +1,57 @@
...
+ * members to the delegate channel passed.
+ */
+public class DelegateMutableMessageChannel extends AbstractMutableMessageChannel {
+
+    protected final MessageChannel delegate;
+
+    public DelegateMutableMessageChannel(final MessageChannel delegate) {
+        super();
+        this.delegate = checkNotNull(delegate, "delegate");
+        this.members.addAll(this.delegate.getMembers());
```
ST-DDT: With my changes proposed above, you could just call `super(delegate.getMembers())`.

src/main/java/org/spongepowered/api/text/channel/type/FixedMessageChannel.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+/**
+ * A message channel that targets the given recipients.
+ */
+public class FixedMessageChannel implements MessageChannel {
```
ST-DDT: This looks like a good candidate to be a `MutableMessageChannel`.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
kashike: done

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
kashike: done

src/main/java/org/spongepowered/api/text/channel/MutableMessageChannel.java
kashike: done

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
ST-DDT: What about naming this method `send` or naming the other message `broadcast` as well.
Because having a/no sender does not really make this a/non broadcast at all.
Broadcast is more about the receivers (send to all) and in this matter both methods send to the same receivers.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
ST-DDT: Then add at least a hint that "it might happen". Nothing is worse than an undocumented and unexpected behaviour.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
kashike: Renamed to `send`.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
ST-DDT: Ok, then please add a javadoc hint that the created instance will use the passed `Set` reference (and not a copy as everywhere else in the API) and thus any change to said `Set` will directly affect the members of this channel.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -298,12 +298,12 @@
...
     /**
-     * Get the sink that messages to be broadcast across the whole server
+     * Get the channel that messages to be broadcast across the whole server
      * should be sent to.
      *
-     * @return The server-wide broadcast sink
+     * @return The server-wide broadcast channel
      */
-    MessageSink getBroadcastSink();
+    MessageChannel getBroadcastChannel();
```
ST-DDT: I'm not sure about that, but this could probably be defaulted to `MessageChannels.TO_ALL`.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
```diff
@@ -0,0 +1,82 @@
...
+package org.spongepowered.api.text.channel;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+/**
+ * An abstract implementation of {@link MutableMessageChannel}, using a
+ * {@link Set} as the underlying {@link Collection}.
```
ST-DDT: > as the underlying member list.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * <p>The {@link #AbstractMutableMessageChannel() default constructor} uses
```
ST-DDT: Move this to the associated ctor.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -0,0 +1,179 @@
...
+     */
+    MessageChannel TO_PLAYERS = () -> ImmutableSet.copyOf(Sponge.getGame().getServer().getOnlinePlayers());
+    /**
+     * A channel with the server console as a member.
+     */
+    MessageChannel TO_CONSOLE = () -> ImmutableSet.of(Sponge.getGame().getServer().getConsole());
+    /**
+     * A channel with all online players, as well as the server console, as members.
+     */
+    MessageChannel TO_ALL = () -> ImmutableSet.<MessageReceiver>builder()
```
simon816: What about `RconSource`s?

ST-DDT: Since there is no `server.getRconSources()` this needs to be handled by the server itself (exchange the impl) or do you have an alternative suggestion?

simon816: It's something that's not currently handled in the implementation but there is a TODO marker on it
https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/text/sink/SpongeMessageSinkFactory.java#L90 Perhaps just add a TODO on this so we can think about it another time

ST-DDT:
````java
List<Consumer<ImmutableSet.Builder<MessageReceiver>>> TO_ALL_PROVIDERS = new ArrayList<Consumer<ImmutableSet.Builder<MessageReceiver>>>() {
	{
		add(builder -> builder.addAll(Sponge.getGame().getServer().getOnlinePlayers()));
		add(builder -> builder.add(Sponge.getGame().getServer().getConsole()));
	}
};

MessageChannel TO_ALL = () -> TO_ALL_PROVIDERS.stream()
		.<ImmutableSet.Builder<MessageReceiver>> collect(ImmutableSet::builder,
				(builder, consumer) -> consumer.accept(builder),
				(left, right) -> left.addAll(right.build()))
		.build();
````

Comment (kashike) [inline src/main/java/org/spongepowered/api/text/channel/MessageChannel.java]: @simon816 In vanilla, only commands are sent to rcon. If we start sending chat messages as well, anything using rcon will get confused, possibly flooded, etc.

File: src/main/java/org/spongepowered/api/Server.java (hunk 29 above)
Comment (kashike) [inline src/main/java/org/spongepowered/api/Server.java]: Nope, changed in a newer push.

File: src/main/java/org/spongepowered/api/text/channel/MessageChannel.java (hunk 32 above)
Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/text/channel/MessageChannel.java]: @kashike  Then maybe we should rename it to `TO_CONSOLE_AND_PLAYERS` or `TO_CONSOLE_PLAYERS` (or maybe `TO_MAIN` if the others are too long) so there is no confusion over the actual targets and allow the addition of an actual `TO_ALL` later on if necessary.

File: src/main/java/org/spongepowered/api/event/message/MessageEvent.java [hunk 33]
```diff
@@ -22,37 +22,39 @@
...
     /**
-     * Gets the {@link Text} message created by the {@link CommandSource} before
-     * the calling of this event.
+     * Gets the original {@link Text} message.
      *
-     * @return The message
+     * @return The message if present, otherwise {@link Optional#empty()}
      */
-    Text getOriginalMessage();
+    Optional<Text> getOriginalMessage();
```
gabizou: How can you have a null original message?

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.Collection;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a channel that takes a message and transforms it for distribution
+ * to the members.
+ */
+public interface MessageChannel {
```
gabizou: If this interface has only one method, mark it as `@FunctionalInterface`.

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
ST-DDT: If a non-player and non-tamed `Entity` is killed, then the message event is triggered but it does not contain a message, but you can set a message so each `Entity` that is killed shows a message like it would have been a `Player`.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -0,0 +1,80 @@
...
+    void sendMessage(Text message);
+
+    /**
+     * Sends the formatted text message(s) to source when possible. If text formatting
+     * is not supported in the implementation it will be displayed as plain text.
+     *
+     * @param messages The message(s)
+     */
+    default void sendMessages(Text... messages) {
+        for (Text message : messages) {
```
gabizou: Missing some check not nulls.

kashike: It assumes that the `sendMessage(Text)` method will be performing null checks (one null check vs two). Would you prefer double checking?

ST-DDT: I guess he is referring to:
````java
for (Text message : checkNotNull(messages, "messages")) {
````

Comment (Zidane) [inline src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java]: A varargs array reference can be null, not just the elements inside of it. You have to check to make sure IT isn't null.

Comment (kashike) [inline src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java]: done

File: src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java [hunk 36]
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Sends the formatted text message(s) to source when possible. If text formatting
+     * is not supported in the implementation it will be displayed as plain text.
+     *
+     * @param messages The message(s)
+     */
+    default void sendMessages(Text... messages) {
+        checkNotNull(messages, "messages");
+        checkArgument(messages.length > 0, "empty array of messages");
```
Minecrell: Although it doesn't make sense it should be fine anyway (should just do nothing in this case)

kashike: @gabizou wanted it

gabizou: Yep, It's better that people can't just send `sendMessages();` and expect it to fail silently.

src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/KickPlayerEvent.java
```diff
@@ -25,11 +25,11 @@
...
-import org.spongepowered.api.event.command.MessageSinkEvent;
+import org.spongepowered.api.event.message.MessageChannelEvent;
 
 /**
  * Fired when a {@link Player} is kicked.
  */
-public interface KickPlayerEvent extends TargetPlayerEvent, MessageSinkEvent {
+public interface KickPlayerEvent extends TargetPlayerEvent, MessageChannelEvent {
 
 }
```
Minecrell: :(

kashike:
```
16:47:31 minecrell | Can't promise I'll be here later
16:47:39 minecrell | make sure to poke the others to merge your PR
```

PR #986, opened 2015-12-23
Refactor tab list API

kashike:
This is a more in-depth change of the tab list API (minor change was done in https://github.com/SpongePowered/SpongeAPI/pull/979).

This does the following:
- renames `PlayerTabInfo` to `TabListEntry`
- removes `setGameProfile` from `TabListEntry`
- renames the methods in `TabList` to reference "entries" instead of "players"


simon816:
How about adding the things you mentioned in #623?
Also it was mentioned here that there could be a `reset()` method https://github.com/SpongePowered/SpongeAPI/commit/d37d3c5c4a95ae219a51bf0c5c9887a6bf544679#commitcomment-9978558

kashike:
@simon816 The things mentioned in that issue I opened will be possible after this PR, as well as others (both current and future), are merged.
As for reset, that is doable.

AlphaModder:
Why remove all the stuff from `TabListEntry`?

JBYoshi:
@AlphaModder They were just moved within the file.

Issue #932, opened 2015-11-24
Move all builders into their associated built interfaces


simon816:
The following builders are not following the new convention to have the builder interface within the target interface they build:

Data builders won't be changing:
~~`org.spongepowered.api.data.manipulator.DataManipulatorBuilder`~~
~~`org.spongepowered.api.data.manipulator.ImmutableDataManipulatorBuilder`~~
~~`org.spongepowered.api.data.value.ValueBuilder`~~
~~`org.spongepowered.api.data.DataTransactionBuilder`~~
~~`org.spongepowered.api.data.ImmutableDataBuilder`~~
~~`org.spongepowered.api.service.persistence.DataBuilder`~~

PaginationBuilder doesn't "build" an object in the same way other builders do
~~`org.spongepowered.api.service.pagination.PaginationBuilder`~~

* [ ] `org.spongepowered.api.entity.ai.task.AITaskBuilder`
* [x] `org.spongepowered.api.item.inventory.transaction.InventoryTransactionBuilder`
* [x] `org.spongepowered.api.text.selector.SelectorBuilder`
* [x] \(`org.spongepowered.api.text.title.TitleBuilder`)
* [x] \(`org.spongepowered.api.text.TextBuilder`)
* [x] `org.spongepowered.api.util.ban.BanBuilder`
* [x] `org.spongepowered.api.world.WorldBuilder`

gabizou:
Data related stuff will stay as is.

simon816:
OK I'll remove them from the list

Minecrell:
Actually, what is with the AI task builders?

gabizou:
`AITaskBuilder` will not be moving into `AITask` as that makes some assumptions that all `AITask`s can be built given certain conditions.

PR #988, opened 2015-12-26
Text API revision 3

Minecrell:
This is the draft for the third revision of the Text API open for review in the next few days.

_"Short"_ summary of changes:
- Split Text classes into separate classes and move the builders into these classes:
  - `Text` -> `Text`, `TextBuilder` -> `Text.Builder`
  - `Text.Literal` -> `LiteralText`, `TextBuilder.Literal` -> `LiteralText.Builder`
  - `TitleBuilder` -> `Title.Builder`
- Remove static utility classes (e.g. `Texts`, `Titles`, `Selectors`), moved into the appropriate class
- Added locales class with a list of locales in Vanilla Minecraft
- Refactor `TextRepresentation` -> `TextSerializer`, rename methods
  - `Texts.toPlain()` -> `TextSerializers.PLAIN.serialize(Text)` or `text.toPlain()`
  - `Texts.json().to(Text)` -> `TextSerializers.JSON.serialize(Text)` or `text.to(TextSerializers.JSON)`
  - `Texts.json().from(String)` -> `TextSerializers.JSON.parse(String)`
  - `Texts.replaceCodes(String, char, char)` -> `TextSerializers.LEGACY.replaceCodes(String, char)`
- Added new text serializer (`TextSerializers.COLOR_CODE`, **TODO: Naming**) for the usual color codes (using `&`), which is unlike `TextSerializers.LEGACY` not deprecated and safe for use
- Remove `Text.Placeholder` and `Texts.format(Text)`
- Much more things I can't remember.


src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -222,15 +282,18 @@ public int hashCode() {
...
-    public String toString() {
+    protected Objects.ToStringHelper toStringHelper() {
         return Objects.toStringHelper(Text.class)
                 .add("format", this.format)
                 .add("children", this.children)
                 .add("clickAction", this.clickAction)
                 .add("hoverAction", this.hoverAction)
-                .add("shiftClickAction", this.shiftClickAction)
-                .toString();
+                .add("shiftClickAction", this.shiftClickAction);
```
ST-DDT: Please change the method to return a reduced output, whenever possible.

````java
return Objects.toStringHelper(Text.class)
	.omitNullValues()
	.add("format", this.format.isEmpty() ? null : this.format) // <- Optional, but recommended
	.add("children", this.children.isEmpty() ? null : this.children) // <- Optional, but recommended
	.add("clickAction", this.clickAction.orElse(null))
	.add("hoverAction", this.hoverAction.orElse(null))
	.add("shiftClickAction", this.shiftClickAction.orElse(null));
````
Here an example: `Text.of()` aka empty `Text`
Currently:
`Text{format=TextFormat{style=NoneTextStyle{}, color=NONE}, children=[], clickAction=Optional.empty, hoverAction=Optional.empty, shiftClickAction=Optional.empty, }`

Improved:
`Text{}`

File: src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java [hunk 2]
```diff
@@ -25,17 +25,16 @@
...
-    public static final Text LT_TEXT = Texts.of("<");
-    public static final Text GT_TEXT = Texts.of(">");
-    public static final Text ELIPSES_TEXT = Texts.of("…");
+    public static final Text PIPE_TEXT = Text.of("|");
+    public static final Text SPACE_TEXT = Text.of(" ");
+    public static final Text PLUS_TEXT = Text.of("*");
+    public static final Text NEWLINE_TEXT = Text.of('\n');
+    public static final Text LT_TEXT = Text.of("<");
+    public static final Text GT_TEXT = Text.of(">");
+    public static final Text ELIPSES_TEXT = Text.of("…");
```
simon816: ELLIPSIS_TEXT

src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java
```diff
@@ -25,17 +25,16 @@
...
-    public static final Text PIPE_TEXT = Texts.of("|");
-    public static final Text SPACE_TEXT = Texts.of(" ");
-    public static final Text PLUS_TEXT = Texts.of("*");
-    public static final Text NEWLINE_TEXT = Texts.of('\n');
-    public static final Text LT_TEXT = Texts.of("<");
-    public static final Text GT_TEXT = Texts.of(">");
-    public static final Text ELIPSES_TEXT = Texts.of("…");
+    public static final Text PIPE_TEXT = Text.of("|");
+    public static final Text SPACE_TEXT = Text.of(" ");
+    public static final Text PLUS_TEXT = Text.of("*");
```
simon816: STAR_TEXT or Text.of("+")

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -59,6 +59,7 @@
      *
      * @return The player's locale
      */
+    @Override
```
simon816: Could just remove this method now

src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java
```diff
@@ -25,17 +25,16 @@
...
+    public static final Text NEWLINE_TEXT = Text.of('\n');
```
simon816: Should be string "\n" not char '\n'

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -54,18 +64,34 @@
...
 
+    /**
+     * The empty, unformatted {@link Text} instance.
+     */
+    public static final Text EMPTY = LiteralText.EMPTY;
+
+    /**
+     * An unformatted {@link Text} that will start a new line (if supported).
+     */
+    public static final LiteralText NEW_LINE = new LiteralText("\n");
```
simon816: Already go `CommandMessageFormatting.NEWLINE_TEXT`

src/main/java/org/spongepowered/api/text/TextIterator.java
```diff
@@ -35,7 +33,7 @@
  * Represents a recursive {@link Iterator} for {@link Text} including the text
  * itself as well as all children texts.
  */
-class TextIterator extends UnmodifiableIterator<Text> {
+class TextIterator implements Iterator<Text> {
```
simon816: Why no longer unmodifiable?

src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java
ST-DDT: The char version is supported now as well. [`Text.of(char)`](https://github.com/SpongePowered/SpongeAPI/pull/988/files#diff-c94a2b7359796a94a693e89c08e6dfacR689)

src/main/java/org/spongepowered/api/text/LiteralText.java
```diff
@@ -0,0 +1,298 @@
...
+ *
+ * @see Builder
+ */
+public class LiteralText extends Text {
+
+    static final LiteralText EMPTY = new LiteralText();
+
+    protected final String content;
+
+    LiteralText() {
```
simon816: Constructor not really needed. Could just make `EMPTY = new LiteralText("");`

src/main/java/org/spongepowered/api/text/TextIterator.java
ST-DDT: Because Java 8 does this by default.

src/main/java/org/spongepowered/api/text/Text.java
ST-DDT: Yeah, but the LiteralText.Builder will automatically point to this one. 
=>
`CommandMessageFormatting.NEWLINE_TEXT` <= `Text.NEW_LINE`

[LiteralText.Builder#build()](https://github.com/SpongePowered/SpongeAPI/pull/988/files#diff-f4d8dd0ce6550f824748066654468e58R198)

src/main/java/org/spongepowered/api/text/LiteralText.java
```diff
@@ -0,0 +1,293 @@
...
+import org.spongepowered.api.text.format.TextStyle;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a {@link Text} containing a plain text {@link String}.
+ *
+ * @see Builder
+ */
+public class LiteralText extends Text {
```
Minecrell: Any opinions on `LiteralText` vs `TextLiteral`? Basically both work, and have both their advantages:
- `LiteralText`, `TranslatableText`, ... -> Better readable
- `TextLiteral`, `TextTranslatable`, ... -> Makes auto-completion easier

ST-DDT: You usually only pass around `Text`s and obtain most stuff from `Text.of(X)` or `Text.builder()`, so I would leave this at it is. And if want to set it explicitly then you know what you want and just type the first letters for it, so `LiteralText` is faster than `TextLiteral` in auto completion (especially since IDEs automatically check for applicable types).

PS: `TextTranslatable` sounds weird as name. (Same for the other types except `TextLiteral`)

ST-DDT:
[`Nameable`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/Nameable.java#L32) (currently used for `Inventory` only) should be replaced with [`Translatable`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/translation/Translatable.java#L30).

src/main/java/org/spongepowered/api/text/translation/Translation.java
```diff
@@ -24,21 +24,23 @@
...
  * be translated server-side, for example using Gettext or a
  * {@link ResourceBundle}
  *
  * <p>Some translations require parameters to be sent together with them, if
  * they're not given they will be filled with empty text.</p>
  *
- * <p>While the client has multiple locales available, server-side vanilla translations
- * support only {@link Locale#ENGLISH}.</p>
+ * <p>While the client has multiple locales available, server-side vanilla
+ * translations support only {@link Locale#ENGLISH}.</p>
```
ST-DDT: Shall we keep `Locale.ENGLISH` here or shall we use `Locales.DEFAULT` here?

ST-DDT:
How about this: [Similarize `Text.join()`, `Text.joinWith`, `Text.of`](https://github.com/SpongePowered/SpongeAPI/compare/feature/text3...ST-DDT:feature/text3-joinOf)

ST-DDT:
I started working on the SpongeCatalogType super class.

ST-DDT:
How about this change: [Add `CatalogType.Translatable`](https://github.com/SpongePowered/SpongeAPI/compare/feature/text3...ST-DDT:feature/text3-Catalogs)?

src/main/java/org/spongepowered/api/text/LiteralText.java
kashike: `LiteralText` in my opinion. Better readability and matches the rest of the API.

src/main/java/org/spongepowered/api/text/translation/locale/Locales.java
```diff
@@ -0,0 +1,130 @@
...
+    public static final Locale AZ_AZ = new Locale("az", "AZ");
+    public static final Locale BG_BG = new Locale("bg", "BG");
+    public static final Locale CA_ES = new Locale("ca", "ES");
+    public static final Locale CS_CZ = new Locale("cs", "CZ");
+    public static final Locale CY_GB = new Locale("cy", "GB");
+    public static final Locale DA_DK = new Locale("da", "DK");
+    public static final Locale DE_DE = Locale.GERMANY;
+    public static final Locale EL_GR = new Locale("el", "GR");
+    public static final Locale EN_AU = new Locale("en", "AU");
+    public static final Locale EN_CA = new Locale("en", "CA");
```
kashike: `Locale.CANADA`

src/main/java/org/spongepowered/api/text/title/Title.java
```diff
@@ -204,4 +207,390 @@ public String toString() {
                 .toString();
     }
 
+    /**
+     * Represents a builder class to create immutable {@link Title}
+     * configurations.
+     *
+     * @see Title
+     */
+    public static class Builder {
```
kashike: This looks out of place between methods, move it to the bottom of the class?

gabizou:
> Nameable (currently used for Inventory only) should be replaced with Translatable.

Pretty sure @Mumfrey was adamant about keeping `Nameable` as it was.

src/main/java/org/spongepowered/api/text/title/Title.java
```diff
@@ -204,4 +207,390 @@ public String toString() {
...
+    /**
+     * Creates a new empty {@link Title} configuration builder. Unlike
+     * {@link #builder} this won't reset the current Title on the client before
+     * displaying the current one. This has less use cases but should be used if
+     * just the previously sent Title should be updated.
+     *
+     * @return A new {@link Builder}
+     * @see #builder
+     */
+    public static Builder update() {
```
kashike: This method is... strange. I'm not sure if it is the description or the name (or both) that will cause confusion.

src/main/java/org/spongepowered/api/text/translation/locale/Locales.java
gabizou: Isn't there an `FR_CA` as well? so `CANADA` wouldn't really quite 100% make sense in this case.

src/main/java/org/spongepowered/api/text/serializer/TextSerializers.java
```diff
@@ -0,0 +1,110 @@
...
+     * this {@link FormattingCodeTextSerializer} is broken in many ways and
+     * should not be used anymore unless really necessary. It is recommended to
+     * use a different text serializer such as {@link #FORMATTING_CODE}
+     * instead.</p>
+     *
+     * @see FormattingCodeTextSerializer
+     * @deprecated Legacy formatting codes are being phased out of Minecraft
+     */
+    @Deprecated
+    public static final FormattingCodeTextSerializer LEGACY = null;
```
kashike: `LEGACY_FORMATTING_CODE`

src/main/java/org/spongepowered/api/text/serializer/TextSerializers.java
```diff
@@ -0,0 +1,110 @@
...
+     * text.</p>
+     *
+     * <p><b>Note:</b> Unlike the {@link #LEGACY} {@link TextSerializer} this
+     * is safe to use as simple {@link Text} format for plugins as long as you
+     * keep in mind that it does not support additional {@link Text} features
+     * such as {@link ClickAction}s or {@link HoverAction}s.</p>
+     *
+     * @see FormattingCodeTextSerializer
+     */
+    public static final FormattingCodeTextSerializer FORMATTING_CODE = null;
```
kashike: `AMPERSAND_FORMATTING_CODE`

src/main/java/org/spongepowered/api/text/serializer/TextSerializer.java
```diff
@@ -22,51 +22,51 @@
...
+     * However, if the input string is not of a valid format, the returned
+     * {@link Text} object will be of the raw input, rather than throwing an
+     * exception.
      *
      * @param input The raw input to try to parse into a text
-     * @return An appropriate {@link Text} object from the input string, or the raw input
+     * @return The parsed text for the input string, or the raw input
      */
-    Text fromUnchecked(String input);
+    default Text parseUnchecked(String input) {
```
kashike: `deserializeUnchecked`

src/main/java/org/spongepowered/api/text/serializer/TextSerializer.java
```diff
@@ -22,51 +22,51 @@
...
+     * Returns a {@link Text} instance from an appropriately formatted string.
      *
      * @param input The raw input to parse into a text
-     * @return An appropriate {@link Text} object from the input string
-     * @throws TextMessageException if an error occurs while parsing the input
+     * @return The parsed text for the input string
+     * @throws TextParseException If an error occurs while parsing the text
      */
-    Text from(String input) throws TextMessageException;
+    Text parse(String input) throws TextParseException;
```
kashike: `deserialize`

src/main/java/org/spongepowered/api/text/serializer/FormattingCodeTextSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+ *     Formatting codes on the Minecraft Wiki</a>
+ */
+public interface FormattingCodeTextSerializer extends SafeTextSerializer {
+
+    /**
+     * Returns the formatting character for this {@link FormattingCodeTextSerializer}.
+     *
+     * @return The formatting character
+     */
+    char getCharacter();
```
kashike: `getCode`

src/main/java/org/spongepowered/api/text/title/Title.java
Minecrell: It has always been like this :P

src/main/java/org/spongepowered/api/text/serializer/TextSerializers.java
Minecrell: No, that's too long. It's supposed to be the default one anyway.

src/main/java/org/spongepowered/api/text/serializer/TextSerializers.java
Minecrell: Eh, I'd rather keep it the way it is, it's used quite frequently in the implementation (and probably less in plugins) so it would be annoying having to type that every time.

src/main/java/org/spongepowered/api/text/translation/locale/Locales.java
kashike: @gabizou talking about the `new Locale("en", "CA")` here. @Minecrell uses `CANADIAN_FRENCH` below already.

src/main/java/org/spongepowered/api/text/title/Title.java
```diff
@@ -204,4 +207,390 @@ public String toString() {
...
+        public Builder() {
+        }
+
+        /**
+         * Constructs a new {@link Builder} with the properties of the given
+         * {@link Title} as initial values.
+         *
+         * @param title The title to copy the values from
+         */
+        public Builder(Title title) {
```
TheRaspPie: Why are the constructors public? Shouldn't devs use the toBuilder() and static builder() methods?

ST-DDT:
How about:
* Adding `TextFormat.DEFAULT` to prevent the instantiation of `new TextFormat()` all the time? (As a result that constructor should be made private.)
* Adding `TextActions.RESET` to reset (all) the `TextAction`s (at once) in `Text.of(Object...)`?
* Adding a new `TextProperty` super-interface to `TextFormat`, `TextColor`, `TextStyle` and `TextAction` to reduce the average amount of casting in `Text.of(Object...)`?

Here the [branch](https://github.com/SpongePowered/SpongeAPI/compare/feature/text3...ST-DDT:feature/text3-simplify) containing
* [a small javadoc and code fix including the removal of 2 random semicolons](https://github.com/SpongePowered/SpongeAPI/commit/9d8ba674616bce8e55630e03b819baa29f6f8025)
* [the changes outlined above](https://github.com/SpongePowered/SpongeAPI/commit/e7e52e629250c3102253ba41eb421ac340ac5c1d)

gabizou:
@Minecrell move the builders into nested classes and if you can, make them resettable builders, I'd rather see the `SelectorBuilder` not require a `SelectorType` to construct one, especially since no special handling is done after it. The same goes for the other builders. When you've done that, #932 can be closed.

ST-DDT:
Why are all those static methods (from `Texts`) pushed into `Text`?

IMO thats just messes everything up.

`Text` is already 700 lines long. And now we add 500 more and when we add Placeholders it will become even more (~500). And if I try to just use the autocompletion on `Text`s I get 90% static methods I don't need.

src/main/java/org/spongepowered/api/text/title/Title.java
Minecrell: I prefer having the static methods at the end of the file personally, does it really matter?

Minecrell:
@ST-DDT Going to take a look at your changes tomorrow. If you have time, mind coming on IRC so we can talk about them?

@gabizou Oh I forgot moving the selector package, going to do that tomorrow. I think I can also make the selector builder and possibly the title builder resettable, however the Text builders are most likely going to stay the way they are because they cannot be properly reset.

@ST-DDT I have removed a lot of the static methods that were in the Texts class, it's only building text what's in there now and joining which is going to stay there as an unified way to create instances/builders of the text types. Placeholders/Templates could very well also be in a separate file, depending on the design. And if your IDE suggests using static methods on instance types there is generally something really wrong...

src/main/java/org/spongepowered/api/text/title/Title.java
Minecrell: Oops.

ST-DDT:
@Minecrell propose a time (with timezone) and I will check whether I have time.

Minecrell:
@ST-DDT Can't say exactly but I will be online tomorrow sometime in between 11am - 8 pm (CET, UTC+1)

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -107,6 +130,8 @@
         this.clickAction = Optional.<ClickAction<?>>ofNullable(clickAction);
         this.hoverAction = Optional.<HoverAction<?>>ofNullable(hoverAction);
         this.shiftClickAction = Optional.<ShiftClickAction<?>>ofNullable(shiftClickAction);
+        this.childrenIterable = () ->
+                this.children.isEmpty() ? Iterators.singletonIterator(this) : new TextIterator(this);
```
ST-DDT: This a fix value, you can move it to the field declaration (as it was before)

````java
protected final Iterable<Text> childrenIterable = () ->
        this.children.isEmpty() ? Iterators.singletonIterator(this) : new TextIterator(this);
````
or change it check for `isEmpty()` before it is set.
````java
if (this.children.isEmpty()) {
    this.childrenIterable = () -> Iterators.singletonIterator(this);
} else {
    this.childrenIterable = new TextIterator(this);
}
````

File: src/main/java/org/spongepowered/api/util/SpongeApiTranslationHelper.java [hunk 21]
```diff
@@ -62,7 +61,7 @@ private SpongeApiTranslationHelper() {
      * @return The translatable text
      */
     public static Text t(String key, Object... args) {
-        return Texts.of(new ResourceBundleTranslation(key, LOOKUP_FUNC), args);
+        return Text.of(new ResourceBundleTranslation(key, LOOKUP_FUNC), args);
```
ST-DDT: I know this isn't the right spot, but I cannot comment the declaration above.
`LOOKUP_FUNC` could be replaced with a functional expression.
````java
private static final Function<Locale, ResourceBundle> LOOKUP_FUNC = (locale) ->
        ResourceBundle.getBundle("org.spongepowered.api.Translations", locale);
````

File: src/main/java/org/spongepowered/api/text/Text.java (hunk 20 above)
Comment (Minecrell) [inline src/main/java/org/spongepowered/api/text/Text.java]: Your first example has no difference to what it is currently (keep in mind it is a lambda, so `isEmpty()` would still be checked every time), except that it causes a compile error.
I may change it like in your second example, however not sure if that is really worth it...

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/text/Text.java]: In the first example the values is set outside the constructor.
Usually if something is set within the ctor I assume it is dependent on the ctor arguments, but `childrenIterable` isn't.

Comment (Minecrell) [inline src/main/java/org/spongepowered/api/text/Text.java]: You can test it, it will fail to compile in some cases (for whatever reason), it complains that children might not be initialized yet.

Comment (Minecrell) [inline src/main/java/org/spongepowered/api/text/Text.java]:
```
org/spongepowered/api/text/Text.java:88: error: variable children might not have been initialized

this.children.isEmpty() ? Iterators.singletonIterator(this) : new TextIterator(this);

     ^
```

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/text/Text.java]: Mhh, looks liek another weakness of the eclipse compiler. (Eclipse compiles it without problems)

An alternative solution:
````java
protected final Iterable<Text> childrenIterable = () ->
        getChildren().isEmpty() ? Iterators.singletonIterator(this) : new TextIterator(this);
````

Comment (JBYoshi) [inline src/main/java/org/spongepowered/api/text/Text.java]: How about:

```java
protected final Iterable<Text> childrenIterable = getChildren().isEmpty() ? () -> Iterators.singletonIterator(this) : () -> new TextIterator(this);

Comment (Minecrell) [inline src/main/java/org/spongepowered/api/text/Text.java]: That would only work in the constructor, because the children text wouldn't be set at the point the field initializers are getting called.

File: src/main/java/org/spongepowered/api/util/SpongeApiTranslationHelper.java (hunk 21 above)
Comment (Minecrell) [inline src/main/java/org/spongepowered/api/util/SpongeApiTranslationHelper.java]: Fixed.

File: src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java [hunk 22]
```diff
@@ -25,17 +25,19 @@
...
+
+    private CommandMessageFormatting() {
+    }
+
+    public static final Text PIPE_TEXT = Text.of("|");
+    public static final Text SPACE_TEXT = Text.of(" ");
+    public static final Text STAR_TEXT = Text.of("*");
+    public static final Text LT_TEXT = Text.of("<");
+    public static final Text GT_TEXT = Text.of(">");
+    public static final Text ELLIPSIS_TEXT = Text.of("…");
```
phroa: Could these just be moved to the Text class itself?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -163,7 +161,7 @@ public static CommandElement location(Text key) {
 
     static class MarkTrueCommandElement extends CommandElement {
```
phroa: not `private`?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -901,7 +899,7 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
 
         @Override
         public Text getUsage(CommandSource src) {
-            return Texts.of(Joiner.on(' ').join(this.expectedArgs));
+            return Text.of(Joiner.on(' ').join(this.expectedArgs));
```
phroa: I'm pretty sure `String::join` would fit in here

src/main/java/org/spongepowered/api/command/CommandMessageFormatting.java
phroa: oh, and these could be `LiteralText`s, right?

kashike: I don't see a need for these to be in `Text`, it's a large enough class already.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
Minecrell: Not touching commands, that would be something for @zml2008

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
Minecrell: Same as above

Minecrell:
https://github.com/SpongePowered/SpongeAPI/commit/6ee796d6f3e53f155976f485a030ed8c7848985f

PR #991, opened 2015-12-29
Add a getServer method to Sponge

kashike:
`Server` is something constantly accessed, I see no point in not having this method if we have ones for far less used things.


ST-DDT:
Nice addition!

simon816:
https://github.com/SpongePowered/SpongeAPI/commit/50ed63e887d678d0da93711ee9220e900bf8f92a

Issue #923, opened 2015-11-18
Plugin signing


kashike:
As seen in Forge, mods are able to verify jar integrity using fingerprints/signatures:
https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/fml/common/Mod.java#L151
https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/fml/common/event/FMLFingerprintViolationEvent.java
https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/fml/common/FMLModContainer.java#L452-L540

I think this would be a useful thing to have as part of the Sponge API.

Deamon5550:
Yeah I certainly agree, I was already talking to gabizou about this earlier. There are several validation steps that we could perform. Another thing we can do with Ore served plugins is check if plugins are up to date and notify the server admin if there are updates available.

JBYoshi:
@Deamon5550 The Ore stuff should probably go in a separate issue.

gabizou:
I'm curious whether this is still valid to contain in the API. As I can see at the moment, many things have changed since 2015, and while Ore does have a system to verify the validity of a plugin's jar being signed, I'm not sure whether we should be relying on plugin jars being verified as signed and opening remote connects to grab the signature files for verification. @windy1 may have more input on this.

ryantheleach:
My understanding is that the way that Ore signs is completely different to the forge jar signing system?

Will need someone to look into this if we plan on supporting it.

Personally, the large reason I think forge jar's were being signed was to prevent mod rehosting modifying the jar files, and having different mods and authors back each other up.

Do we have this problem yet? And if we did, I'm not sure we want the same solution.

pie-flavor:
Perhaps instead of a per-plugin thing it could be a per-plugin-host thing, where the server by default only accepts signatures from the Ore public key.

PR #1007, opened 2016-01-05
Provide a way for plugins to check certificate fingerprints

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1007) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/485) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/200)

This adds the ability for plugins to provide a certificate fingerprint to verify jar signatures. #923

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -29,6 +29,7 @@ import com.flowpowered.math.vector.Vector3i;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import java.net.InetSocketAddress;
+import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


JBYoshi:
Wait... couldn't someone who wanted to modify the plugin JAR get around the warnings by removing the signature checking code?

kenzierocks:
My thinking on the signatures is not security, but corruption. Easy to forge, but corrupt data will fail to load in most cases.

kashike:
@JBYoshi Of course there's way around checking, there's no way to prevent removing checking.

ryantheleach:
I've also heard of mod authors teaming up to check each others fingerprints and stuff I think.

ST-DDT:
IMO its the servers task to check.

TOFU principle.

Plugin is installed
Server starts and detects that said plugin has a signature. The certificate/signature details are stored by the server in a config file.
Server stops
Plugin will be replaced/gets corrupted
Server starts. The server detects that the plugin should have a signature but detects that the signature is broken or missing and can warn the user appropriately.

PR #1009, opened 2016-01-05
Refactor explosion events

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1009) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/401) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/486)

- Renames the original `ExplosionEvent.Pre` event to `ExplosionEvent.Start`
- Adds a new `ExplosionEvent.Pre` event
 - This event is fired **before** an `Explosion` is created, so the `getExplosion` method has been moved to `Start`, `Detonate`, and `Post`
- Adds `getOriginalExplosion` to `ExplosionEvent.Start`
- Adds `NamedCause.igniter(Object)` method.

Implementation PRs are coming as soon as they are cleaned and tested.


Zidane:
Not a bad PR @kashike but I see little usefulness of the Pre event...knowing only the world is of little use beyond "Are explosions allowed in world".

kashike:
@Zidane More than the `World` is included in the Cause, the impl PR will be coming later today.

ST-DDT:
@kashike IIRC The explosion itself is immutable, due to that the pre event should contain getters/setters for the radius, incinerate etc that can be used to create the explosion between pre and explode. Or do I get the event order wrong?.

src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
```diff
@@ -36,16 +36,33 @@
...
      */
-    interface Pre extends ExplosionEvent, Cancellable {
+    interface Start extends ExplosionEvent, Cancellable {
+
+        /**
+         * Gets the original {@link Explosion} involved in this event.
+         *
+         * @return The original explosion that this event is involved in
+         */
+        Explosion getOriginalExplosion();
```
gabizou: Can't you make this a `Transaction<Explosion>`?

kashike:
@ST-DDT The event order is:
- `Pre`
- `Start`
- `Detonate`
- `Post`
@gabizou I can, yes.

kashike:
and to answer the second part of your question, @ST-DDT, there's no easy way to handle explosion properties in `Pre`.

JBYoshi:
@kashike What's the difference between `ExplosionEvent.Pre` and `ExplosionEvent.Start`?

kashike:
@JBYoshi See the description.

src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
Zidane: ^ this as well as the original explosion should be available in ```Post```.

gabizou:
@kashike Wouldn't this have to be merged into bleeding and not master?

kashike:
@gabizou As I have said to you many times now, I can't work off of bleeding because it does not contain master's changes. So no.

src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
kashike: I can't make this a `Transaction<Explosion>` without finding a way to set the `explosion` field on the Forge `ExplosionEvent`.

gabizou:
> @gabizou As I have said to you many times now, I can't work off of bleeding because it does not contain master's changes. So no.

Dunno how to say it aside from saying it again: This PR is a breaking change. If anything bleeding should be rebased onto master at some point this week, but that doesn't change the fact that this is a ***BREAKING*** change, which means it's slated for 4.0's release.

Deamon5550:
@kashike @gabizou I just rebased bleeding for you :)

kashike:
https://github.com/SpongePowered/SpongeAPI/pull/1011

kashike:
@gabizou Re-opened on bleeding.
thanks, @Deamon5550 :heart_eyes_cat:

PR #1012, opened 2016-01-06
Added more vanilla AIs

liach:
- Needs a lot of review
- Added `RangedAttackingAgent` for special AIs and resolved "SnowGolem should implement `ProjectileSource`." issue.
- Add `Tick` for `AITaskEvent`, needs further discussion.
- Javadocs will be added later.

TODO:
- [ ] Add javadoc

P. S. I am new to Sponge contribution, and I hope that Zidane or Aaron could give me some suggestions. Common and Forge implementations should come later, since only these AIs are already a lot.


src/main/java/org/spongepowered/api/entity/ai/task/MutexBitsConstants.java
```diff
@@ -0,0 +1,63 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.ai.task;
+
+/**
+ * Few utilities and constants for AIs.
+ */
+public final class MutexBitsConstants {
```
kenzierocks: These could probably be done as a `CatalogType` instead.

liach: I thought it was like asm constants. I'd hear comments from more people first.

src/main/java/org/spongepowered/api/entity/living/ArrowAttacker.java
```diff
@@ -0,0 +1,36 @@
...
+package org.spongepowered.api.entity.living;
+
+import org.spongepowered.api.entity.ai.Goal;
+import org.spongepowered.api.entity.ai.task.builtin.creature.arrowattacker.AttackWithArrowAITask;
+import org.spongepowered.api.entity.projectile.source.ProjectileSource;
+
+/**
+ * {@link Agent}s which could have {@link AttackWithArrowAITask} in its {@link Goal}s.
+ */
+public interface ArrowAttacker extends Agent, ProjectileSource {
```
kenzierocks: IMO this could use a rename since things like withers and snow golems use this, and they don't use arrows at all.

liach: So what should it be called? In vanilla this interface is called `net.minecraft.entity.IRangedAttackMob`, but some `ProjectileSource`s such as `Blaze` and `Guardian` do not implement this interface. The only method in this interface is only used by the AI in MCP code.

src/main/java/org/spongepowered/api/entity/ai/task/MutexBitsConstants.java
ST-DDT: This is API so it should be mostly about readability and abstraction. It should be easy to use even if the impl  does not use int/byte masks.
I would recommend  using a `CatalogType`. Then you can specify them as `Set` in tasks/goals.

src/main/java/org/spongepowered/api/entity/living/ArrowAttacker.java
kashike: `RangedAgent`
`RangedAttacker`

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -95,4 +95,16 @@
...
+     * Resets this task.
+     */
+    void reset();
+
+    /**
+     * Gets the mutex bits. When the | operation result of two mutex bits is 0, they can run
+     * concurrently.
+     * @return The mutex bits which categorizes this task
+     */
+    int getMutexBits();
```
kashike: This was actually removed before the AI API was merged into master. It was replaced with `canRunConcurrentWith`.

kashike:
This appears to be breaking, so this should be targeted to the `bleeding` branch, not the `master` branch.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
ST-DDT: The issue with canRunConcurrentWith is that you have to know details about every task/goal. This is impossible for plugin based tasks and thus a definition of affected elements from the entity movement/view etc are important to know to check whether the actualy can concurrently.

If you dont want them everywhere then you could still move it to a seperate interface.

src/main/java/org/spongepowered/api/entity/living/ArrowAttacker.java
ryantheleach: I'm guessing it's used by things that need a PROJECTILE (A real projectile undergoing https://en.wikipedia.org/wiki/Projectile_motion, not gravity defying "projectiles") But given the context naming it a ProjectileAttacker could be problematic...

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
kashike: I just stated it so that it's known. Best to poke @Zidane about exposing mutex bits - I have no problem with it personally.

liach:
I will leave it for discussion right now. I hope some collaborator could rebase and commit this change onto `bleeding`, though I did not know about what was `bleeding` branch for at first.

Tzky:
The documentation on our all new branching model is still in the works. However ``master`` is only used for non-breaking changes (minor changes: 3.0.0 -> 3.1.0). Everything else will automatically target our 4.0.0 release which will contain breaking changes that were merged into ``bleeding``

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
Zidane: The mutex bits are an impl detail. For our API they are not required. If I can get a usecase for having them then I'll consider a child interface for a VanillaAITask of sorts but they won't go into the base interface.

@ST-DDT ....which you can get those details by calling other.getGoal().get().getOwner()...etc

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -75,7 +75,7 @@
      * @param other The other task
      * @return True if it can be, false if not
      */
-    boolean canRunConcurrentWith(AITask<O> other);
+    boolean canRunConcurrentWith(AITask<?> other);
```
Zidane: Sorry but it is by design that you'll always get a task passed in with the same GoalType. Change it back or provide a reason why you changed this.

src/main/java/org/spongepowered/api/entity/living/ArrowAttacker.java
gabizou: Why does the interface need to even be exposed to the API?

src/main/java/org/spongepowered/api/entity/ai/task/builtin/GroundNavigationOnly.java
```diff
@@ -0,0 +1,43 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.ai.task.builtin;
+
+import org.spongepowered.api.entity.ai.task.AITask;
+import org.spongepowered.api.entity.ai.task.AITaskBuilder;
+import org.spongepowered.api.entity.living.Agent;
+
+public interface GroundNavigationOnly<O extends Agent> extends AITask<O> {
```
Zidane: Not seeing the point of this classification. Could you elaborate on where you are going with this?

src/main/java/org/spongepowered/api/entity/ai/task/builtin/doorinteract/package-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault package org.spongepowered.api.entity.ai.task.builtin.doorinteract;
```
gabizou: Missing new line at end of file.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/movetolocation/MoveToBlockAITask.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Predicate<Location<World>> getDestinationSelector();
+
+    A setDestinationSelector(Predicate<Location<World>> destinationSelector);
+
+    Location<World> getDestination();
+
+    A setDestination(Location<World> destination);
+
+    interface MoveToBlockAITaskBuilder<
```
Zidane: See GroundNavigationAITask above.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/wolf/package-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault package org.spongepowered.api.entity.ai.task.builtin.creature.wolf;
```
gabizou: Missing new line at end of file.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/target/FindNearestAttackableTargetAITask.java
```diff
@@ -24,31 +24,39 @@
  */
 package org.spongepowered.api.entity.ai.task.builtin.creature.target;
 
+import org.spongepowered.api.entity.living.Creature;
 import org.spongepowered.api.entity.living.Living;
 
 import java.util.function.Predicate;
 
-public interface FindNearestAttackableTargetAITask extends TargetAITask<FindNearestAttackableTargetAITask> {
+public interface FindNearestAttackableTargetAITask<
```
Zidane: Formatting

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/villager/package-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault package org.spongepowered.api.entity.ai.task.builtin.creature.villager;
```
gabizou: Seems like all your package-infos are missing a new line at the end of file.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/villager/VillagerMateAITask.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.spongepowered.api.entity.ai.task.builtin.creature.villager;
+
+import org.spongepowered.api.entity.ai.task.AITask;
+import org.spongepowered.api.entity.ai.task.AITaskBuilder;
+import org.spongepowered.api.entity.living.Villager;
+
+import java.util.Optional;
+
+public interface VillagerMateAITask extends AITask<Villager> {
```
gabizou: This sounds like some what-should-really-be-generic-passive-ageable-task instead of strictly being for `Villager`s.

src/main/java/org/spongepowered/api/event/entity/ai/AITaskEvent.java
```diff
@@ -77,6 +77,14 @@
     }
 
     /**
+     * Fired when an {@link AITask} is updated every
+     */
+    @FactoryCodeCheck(errorMessage = "The target entity '%s' is not the owner of the goal '%s'!")
```
Zidane: @Aaron1011 what is this?

src/main/java/org/spongepowered/api/entity/ai/task/builtin/GroundNavigationOnly.java
liach: Because Those AI Tasks throws an exception when the owner entity's PathFinder is not `PathNavigateGround` (Name I forgotten, but it must be ground navigator)

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/villager/VillagerMateAITask.java
liach: Restricted to villagers in vanilla.

src/main/java/org/spongepowered/api/event/entity/ai/AITaskEvent.java
liach: Maybe it generates the code [here](https://github.com/SpongePowered/SpongeAPI/pull/1012/files#diff-1378043d9608b3c46ff7baa499da8e70R2343)

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
liach: Here `O` is the type of `Agent`. The `other` parameter's generic could either be `O`'s subclass or superclass. For example, if this task applies to `Monster`s, the `other` task could apply to either `Agent`s and `Spider`, if both tasks exists on the goal of a spider.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/movetolocation/MoveToBlockAITask.java
liach: Oops at least I will fix this builder.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/target/FindNearestAttackableTargetAITask.java
liach: To break the line before `<`?

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -72,10 +73,12 @@
      * 3. Returning "true" will add the provided task to the list of updated tasks, if not there
      *    already.
      *
-     * @param other The other task
+     * @param that The other task
      * @return True if it can be, false if not
      */
-    boolean canRunConcurrentWith(AITask<O> other);
+    default boolean canRunConcurrentWith(AITask<?> that) {
```
liach: Topic continued from [Zidane's](https://github.com/SpongePowered/SpongeAPI/pull/1012#discussion_r48967014)

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
liach: So are we going to replace mutex bits with [this catalog type](https://github.com/SpongePowered/SpongeAPI/pull/1012/files#diff-c108db2d21314ed0b6196b9614fd0dc6R31)?

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
Zidane: Then O should be an upper-bound

```
AITask<? super O> other
```

Having ? is worse than O as you'll not be able to infer ANYTHING about it.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/target/FindNearestAttackableTargetAITask.java
Zidane: Run the line to length then break...don't do this barely half-line break.

Common sense > auto-formatter.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -95,4 +98,24 @@
...
+     */
+    default int getMutexBits() {
+        int ret = 0;
+        for (AITaskCategory aiTaskCategory : getTaskCategories()) {
+            ret |= aiTaskCategory.getMutexBits();
+        }
+        return ret;
+    }
+
+    List<AITaskCategory> getTaskCategories();
```
ST-DDT: Use either a `Set` or a `Collection` since categories shouldn't be in there twice.

src/main/java/org/spongepowered/api/entity/ai/task/AITaskCategory.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.spongepowered.api.entity.ai.task;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(AITaskCategories.class)
+public interface AITaskCategory extends CatalogType {
+
+    int getCategoryNumerialId();
```
ST-DDT: Impl detail, should't be exposed IMO.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
gabizou: Also note that with generics, at runtime, the only thing you'll know is the lower bound (which is still going to be `AITask<?>` unless you explicitly cast for generics.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/creature/villager/VillagerMateAITask.java
gabizou: Why not make it agreeable since those are the only ones that can actually mate? (which covers animals and villagers btw).

Zidane:
@liach 

I think you misunderstood. AITask should have no representation of mutex bits since that is a IMPLEMENTATION detail. Also, the impl accounts for the mutex bits when checking to see if two tasks are compatible.

Should you really...really...need it for your custom ```AbstractAITask```, what you ought to do is simply make a child interface (so we can keep the Minecraft stuff separate).

```
interface VanillaAITask<O extends Agent, V extends VanillaAITask<O, V> extends AITask<O> {
    int getMutexBits();
    V setMutexBits(int bits);
}
```

Do NOT default implement the method. Again, my impl is already doing this. All Vanilla AI interfaces would extend the above.

ST-DDT:
@Zidane How would I check that my `HeadShakerAITask` (LOOK) can run at the same time as `RandomStrollAITask` (MOVEMENT) or `PluginYFollowPlayerAITask` (MOVEMENT), but not at the same time as `PluginYRandomLookAroundAITask` (LOOK), if we don't have that method?

PS: I'm against using a bitmask here since that's really an impl detail. (Use a `CatalogType`)

Zidane:
@ST-DDT 

You implement canRunConcurrentWith and check the task being passed in. There is no reason to have the mutex bits or use them besides....

A. You want to use them for your custom tasks (lol whyyy) and if so you can just have your own numbers.
B. You want to utilize the numbers coming from Vanilla...which you can if the changes I made above would work.

Regardless, its a method that takes a task and returns a bool...how you determine that bool return is up to you.

liach:
So should I represent it via the catalogtype? And the conversion between the type and mutix will be in implementations. Mutex will be 100% removed. Will work on that hours later.

Zidane:
@liach 

Lets keep the mutex stuff to a separate PR as its muddying this one.

ST-DDT:
Sorry I was too slow finishing this one

> You implement canRunConcurrentWith and check the task being passed in. There is no reason to have the mutex bits or use them besides....

````java
public boolean canRunConcurrentWith(AITask task) {
    return allowedConcurrentTaskTypeSet.contains(task.getTaskType());
}
````

Really?

And if we flip the situation: How is vanilla's `RandomStrollAITask` supposed to know that it can run concurrently with my custom `HeadShakerAITask`?

> A. You want to use them for your custom tasks (lol whyyy) and if so you can just have your own numbers.

Example: I would like to make all my `Zombie`s shake their head the entire time, regardless what he is doing otherwise (moving and attacking). So I add my task first in the task list/goals. But I have no intention to rewrite the entire MC AI and every plugins AI just to ensure that I know which tasks are compatible and they have to do the same to know which of their tasks are compatible with mine.

My task knows it deals only with the LOOK why do I have to keep track of all other plugins AI tasks that also use LOOK? IMO it should be possible to ask the task, "What do you need?" "Well I need LOOK and MOVEMENT." "Ok, sorry task we cannot run concurrently!"
A plugin/ai task may ignore a requirements clash, but AITasks should be as stupid as possible.

> B. You want to utilize the numbers coming from Vanilla...which you can if the changes I made above would work.

Which changes are you referring to?
But again, then every server impl that want to allow their smart plugin users to use the mutex will be forced to use ints themselves. Why not long? What is `0x1`?  Solution, use an abstraction logic aka `CatalogType`

````java
Set<AIRequirement> getRequirements();
````

Also a setter is not necessary. The requirements should NEVER change after the initial creation.

> Regardless, its a method that takes a task and returns a bool...how you determine that bool return is up to you.

And `getRequirements()` is a simple and abstract way for plugins to calculate that bool.

End of mutex discussion from my side here.

File: src/main/java/org/spongepowered/api/entity/ai/task/builtin/GroundNavigationOnly.java (hunk 5 above)
Comment (liach) [inline src/main/java/org/spongepowered/api/entity/ai/task/builtin/GroundNavigationOnly.java]: Would you like me to remove this interface? Game might crash if plugins put something like a `Squid` in the `Builder`'s parameter. As you've said, you will expose pathfinders in phase 2, I hope you could give me suggestions to make this interface prepared for that change.

**Comment by liach** [comment]
Added vanilla AI task and removed mutex related stuff.

File: src/main/java/org/spongepowered/api/entity/ai/task/builtin/doorinteract/package-info.java (hunk 6 above)
Comment (liach) [inline src/main/java/org/spongepowered/api/entity/ai/task/builtin/doorinteract/package-info.java]: Fixed.

File: src/main/java/org/spongepowered/api/entity/ai/task/AITaskTypes.java [hunk 16]
```diff
@@ -25,37 +25,307 @@
...
 
 import org.spongepowered.api.entity.Entity;
-import org.spongepowered.api.entity.living.Agent;
-import org.spongepowered.api.entity.living.Creature;
-import org.spongepowered.api.entity.living.Humanoid;
-import org.spongepowered.api.entity.living.animal.Horse;
+import org.spongepowered.api.entity.Item;
+import org.spongepowered.api.entity.Tamer;
+import org.spongepowered.api.entity.living.*;
+import org.spongepowered.api.entity.living.animal.*;
```
liach: Well this is an accident. Will fix tomorrow.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
```diff
@@ -72,10 +73,10 @@
      * 3. Returning "true" will add the provided task to the list of updated tasks, if not there
      *    already.
      *
-     * @param other The other task
+     * @param that The other task
      * @return True if it can be, false if not
      */
-    boolean canRunConcurrentWith(AITask<O> other);
+    boolean canRunConcurrentWith(AITask<?> that);
```
Zidane: This still needs to be ```? super O```

liach: Accepted.

ST-DDT: Considering the situation that you are currently in
`LookIdleAITask extends AITask<Agent>` 
why wouldn't it be possible to execute 
`TargetAITask<A extends TargetAITask<A>> extends AITask<Creature>` concurrently?

````java
SwimmingAITask swim = ...;
FindNearestAttackableTargetAITask nearestAttackable = ...;
swim.canRunConcurrentWith(nearestAttackable); // Does not compile
````

Or is this fact silently cast away on the usage side? ([Yes it is](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/mixin/core/entity/ai/MixinEntityAITasks.java#L189))
If this fact is cast away silently then we can just use `<O>` anyway because we know, that we will only check tasks for the same entity/the owner of the task we currently use.
But `<?>` is the correct one if you ask me, why/where would you need the generic type here anyway?

Comment (liach) [inline src/main/java/org/spongepowered/api/entity/ai/task/AITask.java]: @ST-DDT Your idea is exactly the same as [mine](https://github.com/SpongePowered/SpongeAPI/pull/1012#discussion_r48972990). Status changed to pending, as this is just a minor issue.

Comment (Zidane) [inline src/main/java/org/spongepowered/api/entity/ai/task/AITask.java]: If its changed to the ```?``` then you put the pressure on the dev to ensure that the task you pass in is relevant to the task at hand.

At least with what is there, you will be confident that the task being passed in will have an owner that is of the same type of your task's owner (when I make it obey the contract, thanks @ST-DDT ).

With that said, thoughts?

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/entity/ai/task/AITask.java]: > If its changed to the ? then you put the pressure on the dev to ensure that the task you pass in is relevant to the task at hand.

Well, what are relevant tasks? IMO only those how could be executed concurrently which this method will determine.

> At least with what is there, you will be confident that the task being passed in will have an owner that is of the same type of your task's owner

I hope you will never check for a different entities' tasks with this method. IMO that method should only be called with matching owners. Using `<? super O>` is just garbage in the current setup because you cannot use the results any better than just `<?>` and you still have some of the compile errors from `<O>`. Using `<O>` has the potential future hazard that if you call a method that returns `XXXX<O>` for the other task you might get a `ClassCastException` since the real return type is only `XXXX<Agent>`. But more importantly. I don't see a usecase for being that specific since the only thing you can do with the generic type right now is the Owner, but since the owners match anyway you should not access the other one. On the other hand I must admit that there is currently no necessity to use `<?>` besides the compile issue. But the methods aren't supposed to be called that way anyway.

However there are a lot/some more generics issues in the AI API. (See also `Goal<O>.addTask(AITask<O> task)` that currently needs raw/unchecked casts, which is the entrance gate of evil)

We could solve this if we replace all methods parameters and results in `Goal` with <? super O> though => use `<?>` here
or replace all `___AITask<Creature>` with `___AITask<O extends Creature>` => Then we can use `<O>` here.
For all other solutions we have to do raw/unchecked casts in different places.

File: src/main/java/org/spongepowered/api/entity/ai/task/builtin/VanillaAITask.java [hunk 18]
```diff
@@ -0,0 +1,40 @@
...
+
+import org.spongepowered.api.entity.ai.task.AITask;
+import org.spongepowered.api.entity.living.Agent;
+
+/**
+ * Represents an {@link AITask} that exists in vanilla minecraft.
+ *
+ * @param <O> The type of agent
+ */
+public interface VanillaAITask<O extends Agent> extends AITask<O> {
```
ST-DDT: Add an `@Deprecated`since this is basically raw NMS behind the scenes.
Also add a setter here otherwise you cannot access/set the underlying mutex bits for custom impls.
Avoid extending this interface in the API even for the built in AITasks. That way we reduce/avoid breaking changes if we find a better solution to this issue.

Maybe you can also add a static helper method 
`static haveMutexClash(AITask<O>, AITask<O> )` (if either of them does not implement VanillaAITask return false)
and 
`static haveMutexClash(VanillaAITask<O>, VanillaAITask<O> )`

so we can do
````java
boolean canRunConcurrentWith(AITask<O> other) {
    return VanillaAITask.haveMutexClash(this. other);
}
````
and also add the constants here again so you can use them with the setter.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
Zidane: @ST-DDT 

```Well, what are relevant tasks? IMO only those how could be executed concurrently which this method will determine.```

Relevant being defined as a Task whose owner is, ideally, the same as this task's owner or, a bit worse, the same type as this task's owner.

If this was a "Vanilla only AI API", I'd not worry too much about some of this in the API and would ensure that any task you are passed would be the same owner as this task. Unfortunately, custom goals are an objective of the AI framework.

Strictly speaking, the compile errors are more of an implementation problem (as you said above) and not my problem as the API (not to be rude). This means ```? super O``` IS "better" than just ```?``` as you have an upper-bound. With just ```?```, any AITask with any type can be passed in.

With that said, I appreciate and get your sentiments (and am giving them serious thought). Even with the ability to do Custom Goals with the API, I may relax the reins a bit and let the plugin dev take heat if they start inter-twining tasks in such a way.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/VanillaAITask.java
Zidane: Why would you do @Deprecated here?

This can be done in such a way that you do NOT expose mutexes but do something akin to the categories. Categories aren't getting removed from Vanilla sooo.

src/main/java/org/spongepowered/api/entity/ai/task/AITask.java
JBYoshi: > Strictly speaking, the compile errors are more of an implementation problem (as you said above) and not my problem as the API (not to be rude). This means `? super O` IS "better" than just `?` as you have an upper-bound. With just ?, any AITask with any type can be passed in.

Question: What happens when you try to pass, say, `AITask<Monster>` to `AITask<Living>.canRunConcurrentWith()`? `Monster` is not a superinterface of `Living`.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/VanillaAITask.java
liach: But it would be a good idea to add `@Deprecated` to `getMutexBits()`. :+1:

liach:
Closing as retargeting to `bleeding` branch.

src/main/java/org/spongepowered/api/entity/ai/task/builtin/VanillaAITask.java
liach: Moreover we will talk about task categories in #1014. Feel free to post over there.

PR #1024, opened 2016-01-09
Add Fluid API representation

gabizou:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/410) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/489)
Forge has a Fluid API system that mods use to interact with each other for various purposes, and this exposes that system with the Data API. Plugins will be able to interact and interface with `TileEntity` based `IFluidHandler`s and `IFluidContainerItem`s.

The API itself was aided with @diesieben07's comments, and likewise the implementation of this will rely on Forge, with SpongeCommon providing some interfacing with vanilla based containers (namely buckets and such).

### DIFF
--- a/src/main/java/org/spongepowered/api/extra/fluid/package-info.java
+++ b/src/main/java/org/spongepowered/api/extra/fluid/package-info.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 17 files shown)


src/main/java/org/spongepowered/api/extra/fluid/FluidKeys.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.MapValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.util.Direction;
+
+import java.util.List;
+
+public final class FluidKeys {
```
simon816: Why separate from `Keys`?

liach: Because it is extra.

src/main/java/org/spongepowered/api/extra/fluid/package-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+/**
+ * The base Fluid API that attempts to bridge the Data API and Forges Fluid API.
+ * While ItemStacks are for inventories, FluidStacks are for
```
Zidane: y u leave me hanging.

SWEET GUTHIX, WHAT ARE FLUIDSTACKS FOR???

gabizou: THEY'RE LITERALLY JUST STACKS OF FLUID. JUST LIKE HOW ITEM STACKS ARE STACKS OF ITEMS.

PR #1031, opened 2016-01-11
Update for event gen changes

DDoS:
Merging this tonight unless someone objects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,12 +6,16 @@ buildscript {
             name = 'sponge'
             url = 'http://repo.spongepowered.org/maven'
         }
+        maven {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1,11 +1,16 @@
 // Gradle plugins
 buildscript {
     repositories {
+        mavenLocal()
```
gabizou: is this necessary?

DDoS: Forgot to remove that.

Minecrell: If it is then it's broken because that will never work on the Jenkins nor Travis ;)

DDoS: I know, it was just used for development, I forgot to remove it.

PR #1032, opened 2016-01-12
[WIP] ItemStack and TradeOffer Generators

gabizou:
This adds some dynamic "recipe" generators that can be used to not only generate `ItemStack`s and `TradeOffer`s. Namely, this adds `TradeOfferGenerator`s to be used and queried for with `Villager` `Career`s.

Granted, there was #662, but I feel that with the various "chance" utils introduced from #734 can be of better use and instead of adding new interfaces to populate an `ItemStack.Builder`, when a `BiConsumer` is perfectly acceptable, and not only does it achieve full customizability, but it's base intentions are to make semi-random generation simple. @Deamon5550 mentioned he'd be adding some utility methods to `WeightedTable` so that some cases where the `WeightedTable` does not have to be re-created every single invocation of `BiConsumer.apply` for the various `ItemStackBuilderPopulators` used.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
@@ -35,8 +35,10 @@ import org.spongepowered.api.block.tileentity.TileEntity;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.value.BaseValue;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.ResettableBuilder;
@@ -137,6 +139,21 @@ public interface ItemStack extends DataHolder, DataSerializable, Translatable {
          * @throws IllegalArgumentException If the quantity is outside the allowed bounds
          */
         Builder quantity(int quantity) throws IllegalArgumentException;
+
+        /**
+         * Adds a {@link Key} and related {@link Object} value to apply to the
+         * resulting {@link ItemStack}. Note that the resulting
+         * {@link ItemStack} may not actually accept the provided {@code Key}
+         * for various reasons due to support or simply that the value itself
+         * is not supported. Offering custom data is not supported through this,
+         * use {@link #itemData(DataManipulator)} instead.
+         *
+         * @param key The key to identiy the value to
+         * @param value The value to apply
+         * @param <E> The type of value
+         * @return This builder, for chaining
+         */
+        <E> Builder keyValue(Key<? extends BaseValue<E>> key, E value);
 
         /**
          * Sets the {@link DataManipulator} to add to the {@link ItemStack}.
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStackBuilderPopulators.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStackBuilderPopulators.java
@@ -1,0 +1,645 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spongepowered.api.util.weighted.VariableAmount.baseWithRandomAddition;
+import static org.spongepowered.api.util.weighted.VariableAmount.fixed;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.spongepowered.api.data.DataTransactionResult;
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.key.Keys;
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.data.value.BaseValue;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.SetValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.item.Enchantment;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.util.Tuple;
+import org.spongepowered.api.util.weighted.VariableAmount;
+import org.spongepowered.api.util.weighted.WeightedTable;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * A factory for generating {@link BiConsumer}s to apply to an
+ * {@link ItemStack.Builder}, usually through an {@link ItemStackGenerator}.
+ *
+ * <p>Note that the {@link BiConsumer}s are expected to utilize the passed in
+ * {@link Random} and use the builder as necessary.</p>
+ */
+public final class ItemStackBuilderPopulators {
+
+    /**
+     * Creates a new {@link BiConsumer} to set the {@link ItemStack.Builder}
+     * to use the provided {@link ItemStackSnapshot} as a "default". Note
+     * that the normal behavior of the builder is to reset according to
+     * the snapshot.
+     *
+     * @param snapshot The snapshot to set the builder to use
+     * @return The new biconsumer to apply to an itemstack builder
+     */
+    public static BiConsumer<ItemStack.Builder, Random> itemStack(ItemStackSnapshot snapshot) {
+        checkNotNull(snapshot, "ItemStackSnapshot cannot be null!");
+        return (builder, random) -> builder.fromSnapshot(snapshot);
+    }
+
+    /**
+     * Creates a new {@link BiConsumer} that uses a randomized selection
+     * of the provided {@link ItemStackSnapshot}s. The builder, when called will
+     * only use one at random selection to default to.
+     *
+     * @param snapshot The first snapshot
+     * @param snapshots The additional snapshots
+     * @return The new biconsumer to apply to an itemstack builder
+     */
+    public static BiConsumer<ItemStack.Builder, Random> itemStacks(ItemStackSnapshot snapshot, ItemStackSnapshot... snapshots) {
+        checkNotNull(snapshot, "ItemStackSnapshot cannot be null!");
+        final WeightedTable<ItemStackSnapshot> table = new WeightedTable<>(1);
+        table.add(snapshot, 1);
+        for (ItemStackSnapshot stackSnapshot : snapshots) {
+            table.add(checkNotNull(stackSnapshot, "ItemStackSnapshot cannot be null!"), 1);
+        }
+        return (builder, random) -> builder.fromSnapshot(table.get(random).get(0));
+    }
+
+    /**
+     * Creates a new {@link BiConsumer} that defines the provided
+     * {@link ItemType}.
+     *
+     * @param itemType The given item type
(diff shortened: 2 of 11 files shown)


src/main/java/org/spongepowered/api/item/inventory/generation/TestItemGenrator.java
```diff
@@ -0,0 +1,31 @@
...
+import org.spongepowered.api.item.Enchantments;
+import org.spongepowered.api.item.ItemTypes;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.format.TextColors;
+import org.spongepowered.api.text.format.TextStyles;
+
+public class TestItemGenrator {
+
+
+    public static ItemStackGenerator provide() {
```
gabizou: Note this is only provided as a test of functionality.

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackBuilderPopulators.java
```diff
@@ -0,0 +1,197 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Random;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public final class ItemStackBuilderPopulators {
+
```
gabizou: Should add a default `BiConsumer` taking an existing `ItemStack` for cases where plugins want to provide variance of mod-specific item data.

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackBuilderPopulators.java
```diff
@@ -0,0 +1,197 @@
...
+        return enchantment(VariableAmount.fixed(1), enchantment);
+    }
+
+    public static BiConsumer<ItemStack.Builder, Random> enchantment(VariableAmount level, Enchantment enchantment) {
+        checkNotNull(level, "VariableAmount cannot be null!");
+        checkNotNull(enchantment, "Enchantment cannot be null!");
+        return enchantments(VariableAmount.fixed(1), ImmutableList.of(new Tuple<>(enchantment, level)));
+    }
+
+    public static BiConsumer<ItemStack.Builder, Random> enchantmentsWithDefaultVariance(Collection<Enchantment> enchantments) {
```
gabizou: Probably needs a better name to coincide with what the method is actually doing.

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackBuilderPopulators.java
```diff
@@ -0,0 +1,197 @@
...
+        return enchantmentsWithDefaultVariance(VariableAmount.fixed(1), ImmutableList.copyOf(enchantments));
+    }
+
+    public static BiConsumer<ItemStack.Builder, Random> enchantmentsWithDefaultVariance(VariableAmount amount, Enchantment enchantment, Enchantment... enchantments) {
```
gabizou: Likewise here.

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackBuilderPopulators.java
```diff
@@ -0,0 +1,197 @@
...
+        return enchantmentsWithDefaultVariance(amount, ImmutableList.<Enchantment>builder().add(enchantment).addAll(Arrays.asList(enchantments)).build());
+    }
+
+    public static BiConsumer<ItemStack.Builder, Random> enchantmentsWithDefaultVariance(VariableAmount amount, Collection<Enchantment> itemEnchantments) {
```
gabizou: And here.

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackBuilderPopulators.java
```diff
@@ -0,0 +1,197 @@
...
+        checkNotNull(enchantments, "Supplier cannot be null!");
+        checkNotNull(amount, "VariableAmount cannot be null!");
+        return (builder, random) -> {
+            final ItemStack temp = builder.build();
+            final EnchantmentData enchantmentData = temp.getOrCreate(EnchantmentData.class).get();
+            ListValue<ItemEnchantment> enchantmentListValue = enchantmentData.enchantments();
+            final ImmutableList<Tuple<Enchantment, VariableAmount>> list = ImmutableList.copyOf(enchantments.get());
+            final int rollCount = amount.getFlooredAmount(random);
+            checkArgument(rollCount >= 1, "Must have at least 1 enchantment!");
+            WeightedTable<Tuple<Enchantment, VariableAmount>> table = new WeightedTable<>(rollCount);
```
gabizou: This will be moved outside of the `BiConsumer` when a new `WeightedTable` can be constructed with a `VariableAmount` for rolls.

Deamon5550:
`VariableAmount` rolls on weighted and chance tables are now in bleeding. eaa2a0eb02c4e7cad44a401d6098f26025fe0337

src/main/java/org/spongepowered/api/item/merchant/generation/TradeOfferGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+        this.firstItemGenerator = builder.firstGenerator;
+        this.secondItemGenerator = builder.secondGenerator;
+        this.sellingItemGenerator = builder.sellingGenerator;
+        this.experience = builder.experience;
+        this.baseUses = builder.baseUses;
+        this.maxUses = builder.maxUses;
+    }
+
+    @Override
+    public TradeOffer apply(Random random) {
```
kashike: `checkNotNull` on `random`

src/main/java/org/spongepowered/api/item/inventory/generation/ItemStackGenerator.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public Builder baseItem(ItemType itemType) {
+            this.baseItem = itemType;
+            return this;
+        }
+
+        public ItemStackGenerator build() {
+            checkState(this.baseItem != null || !this.consumers.isEmpty(), "Must have at least a defined amount of consumers or a base item type!");
+            return new ItemStackGenerator(this);
```
ST-DDT: Base Item is not used here.

ST-DDT:
IMO this use interfaces.
TradeOffer and ItemStackGenerator should be interfaces and having the current implementations renamed.
Or are planning to serialize this? Or is there any other need to use final classes?

ST-DDT:
How about the ability to upgrade existing trade offers?

gabizou:
> IMO this should use interfaces. TradeOffer and ItemStackGenerator should be interfaces and having the current implementations renamed.

It's absolutely useless, the only thing that is really necessary is representing an `ItemStackGenerator` and a `TradeOfferGenerator`. The latter of which could be an interface instead of an API final class, however, the `ItemStackGenerator` would very well do with being in the API as it's very straight forward. There are no real current implementations of either, except for the more obvious one for `TradeOfferGenerator`s which are the main point of the representation in the API.

> Or are planning to serialize this?

Not at all, they're generators, much like the [World Gen Populator API](https://github.com/SpongePowered/SpongeAPI/pull/734), they're configurable but not serializable, based on constructs and `BiFunction`s that are entirely programmable, but not limited to being generated through configuration by a plugin.

> Or is there any other need to use final classes?

Aside from control, the `ItemStackGenerator` doesn't really need to be declared as a class in the API, but given that it does it's job very well, it's a nice addition.



> How about the ability to upgrade existing trade offers?

See [VillagerRegistry::populateOffers](https://github.com/SpongePowered/SpongeAPI/pull/1032/files#diff-1535a6cdeae2033783d651e11faa3287R44).

ST-DDT:
>> How about the ability to upgrade existing trade offers?

> See VillagerRegistry::populateOffers.

That impl and the entire TradeOfferGenerator design is for appendind trade offers not altering existing ones.

gabizou:
> altering existing ones.

Notice that `TradeOffer` is immutable, which kinda makes it difficult to say "Go ahead and change these trade offers with these generators."

ST-DDT:
@gabizou You are over simplifying the matter.

`TradeOffer`s are immutable, but the list they are stored in isn't.

````java
TradeOffer old = offers.get(x);
TradeOffer.Builder builder = TradeOffer.builder().from(old);
builder.maxUses(old.getMaxUses() + 1);
builder.uses(old.getMaxUses() + 1);
TradeOffer updated = builder.build();
offers.set(x, updated);
````

**Comment by gabizou** [comment]
That's easy to do with working with the data value from `Keys.TRADE_OFFERS`. Not something the `VillagerRegistry` should be doing on it's own.

**Comment by ST-DDT** [comment]
@gabizou It isn't doing anything on its own. It just passes the list of existing trade offers to the `TradeOfferGenerator` (trade offer upgrade tasks), which explicitly do that/are told to do that.

Another benefit is that you can add a group of `TradeOffer`s instead of just one.
For example make an armourer skilled in a specific `Enchantment` and add the next level on upgrade.

> Level 1: Unenchanted leather armor
Level 2: Unenchanted iron armor
Level 3: Unenchanted diamond armor
Level 4: Diamond armor with protection 1 // Replace protection with any `Enchantment`
Level 5: Diamond armor with protection 2
Level 6: Diamond armor with protection 3
Level 7: Diamond armor with protection 4
....

File: src/main/java/org/spongepowered/api/item/inventory/ItemStackBuilderPopulators.java [hunk 9]
```diff
@@ -0,0 +1,236 @@
...
+    public static BiConsumer<ItemStack.Builder, Random> items(final Supplier<Collection<ItemType>> supplier) {
+        return (builder, random) -> {
+            final ImmutableList<ItemType> collection = ImmutableList.copyOf(supplier.get());
+            builder.itemType(collection.get(random.nextInt(collection.size())));
+        };
+    }
+
+    public static BiConsumer<ItemStack.Builder, Random> quantity(VariableAmount amount) {
+        checkNotNull(amount, "VariableAmount cannot be null!");
+        return ((builder, random) -> builder.quantity(amount.getFlooredAmount(random)));
```
phroa: I think you've got an extra layer of parens here

gabizou: I'm pretty sure I don't.

gabizou:
> @gabizou It isn't doing anything on its own. It just passes the list of existing trade offers to the TradeOfferGenerator (trade offer upgrade tasks), which explicitly do that/are told to do that.
> @gabizou It isn't doing anything on its own. It just passes the list of existing trade offers to the TradeOfferGenerator (trade offer upgrade tasks), which explicitly do that/are told to do that.
> 
> Not the same but similar to this: #662: TradeOfferGenerator.
> 
> Another benefit is that you can add a group of TradeOffers instead of just one.
> For example make an armourer skilled in a specific Enchantment and add the next level on upgrade.
> 
> Level 1: Unenchanted leather armor
> Level 2: Unenchanted iron armor
> Level 3: Unenchanted diamond armor
> Level 4: Diamond armor with protection 1 // Replace protection with any Enchantment
> Level 5: Diamond armor with protection 2
> Level 6: Diamond armor with protection 3
> Level 7: Diamond armor with protection 4
> ....
> This is possible with NMS generators.


On it's own, the implementation has a custom interface that does modify a `List<TradeOffer>` with the use of a `Random` object. However, specifying that it's a `TradeOfferGenerator` is not within the contract of that interface. For all I care, it can just mutate the current lists by removing all `TradeOffer`s after level 1.

I've since revised the API PR to include both a `TradeOfferGenerator` and the ability to have `TradeOfferListMutator`s that are accessible in the `VillagerRegistry`.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -139,6 +141,20 @@ static ItemStack of(ItemType itemType, int quantity) {
         Builder quantity(int quantity) throws IllegalArgumentException;
 
         /**
+         * Adds a {@link Key} and related {@link Object} value to apply to the
+         * resulting {@link ItemStack}. Note that the resulting
+         * {@link ItemStack} may not actually accept the provided {@code Key}
+         * for various reasons due to support or simply that the value itself
+         * is not supported.
```
ST-DDT: What happens if the keyValue isnt supported?
Will they be discarded or will they cause an Exception during build()?

src/main/java/org/spongepowered/api/item/merchant/VillagerRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Gets an immutable {@link Multimap} of the {@link Career}'s registered
+     * {code level} to {@link TradeOfferGenerator}s. Note that the map is
+     * immutable and cannot be modified directly.
+     *
+     * @param career The career
+     * @return The immutable multimap
+     */
+    Multimap<Integer, TradeOfferGenerator> getTradeOfferLevelMap(Career career);
```
ST-DDT: These methods need to return `TradeOfferListMutator`s instead of `TradeOfferGenerator`s.

src/main/java/org/spongepowered/api/item/merchant/VillagerRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+     * {@link Career} and {@code level}. Note that the level
+     * must be at least 1. There can be multiple {@link TradeOfferGenerator}s
+     * per level.
+     *
+     * @param career The career
+     * @param level The level
+     * @param generator The generator to register
+     * @return This registry, for chaining
+     */
+    VillagerRegistry addGenerator(Career career, int level, TradeOfferGenerator generator);
```
ST-DDT: The add and set methods can be default implemented as well. (Except the last)

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
gabizou: They'll be discarded. Good to note on the javadocs.

src/main/java/org/spongepowered/api/item/merchant/VillagerRegistry.java
gabizou: I'm hesitant as it's possible to stipulate that there's a default set of `TradeOfferGenerator`s *AND* a set of `TradeOfferListMutator`s.

src/main/java/org/spongepowered/api/item/inventory/ItemStackGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Random;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+public interface ItemStackGenerator extends Function<Random, ItemStack> {
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
```
ST-DDT: There should be a shortcut method to create an `ItemStackGenerator` from an existing stack or snaphot.

````java
static ItemStackGenerator of(ItemStack stack) {
    return random -> stack
}

static ItemStackGenerator randomOf(Iterable<ItemStack> stacks) {
    List copy = new List(stacks);
    int size = copy.size();
    return random -> copy.get(random.nextInt(size));
}
````

src/main/java/org/spongepowered/api/item/merchant/VillagerRegistry.java
ST-DDT: Then add a new interface for TradeOfferListMutators that consist only of a single TradeOfferGenerator and note in the javadocs that they always append.

ST-DDT: ^ Updated

PR #1041, opened 2016-01-18
[WIP]Conversation API

intronate67:
# ConversationAPI
I've seen very little talk about a conversation API and it did not seem that hard to try to port over Bukkit's API to work with Sponge. Now if this isn't the direction you guys wanted for the conversation API then it doesn't have to go through. But i feel like this is a good option because 1. A lot of developers are already used it. 2. it can be improved throughout the beta. & 3. It is a very easy system to use. 

## How it works.
Basically the same as Bukkit's Conversation API
untitled drawing
 
* All sub prompts extend prompt. So all custom prompts will need to implement these methods.

```java
@Override
public Text getPromptText(ConversationContext context) {
    return Text.of("Message Here");
}
```

* And some method to accept a valid input of primitive type.

#### String Prompt
- Accepts any type of string inputted by user.

#### Message Prompt
- Does not accept user input. Sends a message to the player.

#### Validating Prompt - 5 Sub types
** Chooses from a preset type and input **
- Boolean Prompt
- Numeric Prompt
- Regex Prompt
- PlayerName Prompt
- FixedSet Prompt

## How to use

Right now i have been using this inside my execute method for commands, like as such

```java
private MAIN_CLASSNAME plugin;
private ConversationFactory factory = new ConversationFactory(plugin.getGame());

private Conversation conv;
public CommandResult execute(CommandSource src, CommandContext args){
     if(!(src instanceof Player)){
            src.sendMessage(Text.of(TextColors.RED, "Only players can run this command!"));
       }
      Player p = (Player)src;
      conv = factory.withFirstPrompt(new CustomPrompt())).withLocalEcho(false).buildConversation((Conversable) p);
      conv.begin();
}

with CustomPrompt.java looking like

```java
public class CustomPrompt extends StringPrompt {

@Override
    public Text getPromptText(ConversationContext context) {
        return Text.of("Say your name");
    }

    @Override
    public Prompt acceptInput(ConversationContext context, String input) {
        context.getForWhom().sendRawMessage(Text.of("Hello ", input));
        return null;
    }

}

### DIFF
--- a/src/main/java/org/spongepowered/api/conversation/Conversation.java
+++ b/src/main/java/org/spongepowered/api/conversation/Conversation.java
@@ -32,6 +32,31 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+/**
+ * The Conversation class is responsible for tracking the current state of a
+ * conversation, displaying prompts to the user, and dispatching the user's
+ * response to the appropriate place. Conversation objects are not typically
+ * instantiated directly. Instead a {@link ConversationFactory} is used to
+ * construct identical conversations on demand.
+ * <p>
+ * Conversation flow consists of a directed graph of {@link Prompt} objects.
+ * Each time a prompt gets input from the user, it must return the next prompt
+ * in the graph. Since each Prompt chooses the next Prompt, complex
+ * conversation trees can be implemented where the nature of the player's
+ * response directs the flow of the conversation.
+ * <p>
+ * Each conversation has a {@link ConversationPrefix} that prepends all output
+ * from the conversation to the player. The ConversationPrefix can be used to
+ * display the plugin name or conversation status as the conversation evolves.
+ * <p>
+ * Each conversation has a timeout measured in the number of inactive seconds
+ * to wait before abandoning the conversation. If the inactivity timeout is
+ * reached, the conversation is abandoned and the user's incoming and outgoing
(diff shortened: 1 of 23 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/api/conversation/Conversation.java [hunk 1]
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /**
+     * Passes player input into the current prompt. The next prompt (as
+     * determined by the current prompt) is then displayed to the user.
+     *
+     * @param input The user's chat text.
+     */
+    public void acceptInput(String input) {
+        try { // Spigot
```
simon816: Spigot?

mbax:
You forgot to mention the part where this code is under the GPL license, and **isn't yours**

ryantheleach:
Having a conversation API is a good thing, but unless you are able to contact the original contributors and able to get it re-licensed, it can't be copied whatsoever.

intronate67:
Ah, now i see in section 5 of the  GPL license

mbax:
Not to mention the part where you were pretending this code was your own work.

intronate67:
I don't know where I said that. If it was implied like that, it wasn't intentional. Simply just wanted to port over Bukkits conversation API to work with Sponge like it says above. But now that cant happen with current circumstances.

PR #1043, opened 2016-01-18
Add ChatType methods to MessageChannel

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1043) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/445) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/523) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/221)

### DIFF
--- a/src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
+++ b/src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
@@ -24,18 +24,19 @@
  */
 package org.spongepowered.api.text.channel;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Set;
 import java.util.WeakHashMap;
 
 /**
- * An abstract implementation of {@link MutableMessageChannel}, using a
- * {@link Set} as the underlying member list
+ * An abstract implementation of {@link MutableMessageChannel}.
  */
 public abstract class AbstractMutableMessageChannel implements MutableMessageChannel {
 
-    protected final Set<MessageReceiver> members;
+    protected final Collection<MessageReceiver> members;
 
     /**
      * The default implementation uses a {@link WeakHashMap} implementation of {@link Set}.
@@ -46,26 +47,28 @@ public abstract class AbstractMutableMessageChannel implements MutableMessageCha
 
     /**
      * Creates a new instance of {@link AbstractMutableMessageChannel} with the
-     * provided {@link Set} as the underlying member list.
+     * provided {@link Collection} as the underlying member list.
      *
-     * <p>The passed {@link Set} directly affects the members of this channel.</p>
+     * <p>The passed {@link Collection} directly affects the members of this channel.</p>
      *
-     * <p>It is recommended to use a weak set to avoid memory leaks. If you do not use
-     * a weak set, please ensure that members are cleaned up properly.</p>
+     * <p>It is recommended to use a weak collection to avoid memory leaks. If you do not
+     * use a weak collection, please ensure that members are  cleaned up properly.</p>
      *
-     * @param members The set of members
+     * @param members The collection of members
      */
-    protected AbstractMutableMessageChannel(Set<MessageReceiver> members) {
-        this.members = members;
+    protected AbstractMutableMessageChannel(Collection<MessageReceiver> members) {
+        this.members = checkNotNull(members, "members");
     }
 
     @Override
     public boolean addMember(MessageReceiver member) {
-        return this.members.add(member);
+        checkNotNull(member, "member");
+        return !this.members.contains(member) && this.members.add(member);
     }
 
     @Override
     public boolean removeMember(MessageReceiver member) {
+        checkNotNull(member, "member");
         return this.members.remove(member);
     }
 
--- a/src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
+++ b/src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
@@ -33,6 +33,8 @@ import org.spongepowered.api.text.channel.impl.DelegateMutableMessageChannel;
 import org.spongepowered.api.text.channel.type.CombinedMessageChannel;
 import org.spongepowered.api.text.channel.type.FixedMessageChannel;
 import org.spongepowered.api.text.channel.type.PermissionMessageChannel;
+import org.spongepowered.api.text.chat.ChatType;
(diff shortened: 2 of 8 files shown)


src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -134,15 +137,45 @@ default void send(Text original) {
...
+     * Sends a message to this channel, transforming and sending it to
+     * the members.
+     *
+     * <p>By default, the {@link ChatTypes#SYSTEM} type is used when sending
+     * messages to applicable recipients of this channel.</p>
+     *
+     * @param original The original text to send
+     * @param type The type of message
+     */
+    default void send(Text original, ChatType type) {
```
ST-DDT: IMO. The chat type should always be the same for a MessageChannel/the caller should not be the one to controll it.

kashike: `send` methods should be able to control it - a message channel is a way to distribute and transform messages in bulk, there's no reason at all this should not be supported.

src/main/java/org/spongepowered/api/text/ChatTypeMessageReceiver.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Sends the message(s) with the specified {@link ChatType} on the client.
+     *
+     * @param type The chat type to send the messages to
+     * @param messages The message(s) to send
+     */
+    default void sendMessages(ChatType type, Iterable<Text> messages) {
+        for (Text message : checkNotNull(messages, "messages")) {
```
ST-DDT: Shouldn't we check here for an empty list of messages as well? This would be more consistent with the method above.

phroa: Is it really an error though?

windy1: It won't matter anyways, in fact it's less efficient to check the size.

phroa: +1

kashike: @gabizou wanted it originally, but I'll remove it

gabizou: Just keep the check that the messages are not null and the type.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -134,15 +136,45 @@ default void send(Text original) {
...
+
+    /**
+     * Send a message to this channel, transforming and sending it to the members.
+     *
+     * @param sender The sender of the message
+     * @param original The original message to send
+     * @param type The type of message
+     */
+    default void send(@Nullable Object sender, Text original, ChatType type) {
         checkNotNull(original, "original text");
```
Minecrell: `original` :)

kashike:
@Minecrell could you look at merging this soon? @gabizou is attacking me for not having another PR done which depends on this :sos:

Minecrell:
It looks fine to me. I can either merge it tomorrow or you can try to yell at someone else to do that ;)

src/main/java/org/spongepowered/api/text/channel/impl/DelegateMessageChannel.java
```diff
@@ -0,0 +1,60 @@
...
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.text.channel.MessageReceiver;
+import org.spongepowered.api.text.chat.ChatType;
+
+import java.util.Collection;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
```
windy1: This doc is very ambiguous. What's the use case? Where is it used? Why is it necessary?

src/main/java/org/spongepowered/api/text/channel/impl/SimpleMutableMessageChannel.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.spongepowered.api.text.channel.impl;
+
+import org.spongepowered.api.text.channel.AbstractMutableMessageChannel;
+import org.spongepowered.api.text.channel.MessageReceiver;
+import org.spongepowered.api.text.channel.MutableMessageChannel;
+
+import java.util.Collection;
+
+/**
```
windy1: Same as above.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
```diff
@@ -24,18 +24,19 @@
...
 
 /**
- * An abstract implementation of {@link MutableMessageChannel}, using a
- * {@link Set} as the underlying member list
+ * An abstract implementation of {@link MutableMessageChannel}.
  */
 public abstract class AbstractMutableMessageChannel implements MutableMessageChannel {
 
-    protected final Set<MessageReceiver> members;
+    protected final Collection<MessageReceiver> members;
```
windy1: Why the change to Collection? Should duplicate MessageReceivers be permitted?

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
```diff
@@ -46,26 +47,28 @@ protected AbstractMutableMessageChannel() {
...
-        this.members = members;
+    protected AbstractMutableMessageChannel(Collection<MessageReceiver> members) {
+        this.members = checkNotNull(members, "members");
     }
 
     @Override
     public boolean addMember(MessageReceiver member) {
-        return this.members.add(member);
+        checkNotNull(member, "member");
+        return !this.members.contains(member) && this.members.add(member);
```
windy1: This is why a Set was being used. Sets don't allow duplicates.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
kashike: It's up to the developer to avoid this - by default a `WeakHashMap`-based `Set` is used.

src/main/java/org/spongepowered/api/text/channel/AbstractMutableMessageChannel.java
kashike: I wrote this originally, see above comment.

PR #1056, opened 2016-01-30
Add Pitch values for sound effects

windy1:
Example usage:

```java
world.playSound(SoundType.NOTE_PIANO, position, 1, Pitch.C1);
```

will play a middle C.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/MutableBlockVolume.java
+++ b/src/main/java/org/spongepowered/api/world/extent/MutableBlockVolume.java
@@ -131,6 +131,45 @@ public interface MutableBlockVolume extends BlockVolume {
     }
 
     /**
+     * Sets the block at the given position in the world with the provided
+     * {@link Cause} will be used for any events thrown. Note that the
+     * difference between this an {@link #setBlockType(Vector3i, BlockType)} is
+     * that no block tracking chaining will take place. Note that there is
+     * a requirement that the {@link PluginContainer} of the plugin calling
+     * this method is <strong>REQUIRED</strong>.
+     *
+     * @param position The position
+     * @param type The block type
+     * @param cause The cause to use
+     * @throws PositionOutOfBoundsException If the position is outside of the
+     *         bounds of the volume
+     */
+    default void setBlockType(Vector3i position, BlockType type, Cause cause) {
+        setBlock(position.getX(), position.getY(), position.getZ(), type.getDefaultState(), cause);
+    }
+
+    /**
+     * Sets the block at the given position in the world with the provided
+     * {@link Cause} will be used for any events thrown. Note that the
+     * difference between this an {@link #setBlockType(Vector3i, BlockType)} is
+     * that no block tracking chaining will take place. Note that there is
+     * a requirement that the {@link PluginContainer} of the plugin calling
+     * this method is <strong>REQUIRED</strong>.
+     *
+     * @param x The X position
+     * @param y The Y position
+     * @param z The Z position
+     * @param type The block
+     * @param cause The cause to use
+     * @throws PositionOutOfBoundsException If the position is outside of the
+     *         bounds of the volume
+     */
+    default void setBlockType(int x, int y, int z, BlockType type, Cause cause) {
+        setBlock(x, y, z, type.getDefaultState(), cause);
+    }
+
+
+    /**
      * Returns a new volume that is the same or smaller than the current volume.
      * This does not copy the blocks, it only provides a new view of the
      * storage.


simon816:
There already is a [`NotePitches`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/type/NotePitches.java) class

windy1:
@simon816 Yes I understand, this is different though. The note block pitches are a value between 0 and 24 I believe while the sound effect pitches are between 0 and 2.

simon816:
Perhaps there could just be a method `NotePitch#asDouble()` or something?
`world.playSound(SoundType.NOTE_PIANO, position, 1, NotePitches.C1.asDouble());`

gabizou:
Note that by default, the pitch is actually separate from `NotePitch`es since the sounds are normally played at `1`, any higher or lower alters the pitch of the sound itself (in various ways, some sounds are played slower and some sounds are played faster).

windy1:
I'll change it if you guys want I just think this way is more precise and clear. A developer is more likely to get confused if we combine these two things that are in reality two entirely different functions of the MC protocol.

gabizou:
Perhaps you could rename `Pitch` to `PitchModulation` since that's realistically what it's doing (nothing to do with setting the sound's pitch, but more or less modifying the pitch).

windy1:
@gabizou Commit updated

src/main/java/org/spongepowered/api/effect/sound/PitchModulation.java
```diff
@@ -0,0 +1,56 @@
+package org.spongepowered.api.effect.sound;
```
simon816: Missing license header

src/main/java/org/spongepowered/api/effect/sound/PitchModulation.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.effect.Viewer;
+
+/**
+ * Represents a pitch value for a {@link SoundType}. Minecraft's accepted pitches span two octaves from F#0 to F#2.
```
simon816: Maximum javadoc line length is 80

src/main/java/org/spongepowered/api/effect/sound/PitchModulation.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @see Viewer#playSound(SoundType, Vector3d, double)
+ */
+public final class PitchModulation {
+    public static double FSHARP0 = 0;
```
simon816: Need newline after class declaration

src/main/java/org/spongepowered/api/effect/sound/PitchModulation.java
```diff
@@ -0,0 +1,56 @@
...
+    public static double AFLAT1 = GSHARP1;
+    public static double BFLAT1 = ASHARP1;
+    public static double CFLAT1 = B1;
+    public static double DFLAT2 = CSHARP2;
+    public static double EFLAT2 = DSHARP2;
+    public static double FFLAT2 = E2;
+    public static double GFLAT2 = FSHARP2;
+
+    private PitchModulation() {
+    }
```
simon816: Need newline after last member in the class. (before final `}`)

windy1:
@simon816 updated

Cybermaxke:
The fields should be final

windy1:
whoops don't know how i missed that. fixed.

windy1:
Do not pull this yet. The values are out of tune.

windy1:
Okay fixed now.

src/main/java/org/spongepowered/api/effect/sound/PitchModulation.java
```diff
@@ -0,0 +1,83 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.effect.sound;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.effect.Viewer;
+
+/**
+ * Represents a pitch value for a {@link SoundType}. Minecraft's accepted
```
gabizou: probably should update the javadoc for what these do ;)

windy1:
@gabizou updated

gabizou:
Wait why was this closed? @windy1

windy1:
I screwed up my master branch on my fork. I'll resubmit it when i get home later tonight.

PR #1058, opened 2016-02-02
Switch from using ListenableFuture to CompletableFuture

zml2008:
**SpongeAPI PR | [SpongeCommon PR](https://github.com/SpongePowered/SpongeCommon/pull/473)**

CompletableFuture is the jdk8 equivalent of guava's ListenableFuture
class, with similar capabilities but better lambda integration.

There shouldn't be any reason to prefer guava classes over java classes, but this PR is here in case anyone wants to discuss that.

### DIFF
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -25,7 +25,6 @@
 package org.spongepowered.api;
 
 import com.flowpowered.math.vector.Vector3d;
-import com.google.common.util.concurrent.ListenableFuture;
 import org.spongepowered.api.entity.Entity;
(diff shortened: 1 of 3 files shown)


kashike:
@zml2008 what would the equivalent of this be using `CompletableFuture`s?
http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/Futures.html#addCallback(com.google.common.util.concurrent.ListenableFuture,%20com.google.common.util.concurrent.FutureCallback)

zml2008:
@kashike probably something like https://gist.github.com/anonymous/89f6c2093204bc82c40a -- untested, you'd have to mess around a bit.

PR #1061, opened 2016-02-02
Add ability to restrict commands based on the CommandSource type

windy1:
Example usage:

```java
CommandSpec foo = CommandSpec.builder()
    .sourceType(Player.class)
    .executor(new FooExecutor())
    .build();

class FooExecutor extends CommandExecutor<Player> {
    public CommandResult execute(Player player, CommandContext context) {
        // blah blah blah
    }
}
```

Note that while `CommandExecutor#execute(T, CommandContext);`'s signature has been changed, this is *not* a breaking change for old implementations.

SpongeCommon PR: https://github.com/SpongePowered/SpongeCommon/pull/477

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandCallable.java
+++ b/src/main/java/org/spongepowered/api/command/CommandCallable.java
@@ -79,6 +79,20 @@ public interface CommandCallable {
     boolean testPermission(CommandSource source);
 
     /**
+     * Test whether this command is allowed to be executed by the specified
+     * source depending on it's type.
+     *
+     * <p>Implementations should use this to restrict command execution of
+     * certain command source types. This varies from permissions in that
+     * a source may have permission to execute a command but must be, for
+     * example, a Player because the command uses locational data.</p>
+     *
+     * @param source called of the command
+     * @return whether the source type is permitted to execute the command
+     */
+    boolean testSourceType(CommandSource source);
+
+    /**
      * Get a short one-line description of this command.
      *
      * <p>The help system may display the description in the command list.</p>
--- a/src/main/java/org/spongepowered/api/command/CommandSourceTypeException.java
+++ b/src/main/java/org/spongepowered/api/command/CommandSourceTypeException.java
@@ -22,24 +22,43 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.command.spec;
+package org.spongepowered.api.command;
 
-import org.spongepowered.api.command.CommandException;
-import org.spongepowered.api.command.CommandResult;
-import org.spongepowered.api.command.CommandSource;
-import org.spongepowered.api.command.args.CommandContext;
+import org.spongepowered.api.text.Text;
 
 /**
- * Interface containing the method directing how a certain command will be executed.
+ * Exception thrown when a CommandSource tries to execute a command that cannot
+ * be executed by that CommandSource type.
  */
-public interface CommandExecutor<T extends CommandSource> {
+public class CommandSourceTypeException extends CommandException {
+
+    private static final long serialVersionUID = -4945570277201224948L;
+
     /**
-     * Callback for the execution of a command.
+     * Creates a new exception with the specified expected class.
      *
-     * @param source The commander who is executing this command
-     * @param args The parsed command arguments for this command
-     * @return the result of executing this command
-     * @throws CommandException If a user-facing error occurs while executing this command
+     * @param expected class of source type
      */
-    CommandResult execute(T source, CommandContext args) throws CommandException;
+    public CommandSourceTypeException(Class<? extends CommandSource> expected) {
+        super(Text.of("Only users of type " + expected.getName() + " may execute this command."));
+    }
+
+    /**
+     * Creates a new exception with the specified message.
+     *
+     * @param message to display
+     */
+    public CommandSourceTypeException(Text message) {
+        super(message);
+    }
+
+    /**
+     * Creates a new exception with the specified message and original cause.
+     *
+     * @param message to display
+     * @param cause original cause
+     */
+    public CommandSourceTypeException(Text message, Throwable cause) {
+        super(message, cause);
+    }
 }
--- a/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
+++ b/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
@@ -25,19 +25,18 @@
 package org.spongepowered.api.command.spec;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 import static org.spongepowered.api.command.args.GenericArguments.firstParsing;
 import static org.spongepowered.api.command.args.GenericArguments.optional;
+import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
-import org.spongepowered.api.text.Text;
 import org.spongepowered.api.command.CommandCallable;
 import org.spongepowered.api.command.CommandException;
-import org.spongepowered.api.command.CommandMessageFormatting;
 import org.spongepowered.api.command.CommandPermissionException;
 import org.spongepowered.api.command.CommandResult;
 import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.command.CommandSourceTypeException;
 import org.spongepowered.api.command.args.ArgumentParseException;
 import org.spongepowered.api.command.args.ChildCommandElementExecutor;
 import org.spongepowered.api.command.args.CommandArgs;
@@ -46,6 +45,7 @@ import org.spongepowered.api.command.args.CommandElement;
 import org.spongepowered.api.command.args.GenericArguments;
 import org.spongepowered.api.command.args.parsing.InputTokenizer;
 import org.spongepowered.api.command.args.parsing.InputTokenizers;
+import org.spongepowered.api.text.Text;
 
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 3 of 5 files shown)


src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -228,6 +235,33 @@ public Builder inputTokenizer(InputTokenizer parser) {
...
+         * deep is true than the specified class and all subsequent sub-classes
+         * will be permitted to execute the command.
+         *
+         * @param senderType type of command source
+         * @param deep whether the type's sub-classes should be included
+         * @return this
+         */
+        public Builder sourceType(Class<? extends CommandSource> senderType, boolean deep) {
+            this.sourceType = senderType;
+            sourceTypeDeep = deep;
```
simon816: Qualify with `this.`

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -273,6 +307,20 @@ public void checkPermission(CommandSource source) throws CommandException {
...
+     * Checks if the specified CommandSource is permitted to execute this
+     * command based upon their type.
+     *
+     * @param source source to check
+     * @throws CommandException if the source is not permitted to execute the command
+     */
+    public void checkSourceType(CommandSource source) throws CommandException {
+        checkNotNull(source, "source");
+        if (!testSourceType(source)) {
+            throw new CommandSourceTypeException(sourceType);
```
simon816: `this.sourceType`

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -343,6 +392,12 @@ public boolean testPermission(CommandSource source) {
         return this.permission == null || source.hasPermission(this.permission);
     }
 
+    @Override
+    public boolean testSourceType(CommandSource source) {
+        Class<? extends CommandSource> clazz = source.getClass();
+        return sourceTypeDeep ? sourceType.isAssignableFrom(clazz) : sourceType.equals(clazz);
```
simon816: `this.sourceType` and `this.sourceTypeDeep`

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -254,7 +288,7 @@ public CommandSpec build() {
             }
 
             return new CommandSpec(this.args, this.executor, this.description, this.extendedDescription, this.permission,
-                    this.argumentParser);
+                    this.argumentParser, sourceType, sourceTypeDeep);
```
simon816: `this.sourceType` and `this.sourceTypeDeep`

windy1:
Fixed @simon816

src/main/java/org/spongepowered/api/command/CommandSourceTypeException.java
```diff
@@ -0,0 +1,63 @@
...
+package org.spongepowered.api.command;
+
+import org.spongepowered.api.text.Text;
+
+/**
+ * Exception thrown when a CommandSource tries to execute a command that cannot
+ * be executed by that CommandSource type.
+ */
+public class CommandSourceTypeException extends CommandException {
+    private static final long serialVersionUID = -4945570277201224948L;
```
kashike: New line above this.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -46,17 +45,18 @@
 import org.spongepowered.api.command.args.GenericArguments;
 import org.spongepowered.api.command.args.parsing.InputTokenizer;
 import org.spongepowered.api.command.args.parsing.InputTokenizers;
+import org.spongepowered.api.text.Text;
 
+import javax.annotation.Nullable;
```
kashike: This shouldn't be moved from where it was.

src/main/java/org/spongepowered/api/command/spec/CommandExecutor.java
```diff
@@ -41,5 +41,5 @@
      * @return the result of executing this command
      * @throws CommandException If a user-facing error occurs while executing this command
      */
-    CommandResult execute(CommandSource src, CommandContext args) throws CommandException;
+    CommandResult execute(T src, CommandContext args) throws CommandException;
```
kashike: Rename `src` to `source`? It will match args in other command-related ifaces/classes.

windy1: I didn't change that, but I will rename it if you want.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -228,6 +237,33 @@ public Builder inputTokenizer(InputTokenizer parser) {
...
+         * Restricts the command to the specified CommandSource sub-class. If
+         * deep is true than the specified class and all subsequent sub-classes
+         * will be permitted to execute the command.
+         *
+         * @param senderType type of command source
+         * @param deep whether the type's sub-classes should be included
+         * @return this
+         */
+        public Builder sourceType(Class<? extends CommandSource> senderType, boolean deep) {
+            this.sourceType = senderType;
```
kashike: `checkNotNull(senderType, "sender type");`

src/main/java/org/spongepowered/api/command/CommandCallable.java
```diff
@@ -79,6 +79,20 @@
...
     /**
+     * Test whether this command is allowed to be executed by the specified
+     * source depending on it's type.
+     *
+     * <p>Implementations should use this to restrict command execution of
+     * certain command source types. This varies from permissions in that
+     * a source may have permission to execute a command but must be, for
+     * example, a Player because the command uses locational data.</p>
+     *
+     * @param source called of the command
```
simon816: Is this supposed to be "Caller of the command"? (Also note the first word after param name must be capitalized)

src/main/java/org/spongepowered/api/command/CommandCallable.java
```diff
@@ -79,6 +79,20 @@
...
+     * @return whether the source type is permitted to execute the command
```
simon816: "Whether" (capitals)

src/main/java/org/spongepowered/api/command/CommandCallable.java
```diff
@@ -79,6 +79,20 @@
     boolean testPermission(CommandSource source);
 
     /**
+     * Test whether this command is allowed to be executed by the specified
```
simon816: "Tests", not "Test"

src/main/java/org/spongepowered/api/command/CommandSourceTypeException.java
```diff
@@ -0,0 +1,64 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.command;
+
+import org.spongepowered.api.text.Text;
+
+/**
+ * Exception thrown when a CommandSource tries to execute a command that cannot
```
simon816: `{@link CommandSource}`

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -57,23 +57,30 @@
...
  */
+@SuppressWarnings({"rawtypes", "unchecked"})
 public final class CommandSpec implements CommandCallable {
 
     private final CommandElement args;
     private final CommandExecutor executor;
     private final Optional<Text> description;
-    @Nullable private final Text extendedDescription;
+    @Nullable
+    private final Text extendedDescription;
```
simon816: This can stay on the same line as the annotation

src/main/java/org/spongepowered/api/command/CommandCallable.java
```diff
@@ -79,6 +79,20 @@
...
+     *
+     * <p>Implementations should use this to restrict command execution of
+     * certain command source types. This varies from permissions in that
+     * a source may have permission to execute a command but must be, for
+     * example, a Player because the command uses locational data.</p>
+     *
+     * @param source called of the command
+     * @return whether the source type is permitted to execute the command
+     */
+    boolean testSourceType(CommandSource source);
```
ST-DDT: This should be default implemented and return true to be less breaking.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -254,7 +291,7 @@ public CommandSpec build() {
             }
 
             return new CommandSpec(this.args, this.executor, this.description, this.extendedDescription, this.permission,
```
dualspiral: Should there not be a check here that ensures that the `CommandExecutor<T>` and the specified source type match up? It's probably not so much of a problem if you have a `Player` source type and a `CommandExecutor<CommandSource>` executor, but if you have a `Player` source type and a `CommandExecutor<ConsoleSource>` executor, this is still going to be accepted, and the command just won't work at runtime. Worse, you might have not specified the `sourceType` and the executor is of type `CommandExecutor<Player>`, so running as the console will cause a runtime error.

I realise that type erasure makes this difficult though.

windy1: Yeah I did think about this (also this PR will likely end up being inplemented differently) but I failed to find an elegant solution and in the end I decided to keep it simple as to not have to hold the developers hand through this.

src/main/java/org/spongepowered/api/command/CommandCallable.java
20zinnm: @simon816 I'm not familiar with _exactly_ the JavaDoc standards for tense but "Test" works grammatically. It's saying "this method will test" rather than "this method tests." However, this might be changed due to standards, not quite sure.

simon816: Take a look at javadocs in other parts of the API, you'll see we consistently use "Gets" "Sets" etc

20zinnm: @simon816 Ok, my bad. If it's usually in that form then it should be added to the format guide.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
dualspiral: Well, the whole point of an API is that it should pretty much hold the developer's hand! It's supposed to provide type safety!

Would it be better if you provided two executor methods instead and make use of compile time safety? For the current case, no source type filtering (though, this will cause a warning in current code, but that's what bleeding is for, right?)

```
public Builder executor(CommandExecutor<CommandSource> executor) 
```

and, for what you're trying to do:

```
public <T extends CommandSource> Builder executor(CommandExecutor<? super T> executor, Class<T> sourceType) 
```

This will then make the compiler check that the source type is correct. Having a two argument method does feel like it goes against the grain of the builder pattern, but given the source type and the fact Java has type erasure, maybe it's a better solution?

src/main/java/org/spongepowered/api/command/CommandSourceTypeException.java
kinggoesgaming: As a general rule use `@link` when referring any classes part of that particular project which here is SpongeAPI

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
windy1: Yes but as I said this is more of a proof of concept than something that will actually go into the API. I originally implemented it exactly as you've written but thought it was redundant since you would have specified it in sourceType already. IMO with generics it gets to a certain point where you just need to trust that the developer implements it correctly but that's irrelevant anyways since, as I said, this won't be getting pulled.

gabizou:
I think as a good note to leave here is that the idea was good, but it'd be better to have a sort of addition for .executor(Class<T extends CommandSource>, GenerifiedCommandExecutor<T>) where the parameters are soft compiled with some asm as event filtering does. That ultimately would give the better option IMO. Quite possible to do in the future if you want to write it out (the implementation as well perhaps).

PR #1063, opened 2016-02-06
Text Representables

ST-DDT:
Implemenation: https://github.com/SpongePowered/SpongeCommon/pull/483

## The issue

Currently it is not possible to have a default representation for a particular instance.

This has the following disadvantages:
* Code duplication (both in different plugins and in the same plugin) to format the instances
* Dismatching presentation style of the same instance across plugins
* Ugly fallback to `toString()` if no custom representation is set

````java
Text myMessage = Text.of("I got an ", toText(itemStack));

// [...]

public static Text toText(ItemStack stack) {
    final TextBuilder builder;
    final Optional<DisplayNameData> optName = stack.get(DisplayNameData.class);
    if (optName.isPresent()) {
        Value<Text> displayName = optName.get().displayName();
        if (displayName.exists()) {
            builder = displayName.get().builder();
        } else {
            builder = Texts.builder(stack.getTranslation());
        }
    } else {
        builder = Texts.builder(stack.getTranslation());
    }
    builder.onHover(TextActions.showItem(stack));
    return builder.build();
}
````

## Solutions

* Level 1: Allow classes to specify their default representation
* Level 2: Level 1 + Use special format if available
* Level 3: Level 2 + Use the human readable name if available

### The minimal solution (level 1)

Expose the `TextRepresentable` interface and allow the server and plugins to implement it where they want so a lot of instances (can) have a default representation and can easily be passed to `Text.of(Object...)`. This solution does not force any implementation to implement anything at all, because it. With this solution you can pass the single instance to `Text.of(Object....)` and can style the resulting instance without knowledge of the original class and its methods (e.g. `getName()` or `getDisplayName()`). IMO everything that has a human readable name (e.g. `getName()` method) should implement this class.

**None breaking.**

````java
Text myMessage = Text.of("I got an ", itemStack);
````

### The medium solution (level 2)

Expose the `TextRepresentable` interface and mark some classes (mainly those classes which have a representation in vanilla by default) with this interface, so there is a basic set of classes that are forced to have a special representation. Server and plugins can still implement it how they want and can also implement it in a lot of other classes. With this solution you can easily pass the instances to `Text.of(Object...)` or adjust the `Text` representation to your needs (replace click action or alike) without having to style/write it from scratch.

* `Achievement`
* `Entity` and `Player` (maybe `User` as well for consistency reasons)
* `ItemStack`
* `Statistic` (has a representation, but not yet tested)

**Minimal implementation changes required**
* 1x `shadow$getDisplayName()` + 1 line implemenation per class

````java
// Optional customization
Object blueStack = itemStack.toText().toBuilder().color(TextColors.BLUE);

Text myMessage = Text.of("I got an ", blueStack);
````

### The advanced solution (level 3)

Expose the `TextRepresentable` interface and mark all classes with this interface that are used in chat or commands frequently. (All classes that have a `getName()` or a similar method)
* The classes mentioned in level 2
* `CatalogType`s are often used in commands and they have a discrepancy
between their human readable name and their internal id which requires
them to use special representations to be usable in commands.
* `World`s are often used in chat messages and they have a name so, the name
should be displayed.

There are probably more that will match the criteria for this level.

**Minimal implementation changes required**
* Same as level 2
* Most of the other necessary changes can be `default` implemented.

## Noteworthy

Server and plugins are still able to overwrite the default `Text` representation of the object. They just have to do it before `Text.of(Object...)`. They could also rely on the default representation and just optimize it for their current needs.

I added the required code for all three levels, but we can drop one or two of them if you disagree with them. Although I would like to see them all.

### DIFF
--- a/src/main/java/org/spongepowered/api/CatalogType.java
+++ b/src/main/java/org/spongepowered/api/CatalogType.java
@@ -24,6 +24,10 @@
  */
 package org.spongepowered.api;
 
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.TextRepresentable;
+import org.spongepowered.api.text.action.TextActions;
+
 /**
  * Represents a type of a catalog that can be used to identify types without
  * using an {@link Enum}.
@@ -35,7 +39,7 @@ package org.spongepowered.api;
  * <code>`a.getId().equalsIgnoreCase(b.getId())`</code> are true then all must
  * be true.</p>
  */
-public interface CatalogType {
+public interface CatalogType extends TextRepresentable {
 
     /**
      * Gets the unique identifier of this {@link CatalogType}. The identifier is
@@ -60,5 +64,13 @@ public interface CatalogType {
      * @return The human-readable name of this catalog type
      */
     String getName();
+    
+    @Override
+    default Text toText() {
+        final Text.Builder builder = Text.builder(getName());
+        builder.onHover(TextActions.showText(Text.of("Id: " + getId() + "\nName: " + getName())));
+        builder.onShiftClick(TextActions.insertText(getId()));
+        return builder.toText();
+    }
 
 }
--- a/src/main/java/org/spongepowered/api/world/World.java
+++ b/src/main/java/org/spongepowered/api/world/World.java
@@ -30,6 +30,8 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.effect.Viewer;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.service.context.ContextSource;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.TextRepresentable;
 import org.spongepowered.api.world.difficulty.Difficulty;
 import org.spongepowered.api.world.explosion.Explosion;
 import org.spongepowered.api.world.extent.Extent;
@@ -47,7 +49,7 @@ import java.util.UUID;
 /**
  * A loaded Minecraft world.
  */
-public interface World extends Extent, WeatherUniverse, Viewer, ContextSource {
+public interface World extends Extent, WeatherUniverse, Viewer, ContextSource, TextRepresentable {
 
     @Override
     default Location<World> getLocation(Vector3i position) {
@@ -301,5 +303,10 @@ public interface World extends Extent, WeatherUniverse, Viewer, ContextSource {
 
     @Override
     MutableBlockVolumeWorker<? extends World> getBlockWorker();
+
+    @Override
+    default Text toText() {
+        return Text.of(getName());
+    }
 
 }
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -38,6 +38,7 @@ import org.spongepowered.api.data.manipulator.mutable.TargetedLocationData;
 import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.cause.NamedCause;
 import org.spongepowered.api.event.cause.entity.damage.source.DamageSource;
+import org.spongepowered.api.text.TextRepresentable;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.Identifiable;
 import org.spongepowered.api.util.RelativePositions;
@@ -71,7 +72,7 @@ import javax.annotation.Nullable;
  *
  * <p>Blocks and items (when they are in inventories) are not entities.</p>
  */
-public interface Entity extends Identifiable, DataHolder, DataSerializable, Translatable {
+public interface Entity extends Identifiable, DataHolder, DataSerializable, Translatable, TextRepresentable {
 
     /**
      * Get the type of entity.
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
@@ -38,6 +38,7 @@ import org.spongepowered.api.data.DataView;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.text.TextRepresentable;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.ResettableBuilder;
 
@@ -49,7 +50,7 @@ import org.spongepowered.api.util.ResettableBuilder;
  * use {@link DataHolder#get(Class)} to retrieve different information
  * regarding this item stack.</p>
  */
-public interface ItemStack extends DataHolder, DataSerializable, Translatable {
+public interface ItemStack extends DataHolder, DataSerializable, Translatable, TextRepresentable {
 
     /**
      * Creates a new {@link Builder} to build an {@link ItemStack}.
--- a/src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
+++ b/src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
@@ -27,6 +27,7 @@ package org.spongepowered.api.statistic.achievement;
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.statistic.Statistic;
+import org.spongepowered.api.text.TextRepresentable;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.api.util.ResettableBuilder;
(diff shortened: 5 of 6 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/api/CatalogType.java [hunk 1]
```diff
@@ -60,5 +64,13 @@
      * @return The human-readable name of this catalog type
      */
     String getName();
+    
+    @Override
+    default Text toText() {
```
gabizou: Didn't we have a discussion about `CatalogType`s being `TextRepresentable`? I distinctly remember having an issue with it, and I still do.

ST-DDT: Yes, we had (See https://github.com/SpongePowered/SpongeAPI/pull/808#discussion_r49005018). Thats why i explicitly made this three levelled approach.

I removed it from that PR since it was a separate problem of the current text API and I didn't want to push everything in there.

I have since thought about it a lot.

> Convince me why **every** `CatalogType` needs to be `TextRepresentable`.

`CatalogType`s have a name. Not using it for displaying it to the user is just wrong, this should be prevented escpecially if the `toString()` fallback is unusable for displaying in chat.
````
toString: BlockColored{id=minecraft:stained_hardened_clay, name=minecraft:stained_hardened_clay, translation=SpongeTranslation{id=tile.clayHardenedStained.name}, toString=BlockColored{id=minecraft:stained_hardened_clay, name=minecraft:stained_hardened_clay, translation=SpongeTranslation{id=tile.clayHardenedStained.name}, class=net.minecraft.block.BlockColored}}

toText: Stained Clay
````

IMO everything that has a name should be forced to display it in chat unless it has something better to display (custom name etc).
I just don't want to go the road down to a interface like this:

````java
public interface Named extends TextRepresentable {

    String getName();

    default Text toText() {
        return Text.of(getName());
    }

}
````

Default implementations can be overwritten and don't impose anything on implementations except their existance.

I'm not sure about the current implementation though.
(I though about moving it to a `Function<CatalogType,Text>`)

Comment (gabizou) [inline src/main/java/org/spongepowered/api/CatalogType.java]: Again, the thing is, this requires that every `CatalogType` has a default implementation of a legible name that is presented for text, which as an object, is entirely immutable. presenting what a `BlockType` can be as a `Text` object is perfectly fine, but telling me that every `CatalogType` needs to be representable as `Text` objects with little to no configurable alternative is equivalent to ignoring that a `CatalogType` is `TextRepresentable` at all. What would make sense is if you exposed the ability to register `TextRepresenters` that could create a `Text` object for a `CatalogType` such that plugins can dictate how they want certain ones to be read in chat.

Default implementations or not, I seriously don't want to make this a contracted method for `CatalogType` in general as it's simply in the wrong place.

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/CatalogType.java]: `TextRepresentable` is **not** about `How I want it to be displayed  in chat`.
It is about `How IT want to represent itself in chat`.
Its the best-effort approach if you don't know anything about the internals of the class.

Using a `TextRepresenter`s is something totally different. That would be the equivalent of adding a `Function<Object,Text>` to `Text.of(Object...)` which would make the method useless since it would only be a for each loop.
````java
Arrays.stream(objects).map(textRepresenter).forEach(builder::append);
````
Maybe a `Function<Object,TextBuilder>` would be a better choice of param. But this still wouldn't solve the issue of the ugly `toString()` default if nothing is specified.

Comment (gabizou) [inline src/main/java/org/spongepowered/api/CatalogType.java]: If the developer knows what they are including in a `Text` object, then why not leave it up to the developer to decide on how they want a `CatalogType` to be seen as `Text`. `CatalogType` is so ultimately vague that no one can dictate how it should be seen as a `Text` object short of writing `Text` as if it's the `toString()` including all the known variables of the `CatalogType`, which honestly is much more information than needed to present itself as a `Text` object.

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/CatalogType.java]: I just had a good example for this.

Lets assume you have a random `CatalogType` instance. You want to display it with green color.

How would you do that?
````java
return Text.of(TextColors.GREEN, instance.getName());
````

Looks easy, but there might be a `Tranlsation` for your type as well.
Okay, so we change it to this

````java
if (instance instanceOf Translatable) {
    return Text.of(TextColors.GREEN, instance.getTranslation());
} else {
    return Text.of(TextColors.GREEN, instance.getName());
}
````
There might be plugin provided types that don't have a translation although they are `Translatable`.

````java
if (instance instanceOf Translatable) {
    if (instance.getTranslation()==null) {
        return Text.of(TextColors.GREEN, instance.getName());
    } else {
        return Text.of(TextColors.GREEN, instance.getTranslation());
    }
} else {
    return Text.of(TextColors.GREEN, instance.getName());
}
````

Mhh, not very beautiful. 

"Oh I forgot if this instance if from Plugin XY then there is a pretty print method."

````java
if (instance instanceOf PluginABCType) {
   return Text.of(TextColors.GREEN, instance.getText());
} else if (instance instanceOf PluginXYType) {
   return Text.of(TextColors.GREEN, instance.getPrettyText());
} else if (instance instanceOf Translatable) {
    if (instance.getTranslation()==null) {
        return Text.of(TextColors.GREEN, instance.getName());
    } else {
        return Text.of(TextColors.GREEN, instance.getTranslation());
    }
} else {
    return Text.of(TextColors.GREEN, instance.getName());
}
````
Ugly...

and with my PR
````java
return Text.of(TextColors.GREEN, instance);
````

"But plugin XY does not implement that method (like I want)."

Do you know the plugin / Do you have some kind of class dependency?

"No"

Then you couldn't have formatted it properly by yourself as well.

**Comment by ST-DDT** [comment]
@gabizou Besides disagreeing on level 3.

What do you think about level 1 and 2?

**Comment by windy1** [comment]
@ST-DDT Now that `TextRepresentable` is public, is there still anything in this PR you would like to see merged?

**Comment by ST-DDT** [comment]
I would still like to see some API classes implement TextRepresentable.

**Comment by windy1** [comment]
I'm not sure I really agree to be honest. Theres not really a strong case to make these kinds of things TextRepresentable IMO. We probably shouldn't be passing around Text objects when a String will do just as well. I'm sure it would result in a decent amount of overhead and sets and ugly precedent for the API. Text != String

**Comment by ST-DDT** [comment]
Yes. Text != String. You cannot hover a string.
How would you display an achievement or itemstack in chat (using the default style)? The only option you have would be to duplicate code.

**Comment by windy1** [comment]
It just doesn't seem worth it though; memory better spent elsewhere. There are already ways to display achievements or item stacks through chat, if you want to make the default "chat style" for ItemStacks available a better alternative would probably be to make a statically available TextTemplate for it.

**Comment by ST-DDT** [comment]
Memory, really?
How much memory does implementing an interface in a Mixin take?
The real implementation is already provided by vanilla itself.

**Comment by windy1** [comment]
Mixins are an integral part of Sponge and have a very strong use case, sure it might not seem like a big difference but many small optimizations make big optimization.

**Comment by ST-DDT** [comment]
IMO its a bad trade:
Always 1 in the API vs N in plugins

**Comment by ryantheleach** [comment]
Wouldn't the methods that the proposed mixins that implement text representable just end up creating the text lazily anyway when it needs to be displayed? If anything it just seems like convenience methods so you can pass an entity/achievement/itemstack straight into chat?

**Comment by ST-DDT** [comment]
@ryantheleach 
Yes the `Text` is created lazily.

If you consider exposing an impl method a convenience method, then yes.
But this convenience method safes you from writing/duplicating the entire display code from the impl in your plugin or using reflection.

**With**
````java
textBuilder.append(item.toText());
````

**vs.**

**Without**
````java
textBuilder.append(toText(item));

[...]

public static Text toText(ItemStack stack) {
    final TextBuilder builder;
    final Optional<DisplayNameData> optName = stack.get(DisplayNameData.class);
    if (optName.isPresent()) {
        Value<Text> displayName = optName.get().displayName();
        if (displayName.exists()) {
            builder = displayName.get().builder();
        } else {
            builder = Texts.builder(stack.getTranslation());
        }
    } else {
        builder = Texts.builder(stack.getTranslation());
    }
    builder.onHover(TextActions.showItem(stack));
    return builder.build();
}
````

It is possible but error prone, code duplication, tend to differ across plugins unintentionally and new MC features might easily be forgotten.

**Comment by windy1** [comment]
@ryantheleach The problem comes when people start storing these Texts in memory. One might be tempted to start storing Text objects where the really should just be storing Strings. Honestly, it's not a huge issue, but I could see it becoming a problem. Again, I really think this would be better implemented with static TextTemplates @ST-DDT.

**Comment by ryantheleach** [comment]
I see what you are saying now.

**Comment by ST-DDT** [comment]
@ryantheleach The entity is not saved/referenced due to the TextRepresentable.
Unless you store it directly.

File: src/main/java/org/spongepowered/api/world/World.java [hunk 2]
```diff
@@ -302,4 +304,9 @@
     @Override
     MutableBlockVolumeWorker<? extends World> getBlockWorker();
 
+    @Override
+    default Text toText() {
```
Minecrell: Don't randomly implement it for one interface just because Minecraft has no implementation for it internally. This should be in SpongeCommon

Minecrell:
Can you rebase this on the latest master as soon as you find some time? Also make sure to keep the levels separate.

gabizou:
Looking over this PR, there's really no reason to add this as most every single instance that you want to represent any of the objects that are being marked as `TextRepresentable`, you're going to customize it for each instance. Hell, even a `TextTemplate` can be used for customization. Seeing as there hasn't been much of a request for this sort of feature, I'm closing this PR.

PR #1065, opened 2016-02-06
TargetedLocation uses Vector3d, not Location<World>

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1065) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/485)
Supersedes #1064

This changes `TargetedLocation` to use `Vector3d` instead of `Location<World>`.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -848,7 +848,7 @@ public final class Keys {
 
     public static final Key<OptionalValue<UUID>> TAMED_OWNER = null;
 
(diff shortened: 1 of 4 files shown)


windy1:
I think keeping this as a `Vector3d` is OK because the target location will never be in a different world.

kenzierocks:
That's what the change is @windy1.

windy1:
Lol whoops, carry on.

gabizou:
Merged in https://github.com/SpongePowered/SpongeAPI/commit/1a2916f85c468b1fa4f8e6f3ae1f9eee37a7627a

PR #1068, opened 2016-02-10
Introduce GameProfileCache, and extract existing cache method to it

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1068) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/495)

This introduces a replaceable `GameProfileCache` that is used (internally) for resolving profiles and profile related data.

### DIFF
--- a/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
@@ -1,0 +1,314 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.profile;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import org.spongepowered.api.entity.living.player.User;
+import org.spongepowered.api.service.user.UserStorageService;
+import org.spongepowered.api.util.GuavaCollectors;
+
+import java.util.Collection;
+import java.util.Date;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a cache of {@link GameProfile}s.
+ */
+public interface GameProfileCache {
+
+    /**
+     * Add an entry to this cache.
+     *
+     * @param profile The profile to cache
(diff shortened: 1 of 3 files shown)


ryantheleach:
Whats a usecase of being able to replace the cache?

kashike:
@ryantheleach Looking up from your own profile cache instead of hitting Mojang's API, which is rate limited.

JBYoshi:
@kashike I believe the built-in implementation uses a local cache to avoid making multiple requests for the same profile in a short amount of time, so that shouldn't be needed.

kashike:
I'm well aware of the internal cache. Keep in mind networks that have many servers can utilise a central cache (database, etc), instead of requesting from Mojang.

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
```diff
@@ -0,0 +1,313 @@
...
+
+    /**
+     * Gets a {@link GameProfile} from this cache by its unique id.
+     *
+     * @param uniqueId The unique id of the profile
+     * @return The profile, if present, or {@link Optional#empty()} if
+     * the cache did not contain a profile with the provided id
+     */
+    default Optional<GameProfile> get(UUID uniqueId) {
+        checkNotNull(uniqueId, "unique id");
```
Minecrell: `uniqueId`, name this the same as the parameter

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
```diff
@@ -0,0 +1,313 @@
...
+        return Optional.ofNullable(Iterables.getFirst(profiles, null));
+    }
+
+    /**
+     * Gets {@link GameProfile}s in bulk by their unique id.
+     *
+     * @param uniqueIds The unique ids
+     * @return A collection of successfully found up profiles
+     */
+    Collection<GameProfile> getBulkId(Iterable<UUID> uniqueIds);
```
ST-DDT: Maybe a Map would be a better return type.
That way you can implizitly handle absent profiles better. And also use them easier.

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * Looks up {@link GameProfile}s in bulk by their unique id and
+     * loads them into this cache.
+     *
+     * @param uniqueIds The unique ids
+     * @return A collection of successfully looked up profiles
+     */
+    Collection<GameProfile> lookupBulkId(Iterable<UUID> uniqueIds);
```
ST-DDT: I would prefer a Map here too.
I would also change the method name to `bulkLookup` or `lookupBulk`.

kashike: Method name must retain the "Id" part, signature conflict otherwise. `bulkLookupId` is the alternative, but I kept `lookup` first to "group" methods

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
kashike: `Map<UUID, Optional<GameProfile>>`?

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
ST-DDT: Yeah. I edited my post accordingly.

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
ST-DDT: Maybe just `Map<UUID,GameProfile>`. You could skip uuids that aren't present. That way you can check is the key present and don't have to check on the result for null and absent.
But this is just a matter of preference.

src/main/java/org/spongepowered/api/profile/GameProfileCache.java
kashike: What would you want it named? Keep in mind that it is bulk.

ST-DDT: * Single `lookupId`
* Bulk `lookupIds`

I would also default implement the bulk methods using the single method, because some impls aren't capeable to do real bulk lookups anyway. If an impl can optimize performance by using custom impls they can still overwrite the impl if they want.

kashike: The single id method is defaulting to doing it in bulk - you're saying to reverse?

kashike: @ST-DDT changes pushed. method names are a bit weird now

ST-DDT: What is the vanilla minecraft default lookup: single or bulk? I guess we should stick to that.

ST-DDT: > method names are a bit weird now

Which ones?  The names are just proposals.
`getId` should propably be `getById`

kashike: @ST-DDT how's API and impl look now?

me4502:
This is just my opinion and should probably not be done until others agree, but it may make more sense to have this as an overridable service. That way by default it can use the vanilla behaviour, but another plugin can provide an implementation that uses their own database etc..

Seeing as we already have a system like that in place, it only makes sense to use it.

@gabizou Thoughts?

kashike:
@me4502 The reason this isn't a service is because it would be very different from others - this provides (assuming vanilla-based in this context) Mojang's implementation, by default, with `getDefaultCache`.

A possible "solution" would be to keep `getDefaultCache` where it is now and turn the rest into a service

Issue #1073, opened 2016-02-11
Allow determination of whether or not a command execution has failed


minecrafter:
I am working on the new official plugin for [Buycraft](https://www.buycraft.net). An integral and required part of the plugin is to determining whether or not an external plugin command has generated an exception or has otherwise failed to execute so that the command execution can be delayed and retried at a later time. Currently, Sponge does not provide this functionality, leaving us to simply execute the command and hope it successfully completed.

minecrafter:
As it turns out, `InvocationCommandException` is mostly what I need. However, it doesn't handle uncaught exceptions from commands.

me4502:
> As it turns out, InvocationCommandException is mostly what I need. However, it doesn't handle uncaught exceptions from commands.

Would it not make more sense to have it just handle all exceptions? @SpongePowered/developers ?

ImMorpheus:
https://github.com/SpongePowered/SpongeAPI/blob/94d7fb22fcc0b77e5af44b9f9e7dc68534ecc3da/src/main/java/org/spongepowered/api/command/CommandException.java#L31-L32

API wise, this is handled by `CommandException`

PR #1075, opened 2016-02-13
Allow setting throwable for CommandResult.

minecrafter:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/503)

Resolves SpongePowered/SpongeAPI#1073

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandResult.java
+++ b/src/main/java/org/spongepowered/api/command/CommandResult.java
@@ -39,6 +39,7 @@ public class CommandResult {
     private final Optional<Integer> affectedEntities;
     private final Optional<Integer> affectedItems;
     private final Optional<Integer> queryResult;
+    private final Optional<Throwable> throwable;
 
     /**
      * Returns a {@link Builder}.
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -70,20 +71,21 @@ public static CommandResult success() {
 
     /**
      * Constructs a new command result.
-     *
-     * @param successCount The success count
+     *  @param successCount The success count
```
kashike: Oops.

src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -209,12 +222,23 @@ public Builder queryResult(@Nullable Integer queryResult) {
...
          *
          * @return A CommandResult with the specified settings
          */
         public CommandResult build() {
-            return new CommandResult(this.successCount, this.affectedBlocks, this.affectedEntities, this.affectedItems, this.queryResult);
+            return new CommandResult(this.successCount, this.affectedBlocks, this.affectedEntities, this.affectedItems, this.queryResult, this.throwable);
         }
     }
-}
+}
```
kashike: Keep newline.

src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -70,20 +71,21 @@ public static CommandResult success() {
 
     /**
      * Constructs a new command result.
-     *
```
kashike: Don't do this, keep the new line

src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -209,12 +222,23 @@ public Builder queryResult(@Nullable Integer queryResult) {
         }
 
         /**
+         * Sets the exception thrown from this command.
+         *
+         * @param throwable the exception thrown from this command
```
kashike: `@param throwable The ...` (capital `t`)

gabizou:
@zml2008 I'd imagine this wouldn't really change much for the command refactors you have planned later on?

zml2008:
On February 28, 2016 1:03:31 PM PST, Gabriel Harris-Rouquette <notifications@github.com> wrote:
>@zml2008 I'd imagine this wouldn't really change much for the command
>refactors you have planned later on?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/SpongePowered/SpongeAPI/pull/1075#issuecomment-189946068

While it probably wouldn't change much, it's not implemented in a way that fits at all in with the rest of the commands api, since its kinda shoved in at the highest level after all other processing occurs (so for example if another executor returned a command result with a set throwable it would be ignored. The way my work on the commands api is evolving it seems like a failure result may replace throwing exceptions, but this is probably not the way to do it.

zml2008:
The earlier issues I have with this PR still stand and nothing else has come of it, so closing for now.

PR #1079, opened 2016-02-16
TextTemplate API

windy1:
This PR adds re-usable, serializable, text templates to SpongeAPI. This is useful in plugin development when a developer wishes to store a message in a config file with variable elements.

## Example

```java
import static org.spongepowered.api.text.template.TextTemplate.*;

TextTemplate template = of(
    TextColors.BLUE, "Hello, ", arg("playerName").color(TextColors.AQUA).style(TextStyles.BOLD), "!"
);

void onPlayerJoin() {
    player.sendMessage(template, ImmutableMap.of("playerName", Text.of(player.getName()));
}
```

A few things to note:
* Repeated arguments are permitted but must be "equivalent", meaning that if argument 1 named 'x' is optional argument 2 named 'x' must also be optional.
* Elements are appended to the text builder in the order in which they appear.
* Arguments are required by default. You can change this with `arg("name").optional()`.
* TextTemplates are immutable.
* Arguments can have a TextFormat which can be overwritten (or not) by the TextElement supplied to the parameter map.
* A "placeholder" representation of the template can be obtained with `TextTemplate.toText()`
* TextTemplates are Iterable and `iterator()` returns an unmodifiable Iterator of it's elements.
* Green is my favorite color.

## Serialization

TextTemplates are config serializable with @zml2008's Configurate. The above template will produce the following serialized output:

```hocon
template {
    arguments {
        playerName {
            optional=false
        }
    }
    content {
        color=blue
        extra=[
            "Hello, ",
            {
                bold=true
                color=aqua
                text="{playerName}"
            },
            "!"
        ]
        text=""
    }
    options {
        openArg={
        closeArg=}
    }
}
```

Templates are serialized and deserialized like so:

```java
node.getNode("template").setValue(TypeToken.of(TextTemplate.class), template); // serialize
node.getNode("template").getValue(TypeToken.of(TextTemplate.class)); // deserialize
```

This PR also makes Text objects config serializable:

```java
node.getNode("text").setValue(TypeToken.of(Text.class), text); // serialize
node.getNode("text").getValue(TypeToken.of(Text.class)); // deserialize
```

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -29,6 +29,12 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
+import com.google.common.reflect.TypeToken;
+import ninja.leaping.configurate.objectmapping.serialize.TypeSerializers;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.MemoryDataContainer;
+import org.spongepowered.api.data.Queries;
 import org.spongepowered.api.scoreboard.Score;
 import org.spongepowered.api.text.action.ClickAction;
 import org.spongepowered.api.text.action.HoverAction;
@@ -39,6 +45,7 @@ import org.spongepowered.api.text.format.TextFormat;
 import org.spongepowered.api.text.format.TextStyle;
 import org.spongepowered.api.text.format.TextStyles;
 import org.spongepowered.api.text.selector.Selector;
+import org.spongepowered.api.text.serializer.TextConfigSerializer;
 import org.spongepowered.api.text.serializer.TextSerializers;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.text.translation.Translation;
@@ -75,7 +82,11 @@ import javax.annotation.Nullable;
  * @see SelectorText
  * @see ScoreText
  */
-public abstract class Text implements TextRepresentable {
+public abstract class Text implements TextRepresentable, DataSerializable, Comparable<Text> {
+
+    static {
+        TypeSerializers.getDefaultSerializers().registerType(TypeToken.of(Text.class), new TextConfigSerializer());
+    }
 
     /**
      * The empty, unformatted {@link Text} instance.
@@ -246,6 +257,44 @@ public abstract class Text implements TextRepresentable {
      */
     public final String toPlain() {
         return TextSerializers.PLAIN.serialize(this);
+    }
+
+    /**
+     * Concatenates the specified {@link Text} to this Text and returns the
+     * result.
+     *
+     * @param other To concatenate
+     * @return Concatenated text
+     */
+    public final Text concat(Text other) {
+        return toBuilder().append(other).build();
+    }
+
+    /**
+     * Removes all empty texts from the beginning and end of this
+     * text.
+     *
+     * @return Text result
+     */
+    public final Text trim() {
+        return toBuilder().trim().build();
+    }
+
+    @Override
+    public int getContentVersion() {
+        return 1;
+    }
+
+    @Override
+    public DataContainer toContainer() {
+        return new MemoryDataContainer()
+                .set(Queries.CONTENT_VERSION, getContentVersion())
+                .set(Queries.JSON, TextSerializers.JSON.serialize(this));
+    }
+
+    @Override
+    public int compareTo(Text o) {
+        return PLAIN_COMPARATOR.compare(this, o);
     }
 
     @Override
@@ -652,6 +701,32 @@ public abstract class Text implements TextRepresentable {
          */
         public Builder removeAll() {
             this.children.clear();
+            return this;
+        }
+
+        /**
+         * Removes all empty texts from the beginning and end of this
+         * builder.
+         *
+         * @return This builder
+         */
+        public Builder trim() {
+            Iterator<Text> front = this.children.iterator();
+            while (front.hasNext()) {
+                if (front.next().isEmpty()) {
+                    front.remove();
+                } else {
+                    break;
+                }
+            }
+            ListIterator<Text> back = this.children.listIterator(this.children.size());
+            while (back.hasPrevious()) {
+                if (back.next().isEmpty()) {
+                    back.remove();
+                } else {
+                    break;
+                }
+            }
             return this;
         }
 
--- a/src/main/java/org/spongepowered/api/text/TextElement.java
+++ b/src/main/java/org/spongepowered/api/text/TextElement.java
@@ -22,45 +22,18 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.text.action;
-
-import org.spongepowered.api.text.Text;
+package org.spongepowered.api.text;
 
 /**
- * Represents a {@link TextAction} that responds to shift-clicks.
- *
- * @param <R> the type of the result of the action
+ * Represents anything that can be applied to a {@link Text.Builder}.
  */
-public abstract class ShiftClickAction<R> extends TextAction<R> {
+public interface TextElement {
 
     /**
-     * Constructs a new {@link ShiftClickAction} with the given result.
+     * Applies this element to the end of the specified builder.
      *
-     * @param result The result of the shift click action
+     * @param builder Text builder to apply to
      */
-    ShiftClickAction(R result) {
-        super(result);
-    }
+    void applyTo(Text.Builder builder);
 
-    @Override
-    public void applyTo(Text.Builder builder) {
-        builder.onShiftClick(this);
-    }
-
-    /**
-     * Inserts some text into the chat prompt.
-     */
-    public static final class InsertText extends ShiftClickAction<String> {
-
-        /**
-         * Constructs a new {@link InsertText} instance that will insert text at
-         * the current cursor position in the chat when it is shift-clicked.
-         *
-         * @param text The text to insert
-         */
-        InsertText(String text) {
-            super(text);
-        }
-
-    }
 }
--- a/src/main/java/org/spongepowered/api/text/TextTemplate.java
+++ b/src/main/java/org/spongepowered/api/text/TextTemplate.java
@@ -1,0 +1,614 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.reflect.TypeToken;
+import ninja.leaping.configurate.objectmapping.Setting;
+import ninja.leaping.configurate.objectmapping.serialize.ConfigSerializable;
+import ninja.leaping.configurate.objectmapping.serialize.TypeSerializers;
+import org.spongepowered.api.text.format.TextColor;
+import org.spongepowered.api.text.format.TextFormat;
+import org.spongepowered.api.text.format.TextStyle;
+import org.spongepowered.api.text.serializer.TextTemplateConfigSerializer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a re-usable template that produces a formatted
+ * {@link Text.Builder}. Elements will be appended to the result builder in the
+ * order that they are specified in {@link #of(Object...)}.
+ */
+public final class TextTemplate implements TextRepresentable, Iterable<Object> {
+
+    static {
+        TypeSerializers.getDefaultSerializers().registerType(TypeToken.of(TextTemplate.class), new TextTemplateConfigSerializer());
+    }
+
+    /**
+     * Default "open" String for how arguments are contained within the template.
+     */
+    public static final String DEFAULT_OPEN_ARG = "{";
+
+    /**
+     * Default "close" String for how arguments are contained within the template.
+     */
+    public static final String DEFAULT_CLOSE_ARG = "}";
+
+    /**
+     * Empty representation of a {@link TextTemplate}. This is returned if the
+     * array supplied to {@link #of(Object...)} is empty.
+     */
+    public static final TextTemplate EMPTY = new TextTemplate(DEFAULT_OPEN_ARG, DEFAULT_CLOSE_ARG, new Object[]{});
+
+    final ImmutableList<Object> elements;
+    final ImmutableMap<String, Arg> arguments;
+    final Text text;
+    final String openArg;
+    final String closeArg;
+
+    TextTemplate(String openArg, String closeArg, Object[] elements) {
+        this.openArg = openArg;
+        this.closeArg = closeArg;
+
+        // collect elements
+        ImmutableList.Builder<Object> elementList = ImmutableList.builder();
+        Map<String, Arg> argumentMap = new HashMap<>();
+        for (Object element : elements) {
+            if (element instanceof Arg.Builder) {
+                element = ((Arg.Builder) element).build();
+            }
+            if (element instanceof Arg) {
+                // check for non-equal duplicate argument
+                Arg newArg = new Arg((Arg) element, this.openArg, this.closeArg);
+                Arg oldArg = argumentMap.get(newArg.name);
+                if (oldArg != null && !oldArg.equals(newArg)) {
+                    throw new TextTemplateArgumentException("Tried to supply an unequal argument with a duplicate name \""
+                            + newArg.name + "\" to TextTemplate.");
+                }
+                argumentMap.put(newArg.name, newArg);
+                element = newArg;
+            }
+            elementList.add(element);
+        }
+        this.elements = elementList.build();
+        this.arguments = ImmutableMap.copyOf(argumentMap);
+
+        // build text representation
+        Text.Builder builder = null;
+        for (Object element : this.elements) {
+            builder = apply(element, builder);
+        }
+        this.text = Optional.ofNullable(builder).orElse(Text.builder()).build();
+    }
+
+    /**
+     * Returns the elements contained in this TextTemplate.
+     *
+     * @return The elements within the template
+     */
+    public List<Object> getElements() {
+        return this.elements;
+    }
+
+    /**
+     * Returns the arguments contained within the TextTemplate.
+     *
+     * @return The arguments within this TextTemplate
+     */
+    public Map<String, Arg> getArguments() {
+        return this.arguments;
+    }
+
+    /**
+     * Returns the string used for containing Args within the template.
+     *
+     * @return String containing args
+     */
+    public String getOpenArgString() {
+        return this.openArg;
+    }
+
+    /**
+     * Returns the string used for containing Args within the template.
+     *
+     * @return String containing args
+     */
+    public String getCloseArgString() {
+        return this.closeArg;
+    }
+
+    /**
+     * Concatenates the specified {@link TextTemplate} to this template and
+     * returns the result. In the event that the two templates' open/close
+     * argument containers vary, this template's argument containers will be
+     * used.
+     *
+     * @param other To concatenate
+     * @return Concatenated template
+     */
+    public TextTemplate concat(TextTemplate other) {
+        List<Object> elements = new ArrayList<>(this.elements);
+        elements.addAll(other.elements);
+        return of(this.openArg, this.closeArg, elements.toArray(new Object[elements.size()]));
+    }
+
+    /**
+     * Applies an empty map of parameters to this TextTemplate and returns the
+     * result in a {@link Text.Builder}.
+     *
+     * @return Text builder containing result
+     * @throws TextTemplateArgumentException if required parameters are missing
+     */
+    public Text.Builder apply() {
+        return apply(Collections.emptyMap());
+    }
+
+    /**
+     * Applies the specified parameters to this TextTemplate and returns the
+     * result in a {@link Text.Builder}.
+     *
+     * @param params Parameters to apply
+     * @return Text builder containing result
+     * @throws TextTemplateArgumentException if required parameters are missing
+     */
+    public Text.Builder apply(Map<String, TextElement> params) {
+        return apply(null, params);
+    }
+
+    private Text.Builder apply(@Nullable Text.Builder result, Map<String, TextElement> params) {
+        checkNotNull(params, "params");
+        for (Object element : this.elements) {
+            result = apply(element, result, params);
+        }
+        return Optional.ofNullable(result).orElse(Text.builder());
+    }
+
+    @Nullable
+    private Text.Builder apply(Object element, @Nullable Text.Builder builder, Map<String, TextElement> params) {
+        // Note: The builder is initialized as null to avoid unnecessary Text nesting
+        if (element instanceof Arg) {
+            Arg arg = (Arg) element;
+            TextElement param = params.get(arg.name);
+            if (param == null) {
+                arg.checkOptional();
+                if (arg.defaultValue != null) {
+                    builder = applyArg(arg.defaultValue, arg, builder);
+                }
+            } else {
+                builder = applyArg(param, arg, builder);
+            }
+        } else {
+            builder = apply(element, builder);
+        }
+        return builder;
+    }
+
+    private Text.Builder applyArg(TextElement param, Arg arg, @Nullable Text.Builder builder) {
+        // wrap the parameter in the argument format
+        Text.Builder wrapper = Text.builder().format(arg.format);
+        param.applyTo(wrapper);
+        if (builder == null) {
+            builder = wrapper;
+        } else {
+            builder.append(wrapper.build());
+        }
+        return builder;
+    }
+
+    private Text.Builder apply(Object element, @Nullable Text.Builder builder) {
+        if (element instanceof Text) {
+            Text text = (Text) element;
+            if (builder == null) {
+                builder = text.toBuilder();
+            } else {
+                builder.append(text);
+            }
+        } else if (element instanceof TextElement) {
+            if (builder == null) {
+                builder = Text.builder();
+            }
+            ((TextElement) element).applyTo(builder);
+        } else {
+            String str = element.toString();
+            if (builder == null) {
+                builder = Text.builder(str);
+            } else {
+                builder.append(Text.of(str));
+            }
+        }
+        return builder;
+    }
+
+    /**
+     * Constructs a new TextTemplate for the given elements. The order of the
+     * elements is the order in which they will be appended to the result
+     * builder via {@link #apply(Map)}.
+     *
+     * <p>The provided elements may be of any type.</p>
+     *
+     * <p>In the case that an element is a {@link TextElement},
+     * {@link TextElement#applyTo(Text.Builder)} will be used to append the
+     * element to the builder.</p>
+     *
+     * <p>In the case that an element is an {@link Arg} the argument will be
+     * replaced with the {@link TextElement} provided by the corresponding
+     * parameter supplied by {@link #apply(Map)}</p>
+     *
+     * <p>In the case that an element is any other type, the parameter value's
+     * {@link Object#toString()} method will be used to create a {@link Text}
+     * object.</p>
+     *
+     * @param elements Elements to append to builder
+     * @param openArg String to use for beginning of Arg containers
+     * @param closeArg String to use for end of Arg containers
+     * @return Newly constructed TextTemplate
+     */
+    public static TextTemplate of(String openArg, String closeArg, Object[] elements) {
+        checkNotNull(openArg, "open arg");
+        checkArgument(!openArg.isEmpty(), "open arg cannot be empty");
+        checkNotNull(closeArg, "close arg");
+        checkArgument(!closeArg.isEmpty(), "close arg cannot be empty");
+        checkNotNull(elements, "elements");
+        if (elements.length == 0) {
+            return of();
+        }
+        return new TextTemplate(openArg, closeArg, elements);
+    }
+
+    /**
+     * Constructs a new TextTemplate for the given elements. The order of the
+     * elements is the order in which they will be appended to the result
+     * builder via {@link #apply(Map)}.
+     *
+     * <p>The provided elements may be of any type.</p>
+     *
+     * <p>In the case that an element is a {@link TextElement},
+     * {@link TextElement#applyTo(Text.Builder)} will be used to append the
+     * element to the builder.</p>
+     *
+     * <p>In the case that an element is an {@link Arg} the argument will be
+     * replaced with the {@link TextElement} provided by the corresponding
+     * parameter supplied by {@link #apply(Map)}</p>
+     *
+     * <p>In the case that an element is any other type, the parameter value's
+     * {@link Object#toString()} method will be used to create a {@link Text}
+     * object.</p>
+     *
+     * @param elements Elements to append to builder
+     * @return Newly constructed TextTemplate
+     */
+    public static TextTemplate of(Object... elements) {
+        return of(DEFAULT_OPEN_ARG, DEFAULT_CLOSE_ARG, elements);
+    }
+
+    /**
+     * Returns the empty representation of a TextTemplate.
+     *
+     * @return Empty TextTemplate
+     */
+    public static TextTemplate of() {
+        return EMPTY;
+    }
+
+    /**
+     * Constructs a new {@link Arg} to be supplied to {@link #of(Object...)}.
+     * This argument expects a {@link TextElement} parameter.
+     *
+     * @param name name of argument
+     * @return argument builder
+     */
+    public static Arg.Builder arg(String name) {
+        return new Arg.Builder(name);
+    }
+
+    @Override
+    public Text toText() {
+        return this.text;
+    }
+
+    @Override
+    public Iterator<Object> iterator() {
+        return this.elements.iterator();
+    }
+
+    @Override
+    public String toString() {
+        return Objects.toStringHelper(this)
+                .add("elements", this.elements)
+                .add("arguments", this.arguments)
+                .add("text", this.text)
+                .add("openArg", this.openArg)
+                .add("closeArg", this.closeArg)
+                .toString();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(this.elements, this.openArg, this.closeArg);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!(obj instanceof TextTemplate)) {
+            return false;
+        }
+        TextTemplate that = (TextTemplate) obj;
+        return that.elements.equals(this.elements)
+                && that.openArg.equals(this.openArg)
+                && that.closeArg.equals(this.closeArg);
+    }
+
+    /**
+     * Represents a variable element within a TextTemplate. Arguments are
+     * replaced by parameters in {@link #apply(Map)}.
+     */
+    @ConfigSerializable
+    public final static class Arg implements TextRepresentable {
+
+        @Setting final boolean optional;
+        @Setting @Nullable final Text defaultValue;
+        final String name; // defined by node name
+        final TextFormat format; // defined in "content" node
+        final String openArg;
+        final String closeArg;
+
+        Arg(String name, boolean optional, @Nullable Text defaultValue, TextFormat format, String openArg, String closeArg) {
+            this.name = name;
+            this.optional = optional;
+            this.defaultValue = defaultValue;
+            this.format = format;
+            this.openArg = openArg;
+            this.closeArg = closeArg;
+        }
+
+        Arg(String name, boolean optional, @Nullable Text defaultValue, TextFormat format) {
+            this(name, optional, defaultValue, format, DEFAULT_OPEN_ARG, DEFAULT_CLOSE_ARG);
+        }
+
+        Arg(Arg arg, String openArg, String closeArg) {
+            this(arg.name, arg.optional, arg.defaultValue, arg.format, openArg, closeArg);
+        }
+
+        void checkOptional() {
+            if (!this.optional) {
+                throw new TextTemplateArgumentException("Missing required argument in TextTemplate \"" + this.name + "\".");
+            }
+        }
+
+        /**
+         * Returns the name of this argument to be matched with incoming
+         * parameters.
+         *
+         * @return Argument name
+         */
+        public String getName() {
+            return this.name;
+        }
+
+        /**
+         * Returns true if this Arg is optional. If a parameter is missing for
+         * a non-optional Arg, a {@link TextTemplateArgumentException} will be
+         * thrown.
+         *
+         * @return True if optional
+         */
+        public boolean isOptional() {
+            return this.optional;
+        }
+
+        /**
+         * Returns the default value to use if the Arg {@link #isOptional()}
+         * and no parameter is supplied.
+         *
+         * @return Default value
+         */
+        public Optional<Text> getDefaultValue() {
+            return Optional.ofNullable(this.defaultValue);
+        }
+
+        /**
+         * Returns the base format to be applied to this Arg.
+         *
+         * @return Base format
+         */
+        public TextFormat getFormat() {
+            return this.format;
+        }
+
+        /**
+         * Returns the beginning string of the Arg's container.
+         *
+         * @return Open string
+         */
+        public String getOpenArgString() {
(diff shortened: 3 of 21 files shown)


src/main/java/org/spongepowered/api/text/TextElement.java
```diff
@@ -0,0 +1,38 @@
...
+ * Represents anything that can be appended to a {@link Text.Builder}.
+ */
+public interface TextElement {
+
+    /**
+     * Appends this element to the specified builder.
+     *
+     * @param builder to append to
+     */
+    void appendTo(Text.Builder builder);
```
liach: Note that you need a new line after the last class member.

liach: And new line before the first class member, as in those custom exceptions.

windy1: Fixed, thanks for pointing that out @liach

Minecrell: I think you should name this differently, e.g. `apply` or something because most of the implementations (e.g. on color) don't actually _append_ anything.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,214 @@
...
+    public void send(MessageReceiver receiver, Map<String, Object> params) {
+        receiver.sendMessage(eval(params).build());
+    }
+
+    /**
+     * Constructs a new TextTemplate for the given elements. The order of the
+     * elements is the order in which they will be appended to the result
+     * builder via {@link #eval(Map)}.
+     *
+     * The provided elements may be of any type.
```
Minecrell: There should be `<p>` `</p>` around all extra paragraphs (so they actual get displayed as paragraphs)

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * In the case that an element is any other type, the parameter value's
+     * {@link Object#toString()} method will be used to create a {@link Text}
+     * object.
+     *
+     * @param elements Elements to append to builder
+     * @return Newly constructed TextTemplate
+     */
+    public static TextTemplate of(Object... elements) {
+        return new TextTemplate(elements);
```
Minecrell: It would be good to have this do something special if elements is empty (e.g. return constant which results in an empty builder everytime or throw an exception)

src/main/java/org/spongepowered/api/text/TextElement.java
windy1: Agreed, renamed.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
windy1: Fixed.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
windy1: Added some preconditions to the class as a whole as well as an EMPTY constant.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+     * {@link Text.Builder}.
+     *
+     * @return Text builder containing result
+     * @throws RequiredTextTemplateArgumentException if required parameters are missing
+     */
+    public Text.Builder eval() {
+        return eval(Collections.emptyMap());
+    }
+
+    private void appendElement(Text.Builder builder, Object element, Map<String, Object> params) {
```
kashike: Along with previous method regarding visibility, this would be nice to have `protected`.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+     * Evaluates this TextTemplate and returns the result in a
+     * {@link Text.Builder}.
+     *
+     * @param params Parameters to evaluate
+     * @return Text builder containing result
+     * @throws RequiredTextTemplateArgumentException if required parameters are missing
+     */
+    public Text.Builder eval(Map<String, Object> params) {
```
kashike: `parse`? `eval` sounds odd.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+        this.elements = elements;
+    }
+
+    /**
+     * Returns the elements contained in this TextTemplate.
+     *
+     * @return The elements within the template
+     */
+    public List<Object> getElements() {
+        return Collections.unmodifiableList(Arrays.asList(elements));
```
kashike: Could do `return ImmutableList.copyOf(this.elements);` instead.

Minecrell: Or if you just iterate it and do nothing array specific, you could just do the `copyOf` in the constructor, store that instead and return it directly.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Empty representation of a {@link TextTemplate}. This is returned if the
+     * array supplied to {@link #of(Object...)} is empty.
+     */
+    public static final TextTemplate EMPTY = new TextTemplate();
+
+    final Object[] elements;
+
+    TextTemplate(Object... elements) {
+        this.elements = elements;
```
Minecrell: Actually this **should** make a copy of some sort (preferably `ImmutableList.copyOf`), because otherwise you can still modify it from the outside by modifying the array given to the method below, which doesn't make much sense.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
ST-DDT: Or just `apply` as well.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param type Type of parameter
+     * @param f Function that takes "type" as the input and returns Text as the output
+     * @param optional Whether this argument is allowed to be omitted
+     * @param <T> Input type of argument
+     * @return Newly constructed Arg
+     */
+    public static <T> Arg<T> arg(String name, Class<T> type, Function<T, TextElement> f, boolean optional) {
+        checkNotNull(name, "name");
+        checkArgument(!name.isEmpty(), "name");
+        checkNotNull(type, "argument type");
```
ST-DDT: IMO this type is just overhead. Just let the default ClassCastException explode or wrap it if it occurs.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param name Name of parameter
+     * @param type Type of parameter
+     * @param f Function that takes "type" as the input and returns Text as the output
+     * @param optional Whether this argument is allowed to be omitted
+     * @param <T> Input type of argument
+     * @return Newly constructed Arg
+     */
+    public static <T> Arg<T> arg(String name, Class<T> type, Function<T, TextElement> f, boolean optional) {
+        checkNotNull(name, "name");
+        checkArgument(!name.isEmpty(), "name");
```
Minecrell: Should be something like `name cannot be empty`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
Mumfrey: It's a little known fact that it's also good practice to throw exceptions and catch them than to return from methods as it's much more efficient than just using a regular, boring old `return` statement. Sponge is committed to avoiding best practices at all costs and never follows conventions when they can be needlessly flouted. Likewise, type checking is overrated and unnecessary. I mean, secretly the JVM stores everything internally as an `Object` and derived classes are just a lie programmers tell themselves in order to justify making code harder to read.

ST-DDT: @Mumfrey IMO there is no readability improvement here, except there is one method that contains the same details in less characters:
Example Usage:
````java
TextTemplate.arg("player", Player.class, Player::getName);
TextTemplate.arg("player", Player::getName);

TextTemplate.arg("player", Player.class, ConversionUtil::prettyPlayerName);
TextTemplate.arg("player", ConversionUtil::prettyPlayerName);

TextTemplate.arg("player", Player.class, arena::getScore);
TextTemplate.arg("player", arena::getScore);

TextTemplate.arg("player", Player.class, player -> player.getName());
TextTemplate.arg("player", (Player player) -> player.getName());
TextTemplate.<Player>arg("player", player -> player.getName());
````

> It's a little known fact that it's also good practice to throw exceptions and catch them

Can you explain how this one is any better than the other one?

````java
try {
    value = textTemplate.eval(params);
} catch (MismatchedTextTemplateArgumentException e) {
    value = Text.of("Well something wen't wrong but thats a programming error so I don't know how to handle it here anyway: " + e.getMessage());
}
````
vs
````java
try {
    value = textTemplate.eval(params);
} catch (ClassCastException e) {
    value = Text.of("Well something wen't wrong but thats a programming error so I don't know how to handle it here anyway: " + e.getMessage());
}
````
If you have the wrong type in your param mapping it is too late.

Comment (Deamon5550) [inline src/main/java/org/spongepowered/api/text/template/TextTemplate.java]: Except that the message on the MismatchedTextTemplateArgumentException is far more useful than a random ClassCastException. I'm not sure what you're trying to say with those examples since catching either exception is a little odd to say the least.

This isn't a code golf challenge we are allowed to use additional characters to follow some good and standard practices.

Comment (ST-DDT) [inline src/main/java/org/spongepowered/api/text/template/TextTemplate.java]: The examples show the usage with and without the type param.

If the message is the issue just create it appropriately?!

IMO the entire eval method should be wrapped in a try catch if the message is your issue:

````java
eval(Object param) {
try {
    return f.apply((T) param);
} catch (RuntimeException | ClassCastException e) {
    throw new ArgumentEvalException("Failed to evaluate param " + name + " with value " + param + " in function " + f, e);
}
````

File: src/main/java/org/spongepowered/api/text/template/TextTemplate.java [hunk 10]
```diff
@@ -0,0 +1,244 @@
...
+
+        /**
+         * Evaluates the argument for the specified parameter.
+         *
+         * @param param Parameter to evaluate
+         * @return TextElement output
+         */
+        @SuppressWarnings("unchecked")
+        TextElement eval(Object param) {
+            Class<?> paramType = param.getClass();
```
ST-DDT: Potential NPE here.
Null should be an allowed value.

windy1: Not really, this method is not accessible externally and is only called with non-null params. Null parameters are skipped if they are optional.

ST-DDT: Is there any other way to default/fallback to a certain value?

windy1:
```java
param.put("key", value == null ? fallback : value);
template.eval(params);
```

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,244 @@
...
+     *
+     * @param name Name of parameter
+     * @param type Type of parameter
+     * @param f Function that takes "type" as the input and returns Text as the output
+     * @param optional Whether this argument is allowed to be omitted
+     * @param <T> Input type of argument
+     * @return Newly constructed Arg
+     */
+    public static <T> Arg<T> arg(String name, Class<T> type, Function<T, TextElement> f, boolean optional) {
+        checkNotNull(name, "name");
```
ST-DDT: These checks should be moved to the ctor to match the rest of the API.

windy1: It's more efficient to check the parameters before invoking a new Arg object.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
ST-DDT: The fallback is hard to fake if it is a Player or any other komplex interface. But i could optionals...

windy1: Not sure what you mean by it's "hard to fake". Hard to fake what?

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
ST-DDT: Yes, it safes you from creating a new instance in the case that some params are wrong. But invalid input is not to be expected in this method in prod systems. And on dev systems it does not matter since the bad call is a programming bug that must be fixed anyway and 10us are neglectible. And this prevents bypassed checks because something is added later on/called internally.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
windy1: Fixed.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
gabizou: It's better to have the precondition checks before invoking a new instance. All around, if you're worried about adding a new feature on top and the check isn't used, that's where the simple answer "Unit tests" come into play. Since Text API is almost all self contained, it's very well possible to unit test a majority of these types of things.

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+        this.elements = ImmutableList.copyOf(elements);
+    }
+
+    /**
+     * Returns the elements contained in this TextTemplate.
+     *
+     * @return The elements within the template
+     */
+    public List<Object> getElements() {
+        return elements;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+     * an input type and all arguments return a {@link TextElement} as it's
+     * output.
+     *
+     * @param <T> Input type
+     */
+    static final class Arg<T> {
+        final String name;
+        final Class<T> type;
+        final Function<T, TextElement> f;
+        final boolean optional;
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+     * Represents a variable element within a TextTemplate. Each argument has
+     * an input type and all arguments return a {@link TextElement} as it's
+     * output.
+     *
+     * @param <T> Input type
+     */
+    static final class Arg<T> {
+        final String name;
+        final Class<T> type;
+        final Function<T, TextElement> f;
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * Represents a variable element within a TextTemplate. Each argument has
+     * an input type and all arguments return a {@link TextElement} as it's
+     * output.
+     *
+     * @param <T> Input type
+     */
+    static final class Arg<T> {
+        final String name;
+        final Class<T> type;
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * Represents a variable element within a TextTemplate. Each argument has
+     * an input type and all arguments return a {@link TextElement} as it's
+     * output.
+     *
+     * @param <T> Input type
+     */
+    static final class Arg<T> {
+        final String name;
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Represents a variable element within a TextTemplate. Each argument has
+     * an input type and all arguments return a {@link TextElement} as it's
+     * output.
+     *
+     * @param <T> Input type
+     */
+    static final class Arg<T> {
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+        final boolean optional;
+
+        /**
+         * Evaluates the argument for the specified parameter.
+         *
+         * @param param Parameter to evaluate
+         * @return TextElement output
+         */
+        @SuppressWarnings("unchecked")
+        TextElement eval(Object param) {
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+            Class<?> paramType = param.getClass();
+            if (!this.type.isAssignableFrom(paramType)) {
+                throw new MismatchedTextTemplateArgumentException(this.type, paramType, this.name);
+            }
+            return this.f.apply((T) param);
+        }
+
+        Arg(String name, Class<T> type, Function<T, TextElement> f, boolean optional) {
```
kashike: `protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
```diff
@@ -0,0 +1,245 @@
...
+            this.name = name;
+            this.type = type;
+            this.f = f;
```
kashike: x`protected`

src/main/java/org/spongepowered/api/text/template/TextTemplate.java
kashike: `function`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+                // check for non-equal duplicate argument
+                Arg newArg = (Arg) element;
+                newArg.openArg = this.openArg;
+                newArg.closeArg = this.closeArg;
+                Arg oldArg = this.arguments.get(newArg.name);
+                if (oldArg != null && !oldArg.equals(newArg)) {
+                    throw new TextTemplateArgumentException("Tried to supply an unequal argument with a duplicate name \""
+                            + newArg.name + "\" to TextTemplate.");
+                }
+                arguments.put(newArg.name, newArg);
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        return ImmutableBiMap.copyOf(this.arguments);
+    }
+
+    /**
+     * Returns the string used for containing Args within the template.
+     *
+     * @return String containing args
+     */
+    public String getOpenArgString() {
+        return openArg;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /**
+     * Returns the string used for containing Args within the template.
+     *
+     * @return String containing args
+     */
+    public String getCloseArgString() {
+        return closeArg;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        /**
+         * Returns the name of this argument to be matched with incoming
+         * parameters.
+         *
+         * @return Argument name
+         */
+        public String getName() {
+            return name;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+
+        /**
+         * Returns true if this Arg is optional. If a parameter is missing for
+         * a non-optional Arg, a {@link TextTemplateArgumentException} will be
+         * thrown.
+         *
+         * @return True if optional
+         */
+        public boolean isOptional() {
+            return optional;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        /**
+         * Returns the base format to be applied to this Arg.
+         *
+         * @return Base format
+         */
+        public TextFormat getFormat() {
+            return format;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        /**
+         * Returns the beginning string of the Arg's container.
+         *
+         * @return Open string
+         */
+        public Optional<String> getOpenArgString() {
+            return Optional.ofNullable(openArg);
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        /**
+         * Returns the end string of the Arg's container.
+         *
+         * @return Close string
+         */
+        public Optional<String> getCloseArgString() {
+            return Optional.ofNullable(closeArg);
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        return ImmutableList.copyOf(this.elements);
+    }
+
+    /**
+     * Returns the arguments contained within the TextTemplate.
+     *
+     * @return The arguments within this TextTemplate
+     */
+    public Map<String, Arg> getArguments() {
+        return ImmutableBiMap.copyOf(this.arguments);
```
kashike: ...what? The original is not a `BiMap`. I assume you meant `ImmutableMap.copyOf(this.arguments)`?

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
```diff
@@ -0,0 +1,131 @@
...
+ * "arguments" nodes. If so, we use the name and format from the original
+ * LiteralText and obtain whether the argument is optional from the definition.
+ * This is repeated until there are no more Text children to check and we
+ * return a TextTemplate of the elements we have collected.</p>
+ */
+public class TextTemplateConfigSerializer implements TypeSerializer<TextTemplate> {
+
+    private static final String NODE_CONTENT = "content";
+    private static final String NODE_ARGS = "arguments";
+    private static final String NODE_OPT = "optional";
```
kashike: `NODE_OPTIONAL` to be consistent with naming

windy1: Private field, shorter name, I honestly dgaf to be frank lol

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+                return false;
+            }
+            Arg that = (Arg) obj;
+            return that.name.equals(this.name) && that.optional == this.optional;
+        }
+
+        /**
+         * Represents a builder for {@link Arg}s.
+         */
+        public static class Builder {
```
kashike: Majority of classes have a new line between here and the rest of the class

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+        }
+        return Optional.ofNullable(builder).orElse(Text.builder()).build();
+    }
+
+    /**
+     * Represents a variable element within a TextTemplate. Arguments are
+     * replaced by parameters in {@link #apply(Map)}.
+     */
+    @ConfigSerializable
+    public static class Arg implements TextRepresentable {
```
kashike: Is a `toString` possible on this class?

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,473 @@
...
+     *
+     * @param name name of argument
+     * @return argument builder
+     */
+    public static Arg.Builder arg(String name) {
+        return new Arg.Builder(name);
+    }
+
+    @Override
+    public Text toText() {
```
kashike: Is it possible to have a `toText(Text.Builder)` method? It'd be nice to pass an already-made builder here, if available

src/main/java/org/spongepowered/api/text/TextElement.java
```diff
@@ -0,0 +1,39 @@
...
+ * Represents anything that can be applied to a {@link Text.Builder}.
+ */
+public interface TextElement {
+
+    /**
+     * Applies this element to the end of the specified builder.
+     *
+     * @param builder Text builder to apply to
+     */
+    void applyTo(Text.Builder builder);
```
kashike: Could this interface `extend Consumer<Text.Builder>` and use the `accept` method instead, or is this not desired?

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,475 @@
...
+
+    /**
+     * Empty representation of a {@link TextTemplate}. This is returned if the
+     * array supplied to {@link #of(Object...)} is empty.
+     */
+    public static final TextTemplate EMPTY = new TextTemplate(DEFAULT_OPEN_ARG, DEFAULT_CLOSE_ARG, new Object[]{});
+
+    protected final List<Object> elements = new ArrayList<>();
+    protected final Map<String, Arg> arguments = new HashMap<>();
+    protected final String openArg, closeArg;
```
kashike: Separate these

src/main/java/org/spongepowered/api/text/TextTemplate.java
windy1: That seems a bit misleading, we want toText() to return the exact Text representation of the template. Adding another method here only blurs that line and also is redundant when you can just do

```java
builder.append(template.toText());
```

src/main/java/org/spongepowered/api/text/TextElement.java
liach: No it has a quite different meaning.

windy1: `TextElement` is not a functional interface so again I think having it extend Consumer is misleading.

src/main/java/org/spongepowered/api/text/TextTemplate.java
kashike: Fair enough

src/main/java/org/spongepowered/api/text/TextElement.java
kashike: Fair enough.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -77,6 +80,10 @@
  */
 public abstract class Text implements TextRepresentable {
 
+    static {
+        TypeSerializers.getDefaultSerializers().registerType(TypeToken.of(Text.class), new TextConfigSerializer());
```
gabizou: By making `Text` serializable through Configurate, it should very well be possible that you can make `Text` itself `DataSerializable` since it's very similar with regards to making a `DataBuilder` for a `Text` object. It would certainly allow for using `Text` objects within `DataContainer`s far easier in plugin usage.

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final String NODE_OPEN_ARG = "openArg";
+    private static final String NODE_CLOSE_ARG = "closeArg";
+
+    @Override
+    public TextTemplate deserialize(TypeToken<?> type, ConfigurationNode value) throws ObjectMappingException {
+        String openArg = value.getNode(NODE_OPEN_ARG).getString(TextTemplate.DEFAULT_OPEN_ARG);
+        String closeArg = value.getNode(NODE_CLOSE_ARG).getString(TextTemplate.DEFAULT_CLOSE_ARG);
+        Text content = value.getNode(NODE_CONTENT).getValue(TypeToken.of(Text.class));
+        List<Object> elements = new ArrayList<>();
+        elements.add(content.getFormat());
```
windy1: Adding note for myself: Need to recurse through *all* children for each Text object including children's children, children's children's children, etc.

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,499 @@
...
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a re-usable template that produces a formatted
+ * {@link Text.Builder}. Elements will be appended to the result builder in the
+ * order that they are specified in {@link #of(Object...)}.
+ */
+public class TextTemplate implements TextRepresentable {
```
Minecrell: Is `TextTemplate` supposed to be immutable like text?

src/main/java/org/spongepowered/api/text/TextTemplate.java
```diff
@@ -0,0 +1,511 @@
...
+
+    final ImmutableList<Object> elements;
+    final ImmutableMap<String, Arg> arguments;
+    final String openArg;
+    final String closeArg;
+
+    TextTemplate(String openArg, String closeArg, Object[] elements) {
+        this.openArg = openArg;
+        this.closeArg = closeArg;
+        List<Object> elementList = new ArrayList<>();
```
Minecrell: Using `ImmutableList.builder()` would probably simplify this even a bit more.

src/main/java/org/spongepowered/api/text/TextTemplate.java
windy1: Yes, I've strengthened that definition a bit now in https://github.com/SpongePowered/SpongeAPI/commit/747203f1620870b39a7d643f443c4be0b62a7649

windy1:
Happy to report TextTemplates are now fully functional. Could use some testers to try and break it with fringe use-cases but so far I have been unable to do so myself.

windy1:
This is pretty much ready to be merged now. Work on localization is being done in https://github.com/SpongePowered/SpongeAPI/tree/feature/locale

src/main/java/org/spongepowered/api/text/serializer/TextConfigSerializer.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Represents a {@link TypeSerializer} for {@link Text} objects. Serialization
+ * is handled by serializing the text to String with the
+ * {@link TextSerializers#JSON} serializer, loading the String into a
+ * {@link GsonConfigurationLoader}, and setting the value of the
+ * {@link ConfigurationNode} to the root node of the GsonConfigurationLoader.
+ * Although JSON is used for serialization internally, this has no effect on
+ * the actual configuration format the developer chooses to use.
+ */
+public class TextConfigSerializer implements TypeSerializer<Text> {
```
kashike: final?

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
```diff
@@ -0,0 +1,128 @@
...
+ * <p>Deserialization is a bit more complicated. We start by loading the
+ * "content" Text and check the root Text element as well as it's children. If
+ * a {@link LiteralText} value is found that is wrapped in curly braces we
+ * check to see if the value inside the braces is defined as an argument in the
+ * "arguments" nodes. If so, we use the name and format from the original
+ * LiteralText and obtain whether the argument is optional from the definition.
+ * This is repeated until there are no more Text children to check and we
+ * return a TextTemplate of the elements we have collected.</p>
+ */
+public class TextTemplateConfigSerializer implements TypeSerializer<TextTemplate> {
```
kashike: final?

src/main/java/org/spongepowered/api/text/serializer/TextConfigSerializer.java
windy1: I think developers should be allowed to extend this IMO so they can serialize text in their own way.

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
windy1: Same as above.

src/main/java/org/spongepowered/api/text/Text.java
gabizou: @windy1 I don't see the `DataSerializable` changes yet, unless you changed your mind?

windy1: Totally forgot about this :p I'll look at it later although I have pretty much zero prior experience with the data API.

gabizou: If anything, I can explain it tomorrow afternoon or something.

CrystalMageX:
Do these `TextTemplates` nest? For example, if a user wanted to have three groups,*default*,  *playerOfTheMonth* and *vip*, each having the `TextTemplates` *<playername> message*, *!!!playername!!!*, and *[VIP] playername*, and I made *vip* a child of *playerOfTheMonth*, would applying a the second template to the third, then that to the first, caching the result, and applying a String to that work? Will a TextTemplate be able return another TextTemplate?

Flow of nested TextTemplates:
```
Player "Notch"
........|----------- Group default - "<playername> message"
........|----------- Group playerOfTheMonth - "!!!playername!!!"
................|----------- Group vip - "[VIP] playername"

Apply template playerOfTheMonth to template default to get another template:
TextTemplate("<playername> message")
+ Map("playername", TextTemplate("!!!playername!!!"))
= TextTemplate("<!!!playername!!!> message)

Apply template vip to the template we got above to get another template.
TextTemplate("<!!!playername!!!> message")
+ Map("playername", TextTemplate("[VIP] playername")
= TextTemplate("<!!![VIP] playername!!!> message")

Apply Notch's chat, "hi", to this template:
TextTemplate("<!!![VIP] playername!!!> message")
+ Map("playername", "Notch", "message", "Hi")
= LiteralText("<!!![VIP] Notch!!!> Hi")
```

Is this possible, and if so how?

windy1:
@CrystalMageX I have no plans to support nested templates at this time. I toyed with the idea for a bit but I came to a conclusion that there are too many variables involved with text templates (argument validation, open/close arguments) too make nested templates desirable.

windy1:
A little preview from the `feature/messages` branch:

https://gist.github.com/windy1/bfcf2636b42b4fa9e1da#file-myplugin-java

kashike:
@windy1 My thoughts on `feature/messages`:
- `MessageEvent`'s `getMessage` method should return `Optional<Text>`, with `setMessage(@Nullable Text)`
 - This allows "cancelling" the sending of a message, because making this event `Cancellable` won't work
   - Cancelling this event would cause join, etc events to be cancelled, which is unwanted.
 - Allows someone to check if the message is not set before setting it again

What is this "header" and "footer" stuff, I can't really tell?

ST-DDT:
@kashike Just set the message sink to `TO_NONE`. 

And I don't get the header and footer stuff as well.

kashike:
@ST-DDT That doesn't solve anything. Mutating the **message** vs the **message channel**.

windy1:
@kashike `setMessage` is still nullable, `getMessage` now returns an empty text instead of an optional.

The idea behind the header, body, and footer is to partition the message into different parts to make it easier for developers to modify the message. Say I want to add a prefix to the message but I also want to keep existing prefixes.

```java
    TextTemplateApplier applier = new SimpleTextTemplateApplier(of('[', arg("prefix"), ']'));
    applier.setParameter("prefix", Text.of("whatever"));
    event.getFormatter().getHeader().add(applier);
}
```

This will append `"[whatever]"` to the header. Additionally someone else can set "prefix" to something else since "whatever" is not applied until the event cycle is finished.

kashike:
@windy1 `getMessage` **intentionally* returns `Optional<Text>` because sending empty text is **valid**, and there must be a way to differentiate from "empty" and "not present"/"don't send".

windy1:
@kashike Why would you *ever* want to send empty text.

kashike:
@windy1 Sending empty text is valid - an example is filling the chat with empty messages to clear it, etc.

kashike:
https://github.com/SpongePowered/SpongeForge/issues/435
https://github.com/SpongePowered/SpongeCommon/commit/bc2947137e14830d85ee0c8e720e8a8bfeee8cfc
and a few more, but I'm not going to go tracking them down

windy1:
Fair point @kashike. I'll change it back later.

src/main/java/org/spongepowered/api/text/serializer/TextConfigSerializer.java
```diff
@@ -0,0 +1,79 @@
...
+            gsonLoader.save(value);
+        } catch (IOException e) {
+            throw new ObjectMappingException(e);
+        }
+        return TextSerializers.JSON.deserialize(writer.getBuffer().toString());
+    }
+
+    @Override
+    public void serialize(TypeToken<?> type, Text obj, ConfigurationNode value) throws ObjectMappingException {
+        String json = (String) obj.toContainer().get(Queries.JSON).get();
```
Minecrell: Any reason you don't call the text serializer directly here as above?

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static final String NODE_CONTENT = "content";
+    private static final String NODE_ARGS = "arguments";
+    private static final String NODE_OPT = "optional";
+    private static final String NODE_DEF_VAL = "defaultValue";
+    private static final String NODE_OPTIONS = "options";
+    private static final String NODE_OPEN_ARG = "openArg";
+    private static final String NODE_CLOSE_ARG = "closeArg";
+
+    @SuppressWarnings("NullableProblems") private ConfigurationNode root;
```
Minecrell: Why suppress `NullableProblems` here?

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
```diff
@@ -0,0 +1,129 @@
...
+        List<Object> elements = new ArrayList<>();
+        parse(content, elements);
+        return TextTemplate.of(elements.toArray(new Object[elements.size()]));
+    }
+
+    @Override
+    public void serialize(TypeToken<?> type, TextTemplate obj, ConfigurationNode value) throws ObjectMappingException {
+        value.getNode(NODE_OPTIONS, NODE_OPEN_ARG).setValue(obj.getOpenArgString());
+        value.getNode(NODE_OPTIONS, NODE_CLOSE_ARG).setValue(obj.getCloseArgString());
+        value.getNode(NODE_ARGS).setValue(new TypeToken<Map<String, TextTemplate.Arg>>() {}, obj.getArguments());
```
Minecrell: IMO you should store all these type tokens in a constant somewhere.

src/main/java/org/spongepowered/api/text/transform/TextFormatter.java
```diff
@@ -0,0 +1,214 @@
...
+    ImmutableList<E> getAll();
+
+    /**
+     * Returns the element at the specified index.
+     *
+     * @param i Index to retrieve from
+     * @return Element at index
+     */
+    default E get(int i) {
+        return getAll().get(i);
```
Minecrell: It shouldn't be necessary to make a copy of the whole list before being able to get a single element.

src/main/java/org/spongepowered/api/text/serializer/TextConfigSerializer.java
windy1: Not really, changed it to make sure the container was working and just never changed it back.

src/main/java/org/spongepowered/api/text/serializer/TextTemplateConfigSerializer.java
windy1: Suppresses the warning that it has to be initialized, which it is by deserialized, but that's not recognized. Could make it Nullable but then it would complain if you don't check if it's not null which is unnecessary.

src/main/java/org/spongepowered/api/text/SelectorText.java
```diff
@@ -113,6 +113,11 @@ public int hashCode() {
                 .addValue(this.selector);
     }
 
+    @Override
+    public int compareTo(Text o) {
+        return 0;
```
ST-DDT: This should be implemented or add a TODO marker

src/main/java/org/spongepowered/api/text/transform/SimpleTextFormatter.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * A basic implementation of {@link TextFormatter} backed by an
+ * {@link ArrayList}.
+ */
+public class SimpleTextFormatter implements TextFormatter<SimpleTextTemplateApplier> {
```
ST-DDT: How are these formatter any more usefull than a simple list of text builders or text templates?

windy1:
I would like to try and get this merged today, so unless anyone has any pressing objections...speak now or forever hold your peace.

src/main/java/org/spongepowered/api/text/transform/FixedPartitionedTextFormatter.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.util.Collection;
+
+/**
+ * Implementation of {@link PartitionedTextFormatter} that has a fixed amount
+ * of partitions.
+ */
+public class FixedPartitionedTextFormatter implements PartitionedTextFormatter<SimpleTextFormatter> {
+
+    final SimpleTextFormatter[] partitions;
```
kashike: `protected`?

src/main/java/org/spongepowered/api/text/transform/FixedPartitionedTextFormatter.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public ImmutableList<SimpleTextFormatter> getAll() {
+        return ImmutableList.copyOf(this.partitions);
+    }
+
+    @Override
+    public SimpleTextFormatter get(int i) {
+        return partitions[i];
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/transform/FixedPartitionedTextFormatter.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public SimpleTextFormatter set(int i, SimpleTextFormatter element) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public int size() {
+        return partitions.length;
```
kashike: `this.`

src/main/java/org/spongepowered/api/text/transform/FixedPartitionedTextFormatter.java
windy1: Nah, this array is supposed to be immutable.

kashike: make the class final?

windy1: No because while the array is immutable, the whole class isn't "immutable" by the standard definition. The elements within the partitions can be modified so we should allow for extensions of this class, just not modifications on the base array itself.

src/main/java/org/spongepowered/api/text/serializer/TextConfigSerializer.java
gabizou: This could also implement `DataBuilder<Text>` quite easily.

PR #1080, opened 2016-02-17
Add ability to set a custom access checker when using CommandSpec

CrystalMageX:
Instead of only allow *permissions* to be checked in `CommandSpec`, this will allow plugin developers to supply an *custom* ~~`Predicate<CommandSource>`~~ `Function<CommandSource, Optional<Text>>` instead. This is a **better idea** than putting the code in the beginning of the `CommandExecutor`, because the access control checking can be kept in a **central location** now, at `testPermission(CommandSource)`. When the returned `Optional`'s `Text` is not present, it means that there is no error message, so no error.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
+++ b/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
@@ -50,7 +50,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.function.Predicate;
+import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
@@ -63,14 +63,14 @@ public final class CommandSpec implements CommandCallable {
     private final CommandExecutor executor;
     private final Optional<Text> description;
     @Nullable private final Text extendedDescription;
-    @Nullable private final Predicate<CommandSource> accessChecker;
+    @Nullable private final Function<CommandSource, Optional<Text>> permissionTester;
     private final InputTokenizer argumentParser;
 
     private CommandSpec(CommandElement args, CommandExecutor executor, @Nullable Text description, @Nullable Text extendedDescription,
-            @Nullable Predicate<CommandSource> checker, InputTokenizer parser) {
+            @Nullable Function<CommandSource, Optional<Text>> permissionTester, InputTokenizer parser) {
         this.args = args;
         this.executor = executor;
-        this.accessChecker = checker;
+        this.permissionTester = permissionTester;
         this.description = Optional.ofNullable(description);
         this.extendedDescription = extendedDescription;
         this.argumentParser = parser;
@@ -97,7 +97,7 @@ public final class CommandSpec implements CommandCallable {
         @Nullable
         private Text extendedDescription;
         @Nullable
-        private Predicate<CommandSource> accessChecker;
+        private Function<CommandSource, Optional<Text>> permissionTester;
         @Nullable
         private CommandExecutor executor;
         @Nullable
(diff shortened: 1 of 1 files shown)


liach:
I like this, and I suggest you to squash commits.

kashike:
@CrystalMageX Why not call this method `testPermission` as well?

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -113,7 +113,21 @@ private Builder() {}
...
+            return this;
+        }
+
+        /**
+         * Set the permission tester that will be checked before using this command.
+         *
+         * @param tester The permission tester to consult
+         * @return this
+         */
+        public Builder permissionTester(Function<CommandSource, Optional<Text>> tester) {
```
kashike: This should just accept a `Predicate` in my opinion. Add a new method for the error message, will allow it to be handled nicer as well.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -266,9 +281,12 @@ public CommandSpec build() {
      * @throws CommandException if the source does not have permission
      */
     public void checkPermission(CommandSource source) throws CommandException {
-        checkNotNull(source, "source");
-        if (!testPermission(source)) {
-            throw new CommandPermissionException();
+        if (this.permissionTester != null) {
```
kashike: See above comment.

Zidane:
I don't want to pull this without a handy singleton Function somewhere that keeps the same functionality present here as many devs won't care to do more indepth permission checks.

CrystalMageX:
@kashike Having a `Function` instead of a `Predicate` would avoid having redundant code repeating the same code.

Consider the `Predicate` version of a magic plugin's */fireball* command:
```
public boolean hasAccess(CommandSource src) {
    if (!(src instanceof Player)) return false;
    Player p = (Player) src;

    if (MagicalFactions.WIZARD != getFaction(p)) return false;
    if (getLevel(p) < 10) return false;

    return true;
}

public Optional<Text> getError(CommandSource src) {
    if (src instanceof Player)
        return Optional.of(Text.of("You must be a player to cast spells!"));
    Player p = (Player) src;

    if (MagicalFactions.WIZARD != getFaction(p))
        return Optional.of(Text.of("You must be a wizard to firewall players!"));
    if (getLevel(p) < 10)
        return Optional.of(Text.of("You aren't a wise enough wizard yet for this skill!"));

    // should never happen and only needed in this Predicate version
    throw new IllegalStateException("???");
}
```

Consider the same command with a `Function`:
```
public Optional<Text> hasAccess(CommandSource src) {
    if (!(src instanceof Player))
        return Optional.of(Text.of("You must be a player to cast spells!"));
    Player p = (Player) src;

    if (MagicalFactions.WIZARD != getFaction(p))
        return Optional.of(Text.of("You must be a wizard to firewall players!"));
    if (getLevel(p) < 10)
        return Optional.of(Text.of("You aren't a wise enough wizard yet for this skill!"));

    // No error message, not even a generic one, means no error
    return Optional.empty();
}
```

The `Function` version is **9** lines shorter, does not repeat code that could have put bugs in multiple places, and does not have to worry about sending an error message when there is no error.

@Zidane The method `CommandSpec.builder().permission("some.permission.node")` is still present. (https://github.com/SpongePowered/SpongeAPI/pull/1080/files#diff-4201653595f5b8d8b5f941fef8403bdbR115). It has only be re-written *internally* to use the `permissionTester` field, to simplify the code. All current plugins using the `CommandSpec.builder().executor(commandExecutor).permission("pluginName.category,permission").build()` syntax will **still work**. 

> I don't want to pull this without a handy singleton Function somewhere that keeps the same functionality present here as many devs won't care to do more indepth permission checks.

The "handy singleton Function", `Builder.permission()`, remains available, and developers that don't care about more in depth permission checks can safely ignore `Builder.permissionTester()` without any effect on the behavior of their plugin.

Zidane:
@CrystalMageX 

Upon reviewing @kashike 's proposed change and your rebuttal, I still agree with @kashike solely on the premise that I seriously dislike the fact that a present Text object in the Optional means that they failed the permission check. That reeks of bad design.

Closing this as I think it needs re-thinking.

PR #1090, opened 2016-02-21
Fix casting issue preventing compile in mac with oracle jdk.

MatthewDolan:


CrystalMageX:
Java is a *platform* independent programming language. SpongeAPI, an API, is both *platform* and *implementation* independent. Any Java code that compiles on one OS should compile on any other OS. If this does not work on Mac, then it is **Apple's or Oracle's bug, not Sponge's.** Please report this bug to them instead. This change only makes the code less organized for non-Mac users.

Zidane:
@gabizou you use a mac and I assume Oracle's JDK on it...do you have this problem?

gabizou:
I've been working on Mac OS X for the entire lifetime of this project, I do know that older variants of JDK 8 have had issues with lambdas and method references, so if you can, please update your JDK and report back.

windy1:
I have gotten this compile error through IntelliJ but never through gradle itself.

gabizou:
> I have gotten this compile error through IntelliJ but never through gradle itself.

Which JDK did you give IntelliJ?

windy1:
Previously 1.8.0_20, I've updated to 1.8.0_71 and the issue appears to be resolved.

gabizou:
Yeah, it's mostly due to old JDK compilers having their bugs that have been fixed for a very long time now. Update your JDK and it should be fine.

Issue #1052, opened 2016-01-24
Add ability to make armor stand a "Marker"


MichaelPriebe:
Marker: 1 or 0 (true/false) - if true, ArmorStand's size will be set to 0, have a tiny hitbox and disable interactions with it. May not exist.

http://minecraft.gamepedia.com/Chunk_format#ArmorStand

phroa:
Is that 1 negative or positive? Other data values for that entity use positive 1 or 0

JBYoshi:
All boolean-like tags (byte tags that represent boolean values) are restricted to 0 and positive 1.

MichaelPriebe:
I just copied it from the wiki, seems they accidentally put that negative sign. Ill fix it.

PR #1100, opened 2016-02-24
Add armour stand marker data

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1100) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/533)

Implements https://github.com/SpongePowered/SpongeAPI/issues/1052


simon816:
Perhaps add `getMarkerData()` to `ArmorStand` since it's always available?

kashike:
I've reverted this to using methods on `ArmorStand` instead of adding data for this - someone should convert to data in the future, having parts in different places is bad.

gabizou:
Why isn't this data?

PR #1102, opened 2016-02-24
Detach LocatedSource from CommandSource heirarchy

windy1:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1102) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/535)

Pretty self-explanatory. This opens up the opportunity for a variety of use cases. For instance, say I want to track a player's eye location over time.

```java
Locatable locatable = () -> new Location(player.getWorld(), player.getProperty(EyeLocationProperty.class).get().getValue())
```

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Locatable.java
+++ b/src/main/java/org/spongepowered/api/world/Locatable.java
@@ -22,16 +22,13 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.command.source;
-
-import org.spongepowered.api.command.CommandSource;
-import org.spongepowered.api.world.Location;
-import org.spongepowered.api.world.World;
+package org.spongepowered.api.world;
 
 /**
- * Location sources are {@link CommandSource}s that have a current location.
+ * Location sources are anything that have a current location.
  */
-public interface LocatedSource extends CommandSource {
+@FunctionalInterface
+public interface Locatable {
 
     /**
      * Gets the location of the source.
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/world/Locatable.java
```diff
@@ -22,16 +22,13 @@
...
-import org.spongepowered.api.world.World;
+package org.spongepowered.api.world;
 
 /**
- * Location sources are {@link CommandSource}s that have a current location.
+ * Location sources are anything that have a current location.
  */
-public interface LocatedSource extends CommandSource {
+@FunctionalInterface
+public interface Locatable {
```
kashike: `<E extends Extent>` & `default E getExtent()`, assuming this isn't limited to `World` purposefully.

windy1: I honestly think limiting this two `World` is not such a bad idea. The command API at the moment pretty much expects world and while that could be changed / checked, if you have a world location then you can use that to retrieve a chunk location anyways. K.I.S.S

kashike: As far as I can see this isn't limited to the command API anymore, and people can technically create their own extensions of `Extent`

windy1: Well yes, and it very well may be worth generifying this. Just trying to play devils advocate.

windy1: @kashike Here's the solution I came up with https://github.com/SpongePowered/SpongeAPI/commit/b7df3bce478666c913934e48863fceb12a4f9b3f

kashike:
+1 to what @Zidane said, was about to say something similar

windy1:
@kashike Well then it is *impossible* to generify this without breaking commands completely. Can't have it both ways ;)

windy1:
Updates, thoughts on this?

PR #1104, opened 2016-02-25
Pass original plaintext through MessageChannel

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1104) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/553) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/228)

Resolves https://github.com/SpongePowered/SpongeCommon/issues/538

### DIFF
--- a/src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
+++ b/src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
@@ -132,7 +132,19 @@ public interface MessageChannel {
      * @param original The original text to send
      */
     default void send(Text original) {
-        this.send(null, original);
+        this.send(null, null, original);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


ST-DDT:
I disagree with this change.
As it does not work everywhere and adds more complexity to the MessageChannel.

IMO it would be better to pass the raw Text to the MessageChannel and let that add the playername and any prefixes.

kashike:
@ST-DDT The problem with that is unless a plugin/mod changes the `Text` in the event (or through a `MessageChannel`) the message will be sent without the player name.

kashike:
Forge passes both the formatted text and the original string as well: https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/common/ForgeHooks.java#L552-L561

PR #1111, opened 2016-02-28
Changes to MessageEvent

windy1:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1111) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/548) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/229) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/559)

This PR makes some substantial changes to how MessageEvents are handled. Instead of passing around a `Text` object, the event now contains a `MessageFormatter` which can be used to modify the event's message in a more reliable and modular fashion. This will make interactions and compatibility between chat plugins a lot easier. Previous methods like `getOriginalMessage`, `getRawMessage`, `getMessage`, and `setMessage` still behave as expected.

----

Some example uses:

```java
    public GroupApplier() {
    }
}
```

```java
@Listener
}
```

```java
```


windy1:
I seem to have royally messed up during rebasing this commit.

PR #1112, opened 2016-02-28
Changes to MessageEvent

windy1:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1111) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/548) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/229) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/559)

----

Some example uses:

```java
    public GroupApplier() {
    }
}
```

```java
@Listener
}
```

```java
```

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/TargetedLocationData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/TargetedLocationData.java
@@ -27,18 +27,19 @@ package org.spongepowered.api.data.manipulator.mutable;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.ImmutableTargetedLocationData;
 import org.spongepowered.api.data.value.mutable.Value;
-import org.spongepowered.api.entity.projectile.EnderPearl;
-import org.spongepowered.api.item.ItemTypes;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.entity.projectile.EyeOfEnder;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 
 /**
  * An {@link DataManipulator} handling the supposed targeted {@link Location}.
- * Usually for the case of {@link EnderPearl}s, the targeted {@link Location} is
- * where the {@link EnderPearl} will move towards until it's expiration time. In
- * the case of {@link ItemStack}s of type {@link ItemTypes#COMPASS}, the
- * targeted {@link Location} is where the compass will point towards.
+ * Usually for the case of {@link EyeOfEnder}s, the targeted {@link Location} is
+ * where the {@link EyeOfEnder} will move towards until it's expiration time. In
+ * the case of {@link Player}s, the targeted {@link Location} is where any
+ * compasses seen by the player will point towards (for example, a compass {@link ItemStack}
+ * in the player's inventory).
  */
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/KickPlayerEvent.java
```diff
@@ -25,11 +25,14 @@
...
 import org.spongepowered.api.event.message.MessageChannelEvent;
+import org.spongepowered.api.eventgencore.annotation.ImplementedBy;
 
 /**
  * Fired when a {@link Player} is kicked.
  */
+@ImplementedBy(AbstractMessageEvent.class)
 public interface KickPlayerEvent extends TargetPlayerEvent, MessageChannelEvent {
 
 }
```
kashike: Add a newline since you're editing this file anyway?

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
```diff
@@ -24,37 +24,254 @@
...
+
+    /**
+     * Parameter for body section in body partition.
+     */
+    String PARAM_MESSAGE_BODY = "body";
+
+    /**
+     * Parameter for footer section in footer partition.
+     */
+    String PARAM_MESSAGE_FOOTER = "footer";
```
kashike: "footer" doesn't make sense here, "suffix" does

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
```diff
@@ -24,37 +24,254 @@
...
 /**
  * Describes events when a involving a {@link Text} message.
  */
+@ImplementedBy(AbstractMessageEvent.class)
 public interface MessageEvent extends Event {
 
     /**
+     * Parameter for header section in header partition.
+     */
+    String PARAM_MESSAGE_HEADER = "header";
```
kashike: "header" doesn't make sense here, "prefix" does

src/main/java/org/spongepowered/api/service/permission/Subject.java
```diff
@@ -113,12 +112,15 @@
...
      * Test whether the subject is permitted to perform an action given as the
-     * given permission string.
+     * given permission string. This must return the same value as
+     * {@link #hasPermission(Set, String)} using {@link #getActiveContexts()}.
      *
      * @param permission The permission string
      * @return True if permission is granted
      */
-    boolean hasPermission(String permission);
+    default boolean hasPermission(String permission) {
```
kashike: Are these changes intentional? They seem unrelated to this PR.

kashike:
It's quite hard to tell what's part of this PR and what's from #1079, so I apologise if any comments are duplicates.

src/main/java/org/spongepowered/api/service/permission/Subject.java
windy1: These are @zml2008's changes. It's displayed in the diff because I branched off of bleeding, merged master into my new branch, and then applied my changes.

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
windy1: I disagree, footer is more general than suffix. You can add "suffixes" to the "footer" but that doesn't make the whole footer a suffix.

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
windy1: Same as above.

windy1:
@kashike Look at the individual commits instead of the overall diff. This is a result of merging master into my bleeding branch.

PR #1113, opened 2016-02-28
Append 'Minecart' to names instead of prepending

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1113) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/549)

This PR changes all of the Minecart entities to be appended with `Minecart` instead of pre-pended: `HopperMinecart` instead of `MinecartHopper`.
This is a difference compared to the rest of the API - for example: `Fireball` & `LargeFireball`, etc


DDoS:
:100:

PR #1118, opened 2016-02-28
Changes to MessageEvent

windy1:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1118) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/548) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/229) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/559)

----

Some example uses:

```java
    public GroupApplier() {
    }
}
```

```java
@Listener
}
```

```java
```


src/main/java/org/spongepowered/api/event/message/MessageEvent.java
```diff
@@ -24,37 +24,262 @@
...
 /**
  * Describes events when a involving a {@link Text} message.
  */
+@ImplementedBy(AbstractMessageEvent.class)
 public interface MessageEvent extends Event {
 
     /**
+     * Parameter for header section in header partition.
+     */
+    String PARAM_MESSAGE_HEADER = "header";
```
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1114#discussion_r54342803

src/main/java/org/spongepowered/api/event/message/MessageEvent.java
```diff
@@ -24,37 +24,262 @@
...
+
+    /**
+     * Parameter for body section in body partition.
+     */
+    String PARAM_MESSAGE_BODY = "body";
+
+    /**
+     * Parameter for footer section in footer partition.
+     */
+    String PARAM_MESSAGE_FOOTER = "footer";
```
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1114#discussion_r54342802

src/main/java/org/spongepowered/api/text/transform/TextFormatter.java
```diff
@@ -0,0 +1,329 @@
...
+     */
+    boolean add(E element);
+
+    /**
+     * Adds the specified elements to the end of this formatter.
+     *
+     * @param elements Elements to add
+     * @return True if the formatter changed as a result of the call
+     */
+    boolean add(Collection<E> elements);
```
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1114#discussion_r54342840

ST-DDT:
@windy1 Instead of continuously closing and opening new PRs you could just update them. 
This makes tracking the discussion a lot easier.

windy1:
@ST-DDT Trust me it's not intentional :P I keep messing up rebasing the feature/text-template branch....

JBYoshi:
@windy1 You can undo the changes with `git reflog` and then force-push the old versions.

ST-DDT:
Well you can still fix / recreate your changes and then force push to the branch associated with the PRs. Even if the original branch is horribly broken a simple force push with a working state will also fix the PR.

windy1:
Okay -- I'd like to get this merged ASAP now that TextTemplates are in. I'm still partial to "header, body, footer" but if someone can make a compelling argument against it other than "I like 'x' better" I'll change it.

PR #1120, opened 2016-03-01
Add ability to get/set forced bed location. Resolves #955

### LINKED ISSUE #955: Add a way to force non-bed spawns with RespawnLocationData

simon816:
#961 on the bleeding edge

Changes `Vector3d` to `Tuple<Vector3d, Boolean>`
See #955

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/RespawnLocation.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/RespawnLocation.java
@@ -1,0 +1,80 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


Mumfrey:
I have a couple of problems with this, though they're easily addressed.

I'm generally against the use of `Tuple`s as a lazy way of doing what really needs to be a custom struct (just as @gabizou, I [made him change some the other day](https://github.com/SpongePowered/SpongeCommon/commit/67972897c9e4ba51c1493b2d40b15b134b130d4b)) as it suffers from the same type of issue as *[the boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html)* in that it instantly obfuscates the purpose of the data structure in question. This is not to say there aren't good uses of `Tuple`, but this isn't one of them.

Secondly, this also falls directly into the *boolean trap*, and it's precisely because the use of the `Tuple` hides the purpose of the `bool`.

What you're trying to indicate is that a particular `Vector` is "forced" and pairing it with a boolean using a `Tuple` is just smelly. You should either:

* Create a struct to hold your vec3 and bool, for example `SpawnLocation`
* Extend `Vector3` with a subclass which implements (for example) a new interface `IForceable` (or whatever, maybe `IEnforceable` might be better) containing an `isForced()` accessor and use the new subclass `ForcedLocation`

simon816:
I have addressed @Mumfrey's issue by creating a `RespawnLocation` interface

src/main/java/org/spongepowered/api/entity/living/player/RespawnLocation.java
```diff
@@ -0,0 +1,80 @@
...
+import org.spongepowered.api.world.World;
+
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Represents a location for a player to respawn in in a particular world.
+ * Instances can be obtained using {@link RespawnLocationData}.
+ */
+public interface RespawnLocation {
```
kashike: Does this need to be represented as an interface, or can it be class?

simon816: I suppose it could just be a class

PR #1121, opened 2016-03-01
Add API for handedness

ZephireNZ:
SpongePowered/SpongeCommon#630

This PR adds the ability for plugins to query the handedness of entities.

#### Players
Players are able to configure handedness through client settings. This is not configurable serverside, as such the implementation can get a player's handedness but will not be able to set it.

### Entities
Many living entities such as Skeletons or Zombies can have a dominant hand (Ghasts, for example, cannot). This is configurable with the `LeftHanded` NBT tag.

Implementation wise this will likely only be represented by a single boolean, with each value inversely toggling the it.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -543,6 +543,14 @@ public final class Keys {
 
     public static final Key<Value<Vector3d>> LEFT_ARM_ROTATION = KeyFactory.fake("LEFT_ARM_ROTATION");
 
+    /**
+     * Represents the {@link Key} for representing the dominant {@link HandType}
+     * of a {@link Living} entity.
+     *
+     * @see DominantHandData#dominantHand()
+     */
+    public static final Key<Value<HandType>> DOMINANT_HAND = KeyFactory.fake("DOMINANT_HAND");
+
     public static final Key<Value<Vector3d>> LEFT_LEG_ROTATION = KeyFactory.fake("LEFT_LEG_ROTATION");
 
     public static final Key<Value<String>> LOCK_TOKEN = KeyFactory.fake("LOCK_TOKEN");
--- a/src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
@@ -49,6 +49,7 @@ import org.spongepowered.api.data.manipulator.mutable.entity.ChargedData;
 import org.spongepowered.api.data.manipulator.mutable.entity.CriticalHitData;
 import org.spongepowered.api.data.manipulator.mutable.entity.DamageableData;
 import org.spongepowered.api.data.manipulator.mutable.entity.DamagingData;
+import org.spongepowered.api.data.manipulator.mutable.entity.DominantHandData;
 import org.spongepowered.api.data.manipulator.mutable.entity.ElderData;
 import org.spongepowered.api.data.manipulator.mutable.entity.ExpOrbData;
 import org.spongepowered.api.data.manipulator.mutable.entity.ExperienceHolderData;
@@ -93,6 +94,7 @@ import org.spongepowered.api.data.manipulator.mutable.entity.VelocityData;
 import org.spongepowered.api.data.manipulator.mutable.entity.VillagerZombieData;
 import org.spongepowered.api.data.type.Art;
 import org.spongepowered.api.data.type.DyeColor;
+import org.spongepowered.api.data.type.HandType;
 import org.spongepowered.api.data.type.HorseColor;
 import org.spongepowered.api.data.type.HorseStyle;
 import org.spongepowered.api.data.type.HorseVariant;
@@ -234,6 +236,12 @@ public final class CatalogEntityData {
      * {@link Player}s and {@link Living} entities.
      */
     public static final Class<DisplayNameData> DISPLAY_NAME_DATA = DisplayNameData.class;
+    /**
+     * Represents the dominant {@link HandType} used by an entity for for "main"
+     * interactions such as such as tool use or block breaking. Usually
+     * applicable to {@link Living} entities.
+     */
+    public static final Class<DominantHandData> DOMINANT_HAND_DATA = DominantHandData.class;
     /**
      * Signifies that the entity can be dyed a specific {@link DyeColor}.
      * Usually applies to {@link Sheep}.
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDominantHandData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDominantHandData.java
@@ -25,6 +25,7 @@
 package org.spongepowered.api.data.manipulator.immutable.entity;
 
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.manipulator.immutable.ImmutableVariantData;
 import org.spongepowered.api.data.manipulator.mutable.entity.DominantHandData;
 import org.spongepowered.api.data.type.HandType;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
@@ -39,7 +40,7 @@ import org.spongepowered.api.entity.living.Living;
  *     such as tool use or block breaking.
  * </p>
  */
-public interface ImmutableDominantHandData extends ImmutableDataManipulator<ImmutableDominantHandData, DominantHandData> {
+public interface ImmutableDominantHandData extends ImmutableVariantData<HandType, ImmutableDominantHandData, DominantHandData> {
 
     /**
      * Gets the {@link ImmutableValue} representing the dominant {@link HandType} of an
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDominantHandData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDominantHandData.java
@@ -27,16 +27,21 @@ package org.spongepowered.api.data.manipulator.immutable.entity;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.ImmutableVariantData;
 import org.spongepowered.api.data.manipulator.mutable.entity.DominantHandData;
+import org.spongepowered.api.data.property.entity.DominantHandProperty;
 import org.spongepowered.api.data.type.HandType;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.Living;
+import org.spongepowered.api.entity.living.player.Player;
 
 /**
  * A {@link ImmutableDataManipulator} representing the dominant {@link HandType}
  * of a {@link Living} entity.
  * <p>Handedness usually determines which hand is used for "main" interactions,
  * such as tool use or block placing/breaking.</p>
+ *
+ * <p><i>NOTE: </i> This does not apply to {@link Player}s, for Player
+ * entities see {@link DominantHandProperty}.</p>
  */
 public interface ImmutableDominantHandData extends ImmutableVariantData<HandType, ImmutableDominantHandData, DominantHandData> {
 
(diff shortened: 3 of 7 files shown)


src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -708,6 +714,12 @@
 
     public static final Key<Value<Vector3d>> RIGHT_ARM_ROTATION = null;
 
+    /**
+     * Represents the {@link Key} for representing whether something is right
+     * handed.
+     */
+    public static final Key<Value<Boolean>> RIGHT_HANDED = null;
```
windy1: Is it really necessary to have two keys? If an entity is not left handed, wouldn't it be safe to assume it's right handed?

gabizou:
> Players are able to configure handedness through client settings. This is not configurable serverside

Why does this smell then that this should be a `Property` instead of a `DataManipulator`? Plus, I'm sure you can figure out a way to make this an enum instead of two booleans?

ZephireNZ:
Yeah, I was conflicted on that. But then would you be able to get the property from an entity?

src/main/java/org/spongepowered/api/data/key/Keys.java
Mumfrey: Honestly it should be a single pseudo enum `Handedness` with two values.

ZephireNZ: pseudo enum? I imagine it's very unlikely that a 3rd handedness will be invented in the future haha

gabizou:
Yes... that's the whole point of Data API... There are two already.

ZephireNZ:
If I understand the contract correctly, Properties are meant to be immutable, at least within a tick (

If you have the manipulator for the entity, and property for both player and entities, then can't that "immutable" property be changed mid tick by using the manipulator?

src/main/java/org/spongepowered/api/data/key/Keys.java
Mumfrey: In vanilla.

Mumfrey: @ZephireNZ Encoding assumptions is unecessary, the structure in this PR is better than handedness being a boolean. Because if you assume that "not left hand" is equal to "right hand" then you instantly rule out any other hand. Maybe a mod lets you hold stuff in your tail or wear it on your head. It's no up to us to make that decision, we just need to be able to represent the possible states.

gabizou:
You can provide the manipulator for entities that we *can* change, but the docs need to explicitly mention that because by no means are we going to try to fool people into believing that we'll be able to change the dominant hand for clients.

src/main/java/org/spongepowered/api/data/key/Keys.java
ZephireNZ: Ha, good point. Will rewrite it now.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -554,6 +554,14 @@
 
     public static final Key<Value<Vector3d>> LEFT_ARM_ROTATION = null;
 
+    /**
+     * Represents the {@link Key} for representing the dominant {@link HandType}
+     * of a {@link Living} entity.
+     *
+     * @see DominantHandData#dominantHand()
+     */
+    public static final Key<Value<Boolean>> DOMINANT_HAND = null;
```
kashike: How is this a boolean?

gabizou:
How about a property for `DominantHand`?

ZephireNZ:
Working on that now

simon816:
I don't think this is suited as a `DataManipulator`. Maybe it could just be a `Property` or some interface that some entities implement.

gabizou:
@simon816 the thing is, it's something that is data related for entities, except for the case of `Player`s, it fits entirely within the contract of a `DataManipulator` and having it return false for a supports is easy enough.

CrystalMageX:
> Players are able to configure handedness through client settings. This is not configurable serverside, as such the implementation can get a player's handedness but will not be able to set it.

We **can** set this serverside, but the target player just won't see it. This is done by changing the outgoing packets (I don't know which) to *other* players, and faking the main hand of the player.

ryantheleach:
Sounds hacky without contextual data for no good reason..

gabizou:
> We can set this serverside, but the target player just won't see it. This is done by changing the outgoing packets (I don't know which) to other players, and faking the main hand of the player.

Yeah.... no. This is a poor excuse for contextual data that knocks on a very similar door to what ProtocolLib used to do with `ItemStack`s in the past.

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDominantHandData.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.api.data.manipulator.mutable.entity.DominantHandData;
+import org.spongepowered.api.data.type.HandType;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.Living;
+
+/**
+ * A {@link ImmutableDataManipulator} representing the dominant {@link HandType}
+ * of a {@link Living} entity.
+ * <p>
```
modwizcode: These should wrap the paragraph directly and have a blank line between the description and the second paragraph.

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/DominantHandData.java
```diff
@@ -0,0 +1,53 @@
...
+import org.spongepowered.api.data.manipulator.mutable.VariantData;
+import org.spongepowered.api.data.type.HandType;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.Living;
+
+/**
+ * A {@link DataManipulator} representing the dominant {@link HandType} of a
+ * {@link Living} entity.
+ * <p>
```
modwizcode: These should wrap the paragraph directly and have a blank line between the description and the second paragraph.

src/main/java/org/spongepowered/api/data/property/entity/DominantHandProperty.java
```diff
@@ -0,0 +1,66 @@
...
+import org.spongepowered.api.entity.living.Living;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.util.Coerce;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a {@link Property} for the dominant {@link HandType} of a {@link Player}.
+ *
+ * <p>
```
modwizcode: These should wrap the paragraph directly and have a blank line between the description and the second paragraph.

src/main/java/org/spongepowered/api/data/property/entity/DominantHandProperty.java
```diff
@@ -0,0 +1,66 @@
...
+ *     Handedness usually determines which hand is used for "main" interactions,
+ *     such as tool use or block breaking.
+ * </p>
+ *
+ * <p>
```
modwizcode: Same as above

src/main/java/org/spongepowered/api/data/type/HandType.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.spongepowered.api.data.type;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.entity.living.Living;
+import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents the "handedness" of a {@link Living} entity.
```
modwizcode: Maybe elaborate on what "Handedness" actually means in this context?

kashike:
@ZephireNZ There should be a sponge-style enum for the hand types too - `MAIN_HAND` and `OFF_HAND`. When in configurations one could use `main_hand`, and then resolve that to either `left` or `right` on the `Entity` at runtime.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -543,6 +543,14 @@
 
     public static final Key<Value<Vector3d>> LEFT_ARM_ROTATION = KeyFactory.fake("LEFT_ARM_ROTATION");
 
+    /**
+     * Represents the {@link Key} for representing the dominant {@link HandType}
+     * of a {@link Living} entity.
```
gabizou: `Agent`, not `Living`.

src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
```diff
@@ -235,6 +237,12 @@
      */
     public static final Class<DisplayNameData> DISPLAY_NAME_DATA = DisplayNameData.class;
     /**
+     * Represents the dominant {@link HandType} used by an entity for for "main"
+     * interactions such as such as tool use or block breaking. Usually
+     * applicable to {@link Living} entities.
```
gabizou: Same here.

src/main/java/org/spongepowered/api/data/key/Keys.java
kashike: Nope, this is supposed to be `Living`.

src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
kashike: Nope.

src/main/java/org/spongepowered/api/data/key/Keys.java
gabizou: @kashike Not at all. `Player`s can have their handedness changed, which leaves the only other living type to be `Agent`s (which the implementation follows fyi).

kashike: @gabizou `EntityLivingBase` (`Living`) contains the `getPrimaryHand` method - all living entities can be handed, it's just up to the implementations to handle the **setting** of the hand.

gabizou: @kashike I don't want to elude the idea that plugins can set the handedness of objects that clearly will not support it (Players).

kashike: I mean, players support hands, so the javadocs should state so. The javadocs should simply state that you cannot **set** a player's primary hand.

gabizou: @kashike You can't have support for getting and no support for setting, that's not how data manipulators work. It's either they're supported for all `Living` with corner cases that are clearly defined in the javadoc, or, they're supported only by `Agent`s.

Considering that `ArmorStand`s are extending `Living` and not `Agent`s, this should target `Living`, but *clearly* state that you cannot support `Player`s, and instead use the handed property.

ZephireNZ:
@gabizou OK so this should be good now. The implementation targets `EntityLiving`, so players will not return data. I have mentioned in the javadocs that `DominantHandProperty` should be used instead.

kashike:
@ZephireNZ https://github.com/SpongePowered/SpongeAPI/pull/1121#issuecomment-204963653

ZephireNZ:
Can you explain a bit more what you mean?

kashike:
@ZephireNZ Take a look at `EnumHand` (not `EnumHandSide`) in MCP/Forge - you'll see that it has two enum values (`MAIN_HAND` and `OFF_HAND`). This enum is used to represent the main and off hands, since we don't know *which* hand is the active one when we're in configuration/etc.
Take this example:
```java
Hand hand = Sponge.getRegistry().getType(Hand.class, "main_hand"); // assume "main_hand" comes from a configuration value
living.setHeldItem(hand, stack); // simple example, would be data API usually
```

This sets the `Living` entity's held item in the `MAIN_HAND` (could be left or right) to the `stack` provided.

ZephireNZ:
Ah, so you want me to add hand APIs to the relevant entities, and add a
main/offhand overload as well?

kashike:
@ZephireNZ 
- rename `HandType` to `HandSide` (or something similar)
- add `HandType` with `MAIN_HAND` and `OFF_HAND` values

`HandType` should be used for setting item stacks in hand, etc.
`HandSide` should be used for querying/setting (where applicable) the current hand

See MCP/Forge for more details on this.

ZephireNZ:
Why not have an overload methods for both the side and type though?
Relatively simply logic to do to convert side to type, and vice versa

kashike:
Well yeah, allow both, just make sure `MAIN_HAND` and `OFF_HAND` "enum" exists to query/set values/data as well.

PR #1128, opened 2016-03-05
MemorySubjectData Option : NullPointerException

rexbut:
All other methods return an empty map or an immutable copy


zml2008:
Whoops! This is probably correct -- I'll merge it in a bit.

PR #1138, opened 2016-03-14
Add `getExtendedDescription` to the `CommandSpec` class.

MichaelFedora:
Basically a Quality of Life thing, I guess. I found it annoying that the only way to access the `extendedDescription` for a `CommandSpec` was from the `getHelp` method, and as such worked this up.

(Previously, was adding it to the `CommandCallable` interface, but @kashike redirected me to just implement it for the `CommandSpec` class.)

### DIFF
--- a/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
+++ b/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
@@ -61,7 +61,7 @@ public final class CommandSpec implements CommandCallable {
     private final CommandElement args;
     private final CommandExecutor executor;
     private final Optional<Text> description;
-    @Nullable private final Text extendedDescription;
+    private final Optional<Text> extendedDescription;
     @Nullable private final String permission;
     private final InputTokenizer argumentParser;
 
@@ -71,7 +71,7 @@ public final class CommandSpec implements CommandCallable {
         this.executor = executor;
         this.permission = permission;
         this.description = Optional.ofNullable(description);
-        this.extendedDescription = extendedDescription;
+        this.extendedDescription = Optional.ofNullable(extendedDescription);
         this.argumentParser = parser;
     }
 
@@ -354,6 +354,15 @@ public final class CommandSpec implements CommandCallable {
     }
 
     /**
+     * Gets the extended description used with this command if any is present
+     *
+     * @return the extended description.
+     */
+    public Optional<Text> getExtendedDescription(CommandSource source) {
+        return this.extendedDescription;
+    }
+
+    /**
      * Gets the usage for this command appropriate for the provided command
      * source.
      *
@@ -378,14 +387,9 @@ public final class CommandSpec implements CommandCallable {
     public Optional<Text> getHelp(CommandSource source) {
         checkNotNull(source, "source");
         Text.Builder builder = Text.builder();
-        Optional<Text> desc = getShortDescription(source);
-        if (desc.isPresent()) {
-            builder.append(desc.get(), Text.NEW_LINE);
-        }
+        this.getShortDescription(source).ifPresent((a) -> builder.append(a, Text.NEW_LINE));
         builder.append(getUsage(source));
-        if (this.extendedDescription != null) {
-            builder.append(Text.NEW_LINE, this.extendedDescription);
-        }
+        this.getExtendedDescription(source).ifPresent((a) -> builder.append(Text.NEW_LINE, a));
         return Optional.of(builder.build());
     }
 


20zinnm:
Is this breaking? In SemVer, what would it bump?

MichaelFedora:
No, it's more of a feature request. Should it have been an 'issue' instead? Sorry.

20zinnm:
It depends on the functionality. Seems like a small change for a PR but it’s not without merit. Wait for a team member to judge it, I suppose.

kashike:
`getHelp` in `CommandCallable` is what this is already.
`CommandSpec` implements that method and provides the `extendedDescription` method to specify output. A more proper name for `extendedDescription` is `help` in my opinion.

20zinnm:
This sounds like a purely cosmetic change then. Isn’t this the sort of thing that is supposed to be included in a bigger PR? Also, isn’t renaming a function breaking? Would this then be for 5.0.0?

MichaelFedora:
This isn't renaming anything, but simply exposing the `extendedDescription` variable to the public via an optional-get function. `getHelp` is something a little different, I feel.

kashike:
`CommandSpec.Builder#extendedDescription`: Sets an extended description to use in longer help listings for this command.
`CommandCallable#getHelp`: Get a longer formatted help message about this command.

If anything, you should be adding this method directly to `CommandSpec`, as this is `CommandCallable#getHelp` normally.

MichaelFedora:
I suppose.. it seems to be halfway between excess and useful, haha.
I'll change it then, and implement `extendedDescription` as an Optional (as `description` already is).

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -383,8 +392,9 @@ public Text getUsage(CommandSource source) {
             builder.append(desc.get(), Text.NEW_LINE);
         }
         builder.append(getUsage(source));
-        if (this.extendedDescription != null) {
-            builder.append(Text.NEW_LINE, this.extendedDescription);
+        Optional<Text> extdesc = getExtendedDescription(source);
+        if (extdesc.isPresent()) {
+            builder.append(Text.NEW_LINE, extdesc.get());
```
kashike: `this.getExtendedDescription(source).ifPresent(builder::append);` or so, instead of these 3 lines

MichaelFedora: butbutbut that's what line 390-3 do too!
I'll clean it up though~

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -354,6 +354,15 @@ public boolean testPermission(CommandSource source) {
     }
 
     /**
+     * Gets the extended description used with this command if any is present
+     *
+     * @return the extended description.
+     */
+    public Optional<Text> getExtendedDescription(CommandSource source) {
```
MichaelFedora: Technically, this doesn't even need to require the `source` param, though keeping it for consistencies sake might be a good idea. Thoughts?

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
kashike: Yes, still okay.

MichaelFedora:
Squashed commits into one.

liach:
Um... I think you should leave the private field `@Nullable` and for the getter method:
```java
public Optional<Text> getExtendedDescription(CommandSource source) {
    return Optional.ofNullable(extendedDescription);
}
```

MichaelFedora:
I changed it to be "act" the *exact* same as the `description` field.

If we want to change the `extendedDescription` to a nullable, not-`Optional` field, then should we do the same with `description` as well?

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
liach: Why not remove this useless param?

liach:
Actually it better to make less change, so the best solution is to just add a method.

liach:
And that commandsender is used in usage phrase but nowhere else. Remove it.

ST-DDT:
You may need the sender arg to implement multi language support.

MichaelFedora:
I kept the `CommandSource` arg to keep it in the same style as the other functions, as if it was required by the `CommandCallable` interface.

If it is deemed excess, then I could remove it..

@ST-DDT What do  you mean?

ST-DDT:
Nevermind. I thought this was added to the CommandCallable interface. 
Is there a reason to not do it?

MichaelFedora:
Not to add to the `CommandCallable` interface? See @kashike 's [post](https://github.com/SpongePowered/SpongeAPI/pull/1138#issuecomment-196115832). It lands halfway between "excess" and "necessary", because some `CommandCallable` implementations may not need an extended description (because they use the `getHelp` method).

(See: [`SimpleDispatcher`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/command/dispatcher/SimpleDispatcher.java), [`MinecraftCommandWrapper`](https://github.com/SpongePowered/SpongeCommon/blob/master/src/main/java/org/spongepowered/common/command/MinecraftCommandWrapper.java), etc.)

kashike:
The tl;dr is that `getHelp` is what this is in `CommandCallable`.

MichaelFedora:
As made known to me by @kashike , this does somewhat come into play with #1088 , as by [this post](https://github.com/SpongePowered/SpongeAPI/pull/1088#issuecomment-192623913) (or something, and implementing `getExtendedDescription` into the CommandCallable?).

Suddenly things get really muddy. Unless there is multiple people who want it changed to be more "streamlined"/without the param, then I shall leave it be.

MichaelFedora:
Anything I need to do to get this moved, or should this be closed/merged with a "small improvements" pr?

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
ryantheleach: I thought this was needed for translations?

kashike: The parameter is not useless.

MichaelFedora:
Again, just re-iterating why I've wanted this; I want to be able to get the extended description of a command *without* having to get the whole;
```
[usage]
Short description
Long description
```
thing that usually comes along with it using `getHelp()`. I *just* want the extended description. While it could've been added to the entire `CommandCallable` interface, I just put it on the CommandSpec class after some discussion because that's the only class (afaik) that uses an extendedDescription anyways.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
Minecrell: `CommandSpec` is a final class that will always return the field, so the parameter is really not necessary here.

ST-DDT: Well we could change `extendedDescription` to a `Function<CommandSource , Text>` to show some target specific help.
This would allow the help to show user specific limits or hide components the user does not have the permission to use.

MichaelFedora: That would be cool ST, but sounds pretty intense refactor-wise; this PR was simply to be able to get access to the `extendedDescription` value of commands, not introduce this whole abstract layer of permission handling and stuff 😛 .

Deamon5550:

MichaelFedora:
Done, I think?

Also, I can't re-open. Dunno why.

parlough:
This seems like a fairly harmless PR that may help some use cases until commands are refactored/redone so looks good to me.

kashike:
Thank you, @MichaelFedora.

MichaelFedora:
And that ends one of the longest PRs on here, haha. Thanks @kashike !

PR #1144, opened 2016-03-16
Virtual block changes

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1144) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/591)

This adds the ability to send block changes to `Viewer`s without actually having the block change in the world.

Resolves https://github.com/SpongePowered/SpongeAPI/issues/619

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/Viewer.java
+++ b/src/main/java/org/spongepowered/api/effect/Viewer.java
@@ -24,11 +24,16 @@
  */
 package org.spongepowered.api.effect;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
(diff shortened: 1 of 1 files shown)


JBYoshi:
Cool :smile: I started to make a plugin that did something like Orebfuscator, and this will probably be the foundation of it (although I'll need to make another PR to add a `ChunkSendEvent`).

ryantheleach:
What about supplying an area that needs to be reset, rather then having to loop and create a vector3i for every block that needs to be reset?

kashike:
@ryantheleach There are `int, int` methods here as well.

bloodmc:
Looks good to me.

theboomer:
I'm pumped to use this, and can only imagine how JBYoshi might be able to make that orebfuscator-like system 100x more efficient with this :)

PR #1146, opened 2016-03-18
[WIP] Add new 1.9 block entities

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1146) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/619)

This is a work-in-progress PR to add 1.9 block entities and related data.


liach:
default method for accessing the data?

src/main/java/org/spongepowered/api/data/manipulator/immutable/tileentity/ImmutableEndGatewayData.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public interface ImmutableEndGatewayData extends ImmutableDataManipulator<ImmutableEndGatewayData, EndGatewayData> {
+
+    /**
+     * Gets the {@link ImmutableValue} for the exit portal location of
+     * the {@link EndGateway}.
+     *
+     * @return The value for the exit portal location
+     */
+    ImmutableValue<Vector3i> exitPortal();
```
liach: Is this the position where player goes after entering the portal? If yes, then I do not suggest to use `Vector3i` since in this case, the player will be transported to another world and `Location<World>` would become the best choice.

kashike:
@liach You're confusing an end **gateway** with an end **portal**. Gateways are to the same dimension, portal is to other dimension.

liach:
Yes so I deleted that comment. By the way, is it necessary to include the age in the date of the end gateway?

kashike:
This isn't complete, I was just pushing my changes after the rebase :)

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -300,8 +303,24 @@
 
     public static final Key<Value<Boolean>> ELDER_GUARDIAN = KeyFactory.fake("ELDER_GUARDIAN");
 
+    /**
+     * Represents the {@link Key} for representing if the exact teleport location
```
simon816: 'represent' used twice in one sentence reads a bit oddly. Don't need the first one. could just read "Key for representing..."

src/main/java/org/spongepowered/api/data/manipulator/mutable/tileentity/EndGatewayData.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Gets the {@link Value} for the exit portal location of
+     * the {@link EndGateway}.
+     *
+     * @return The value for the exit portal location
+     */
+    Value<Vector3i> exitPortal();
+
+    /**
+     * Gets the {@link Value} for the "should use exact teleport location"
```
simon816: Could explain what this means (I'm guessing this is about finding a safe location but I don't know)

src/main/java/org/spongepowered/api/block/tileentity/EndGateway.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Represents an End Gateway.
+ *
+ * @see EndGatewayData
+ */
+public interface EndGateway extends TileEntity {
+
+    default EndGatewayData getEndGatewayData() {
+        return this.get(EndGatewayData.class).get();
```
liach: Only local **field** access needs `this` qualifier. This is a method.

src/main/java/org/spongepowered/api/block/tileentity/Structure.java
```diff
@@ -0,0 +1,32 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.block.tileentity;
+
+/**
+ * Represents a Structure.
+ */
+public interface Structure extends TileEntity {
```
liach: This name is confusing... Call it `Structure Block` in javadoc?

src/main/java/org/spongepowered/api/data/manipulator/immutable/tileentity/ImmutableEndGatewayData.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Gets the {@link ImmutableValue} for the "should use exact teleport
+     * location" tate of the {@link EndGateway}.
```
ZephireNZ: tate?

kinggoesgaming: Mate I rate that 8 out of 8 :P

kashike:
Superseded by #1146

PR #1150, opened 2016-03-19
Fixing ASM version due to https://github.com/SpongePowered/SpongeVanilla/issues/233

vorburger:


kashike:
`asm-debug-all` version `5.0.3` is a [Forge dependency](https://github.com/MinecraftForge/MinecraftForge/blob/master/jsons/1.8.9-dev.json#L22-L26).

Minecrell:
This needs to be fixed in the Eclipse setup of SpongeVanilla, not here.

PR #1155, opened 2016-03-21
Add methods to create TextFormat

kashike:
`TextFormat` is missing methods to create it.


windy1:
Hmm well you can actually do `TextFormat.NONE.color(color).style(style);` right now but these seem like a good convenience methods if nothing else. Assigning @Minecrell since `TextFormats` are kind of his deal IIRC.

MichaelFedora:
If nothing else, you could get rid of a lot of the constructors to just have one, and then make the `of` functions supply the `NONE`'s.

Minecrell:
Yep, remove the additional constructors, these were from when they were still public.

kashike:
@Minecrell done

MichaelFedora:
@kashike no your not :P

Issue #1105, opened 2016-02-25
There should be non-center upright Directions


CrystalMageX:
*SpongeAPI* is an API for *Minecraft*. In *Minecraft* there **26 neighbors** for a block, 6 away from the faces of the cube, 8 away from the vertices, and 12 away from the edges. The `Direction` class should represent this.

~~The secondary ordinal directions should be removed because it is **impossible** for a block to be something like **"NORTH_NORTHEAST" from a block**. They are also unused in SpongeForge, SpongeVanilla, SpongeCommon, and even SpongeAPI itself.~~ Edit: Seems like they are used indeed in item frames and other things.

Sideways up and down directions (vertex normal directions) should be added as a way to **describe blocks** next to another block that **isn't directly up or down**. It also allows creative voxel manipulation plugins (like *WorldEdit*) to use `GenericArguments.enumValue(Text.of("direction"), Direction.class)`. Edit: Also, secondary ordinal up/down directions should also be added for consistency that directions pointing up/down be as accurate as directions on a compass.

simon816:
They _are_ used. For things like rotation - items in an item frame, skull tileentity head rotation etc

gravityfox:
But I have to ask...

Why would you ever need this?

Could you provide a use case?

liach:
Custom plugin that create tools that destroy connected blocks.

kinggoesgaming:
@liach did you mean mods :P

gravityfox:
I still don't see that as a valid use case.

Tell my why SPONGEAPI would have a use for those directions.

CrystalMageX:
@gravityfox A use case would be, as described in the original issue, a worldedit plugin that accepts directions to be specified using chat, like `//expand halfup_northeast 10`, and have it processed easily using `GenericArguments.enumValue(Text.of("direction"), Direction.class)`.

> Tell my why SPONGEAPI would have a use for those directions.

SpongeAPI does not use *anything*. It is an API, it provides stuff for plugins to use.

gravityfox:
Let me give you two different perspectives...

From a user perspective, `halfup_northeast` makes absolutely no sense to me, and I could never be bothered to type that out. i would just type `north`, `east`, and then `up`.

From a plugin dev perspective, If you really wanted to do something like that, you would be using your own command parser/engine, and you wouldn't be using generic arguments. Generic arguments is a general purpose convenient way of making commands. If you want to do something crazy and out there, do it yourself.

[...]

PR #1160, opened 2016-03-28
Add non-center upright Directions

CrystalMageX:
Currently the [`org.spongepowered.api.util.Direction`](https://github.com/SpongePowered/SpongeAPI/blob/751873a73cd4ffa2583361181a8d20879393cc51/src/main/java/org/spongepowered/api/util/Direction.java) class only supports the 4 cardinal directions, the 4 ordinal directions, and the 8 secondary ordinal directions, plus the up&down vertical directions and the "NONE" direction. This adds **half, quarter, and three quarters up&down** directions for **each horizontal** direction, minus "NONE", totaling 96 new directions. This also add a convenient `DirectionVectors` class to get the vectors these directions use without getting the direction and calling `toVector3d()` on it.

This is based on bleeding, because I have re-organized the fields, [breaking](http://www.benf.org/other/cfr/switch-on-enum.html) already-compiled code. Breaking could be avoided, but the code would be *messy and unreadable* with this amount of fields.

**More info** about *why* we should add these upright directions is found in: #1105

### DIFF
--- a/src/main/java/org/spongepowered/api/util/Direction.java
+++ b/src/main/java/org/spongepowered/api/util/Direction.java
@@ -24,13 +24,16 @@
  */
 package org.spongepowered.api.util;
 
-import com.flowpowered.math.GenericMath;
-import com.flowpowered.math.TrigMath;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkArgument;
+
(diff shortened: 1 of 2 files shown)


CrystalMageX:
I am closing this pull request, because it is consensus in #1105 that there currently no use for it. Once a mod/version is made that makes use of these directions, another, improved pull request can be made.

Issue #1142, opened 2016-03-14
Support for serializing sections of the world to disk


Deamon5550:
We would like to add support for serializing sections of the world to disk. This issue is to gather comments on the layout of the data.

This is a mockup of the updated variant of the MCEdit scheamtic format, this variant would not be backwards compatible. The key changes here are the inclusion of a version tag to assist with future content updates and the inclusion of an id palette to both increase compression and allow a much larger range of block states. The format of the block state palette is mostly inspired from the 1.9 chunk data protocol.

The format is mocked out here in json although the preferred serialization format for this data would be NBT.

``` json
"Schematic": {
    "Metadata": {
        "Name": "my_schematic",
        "Author": "Deamon",
        "RequiredMods":[
            "a_mod",
            "another_mod"
        ],
        ...
    },
    "Version": 1,
    "Width": 1,
    "Height": 1,
    "Length": 1,
    "Offset": [0, 0, 0],
    "PaletteLength": 1,
    "Palette": {
        "minecraft:air": 0
    },
    "BlockData":[0],
    "TileEntities":[
        {
            "ContentVersion": 0,
            "x": 0,
            "y": 0,
            "z":0,
            ...
        }
    ],
    "Entities":[
        {
            "ContentVersion": 0,
            "x": 0,
            "y": 0,
            "z":0,
            ...
        }
    ]
}
```

### Usage
Usage of this format within sponge depends on #1137 and would likely consist of a Builder pattern of some sort.

20zinnm:
YES! Finally! I'm actually happy right now :smile:

20zinnm:
As for the Name tag, how about
```json
"Schematic": {
  ...
  "Info": {
    "Name": "The Magical Build",
    "Author": "AniSkywalker",
    "Date": "TIMESTAMP"
    ...
  }
}
```
Where everything in Info (and Info itself) is optional? It would then be loaded as an `Optional<Properties>` in a `Schematic` object. The builder could use a `.property("Name", "The Magical Build")` and we could then add an ENUM containing universally recognized properties.

gabizou:
Some thoughts:
- What about possible version for the entity/tileentity tag information? Since 1.9 takes the precaution of upgrading the tags as it's being read from disk, I'd likely see that schematics for different versions of MC would eventually be supported as such. (this would likely require not a `"SchematicVersion"` but also a `"ContentVersion"` that is shared between Mojang's new content versions and future ones.
[...]

PR #1172, opened 2016-04-04
[WIP] Schematic API

20zinnm:
See #1142 and the actual specification [here](https://github.com/SpongePowered/Schematic-Specification) (https://github.com/SpongePowered/Schematic-Specification). Big thanks to @Deamon5550 for the actual specification and everyone in the SpongeDev IRC. This is my first PR, so I hope it's not _too_ bad. Input is appreciated--this is a fairly important API because the Structures system would likely be based on it. This is a non-breaking API addition (at the moment).

### DIFF
--- a/src/main/java/org/spongepowered/api/schematic/Schematic.java
+++ b/src/main/java/org/spongepowered/api/schematic/Schematic.java
@@ -1,0 +1,185 @@
+package org.spongepowered.api.schematic;
+
+
+import com.flowpowered.math.vector.Vector3i;
+import com.google.common.collect.BiMap;
+import org.spongepowered.api.block.BlockSnapshot;
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.extent.Extent;
+import org.spongepowered.api.world.extent.worker.MutableBlockVolumeWorker;
+
+import java.util.Optional;
+
+/**
+ * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
+ * .com/SpongePowered/Schematic-Specification}.
+ */
+public interface Schematic extends DataSerializable {
+
+    /**
+     * Gets the format version used. Correlates to the `Version` field of the specification.
+     *
+     * @return The format version used
+     */
+    int version();
+
+    /**
+     * Gets the attached {@link DataContainer}. Correlates to the optional `Metadata` field of the specification.
+     *
+     * @return The attached {@link DataContainer} or {@link Optional#empty()} if none
+     */
+    Optional<DataContainer> metadata();
+
+    /**
+     * Gets the width of the Schematic. Correlates to the `Width` field of the specification.
+     *
+     * @return The width of the Schematic.
+     */
+    int width();
+
+    /**
+     * Gets the height of the Schematic. Correlates to the `Height` field of the specification.
+     *
+     * @return The height of the Schematic.
+     */
+    int height();
+
+    /**
+     * Gets the length of the Schematic. Correlates to the `Length` field of the specification.
+     *
+     * @return The length of the Schematic.
+     */
+    int length();
+
+    /**
+     * Gets the origin of the Schematic (relative to the minimum point) or the minimum point if none. Correlates to the `OffsetX`, `OffsetY`, and
+     * `OffsetZ` fields of the specification.
+     *
+     * @return The origin of the Schematic
+     */
+    Vector3i origin();
+
+    /**
+     * Gets the largest key in the palette.
+     * @return
+     */
+    int paletteMax();
+
+    // TODO how to say the position in the set == the palette value
+
+    /**
+     * Gets the palette of the Schematic. Correlates backwards to the `Palette` field of the specification.
+     *
+     * @return The palette
+     */
+    BiMap<Integer, BlockState> palette();
+
+    /**
+     * Gets the {@link BlockState} at the relative location, or {@link Optional#empty()} if none.
+     *
+     * @param location The location (relative to the origin of the schematic) of the BlockSnapshot desired
+     * @return The BlockState found
+     */
+    Optional<BlockState> blockAt(Vector3i location);
+
+    /**
+     * Gets all {@link BlockState}s in the Schematic.
+     *
+     * @return All BlockStates
+     */
+    BlockSnapshot[] blocks();
+
+    /**
+     * Gets all {@link SchematicEntity} in the Schematic.
+     *
+     * @return All entities
+     */
+    SchematicEntity[] entities();
+
+    /**
+     * Gets all TileEntities in the Schematic.
+     *
+     * @return All tile entities
+     */
+    SchematicTileEntity[] tileEntities();
+
+    
+
+    interface Builder extends DataBuilder<Schematic> {
+
+
+        /**
+         * Adds the given {@link DataContainer} to the schematic's metadata.
+         *
+         * @param metadata The metadata to add
+         * @return This schematic builder
+         */
+        Builder metadata(DataContainer metadata);
+
+        /**
+         * Sets the minimum point of the schematic.
+         *
+         * @param point The point to set
+         * @return This schematic builder
+         */
+        Builder minimumPoint(Vector3i point);
+
+        /**
+         * Sets the maximum point of the schematic.
+         *
+         * @param point The point to set
+         * @return This schematic builder
+         */
+        Builder maximumPoint(Vector3i point);
+
+        /**
+         * Sets the origin of the schematic. Correlates to the `OffsetX`, `OffsetY`, and `OffsetZ` fields of the Schematic Specification (v1.
+         *
+         * @param origin The origin point to set
+         * @return This schematic builder
+         */
+        Builder origin(Vector3i origin);
+
+        /**
+         * Adds a {@link BlockState} to the palette. Duplicate BlockStates will be ignored.
+         *
+         * @param blockState
+         * @return This schematic builder
+         */
+        Builder addToPalette(BlockState blockState);
+
+        /**
+         * Should offer automated conversions between versions.
+         *
+         * @param holder
+         * @return This schematic builder
+         */
+        Builder from(Schematic holder);
+
+        /**
+         * Builds the schematic object
+         *
+         * @return The built Schematic object
+         * @throws InvalidDataException if any required fields are not completed
+         */
+        Optional<Schematic> build() throws InvalidDataException;
+
+        /**
+         * Builds the schematic object using locations from the given world
+         *
+         * @param world
+         * @return
+         * @throws InvalidDataException
+         */
+        Optional<Schematic> build(World world) throws InvalidDataException;
+
+    }
+
+}
--- a/src/main/java/org/spongepowered/api/schematic/Schematic.java
+++ b/src/main/java/org/spongepowered/api/schematic/Schematic.java
@@ -26,63 +26,71 @@ package org.spongepowered.api.schematic;
 
 import com.flowpowered.math.vector.Vector3i;
 import com.google.common.collect.BiMap;
-import org.spongepowered.api.block.BlockSnapshot;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.data.persistence.InvalidDataException;
 import org.spongepowered.api.world.World;
-import org.spongepowered.api.world.extent.Extent;
 
 import java.util.Optional;
 
 /**
- * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
+ * Unless otherwise noted, "the specification" refers to the latest version of
+ * the Schematic Specification {@url https://github
  * .com/SpongePowered/Schematic-Specification}.
  */
 public interface Schematic extends DataSerializable {
 
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
     /**
      * Gets the attached {@link DataContainer}.
      *
-     * <p>Correlates with the optional `Metadata` field of the specification.</p>
+     * <p>Correlates to the optional `Metadata` field of the specification.</p>
      *
-     * @return The attached {@link DataContainer} or {@link Optional#empty()} if none
+     * @return The attached {@link DataContainer} or {@link Optional#empty()} if
+     * none
      */
     Optional<DataContainer> getMetadata();
 
     /**
      * Gets the width of the Schematic.
      *
-     * <p>Correlates with the `Width` field of the specification.</p>
+     * <p>Correlates to the `Width` field of the specification.</p>
      *
-     * @return The width of the Schematic.
+     * @return The width of the Schematic
      */
     int getWidth();
 
     /**
      * Gets the height of the Schematic.
      *
-     * <p>Correlates with the `Height` field of the specification.</p>
+     * <p>Correlates to the `Height` field of the specification.</p>
      *
-     * @return The height of the Schematic.
+     * @return The height of the Schematic
      */
     int getHeight();
 
     /**
      * Gets the length of the Schematic.
      *
-     * <p>Correlates with the `Length` field of the specification.</p>
+     * <p>Correlates to the `Length` field of the specification.</p>
      *
-     * @return The length of the Schematic.
+     * @return The length of the Schematic
      */
     int getLength();
 
     /**
-     * Gets the origin of the Schematic (relative to the minimum point) or the minimum point if none.
+     * Gets the origin of the Schematic (relative to the minimum point) or the
+     * minimum point if none.
      *
-     * <p>Correlates with the `OffsetX`, `OffsetY`, and `OffsetZ` fields of the specification.</p>
+     * <p>Correlates to the `OffsetX`, `OffsetY`, and `OffsetZ` fields of the
+     * specification.</p>
      *
      * @return The origin of the Schematic
      */
@@ -105,13 +113,14 @@ public interface Schematic extends DataSerializable {
     BiMap<Integer, BlockState> getPalette();
 
     /**
-     * Gets the {@link BlockState} at the relative location, or {@link Optional#empty()} if none.
+     * Gets the {@link BlockState} at the relative location, or
+     * {@link Optional#empty()} if none.
      *
-     * @param extent The extent
-     * @param location The location (relative to the origin of the schematic) of the BlockSnapshot desired
+     * @param location The location (relative to the origin of the schematic)
+     * of the BlockSnapshot desired
      * @return The BlockState found
      */
-    Optional<BlockState> getBlockAt(Extent extent, Vector3i location);
+    Optional<BlockState> getBlockAt(Vector3i location);
 
     /**
      * Gets all {@link BlockState}s in the Schematic.
@@ -165,7 +174,8 @@ public interface Schematic extends DataSerializable {
         /**
          * Sets the origin of the schematic.
          *
-         * <p>Correlates to the `OffsetX`, `OffsetY`, and `OffsetZ` fields of the Schematic Specification (v1).</p>
+         * <p>Correlates to the `OffsetX`, `OffsetY`, and `OffsetZ` fields of
+         * the Schematic Specification (v1).</p>
          *
          * @param origin The origin point to set
          * @return This schematic builder
@@ -173,7 +183,8 @@ public interface Schematic extends DataSerializable {
         Builder origin(Vector3i origin);
 
         /**
-         * Adds a {@link BlockState} to the palette. Duplicate BlockStates will be ignored.
+         * Adds a {@link BlockState} to the palette. Duplicate BlockStates will
+         * be ignored.
          *
          * @param blockState
          * @return This schematic builder
@@ -192,9 +203,10 @@ public interface Schematic extends DataSerializable {
          * Builds the schematic object
          *
          * @return The built Schematic object
-         * @throws InvalidDataException if any required fields are not completed
+         * @throws InvalidDataException if any required fields are not
+         * completed
          */
-        Optional<Schematic> build() throws InvalidDataException;
+        Schematic build() throws InvalidDataException;
 
         /**
          * Builds the schematic object using locations from the given world
@@ -203,7 +215,7 @@ public interface Schematic extends DataSerializable {
          * @return
          * @throws InvalidDataException
          */
-        Optional<Schematic> build(World world) throws InvalidDataException;
+        Schematic build(World world) throws InvalidDataException;
 
     }
 
--- a/src/main/java/org/spongepowered/api/schematic/Schematic.java
+++ b/src/main/java/org/spongepowered/api/schematic/Schematic.java
@@ -33,6 +33,7 @@ import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.data.persistence.InvalidDataException;
 import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.extent.Extent;
 
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,205 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.schematic;
+
+
```
kashike: One too many new lines here.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,205 @@
...
+ * .com/SpongePowered/Schematic-Specification}.
+ */
+public interface Schematic extends DataSerializable {
+
+    /**
+     * Gets the format version used. Correlates to the `Version` field of the specification.
+     *
+     * @return The format version used
+     */
+    int version();
```
kashike: Getters should be prefixed with `get`.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.Optional;
+
+/**
+ * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
+ * .com/SpongePowered/Schematic-Specification}.
+ */
+public interface Schematic extends DataSerializable {
+
+    /**
+     * Gets the format version used. Correlates to the `Version` field of the specification.
```
kashike: Should be formatted like (for all cases of this):
```
* Gets the format version used.
*
* <p>Correlates to the `Version` field of the specification.</p>
```

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,202 @@
...
+     */
+    int getLength();
+
+    /**
+     * Gets the origin of the Schematic (relative to the minimum point) or the minimum point if none. Correlates to the `OffsetX`, `OffsetY`, and
+     * `OffsetZ` fields of the specification.
+     *
+     * @return The origin of the Schematic
+     */
+    Vector3i getOrigin();
```
kashike: This is called "Origin", but correlated to the "Offset" fields?

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,202 @@
...
+     */
+    BiMap<Integer, BlockState> getPalette();
+
+    /**
+     * Gets the {@link BlockState} at the relative location, or {@link Optional#empty()} if none.
+     *
+     * @param location The location (relative to the origin of the schematic) of the BlockSnapshot desired
+     * @return The BlockState found
+     */
+    Optional<BlockState> getBlockAt(Vector3i location);
```
kashike: You require passing an `Extent` as well, to obtain the `BlockState` from.

src/main/java/org/spongepowered/api/schematic/Schematic.java
gabizou: Shouldn't need this as it's inherited from `DataSerializable`.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Gets all {@link BlockState}s in the Schematic.
+     *
+     * @return All BlockStates
+     */
+    BlockSnapshot[] getBlocks();
```
gabizou: Don't use BlockSnapshot.

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.spongepowered.api.schematic;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.entity.EntityType;
+
+public interface SchematicEntity extends DataContainer, DataSerializable {
```
gabizou: Uh, no.... You don't want to extend `DataContainer`.

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.spongepowered.api.schematic;
+
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.block.tileentity.TileEntityType;
+import org.spongepowered.api.data.DataContainer;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+
+public interface SchematicTileEntity extends DataContainer, DataSerializable {
```
gabizou: And here.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: The Offset fields define an origin for recreating the build. Might as well be a Vector3i, because it'd just be three int fields.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+import org.spongepowered.api.world.extent.Extent;
+
+import java.util.Optional;
+
+/**
+ * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
+ * .com/SpongePowered/Schematic-Specification}.
+ */
+public interface Schematic extends DataSerializable {
+
```
gabizou: Should add a static method to get a `Builder` instance, just like `ItemStack`.

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
```diff
@@ -0,0 +1,67 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.schematic;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.entity.EntityType;
+
+public interface SchematicEntity extends DataSerializable {
```
gabizou: You probably want to be able to extend `DataHolder`

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
```diff
@@ -0,0 +1,67 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.schematic;
+
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.block.tileentity.TileEntityType;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+
+public interface SchematicTileEntity extends DataSerializable {
```
gabizou: Likewise here, extend `DataHolder` to interact with the Data API (otherwise, it's all raw data container manipulation).

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
```diff
@@ -0,0 +1,67 @@
...
+        /**
+         * Gets the location of the entity relative to the origin.
+         *
+         * @return The location
+         */
+        Builder location(Vector3d location);
+
+    }
+
+}
```
windy1: Newline here

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
```diff
@@ -0,0 +1,67 @@
...
+        /**
+         * Gets the location of the entity relative to the origin.
+         *
+         * @return The location
+         */
+        Builder location(Vector3i location);
+
+    }
+
+}
```
windy1: Same

kashike: Newline here

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
```diff
@@ -0,0 +1,67 @@
...
+         * @return This SchematicEntity builder
+         */
+        Builder type(TileEntityType type);
+
+        /**
+         * Gets the location of the entity relative to the origin.
+         *
+         * @return The location
+         */
+        Builder location(Vector3i location);
```
kashike: This is more of "position", no?

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
```diff
@@ -0,0 +1,67 @@
...
+     * Gets the location of the entity relative to the origin.
+     *
+     * @return The location
+     */
+    Vector3i getLocation();
+
+    interface Builder extends DataBuilder<SchematicEntity> {
+
+        /**
+         * Sets the EntityType for the SchematicEntity
```
kashike: `TileEntityType`, not `EntityType`.

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
```diff
@@ -0,0 +1,67 @@
...
+         * @return This SchematicEntity builder
+         */
+        Builder type(EntityType type);
+
+        /**
+         * Gets the location of the entity relative to the origin.
+         *
+         * @return The location
+         */
+        Builder location(Vector3d location);
```
kashike: This is more of "position", no?

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+        Optional<Schematic> build() throws InvalidDataException;
+
+        /**
+         * Builds the schematic object using locations from the given world
+         *
+         * @param world
+         * @return
+         * @throws InvalidDataException
+         */
+        Optional<Schematic> build(World world) throws InvalidDataException;
```
kashike: In what cases can this return `null` (you're using `Optional`)?

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+         */
+        Builder from(Schematic holder);
+
+        /**
+         * Builds the schematic object
+         *
+         * @return The built Schematic object
+         * @throws InvalidDataException if any required fields are not completed
+         */
+        Optional<Schematic> build() throws InvalidDataException;
```
kashike: In what cases can this return `null` (you're using `Optional`)?

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+    BiMap<Integer, BlockState> getPalette();
+
+    /**
+     * Gets the {@link BlockState} at the relative location, or {@link Optional#empty()} if none.
+     *
+     * @param extent The extent
+     * @param location The location (relative to the origin of the schematic) of the BlockSnapshot desired
+     * @return The BlockState found
+     */
+    Optional<BlockState> getBlockAt(Extent extent, Vector3i location);
```
kashike: `position`, not `location`?

gabizou:
You should review the javadocs as there's a few issues of copy pasting and/or line length. Javadocs should be limited to 80 character widths.

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
20zinnm: I suppose.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+     * @return The origin of the Schematic
+     */
+    Vector3i getOrigin();
+
+    /**
+     * Gets the largest key in the palette.
+     *
+     * @return The largest key
+     */
+    int getPaletteMax();
```
kashike: `getMaxPalette`?

src/main/java/org/spongepowered/api/schematic/Schematic.java
liach: If you return `Optional.empty()` you'd just throw an exception.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
+ * .com/SpongePowered/Schematic-Specification}.
+ */
+public interface Schematic extends DataSerializable {
+
+    /**
+     * Gets the attached {@link DataContainer}.
+     *
+     * <p>Correlates with the optional `Metadata` field of the specification.</p>
```
Aaron1011: 'Corresponds to' would make more sense than 'Correlates with', I think.

src/main/java/org/spongepowered/api/schematic/Schematic.java
Aaron1011: I know that this is in the specification, but it's more of an implementation detail. `getPalette#size` already does this.

src/main/java/org/spongepowered/api/schematic/Schematic.java
Aaron1011: Why is there an `Extent` parameter?

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+     * @return The BlockState found
+     */
+    Optional<BlockState> getBlockAt(Extent extent, Vector3i location);
+
+    /**
+     * Gets all {@link BlockState}s in the Schematic.
+     *
+     * @return All BlockStates
+     */
+    BlockState[] getBlocks();
```
Deamon5550: Arrays are bad.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: @kashike told me to add it.

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
Aaron1011: I think an `EntitySnapshot` should be used instead of a new interface, as a schematic entity can store additional data (NBT) as well.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: I was using the basic spec. I'll change it, you're correct--it should only return empty if there's an error.

src/main/java/org/spongepowered/api/schematic/Schematic.java
Zidane: the build method of builders should never return Optional so kashike is right. This will be fixed elsewhere in the API.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: Wat?

src/main/java/org/spongepowered/api/schematic/Schematic.java
Aaron1011: But, this method is just getting a `BlockState` stored in the schematic. What does the `extent` parameter even mean?

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
20zinnm: @Aaron1011 EntitySnapshot is Locatable.

src/main/java/org/spongepowered/api/schematic/Schematic.java
Zidane: @kashike 

Why in the world does this take an Extent? The Schematic stores its blockstates....

src/main/java/org/spongepowered/api/schematic/Schematic.java
Zidane: @Deamon5550 

What would you suggest?

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: I'll remove it--I thought kashike had some god-reason.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: @Deamon5550 Once the BlockData is deserialized, it's just an array...

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: It doesn’t. The palette could have spots, so size isn’t the biggest key. You have to take the Set of keys and sort them.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: You are correct. http://english.stackexchange.com/questions/142495/correspond-to-vs-correspond-with

20zinnm:
So, there's a few small(ish) changes to be made. Is the overall API logical?

src/main/java/org/spongepowered/api/schematic/Schematic.java
Deamon5550: Well either the schematic is immutable in which case exposing arrays breaks that and allows you to mutate it willy nilly or its mutable in which case your enforcing the contract on all implementations that it is stored in memory as an array of blockstates (which in all honestly it probably wouldn't be as its cheaper to just keep it encoded to a byte/short/int array) better to just have getters and setters.

20zinnm: How about using an immutable list?

src/main/java/org/spongepowered/api/schematic/Schematic.java
kashike: A `getExtent` method should be added to `Schematic`, then.

20zinnm: What’s the extent for?

kashike: ...to get the `Extent` that the `BlockState`s are coming from?

Zidane: @kashike 

The blockstates aren't coming from an Extent...they are what was stored in a file. The application of the schematic onto an Extent is another story. Schematics are simply NOT Extent-bound.

src/main/java/org/spongepowered/api/schematic/Schematic.java
Deamon5550: Seriously getters and setters, the api should not be dictating how the implementation stores the data.

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.extent.Extent;
+
+import java.util.Optional;
+
+/**
+ * Unless otherwise noted, "the specification" refers to the latest version of the Schematic Specification {@url https://github
```
simon816: Maximum line length for javadoc is 80

src/main/java/org/spongepowered/api/schematic/Schematic.java
```diff
@@ -0,0 +1,210 @@
...
+     * @return The attached {@link DataContainer} or {@link Optional#empty()} if none
+     */
+    Optional<DataContainer> getMetadata();
+
+    /**
+     * Gets the width of the Schematic.
+     *
+     * <p>Correlates with the `Width` field of the specification.</p>
+     *
+     * @return The width of the Schematic.
```
simon816: Shouldn't put periods at the end of `@return` descriptions

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.spongepowered.api.schematic;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.data.DataHolder;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.entity.EntityType;
+
+public interface SchematicEntity extends DataHolder, DataSerializable {
```
Aaron1011: Shouldn't an `EntitySnapshot` be used instead of this? The schematic spec allows custom data (i.e. NBT) to be attached to entities.

20zinnm: @Aaron1011 The reason this is a DataHolder is because it supports custom data. EntitySnapshot requires a location and a unique ID (entityID?) so it can't work in the Schematic. We determined creating another representation was needed.

simon816: IIRC EntitySnapshot was designed with the schematic spec in mind. If it won't work then maybe we could make a common shared interface. (perhaps create a `LocatableEntitySnapshot` if location is the only issue)

20zinnm: Location and, IIRC, the Unique ID. For it to be usable in this case, change the restore functionality to take a Location<World> and remove the restore and UUID from the state.

gabizou: What `SchematicEntity` needs is a simple Vector3d relative to the offset perhaps, and of course rotations.

20zinnm: Rotations is (according to Deamon) stored in the custom data. The Vector3d is already relative to the origin (which is basically the offset + lower bound).

gabizou: And in my opinion, isn't an Entity's data such as rotation and location going to belong in `SchematicEntity`?

20zinnm: I suppose so. Could we modify EntitySnapshot and BlockSnapshot to suit our purposes?

gabizou: No, they are for different use cases.

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
20zinnm: What does "newline" mean? Omit the extra line?

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
20zinnm: I thought it was location.

src/main/java/org/spongepowered/api/schematic/Schematic.java
20zinnm: @Deamon5550 I don't understand what you mean by this. Can you explain more?

src/main/java/org/spongepowered/api/schematic/SchematicEntity.java
20zinnm: So what are we doing with this? Should I add the rotation to it?

src/main/java/org/spongepowered/api/schematic/SchematicTileEntity.java
kashike: You need a newline at the end of the file.

20zinnm:
Sorry, just got back from a bit of a break. What else needs to be done?

kinggoesgaming:
@20zinnm Name your next commit More changes ++ :P

src/main/java/org/spongepowered/api/schematic/Schematic.java
simon816: Should be a `Collection` or perhaps `ImmutableList` if this should be exposed.
What I think @Deamon5550 is saying is that by providing access to an indexed list (array) you're dictating that the implementation must store data in some defined order. The API consumer must also agree on this order so that it's actually useful defining such an order.

All that's really needed is things like `getBlock(x, y, z)`, perhaps add utility methods to get in a range(min, max) etc.

For `getEntities()` and `getTileEntities()` arrays should also not be given, though `List`/`ImmutableList` would be acceptable for these

RobertHerhold:
Status update?

gabizou:
Closing as @Deamon5550 is close to finishing a branch merge in `feature/schematics`

PR #1176, opened 2016-04-10
[WIP]1.9 Multi-World API Changes

Zidane:


src/main/java/org/spongepowered/api/Server.java
```diff
@@ -220,7 +214,7 @@
      * @param settings The settings for creation
      * @return The new or existing world properties, if creation was successful
```
simon816: Remove "if creation was successful"

PR #1182, opened 2016-04-14
[WIP] Add ClientConnectionEvent.Handshake to let plugins handle client handshaking

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1182) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/649) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/609)

This adds a new `ClientConnectionEvent.Handshake` event which is fired during a client `HANDSHAKE`. This event allows for more versatile handshaking to occur - the Sponge implementations are limited to BungeeCord-specific handshaking currently.

To replicate the BungeeCord-style handshaking:
```java
    @Listener
    public void handshake(final ClientConnectionEvent.Handshake event) {
        String[] split = event.getHandshakeString().split("\00");
        if (split.length == 3 || split.length == 4) {
            event.setServerHostname(split[0]);
            event.setSocketAddressHostname(split[1]);

            GameProfile profile = GameProfile.of(UUID.fromString(split[2].replaceFirst("(\\w{8})(\\w{4})(\\w{4})(\\w{4})(\\w{12})", "$1-$2-$3-$4-$5")), null);

            if (split.length == 4) {
                ProfileProperty[] properties = GSON.fromJson(split[3], ProfileProperty[].class);
                for (ProfileProperty property : properties) {
                    profile.getPropertyMap().put(property.getName(), property);
                }
            }

            event.setGameProfile(profile);
        } else {
            event.setFailed(true);
        }
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
+++ b/src/main/java/org/spongepowered/api/event/SpongeEventFactory.java
@@ -3982,6 +3982,34 @@ public class SpongeEventFactory {
     /**
      * AUTOMATICALLY GENERATED, DO NOT EDIT.
      * Creates a new instance of
+     * {@link org.spongepowered.api.event.network.ClientConnectionEvent.Handshake}.
+     * 
+     * @param cause The cause
+     * @param formatter The formatter
+     * @param gameProfile The game profile
+     * @param handshakeString The handshake string
+     * @param serverHostname The server hostname
+     * @param socketAddressHostname The socket address hostname
+     * @param failed The failed
+     * @param messageCancelled The message cancelled
+     * @return A new handshake client connection event
+     */
+    public static ClientConnectionEvent.Handshake createClientConnectionEventHandshake(Cause cause, MessageEvent.MessageFormatter formatter, Optional<GameProfile> gameProfile, String handshakeString, Optional<String> serverHostname, Optional<String> socketAddressHostname, boolean failed, boolean messageCancelled) {
+        HashMap<String, Object> values = new HashMap<>();
+        values.put("cause", cause);
+        values.put("formatter", formatter);
+        values.put("gameProfile", gameProfile);
+        values.put("handshakeString", handshakeString);
+        values.put("serverHostname", serverHostname);
+        values.put("socketAddressHostname", socketAddressHostname);
+        values.put("failed", failed);
+        values.put("messageCancelled", messageCancelled);
+        return SpongeEventFactoryUtils.createEventImpl(ClientConnectionEvent.Handshake.class, values);
+    }
+
+    /**
+     * AUTOMATICALLY GENERATED, DO NOT EDIT.
+     * Creates a new instance of
      * {@link org.spongepowered.api.event.network.ClientConnectionEvent.Join}.
      * 
      * @param cause The cause
--- a/src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
@@ -42,6 +42,8 @@ import org.spongepowered.api.text.TextRepresentable;
 import org.spongepowered.api.world.World;
 
 import java.net.InetAddress;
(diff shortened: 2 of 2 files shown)


Zidane:
@kashike 

I'm not sure about this one. I'm quite weary about declaring official support of modifying the output of the handshake like this...basically I'm on the fence about it

@SpongePowered/contributors 
@SpongePowered/developers 

Thoughts team?

Zidane:
@kashike 

Instead of a fail toggle, could you use cancellation status instead?

dualspiral:
I imagine you're doing this because you want to do something different with Waterfall. I can see what this is meant to achieve, but I really don't think that exposing this as an API, at least in this way, is a good idea. I like the abstracting done in Common/Forge though.

**Idealistically**, I would argue this is definitely not something that should be exposed to plugins. While the event looks like it would only fire if the mixin plugin was enabled, I can see that there is huge potential for abuse. IP forwarding, client detail forwarding, whatever you want to call it is already insecure enough that opening it up like this is potentially asking for trouble. Indeed, should plugins be touching anything that affects the login process, other than to disconnect the user? I really feel like they shouldn't be able to, because the server - read Sponge - can no longer trust whether the client is who they say they are.

Yes, when putting the Bungee support in, we've modified the login sequence, but that's in the realm of Sponge. That decision on whether to trust a set of credentials _should not_ be taken out of Sponge's hands.

While it's out of the scope of the PR, I personally would much rather see a standardised IP forwarding protocol that proxy implementations are required to implement. 

**Realistically**, the problem is that IP forwarding is that it is fragmented. I've been waiting for my Bungee PR to be accepted for a long time (for whatever reason, not going to speculate here), and it seems that the _de facto_ forwarding spec is controlled by one group, and I don't see there being any movement in the way things are done there soon. It does make it difficult to deliver alternative methods without having such a plugin system, so I can see what you're trying to do.

However, what happens if more than one plugin tries to handle the event? If you do this, it needs to be done in such a way such that only one plugin can handle this.

I think that you're better off having abstracted Proxy support that is managed and accepted by the Sponge implementations themselves. That way, the burden of trust lies with Sponge/Forge/Minecraft, how _they_ choose to accept clients. Gut feeling says that a plugin accessible API for this should not exist.

Mumfrey:
> I think that you're better off having abstracted Proxy support that is managed and accepted on a case by case basis by the Sponge implementations themselves. That way, the burden of trust lies with Sponge/Forge/Minecraft, how they choose to accept clients. Gut feeling says that a plugin accessible API for this should not exist.

**+1**

gabizou:
> the burden of trust lies with Sponge/Forge/Minecraft, how they choose to accept clients.

@dualspiral, you hit it right on the nail.

I won't reiterate what was said because it's been said quite well. I will have to decline this PR getting added to the API.

kashike:
@gabizou @Zidane 
As it stands, if someone makes a new proxy (or changes something in their fork), it would have to be used by many people to get it put into Sponge itself - maintaing a fork just for authentication is annoying. It'd be the same as requiring every dependency to be managed by Sponge - "we don't know why the fuck you'd need XYZ, so we're going to close this" as it is currently, Sponge is just forcing people to use BungeeCord-based proxies.
"Oh, it's a great idea to require the implementation to dictate what proxy solution I can use."

 LIlyPad suffers badly from this in Spigot/forks:  https://github.com/LilyPad/Bukkit-Connect/tree/master/src/main/java/lilypad/bukkit/connect

Would something written in just SpongeCommon/SpongeForge (non-SpongeAPI) that people can use instead? It seems like the main issue is this being in the API.

gabizou:
> it would have to be used by many people to get it put into Sponge itself

On what premise?

If a new proxy is wanting to be used with Sponge, it's pretty easy to make a PR to Sponge to add support, if anything, modularizing the proxy support in Sponge would greatly benefit the community in general. Then multiple proxy setups could be supported officially and maintenance would be pretty much the same idea.

PR #1189, opened 2016-04-18
Player GUI API

simon816:
[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/935) | **SpongeAPI**

This is a very WIP API. I'm open to suggestions on how the API should be designed.  
If I have missed and open-able GUIs then let me know.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/Player.java
@@ -38,6 +38,7 @@ import org.spongepowered.api.entity.living.Humanoid;
 import org.spongepowered.api.entity.living.player.gamemode.GameMode;
 import org.spongepowered.api.entity.living.player.tab.TabList;
 import org.spongepowered.api.event.cause.Cause;
+import org.spongepowered.api.gui.window.Window;
 import org.spongepowered.api.item.inventory.Container;
 import org.spongepowered.api.item.inventory.Inventory;
 import org.spongepowered.api.network.PlayerConnection;
@@ -101,6 +102,31 @@ public interface Player extends Humanoid, User, Locatable, RemoteSource, Viewer,
      * @return whether or not closing the inventory succeeded
      */
     boolean closeInventory(Cause cause) throws IllegalArgumentException;
+
+    /**
+     * Gets the active {@link Window} displayed to the player, or
+     * {@link Optional#empty} if no window is shown or otherwise detected from
+     * the server.
+     *
+     * @return The active window, if available
+     */
+    Optional<Window> getActiveWindow();
+
+    /**
+     * Closes the active window if possible.
+     *
+     * @return True if the window was closed
+     */
+    boolean closeActiveWindow();
+
+    /**
+     * Shows the window to the player. If the window is already being shown to
+     * another player then this returns false.
+     *
+     * @param window The window to show
+     * @return Whether it was shown to the player
+     */
+    boolean showWindow(Window window);
 
     /**
      * Gets the view distance setting of the player. This value represents the
--- a/src/main/java/org/spongepowered/api/entity/living/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/Player.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.api.entity.living.player;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.spongepowered.api.Server;
 import org.spongepowered.api.block.tileentity.EnderChest;
 import org.spongepowered.api.command.source.RemoteSource;
@@ -113,20 +115,26 @@ public interface Player extends Humanoid, User, Locatable, RemoteSource, Viewer,
     Optional<Window> getActiveWindow();
 
     /**
-     * Closes the active window if possible.
+     * Closes the active window.
      *
      * @return True if the window was closed
      */
-    boolean closeActiveWindow();
+    default void closeActiveWindow() {
+        Optional<Window> active = getActiveWindow();
+        if (active.isPresent()) {
+            active.get().close(this);
+        }
+    }
 
     /**
-     * Shows the window to the player. If the window is already being shown to
-     * another player then this returns false.
+     * Shows the window to the player if possible.
      *
      * @param window The window to show
      * @return Whether it was shown to the player
      */
-    boolean showWindow(Window window);
+    default boolean showWindow(Window window) {
+        return !getActiveWindow().isPresent() && checkNotNull(window, "window").show(this);
+    }
 
     /**
      * Gets the view distance setting of the player. This value represents the
--- a/src/main/java/org/spongepowered/api/gui/window/VillagerTradeWindow.java
+++ b/src/main/java/org/spongepowered/api/gui/window/VillagerTradeWindow.java
@@ -1,0 +1,50 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.gui.window;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.manipulator.mutable.entity.TradeOfferData;
+import org.spongepowered.api.item.merchant.Merchant;
+
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+public interface VillagerTradeWindow extends ContainerWindow {
+
+    Optional<Merchant> getMerchant();
+
+    void setMerchant(@Nullable Merchant merchant);
+
+    void setVirtualTradeData(@Nullable TradeOfferData tradeOfferData);
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    interface Builder extends Window.Builder<VillagerTradeWindow, Builder> {
+    }
+
+}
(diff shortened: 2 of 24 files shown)


ewized:
I think it should be called Window as that's what it's called internally and is known by the packet name.

liach:
I suggest to name it like `AnvilGui`

ewized:
@liach what about `AnvilWindow`

liach:
Anyways name like `Window` or `Gui` should be suffixes, as the names of `AITask`s and `Minecart`s.

src/main/java/org/spongepowered/api/gui/Gui.java
```diff
@@ -0,0 +1,44 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.gui;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
+
+public interface Gui {
```
liach: A list of buttons?

src/main/java/org/spongepowered/api/gui/GuiBook.java
```diff
@@ -0,0 +1,29 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.gui;
+
+public interface GuiBook extends Gui {
+// TODO Only possible to open editor client side (right-click book)
```
windy1: You actually can't even open the "editor" you can only open signed books so it will always be read-only.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -96,6 +97,11 @@
      */
     void closeInventory(Cause cause) throws IllegalArgumentException;
 
+    boolean isViewingGui();
```
ST-DDT: I would remove this method because getOpenGui().isPresent() is all you need.

ST-DDT:
The GUIs should expose their inventory if they have one.

liach:
@ST-DDT Actually I think it's better to relate some Guis with their Container inventories, like forge does.

VcSaJen:
> If I have missed and open-able GUIs then let me know.

Also there's a demo screen:
screenshot
I dunno about the use case (text override via resource packs?), but it's possible.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -96,6 +97,11 @@
...
+    Optional<Gui> getOpenGui();
```
kashike: `getCurrentGui` sounds better

src/main/java/org/spongepowered/api/gui/GuiAnvil.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.gui;
+
+public interface GuiAnvil extends GuiContainer {
```
kashike: Should be `AnvilGui` (applies to all `Gui*`).

src/main/java/org/spongepowered/api/gui/GuiDemoMessage.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.gui;
+
+public interface GuiDemoMessage extends Gui {
```
kashike: `DemoGui`, it isn't just a "message"

src/main/java/org/spongepowered/api/gui/Gui.java
simon816: @liach buttons are client-side, this is server-side and is only able to send data the protocol allows

src/main/java/org/spongepowered/api/gui/GuiBook.java
simon816: I've modified this to use `BookView`

src/main/java/org/spongepowered/api/gui/window/Window.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Optional;
+
+public interface Window {
+
+    /**
+     * Gets the player being shown this window, if currently shown.
+     *
+     * @return The player, if available
+     */
+    Optional<Player> getPlayerShown();
```
kashike: I still don't agree with this

src/main/java/org/spongepowered/api/gui/window/VillagerTradeWindow.java
```diff
@@ -0,0 +1,50 @@
...
+
+import javax.annotation.Nullable;
+
+public interface VillagerTradeWindow extends ContainerWindow {
+
+    Optional<Merchant> getMerchant();
+
+    void setMerchant(@Nullable Merchant merchant);
+
+    void setVirtualTradeData(@Nullable TradeOfferData tradeOfferData);
```
kashike, line 41: Getter?

simon816:
One problem I can see is detecting when the player opens and closes a window. It could be possible to detect when a player right-clicks on blocks/items that cause windows to open, though I'm not sure whether that would be a great idea.
So either `getActiveWindow` will only return a window that's been opened by calling `showWindow`, I remove the method, or I try to detect when a window will be shown.

The other issue is that not all Windows report when they are closed. Containers do, but for example the BookWindow (`GuiScreenBook`) doesn't. Not sure on a good way to get around this constraint.

liach:
Status?

src/main/java/org/spongepowered/api/gui/window/Window.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Shows this window to the player.
+     *
+     * @param player The player
+     * @return Whether the window was shown to the player
+     */
+    boolean show(Player player);
+
+    void close(Player player);
```
liach: Move this method to `ContainerWindow`?

liach:
You can create a class called `ServerWindow` which server-controllable windows extends from.

Aaron1011:
Closing in favor of https://github.com/SpongePowered/SpongeAPI/pull/1258

simon816:
@Aaron1011 not really? This is for displaying windows on the client (could be inventory, could be others).
Custom inventory would be something that provides the backing data for an inventory, a Window (the view) is what it gets displayed on

Aaron1011:
@simon816: No, the custom inventory PR I linked handles all aspects of the creation process - including the underlying window type that's used.

simon816:
You're not going to create a `BookWindow` or a `WinGameWindow` from an inventory system.
The Window API is more to do with showing `GuiScreen`s to a client with _full_ control of the things displayed (e.g. for a horse inventory window you can customize the model horse in the gui)

simon816:
I want to make sure I don't overlap and conflict with the current proposal for a custom inventory API.
Feedback would be appreciated.

My take on the design is that a custom inventory defines the backing model for an inventory, kind of like a database model.
This GUI API is used to display that to the client. The custom inventory implementation will encompass the ContainerWindow-based GUIs and handle interaction from the player to the backend.

Zidane:
@simon816 

This isn't really needed with InventoryArchetypes in that branch...you can build and display any Vanilla container in the game....in what scenario would you open the Window to the client and have no backing inventory?

Zidane:
To add further, you can figure out which Container you have open by calling getOpenInventory (in the branch) as that returns a Container and then call getArchetype... You can THEN also figure out the carrier as well :P.

What does this offer us? Not trying to stomp on your work, just want to make sure it all makes sense.

simon816:
Ah well all that didn't exist when I made this PR. I'll go over the current custom inventory proposal and see whether there's anything this window API can add to it.

simon816:
OK so I think most inventory-related Windows could be handled by `InventoryArchetype`, but not completely (more details below).
There are still some Windows that aren't covered by inventories.

My original inspiration for this was way back when the method `editSign()` was proposed (#682) and the subsequent issue (#696) that I based this PR on.

Here is a list of Windows that are non-inventory based:
 * `BookWindow`
 * `CommandBlockWindow`
 * `DemoWindow`
 * `SleepingWindow`
 * `WinGameWindow`
 * `SignWindow`

All others are types of `InventoryArchetype`

I can definitely see usecases for `SignWindow` and `BookWindow` because you might not want to have a physical object in the world in order to show these to players.

In terms of the inventory-based Windows, I'd like it to be possible to control as much of the Window as possible, for example a virtual furnace that can cook items in your backpack. You'd need to be able to offer `FurnaceData` to the Window to update the display, another example could be a virtual brewing table.

I can't see any way to actually offer this data back to an inventory instance and I don't think it makes sense to do so. It's not part of the _inventory_, rather, it's the state of the virtual furnace.

I'm not really sure how best to create something like this without creating a `Window` type for each `InventoryArchetype`.

Deamon5550:

ryantheleach:
This should probably be the preferred way to show inventory menus to clients as well right?
Where the current inventory API is better used for virtual inventories.

The main problem is potential dupe issues if events arn't cancelled correctly, where as if the inventory is only a window/gui rather then actually backed by something means that items can't be mistakenly taken onto another slot / container / dropped / cursor right?

XakepSDK:
@simon816 any news?

Cybermaxke:
@simon816 Are you still planning to work on this? If so, you can also take a look at some changes I am making in Lantern which could be useful for the implementation of this API. I seperated a container implementation into two layers, a client one (which controls the window contents) and a server one (which handles clicking logic, number keys, etc.)
You can find it here: https://github.com/LanternPowered/LanternServer/pull/37
Just know that it still WIP, but it is almost finished.

simon816:
Yes, I will update this soon&trade;. I was discussing with a few sponge devs yesterday about whether this is redundant now that we have the custom inventory API.
I will re-do this PR to extend the custom inventory API, since a lot is already covered by it.
There's some pretty cool things I want to be able to do, e.g. custom control of the furnace/brewing stand progress bars, customising the horse model in the horse inventory gui etc.

BrainStone:
Updates on this?

parlough:
@simon816 When you eventually come back to this, will the PR be able to cover controlling repair costs in anvils as well as the name changing field? Or is that not relevant to this?

ryantheleach:
@simon816 is this still in WIP stage? 
can you comment with "~wip" to mark it as Work In Progress, or "~qa" if it is ready for review?

Also does it have breaking changes? I've milestone'd it to API 8 for now.

Alternatively, since there has been demand, if you don't plan on updating this soon, maybe talk to @Meronat and @Cybermaxke about potentially handing it off, or when you will next have time to work on this.

Zidane:

PR #1195, opened 2016-04-22
Add helper methods to CommandResult

kashike:
This adds the following helper methods to `CommandResult` to prevent lots of duplication in plugins:
- `CommandResult.successCount(int)`
- `CommandResult.affectedBlocks(int)`
- `CommandResult.affectedEntities(int)`
- `CommandResult.affectedItems(int)`
- `CommandResult.queryResult(int)`

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandResult.java
+++ b/src/main/java/org/spongepowered/api/command/CommandResult.java
@@ -69,6 +69,60 @@ public class CommandResult {
     }
 
     /**
+     * Returns a result indicating the command was processed with a single success.
+     *
+     * @param count The success count
+     * @return The result
+     */
+    public static CommandResult successCount(int count) {
+        return builder().successCount(count).build();
+    }
+
+    /**
+     * Returns a result indicating the command was processed with an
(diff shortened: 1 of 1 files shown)


ryantheleach:
When should reach result type be returned? does this automatically set success? Is it possible to set affected entities, but for the command to still fail?

kashike:
@ryantheleach These are simply helper methods.
- success:
 - old: `CommandResult.builder().successCount(count).build()`
 - new `CommandResult.successCount(1)`
- affected blocks:
 - old: `CommandResult.builder().affectedBlocks(count).build()`
 - new `CommandResult.affectedBlocks(1)`
- ... etc

ryantheleach:
Yes but should developers use them mutually exclusively? or is there times when you would want to use both in a builder? I don't understand it from reading the docs of the builder, and figured while you are editing the class you could provide some guidance to developers.

kashike:
@ryantheleach Vanilla only updates one at a time, yes, but there isn't anything from preventing vanilla, mods, or plugins from updating multiple values at once.

PR #1197, opened 2016-04-24
Item pickup and despawn delay

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1197) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/667)


simon816:
Just though I'd point out that @gabizou said that pickup delay would be the same as the current `DelayableData`. https://forums.spongepowered.org/t/item-pickup-delays/10629
Don't know that should still be the case or not.

kashike:
@simon816 No. `DelayableData`'s min/max are different from an item's delay.

Flibio:
Any progress on this issue?

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableDespawnDelayData.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@link Item}.
+ */
+public interface ImmutableDespawnDelayData extends ImmutableDataManipulator<ImmutableDespawnDelayData, DespawnDelayData> {
+
+    /**
+     * Gets the {@link ImmutableBoundedValue} for the "despawn delay".
+     *
+     * @return The immutable bounded value for the "despawn delay"
+     */
+    ImmutableBoundedValue<Integer> delay();
```
Zidane: Maybe javadoc should state upper and lower bounds.

Issue #1168, opened 2016-04-02
Custom Data Factory


gabizou:
It should be possible at this point to provide a class factory that will create given a certain set of parameters (or simply using a manipulator generator blueprint of sorts) to have Sponge generate a custom class with some unique identification that gives minimal code base requirements from plugin developers. Effectively, this can eliminate the need altogether for plugin developers to need to create at least 3 classes for each custom data themselves.

What it could look like:
```java

public class MyDataStuff {

  public static final Key<Value<String>> FOO = KeyFactory.create(String.class, Value.class, DataQuery.of("myData"));

  public static Class<? extends DataManipulator<?, ?>> MUTABLE_FOO = CustomDataGenerator.builder()
    .id("FooData")
    .withString(FOO)
    .version(1)
    .create(PluginClass.getInstance());
  public static Class<? extends ImmutableDataManipulator<?, ?>> IMMUTABLE_FOO = CustomDataGenerator.immutableOf(MUTABLE_FOO);
 
}

```

The only issue of course is that usability of the manipulator as a class itself is completely useless since it's type erased to a wild carded class. The advantage is that what takes 3 lines of code from a developer is all it would take to get started using custom data relatively easily.

The disadvantage of this all is that of course the plugin would not be able to see the data class itself, unless in the implementation. 

My idea is that the class name generated itself would utilize the plugins' id as a root and the ending being retrieved from the id provided though the generator.

I'm hoping that the community could see this for what it is as an addition and not a replacement of some serious custom data.

A note: The generator would only be able to accept some default serializable/supported objects like primitives, strings, and potentially other `DataSerializable`s.

gabizou:
Pinging @Aaron1011 and @viveleroi as I've seen both to have some potential interest in such a thing being discussed.

viveleroi:
I think this would be great. It's still a bit lengthy for situations where we're only storing a boolean or a single primitive value but it's quite a lot better than the three class files I had to a) learn and b) implement just to store a single boolean. Not only is this example more compact but it's a lot more understandable.

gabizou:
@viveleroi The only loss here though is that the factory will not be able to provide you with a nice generic class to use with data manipulators themselves. The fortunate part is that you'll still be able to generate the generic keys to interact with that way.

kashike:
@gabizou `withString`? Why not just `string`? No other builders use `with`, and it'll be weird to start doing so

gabizou:
[...]

PR #1198, opened 2016-04-25
Custom Data Generation

gabizou:
Following up on #1168.

While custom data has been supported for many months now, and a few plugin developers have started using it, I've been listening to the qualms with _implementing_ custom data. For the most part, the issues are:
- Too many methods to write that are somewhat confusing
- Forgetting to call `registerGettersAndSetters()`
- Using a mutable variant in the immutable object as a field for reference
- Writing serialization code that works is hard (not really, just not super obvious)
- Having to write 3 classes just to store a `String` onto an `Entity` or `ItemStack` is a bit excessive

More elaborate descriptions will come.

Examples of use for a single/multiple value based `DataManipulator` without the requirement of having hard interfaced classes (mainly where you focus on using `Key`s only):

``` java


public class SimpleStringCustomData {

    // For obvious reasons, we want to be able to use the key for various things. so we generate it!
    public static final Key<Value<String>> MY_STRING = KeyFactory.makeSingleKey(String.class, Value.class, DataQuery.of("MyString"), "com.gabizou.my_string", "MyStringKey");

    // We get to use the CustomDataProvider, however, since there are no hard interfaces provided,
    // we can't use any form of generics. Fortunately for us, we don't really care about generics,
    // except for using our Key. With that, we still need to provide some information about the
    // default values of that key.
    // So, in essence, we still get our DataRegistration object back, which can be private.
    // The nice thing about this is that the registration, once validated and classes generated,
    // the classes and builder instance are also validated and registered with the DataManager.
    // No further registration is required
    private static final DataRegistration<?, ?> MY_STRING_REGISTRATION = CustomDataProvider.builder()
            .id("CustomStringData")
            //.withInt(MY_STRING, 54) // Will not compile due to method requirements :D
            .withString(MY_STRING, "DefaultString")
            .build(null); // Not actually supposed to be null, but it can be for the sake of demonstration

    // We now have our generated "super class" which can be used in any methods accepting manipulator classes
    public static final Class<? extends DataManipulator<?, ?>> MY_STRING_MANIPULATOR = MY_STRING_REGISTRATION.getSuperManipulator();
    // We also have our immutable registration
    public static final Class<? extends ImmutableDataManipulator<?, ?>> MY_STRING_IMMUTABLE_MANIPULATOR = MY_STRING_REGISTRATION.getSuperImmutable();
    // And of course, our data builder class
    public static final Class<? extends DataManipulatorBuilder<?, ?>> MY_STRING_BUILDER_CLASS = MY_STRING_REGISTRATION.getBuilderClass();
    // Oh! And I almost forgot, we can have our own static instance of our manipulator builder since that is generated for us as well!
    public static final DataManipulatorBuilder<?, ?> MY_STRING_BUILDER = MY_STRING_REGISTRATION.getBuilder();

    // After all this is set and done, the DataRegistration is also able to produce new instances
    // without having to use the DataManager, since it already has the instance prepared.


}
```


RysingDragon:
I don't really have much useful input on this but i just wanted to say how i feel about it. Because of how the data api is now, mainly issues 1 and 5 on the list you wrote, i actually tend to avoid it. If needed i just use nms for setting custom nbt since i find it much easier to just get the nbttagcompound and just set/get the data i need.

src/main/java/org/spongepowered/api/data/manipulator/generator/CustomDataProvider.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public static <T extends DataManipulator<T, I>, I extends ImmutableDataManipulator<I, T>> TypeBuilder<T, I> builder(Class<T> manipulatorClass, Class<I> immutableClass) {
+        return Sponge.getDataManager().createCustomBuilder(manipulatorClass, immutableClass);
+    }
+
+
+    public interface RawBuilder extends ResettableBuilder<DataRegistration<?, ?>, RawBuilder> {
+
+        RawBuilder withString(Key<? extends BaseValue<String>> key, String defaultValue);
```
simon816: with\* prefix isn't preferred in builders

src/main/java/org/spongepowered/api/data/manipulator/generator/testing/SimpleStringCustomData.java
```diff
@@ -0,0 +1,47 @@
...
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.DataManipulatorBuilder;
+import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.manipulator.generator.CustomDataProvider;
+import org.spongepowered.api.data.manipulator.generator.DataRegistration;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.data.value.BaseValue;
+import org.spongepowered.api.data.value.mutable.Value;
+
+public class SimpleStringCustomData {
```
gabizou, line 20: @RysingDragon This class should suffice to make an example of what you're looking for.

src/main/java/org/spongepowered/api/data/manipulator/generator/CustomDataProvider.java
SocraticPhoenix: I feel like with\* is well suited in this context, since we're adding "fields" to a "class." Then again, convention trumps all, doesn't it?

src/main/java/org/spongepowered/api/data/manipulator/generator/testing/SimpleStringCustomData.java
DawnTheFawn1, line 20: That definitely does look a lot simpler now, thanks gabizou for all the work you're doing :)

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -131,6 +135,8 @@ static ItemStack of(ItemType itemType, int quantity) {
          */
         Builder itemType(ItemType itemType);
 
+        ItemType getCurrentItem();
```
Zidane, line 160: Builders never return their current values as it should never matter. You're building the object, not looking up properties.

gabizou, line 160: @Zidane The feature was requested by @Wolfzein for adding some precondition based data application.

me4502:
One issue with Custom Data - is that if a plugin is removed, or fails to start for some reason, their data is removed from the world.

Maybe if it was still loaded as a raw NBT tag, and then re-saved to the world if it can't be loaded?

gabizou:
> One issue with Custom Data - is that if a plugin is removed, or fails to start for some reason, their data is removed from the world.
> 
> Maybe if it was still loaded as a raw NBT tag, and then re-saved to the world if it can't be loaded?

Not really relevant to this PR. It's more of an issue with implementation in general and has been discussed at length on multiple occasions.

me4502:
It's a PR on issues with custom data - so I presumed it'd be relevant

kinggoesgaming:
@me4502 
its a PR on custom data generation xD

20zinnm:
Please, we need this 😄 .

src/main/java/org/spongepowered/api/data/key/Key.java
```diff
@@ -49,6 +50,14 @@
     Class<V> getValueClass();
 
     /**
+     * Gets the class of the element of the {@link Value} this {@link Key}
+     * is representing.
+     *
+     * @return The element class
+     */
+    Class<?> getElementClass();
```
Cybermaxke: Wouldn't it be more usefull to provide a `TypeToken` instead of a `Class`, this could contain the generics of `Map`s, `Set`s and can be used to get the serializer of the values (and keys).

gabizou: I would rather not use type tokens because of their requirement that people use them as asnonymous classes to get the generic parameters or fully understand the limitations of type tokens in the first place. The end result still is the same: in a pure abstract system, the only real use for this method is validation at runtime, not compile time.

ryantheleach:
This is for custom/fake data right?

So would you still need to fallback to the old way for implementing "mod" dataManipulators to expose them as API to other plugins?

gabizou:
> So would you still need to fallback to the old way for implementing "mod" dataManipulators to expose them as API to other plugins?

If this is written correctly, we can effectively remove the implemented `DataManipulator`s that Sponge provides.

src/main/java/org/spongepowered/api/util/CommonTokens.java
```diff
@@ -0,0 +1,505 @@
...
+    };
+    public static final TypeToken<ZombieType> ZOMBIE_TYPE_TOKEN = new TypeToken<ZombieType>() {
+        private static final long serialVersionUID = -1;
+    };
+    public static final TypeToken<Value<ZombieType>> ZOMBIE_TYPE_VALUE_TOKEN = new TypeToken<Value<ZombieType>>() {
+        private static final long serialVersionUID = -1;
+    };
+
+    private CommonTokens() {
+        throw new UnsupportedOperationException("Should not be constructing a CommonTokens instance!");
```
Deamon5550, line 502: @DemonWav says that you're dumb.

ZephireNZ:
One thing I discovered with my closed PR - Setting a Key to a custom holder without the datamanipulator present always fails. Perhaps we should have some way of linking Keys to datamanipulatorbuilders, so that it can try proactively add the default data before setting a key.

While that's a minor annoyance for the current system, the RawBuilder system (where only keys need to be exposed) could make things a lot more annoying.

Not relevant to the API part, but there's no impl PR just yet :P

gabizou:
> One thing I discovered with my closed PR - Setting a Key to a custom holder without the datamanipulator present always fails. Perhaps we should have some way of linking Keys to datamanipulatorbuilders, so that it can try proactively add the default data before setting a key.

That's part of the intention of having the builder system ;)

src/main/java/org/spongepowered/api/util/CommonTokens.java
```diff
@@ -0,0 +1,505 @@
...
+    };
+    public static final TypeToken<Value<PickupRule>> PICKUP_VALUE_TOKEN = new TypeToken<Value<PickupRule>>() {
+        private static final long serialVersionUID = -1;
+    };
+    public static final TypeToken<ZombieType> ZOMBIE_TYPE_TOKEN = new TypeToken<ZombieType>() {
+        private static final long serialVersionUID = -1;
+    };
+    public static final TypeToken<Value<ZombieType>> ZOMBIE_TYPE_VALUE_TOKEN = new TypeToken<Value<ZombieType>>() {
+        private static final long serialVersionUID = -1;
+    };
```
liach: You need to sort these fields

src/main/java/org/spongepowered/api/data/manipulator/generator/testing/DummyCustomDataData.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.api.data.key.KeyFactory;
+import org.spongepowered.api.data.manipulator.DataManipulatorBuilder;
+import org.spongepowered.api.data.manipulator.generator.DataRegistration;
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.util.CommonTokens;
+
+public class DummyCustomDataData {
```
liach, line 15: Can you move these dummies to test sources?

gabizou: Clearly you missed the point of "WIP" and the last time that I've updated this PR... There's other PR's/branches that have been made since towards this overall PR.

liach:
It is almost one year... Why is this still "in progress"? Seems it is already functional.

gabizou:
> It is almost one year... Why is this still "in progress"? Seems it is already functional.

Clearly you haven't tried running the API and implementation branches to even *test*.

gabizou:
@Cybermaxke will be taking over this aspect of the API.

PR #1186, opened 2016-04-16
[WIP] Add annotation based commands

jamierocks:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/656)

Example: 
```java
    @Command(
        aliases = {"actions"}
    )
    @TranslatableDescription(
        description = "some.translation.key.description",
        usage = "some.translation.key.usage",
        help = "some.translation.key.help"
    )
    @Group("kittens") // /kittens actions
    @Group("cats") // /cats actions
    @Permission({
        "some.permission",
        "some.other.permission"
    })
    public CommandResult example() {
        return CommandResult.empty();
    }

    @Command(
        aliases = {"pet"}
    )
    @PlainDescription(
        description = "Pet the kitty",
        usage = "/kittens pet",
        help = "how2helppet?"
    )
    @Group({"kittens", "actions"}) // /kittens actions pet
    @Group({"cats", "actions"}) // /cats actions pet
    @Permission({
        "some.permission",
        "some.other.permission"
    })
    public CommandResult anotherExample() {
        return CommandResult.empty();
    }
```

Issue #1171, opened 2016-04-03
@Command annotation system (on top of existing, lower-level, command API)


vorburger:
I hope it's OK if I open this ENHANCEMENT ticket for a new discussion about a @Command annotation system in the Sponge API?

Apparently there were previous takes on this, I've found https://github.com/SpongePowered/SpongeAPI/issues/276 and https://github.com/SpongePowered/SpongeAPI/pull/561, both closed - yet still nada.

This does seem to be something people want - via a quick search I've found 4 (!) different separate outside implementations of this idea, see:

1. https://github.com/sk89q/Intake (@sk89q is that for Sponge? or do you have something else somewhere to use Intake with Sponge?)

2. https://github.com/minnymin3/Directive/ from https://forums.spongepowered.org/t/directive-a-light-weight-annotation-based-command-system/8325

3. https://github.com/WesJD/CommandManager from https://forums.spongepowered.org/t/commandmanager-an-annotation-based-command-system/11529

4. my very own https://github.com/vorburger/SwissKnightMinecraft/tree/ad5954212cc4f1bda3596d13db2e53a6d3ebb3f3/SpongePowered/MyFirstSpongePlugIn/src/main/java/ch/vorburger/minecraft/command (WIP, NOK)

At the very least, perhaps the Sponge documentation could... say something about this? As in "Planned, watch (this ticket?)" or "We'll never, but perhaps you would like to use (sk89q/Intake, with usage example?)"

kashike:
Intake is quite good, I'd be :+1: for including it (or a modified version) in SpongeAPI.

kenzierocks:
I'd love to have a non-annotation based option, personally I think annotations are too much magic w/ commands and also can't be configured easily by a config file.

JonathanxD:
I'm implementing my own Command API https://github.com/JonathanxD/WCommands/, but i need to finish javadoc first.

ryantheleach:
>I'd love to have a non-annotation based option

Isn't that what we already have?

me4502:
I agree with the inclusion of Intake. The way it's written means that it'd be super easy to implement.

zml2008:
[...]

PR #1201, opened 2016-04-28
Annotation-based commands

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1201) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/675)
Implements #1171 

---

This PR adds an annotation-based command API to SpongeAPI.

Example:

``` java
    @Listener
    public void starting(final GameStartingServerEvent event) {
        this.game.getCommandManager().install(this, new AbstractCommandModule() {
            @Override
            protected void configure() {
                // Here I bind any commands registered through this command module who has `Logger` in the signature
                // to `MiserableTest.this.logger` which, in this case, is the Logger of my test plugin.
                this.bind(Logger.class).toProvider((args, modifiers) -> MiserableTest.this.logger);
                // Here I bind any commands registered through this command module who has `World` in the signature
                // to the first world found using the iterator which, in this case, is the Logger of my test plugin.
                this.bind(World.class).toProvider((args, modifiers) -> Sponge.getServer().getWorlds().iterator().next());
                this.register(new ExampleCommands());
            }
        });
    }

    public static class ExampleCommands {

        /*
         * /kittens actions
         * /cats actions
         */
        // @formatter:off
        @Parent("kittens")
        @Parent("cats")
        @Command("actions")
        @Permission({
            "example.kittens",
            "example.kittens.actions"
        })
        @PlainDescription(
            description = "Display a listing of actions",
            usage = "/kittens pet",
            help = ""
        )
        @Default(defaultOnly = true)
        // @formatter:on
        public CommandResult actions(CommandSource source) {
            source.sendMessage(Text.of("Actions"));
        }

        /*
         * /kittens actions pet
         * /cats actions pet
         */
        // @formatter:off
        @Parent({"kittens", "actions"})
        @Parent({"cats", "actions"})
        @Command("pet")
        @Permission({
            "example.kittens",
            "example.kittens.actions",
            "example.kittens.actions.pet"
        })
        // The description, usage, and help texts are static.
        @PlainDescription(
            description = "Pet the kitty",
            usage = "/kittens pet",
            help = "how2helppet?"
        )
        // @formatter:on
        public CommandResult pet(Logger logger, CommandSource source) {
            source.sendMessage(Text.of("Success from pet"));
            logger.info("Meow.");
        }
    }
```

---

Thanks to @jamierocks for his initial work on #1186

### DIFF
--- a/src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
+++ b/src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
@@ -1,0 +1,366 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.command;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import org.slf4j.Logger;
+import org.spongepowered.api.Game;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.command.annotation.Command;
+import org.spongepowered.api.command.annotation.PlainDescription;
+import org.spongepowered.api.command.args.CommandArgs;
+import org.spongepowered.api.command.args.parsing.InputTokenizer;
+import org.spongepowered.api.command.binding.Binder;
+import org.spongepowered.api.command.binding.BindingBuilder;
+import org.spongepowered.api.command.binding.BindingKey;
+import org.spongepowered.api.command.binding.ReadOnlyDelegateBinder;
+import org.spongepowered.api.command.binding.simple.SimpleBinder;
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.text.serializer.TextSerializer;
+import org.spongepowered.api.text.serializer.TextSerializers;
+import org.spongepowered.api.text.translation.TextFunction;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+
+import javax.annotation.Nullable;
+
+/**
+ * A command module contributes configuration information to commands, and
+ * registers commands with a {@link CommandManager}.
+ *
+ * <p>A command module is <b>not</b> re-usable.</p>
+ */
+public abstract class AbstractCommandModule {
+
+    private Binder binder = new SimpleBinder();
+    @SuppressWarnings("NullableProblems")
+    private PluginContainer plugin;
+    @SuppressWarnings("NullableProblems")
+    private CommandManager manager;
+    private InputTokenizer inputTokenizer = InputTokenizer.quotedStrings(false);
+    private TextSerializer textSerializer = TextSerializers.FORMATTING_CODE;
+    @Nullable private CommandThrowableProcessor throwableProcessor;
+    @Nullable private TextFunction textFunction;
+    private CommandAuthorizer authorizer = CommandAuthorizer.CALLABLE;
+    private boolean configured;
+    private boolean registered;
+
+    protected AbstractCommandModule() {
+    }
+
+    /**
+     * Contributes bindings and other configurations for this module to {@code binder}.
+     *
+     * @param plugin The plugin
+     * @param manager The command manager
+     */
+    @SuppressWarnings("ConstantConditions")
+    public final void configure(PluginContainer plugin, CommandManager manager) {
+        checkState(this.manager == null, "Re-entry is not allowed.");
+        this.plugin = checkNotNull(plugin, "plugin");
+        this.manager = checkNotNull(manager, "manager");
+        this.preConfigure();
+        this.configure();
+        // Create a delegating read-only binder
+        this.binder = new ReadOnlyDelegateBinder(this.binder);
+        this.configured = true;
+    }
+
+    private void preConfigure() {
+        // Register a few of the most common arguments by default
+        this.bind(CommandSource.class).toProvider((source, args, modifiers) -> source);
+        this.bind(CommandArgs.class).toProvider((source, args, modifiers) -> args);
+        this.bind(Game.class).toProvider((source, args, modifiers) -> Sponge.getGame());
+        this.bind(Logger.class).toProvider((source, args, modifiers) -> this.plugin.getLogger());
+    }
+
+    private void noChangingRegistration(String what) {
+        checkState(!this.registered, "The %s cannot be changed after commands have been registered", what);
+    }
+
+    /**
+     * Determines if this command module can be installed or configured
+     * through a {@link CommandManager}.
+     *
+     * @param install {@code true} if we should check if this module
+     *     can be installed
+     * @return {@code true} if this command module can be installed
+     *     or configured
+     */
+    @SuppressWarnings("ConstantConditions")
+    public final boolean canBeInstalledOrConfigured(boolean install) {
+        return install ? (this.plugin == null && this.manager == null && !this.configured) : !this.configured;
+    }
+
+    /**
+     * Configure this command module.
+     *
+     * <p>You <b>must</b> {@link #register(Object) register commands} <i>after</i>
+     * you have configured any bindings, input tokenizer, etc.</p>
+     */
+    @SuppressWarnings("WeakerAccess")
+    protected abstract void configure();
+
+    /**
+     * Create a binding builder from a class type.
+     *
+     * @param clazz The class
+     * @param <T> The bound type
+     * @return A binding builder
+     */
+    protected final <T> BindingBuilder<T> bind(Class<T> clazz) {
+        this.noChangingRegistration("bindings");
+        return this.binder.bind(clazz);
+    }
+
+    /**
+     * Create a binding builder from a binding key.
+     *
+     * @param key The binding key
+     * @param <T> The bound type
+     * @return A binding builder
+     */
+    protected final <T> BindingBuilder<T> bind(BindingKey<T> key) {
+        this.noChangingRegistration("bindings");
+        return this.binder.bind(key);
+    }
+
+    /**
+     * Registers all found {@link Command annotation commands}.
+     *
+     * <p>You <b>must</b> register commands <i>after</i> you have
+     * configured any bindings, input tokenizer, etc.</p>
+     *
+     * @param object The object to register commands from
+     */
+    protected final void register(Object object) {
+        this.registered = true;
+        this.manager.register(this, object);
+    }
+
+    /**
+     * Register a given command using the given list of aliases.
+     *
+     * <p>You <b>must</b> register commands <i>after</i> you have
+     * configured any bindings, input tokenizer, etc.</p>
+     *
+     * <p>If there is a conflict with one of the aliases (i.e. that alias
+     * is already assigned to another command), then the alias will be skipped.
+     * It is possible for there to be no alias to be available out of
+     * the provided list of aliases, which would mean that the command would not
+     * be assigned to any aliases.</p>
+     *
+     * <p>The first non-conflicted alias becomes the "primary alias."</p>
+     *
+     * @param callable The command
+     * @param alias An array of aliases
+     * @return The registered command mapping, unless no aliases could be registered
+     * @throws IllegalArgumentException Thrown if {@code plugin} is not a plugin instance
+     * @see CommandManager#register(Object, CommandCallable, String...)
+     */
+    protected final Optional<CommandMapping> register(CommandCallable callable, String... alias) {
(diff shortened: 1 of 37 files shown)


ryantheleach:
What is a group?

jamierocks:
@ryantheleach `What is a group?`
A group is the base command

Zidane:
@jamierocks @kashike 

Group made me think permission. Maybe @Parent ?

jamierocks:
Yeah - that'd make sense

kinggoesgaming:
Does this support flags?

kashike:
@kinggoesgaming Example? It currently adds no additional parsing, though I'm curious to see what you have to say.

jamierocks:
I think @kinggoesgaming is referring to something like /stop -t 10s -m bye

kinggoesgaming:
Yup :)

src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
```diff
@@ -0,0 +1,365 @@
...
+     * <p>You <b>must</b> register commands <i>after</i> you have
+     * configured any bindings, input tokenizer, etc.</p>
+     *
+     * <p>If there is a conflict with one of the aliases (i.e. that alias
+     * is already assigned to another command), then the alias will be skipped.
+     * It is possible for there to be no alias to be available out of
+     * the provided list of aliases, which would mean that the command would not
+     * be assigned to any aliases.</p>
+     *
+     * <p>The first non-conflicted alias becomes the "primary alias."</p>
```
ryantheleach, line 180: So assuming there is no non-conflicting alias, is there a primary alias? or is that when Optional.absent() is returned?

What about the /pluginname:primary fallback aliases?

You mention it in the return type, but shouldn't it always be possible to register the command unless the same plugin is attempting to register the command twice?

kashike, line 180: These work the same as the `register`methods in `CommandManager`

ryantheleach, line 180: I've wondered the same thing about the register method in CommandManager as well. I'm not sure what the lesser of two evils is, duplicating JavaDocs, or having no information except a @see

src/main/java/org/spongepowered/api/command/CommandDescription.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Get a short one-line description of this command.
+     *
+     * <p>The help system may display the description in the command list.</p>
+     *
+     * @param source The source of the help request
+     * @return A description
+     */
+    Optional<? extends Text> getShortDescription(CommandSource source);
```
ryantheleach: https://github.com/SpongePowered/SpongeAPI/issues/1213

Since the interface is moving here anyway(I know it's backwards compatible, but is it binary compatible?), I'd like to once again ask what is the point to returning `Optional<? extends Text>` It prevents doing optShortDescription.orElse(Text);

`Optional<Text>` allows anything implementing/extending Text inside.

src/main/java/org/spongepowered/api/command/annotation/TranslatableDescription.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Annotates a {@link Command} method which has translatable descriptions.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface TranslatableDescription {
+
+    /**
+     * Gets the description of this command.
```
ryantheleach: From our pm's these JavaDocs should probably say translation keys.

src/main/java/org/spongepowered/api/command/CommandDescription.java
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1239

src/main/java/org/spongepowered/api/command/CommandAuthorizer.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A command authorization checker.
+ */
+public interface CommandAuthorizer extends BiPredicate<CommandSource, CommandCallable> {
+
+    /**
+     * A command authorizer which asks the {@link CommandCallable} if
+     * the {@link CommandSource} has permissions.
+     */
+    CommandAuthorizer CALLABLE = (source, callable) -> callable.testPermission(source);
```
ryantheleach, line 38: I'm not 100% sure of this naming.

src/main/java/org/spongepowered/api/command/annotation/Completion.java
```diff
@@ -0,0 +1,49 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Completion {
+
+    /**
+     * Gets the command name parts for this completer.
+     *
+     * @return The command name parts
+     */
+    String[] value();
```
ryantheleach, line 47: Do these need to 100% match the command, or is 1 enough? Should there instead be some sort of CommandKey?

The problems I'm thinking of here, is about failed alias registration + having to repeat the aliases everywhere.

src/main/java/org/spongepowered/api/command/annotation/Standalone.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a a {@link Command} method that should be
+ * registered outside of a parent.
+ *
+ * <p>This annotation should be used with {@link Parent}.</p>
```
ryantheleach, line 36: I'm still not 100% sure what this is used for. It states it needs a parent, yet I thought having a command that wasn't annotated with a parent would be enough?

Is this for cases where a class is annotated with a parent, and you need the command to not have a parent?

Or for cases where you register a command both with a parent, as well as a top level alias?

src/main/java/org/spongepowered/api/command/annotation/TranslatableDescription.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.spongepowered.api.command.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a {@link Command} method which has translatable descriptions.
```
ryantheleach, line 33: Ok, so this has translation keys, but from where are the keys looked up from?

From some plugin specific resource bundle? A resource bundle provided in the module? A Sponge global resource bundle? The base game/resource packs?

src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
```diff
@@ -0,0 +1,366 @@
...
+        this.plugin = checkNotNull(plugin, "plugin");
+        this.manager = checkNotNull(manager, "manager");
+        this.preConfigure();
+        this.configure();
+        // Create a delegating read-only binder
+        this.binder = new ReadOnlyDelegateBinder(this.binder);
+        this.configured = true;
+    }
+
+    private void preConfigure() {
```
ryantheleach, line 96: The abstract Command Module is good, but some plugins may already have a Guice module configured that could handle the commands.

Could we provide another class that uses Guice modules for the bindings?

ryantheleach:
In your example, the usage help text has the alias hard coded in the string, what happens if the first alias isn't available? the usage text will be wrong.

src/main/java/org/spongepowered/api/command/provider/Provider.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Provides an instance of {@link T}.
+     *
+     * @param source The command source
+     * @param args The command arguments
+     * @param modifiers The annotation modifiers
+     * @return An instance of {@link T}
+     * @throws ProvisionException If an instance could not be provided
+     */
+    T get(CommandSource source, CommandArgs args, Collection<? extends Annotation> modifiers) throws ProvisionException;
```
ryantheleach, line 49: These modifiers, are these annotation on the method or the parameters? Because there is also bind(Class).withAnnotations() ?

src/main/java/org/spongepowered/api/command/CommandManager.java
```diff
@@ -101,6 +102,50 @@
     Optional<CommandMapping> register(Object plugin, CommandCallable callable, List<String> aliases, Function<List<String>, List<String>> callback);
 
     /**
+     * Installs a {@link AbstractCommandModule} into this command manager.
+     *
+     * @param plugin The plugin instance
+     * @param module The command module
+     */
+    void install(Object plugin, AbstractCommandModule module);
```
ryantheleach, line 121: Is CommandManager only ever expected to be implemented by Sponge? if not, you might need a separate interface for CommandManagers that can accept an AbstractCommandModule.

src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
```diff
@@ -0,0 +1,366 @@
...
+        return install ? (this.plugin == null && this.manager == null && !this.configured) : !this.configured;
+    }
+
+    /**
+     * Configure this command module.
+     *
+     * <p>You <b>must</b> {@link #register(Object) register commands} <i>after</i>
+     * you have configured any bindings, input tokenizer, etc.</p>
+     */
+    @SuppressWarnings("WeakerAccess")
```
windy1, line 128: Ehhh you could probably just disable this inspection in your IDE instead. If we suppressed this warning consistently these would be plastered all over Sponge.

src/main/java/org/spongepowered/api/command/CommandAuthorizer.java
kashike, line 38: `DELEGATE_TO_CALLABLE`?

src/main/java/org/spongepowered/api/command/CommandManager.java
kashike, line 121: There's nothing limiting ACM's from only being used by the Sponge command manager.

src/main/java/org/spongepowered/api/command/annotation/Completion.java
kashike, line 47: Each array element is checked, does not need to match 100%

You can store an array of aliasea for use with commands.

src/main/java/org/spongepowered/api/command/annotation/Standalone.java
kashike, line 36: > Is this for cases where a class is annotated with a parent, and you need the command to not have a parent?

Yes

src/main/java/org/spongepowered/api/command/annotation/TranslatableDescription.java
kashike, line 33: See abstract command module's text function methods

src/main/java/org/spongepowered/api/command/provider/Provider.java
kashike, line 49: On the parameters

src/main/java/org/spongepowered/api/command/AbstractCommandModule.java
kashike, line 96: Guice modules don't work the same, nor is it really a good idea to use it since guice does things differently. feel free to look into it more if you want

src/main/java/org/spongepowered/api/command/CommandDescription.java
ryantheleach: This needs updating since #1239 was merged.

kashike: I'm aware. :smile_cat:

src/main/java/org/spongepowered/api/command/CommandAuthorizer.java
ryantheleach: #1080 and this should probably match depending on whether both of these get accepted.

me4502:
So this is a system where the child command defines the parent? Doesn't that do kinda the opposite of what the non-annotation based command API does?

One thing that I liked about WorldEdit's command system, is that it's really easy in the parenting system.

You create a method, annotate it with something like @CommandChildren(LowerCommands.class), and then any commands in LowerCommands.class are the child commands of the one that the method refers to.

kashike:
Yes. There's no way to do it the same way.

src/main/java/org/spongepowered/api/command/CommandException.java
```diff
@@ -75,11 +93,41 @@ public CommandException(Text message, boolean includeUsage) {
...
+
+    /**
+     * Gets if this exception is "friendly".
+     *
+     * @return If this exception is friendly.
+     */
+    public boolean isFriendly() {
+        return this.friendly;
+    }
 }
```
liach: Line

src/main/java/org/spongepowered/api/command/ImmutableCommandMapping.java
```diff
@@ -36,7 +36,7 @@
  * An immutable command mapping instance that returns the same objects that
  * this instance is constructed with.
  */
-public final class ImmutableCommandMapping implements CommandMapping {
+public class ImmutableCommandMapping implements CommandMapping {
```
liach, line 39: I don't agree with the inheritance here.

src/main/java/org/spongepowered/api/command/annotation/Parent.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Annotates a {@link Command} method which should be part of a parent.
+ *
+ * <p>This annotation can be repeated to assign multiple parents.</p>
+ *
+ * <p>When using {@link Parent}, additionally annotate with
+ * {@link Standalone} to register with no parent.</p>
+ */
+@Repeatable(Parent.Parents.class)
```
liach, line 41: Why do we still need `@Parents` since `@Parent` is already repeatable?

src/main/java/org/spongepowered/api/command/ImmutableCommandMapping.java
kashike: What don't you agree with?

src/main/java/org/spongepowered/api/command/annotation/Parent.java
kashike: This is how repeatable annotations work.
http://docs.oracle.com/javase/tutorial/java/annotations/repeating.html

liach: Oh, thanks.

src/main/java/org/spongepowered/api/command/ImmutableCommandMapping.java
liach: ImmutableCommandMapping should be final.
ImmutableModuleCommandMapping should directly extend Object.

kashike: `ImmutableModuleCommandMapping` is an `ImmutableCommandMapping` which has an associated `AbstractCommandModule`. While it is possible to implement `CommandMapping`, it would just result in a lot of duplicated code.

ryantheleach:
@kashike At the moment this is in API7 Milestone. Is it feasible for it to make it into this release (Also considering that Dual's command refactor and brigadier is planned for API8?) Or should it be scheduled for a later release?

ryantheleach:
This PR has been delayed for inclusion in Dualspires Command API or afterwards.

Zidane:

Additionally this should be in the new command framework PR.

PR #1206, opened 2016-04-28
[WIP] Add Localization API

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1206) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/677) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/618)
This is a continuation of #1088.
___
This PR adds a localization API to SpongeAPI among some other breaking changes. The `provideFirst(Class)` method has been added to ServiceManager to allow for extensible services and `getServiceManager` has been added to `PluginContainer` for internal services. In the context of localization, this allows plugins to register their own `Dictionary` and have it be received with `plugin.getServiceManager().provideFirst(Dictionary.class);` without the need to know the exact implementing class.
___
**Example usage:**
```java
@Listener
public void onGameStarted(GameStartedServerEvent event) {
    DefaultPluginDictionary dict = new DefaultPluginDictionary(this, Locales.DEFAULT, configDir);
    try {
        dict.load();
    } catch (IOException e) {
        e.printStackTrace();
    }
    Sponge.getPluginManager().getServiceManager(this).setProvider(Cause.of(this), Dictionary.class, dict);
}
```
This will create a new dictionary that will first look for the dictionary file in `<configDir>/<pluginId>.dict` and then as a resource at `<pluginId>.dict`. Plugins are of course free to implement their localization through dictionaries however they want.

**Example .dict file:**
```
en_US {
    hello="Hello"
}
en_GB {
    hello="'Ello govna'"
}
```
```java
dictionary.get("hello").get(); // "Hello"
dictionary.get("hello", Locale.UK).get(); // "'Ello govna'"

// alternatively

Text.get(this, "hello").get(); // Text{"Hello"}
Text.get(this, "hello", Locale.UK).get(); // Text{"'Ello govna'"}

// localize for player's client locale

Text.get(this, "hello", player.getLocale()).get();
```


kinggoesgaming:
@kashike question suppose a player's locale is `zh` (Chinese) and the server locale is `us_en` and the server doesnt have dictionary value for a particular key would it default to server locale?

kashike:
@kinggoesgaming yes

ewized:
I suggest that there should be a `DictionaryText` this way if you assign an item or entity the instance of `DictionaryText`, within the `DataWatcher` and/or else where it send's the client the correct translated text.

src/main/java/org/spongepowered/api/locale/DictionarySourceResolver.java
```diff
@@ -57,15 +56,14 @@ public final void setPrimary(Function<Locale, R> primary) {
...
      * given {@link Locale}.
      *
-     * @param locale Locale to supply for
      * @param function The function to supply {@link R}
      * @return This resolver
      */
-    public DictionarySourceResolver<R> add(Locale locale, Function<Locale, R> function) {
-        this.sources.put(checkNotNull(locale, "locale"), checkNotNull(function, "function"));
+    public DictionarySourceResolver<R> add(Function<Locale, R> function) {
+        this.sources.add(checkNotNull(checkNotNull(function, "function")));
```
ewized: two null checks?

ryantheleach:
How well does this correspond with formatted texts?

Is it possible to have localized texts / templates / plurals / colors / formatting?

kashike:
@ryantheleach This is just an extended `ResourceBundle`, technically. It pulls `String`s from a dictionary by `String` key. The methods added to `Text` assume there is no formatting, however you can obtain the `String`s from your dictionary manually and run the value through a `TextSerializer`.

ryantheleach:
Possible to have some way of providing a TextSerializer to a Dictionary to create a TextDictionary?

kashike:
Superseded by #1224

PR #1208, opened 2016-04-29
[WIP] Add Key bindings API

Cybermaxke:
This PR adds a key bindings api that will allow plugin developers to create custom key bindings. The api is always available, on both client and server versions, this is because the server will send packets to register the key bindings on joining the server and cleaned up afterwards, but this is a implementation detail.

I am not sure if I should have waited for the client-phase-1, but since the client isn't required for players to join and the api impl does not require any client. The impl will only send packets to the client and recieve them back.

I am also also only going to implement this in the minecraft versions 1.9+, depending on which version should be supported at that time.

The implementation will follow once the api is mostly finished/reviewed.

Example usage:
```java
private KeyBinding myTestKey;

@Listener
public void onGameInit(GameInitializationEvent event) {
    this.myTestKey = KeyBinding.builder()
            .id("testPluginId:test")
            .category(KeyCategories.GAMEPLAY)
            .build();
    Sponge.getRegistry().register(KeyBinding.class, this.myTestKey);
}

@Listener
public void onInteractKey(InteractKeyEvent event) {
    if (event.getKeyBinding() == this.myTestKey) {
        event.getTargetEntity().sendMessage(Text.of("Ow, that hurt..."));
    }
}
```

Networking draft:
Each key binding will have a internal id which will be used for networking, the default minecraft key bindings will get a inbuild internal id.

On Join:
1. Call the RegisterKeyBindingsEvent
2. Server -> Client: Register the custom key categories, includes string id, internal id and title
3. Server -> Client: Register the applicable custom key bindings, includes string id, internal id, name, category

While playing a player clicks/releases a key:
Client -> Server: Send state of the key

On logout:
Remove all the custom keys and categories.


Zidane:
Must say, impressive.

Few thoughts:

* What packet is the official client sending up to synchronize keybindings as well as when we press a key? I honestly don't remember seeing this; how does this all come together?
* I dislike having some "handler" class as it screams "You can do this with an event"

Zidane:
Continued:

* Since these events ALWAYS come from a Player object, the base KeyboardEvent could have a getter for the Player.

src/main/java/org/spongepowered/api/keyboard/KeyCategory.java
```diff
@@ -0,0 +1,87 @@
...
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets all the {@link KeyBinding}s that are added to this
+     * category.
+     *
+     * @return The bindings
+     */
+    List<KeyBinding> getBindings();
```
Zidane: CatalogTypes are inherently immutable. You should, instead, put this in the Player interface.

src/main/java/org/spongepowered/api/keyboard/KeyBindings.java
```diff
@@ -0,0 +1,179 @@
...
+        // SORTFIELDS:OFF
+
+        private Multiplayer() {}
+    }
+
+    /**
+     * All the {@link KeyBinding}s listed under the
+     * category {@link KeyCategories#STREAM}.
+     */
+    public static class Stream {
```
Zidane: Twitch support is removed in Minecraft 1.9 afaik.

src/main/java/org/spongepowered/api/keyboard/PlayerKeyBinding.java
```diff
@@ -0,0 +1,91 @@
...
+package org.spongepowered.api.keyboard;
+
+import org.spongepowered.api.entity.living.player.Player;
+
+import java.util.Optional;
+
+/**
+ * Represents a {@link KeyBinding} held by a specific {@link Player}.
+ */
+public interface PlayerKeyBinding {
```
Zidane: This interface shouldn't be needed: Do this all through events.

src/main/java/org/spongepowered/api/keyboard/KeyCategory.java
gabizou: Not all of them are immutable. We only recommend they are likely mutable shared instances, but still nonetheless mutable. Some examples are populator objects from @Deamon5550's World Gen API.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -259,4 +261,14 @@ default GameModeData getGameModeData() {
...
 
+    /**
+     * Gets the {@link PlayerKeyBinding} of this {@link Player}. This may be
+     * {@link Optional#empty()} if the player doesn't have the Sponge on the
+     * client side.
+     *
+     * @param keyBinding The key binding
+     * @return The player key binding
+     */
+    Optional<PlayerKeyBinding> getKeyBinding(KeyBinding keyBinding);
```
Zidane: Un-needed

src/main/java/org/spongepowered/api/keyboard/KeyBindings.java
gabizou: It has been.

Zidane:
Reading more, I'm confused:

```
I am not sure if I should have waited for the client-phase-1, but since the api doesn't depend on client I thought I shouldn't really wait for it.
```

Yet in your javadocs, I see it mention it needs Sponge on client?

src/main/java/org/spongepowered/api/event/keyboard/PlayerKeyInteractEvent.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @return The key interaction handlers
+     */
+    List<KeyInteractHandler> getHandlers();
+
+    /**
+     * Is fired when a {@link Player} a {@link KeyCode} pressed
+     * that is assigned to a {@link KeyBinding}.
+     */
+    interface Press extends PlayerKeyInteractEvent {
```
gabizou: Would there be some sort of ticking for a player keeping the key pressed? Versus having that information only available on the pressed event in the past.

src/main/java/org/spongepowered/api/event/keyboard/PlayerChangeKeyCodeEvent.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.api.keyboard.KeyBinding;
+import org.spongepowered.api.keyboard.KeyCode;
+
+import java.util.Optional;
+
+/**
+ * Is fired when a {@link Player} the {@link KeyCode}
+ * of a specific {@link KeyBinding} modifies.
+ */
+public interface PlayerChangeKeyCodeEvent extends PlayerKeyboardEvent {
```
gabizou: Is this even sent over the wire from client to server?

Cybermaxke:
@Zidane The client isn't required for players to connect the server, but if they join without they won't be able to retrieve the PlayerKeyBinding from a Player, so returning a empty optional.

src/main/java/org/spongepowered/api/event/keyboard/PlayerKeyInteractEvent.java
Zidane: Event should follow our standards and be named:

```PlayerInteractKeyEvent```

**Comment by gabizou** [comment]
So this is reliant on the client having Sponge installed, otherwise none of this is being received server side?

File: src/main/java/org/spongepowered/api/keyboard/PlayerKeyBinding.java (hunk 3 above)
Comment (Cybermaxke) [inline src/main/java/org/spongepowered/api/keyboard/PlayerKeyBinding.java]: @Zidane 
The main purpose of this class is to give direct access of the state of a key binding for a specific player. The `setEnabled` and `getEnabled` methods can also only be accessed through this class.
Maybe are the difference in names and purpose between PlayerKeyBinding and KeyBinding a bit confusing, but that was the only good name I could find at that time.

File: src/main/java/org/spongepowered/api/event/keyboard/PlayerKeyInteractEvent.java (hunk 5 above)
Comment (Cybermaxke) [inline src/main/java/org/spongepowered/api/event/keyboard/PlayerKeyInteractEvent.java]: @gabizou You could get the state of the binding through PlayerKeyBinding, but a Tick event could replace that if it should be removed.

**Comment by Cybermaxke** [comment]
@gabizou Yes, that is how it will work.

**Comment by Cybermaxke** [comment]
@Zidane I added some info about how I am planning to handle the client <-> server communication.

File: src/main/java/org/spongepowered/api/event/keyboard/PlayerChangeKeyCodeEvent.java (hunk 6 above)
Comment (Mumfrey) [inline src/main/java/org/spongepowered/api/event/keyboard/PlayerChangeKeyCodeEvent.java]: I don't see the utility of the server knowing what key the client has bound anyway. IMO that should be purely client-side and the server shouldn't know what handler is bound on the client side.

Comment (Zidane) [inline src/main/java/org/spongepowered/api/event/keyboard/PlayerChangeKeyCodeEvent.java]: @Mumfrey 

In the past I would have quite disagreed with you but this is a case where commands were meant for...since you can easily register a command and have the client send up that command.

I mean I do this currently with Backpack for Forge... Press B, send up /backpack.

**Comment by Mumfrey** [comment]
The `KeyCode` and `KeyCodes` stuff doesn't belong in the API in my opinion, and everything referencing it really should be removed.

There should be no expectation that the server can dictate client bindings, it should just advertise what bindings are available and let the client decide what to do with them. Keycodes have poor localisability. It doesn't provide any useful scope for (for example) jinput event piping allowing use of controller buttons and for example keyboards with macro keys (again supported by jinput).

The API's circle of influence should end at advertising key bindings and allowing keybind events to be sunk, anything further is trespassing on the user's control of their own client.

**Comment by Mumfrey** [comment]
Networking:

* Server sends available bindings on connect, consisting of
 * globally unique identifier
 * trigger id (a server-specific integer for triggering this binding, to reduce overhead)
 * binding type: momentary, stateful
* Client receives bindings, any previously known bindings (by unique id) recieve their old assignments, previously unassigned keys are left unassigned
* Client can bind keys to bindings, this is frankly none of the server's business
 * Client stores mapping of globally unique identifier in local config
 * Client remembers (in memory only) mapping of binding to int id
* For stateful bindings, client sends `keyid`+`state` to server
* For momentary bindings, client sends `keyid` to server

**Comment by Mumfrey** [comment]
> So this is reliant on the client having Sponge installed, otherwise none of this is being received server side?

Since I would assume this uses plugin channels, I could for example support it natively in liteloader as well.

**Comment by Cybermaxke** [comment]
@Mumfrey Yeah, your probably right about the `KeyCodes`, I didn't really look at it that way. And about the mementary bindings, are you referring to only send a state when it's pressed, and that the server expects that it's immediately released afterwards?

File: src/main/java/org/spongepowered/api/keyboard/KeyBinding.java [hunk 7]
```diff
@@ -0,0 +1,122 @@
...
+         */
+        Builder keyCode(@Nullable KeyCode keyCode);
+
+        /**
+         * Sets the identifier of the category.
+         *
+         * @param identifier The identifier
+         * @return This builder for chaining
+         */
+        Builder identifier(String identifier);
```
kashike: `id` to be consistent with API

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
```diff
@@ -0,0 +1,122 @@
...
+        Builder handlers(KeyInteractHandler handler, KeyInteractHandler... handlers);
+
+        /**
+         * Builds a new instanceof of a {@link KeyBinding} and registers it.
+         *
+         * @return The key binding
+         * @throws IllegalStateException If the key binding is not completed or if
+         * there already a key binding is registered with the identifier
+         */
+        KeyBinding buildAndRegister();
```
kashike: I suggest having a `build` method instead, and requiring manual registration.

Mumfrey:
> EDIT: Another thing about the There should be no expectation that the server can dictate client bindings, do you also think that setEnabled method which will hide/show to key binding on the client falls under this point?

What situation do you anticipate that having a binding appear and disappear during a session is appropriate? For a consistent and non-confusing UX it makes more sense that keybindings are always present for the duration of a session. Whether they actually *do* anything or not is another matter, but the binding should at least be present.

src/main/java/org/spongepowered/api/event/keyboard/KeyboardEvent.java
```diff
@@ -24,15 +24,31 @@
...
+import org.spongepowered.api.keyboard.PlayerKeyBinding;
 
 public interface KeyboardEvent extends Event {
 
     /**
+     * Gets the {@link Player} that caused this event.
+     *
+     * @return The player
+     */
+    Player getPlayer();
```
kashike: extend `TargetPlayerEvent` instead of `Event` instead

Cybermaxke:
@Mumfrey Developers may want to hide certain key bindings for groups of players, or depending on permissions or something. They will always be registered on the client, but using the `setEnabled` will toggle the visibility in the bindings list.

Mumfrey:
> @Mumfrey Developers may want to hide certain key bindings for groups of players, or depending on permissions or something. They will always be registered on the client, but using the setEnabled will toggle the visibility in the bindings list.

It seems like you're mixing your paradigms though. A binding either is or is not enabled for a certain player so a plugin should either register a binding globally and determine its visibility on a case-by-case basis using an event. Or else it should register bindings on a per-player basis during the player's join event.

Either way there's no utility in changing a binding after the fact, because a binding appearing and disappearing during a session is very bad UX.

src/main/java/org/spongepowered/api/event/keyboard/InteractKeyEvent.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package org.spongepowered.api.event.keyboard;
+
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.keyboard.KeyBinding;
+
+/**
+ * Is fired when a {@link Player} interacts with a {@link KeyBinding}.
+ */
+public interface InteractKeyEvent extends KeyboardEvent {
```
me4502: Could this potentially be Cancellable?

It may be useful if you wanted to make a plugin that could 'snare' players or something, so you just cancel movement key presses - still allowing other sources of movement.

That's just one use case, but I'm sure there are a heap more.

me4502:
Wait, just to get this straight..

Is this an API for making client-side plugins that have keybindings, or is it an API that the server can access, if the player has the sponge client?

Cybermaxke:
@me4502 It is an api that the server can acces.

src/main/java/org/spongepowered/api/event/keyboard/InteractKeyEvent.java
Cybermaxke: @me4502 Since this api is for the server side it won't be possible to modify the behavior of client key bindings, you can just track the changes on the server side.
Making it Cancellable may also allow one plugin to cancel another their key bindings, not sure if it would be useful.

me4502: If this was for client side plugins, I'd argue for it to be cancellable.

Cybermaxke:
I will reopen the PR for the `bleeding` branch because it's targeting 1.9

PR #1209, opened 2016-04-29
Settings

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1209) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/680)

---

This is a work-in-progress pull request, and the description will be updated once more progress has been made.

### DIFF
--- a/src/main/java/org/spongepowered/api/setting/Setting.java
+++ b/src/main/java/org/spongepowered/api/setting/Setting.java
@@ -26,12 +26,32 @@ package org.spongepowered.api.setting;
 
 import org.spongepowered.api.setting.type.SettingType;
 
+import java.util.regex.Pattern;
+
 /**
  * A setting.
  *
  * @param <T> The value type of this setting
  */
 public interface Setting<T> {
+
+    /**
+     * A pattern to validate setting ids.
+     */
+    Pattern ID_PATTERN = Pattern.compile("[a-z][a-z0-9_]*");
+
+    /**
+     * Gets the identifier for this setting.
+     *
+     * <p>Ids should be unique as to not conflict with other settings.</p>
+     *
+     * <p>The id must be lowercase, start with an alphabetic character, and
+     * contain only alphanumeric characters.</p>
+     *
+     * @return The identifier
+     * @see #ID_PATTERN
+     */
+    String getId();
 
     /**
      * Gets the setting type.
--- a/src/main/java/org/spongepowered/api/setting/Setting.java
+++ b/src/main/java/org/spongepowered/api/setting/Setting.java
@@ -24,10 +24,12 @@
  */
 package org.spongepowered.api.setting;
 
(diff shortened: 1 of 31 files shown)


liach:
What kind of settings?

src/main/java/org/spongepowered/api/setting/Setting.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.setting;
+
+public class Setting {
```
liach: Possibly change this to an interface or a catalog type

Zidane:
@kashike 

While this seems obvious, I'm having issues seeing the point of this when we have configurate. Can you go into detail?

src/main/java/org/spongepowered/api/setting/value/EnumSettingValue.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public String serialize() {
+        return this.value == null ? "" : this.value.name();
+    }
+
+    @Override
+    public Optional<T> cycleNext() {
+        T[] constants = this.enumClass.getEnumConstants();
+        if (constants.length == 0) {
```
Electroid: It does not make sense for any setting value to have no possible options, this should be a precondition in the object. That also means you can remove the `Optional` return from `CycleableSettingValue#cycleNext`. If you can cycle something, there should always be a valid setting option.

kashike: Good point, will do that now.

gabizou:
I'm somewhat confused as to the relevance of this PR. That aside, there's a lot of implementation detail that could reside in, well, the implementation of Sponge API. Additionally, making Setting not a `CatalogType` is somewhat silly if the sake of having generic type information is to be desired.

kashike:
@gabizou What do you think can go into the implementation? `SimpleSettingManager` and `SimpleSettingRegistry` can be used by plugins to construct their own registries and managers, `*SettingType` can be extended. `SimpleSetting` can be moved to the implementation if really desired.

kashike:
I'm going to close this for now - I'll revisit this soon.

PR #1210, opened 2016-04-30
Add Key bindings API

Cybermaxke:
API | [Common](https://github.com/SpongePowered/SpongeCommon/pull/682) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/661) | [LiteMod](https://github.com/Cybermaxke/LiteSpongeClient/releases/)

This PR adds a key bindings api that will allow plugin developers to create custom key bindings. The api is always available, on both client and server versions, this is because the server will send packets to register the key bindings on joining the server and cleaned up afterwards, but this is a implementation detail.

I am not sure if I should have waited for the client-phase-1, but since the client isn't required for players to join and the api impl does not require any client. The impl will only send packets to the client and recieve them back.

I am also also only going to implement this in the minecraft versions 1.9+, depending on which version should be supported at that time.

There is already a small LiteLoader mod I made for the client features, it won't work in combination with  Forge. Most of the keyboard code will be able to be reused for the Forge implementation.

Test plugin:

``` java
@Plugin(id = "keyboardtest")
public class TestKeyboardPlugin {

    @Listener
    public void onGameInit(GameInitializationEvent event) {
        final KeyBinding myTestKey = KeyBinding.builder()
                .displayName(Text.of("Test Binding"))
                .category(KeyCategories.GAMEPLAY)
                .listener(InteractKeyEvent.Press.class, e -> e.getTargetEntity().sendMessage(Text.of("Ow, that hurt...")))
                .build(this, "test");
        Sponge.getRegistry().register(KeyBinding.class, myTestKey);
        final KeyCategory testCategory = KeyCategory.builder()
                .title(Text.of("My Test Category"))
                .build(this, "test_category");
        Sponge.getRegistry().register(KeyCategory.class, testCategory);
        final KeyBinding myOtherTestKey = KeyBinding.builder()
                .displayName(Text.of("Other Test Binding"))
                .category(testCategory)
                .context(KeyContexts.IN_GAME)
                .listener(InteractKeyEvent.Press.class, e -> e.getTargetEntity().sendMessage(Text.of("And... Nothing...")))
                .listener(InteractKeyEvent.Release.class, e -> e.getTargetEntity().sendMessage(Text.of("Nope still nothing...")))
                .build(this, "other_test");
        Sponge.getRegistry().register(KeyBinding.class, myOtherTestKey);
        final KeyBinding contextTestKey = KeyBinding.builder()
                .displayName(Text.of("Context Test Binding"))
                .category(testCategory)
                .context(KeyContexts.INVENTORY)
                .listener(InteractKeyEvent.Press.class, e -> e.getTargetEntity().playSound(
                        SoundTypes.UI_BUTTON_CLICK, e.getTargetEntity().getLocation().getPosition(), 0.5))
                .build(this, "context_test");
        Sponge.getRegistry().register(KeyBinding.class, contextTestKey);
    }

    @Listener
    public void onRegisterKeyBinding(RegisterKeyBindingsEvent event) {
        System.out.println(String.format("onRegisterKeyBinding: %s keyBindings, player %s",
                event.getKeyBindings().size(), event.getTargetEntity().getName()));
    }

    @Listener(order = Order.EARLY)
    public void onInteractKeyPress(InteractKeyEvent.Press event) {
        System.out.println(String.format("onInteractKeyPrePress: %s, player %s",
                event.getKeyBinding().getId(), event.getTargetEntity().getName()));
    }
}
```


src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -259,4 +261,13 @@ default GameModeData getGameModeData() {
...
     void setSleepingIgnored(boolean sleepingIgnored);
 
+    /**
+     * Gets all the {@link KeyBinding}s that are available for this {@link Player},
+     * if the player doesn't have a client that supports custom key bindings this
+     * will always return a empty collection.
+     *
+     * @return The key bindings
+     */
+    Collection<KeyBinding> getKeyBindings();
```
me4502, line 304: Would an optional make more sense?

octylFractal, line 304: Not for a collection.

kinggoesgaming, line 304: A collection returns with a size of 0 if nothing is present :P

Hint: List

src/main/java/org/spongepowered/api/keyboard/KeyBindings.java
```diff
@@ -0,0 +1,160 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.keyboard;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * An enumeration of all possible {@link KeyBinding}s in vanilla minecraft.
+ */
+public class KeyBindings {
```
kashike: `final`

src/main/java/org/spongepowered/api/keyboard/KeyBindings.java
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * All the {@link KeyBinding}s listed under the
+     * category {@link KeyCategories#GAMEPLAY}.
+     */
+    public static class Gameplay {
```
kashike: `final`

src/main/java/org/spongepowered/api/keyboard/KeyBindings.java
```diff
@@ -0,0 +1,160 @@
...
+
+        public static final KeyBinding ATTACK = DummyObjectProvider.createFor(KeyBinding.class, "ATTACK");
+
+        public static final KeyBinding PICK_BLOCK = DummyObjectProvider.createFor(KeyBinding.class, "PICK_BLOCK");
+
+        public static final KeyBinding USE_ITEM = DummyObjectProvider.createFor(KeyBinding.class, "USE_ITEM");
+
+        // SORTFIELDS:OFF
+
+        private Gameplay() {}
```
kashike:
``` java
private Gameplay() {
}
```

src/main/java/org/spongepowered/api/event/keyboard/KeyboardEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.keyboard;
+
+import org.spongepowered.api.event.entity.living.humanoid.player.TargetPlayerEvent;
+import org.spongepowered.api.keyboard.KeyBinding;
+
+public interface KeyboardEvent extends TargetPlayerEvent {
```
kashike, line 34: Javadocs?

phroa:
I'd really enjoy an executor interface that the binding builder accepts, just a basic `.executor((KeyBinding, Player) -> void)`.

Cybermaxke:
@phroa I added executor methods for press, release and tick actions ;)

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
```diff
@@ -96,6 +98,34 @@ static Builder builder() {
...
+
+        /**
+         * Sets the tick executor of this key binding, it will be called
+         * when a specific {@link Player} the key binding holds the key
+         * and a tick occurred.
+         *
+         * @param executor The tick executor
+         * @return This builder for chaining
+         */
+        Builder tickExecutor(@Nullable BiConsumer<Player, KeyBinding> executor);
```
phroa: These are great, but why `@Nullable`?

windy1: Presumably so you can remove the executor?

Cybermaxke: @phroa They can be `null`, they will just do nothing for that action if `null` is specified.

phroa: @Cybermaxke, the `CommandSpec` builder does not accept null as an executor, but does internally default to null if you don't specify one.

Unless there's child commands, it will throw an exception if there's a null executor. In this scenario I think it would be acceptable to no-op for null executors -- any or all of the three.

phroa: Maybe I should make that last part more clear. I mean that you should remove `@Nullable` annotations, assume all user-provided executors are not null, and just do an internal check before firing an un-user-set executor.

Though I guess it's really potāto potáto. Three annotations won't kill me.

me4502: > BREAKING NEWS! Phroa found dead in his home, neighbors reported seeing three annotations flee the scene of the crime.

me4502: But yes, I agree. The methods should be there for if you're using them. If you don't want the executor, don't call the method, especially if that's the way the command executor registration works

phroa: While we're talking about this section, your javadocs are really funky. I didn't mention it cause `[WIP]` but... yeah

kashike: Builders can be re-used, and passing `null` to clear the executor is valid.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
phroa, line 304: Until right now I inferred that key binds were supported server-side like this even on vanilla, but `doesn't have a client that supports...` implies the opposite.

If only (forge?) clients support this call then I think it should be `Optional<Collection<KeyBinding>>` because otherwise it might be assumed that the client supports them but hasn't received any from the server yet.

Thoughts?

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
ST-DDT: Also having the `@Nullable` annotation shows whether you have to call  that method in order to successfully create an instance or not.

Mumfrey: Why not just add no-arg overloads with clear contract instead of having `@Nullable` params?

kashike: That'd just be a bunch of bloat.

Mumfrey: You have a weird definition of "bloat" if adding 3 overloads counts.

ryantheleach: a no-args overload doesn't really say "clear executors" to me. either have it nullable, or a separate .clear or .removeExecutor for each type.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
Cybermaxke, line 304: @phroa The list will never be empty if a player has a client side mod, then will it always contain the default key bindings, since you cannot remove them.
And there will be a forge impl once its updated to 1.9,  I also made a standalone client mod for liteloader (link can be found at the top)

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
gabizou: I'd rather have no `@Nullable` annotated parameters and simply have the `reset()` perform the cleaning. If anything, you can very well add `clearExecutors()`, but avoid adding nullable parameters.

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
```diff
@@ -0,0 +1,136 @@
...
+         * is specified then will the builder default to the category
+         * {@link KeyCategories#MISC}.
+         *
+         * @param category The category
+         * @return This builder for chaining
+         */
+        Builder category(@Nullable KeyCategory category);
+
+        /**
+         * Sets the identifier of the category.
```
ST-DDT: The identifier of the `category`?
AFAICT you are referring to the identifier of the new `key binding`.

src/main/java/org/spongepowered/api/keyboard/KeyBinding.java
kashike: `reset` resets the entire builder, passing `null` to builder methods can clear that _single_ builder value. Additionally, having to do the following will get plain annoying really fast:

``` java
@Nullable BiConsumer<Player, KeyBinding> executor = null;
....
if (executor != null) {
    builder.tickExecutor(executor);
} else {
    builder.reset();
    // re-apply all other values
    builder.tickExecutor(executor);
}
```

Cybermaxke: @kashike I think that in this case you are most likely going to use the `reset` anyway because the only setting that you want to keep is the `category`, the other ones are most likely to change, like the id and display name which should be unique and I assume that you won't reuse the same executors for multiple key bindings.

kashike: Sure, `id` and other **unique** properties should be changed, but there **can** be multiple bindings with the same executors.

Cybermaxke: There can be the same executors, but it's less likely to happen that you use the same executor.

src/main/java/org/spongepowered/api/event/keyboard/InteractKeyEvent.java
```diff
@@ -0,0 +1,134 @@
...
+
+        }
+
+        /**
+         * Is fired after the {@link KeyBinding}s press executor and
+         * if the {@link Pre} event wasn't cancelled.
+         *
+         * @see KeyBinding.Builder#pressExecutor(BiConsumer)
+         */
+        interface Post extends Press {
```
me4502: What exactly is the use case of the Post events?

Cybermaxke:
Whoops, my phone happend...

src/main/java/org/spongepowered/api/event/keyboard/InteractKeyEvent.java
Cybermaxke: Devs may want to execute a action after a specific KeyBinding pressExecutor is handled. Logging plugins may for example try to keep track of how many times are specific action is executed in a specific amount of time. I can't come up with any other use cases for now, but it's to perform actions after a executor is guaranteed to be executed.

me4502: But the main event is already called afterwards. InteractKeyEvent.Pre is called before, a InteractKeyEvent is called after.

Cybermaxke: There is not really a main event, if you would listen for `InteractKeyEvent.Press`, it would include the `Pre` and the `Post` events, so your listener would be called twice if the executor was executed.

me4502: To continue with how other events work, it should be a main event without post.

Cybermaxke:
The implementations will be updated later, I already updated the api for further discussion.

Limeth:
This exists? :O I've always wanted to make client-side mods just to have this functionality, and now I won't even need client-side code? Where can I read more about this?

ryantheleach:
Limeth sorry to burst your bubble, but this will only work if the client has Sponge also.

Mumfrey:
I will add support in liteloader too.

Limeth:
@ryantheleach Aww, I thought it was too good to be true. But still, it sounds pretty great.

liach:
I sugest moving the keybind package into the client package, as it does not exist on the server and should not exist even on the logical server.

Cybermaxke:
@liach These key bindings are created on the server, synced to the client like recipes, and the client sends when they are being used. So they shouldn't be in the client package.

liach:
Eh ok, dunno if sponge-dependent custom keybinding is feasible.

ImMorpheus:

PR #1214, opened 2016-05-01
Add the SpongeAPI dependency to the annotation processor.

JBYoshi:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/687) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/622) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/238)

Allows SpongePowered/Ore#95 to be added without giving (most) plugin developers any additional work. This uses a dependency version range, so it will work with any SpongeAPI version that's guaranteed to be compatible.

### DIFF
--- a/api-revision.gradle
+++ b/api-revision.gradle
@@ -1,0 +1,15 @@
+// This task attempts to execute 'git rev-parse --short HEAD' on the SpongeAPI
+// submodule to resolve the current API revision short hash. If successful, it
+// will append it to the SpongeAPI version included in the implementation
+// jar files.
+task('resolveApiRevision') << {
+    def process = ['git', 'rev-parse', '--short', 'HEAD'].execute(null as String[], projectDir)
+
+    def result = process.waitFor()
+    if (result == 0) {
+        version += '-' + process.inputStream.readLines().first().trim()
+    } else {
+        logger.warn('Failed to resolve API revision (Process returned error code {})', result)
+        return
+    }
+}
--- a/gradle/sponge.gradle
+++ b/gradle/sponge.gradle
@@ -50,10 +50,15 @@ test {
 
 // Set manifest entries
 jar {
+    dependsOn api.resolveApiRevision
     manifest {
         attributes(
                 'Built-By': System.properties['user.name'],
                 'Created-By': "${System.properties['java.vm.version']} (${System.properties['java.vm.vendor']})",
+
+                'Specification-Title': api.name,
+                'Specification-Version': api.version,
+                'Specification-Vendor': api.url,
 
                 'Implementation-Title': project.name,
                 'Implementation-Vendor': project.url
@@ -156,4 +161,9 @@ apply from: api.file('gradle/deploy.gradle')
 // Gradle version used for generating the Gradle wrapper
 task wrapper(type: Wrapper) {
     gradleVersion = '2.12'
+}
+
+// Append API revision to the SpongeAPI version
+api.resolveApiRevision << {
+    jar.manifest.attributes('Specification-Version': api.version)
 }
--- a/src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
+++ b/src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
@@ -30,6 +30,7 @@ import static javax.tools.Diagnostic.Kind.ERROR;
 import static javax.tools.Diagnostic.Kind.WARNING;
 import static org.spongepowered.api.plugin.Plugin.ID_PATTERN;
 
+import org.spongepowered.api.Platform;
 import org.spongepowered.api.plugin.Dependency;
 import org.spongepowered.api.plugin.Plugin;
 import org.spongepowered.plugin.meta.PluginMetadata;
@@ -42,6 +43,8 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Paths;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.annotation.processing.Messager;
 import javax.lang.model.element.TypeElement;
@@ -51,6 +54,8 @@ final class PluginElement {
     private final TypeElement element;
     private final AnnotationWrapper<Plugin> annotation;
     private final PluginMetadata metadata;
+
+    private static final Pattern API_VERSION_PATTERN = Pattern.compile("\\A((\\d+)\\.\\d+)(?:\\.\\d+)?(?:-.*)?\\Z");
 
     PluginElement(TypeElement element, AnnotationWrapper<Plugin> annotation, PluginMetadata metadata) {
         this.element = checkNotNull(element, "element");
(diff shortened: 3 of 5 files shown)


api-revision.gradle
```diff
@@ -0,0 +1,15 @@
...
+    def process = ['git', 'rev-parse', '--short', 'HEAD'].execute(null as String[], projectDir)
+
+    def result = process.waitFor()
+    if (result == 0) {
+        version += '-' + process.inputStream.readLines().first().trim()
+    } else {
+        logger.warn('Failed to resolve API revision (Process returned error code {})', result)
+        return
+    }
+}
```
kashike: Missing newline

gradle/sponge.gradle
```diff
@@ -157,3 +162,8 @@ apply from: api.file('gradle/deploy.gradle')
 task wrapper(type: Wrapper) {
     gradleVersion = '2.12'
 }
+
+// Append API revision to the SpongeAPI version
+api.resolveApiRevision << {
+    jar.manifest.attributes('Specification-Version': api.version)
+}
```
kashike: Missing newline

src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
```diff
@@ -157,13 +158,16 @@ void apply(Messager messager) {
...
 
+        boolean hasSpongeDependency = false;
         Dependency[] dependencies = this.annotation.get().dependencies();
         if (dependencies.length > 0) {
             for (Dependency dependency : dependencies) {
                 final String id = dependency.id();
                 if (id.isEmpty()) {
                     messager.printMessage(ERROR, "Dependency ID should not be empty", this.element, this.annotation.getMirror(),
                             this.annotation.getValue("dependencies"));
+                } else if (id.equalsIgnoreCase("sponge")) {
```
kashike: This should be `Platform.API_ID`, `sponge` is the name of implementation.

Minecrell:
Personally I'm not sure if plugins should fail to load on newer or older major versions, there is always the chance they will continue working if no new API features were used or only specific API parts were changed. By adding the version range like that, the plugin will refuse to be loaded on older or newer versions.

JBYoshi:
@Minecrell The plugin can still override the data by specifying the SpongeAPI dependency manually.

src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
Mumfrey: Idiomatically should also be `Platform.API_ID.equalsIgnoreCase(id)` as well.

kashike:
Perhaps output a warning if there is a version mismatch?

src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
```diff
@@ -181,6 +185,16 @@ void apply(Messager messager) {
                 this.metadata.loadAfter(new PluginMetadata.Dependency(id, dependency.version()), !dependency.optional());
             }
         }
+
+        if (!hasSpongeDependency) {
+            String version = Platform.API_VERSION.orElseThrow(
+                    () -> new IllegalStateException("Can't find SpongeAPI version"));
+            if (version.contains("-")) version = version.substring(0, version.indexOf('-'));
```
Mumfrey: Missing braces, also for the love of `<insert deity here>` just use a regex instead of this mess of ifs and splits.

JBYoshi: I don't really know how to do reflex :P

JBYoshi: *regex* Stupid autocorrect.

Mumfrey: Example:
```java
Pattern versionPattern = Pattern.compile(".*-?((\\d+)\\.(\\d+\\.)*\\d)");
Matcher versionMatcher = versionPattern.matcher(version);
if (versionMatcher.find()) {
    // This is now safe because group is guaranteed to be digits
    int major = Integer.parseInt(versionMatcher.group(2));
    String range = String.format("[%s,%d.0)", versionMatcher.group(1), ++major);
    this.metadata.loadAfter(....
}
```
Do not use this verbatim, ensure that you understand it. Consult [this site](http://www.regular-expressions.info/) until you understand regular expressions enough to know what a character class and a capturing subpattern is.

Also note the use of `String.format` instead of messy string concatenation for clarity.

Minecrell:
@kashike That's not really possible the way it is added to mcmod.info :/ We can't just make plugins handle dependency versions on a specific dependency differently.

src/main/java/org/spongepowered/api/Platform.java
```diff
@@ -27,13 +27,20 @@
...
+
+import com.google.common.base.Objects;
 
 /**
  * Represents a possible platform or implementation a {@link Game} could be
  * running on.
  */
 public interface Platform {
 
+    String API_ID = "spongeapi";
```
liach: Leave blank between fields, you'd better have javadocs.

gradle/sponge.gradle
JBYoshi: Done

api-revision.gradle
JBYoshi: Done

src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
JBYoshi: Donw

src/ap/java/org/spongepowered/plugin/processor/PluginElement.java
JBYoshi: Done. Here's the code I used for testing, along with the final regex:
```java
import java.util.regex.*;
public class RegexTest {
    private static final Pattern API_VERSION_PATTERN = Pattern.compile("((\\d+)\\.\\d+(\\.\\d+)*)");

    public static void main(String[] args) {
        Matcher matcher = API_VERSION_PATTERN.matcher("2.1-SNAPSHOT"); // this string can be changed
        if (!matcher.find()) throw new AssertionError("Invalid SpongeAPI version");
        System.out.println(String.format("[%s,%d.0)", matcher.group(1), Integer.parseInt(matcher.group(2)) + 1));
    }
}
```

src/main/java/org/spongepowered/api/Platform.java
JBYoshi: Fixed.

JBYoshi:
Status?

Minecrell:
What I said before is still the biggest problem with this, it's fine to cause warnings at runtime if loaded on a newer/older major version, but it shouldn't prevent plugins from loading like it does with your changes. There is always the chance that the plugin will continue to work anyway. This is not really something `mcmod.info` supports so I'm not sure how to solve it yet.

Minecrell:
https://github.com/SpongePowered/SpongeAPI/pull/1357 implements this in a similar way, but without forcing a specific API version, which solves the problem I mentioned in the comment above.

PR #1215, opened 2016-05-02
Add API for generating error reports

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1215) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/689)
___
This is a continuation of #793.


src/main/java/org/spongepowered/api/service/error/UserErrorException.java
```diff
@@ -0,0 +1,33 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.error;
+
+/**
+ * This is a marker interface for problems that are a result of user error. These are for example errors that are the result of an invalid
+ * configuration or a misconfigured internet connection, that a user is expected to be able to use the information from to be able to resolve issues
+ */
+public interface UserErrorException {
```
phroa: Is this supposed to be an `Exception` subclass?

Or, maybe reuse the `PEBKACException` in Sponge already?

kinggoesgaming: would `PEBKACException` even make sense here?

kinggoesgaming: UserErrorException is more understandable for the enduser imo

src/main/java/org/spongepowered/api/service/error/ErrorReport.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    void dispatchFatal() throws RuntimeException;
+
+    /**
+     * Dispatch this error to a command source. If this command source is not the console, the error report will additionally be mirrored to the
+     * console.
+     *
+     * @param source The source to send information about the error report to
+     */
+    void dispatchToCommandSource(CommandSource source);
```
phroa: Perhaps either replace this with, or add a new method for, dispatching to a `MessageReceiver`

kashike:
@phroa Since you seem to be interested in this, what are your thoughts on renaming `ErrorReport` to `CrashReport` (and related `Error` -> `Crash`)?
When being displayed `Minecraft Crash Report` is used to display and we shouldn't/can't change that really, as well as the fact that this **is** a crash, not an error (we intentionally "crash" the server).

phroa:
Shouldn't it only crash the server and spew the report if `#dispatchFatal` is called?

(If not, I'd say rename it.)

kashike:
Technically, yes, but the issue is that we refer to it as an "error report", but when it is being displayed/etc it is referred to as a "crash report".

kashike:
Another issue is reusability - `ErrorReport` and `ErrorReport.Category` **cannot** be reused due to stacktraces being included, which will be different each call more than likely.

phroa:
Yeah, I'd say rename.

gabizou:
To be honest, I've taken a liking to @Mumfrey's [`PrettyPrinter`](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/event/tracking/CauseTracker.java#L143-L154) as it's powering a lot of the new cause tracker's exception printing and debugging. If anything, it's far easier to manage than having the equivalent of `CrashReporter` from Minecraft itself.

kashike:
I find Minecraft's `CrashReport` to be more useful in a **crash** - a `PrettyPrinter` is fine for a non-crash error.

src/main/java/org/spongepowered/api/service/crash/CrashReport.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param reportable The object to append information from
+     * @return this
+     */
+    CrashReport addReportable(CrashReportable reportable);
+
+    /**
+     * Dispatch this error as a fatal error.
+     *
+     * <p>The platform will be stopped and a full crash report will be generated.</p>
```
gabizou: Isn't this object already generated? If anything this should read: "The platform will be stopped and this crash report will be printed."

src/main/java/org/spongepowered/api/service/crash/CrashReportPublisher.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.spongepowered.api.service.crash;
+
+/**
+ * A crash report publisher is responsible for publishing a {@link CrashReport}.
+ *
+ * <p>An example use is a publisher which publishes crash reports to a
+ * pastebin-like service, such as Gist or Pastie.</p>
+ */
+public interface CrashReportPublisher {
```
liach: Actually it is the same as `CrashReportable`.

kashike: No, it isn't. `CrashReportable` is responsible for populating a `CrashReport`, `CrashReportPublisher` is responsible for publishing a `CrashReport` to a pastebin-like service, such as Gist or Pastie.

20zinnm:
@kashike Did we abstract the Pastebin service?

kashike:
@20zinnm `CrashReportPublisher`

SocraticPhoenix:
Would this API allow for what I'm talking about here:
[https://forums.spongepowered.org/t/crash-handlers/12694](https://forums.spongepowered.org/t/crash-handlers/12694)

kashike:
@SocraticPhoenix What are you wanting to do on crash?

src/main/java/org/spongepowered/api/service/crash/CrashReportPublisher.java
liach: Actually they are all functionally `Consumer<CrashReport>`

ST-DDT:
I guess something like Calling an service or API that informs the admin of the crash or reset the server to a valid state after it had crashed serveral times in a row.

kashike:
@ST-DDT A `CrashReportPublisher` is able to publish reports to an external service if allowed by the server administrator. As for resetting the server, that's best handled by a server wrapper in my opinion.

SocraticPhoenix:
@kashike
I'd like to emergency-save plugin data to disk. Configuration, player data, etc. The problem with this is that it could potentially spawn it's own errors, so I'm not sure if it fits into this API.

Deamon5550:

PR #1217, opened 2016-05-02
Boss Bar API

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1217) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/692)
Supersedes #1149 
Adds #728 

Example:
example
```java
    @Listener
    public void join(final ClientConnectionEvent.Join event) {
        Player player = event.getTargetEntity();
        ServerBossBar bossBarA = ServerBossBar.builder()
            .name(Text.of(
                TextColors.DARK_AQUA, "A",
                TextColors.DARK_PURPLE, "l",
                TextColors.BLUE, "l",
                " ",
                TextColors.DARK_GREEN, "y",
                TextColors.AQUA, "o",
                TextColors.RED, "u",
                TextColors.LIGHT_PURPLE, "r",
                " ",
                TextColors.GOLD, "c",
                TextColors.DARK_AQUA, "o",
                TextColors.LIGHT_PURPLE, "l",
                TextColors.DARK_PURPLE, "o",
                TextColors.DARK_GREEN, "r",
                TextColors.RED, "s",
                " ",
                TextColors.GOLD, "s",
                TextColors.DARK_GREEN, "t",
                TextColors.DARK_PURPLE, "a",
                TextColors.DARK_AQUA, "r",
                TextColors.RED, "t",
                " ",
                TextColors.GOLD, "t",
                TextColors.DARK_PURPLE, "o",
                " ",
                TextColors.DARK_PURPLE, "b",
                TextColors.RED, "u",
                TextColors.DARK_AQUA, "r",
                TextColors.DARK_PURPLE, "n"
            ))
            .percent(0.3f)
            .color(BossBarColors.PURPLE)
            .overlay(BossBarOverlays.PROGRESS)
            .build();
        ServerBossBar bossBarB = ServerBossBar.builder()
            .name(Text.of(
                TextColors.DARK_PURPLE, "F",
                TextColors.BLUE, "i",
                TextColors.AQUA, "n",
                TextColors.DARK_GREEN, "a",
                TextColors.LIGHT_PURPLE, "l",
                TextColors.DARK_AQUA, "l",
                TextColors.YELLOW, "y",
                " ",
                TextColors.RED, "I",
                " ",
                TextColors.AQUA, "f",
                TextColors.DARK_AQUA, "e",
                TextColors.DARK_PURPLE, "e",
                TextColors.RED, "l",
                " ",
                TextColors.DARK_PURPLE, "t",
                TextColors.DARK_AQUA, "h",
                TextColors.YELLOW, "e",
                " ",
                TextColors.DARK_PURPLE, "a",
                TextColors.AQUA, "i",
                TextColors.LIGHT_PURPLE, "r",
                " ",
                TextColors.DARK_PURPLE, "I",
                " ",
                TextColors.DARK_AQUA, "b",
                TextColors.YELLOW, "r",
                TextColors.GOLD, "e",
                TextColors.LIGHT_PURPLE, "a",
                TextColors.RED, "t",
                TextColors.DARK_PURPLE, "h",
                TextColors.DARK_AQUA, "e"
            ))
            .percent(0.6f)
            .color(BossBarColors.BLUE)
            .overlay(BossBarOverlays.NOTCHED_12)
            .build();
        bossBarA.addPlayer(player);
        bossBarB.addPlayer(player);
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/api/boss/BossBarColors.java
+++ b/src/main/java/org/spongepowered/api/boss/BossBarColors.java
@@ -1,0 +1,55 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.boss;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * An enumeration of {@link BossBarColor}s.
+ */
+public final class BossBarColors {
+
+    // SORTFIELDS:ON
+
+    public static final BossBarColor BLUE = DummyObjectProvider.createFor(BossBarColor.class, "BLUE");
+
+    public static final BossBarColor GREEN = DummyObjectProvider.createFor(BossBarColor.class, "GREEN");
+
+    public static final BossBarColor PINK = DummyObjectProvider.createFor(BossBarColor.class, "PINK");
+
+    public static final BossBarColor PURPLE = DummyObjectProvider.createFor(BossBarColor.class, "PURPLE");
+
+    public static final BossBarColor RED = DummyObjectProvider.createFor(BossBarColor.class, "RED");
+
+    public static final BossBarColor WHITE = DummyObjectProvider.createFor(BossBarColor.class, "WHITE");
+
+    public static final BossBarColor YELLOW = DummyObjectProvider.createFor(BossBarColor.class, "YELLOW");
+
+    // SORTFIELDS:OFF
+
+    private BossBarColors() {
+    }
+
+}
--- a/src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
+++ b/src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
@@ -1,0 +1,66 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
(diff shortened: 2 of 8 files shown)


src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A progress bar overlay.
+     */
+    public static final BossBarOverlay PROGRESS = dummy("PROGRESS");
+
+    /**
+     * A 6 notch overlay.
+     */
+    public static final BossBarOverlay NOTCHED_6 = dummy("NOTCHED_6");
```
modwizcode: I don't think notch overlay is a good description. Something like "A boss bar divided into 6 distinct segments" would be more descriptive.

src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A 10 notch overlay.
+     */
+    public static final BossBarOverlay NOTCHED_10 = dummy("NOTCHED_10");
```
modwizcode: Same as above

src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A 12 notch overlay.
+     */
+    public static final BossBarOverlay NOTCHED_12 = dummy("NOTCHED_12");
```
modwizcode: Same as above

src/main/java/org/spongepowered/api/boss/BossBarOverlays.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A 20 notch overlay.
+     */
+    public static final BossBarOverlay NOTCHED_20 = dummy("NOTCHED_20");
```
modwizcode: Same as above

src/main/java/org/spongepowered/api/boss/ServerBossBar.java
```diff
@@ -0,0 +1,159 @@
...
+
+import java.util.Collection;
+
+/**
+ * Represents a boss bar controlled by a {@link Server}.
+ */
+public interface ServerBossBar extends BossBar {
+
+    @Override
+    ServerBossBar setName(Text name);
```
modwizcode: Couldn't we just use generics here?

kashike: Elaborate?

modwizcode: Never mind, the way I was envisioning is much more complicated, it's the recursive generics method used in data that I was thinking about

Minecrell:
:+1:

src/main/java/org/spongepowered/api/boss/BossBarColors.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static final BossBarColor YELLOW = dummy("YELLOW");
+
+    // SORTFIELDS:OFF
+
+    private BossBarColors() {
+    }
+
+    private static BossBarColor dummy(String name) {
+        return DummyObjectProvider.createFor(BossBarColor.class, name);
```
gabizou: Why create a private static method that already does the same thing?

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -259,4 +263,31 @@ default GameModeData getGameModeData() {
...
     void setSleepingIgnored(boolean sleepingIgnored);
 
+    /**
+     * Gets a collection of {@link ServerBossBar}s this player is a member of.
+     *
+     * <p>The returned collection is unmodifiable.</p>
+     *
+     * @return The boss bars
+     */
+    Collection<ServerBossBar> getBossBars();
```
gabizou: It sounds like from the methods below that it's `ServerBossBar`s that have a list of players (and thusly an `M:N` relationship), but everything goes through `ServerBossBar`. So my question I guess is, where are boss bars stored if not otherwise on players? Does the `Server` keep track of them? Does the `World`?

src/main/java/org/spongepowered/api/boss/BossBarColors.java
Zidane: Agreed. No point in this @kashike and defeats the style elsewhere.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
JBYoshi: From what I can tell, the `BossBar` stores a list of `Player`s, not the other way around. But I do like having both methods.

gabizou: @JBYoshi Yes, it does seem that way, but where can one get all available `BossBar`s?

JBYoshi: @gabizou Probably just using an `@Inject` to the underlying NMS boss bar's `addPlayer()` and `removePlayer()` (or whatever they're called).

windy1: Or 

```java
getServer().getPlayers().forEach(p -> bossBars.addAll(p.getBossBars());
```
?

JBYoshi: @windy1 

> the `BossBar` stores a list of `Player`s, not the other way around.

gabizou: Right and then the issue is having duplicate entries of `BossBar`s.

windy1: Use a `Set` then?

JBYoshi: Yeah, a `Set` could work.

ST-DDT: IMO we should keep it a `Collection` to avoid any unnecessary requirements for impls.
Similar to what we have done with [`GameRegistry#getAllOf(Class<T>)`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/GameRegistry.java#L112).

kashike: @gabizou Boss bars are not stored anywhere internally, it's up to the plugin to store/manage it.

Boss bars internally store players - Boss bars store players.

The methods on Player are simply helpers really.

There is no registry for bars, nor will be

src/main/java/org/spongepowered/api/boss/BossBarColors.java
kashike: Intent was to make it cleaner, change one method to change dummy stuff instead of changing many lines.

Will "fix" when I'm able.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
gabizou: Right, I'm not saying there should be, just saying it needs to be documented as such ;) I did poke around the internals, but I'm not really finding a reason to have a difference of `BossBar` and `ServerBossBar`. (Plus, there are entities that by default provide a `BossBar`, so getters could be added to those entities)

kashike: ServerBossBar can have players and visibility changed, BossBar can't. BossBar also will represent the client-sided Boss bar

PR #1187, opened 2016-04-16
Add new 1.9 gamerules, and add a description to all gamerules

kashike:
This adds the new 1.9 gamerules, and adds a [description](http://minecraft.gamepedia.com/Commands#gamerule) to all gamerules.

PR #1220, opened 2016-05-02
Add new 1.9 gamerules, and add a description to all gamerules

kashike:
This adds the new 1.9 gamerules, and adds a [description](http://minecraft.gamepedia.com/Commands#gamerule) to all gamerules.
Supersedes #1187

### DIFF
--- a/src/main/java/org/spongepowered/api/world/gamerule/DefaultGameRules.java
+++ b/src/main/java/org/spongepowered/api/world/gamerule/DefaultGameRules.java
@@ -24,25 +24,194 @@
  */
 package org.spongepowered.api.world.gamerule;
 
+import org.spongepowered.api.block.tileentity.CommandBlock;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.entity.living.Villager;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/gamerule/DefaultGameRules.java
```diff
@@ -24,25 +24,169 @@
...
+     */
     public static final String LOG_ADMIN_COMMANDS = "logAdminCommands";
+
+    /**
+     * Whether creepers, {@link Zombie}, {@link Enderman}, {@link Ghast},
+     * {@link Wither}, {@link EnderDragon}, {@link Rabbit}, {@link Sheep},
+     * and {@link Villager} should be able to change blocks, and whether
+     * {@link Villager} , {@link Zombie}, {@link Skeleton}, and
+     * {@link ZombiePigman} can pick up items.
+     *
```
Cybermaxke: Snow golems and silverfish can also modify blocks, and this also includes every other living entity but only for falling on farmland which will turn it into dirt.

PR #1223, opened 2016-05-04
Add methods to ChannelBuf for writing UTF strings, varints, and byte arrays

Deamon5550:
**API** // [Common](https://github.com/SpongePowered/SpongeCommon/pull/701)

Adds several new data types to ChannelBuf.

- UTF strings written in the style of `DataOutputStream`
- Varints written according to [https://developers.google.com/protocol-buffers/docs/encoding#varints](https://developers.google.com/protocol-buffers/docs/encoding#varints)
- byte arrays written as a varint with the data
- raw byte data

### DIFF
--- a/src/main/java/org/spongepowered/api/network/ChannelBuf.java
+++ b/src/main/java/org/spongepowered/api/network/ChannelBuf.java
@@ -219,12 +219,11 @@ public interface ChannelBuf {
      */
     boolean getBoolean(int index);
 
-
     /**
      * Sets the specified byte at the current writerIndex and increases
      * the writerIndex by 1 in this buffer.
      *
-     * @param data The boolean data
+     * @param data The byte data
      * @return This stream for chaining
      */
     ChannelBuf writeByte(byte data);
@@ -235,7 +234,7 @@ public interface ChannelBuf {
      * of this buffer.
      *
      * @param index The index
-     * @param data The boolean data
+     * @param data The byte data
      * @return This stream for chaining
      */
     ChannelBuf setByte(int index, byte data);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/network/ChannelBuf.java
```diff
@@ -493,22 +691,76 @@
...
+    /**
+     * Sets the string at the specified absolute index in this
+     * buffer. This method does not modify readerIndex or writerIndex
+     * of this buffer.
+     * 
+     * <p>The string will be encoded as the utf length as a short followed by
+     * the UTF-8 bytes of the string.</p>
+     *
+     * @param index The index
+     * @param data The boolean data
```
kashike: This is not a boolean.

src/main/java/org/spongepowered/api/network/ChannelBuf.java
```diff
@@ -493,22 +691,76 @@
...
     String getString(int index);
 
     /**
+     * Sets the specified string at the current writerIndex and increases the
+     * writerIndex by the length of the string.
+     * 
+     * <p>The string will be encoded as the utf length as a short followed by
+     * the UTF-8 bytes of the string.</p>
+     *
+     * @param data The boolean data
```
kashike: This is not a boolean.

src/main/java/org/spongepowered/api/network/ChannelBuf.java
```diff
@@ -493,22 +691,76 @@
...
+     * @return This stream for chaining
+     */
+    ChannelBuf writeUTF(String data);
```
ST-DDT: What is the difference between writeUTF and writeString?

Deamon5550: as documented in the javadoc the difference is the way the length is encoded, mojang uses a varint in its packets, other things using javas DataOutput/InputStream expect a short.

Deamon5550:
merged in 8af3336d834be9dbeede4af250eb69242a40f7d9

PR #1224, opened 2016-05-05
Translation dictionaries

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1224) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/703) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/625)
Supersedes #1088, #1206

---

This PR adds a translation dictionary API to SpongeAPI.


pie-flavor:
noice

src/main/java/org/spongepowered/api/text/translation/dictionary/CachingTranslationDictionary.java
```diff
@@ -0,0 +1,36 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.translation.dictionary;
+
+/**
+ * A {@link TranslationDictionary} that can cache values.
+ */
+public interface CachingTranslationDictionary extends TranslationDictionary {
```
windy1, line 30: What's the use case for something like this? Shouldn't it be irrelevant to the person getting a key from a dictionary whether the values are cached or not?

windy1:
@kashike Is this a WIP?

src/main/java/org/spongepowered/api/text/translation/dictionary/CachingTranslationDictionary.java
kashike, line 30: The point is to allow clearing of an internal translation dictionary cache. Additional resolvers can be added to `TranslationDictionary`s which have a `TranslationDictionarySourceResolver`, and this provides a way to clear that cache if changes are made after things have been loaded, without re-creating the dictionary - same applies for configuration translation dictionaries.

kashike:
@windy1 I'd call this done unless anything is brought up.

windy1:
The only thing I think this is missing is a default implementation of `SimpleConfigDictionary` that points to some uniform path using the plugin's ID and checks for the key in an asset if the config file is missing or the key is missing in the config file so plugins can just use it out of the box without having to worry about paths and they don't have to expose it to their conf dir if they don't want to. This would be nice to persuade the use of a standardized system among developers. Of course, server admins can always explicitly create the dictionary and override values themselves.

Other than that, good stuff here.

src/main/java/org/spongepowered/api/text/translation/dictionary/CachingTranslationDictionary.java
windy1, line 30: That sounds like an implementation detail to me. BTW: ResourceBundles accessed with `getBundle` are already cached internally so this isn't even necessary as far as I can tell.

src/main/java/org/spongepowered/api/text/translation/dictionary/bundle/MultiSourceResourceBundleTranslationDictionary.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Optional;
+import java.util.ResourceBundle;
+import java.util.function.Function;
+
+/**
+ * A {@link ResourceBundleTranslationDictionary} with multiple resource bundle sources.
+ *
+ * @see SimpleResourceBundleTranslationDictionary
+ */
+public class MultiSourceResourceBundleTranslationDictionary extends SimpleResourceBundleTranslationDictionary
```
windy1, line 44: This is kind of a bastardization of `ResourceBundles` IMO. `ResourceBundles` are already associated with multiple locales so mapping them to `Locale`s doesn't really make sense.

src/main/java/org/spongepowered/api/text/translation/dictionary/remote/config/MultiSourceConfigTranslationDictionary.java
```diff
@@ -0,0 +1,74 @@
...
+        this.nodes.put(locale, node);
+        this.bundles.put(locale, new ConfigResourceBundle(node));
+        return node;
+    }
+
+    @Override
+    public ConfigurationNode getNode(Locale locale) {
+        ConfigurationNode node = this.nodes.get(locale);
+        if (node == null) {
+            throw new IllegalStateException("Tried to read MultiSourceConfigTranslationDictionary before locale " + locale + " was loaded.");
```
windy1: You could use `Preconditions.checkState` here.

windy1:
Additionally, not sure if this belongs in the `text` package since it not strictly limited to text. I would recommend a root package of `locale`.

src/main/java/org/spongepowered/api/text/translation/dictionary/bundle/MultiSourceResourceBundleTranslationDictionary.java
kashike, line 44: This is a multi-source translation dictionary. There can be a primary classpath `ResourceBundle`, and plugins can add (to the resolver) extensions such as `PropertyResourceBundle`. The `Locale` is required to resolve file paths: `my/path/translations/{LOCALE}.properties` for example.

windy1, line 44: Ah okay, I see. Fair point.

src/main/java/org/spongepowered/api/text/translation/dictionary/remote/config/SimpleConfigTranslationDictionary.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public ConfigurationNode load(Locale locale) throws IOException {
+        this.root = super.load(locale);
+        return this.root;
+    }
+
+    @Override
+    public ConfigurationNode getNode(Locale locale) {
+        if (this.root == null) {
+            throw new IllegalStateException("Tried to read SimpleConfigTranslationDictionary before it was loaded.");
```
windy1: Could use `checkState` here as well :P

src/main/java/org/spongepowered/api/text/translation/dictionary/CachingTranslationDictionary.java
kashike, line 30: That is not an implementation detail. Plugins can have classpath/shaded translations which it uses as a fallback/default, and filesystem (using `PropertyResourceBundle`) translations which endusers can easily change (without modifying jars).

I'm aware that `ResourceBundle#getBundle` caches, but as I've said this it's dealing with **just** `ResourceBundle#getBundle`. Not caching would result in loading from filesystem, creating `PropertyResourceBundle`, etc each time.

windy1, line 30: No it wouldn't actually. 

https://docs.oracle.com/javase/7/docs/api/java/util/ResourceBundle.html#getBundle%28java.lang.String,%20java.util.Locale,%20java.lang.ClassLoader,%20java.util.ResourceBundle.Control%29

I think `ResourceBundle` is much better equipped to deal with the caching and we probably should just use that. You can also modify the behavior of the cache with the `control` parameter.

kashike, line 30: Do remember that this is a multi-source translation dictionary, and not all resource bundles use `ResourceBundle#getBundle`, which means we need our own internal cache. Not using a cache would result in the translation dictionary loading the properties file from the filesystem and creating a `PropertyResourceBundle` each time, which is bad. With `CachingTranslationDictionary`, the cache can be cleared if any new sources are added (plugin adds a command or something to reload it's configuration, it would want to clear the cache).

windy1, line 30: Oh true, I just realized that `SimpleResourceBundle` isn't a `CachingTranslationDictionary`

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -1010,6 +1012,31 @@ public static Text of(Object... objects) {
...
 
     /**
+     * Gets an optional {@link Text} from the provided
+     * {@link TranslationDictionary} by key.
+     *
+     * @param dictionary The translation dictionary to retrieve from
+     * @param key The translation key
+     * @return The text, if present, {@link Optional#empty()} otherwise
+     */
+    public static Optional<Text> get(TranslationDictionary dictionary, String key) {
```
windy1, line 1028: Dictionaries should be available to be used as services so you can just pass a plugin instance here. See my old PR for details.

kashike:
@windy1 Are you available on IRC right now? If not, when would you be?

windy1:
Yes I am

src/main/java/org/spongepowered/api/text/translation/dictionary/TranslationDictionaries.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Gets the translation dictionary for the provided plugin.
+     *
+     * @param plugin The plugin instance
+     * @return The translation dictionary, if present,
+     *     {@link Optional#empty()} otherwise
+     */
+    public static Optional<TranslationDictionary> plugin(Object plugin) {
+        return Sponge.getPluginManager().fromInstance(plugin)
+                .map(container -> container.getServiceManager().provide(TranslationDictionary.class).orElse(null));
```
windy1, line 48: This won't work. You need something like `provideFirst` in my previous PR. Reason being that `provide` only returns services of the exact class type.

kashike, line 48: @windy1 It works fine, just like every other service.
- `setProvider(plugin, TranscationDictionary.class, myTranslationDictionary)`
- `provide(TranscationDictionary.class)`

ryantheleach:
So from what I understand of this, instead of having a new subclass of Text, that can do the locale checking / translation when sent to the player / console / displayed, plugins instead need to manually fetch from a dictionary every time to support translation? This seems sort of awkward, especially when it relates to translatable command arguments.

Even if it's decided against having a sub-class of Text that can resolve, it would still help to have some form of "internationalized text" class that could be treated like one, then resolved to a Text at the last moment when the locale is known, rather then keeping reference to a dictionary the whole time.

windy1:
@ryantheleach There is a distinction between text that is sent as a translation and text that is translated on the server. This is the latter. Obviously we can't add new translation keys without access to the client. If text is translated on the server then there is no point of subclassing text because all you need is a reference to the key.

Instead of using plain text messages throughout your plugin you would instead get text objects through keys that would translate the text for the server's locale. So instead of

```java
player.sendMessages(Text.of("Hello, world!"));
```

you now have

```java
player.sendMessage(Text.get("hello.world"));
```

This is a very common implementation among modern software and is how we should handle it as well as far as I'm concerned.

ryantheleach:
> server's locale.

That's the problem. It needs to be the Users Locale, or it's useless.

windy1:
@ryantheleach Err that's what I meant, player locales are available on the server. An additional `sendMessage(String)` method could be added to `Player` that overloads to `sendMessage(Text)` with the player's locale.

src/main/java/org/spongepowered/api/text/translation/dictionary/TranslationDictionary.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    Optional<String> get(String key, Locale locale);
+
+    /**
+     * Gets the default {@link Locale} to be used with this dictionary.
+     *
+     * @return The default locale
+     */
+    default Locale getDefaultLocale() {
+        return Locales.DEFAULT;
```
me4502, line 69: In my opinion you should be able to override this per dictionary. Eg, if a plugin is Spanish, and the default language is Spanish, you'd want that to be the fallback as it'd be the most complete and accurate. However I'm assuming DEFAULT is system default.  What if the default is Chinese and the plugin doesn't provide a Chinese translation?

That was not meant to be a review sorry, just used to pressing in that spot to make comments

kashike: `DEFAULT` = `Locale.US`, Minecraft default

me4502: Hmm, okay. The first case is still an issue, however only if we want plugins that don't provide an English translation

Deamon5550:

kinggoesgaming:
Status of this?

liach:
Any progress on implementation?

src/main/java/org/spongepowered/api/text/translation/dictionary/bundle/MultiSourceResourceBundleTranslationDictionary.java
```diff
@@ -0,0 +1,100 @@
...
+    public ResourceBundle getBundle(Locale locale) throws MissingResourceException {
+        return this.cache.computeIfAbsent(locale, this.resolver);
+    }
+
+    @Override
+    public void clearCache() {
+        this.cache.clear();
+    }
+
+    public static class Resolver extends TranslationDictionarySourceResolver<ResourceBundle> implements Function<Locale, ResourceBundle> {
```
liach, line 79: Make this final?

DawnTheFawn1:
What's the status of this?

ryantheleach:
Heard on discord that this pr is stalled until text is reworked in a future update, moving implementation to SpongeCommon.

ryantheleach:
Closing this after a chat with Kashike. This will potentially be handled in a different way once Text is in implementation.

PR #1226, opened 2016-05-06
Update MobSpawnerData for Archetypes

ZephireNZ:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/710) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/794)

Changes `MobSpawnerData` and `Dungeon` to use `EntityArchetype`s instead of `EntitySnapshot`s.

Updates `Dungeon` to have option of a table, as explained in the Common branch.

Adds a convenience method to `EntityArchetype` for creating an archetype based solely on an `EntityType`.

Also adds an immutable/unmodifiable version of `WeightedTable` to allow for wrapping of the default list of entities to use in a Dungeon. Without this, plugins could add/remove from this list potentially causing issues down the line. (if this is desired, maybe need to add better systems for doing so?)

### DIFF
--- a/src/main/java/org/spongepowered/api/world/gen/populator/Dungeon.java
+++ b/src/main/java/org/spongepowered/api/world/gen/populator/Dungeon.java
@@ -26,17 +26,20 @@ package org.spongepowered.api.world.gen.populator;
 
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.manipulator.mutable.MobSpawnerData;
-import org.spongepowered.api.entity.EntitySnapshot;
+import org.spongepowered.api.entity.EntityArchetype;
 import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 import org.spongepowered.api.util.ResettableBuilder;
 import org.spongepowered.api.util.weighted.LootTable;
 import org.spongepowered.api.util.weighted.VariableAmount;
-import org.spongepowered.api.util.weighted.WeightedSerializableObject;
 import org.spongepowered.api.util.weighted.WeightedTable;
 import org.spongepowered.api.world.gen.Populator;
 
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
 /**
- * Represents a which places 'Dungeon's randomly underground. Each dungeon has
+ * Represents a {@link Populator} which places 'Dungeon's randomly underground. Each dungeon has
  * some associated MobSpawnerData, and data regarding the contents of any chests
  * generated within the dungeon.
  */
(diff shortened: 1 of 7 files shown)


Deamon5550:
you'll have to rebase but I've changed your base to bleeding now that schematics are merged

ZephireNZ:
Oh awesome, didn't even realise that was possible! :open_mouth:

ZephireNZ:
@gabizou Same here, all 3 PRs can be pulled at will (after thorough testing, I'm sure :wink: )

src/main/java/org/spongepowered/api/util/weighted/ImmutableProxyWeightedTable.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+/**
+ * Creates a WeightedTable that is completely immutable, but still is able to be changed via its proxy table.
+ * @param <T> Item type
+ */
+public class ImmutableProxyWeightedTable<T> extends WeightedTable<T> {
```
Zidane: Change this name to UnmodifiableWeightedTable instead.

gabizou: Well, if it's supposed to be immutable, it should copy the weighted table, otherwise, yes, I would agree that it is an unmodifiable weighted table. @Deamon5550 thoughts?

src/main/java/org/spongepowered/api/world/gen/populator/Dungeon.java
```diff
@@ -75,12 +78,50 @@ default void setAttemptsPerChunk(int attempts) {
...
-     * @return The mob spawner data
+     * <p>Gets the {@link MobSpawnerData} which represents the MobSpawner which
+     * will be created within the dungeon.</p>
+     *
+     * <p><b>Note: </b> Only one of choices or mob spawner data
+     * will be present.</p>
+     *
+     * @return The mob spawner data, if present
+     */
+    Optional<MobSpawnerData> getMobSpawnerData();
```
Zidane, line 89: Wouldn't a Dungeon always have this data?

gabizou: Dungeons by default randomly choose a random `EntityType` to spawn, so not always going to have the same choice of mob spawner data.

ZephireNZ: As explained in the docs, no. The vanilla system chooses a random String (in the case of the API, `EntityArchetype` to be more flexible) to create the data from. This is different to the MobSpawners "possible entities" as it chooses just one from the list to create. If we were to always force a MobSpawnerData to be present the API would either be inconsistent or lacking. 

So I have made two options - a weighted table of types (which defaults to the unmodifiable vanilla list) with `choices()`, or setting a static MobSpawnerData.

ZephireNZ: one will always be present at a time, as by default if neither is present in the implementation will then return the unmodifiable table representing the default choices.

src/main/java/org/spongepowered/api/util/weighted/ImmutableProxyWeightedTable.java
ZephireNZ: I think I agree - Immutable is not the best way to describe it. It's just meant to wrap around the existing table in the registry to only allow modification to the original (instead of not allowing modification at all)

src/main/java/org/spongepowered/api/util/weighted/ImmutableProxyWeightedTable.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    // PROXY METHODS
+
+    @Override
+    public VariableAmount getRolls() {
+        return this.proxy.getRolls();
+    }
+
+    @Override
```
ZephireNZ: Just realised these need to be reordered so the comments actually make sense

ZephireNZ:
The only other way i could see this being done is to instead have a
weighted table of mob spawner data

Zidane:
@Deamon5550 

Perhaps you could comment?

ZephireNZ:
@gabizou @Deamon5550 @bloodmc 

This has been stagnant for a while now - comments?

PR #1227, opened 2016-05-06
All living entities are a Projectile source

### LINKED ISSUE #1097: A projectile can not be sourced to an entity that would not fire projectiles in Vanilla
There is currently no way to set a projectile's source to a entity which does not shoot projectiles in Vanilla Minecraft such as a Zombie.

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1227) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/711)
Fixes #1097

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/projectile/ProjectileLauncher.java
+++ b/src/main/java/org/spongepowered/api/entity/projectile/ProjectileLauncher.java
@@ -22,26 +22,10 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.entity.living.monster;
-
-import org.spongepowered.api.data.key.Keys;
-import org.spongepowered.api.data.value.mutable.Value;
-import org.spongepowered.api.entity.living.Aerial;
-import org.spongepowered.api.entity.projectile.ProjectileLauncher;
+package org.spongepowered.api.entity.projectile;
 
 /**
- * Represents a Blaze.
+ * Represents a vanilla projectile launcher.
  */
-public interface Blaze extends Monster, Aerial, ProjectileLauncher {
-
-    /**
-     * Gets the current {@link Value value} for whether this {@link Blaze}
-     * is considered "aflame".
-     *
-     * @return The value for whether this blaze is aflame
-     */
-    default Value<Boolean> aflame() {
-        return getValue(Keys.IS_AFLAME).get();
-    }
-
+public interface ProjectileLauncher {
 }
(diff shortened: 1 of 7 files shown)


simon816:
I'm still not too sure about this.
To launch a projectile, all you need is a position and a rotation, which all entities have.
What makes `ProjectileSource` useful is that is marks entities that have the natural ability to launch projectiles.
In terms of #1097, I think setting the the `Source`/`ProjectileSource` name in the cause would suffice to provide information about what launched the projectile.

ST-DDT:
@simon816 You are missing the fact that [`Projectile`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/projectile/Projectile.java#L34)s 
have a method named [`getShooter() : ProjectileSource`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/projectile/Projectile.java#L43).

So we have to either change that to `Entity` (which is what I prefer) or `LivingEntity`
or the changes proposed by @kashike.
However I'm not sure whether this works that simple because AFAIK minecraft stores this value itself, so there might be casting/requirement issues.
(From 1.8.7: Its stored as `Entity` but is set via the constructor as `LivingEntityBase`)

I would prefer `Entity` because you might have some `Entity` like an `ItemFrame` lets call it `CoursedMirror` that will shoot you (with an arrow) if you look at it at night.
(And IMO its irrelevant (for the shooter information) that a plugin triggered the launch code or added the entity)

kashike:
> What makes `ProjectileSource` useful is that is marks entities that have the natural ability to launch projectiles.

Which breaks, however, for all mods - as if mods are going to depend on SpongeAPI to implement `ProjectileSource`.

The alternative to this, which I'll gladly switch this PR to, is to move the `launchProjectile` methods to `Living`.

JBYoshi:
@kashike I'd rather keep what you currently have in this PR. In order for everything to properly work, a `ProjectileSource`, and only a `ProjectileSource`, can have the ability to launch projectiles, because projectiles all have a method to get their `ProjectileSource`.

Zidane:
I've thought about this a bit and do not like the solution presented.

Spawning a projectile should care about it's source but that doesn't mean everything is a ProjectileSource (see below). That concept of launching a projectile should be separated out into the ```World``` interface (for example, see how Explosion is triggered) that takes a Cause which would contain who did it.

As for ProjectileSource, that SHOULD remain on who actually launches Projectiles in Vanilla's terms (for classification purposes). You could leave the launchXXXX methods that are forwarders to the World methods.

Also, while we are discussing this...why does launchProjectile take a class arg? How would this return Optional?

src/main/java/org/spongepowered/api/entity/projectile/ProjectileLauncher.java
```diff
@@ -0,0 +1,31 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.projectile;
+
+/**
+ * Represents a vanilla projectile launcher.
+ */
+public interface ProjectileLauncher {
```
gabizou, line 30: Is this supposed to extend `ProjectileSource` or anything? To be honest, it seems pretty vague comparing the two interfaces with the single sentence javadoc.

PR #1228, opened 2016-05-06
Add API for handedness

ZephireNZ:
*Continuation of #1121*
SpongePowered/SpongeCommon#630

This PR adds the ability for plugins to query the handedness of entities.

#### Players
Players are able to configure handedness through client settings. This is not configurable serverside, as such the implementation can get a player's handedness but will not be able to set it.

### Entities
Many living entities such as Skeletons or Zombies can have a dominant hand (Ghasts, for example, cannot). This is configurable with the `LeftHanded` NBT tag.

Implementation wise this will likely only be represented by a single boolean, with each value inversely toggling the it.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -551,6 +551,14 @@ public final class Keys {
 
     public static final Key<Value<Vector3d>> LEFT_ARM_ROTATION = KeyFactory.fake("LEFT_ARM_ROTATION");
 
+    /**
+     * Represents the {@link Key} for representing the dominant {@link HandType}
+     * of a {@link Living} entity.
+     *
+     * @see DominantHandData#dominantHand()
+     */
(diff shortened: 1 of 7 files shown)


kashike:
https://github.com/SpongePowered/SpongeAPI/pull/1121#issuecomment-204963653

ZephireNZ:
@kashike What do you think I should do with the interact events? Currently they have a bunch of subevents for each hand (left/right) and type (primary/off). Should I replace all of those with a getHand and getType? Or maybe keep both?

Keeping both wouldn't work if I wanted to add a method like setType() to change whether interaction is with primary/offhand as then the type of event would need to be changed.

kashike:
@ZephireNZ Well, it depends. What can you do with one with `getHand`/`getType`, and what can you do with separate events? What can't you do? One event makes sense to me.

kenzierocks:
Not sure if filters cover this, but you can select primary/off-hand by event type right now and only listen to one.

Issue #574, opened 2015-04-18
Allow regenerating a region


zml2008:
Revert it to how it would look if it had when the world generator is first run.

Preferably accept an arbitrary Extent (possibly a plugin-implemented proy), might have to be a Chunk tho since world generation code is rather complex.

Azatik:
1 - I used to be offered to the regeneration of the chunk. The answer was to Deamon5550: https://github.com/SpongePowered/SpongeAPI/issues/360

2 - There is a problem with regeneration. How are we going to deal with it?
![2014-12-29_18 33 14](https://cloud.githubusercontent.com/assets/10154956/7215157/6bcef84a-e5e3-11e4-816e-aef3518f2b3a.png)

zml2008:
The deleteChunk method is no longer present in World. Still needs review.

That problem happens when the world generator changes (between MC versions usually). On the same generator, the same random seed is used meaning that the same results are given.

ST-DDT:
No trees and things like that are generated randomly. And if you try to restore villages the neighbor chunks have to be used.

zml2008:
pseudo-randomly. same seed->same results

Deamon5550:
Yeah, so long as the world generator is not modified the chunk generation is deterministic.

Our options at this point is either a method to regenerate chunks, or one to delete them. Personally I'm in favour of the delete method as it can have uses beyond world regeneration like world trimming.

progwml6:
regeneration is vital in order to retroactively place things in the world that are needed for new plugins

zml2008:
My only concern with delete vs regen is that this has to work if there are players still in the chunk.

progwml6:
regen should work when players are still in the chunk

Deamon5550:
That's a good point, I guess there is no real harm in having both for the two operations eh?

zml2008:
your choice how to do it, I just need to be able to regenerate a portion of a chunk (well, a whole chunk works too) that still contains players.

Deamon5550:
While perfectly possible from an impl standpoint to re apply the generation steps to regenerate a chunk, the populators will currently not play nice with regenerating a single chunk as they populate a 16x16 area which is offset by 8 blocks by design. To allow regenerating a single chunk we're going to have to add parameters to the `Populator.populate(World,Random)` method specifying the area to populate.

[...]

PR #1231, opened 2016-05-08
Change populators to take an extent for the area to be populated

Deamon5550:
[Forge](https://github.com/SpongePowered/SpongeForge/pull/627) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/715) | Api

This makes a small change to `Populator#populate` in order to give it an extent view of the area needing population rather than relying on the plugin to offset to the correct location, this also allows much more flexibility for implementations who wish to use a different pattern for population.

A major usecase for this change is the ability to populate a non offset area to allow a single chunk to be regenerated for #574.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/gen/Populator.java
+++ b/src/main/java/org/spongepowered/api/world/gen/Populator.java
@@ -24,8 +24,8 @@
  */
 package org.spongepowered.api.world.gen;
 
-import com.flowpowered.math.vector.Vector3i;
-import org.spongepowered.api.world.Chunk;
+import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.extent.Extent;
 import org.spongepowered.api.world.gen.populator.RandomObject;
 
 import java.util.Random;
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/world/gen/Populator.java
```diff
@@ -61,59 +61,15 @@
...
-     *
-     * @param chunk The provided chunk.
+     * @param volume The volume to be populated
+     * @param world The World within which the generation in happening
      * @param random A random number generator. This random number generator is
      *        based on the world seed and the chunk position. It is shared with
      *        with other populators.
      */
-    void populate(Chunk chunk, Random random);
+    void populate(Extent volume, World world, Random random);
```
kashike: I'd say this should be `World, Extent, Random`

Zidane: @kashike 

You're not understanding the usage here. Extent in itself represents a VIEW of blocks and that is what you are given here.

Deamon5550: the world is only context 90% of the calls you cake would be to the extent, I think this ordering makes sense.

kashike: It seems reversed like this - the `Extent` is part of the `World`, this makes it sound the opposite.

PR #1244, opened 2016-05-13
Add removal and clear methods to GameProfileCache

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1244) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/720)

### DIFF
--- a/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
@@ -45,7 +45,7 @@ public interface GameProfileCache {
      *
      * @param profile The profile to cache
      * @return {@code true} if the profile was successfully cached,
-     * otherwise {@code false}
+     *     otherwise {@code false}
      */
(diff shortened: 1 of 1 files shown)


simon816:
@kashike One thing I noticed about the `GameProfileCache` was that it isn't the right place to 'lookup' profiles. A cache should simply be a store of profiles, looking them up should be done by another service which is what `GameProfileManager` is for. It would be a bleeding change but what do you think about moving them to the `GameProfileManager`?

kashike:
@simon816 `GameProfileCache` isn't just a "cache", it's a cache *and* a "loader". This allows plugins to change where everything is loaded from, since `GameProfileManager` is not a service (nor should it be, really, in my opinion).

modwizcode:
Maybe `GameProfileManager` needs to be a service then? I think these changes should be done in another PR. This PR seems good.

kashike:
I disagree, there's methods in there that can't be implemented by plugins.

PR #1247, opened 2016-05-16
Add setter for the PotionEffect in ChangeEntityPotionEffectEvent

pie-flavor:
This should be able to be modified without cancelling the event.


liach:
Please use four spaces instead of tabs.

liach:
```java
    /**
     * Gets the potion effect involved in this event, handled by previous plugins.
     *
     * @return The potion effect involved in this event
     */
    PotionEffect getModifiedPotionEffect();
	
    /**
     * Sets the new potion effect to be used in this event.
     *
     * @param effect The potion effect to be used
     */
    void setPotionEffect(PotionEffect effect);
```
These two methods should be in `Gain` sub-interface as this has no meaning to other two interfaces.

ryantheleach:
The second you start letting the effect be mutable, you should also have a getOriginalEffect.

liach:
@ryantheleach If you do this change you should retarget bleeding, because the old `getPotionEffect()V` returns the original effect.

PR #1248, opened 2016-05-16
add child(CommandCallable, Collection<String>) utility method

kinggoesgaming:
add child(CommandCallable, Collection<String>):CommandSpec.Builder to CommandSpec.Builder. This is simply a utility method.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
+++ b/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
@@ -25,16 +25,14 @@
 package org.spongepowered.api.command.spec;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 import static org.spongepowered.api.command.args.GenericArguments.firstParsing;
 import static org.spongepowered.api.command.args.GenericArguments.optional;
+import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
-import org.spongepowered.api.text.Text;
 import org.spongepowered.api.command.CommandCallable;
 import org.spongepowered.api.command.CommandException;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -163,6 +162,24 @@ public Builder child(CommandCallable child, String... aliases) {
...
+         * Add a single child command to this command.
+         *
+         * @param child The child to add.
+         * @param aliases Aliases to make the child available under. First
+         *     one is primary and is the only one guaranteed to be listed in usage
+         *     outputs.
+         *
+         * @return this
+         */
+        public Builder child(CommandCallable child, List<String> aliases) {
```
liach: Actually you can just change the method to `child(CommandCallable child, Iterable<String> aliases)`, as it could still accept `List<String>`s and do not need extra casts below.

kinggoesgaming: Now why didn't I think of that :facepalm:

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -163,6 +162,24 @@ public Builder child(CommandCallable child, String... aliases) {
...
+         * Add a single child command to this command.
+         *
+         * @param child The child to add.
+         * @param aliases Aliases to make the child available under. First
+         *     one is primary and is the only one guaranteed to be listed in usage
+         *     outputs.
+         *
+         * @return this
+         */
+        public Builder child(CommandCallable child, Iterable<String> aliases) {
```
Mumfrey: Why `Iterable`? Really ought to be `Collection`.

kinggoesgaming: I added that as to avoid the cast in c5ece94 but you are right a `Collection<String>` is a better choice.

kinggoesgaming:
Any other changes to this?

PR #1249, opened 2016-05-17
Make the potion effect in ChangeEntityPotionEffect.Gain mutable

pie-flavor:
This targets bleeding because the existing getPotionEffect() method would return the changed effect, and does not currently.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/ChangeEntityPotionEffectEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/ChangeEntityPotionEffectEvent.java
@@ -51,7 +51,22 @@ public interface ChangeEntityPotionEffectEvent extends TargetEntityEvent, Cancel
     /**
      * An event where the {@link PotionEffect} is being added.
      */
-    interface Gain extends ChangeEntityPotionEffectEvent { }
+    interface Gain extends ChangeEntityPotionEffectEvent {
+        
+        /**
+         * Gets the original potion effect involved in this event.
+         *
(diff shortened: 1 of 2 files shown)


modwizcode:
This probably doesn't need to target bleeding, nothing is being renamed or removed, so this shouldn't break existing plugins.

pie-flavor:
@modwizcode The functionality of the original getPotionEffect() method will change, and I think that functionality changing of existing methods can be considered breaking.

windy1:
Is this event implemented? If so, you'll need to update the signatures of any calls to `createChangeEntityPotionEffectEventGain`

pie-flavor:
@windy1 Doesn't look like it.

src/main/java/org/spongepowered/api/event/entity/ChangeEntityPotionEffectEvent.java
```diff
@@ -51,7 +51,22 @@
...
+        
+        /**
+         * Gets the original potion effect involved in this event.
+         *
+         * @return The original potion effect involved in this event
+         */
+        PotionEffect getOriginalPotionEffect();
+        
+        /**
+         * Sets the potion effect to be used in this event.
```
gabizou: How do you plan for if a plugin adds a duplicate `PotionEffect`, what happens?

pie-flavor: It doesn't matter; a custom value is a custom value, and it's only a single value.

Issue #1000, opened 2016-01-01
Ongoing Minor Issue List (OCD List) Beta Edition


Deamon5550:
A new and improved Minor issue list for Beta. This list will be handled a little differently in order to make it much clearer whether a reported issue has been noted and/or fixed. Minor issues reported here will be added to the main comment and the reporting comment will be deleted. Checkstyle issues should not be reported here as those can be more accurately determined by simply running the checkstyle tool. Examples of good issues to report here are misnamed methods/fields, poorly worded javadocs, minor logical errors (like that something could be using a java8 stream).

- [ ] Sink package is left over after rename to channels: https://github.com/SpongePowered/SpongeAPI/tree/master/src/main/java/org/spongepowered/api/text/sink

liach:
[SnowGolem](SpongeAPI/src/main/java/org/spongepowered/api/entity/living/golem/SnowGolem.java) should implement [ProjectileSource](SpongeAPI/src/main/java/org/spongepowered/api/entity/projectile/source/ProjectileSource.java).

JBYoshi:
`Keys.INVULNERABILITY` and `Keys.INVULNERABILITY_TICKS` - only one should be present

kenzierocks:
- [ ] [`ItemStack`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java#L52) doesn't need to extend `DataSerializable`, it already does from `DataHolder`.
- [ ] [`DataSerializable`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/DataSerializable.java#L32) `therefore`
- [ ] [`DataSerializable`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/data/DataSerializable.java#L41) `deemed`

JBYoshi:
Moved from the old issue, posted by @randombyte-developer:
- [ ] https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/EntityTypes.java#L64 shouldn't it be `ZOMBIE_PIGMAN` instead of `PIG_ZOMBIE`?

kashike:
@JBYoshi The "real" name is "PigZombie".

randombyte-developer:
[...]

PR #1252, opened 2016-05-18
Fixes JavaDoc issues in #1000; removes implements DataSerializable from ItemStack

octo-byte:
"OCD" JDoc fixes. DataHolder already extends DataSerializable.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -250,7 +250,7 @@ public final class GenericArguments {
     }
 
     /**
-     * Consumes a series of arguments. Usage is the elements concencated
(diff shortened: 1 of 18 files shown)


src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -250,7 +250,7 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
     }
 
     /**
-     * Consumes a series of arguments. Usage is the elements concated
+     * Consumes a series of arguments. Usage is the elements concencated
```
ryantheleach: concatenated or concat'ed concencated is just wrong.

octo-byte: lol wow, yep thanks

Issue #1255, opened 2016-05-25
Asset API improvement


ryantheleach:
This is a proposition to improvement on the Asset API.

At the moment the copy method on the Asset only copies to a provided file, given as a path.

I'd like to see an improved method, that creates directories if they don't exist already, as well as preserves the filename of the Asset.

e.g.
asset = container.getAsset("myConfig.conf")
asset.easyCopy(Path pathToConfigDir)

Would create the config directory if it didn't already exist, and copy using the filename if the file didn't already exist.

I believe this would offer a choice to users who do not want to use the default as a fallback, in cases where using it as a fallback makes it impossible to remove configuration nodes in the config files.

Zidane:
Closing this, feedback should go on  PR.

PR #1256, opened 2016-05-25
add Asset::copyToDirectory and Asset::getFileName

ryantheleach:
My attempt at fixing https://github.com/SpongePowered/SpongeAPI/issues/1255

Add Asset::copyToDirectory that copies the asset (using it's filename ignoring path) to the provided directory.

Add Asset::getFileName to get the filename portion of the url.

### DIFF
--- a/src/main/java/org/spongepowered/api/asset/Asset.java
+++ b/src/main/java/org/spongepowered/api/asset/Asset.java
@@ -77,6 +77,33 @@ public interface Asset {
     }
 
     /**
+     * Copies this Asset to the specified 'outputDirectory' {@link Path}.
+     *
+     * @param outputDirectory The directory to copy to
+     * @throws IOException
+     */
+    default void copyToDirectory(Path outputDirectory) throws IOException {
+        checkNotNull(outputDirectory, "outputDirectory");
+        copyToFile(outputDirectory.resolve(this.getFileName()));
+    }
+
+    /**
+     * Returns the the last portion of the Asset URL, e.g. the file name.
+     *
+     * @return The file name
+     */
+    default String getFileName() {
+        String path = getUrl().getPath();
+        //We don't need to worry about file system specific file separators as we are dealing with a substring of URL
+        int end = path.lastIndexOf('/');
+        if (end < 0) {
+            return path;
+        } else {
+            return path.substring(end + 1);
+        }
+    }
+
+    /**
      * Reads this Asset in it's entirety as a {@link String} and returns the
      * result.
      *


ryantheleach:
Another addition that I would have liked to make, would be a way of getting the path that corresponds to the "directory" of the asset. e.g. everything between assets\pluginid\ and the filename.

What would be the cleanest way to add this?

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
...
+        checkNotNull(outputDirectory, outputDirectory);
+        copyToFile(outputDirectory.resolve(this.getFileName()));
+    }
+
+    /**
+     *
+     * @return get's the last portion of the Asset URL, e.g. the file name
+     */
+    default String getFileName(){
+        return new File(getUrl().getPath()).getName();
```
Minecrell: I don't really see a need for constructing a `File` object with the path to an URL here, just split it yourself

ryantheleach: Come up with a reliable method of splitting it that works for all input then? I couldn't work one out.

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
     }
 
     /**
+     * Copes this Asset to the specified 'outputDirectory' {@Link Path}.
+     * @param outputDirectory directory to copy to
+     * @throws IOException
+     */
+    default void copyToDirectory(Path outputDirectory) throws IOException {
+        checkNotNull(outputDirectory, outputDirectory);
```
kashike: `checkNotNull(outputDirectory, "output directory');`

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
...
+        copyToFile(outputDirectory.resolve(this.getFileName()));
+    }
+
+    /**
+     *
```
kashike: Incomplete javadocs

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
...
+     * @return get's the last portion of the Asset URL, e.g. the file name
```
kashike: Capital `G`

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
     }
 
     /**
+     * Copes this Asset to the specified 'outputDirectory' {@Link Path}.
+     * @param outputDirectory directory to copy to
```
kashike: `@param outputDirectory The directory to copy to`

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +78,24 @@ default void copyToFile(Path output) throws IOException {
     }
 
     /**
+     * Copes this Asset to the specified 'outputDirectory' {@Link Path}.
```
kashike: New line before `@param`s.

src/main/java/org/spongepowered/api/asset/Asset.java
kinggoesgaming: Also its Gets not Get's.

src/main/java/org/spongepowered/api/asset/Asset.java
kinggoesgaming: Copies not Copes

Zidane:
@ryantheleach 

Can you fix the above comments? Wanna merge this in.

src/main/java/org/spongepowered/api/asset/Asset.java
kinggoesgaming: Copes-> Copies

ryantheleach:
I'll fix it up in ~ 15 hours.

src/main/java/org/spongepowered/api/asset/Asset.java
ryantheleach: The original reasoning was the potential that an asset could contain file separators at the end of the Url, (I don't think it's applicable, but for instance WebAssets with query strings) But I think that would fail if I used a File or Path anyway, so I will split it, since it's in a default method if a strange implementation of Asset needs it it can override the functionality anyway.

I don't expect that standard assets will cause the bug to trigger.

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
...
+    /**
+     * Returns the the last portion of the Asset URL, e.g. the file name.
+     *
+     * @return The file name.
+     */
+    default String getFileName(){
+        String path = getUrl().getPath();
+        //We don't need to worry about file system specific file separators as we are dealing with a substring of URL
+        int end = path.lastIndexOf('/');
+        if(end < 0) {
```
Minecrell: space after if

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
...
 
     /**
+     * Copies this Asset to the specified 'outputDirectory' {@Link Path}.
+     *
+     * @param outputDirectory The directory to copy to
+     * @throws IOException
+     */
+    default void copyToDirectory(Path outputDirectory) throws IOException {
+        checkNotNull(outputDirectory, "outputDirectory");
+        copyToFile(outputDirectory.resolve(this.getFileName()));
```
ryantheleach: Should an asset have a concept of a sub-directory? or should that be delegated to plugins to use if they need anything further?

In my mind copyToDirectory would be used to copy default configs + examples to the correct directories inside the plugin folders.

But if they need sub-directories that falls apart somewhat, as assets currently only know about their own filenames, and not whether they are in a sub-directory. So plugins would be responsible for discovering their assets, and setting the appropriate sub-directory. I guess that isn't so bad, but would appreciate any input others might have.

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
...
+    }
+
+    /**
+     * Returns the the last portion of the Asset URL, e.g. the file name.
+     *
+     * @return The file name.
+     */
+    default String getFileName(){
```
Minecrell: Missing space before `{`

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
     }
 
     /**
+     * Copies this Asset to the specified 'outputDirectory' {@Link Path}.
```
Minecrell: `{@link` not `{@Link`

gabizou:
@ryantheleach I'm willing to bet that this can be merged in for API 5.0.0, but I'll leave it to @Zidane to decide.

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
...
+     */
+    default void copyToDirectory(Path outputDirectory) throws IOException {
+        checkNotNull(outputDirectory, "outputDirectory");
+        copyToFile(outputDirectory.resolve(this.getFileName()));
+    }
+
+    /**
+     * Returns the the last portion of the Asset URL, e.g. the file name.
+     *
+     * @return The file name.
```
Minecrell: There should be no period here

src/main/java/org/spongepowered/api/asset/Asset.java
```diff
@@ -77,6 +77,33 @@ default void copyToFile(Path output) throws IOException {
...
+        checkNotNull(outputDirectory, "outputDirectory");
+        copyToFile(outputDirectory.resolve(this.getFileName()));
+    }
+
+    /**
+     * Returns the the last portion of the Asset URL, e.g. the file name.
+     *
+     * @return The file name
+     */
+    default String getFileName() {
```
windy1: Could probably be shortened to 

```java
Paths.get(getUrl().getPath()).getFileName();
```

Minecrell: Since assets are not specific to a file system (that is directly supported with `Path`) there is no need to touch a filesystem-specific API here in my opinion.

src/main/java/org/spongepowered/api/asset/Asset.java
Minecrell: In my opinion, if I call `copyToDirectory` the asset should be copied to the specified directory, and not to a sub-directory of it. The other functionality can be still added later or implemented separately in plugins.

For example we could add a separate `getPath()` method that returns a string with the relative path from the asset directory to the asset itself, e.g. `plugin.jar!/assets/plugin/textures/test.png` would be `textures/test.png`

ryantheleach: Ok sure, I figured if plugins wanted to change the behaviour they could always extend copyToDirectory anyway.

PR #1258, opened 2016-05-27
Custom Inventories API Refactor

Faithcaio:
[API] (https://github.com/SpongePowered/SpongeAPI/pull/1258) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/752)
___

!! Current Code is now in main repository under ```feature/custom-inventory``` !! so don't bother commenting on code here.

Custom Inventories!

Example showing a *virtual* chest to a Player:

```java
    Inventory inv = Inventory.builder().of(Archetypes.CHEST).build();
    player.openInventory(inv, Cause.of(NamedCause.source(player)));
```


src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,31 @@
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(InventoryTypes.class)
+public interface InventoryType extends CatalogType
+{
```
kashike: Formatting throughout this class is wrong.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,31 @@
+package org.spongepowered.api.item.inventory.custom;
```
kashike: Missing license header

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryTypes.java
```diff
@@ -0,0 +1,22 @@
+package org.spongepowered.api.item.inventory.custom;
```
kashike: Missing license header

simon816:
Not sure about `InventoryType`, it seems to conflict with the Window API I have proposed https://github.com/SpongePowered/SpongeAPI/pull/1189
IMO Inventories should be a representation of the underlying storage of the itemstacks and maybe the `Container` interface is a better target for specific inventory types.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -54,7 +55,7 @@ static Builder builder() {
          * @param name The title
          * @return fluent pattern
          */
-        Builder name(Translation name);
+        Builder name(String name);
```
simon816: I think this could be a `Text` object if that's supported.

Faithcaio: InventoryBasic wants a String. I am not sure how to convert the Text without loss.
It gets recreated using:
`this.hasCustomName() ? new ChatComponentText(this.getName()) : new ChatComponentTranslation(this.getName(), new Object[0]);`

simon816: Find a way to change that method to return the `IChatComponent` from the `Text` object (`SpongeTexts.toComponent`). I assume the protocol allows for sending a text object.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,50 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(InventoryTypes.class)
+public interface InventoryType extends CatalogType {
```
kashike: Missing new line after this one

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,50 @@
...
+    class Block implements InventoryType {
```
kashike: Missing new line after this one

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryTypes.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.block.BlockTypes;
+
+/**
+ * http://wiki.vg/Inventory
+ */
+public class InventoryTypes {
+    public static final InventoryType CHEST = new InventoryType.Block(BlockTypes.CHEST);
```
kashike: Missing new line after this one

Faithcaio:
@simon816 so `InventoryType` beeing a sort of wrapper around `Container.class`?
similar to EntityType

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.item.inventory.Container;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(InventoryTypes.class)
+public interface InventoryType extends CatalogType {
+
+    class Block implements InventoryType {
```
Zidane: Never implement a CatalogType in the API. This is why we have dummy capability.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryTypes.java
```diff
@@ -0,0 +1,46 @@
...
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.block.BlockTypes;
+
+/**
+ * http://wiki.vg/Inventory
+ */
+public class InventoryTypes {
+
+    public static final InventoryType CHEST = new InventoryType.Block(BlockTypes.CHEST);
```
Zidane: This needs to use the dummy system.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
gabizou: This should be an interface.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -25,16 +25,18 @@
...
 import org.spongepowered.api.util.ResettableBuilder;
 
 /**
  * A CustomInventory is an inventory that has no backing in the world that is
  * created by plugins.
  */
-public interface CustomInventory extends OrderedInventory, Identifiable {
+public interface CustomInventory extends Inventory, Identifiable {
+
+    InventoryType getType();
```
kashike: Indentation is wrong.

kashike: Missing javadocs.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -64,6 +66,8 @@ static Builder builder() {
          */
         Builder size(int size);
 
+        Builder type(InventoryType type);
```
kashike: Javadocs

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -54,7 +56,7 @@ static Builder builder() {
          * @param name The title
          * @return fluent pattern
          */
-        Builder name(Translation name);
+        Builder name(Text name);
```
kashike: Should this be `title` instead? (method name, and argument name)

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryType.java
```diff
@@ -0,0 +1,60 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.item.inventory.Container;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(InventoryTypes.class)
```
kashike: Javadocs?

Zidane:
This API has some problems. Namely how you are representing CustomInventory.

While I'm aware the interface is here, the Vanilla container types should actually represent what they are. A chest, for example, is an Inventory2D + Title. A Player Inventory is one Inventory2D + Armor Inventory (slots) + Hotbar (slots). It is important that the Vanilla container types that end up becoming Inventories represent a composition of inventories...not just a CustomInventory with an arbitrary type. The very existence of CustomInventory, imo, goes against the Inventory API's objective OF being about composition.

Perhaps @Mumfrey may agree or debate me on this but I find that we need to provide the Vanilla types to the API, perhaps as some form of Archetype, in-which construct the particular Vanilla type with the multitude of Inventories as you would expect

Zidane:
Thinking further, I believe your intention here is to figure out which container to use (since Vanilla's system is hella bad). This is where the archetype idea is critically important. Basically you wouldn't call a CustomInventory.Builder, instead you would do ChestInventory.Builder or something like this in-which not only construct the Container but the Inventory objects....basically the whole construct in one go.

Zidane:
@Faithcaio 

So, a step forward which is good. You should move the Builder methods out of CustomInventory and into the types. So the ChestInventory Builder should be in the ChestInventory interface as well as the static builder() method should live in the ChestInventory interface.

That way, I can literally call ChestInventory.builder().blah().build()

Zidane:
Also, it shouldn't be the goal of this API to do any persistence. Using events, persistence of this inventory should be entirely up to the plugin.

Mumfrey:
> Perhaps @Mumfrey may agree or debate me on this but I find that we need to provide the Vanilla types to the API, perhaps as some form of Archetype, in-which construct the particular Vanilla type with the multitude of Inventories as you would expect

@Zidane exactly, this PR is kind of moving in the opposite direction to the one we should be going, since it actively reduces abstraction rather than promoting it.

Mumfrey:
Github screwed me over because I accidentally clicked off the window. But the gist of the tail of the message which I lost was "adding vanilla types into the api is not the right approach, rather - like Inventory API does - we should encourage the consumer to specify *intent* and have the implementation decide how that translates into reality. The current API doesn't make provision for non-inventory-like inventories either: eg. "menu" inventories or other intents, and this is something I very much wanted to revise for API v5. I feel like the custom inventory API should be brought closer to the Inventory API, not taken further away from it."

Zidane:
@Mumfrey 

I think his latest changes are taking a step in the right direction. Especially when he moves the Builder interfaces where they should go. Unless you had a different idea in mind.

Faithcaio:
How would you specify the *intent* if not with the Type of Inventory you want to show?

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -25,52 +25,212 @@
...
+        Builder title(Text title);
+
+        /**
+         * Builds the inventory instance.
+         *
+         * @return A new custom inventory
+         */
+        T build();
+
+        interface Merchant extends Builder<CustomInventory.Merchant> {
```
Zidane: Get rid of these altogether from CustomInventory.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -25,52 +25,212 @@
...
+import org.spongepowered.api.text.Text;
 import org.spongepowered.api.util.Identifiable;
 import org.spongepowered.api.util.ResettableBuilder;
 
 /**
  * A CustomInventory is an inventory that has no backing in the world that is
  * created by plugins.
  */
-public interface CustomInventory extends OrderedInventory, Identifiable {
+public interface CustomInventory extends Inventory, Identifiable {
```
Zidane: You could literally rename CustomInventory to TitledInventory or something like that lol.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventory.java
```diff
@@ -25,52 +25,212 @@
...
     /**
-     * Creates a new {@link Builder} to build a {@link CustomInventory}.
-     *
-     * @return The new builder
+     * A Custom MerchantInventory
+     * TODO needs a Merchant
      */
-    static Builder builder() {
-        return Sponge.getRegistry().createBuilder(Builder.class);
+    interface Merchant extends CustomInventory { // , MerchantInventory
```
Zidane: These should go into their own classes.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+public interface CustomInventoryBuilder<T extends Inventory> extends ResettableBuilder<T, CustomInventoryBuilder<T>> {
+
+    /**
+     * Sets the title of the custom inventory, viewable by players looking
+     * at the inventory.
+     *
+     * @param title The title
+     * @return fluent pattern
+     */
+    CustomInventoryBuilder title(Text title);
```
gabizou: If the class is generified, this return type needs to be generified as well.

src/main/java/org/spongepowered/api/item/inventory/type/DispenserInventory.java
```diff
@@ -0,0 +1,57 @@
...
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * A vanilla DispenserInventory with a 3x3 Grid
+ */
+public interface DispenserInventory extends GridInventory {
+
+    @Override
+    default int getColumns() {
```
gabizou: Don't leak these sort of implementation details. That's precisely why we have the inventory API to begin with.

src/main/java/org/spongepowered/api/item/inventory/crafting/CraftingInventory.java
```diff
@@ -55,4 +57,17 @@
      */
     Optional<Recipe> getRecipe();
 
+
+    interface Builder extends CustomInventoryBuilder<CraftingInventory> {
```
kashike: This should be at the bottom of the parent class.

src/main/java/org/spongepowered/api/item/inventory/type/HopperInventory.java
```diff
@@ -0,0 +1,57 @@
...
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * A Custom HopperInventory.
+ * 5 Slots
+ */
+public interface HopperInventory extends GridInventory {
+
+
+    @Override default int getRows() {
```
kashike: Method annotations on their own line

src/main/java/org/spongepowered/api/item/inventory/crafting/CraftingInventory.java
```diff
@@ -55,4 +57,17 @@
      */
     Optional<Recipe> getRecipe();
 
+
+    interface Builder extends CustomInventoryBuilder<CraftingInventory, Builder> {
```
kashike: This should be at the bottom of the parent class.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -25,22 +25,14 @@
...
-import org.spongepowered.api.item.inventory.type.PlayerInventory;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.util.ResettableBuilder;
 
 /**
  * A Builder for inventories that have no backing in the world that are created by plugins.
  */
-public interface CustomInventoryBuilder<T extends Inventory> extends ResettableBuilder<T, CustomInventoryBuilder<T>> {
+public interface CustomInventoryBuilder<T extends Inventory, B extends CustomInventoryBuilder<T, B>>
+        extends ResettableBuilder<T, CustomInventoryBuilder<T, B>> {
```
gabizou: Your second generic would be `B`, not `CustomInventoryBuilder<T, B>`.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+package org.spongepowered.api.item.inventory.custom;
+
+import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A Builder for inventories that have no backing in the world that are created by plugins.
+ */
+public interface CustomInventoryBuilder<T extends Inventory, B extends CustomInventoryBuilder<T, B>>
```
kashike: `T` could be named `I` for "Inventory".

src/main/java/org/spongepowered/api/item/inventory/type/ChestInventory.java
```diff
@@ -22,55 +22,33 @@
...
 
     /**
-     * Creates a new {@link Builder} to build a {@link CustomInventory}.
+     * Creates a new {@link Builder} to build a {@link ChestInventory}.
      *
      * @return The new builder
      */
     static Builder builder() {
-        return Sponge.getRegistry().createBuilder(Builder.class);
+        return Sponge.getRegistry().createBuilder(ChestInventory.Builder.class);
```
kashike: This change shouldn't be required.

src/main/java/org/spongepowered/api/item/inventory/type/AnvilInventory.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.spongepowered.api.item.inventory.type;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * A Custom AnvilInventory
+ * Container-Inventory: 2 Input - 1 Output Slot
```
kashike: Proper javadocs please

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+ * A Builder for inventories that have no backing in the world that are created by plugins.
+ */
+public interface CustomInventoryBuilder<I extends Inventory, B extends CustomInventoryBuilder<I, B>> extends ResettableBuilder<I, B> {
+
+    /**
+     * Sets the title of the custom inventory, viewable by players looking
+     * at the inventory.
+     *
+     * @param title The title
+     * @return fluent pattern
```
kashike: Capital `F`.

src/main/java/org/spongepowered/api/item/inventory/type/BeaconInventory.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * A Custom BrewingStandInventory
+ * Container-Inventory ? 1 Slot
+ */
+public interface BeaconInventory extends Inventory {
+
+    /**
+     * Creates a new {@link CustomInventoryBuilder} to build a {@link BeaconInventory}.
+     *
+     * @return The new builder
```
kashike: I don't think "new" is necessary here.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    B title(Text title);
+
+    /**
+     * Builds the inventory instance.
+     *
+     * @return A new custom inventory
```
kashike: I don't think "new" is necessary here.

src/main/java/org/spongepowered/api/item/inventory/type/ChestInventory.java
```diff
@@ -22,55 +22,33 @@
...
-
-        /**
-         * Builds the inventory instance.
-         *
-         * @return A new custom inventory
-         */
-        CustomInventory build();
-
-        // TODO add persistence methods
+        ChestInventory.Builder rows(int size);
```
kashike: Why rename this?

kinggoesgaming: tbh rows is more understandable as in a chest GUI the number changes the number of rows that are shown.

src/main/java/org/spongepowered/api/item/inventory/type/AnvilInventory.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * Represents the inventory of a vanilla Anvil.
+ * Contains 2 Input and 1 Output Slots.
+ */
+public interface AnvilInventory extends Inventory {
```
Cybermaxke: This should extend OrderedInventory.

src/main/java/org/spongepowered/api/item/inventory/type/EnchantmentTableInventory.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * Represents the Inventory of a vanilla EnchantmentTable.
+ * Contains one Input-Slot for the enchantment item and one Input-Slot for Lapis
+ */
+public interface EnchantmentTableInventory extends Inventory {
```
Cybermaxke: This should extend OrderedInventory.

src/main/java/org/spongepowered/api/item/inventory/type/FurnaceInventory.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.item.inventory.custom.CustomInventoryBuilder;
+
+/**
+ * Represents the Inventory of a vanilla Furnace.
+ * Contains one InputSlot, one FuelInputSlot and one OutputSlot
+ */
+public interface FurnaceInventory extends Inventory {
```
Cybermaxke: This should extend OrderedInventory.

Zidane:
Talked extensively with @Mumfrey on this and we both agree that the direction isn't where we wanna go with this. Instead, we want to take Archetype approach. I'll let him elaborate further.

src/main/java/org/spongepowered/api/item/inventory/type/ChestInventory.java
kashike: And who's to say the number of columns will always be the same?

Faithcaio: Vanilla? But this might all go away anyways.

Mumfrey:
Elaborating further: you can probably tell from looking at a lot of the API that the approach we tend to favour when designing API features is to encode hard *concepts* in the shape of the API, but deliberately separate specifics into catalogued types (aka pseudo-enums) and similar constructs. Accordingly, one of the first changes I made when picking up the task of designing of the Inventory API in the first place was to actually *remove* a lot of the interfaces you're proposing to add here from the previous proposal. The reasons are myriad, but the main ones are:

* Expressiveness of consumer code: encouraging consumers to express *intent* rather than actual detail means that if implementation or the entire shape of the API changes in the future, it is still possible for the implementation to inflect some meaning and adapt it to the new situation. This leads to
* Better future-proof code - I mean, we're talking an increment from "little" to "some", but it's still an improvement
* Flexibility for different implementations which may have to do things differently
* Encoding the very idea of "might not exist" into the fabric of the API also allows for graceful degredation in the future if a feature goes away

Not going to harp on about this too much as it's all been expressed in depth elsewhere. If you want to understand my reasoning behind the Inventory API, go read the original PR.

Anyway, I was planning to introduce a similar "intent"-focused system into custom inventories for API v5, however speaking with @Zidane his reasoning behind making the system Archetype-based (effectively the same thing) is a better term for it, since it already aligns with other aspects of the game design, in this case world archetypes.

> How would you specify the intent if not with the Type of Inventory you want to show?

The answer is, you **do** specify the type of inventory you want to show, but the contract ends there. You declare your requirement, the implementation is free to do what it wants from that point onwards. Consumers should not expect a particular type of inventory back. Why? because nowehere else in the Inventory API is this contract encoded: All `Carrier`s return a plain `Inventory`! Making the Custom Inventory API different doesn't make a lot of sense, partly because it adds a lot of unnecessary cruft - which is actually pretty nicely demonstrated by your example listener code above, which elucidates *precisely* the type of crufty consumer code we're trying to avoid - and partly because it's not good practice if we want to maintain the general shape of the API being concept-hard, detail-soft.

*How does this work in practice?*

Well, the various interfaces you have outlined go away, and are replaced with a catalogued selection of *Archetypes*, which can be used to describe the type of inventory you're wishing to build, provided to the factory in very much the same way you would specify a list of ingredients for a recipe.

Let's assume we have things such as:

* `ArchetypeChest`
* `ArchtetypeWorkbench`
 and sprinkled in with them we have non-vanilla intents which represent other archetypal usage of inventory windows such as virtual menus and such, eg.
* `ArchetypeRowMenu`
* `ArchetypeGridMenu`
* `...`

We create concrete inventories thus by offering an Archetype to a Factory:

Archetypes 1

Further, we also provision for consumable `properties` which provide the customisation capabilities which can be `offer`ed to (consumed by) the Archetypes when they pass through the Factory.

* `DimensionProperty`
* `TitleProperty`
* `...`

Archetypes 2

Since archetypes themselves are a catalogued type, this means they can be retrieved by name (bye bye horrible switch statements) and can also be easily expanded upon by mods or other plugins.

Changing from the idea of singular types into the concept of Archetypes also allows us to employ *composition* when employing archetypes. For example it should be clear that different types of menu archetype can co-exist, or maybe a partial player inventory view and part of a chest inventory view can be combined and the implementation figures out the details, maybe a mod provides modular inventories which can be specified and combined. The point is, by supporting composition we can cover all of those use-cases as well:

Archetypes 3

Plugins could potentially even register these composite archetypes to allow them to be easily consumed by other plugins.

*How does it look in code*

Realistically, the `Archetypes` catalogue will contain all the available archetypes as a regular catalogue, but could also potentially contain the factory capability for making composite archetypes. Ideally properties should be able to be specified either as part of the composite or at the consumer level, this would allow some pre-configuration of Archetypes as they are added to the composite but also configuration of properties at instance time.

**Construction of a simple archetype**
```java
import Archetypes.Properties;

Inventory custom = CustomInventory.builder()
    .of(Archetypes.CHEST)
    .withProperty(Properties.DOUBLE_CHEST)
    .build();
```

**Constructing and using a composite archetype**
```java
import Archetypes.Properties.Delegate;
import Archetypes.Properties.Size;

Archetype myMenu = Archetypes.factory()
    .named("translation.key")
    .of(Archetypes.MENU_ROW)
    .of(Archetypes.MENU_GRID).with(new Delegate("size", Size.class)) // delegated property
    .build();

Inventory menuInv = CustomInventory().builder()
    .of(myMenu)
    .withNamedProperty("size", new Size(9, 2))
    .build();
```

The one remaining thing which needs to be present for convenience is some kind of `Entity` -> `Archetype` mapping which can be used to achieve a "give me the screen for `x`" capability in a simple manner. In particular this should either take the form of

```java
Archetype archetype = Archetype.from(someEntity); // or maybe Carrier
```

or possibly a simple overload on the custom inventory factory:

```java
Inventory inv = CustomInventory.builder().for(someCarrier).build()
```

but I leave that up to the imagination of the readership.

I'm going to stop typing now because my fingers hurt.

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Sets the title of the custom inventory, viewable by players looking
+     * at the inventory.
+     * TODO may not work with all Inventories -- investigate
+     *
+     * @param title The title
+     * @return Fluent pattern
+     */
+    B title(Text title);
```
Mumfrey: Inventories should take `Translation` not `Text`

kashike: I disagree, they should be able to take `Text`. You create a `TranslatableText` that contains a `Translation`, or a non-translatable (plain) text

Zidane:
@Mumfrey 

Excellent post as always. You've touched on exactly what I was thinking of too. I think my old disagreements come solely down to naming of classes/methods and some structuring. Namely:

* Rename Archetype -> InventoryArchetype
* Rename Properties -> VanillaInventoryProperties? Depends on if this is just primitive values or an Object
* Interesting regarding Archetypes.factory, I was gonna put the builder in XXXXArchetype (WorldArchetype in 1.9.4 has the builder interface in it). I'd say we should decide on keeping it there or moving it as all other catalogs would follow that strategy.
* CustomInventory needs some name...not sure the actual name but I strongly dislike anything "Custom" in its name (and lets face it, its not an Inventory whatsoever).

src/main/java/org/spongepowered/api/item/inventory/custom/CustomInventoryBuilder.java
JBYoshi: Actually, the underlying implementation just uses `Text` - as an example of why this would be needed, chests with custom names have the custom name (which may or may not be a translation) as their title.

Mumfrey: I disagree, since containers are very much a prominent interface element, much more so than messages, something I'm keen to maintain throughout inv API is that titles displayed on containers are localisable. By forcing the consumer to make the title localisable, it means server owners and others are free to provide localisations for these critical UI elements.

kashike: It doesn't make sense to force things to be translatable, nor is it really feasible without rewriting everything in the API/impl that uses `Text`/etc. Using `Translation` is not possible for non-vanilla/forge mods on the client, either - they only work with translations available on the client.

Mumfrey: That's not the point here. The point is that Container titles **cannot** utilise rich text features, and so there is no sensible argument to **not** enforce them being localisable other than "I'm a lazy coder", which isn't strong enough to counteract the benefit of container titles being localisable.

This is a decision I made right back at the start of designing the inventory API, and it reflects the importance of the UI elements which are Containers. We're not talking about sending localisation keys to clients, we're talking about server-side translations. Making the title a translation key means that:

* responsible coders will use translation keys in their titles and provide default translations with their plugin, **which can then be set using the normal method by server owners if they have localisations available**
* lazy coders can still pass in non-translatable "keys"

All containers should have translatable titles, literally the only argument I can currently see for this not being the case is lazy coders.

kashike: Then, as I said, the entire Translation system in the API/implementation needs to be redone.

kashike: Container titles **can** utilise text colouring/formatting by the way.

Mumfrey: Any formatting is ignored by the client.

Mumfrey: "Native" codes do pass through apparently, but those could be included in the translation value so it doesn't change anything.

kashike: Unless that's been done very recently, no it isn't.

Zidane: @kashike 

The text component that InventoryBasic and GuiChest uses grabs the unformatted text which means beyond chat colors, the other formatting is lost. You can get around it but you won't get Text-level formatting.

kashike: The point of the Text system is to **avoid** telling people to use *legacy*/*deprecated* formatting techniques, and to instead use `TextFormat`/`TextStyle` (which can't be done when accepting `Translation`).

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetypes.java
```diff
@@ -0,0 +1,80 @@
...
+
+    // Vanilla Archetypes
+    // TODO Dummies
+    public static final InventoryArchetype CHEST = null;
+    public static final InventoryArchetype WORKBENCH = null;
+
+    public static final InventoryArchetype MENU_ROW = null;
+    public static final InventoryArchetype MENU_GRID = null;
+
+    public static class Properties {
```
Zidane: I'd put these off into their own class.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetype.java
```diff
@@ -0,0 +1,62 @@
...
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.block.tileentity.carrier.TileEntityCarrier;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.item.inventory.Carrier;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(InventoryArchetypes.class)
+public interface InventoryArchetype extends CatalogType {
+
+    static InventoryArchetype from(Entity entity) { // TODO add an EntityCarrier interface ?
```
Zidane: Keep the implementation of these into the builder, these would forward to the builder.

Include a static ```builder``` method that returns a new instance of the builder.

Have the Builder interface in here as well.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryBuilder.java
```diff
@@ -24,53 +24,47 @@
...
-import org.spongepowered.api.util.Identifiable;
 import org.spongepowered.api.util.ResettableBuilder;
 
 /**
- * A CustomInventory is an inventory that has no backing in the world that is
- * created by plugins.
+ * A Builder for inventories that have no backing in the world that are created by plugins.
  */
-public interface CustomInventory extends OrderedInventory, Identifiable {
+public interface InventoryBuilder extends ResettableBuilder<Inventory, InventoryBuilder> {
```
Zidane: Put this into InventoryArchetype as ```InventoryArchetype.Builder```

src/main/java/org/spongepowered/api/item/inventory/type/DispenserInventory.java
```diff
@@ -0,0 +1,48 @@
+/*
```
Zidane: I think you accidentally removed this.

src/main/java/org/spongepowered/api/item/inventory/crafting/CraftingInventory.java
```diff
@@ -54,5 +54,4 @@
      * @return The recipe or {@link Optional#empty()} if no recipe is formed
      */
     Optional<Recipe> getRecipe();
-
```
kashike: Don't remove this new line.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetypes.java
```diff
@@ -0,0 +1,80 @@
...
+
+        public static final InventoryProperty DOUBLE_CHEST = null;
```
kashike: Group this with `CHEST` above?

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetypes.java
```diff
@@ -0,0 +1,80 @@
...
+
+        public static InventorySize size(Vector2i dimension) {
```
kashike: `int, int` method as well, `Vector2i` method can be a default which calls `int, int`

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetypes.java
Mumfrey: This was from my example, it's probably temporary.

Zidane: Double Chest should be composed as two distinct chest inventories under the parent double chest inventory. It should be an archetype.

Faithcaio:
So I thought a bit about the Factory and Builder in Archetype.

I think we do not need both as the Builder would just group together Archetypes and Properties to build an inventory like the Factory.

So if you want some custom Inventory you'd first build your own Archetype and then just call `Archetype#newInventory()` allowing to easily reuse your custom Inventory Archetype.

Modified example from above:
```java
Archetype myMenu = Archetype.builder()
    .named("translation.key")
    .of(Archetypes.MENU_ROW)
    .of(Archetypes.MENU_GRID)
    .size(9, 2)
    .build();

Inventory menuInv = myMenu.newInventory();
Inventory menuInv2 = myMenu.newInventory();
```

Thoughts?

Zidane:
@Faithcaio 

Properties shouldn't be hard coded. We never know what properties are available and this is purely up to the impl in question (mod inventories could have different properties all their own). Here is further refinement of what Mumfrey mentioned above:

```
public static final InventoryArchetype BACKPACK = InventoryArchetype.builder()
    .named("translation.key")
    .with(Archetypes.MENU_GRID)
    .withProperty("size", Size.of(9, 54))
    .build(Backpack.PLUGIN_ID + ":inventory_backpack", "Backpack"); // build registers the catalog type and therefore we need the id and name
```

I think everyone can find the above basically what we want. Its hugely flexible and insanely powerful. The issue at this point is "I've got this archetype...how do I make an Inventory instance from it?". Looking at WorldArchetype (which is very similar to this) in 1.9.4, Server has a Server#createWorldProperties and then Server#loadWorld. Ignoring the latter (as its needed for world things), I'm not a fan of the former but that would mean that the Builder would go into Inventory itself. The question at that point is how far do we enforce the archetype? Can you build an Inventory without one? Do to the nature of the base game, this is hard to answer and I'd almost say no. Which brings us to your code above. My only grief is any deviations (with your change) of my BACKPACK archetype means it'll be registered which is not what I want. We do need a layer between Archetype and Instance where the Instance can be configured with specific data for THAT instance. With that in mind, one small addendum to your change:

```
// Some code later
Inventory backpack = BACKPACK.createInstanceBuilder().named().with().size().build() // I'm not very good with names
```

With the above line, I can configure this instance of the Archetype with data that I want solely for this instance. Someone help out with better names. I'm done rambling :).
```

**Comment by Faithcaio** [comment]
Wouldn't that be the same as:
````java
InventoryArchetype smallerBackpack = BACKPACK.toBuilder().withProperty("size", Size.of(9, 1)).build(); // Unregistered Archetype
Inventory instance = smallerBackpack.newInventory();
```

Zidane:
@Faithcaio 

No because archetypes should never ever exist without registration. Also, properties cannot be changed post Inventory construction as their immutable.

Faithcaio:
@Zidane 
My issue with your proposed syntax is that you have 2 Builders one for the Archetype and one for the actual Inventory.
How about:
```java

// Some base Archetype:
public static final InventoryArchetype BACKPACK = InventoryArchetype.builder()
    .with(Archetypes.CHEST)
    .withProperty("title", new FixedTranslation("Backpack"))
    .withProperty("size", Size.of(9, 6))
    .register(Backpack.PLUGIN_ID + ":inventory_backpack", "Backpack"); 

// Build instance by converting to Builder and then building the inventory
Inventory instance = InventoryArchetype.builder().from(BACKPACK).build();
Inventory instance = BACKPACK.toBuilder().build(); // shorter
Inventory instance = BACKPACK.build(); // even shorter

// Register Modify Archetype:
InventoryArchetype SMALL_BACKPACK = BACKPACK.toBuilder().withProperty("size", Size.of(9, 1)).register(..., ...);

// Modify Archetype and build Inventory Directly
Inventory instance = BACKPACK.toBuilder().withProperty("size", Size.of(9, 1)).build(); 
// Same as
Inventory instance = BACKPACK.build();

// Building it the long version:
InventoryArchetype.Builder builder =  InventoryArchetype.builder()
    .with(Archetypes.CHEST)
    .withProperty("title", new FixedTranslation("Backpack"))
    .withProperty("size", Size.of(9, 1));
// builder is like an "unregistered Archetype" 
// can be used multiple times:
Inventory inventory = builder.build();
```

Mumfrey:
@Faithcaio I think having the accessor the get the Inventory factory from the Archetype as a convenience (and also expressive) syntax is okay, but I agree with @Zidane that mixing the paradigms of building an archetype and building an inventory *based on* that archetype is a bad plan.

From a conceptual standpoint these aspects need to be preserved.

* Intrinsic properties of a composite archetype need to be baked into the composite archetype itself, this requires a clear factory pattern for archetypes
* Delegation can be provided for upstream composite properties so that downstream consumers can still access native properties in the component archetypes, as per `Delegate` in my example. This should also be baked during by the Archetype factory
* Properties being applied to an *instance* (downstream) should be applied **to an instance factory, not an archetype factory** because these properties are only consumed by the archetype during instanciation and are not a material part of the archetype itself
* Instance factory consumes an archetype and properties and returns an instance
* Archetype factory consumes an archetype, properties and property delegates, and returns an archetype

In my opinion, yes you could put convenience functionality on the Archetype for returning an instance factory, but the questions you should put to yourself:

* Is this a more expressive syntax? Does it lead to clearer code?
* Does it limit the scope of archetypes and potentially lead to a situation where similar functionality in the future needs to also be bundled into Archetype for consistency.

As I said, while I don't think the idea is entirely a bad one, I have concerns that it's not necessary and doesn't carry all that many benefits. The only benefit I really see is that the syntax is more compact, rather than more expressive.

Mumfrey:
> My issue with your proposed syntax is that you have 2 Builders one for the Archetype and one for the actual Inventory.

Yes, this is exactly what we should have. It needs to be clear to consumers that these are very different operations with differing contracts.

Mumfrey:
This example of yours ensconces the exact problem with this approach

>    ```java
>    Inventory instance = InventoryArchetype.builder().from(BACKPACK).build();
>    Inventory instance = BACKPACK.toBuilder().build(); // shorter
>    Inventory instance = BACKPACK.build(); // even shorter
>    ```

The mixture of paradigms here makes it very hard to tell what builder is being returned (I mean, the first line looks like it returns an Archetype factory, when clearly it returns an instance factory.

I think this demonstrates nicely how it fails on the point

* Is this a more expressive syntax? Does it lead to clearer code?

I think the answer is firmly "no" here.

ST-DDT:
IMO the first to look like they would return InventoryArchetypes. Similar to the ItemStackBuilder.

Mumfrey:
> IMO we need more expressive method names regarding instance creation from archetypes vs creating a archetype builder based on a given archetype. Even if it is just instanceBuilder()/instantiate() vs toBuilder().

There is no need for this if the factories are just kept entirely separate, which basically they should be.

Zidane:
I think @Mumfrey 's comments are spot on. In reading your code @Faithcaio , its really hard to tell which builder I'm getting. Perhaps the World API doesn't "suffer" from the double builder mechanism solely because WorldProperties is mutable (mostly) whereas Inventory is only mutable in the ItemStacks it holds.

I think having the two builders is a perfectly okay strategy in this instance. The next question to be answered before implementation is "Does the instance builder require an Archetype to 'kick off'"

Faithcaio:
I was thinking of having a single Builder. 
Builder#register() creates an Archetype
Builder#build() creates an Inventory

Mumfrey:
> I was thinking of having a single Builder.
> Builder#register() creates an Archetype
> Builder#build() creates an Inventory

Reasoning? This just seems to be mixing paradigms to save space rather than with any underlying logic to it.

Zidane:
@Faithcaio 

That is only assuming that the methods on the instance builder are 1:1 the methods on the archetype builder. I'm not exactly seeing a way of building an instance builder without first starting from one of the builtin archetypes (unless @Mumfrey knows something I don't know).

Mumfrey:
@Zidane even if their signatures overlap somewhat, their ouput `build()` method and fluent types are different. Again I just think that far more than any mechanical concerns, the factories need to be clearly separated.

Separated in intent, separated in code.

Zidane:
@Mumfrey 

No disagreement here, I was merely thinking of a way from a technical standpoint in the default impl of "How to use the proper container and such to show this custom inventory"...which means we need a base to work on. I do agree that separation aims for better code readability overall as these are two fundamentally different things.

Zidane:
@Faithcaio 

Again, steps in the right direction. I think what we've discussed above merits renaming Factory to Builder, having Builder extend ResettableBuilder. The build method needs to take a String id and String name to fulfill CatalogType contract.

The current thing you've called Builder should go right into Inventory as Inventory.Builder.

src/main/java/org/spongepowered/api/item/inventory/custom/InventoryArchetype.java
```diff
@@ -24,53 +24,58 @@
...
 import org.spongepowered.api.util.ResettableBuilder;
+import org.spongepowered.api.util.annotation.CatalogedBy;
 
-/**
- * A CustomInventory is an inventory that has no backing in the world that is
- * created by plugins.
- */
-public interface CustomInventory extends OrderedInventory, Identifiable {
+@CatalogedBy(InventoryArchetypes.class)
+public interface InventoryArchetype extends CatalogType {
```
Mumfrey: `InventoryArchetype extends Archetype`
`Archetype extends CatalogType`

provisions for future archetype-based systems.

Zidane: @Mumfrey 

The schematics API will be adding Archetype and it behaves differently than this. I agree, I'll have WorldArchetype extend Archetype in 1.9.4. Schematics API can adjust accordingly.

Mumfrey:
> The current thing you've called Builder should go right into Inventory as Inventory.Builder.

`Inventory.Builder`
`InventoryArchetype.Builder`

makes sense to me.

Can we also consider moving this new system to API v5 as I originally intended, as this is now such a fundamental change that it basically makes my current CustomInventory progress right in the toilet.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+        CHEST = InventoryArchetype.builder()
+                .with(MENU_GRID)
+                .property(new InventorySize(9, 3))
+                .property(new TitleProperty("minecraft.tile.chest.name"))
+                .build("minecraft:chest", "Chest");
+    }
+    public static final InventoryArchetype DOUBLE_CHEST = null; // Composite of 2 CHEST
+    {
+        DOUBLE_CHEST = InventoryArchetype.builder()
+                .with(CHEST).with(CHEST)
```
Zidane: This is more appropriate:

```
.with(CHEST)
.with(CHEST)
```

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+                .with(CHEST, 2) // alternative?
+                .property(new InventorySize(9, 6))
```
Zidane: Good idea here. A Double Chest is a parent inventory of two chests. If I query for a Chest, that would return two inventories each with 9x3. Querying the Double Chest's size property would return 9x6.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+                .property(new TitleProperty("minecraft.tile.chest.name"))
```
Zidane: Sponge spoils me with XXXX.Of in various places. Lets do ```TitleProperty.Of("minecraft.tile.chest.name");```

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * An enumeration of {@link InventoryArchetype}s.
+ */
+public class InventoryArchetypes {
+
+    // TODO Dummies
+
+    // Base Archetypes
+
+    // TODO how to start?
```
Zidane: This is an interesting question.

@Mumfrey , considering the API...can you help @Faithcaio figure out the correct way to kick off implementation? I'll give a scenario that stumps me as well.

```
InventoryArchetype COMBINED = InventoryArchetype.builder()
    .with(InventoryArchetypes.CHEST)
    .with(InventoryArchetypes.DISPENSER)
    .with(InventoryArchetypes.FURNACE)
    .property(TitleProperty.of(new FixedTranslation("Lulzy Inventory"))
    .build("combined", "Combined");

Inventory myLulzyInventory = Inventory.builder().of(COMBINED).build();

player.openInventory(myLulzyInventory);
```

Considering the above code, I've no idea what Container to show in impl. How is that defined? Is the Container a Property of the Inventory? If you use a Container that doesn't match the Vanilla archetype in question, is that a fail fast?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -451,4 +452,60 @@
...
+         */
+        Builder property(String name, InventoryProperty property);
+
+        /**
+         * Sets the InventoryArchetype and Properties according to the {@link Carrier}s Inventory
+         *
+         * @param carrier The Carrier
+         * @return Fluent pattern
+         */
+        Builder forCarrier(Carrier carrier);
```
kashike: I don't think the `for` prefix is necessary here.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -451,4 +452,60 @@
...
+
+        /**
+         * Sets the InventoryArchetype and Properties for a default Inventory of given {@link Carrier}
+         *
+         * @param carrier The Carrier class
+         * @return Fluent pattern
+         */
+        Builder forCarrier(Class<? extends Carrier> carrier);
```
kashike: I don't think the `for` prefix is necessary here.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -451,4 +452,60 @@
...
+
+        /**
+         * Builds the Inventory
+         *
+         * @return The new Inventory Instance
```
kashike: instance

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+import org.spongepowered.api.item.inventory.property.InventorySize;
+import org.spongepowered.api.item.inventory.property.SlotIndex;
+import org.spongepowered.api.item.inventory.property.TitleProperty;
+
+/**
+ * An enumeration of {@link InventoryArchetype}s.
+ */
+public class InventoryArchetypes {
+
+    // TODO Dummies
```
kashike: This is easy, see any other catalog type.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Zidane: Same as title, InventorySize.of(9, 6)

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Mumfrey: @kashike given that this is **very clearly labelled as "WIP"** can you **lay the hell off the minutiae**, all it's doing is polluting the actual discussion.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+                .with(CHEST).with(CHEST)
+                .with(CHEST, 2) // alternative?
+                .property(new InventorySize(9, 6))
+                .property(new TitleProperty("minecraft.tile.chest.name"))
+                .build("minecraft:chest2", "DoubleChest");
+    }
+    public static final InventoryArchetype FURNACE = null;
+    {
+        FURNACE = InventoryArchetype.builder()
+                .with(SLOT.toBuilder().property(new SlotIndex(0)).build("minecraft:furnace_input", "FurnaceInput"))
```
ST-DDT: Aren't we better of we use a similar strategy here as in the Data API?
````java
SLOT.toBuilder().property(Keys.SlotIndex, 0).build("minecraft:furnace_input", "FurnaceInput")
````

File: src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java (hunk 45 above)
Comment (Mumfrey) [inline src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java]: We already discussed this in TS, since every actual Archetype instance knows its own exclusivity matrix, it's fail-fast either at the call to `build()` or at the second call to `with()`.

This class simply needs to contain the possible archetypes and any obvious composites, the underlying implementation doesn't need to be considered at this stage.

File: src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java [hunk 51]
```diff
@@ -0,0 +1,114 @@
...
+    // TODO how to start?
+    public static final InventoryArchetype SLOT = null;
+    public static final InventoryArchetype MENU_ROW = null;
+    public static final InventoryArchetype MENU_GRID = null;
+
+    // Vanilla Archetypes
+
+    // Examples
+    public static final InventoryArchetype CHEST = null;
+    {
```
Mumfrey: Why all the implementation here?

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,114 @@
...
+        CHEST = InventoryArchetype.builder()
+                .with(MENU_GRID)
+                .property(new InventorySize(9, 3))
```
Mumfrey: The properties really shouldn't be appearing in the API except in the case of delegates for composite properties.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Zidane: That is what I remembered but I wanted it to essentially be on paper. We'll fail fast at build step.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Zidane: @Mumfrey 
 
This is more of "this is how it'll look". This will be turned to dummies shortly.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Mumfrey: The exclusivity matrix is obviously implementation-level though and isn't exposed in the API. This gives scope for the rules to be relaxed where it becomes possible in the future, eg. if containers become more compositional themselves.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Mumfrey: copy that.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Mumfrey: As @Zidane said, this is pseudocode. In reality this information is simply slurped into the Archetype impl from the relevant `Lens`.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Mumfrey: Why? It's returning a builder *for a `Carrier`* it's not adding a `Carrier` to a builder. It has the same semantic equivalence as `of` in other circumstances. If anything it should just be `for` except that it can't be because `for` is a keyword.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Zidane: @Mumfrey 

Here is a thought @Faithcaio chucked out in TS. We could not enforce any rules in the impl from the container perspective until the act of actually showing said Inventory to the Player occurs. Which means fail trigger happens in the act of opening the Inventory, not at building it.

There are use cases for Inventories who are never shown.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Zidane: @Mumfrey 

In this context, the builder is already made and you are setting the carrier of this builder. In that stance, having just ```carrier``` makes sense.

If this was a static method that created a builder with the carrier set, it being ```forCarrier``` would make complete sense.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
Mumfrey: Sounds like a nice idea, under the hood an `archetype -> inventory` conversion is generating 2 things: a segmented Fabric and a collection of Lenses for it. Now you're absolutely right, the conversion to VM doesn't actually need to occur until the consumer tries to dispatch the V to the client. So the MV -> V instantiation doesn't need to happen until that point.

That works as an approach.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Zidane: I'll chuck in some pseudocode based on what we have:

```
Inventory instanceInv = Inventory.builder().carrier(myTileEntity).build();

// Could add following for Mumfrey, just sugar though
Inventory instanceInvCarrier = Inventory.forCarrier(myTileEntity).build();
```

Mumfrey: I don't see how this works though. As I mentioned above, what you need to obtain for a `Carrier` (or any `Entity`) is the `InventoryArchetype` appropriate for that entity (if one exists), this if you already *have* the archetype, you're not going to be doing anything with respect to the carrier.

Faithcaio: I intended `forCarrier` to just fetch & set the appropriate InventoryArchetype in the builder.
And as Mumfrey said I would have named it `for` but that's not possible.

Wouldn't setting the Carrier some kind of property too?

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,57 @@
...
+    public static final InventoryArchetype FURNACE = DummyObjectProvider.createFor(InventoryArchetype.class, "FURNACE");
+    public static final InventoryArchetype ENCHANTING_TABLE = DummyObjectProvider.createFor(InventoryArchetype.class, "ENCHANTING_TABLE");
+    public static final InventoryArchetype ANVIL = DummyObjectProvider.createFor(InventoryArchetype.class, "ANVIL");
+    public static final InventoryArchetype BREWING_STAND = DummyObjectProvider.createFor(InventoryArchetype.class, "BREWING_STAND");
+    public static final InventoryArchetype BEACON = DummyObjectProvider.createFor(InventoryArchetype.class, "BEACON");
+    // Entity
+    public static final InventoryArchetype HORSE = DummyObjectProvider.createFor(InventoryArchetype.class, "HORSE");
+    public static final InventoryArchetype PLAYER = DummyObjectProvider.createFor(InventoryArchetype.class, "PLAYER");
+    // Creative?
+    public static final InventoryArchetype CREATIVE = DummyObjectProvider.createFor(InventoryArchetype.class, "CREATIVE");// ?
```
Zidane: This doesn't feel like an inventory archetype.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
me4502: I agree with Zidane here, it doesn't makes sense for inside a builder. It's more that it's setting the carrier property, and a few other properties along with it. forCarrier implies a static method to me that creates a new builder.

Zidane: @me4502 

Spoke with @Mumfrey  on this. Its purely meant to generate an InventoryArchetype FROM a Carrier and set it to the archetype of this builder. This is perfectly fine.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
me4502: This is a complex one - It representing the creative item grid makes sense, however that doesn't really have a use in the API. That item grid is very specific just for that case.

It's more the distinction of 'PLAYER' and 'CREATIVE' that's an issue as well. It's still the players inventory, just with the creative item selection grid.

IMO, if it's usable for the API (Such as selecting items for something?) it should be an Archetype, as it represents something different to the others.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * An enumeration of {@link InventoryArchetype}s.
+ */
+public class InventoryArchetypes {
+
+    // Base Archetypes
+    // TODO how to start?
+    public static final InventoryArchetype SLOT = DummyObjectProvider.createFor(InventoryArchetype.class, "SLOT");
+    public static final InventoryArchetype MENU_ROW = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_ROW");
```
Mumfrey: Don't forget to add more of these, eg. `MENU_COLUMN`, `MENU_BUTTON` (slot which acts like a button), `MENU_ICON` (slot which acts like a status icon), `MENU_CHECKBOX` (slot which acts like a toggle?).

Use your imagination. It may be useful to add events for these things too. This way a plugin can just construct a menu and then listen to events from its buttons.

ST-DDT: IMO we should not use events (listeners on plugins) for this and use a callback system. `BiPredicate<Inventory(Archetype), Player/Carrier>`. The event handler are just overhead that only have to exist once in the impl with this approach.

Mumfrey: > IMO we should not use events (listeners on plugins) for this and use a callback system. BiPredicate<Inventory(Archetype), Player/Carrier>.

Yes because callbacks in an otherwise entirely event-driven framework is a good practice `</sarcasm>`

> The event handler are just overhead that only have to exist once in the impl with this approach.

The who-what with the where now?

ST-DDT: > > IMO we should not use events (listeners on plugins) for this and use a callback system. BiPredicate.

> Yes because callbacks in an otherwise entirely event-driven framework is a good practice </sarcasm>

Well, we also have [`ExecuteCallback`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/action/ClickAction.java#L125) for `Text` based menus.

And yes IMO its a good idea because its easier to understand and maintain for plugin devs.

**Example:**
Callback:
````java
InventoryArchetype TEAM_SELECTION = InventoryArchetype.builder()
.with(InventoryArchetype.MENU_BUTTON.toBuilder().with(Keys.CALLBACK, (player, inventory) -> this.joinTeam(player, inventory, Team.RED)).build())
.with(InventoryArchetype.MENU_BUTTON.toBuilder().with(Keys.CALLBACK, (player, inventory) -> this.joinTeam(player, inventory, Team.BLUE)).build())
.register();

public void openTeamSelectionMenu(Player player) {
    TEAM_SELECTION.openFor(player);
}

public void joinTeam(Player player, Inventory inventory, Team team) {
    inventory.close();
    player.sendMessage("You joined team " + team.getName());
}
````
Event driven:
````java
InventoryArchetype TEAM_SELECTION = InventoryArchetype.builder()
.with(InventoryArchetype.MENU_BUTTON.toBuilder().with(Keys.CALLBACK_ID, "MYPLUGIN:joinRed").build())
.with(InventoryArchetype.MENU_BUTTON.toBuilder().with(Keys.CALLBACK_ID, "MYPLUGIN:joinBlue").build())
.register();

public void openTeamSelectionMenu(Player player) {
    TEAM_SELECTION.openFor(player);
}

public void joinTeam(Player player, Inventory inventory, Team team) {
    inventory.close();
    player.sendMessage("You joined team " + team.getName());
}

public void onMenuClick(InventoryClickEvent.ClickMenuOption event) {
    Player player = event.getPlayer();
    Inventory inventory = event.getInventory();
    String callbackId = event.foo().get(Keys.CALLBACK_ID);
    switch(callbackId) {
        case "MYPLUGIN:joinRed": 
            joinTeam(player, inventory, Team.RED);
            return;
        case "MYPLUGIN:joinBlue": 
            joinTeam(player, inventory, Team.BLUE);
            return;
        case "MYPLUGIN:leave": 
            [...]
            return;
        case "MYPLUGIN:invite": 
            [...]
            return;
        [...]
    }
}
````

**Possible issues**
Callback:
* (We use events otherwise)
* ???

Event driven:
* Possible conflicts with unqualified callback ids
* The button definition is separated from its action
* More event listeners for those menu events
* The listener method looks ugly/messy since it has to contain all cases, that could appear in the plugin.
* More code for plugins to write -> less readable

**Server implementation effort**
Should be mostly the same for both versions, since you have to check whether you are in a special control menu in both cases.
You have slight overhead in the callback version due to the necessary mapping (But that is basically a map lookup).

--------------------------------------

> > The event handler are just overhead that only have to exist once in the impl with this approach.

> The who-what with the where now?

The event handlers in plugins are overhead, because you could implement a single one in the server impl that does the necessary `button -> action` mapping.
See my above code examples:
The first one does not need a `onMenuClick` event handler since the server does that internally.
In the second example every plugin has to register a `onMenuClick` event handler to deal with its button itself.
Both are able to do exactly the same actions, just with other syntax and code placement.

Comment (Mumfrey) [inline src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java]: That makes more sense, +1 for providing a sensible rationale. To be honest maintaining parity with the text-based menus is sufficient reason to use callback instead, I'd rather have consistency any day of the week.

The listener can't really be passed to the archetype because the archetype may be consumed by another plugin and *that* plugin needs a way to supply appropriate listeners which is why I was thinking events made more sense (since event could just be registered on the archetype with an ID then consumed by *any* plugin) but thinking about it the event could simply be declared on the archetype and then attached to a listener at instance time:

```java
InventoryArchetype menuArchetype = .... ..... .withEvent("button_event_foo") .... .build();
Inventory menuInstance = Inventory.Builder.of(menuArchetype).listener("button_event_foo", Foo::bar).build();
```

> The event handlers in plugins are overhead, because you could implement a single one in the server impl that does the necessary button -> action mapping.

Not buying that at all.

Zidane: @ST-DDT 

Drop the argument that event handlers are overhead unless you are ready to generate facts to prove it. I'll spare you some effort, **they are not**.

ST-DDT: > ````java
InventoryArchetype menuArchetype = .... ..... .withEvent("button_event_foo") .... .build();
Inventory menuInstance = Inventory.Builder.of(menuArchetype).listener("button_event_foo", Foo::bar).build();
````

well this just looks like my example except that it uses a delegate.
And your `Foo::bar` is a callback (you just name it event. IMO a event uses the `@Listener` approach and `event.getMenuEventId()` or alike.
My example has the benefit that it might be enhenced in the future to support stuff like onDoubleClick, onHover onDragAndDrop, whereas your `withEvent` is strictly limited to one.

> Drop the argument that event handlers are overhead unless you are ready to generate facts to prove it. I'll spare you some effort, they are not.

Yes, you are right they are not overhead, because you have to implement them anyway. (The callback system is build on top of the event system). Its just overhead to force the user to reimplement it every time.

Comment (Mumfrey) [inline src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java]: > well this just looks like my example except that it uses a delegate.

You missed the actual point. Yes it's like yours **but it uses a delegate for a reason**. That reason being that the **event handler must be attached to the _instance_ and not to the _archetype_**. I frankly don't give a shit about the syntax but if it's going to be *callbacks* rather than *events* then the callback must be registered against the instance, otherwise a consumer of an archetype registered by another plugin/mod has no way to sink those events!

> Yes, you are right they are not overhead, because you have to implement them anyway. (The callback system is build on top of the event system). Its just overhead to force the user to reimplement it every time.

I'm with @Zidane here. Just drop it, your point is invalid.

Comment (Mumfrey) [inline src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java]: > And your Foo::bar is a callback (you just name it event. IMO a event uses the @Listener approach and event.getMenuEventId() or alike.

I didn't "name it" anything, I was agreeing with your point that callbacks work. I just used that in my example because it was easier to type to make the point.

File: src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java [hunk 55]
```diff
@@ -0,0 +1,63 @@
...
+    public static final InventoryArchetype MENU_GRID = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_GRID");
+    public static final InventoryArchetype MENU_COLUMN = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_COLUMN");
+    public static final InventoryArchetype MENU_BUTTON = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_BUTTON");
+    public static final InventoryArchetype MENU_ICON = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_ICON");
+    public static final InventoryArchetype MENU_CHECKBOX = DummyObjectProvider.createFor(InventoryArchetype.class, "MENU_CHECKBOX");
+    // TODO use my imagination
+
```
Cybermaxke: Maybe something like a _slider_, left click to increase the value and right click to decrease it.

Deamon5550: Better to call that a 'spinner' as thats closer to what you're talking about than a slider.

src/main/java/org/spongepowered/api/item/inventory/property/InventorySize.java
```diff
@@ -31,6 +31,8 @@
 /**
  * property for inventories of a particular size. For example to allow querying
  * for InventoryRows of length 9 or GridInventories of size 3x3
+ *
+ * TODO maybe rename this? InventorySize sounds more like an IntProperty
```
pschichtel: InventoryDimensions ?

JBYoshi: IMO `InventorySize` sounds good.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetype.java
```diff
@@ -0,0 +1,184 @@
...
+    default Builder toBuilder() {
+        return builder().from(this);
+    }
+
+    /**
+     * Builds a new {@link Inventory} based of this InventoryArchetype alone.
+     *
+     * @return The new Inventory
+     */
+    default Inventory newInventory() {
```
kashike: Why not `build` or `create`?

gabizou:
@Faithcaio can you squash the smaller commits into larger commits for either cleanups or what not?

bloodmc:
@Faithcaio please rebase this and let me know when it is ready for testing. We want to get this into bleeding as soon as possible.

Faithcaio:
@bloodmc 
So rebase onto bleeding branch? This (and my Impl.) is up to date with master currently.

As for the testing part: 
My Implementation of the API is somehow working (create / manipulate / open simple vanilla Inventories) but its ugly as hell.

Also I'm really not sure what to do with 
- Properties
- Event-Listener

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetypes.java
ryantheleach: What would be the equivalent of a n-state toggle like a checkbox?

https://css-tricks.com/indeterminate-checkboxes/

gabizou:
@Faithcaio I believe I can help out with a bit of the implementation to iron it out, but as I work from morning until around this time in the afternoon, I won't really be available earlier in the day.

Faithcaio:
Stuff to consider before merging into the main branch:

Remove/Hide non-vanilla based InventoryArchetypes since there is no usable impl for them.
(SLOT, MENU_ROW, MENU_GRID, MENU_COLUMN, MENU_BUTTON, MENU_ICON, MENU_CHECKBOX, MENU_SPINNER, CRAFTING)
And probably also HORSE (not viewable without actual horse entity) and PLAYER (no vanilla Container to view another player)

InventoryProperties:
Remove ```IconProperty```, ```ReturningItemProperty```.
```InventoryCapacity```: (currently using InventorySize for this but it only allows rectangular inventories)
```SlotIndex```: (What about composite Inventories and their index?)
```TitleProperty```: (Changing Title of Inventory after Creation ; also does not work for all Containers)
```IdentifiableProperty```: (Maybe even add this to Inventory?)

Also i've added a returnvalue to ```Player#openInventory```. (Maybe change to Optional if Inventory has no corresponding Container)
This allows tracking of an opened inventory in later events. (e.g. open Custom Chest but disallow any change in the inventory)

And maybe separate from this:
in ClickInventoryEvent it is (i think) impossible to get which inventory was clicked
(upper) viewed Inventory or the (lower) Player Inventory. ```Slot#parent()``` just returns itself.

I've merged the stuff @Aaron1011 did into here (and impl too)

Damn Timezones :/

**Comment by Mumfrey** [comment]
@Faithcaio I'll review it this week some time, I've made some changes to support impl so I will potentially pull this onto a branch to merge my changes.

File: src/main/java/org/spongepowered/api/item/inventory/property/ReturningItemProperty.java [hunk 58]
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Once the {@link Container} with this Property is closed all contained {@link ItemStack}s
+ * are returned to the specified Carrier or if {@code null} dropped on the ground.
+ *
+ * Default Property of {@link BlockTypes#CRAFTING_TABLE}, {@link BlockTypes#ANVIL},
+ * {@link BlockTypes#ENCHANTING_TABLE}, {@link EntityTypes#VILLAGER}, the players crafting grid.
+ */
+// TODO-feature/custominventory remove? was supposed to be what workbenches etc. do in vanilla
+public class ReturningItemProperty extends AbstractInventoryProperty<String, Carrier>{
```
liach: `>{` should be `> {`

Faithcaio:
ec86e3c83311490d482cb038f1d09101efa411ca

PR #1276, opened 2016-06-19
Allow plugins to load chunks asynchronously

Minecrell:
Since both our implementations support asynchronous chunk loading now we should allow plugins to use that feature by being able to load chunks asynchronously if they don't need the loaded chunk immediately.

**Implementation:** [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/303) |  [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1299)

### DIFF
--- a/src/main/java/org/spongepowered/api/world/World.java
+++ b/src/main/java/org/spongepowered/api/world/World.java
@@ -51,6 +51,8 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
 
 /**
  * A loaded Minecraft world.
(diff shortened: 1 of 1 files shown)


kenzierocks:
Why not just return `Future` or `CompleteableFuture` instead of the callback?

aikar:
Not everyone prefers futures, but you can support both.

zml2008:
CompletableFutures tend to be more flexible and are also what a lot of our API uses already, so it's probably best to use those instead.

bloodmc:
@Minecrell can you update PR to use CompletableFuture's so this can be merged

Minecrell:
@bloodmc Done. Still need to implement it though.

gabizou:
@Minecrell I'd want to have this implemented before merging. I you can get it finished, that'd be great.

stephan-gh:
@gabizou Done.

Issue #733, opened 2015-06-21
Add a way to simulate a player chatting


zml2008:
This would call the necessary events to take action as if a player had passed a given message to chat, throwing event & such (Would this work better with causes once those are added @gabizou)

This would add a method like:

``` java
/**
  * Simulate a player having chatted on the server. This will post the appropriate events, and handle commands as necessary
  *
  * @param input The input, as would be provided by this player
  * @return The event that has been thrown. Changes have no effect, this is simply to report back information
*/
PlayerChatEvent chat(String input); // Don't really think it makes sense to allow a Text here? dunno. Makes things more difficult for commands.
```

to `Player`

FerusGrim:
Why wouldn't you have the method accept a Text? `PlayerChatEvent chat(Text input)`

After all, there's plenty of uses for it. I can see maybe having both of them, but that would break the standard that we currently have, of only accepting Text.

ST-DDT:
@FerusGrim I guess because we want to be able to simulate what the player has entered in their chat window and that is just a String. Either a command in which case a `Text` is pretty useless or a chat message which then has to be check whether it is a legacy or json message.

However the method naming is bad. `chat()` sounds to much like chatting between players.

FerusGrim:
@ST-DDT I suppose that makes sense. Though, I still believe it'd be a waste to not have the ability to bedazzle the message.

Maybe at least a `PlayerChatEvent#isSimulated`, so that plugins that use it can modify it afterwards?

ST-DDT:
Is it possible for a player to send a real Json Message at all? If yes you can still convert the Text to a JSON String although this seems a little bit complicated.

Maybe just have two different input methods
* `simulateInput(String)` - Commands and Chat
* `simulateChat(Text)` - Chat only

But I have too few knowledge on the internals here whether this is even possible.

Regarding the `isSimulated()` method i am really not sure whether we should allow the plugin writers to treat messages differently whether its simulated or not.

Zidane:
@zml2008 

Can you send a command as a CommandSource in SpongeAPI? If so, we already have support. Unless you want to make that easier to use.

zml2008:
You can send a command, can't send chat though currently.

Zidane:
@zml2008 

Though, you could send the command "say" with your message (or tellraw for something complex)

zml2008:
Does that call events? I seem to remember it also using a different format for messages. 

[...]

PR #1283, opened 2016-06-26
Added Player chat simulation method. Adds #733

me4502:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1283) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/895)

This PR adds a method to Player, that allows for the simulation of chat.

The exact reasoning of it can be seen in the corresponding issue, https://github.com/SpongePowered/SpongeAPI/issues/733

This method returns the event so that the plugin can see what modifications have been made by other plugins, and to see if it has been cancelled.

Test code,

``` java
    @Listener
    public void onInteract(InteractBlockEvent.Secondary event, @First Player player) {
        player.simulateChat(Text.of("Hello there"));
    }

    @Listener
    public void onMessage(MessageChannelEvent.Chat chat, @First Player player) {
        Sponge.getServer().getBroadcastChannel().send(Text.of("Hello, " + player.getName() + '!'));
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/Player.java
@@ -26,6 +26,11 @@ package org.spongepowered.api.entity.living.player;
 
 import org.spongepowered.api.Server;
 import org.spongepowered.api.block.tileentity.EnderChest;
+import org.spongepowered.api.command.CommandManager;
+import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.event.message.MessageChannelEvent;
+import org.spongepowered.api.text.channel.MessageReceiver;
+import org.spongepowered.api.text.chat.ChatType;
 import org.spongepowered.api.command.source.RemoteSource;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.mutable.DisplayNameData;
@@ -125,6 +130,26 @@ public interface Player extends Humanoid, User, RemoteSource, Viewer, ChatTypeMe
      * @return True if colors are enabled in chat
      */
     boolean isChatColorsEnabled();
+
+    /**
+     * Simulates a chat message from a player.
+     *
+     * <p>This method sends a message as if it came from this player.
+     * To send a message to this player instead, see
+     * {@link MessageReceiver#sendMessage(Text)} or
+     * {@link ChatTypeMessageReceiver#sendMessage(ChatType, Text)}.</p>
+     *
+     * <p>Commands cannot be sent using this method. To send commands, use
+     * {@link CommandManager#process(CommandSource, String)}.</p>
+     *
+     * <p>If text formatting is not supported in the implementation
+     * it will be displayed as plain text.</p>
+     *
+     * @param message The message to send
+     * @param cause The cause for the message
+     * @return The event that was thrown from sending the message
+     */
+    MessageChannelEvent.Chat simulateChat(Text message, Cause cause);
 
     /**
      * Gets the skin parts that this player has allowed to render.


me4502:
As a potential discussion point - should the plugin be added as the root cause, rather than the player? I wasn't sure about this as it'd require passing the plugin - as well as possibly confusing plugins that expect the root cause to be a player.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -119,6 +120,14 @@
     boolean isChatColorsEnabled();
 
     /**
+     * Simulates a chat message from a player.
+     *
+     * @param message The message to send
+     * @return The event thrown from the message
+     */
+    MessageChannelEvent.Chat chat(Text message);
```
kashike: This should accept `String` in my opinion. Clients send a `String` only, and this method should be used to simulate a client sending a `String` to the server.

me4502: That would however limit what the plugin can do. It feels like a severe limitation for no reason other than to prevent them doing something the vanilla client can't.

kashike: It would limit it... how? You convert to a legacy text as it is, which limits things as is.

me4502: That was one part I wish to change, I'd need to discuss it with people first - I wanted to move chat handling over to common, and handle it properly there as a Text.

As well as that, I want to discuss how it'd be handled on a client plugin - which is why I've currently put the code in EntityPlayerMP.

ryantheleach: Maybe the client gains the ability to bold text in the future?

ST-DDT: There should be a hint whether:
- This actually sends messages to someone or not
- Whether the event is already processed or not
- Whether you can send commands

me4502: @ST-DDT What do you mean by "This actually sends messages to someone or not"?

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -119,6 +120,17 @@
...
     /**
+     * Simulates a chat message from a player.
+     *
+     * <p>This method sends a message as if it came from this player.
+     * It does not send a message to this player.</p>
+     *
+     * @param message The message to send
+     * @return The event that was thrown from sending the message
+     */
+    MessageChannelEvent.Chat chat(Text message);
```
ST-DDT: Your changes fixed most of my concerns. However it still does not explicitly states whether this method can be used to send commands.
Maybe there should be a hint which method can be used to send commands because both features are somehow related.

ST-DDT: > It does not send a message to this player.

I'm not quite sure about this one. Because it implies that the used MessageChannel explicitly excludes the given player. Although the player would normally get a formatted version of what he just typed.

me4502: I'll rephrase it to "To send a message to this player, see sendMessage(Text)"

kashike: Perhaps call this `simulateChat` instead?

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -119,6 +123,24 @@
     boolean isChatColorsEnabled();
 
     /**
+     * Simulates a chat message from a player.
+     *
+     * <p>This method sends a message as if it came from this player.
```
kashike, line 137:
``` java
     * <p>This method sends a message as if it came from this player.
     * To send a message to this player instead, see
     * {@link MessageReceiver#sendMessage(Text)} or
     * {@link ChatTypeMessageReceiver#sendMessage(ChatType, Text)}.</p>
     *
     * <p>Commands cannot be sent using this method. To send commands, use
     * {@link CommandManager#process(CommandSource, String)}.</p>
     *
     * <p>If text formatting is not supported in the implementation
     * it will be displayed as plain text.</p>
```

Minecrell:
@me4502 IMO the plugin shouldn't be the root/source of the cause of the event, but at least part of it, so other plugins can identify the plugin sending the message. Maybe the methods could also take a `Cause` directly so the plugins could add more context if they wanted to?

me4502:
Okay, I've added a Cause to the method. That makes the most sense IMO, and is inline with how setBlock etc work.

Issue #1165, opened 2016-04-01
Command usage displayed incorrectly


Flibio:
Command arguments for sub-commands do not display at all in the help message when a user types in a sub-command without the required arguments. A detailed explanation can be found below.

Command `/main` has a sub-command `set`. The sub-command `set` has a required argument, called `arg1`. Running `/main set`, without submitting the required argument, displays an unhelpful usage message: `Usage: /main set`. It fails to display the required argument, `arg1` in the usage message.

The command was created by implementing the `CommandExecutor` and registering a `CommandSpec`.

liach:
How does a `SimpleDispatcher` work? Does it work incorrectly in the same way?

dualspiral:
In the latest SF for API 7 snapshot, I've added method `childArgumentParseExceptionFallback` to the `CommandSpec`. You need to set this to `false` to get the behaviour you expect above, I opted to make this a non-breaking change for API 7.

PR #1284, opened 2016-06-26
Various command fixes and some additions.

Cybermaxke:
- Added the ColorCommandElement
- Allow Vector3d and Location<World> elements to be completed through targeted blocks
- Fixed tab completion behaviour of in command optionals and flags that aren't at the end, for example: `/tp [<target>] <destination>|[-world:-w <world>] <coordinates>`
- Fixed a error caused by the WorldPropertiesElement when there are disabled properties present
- Completely reset the CommandArgs state after parsing every FirstParsingCommandElement entry
- Some other small bug fixes

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandArgs.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandArgs.java
@@ -122,6 +122,10 @@ public final class CommandArgs {
         return this.args;
     }
 
+    void setArgs(List<SingleArg> args) {
+        this.args = args;
+    }
+
     /**
      * Return this arguments object's current state. Can be used to reset with the {@link #setState(Object)} method.
      *
@@ -153,6 +157,20 @@ public final class CommandArgs {
     }
 
     /**
+     * Insert an {@link Iterable} of args as the next args to be
+     * returned by {@link #next()}.
+     *
+     * @param values The arguments to insert
+     */
+    public void insertArgs(Iterable<String> values) {
+        int index = this.index < 0 ? 0 : this.args.get(this.index).getEndIdx();
+        int pos = this.index;
+        for (String value : values) {
+            this.args.add(++pos, new SingleArg(value, index, index));
+        }
+    }
+
+    /**
      * Insert an arg as the next arg to be returned by {@link #next()}.
      *
      * @param value The argument to insert
@@ -163,12 +181,29 @@ public final class CommandArgs {
     }
 
     /**
+     * Remove the arguments parsed at the state.
+     *
+     * @param state The state
+     */
+    public void removeArg(Object state) {
+        if (!(state instanceof Integer)) {
+            throw new IllegalArgumentException("The provided state was not of the correct type!");
+        }
+        int index = (Integer) state;
+        if (this.index > index) {
+            this.index--;
+        }
+        this.args.remove(index);
+    }
+
+    /**
      * Remove the arguments parsed between startState and endState.
      *
      * @param startState The starting state
      * @param endState The ending state
+     * @return The removed arguments
      */
-    public void removeArgs(Object startState, Object endState) {
+    public List<String> removeArgs(Object startState, Object endState) {
         if (!(startState instanceof Integer) || !(endState instanceof Integer)) {
             throw new IllegalArgumentException("One of the states provided was not of the correct type!");
         }
@@ -181,9 +216,11 @@ public final class CommandArgs {
                 this.index -= (endIdx - startIdx) + 1;
             }
         }
+        final List<String> args = new ArrayList<>();
         for (int i = startIdx; i <= endIdx; ++i) {
-            this.args.remove(startIdx);
+            args.add(this.args.remove(startIdx).getValue());
         }
+        return args;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


zml2008:
Thanks for this! I always appreciate folks hacking at the commands API. I haven't looked at this super in depth, but the changes look for the most part quite good. The only thing that I've noticed so far is that you're doing a separate get/set Args in CommandArgs -- the get/set State methods return an object so that at some point in the future both position and arg contents could be snapshotted and restored. Unless there's some special handling you have to do for some reason, I'd prefer if you just reused the state getter and setter, to handle both at once.

src/main/java/org/spongepowered/api/command/args/CommandArgs.java
```diff
@@ -153,6 +157,20 @@ public String getRaw() {
     }
 
     /**
+     * Insert an {@link Iterable} of args as the next args to be
+     * returned by {@link #next()}.
+     *
+     * @param values The arguments to insert
+     */
+    public void insertArgs(Iterable<String> values) {
```
ryantheleach: Should this be called addAll and take an index to match the java collections API? Or does insertArgs work just as well?

Cybermaxke: @ryantheleach I named this method to be consistent with the `insertArg` method.

Meegooo:
Can you also look into this? https://github.com/SpongePowered/SpongeAPI/issues/1165

src/main/java/org/spongepowered/api/command/args/CommandArgs.java
```diff
@@ -204,4 +258,17 @@ public int getRawPosition() {
...
+    /**
+     * A snapshot of the {@link CommandArgs}.
+     */
+    static final class Snapshot {
+
+        private final List<SingleArg> args;
+        private final int index;
+
+        Snapshot(List<SingleArg> args, int index) {
+            this.args = args;
```
ST-DDT: Maybe this should use a defensive immutable copy. ImmutableList.copyOf

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -164,4 +234,16 @@ public boolean hasAny(String key) {
...
+
+    /**
+     * A snapshot of the {@link CommandContext}.
+     */
+    static final class Snapshot {
+
+        private final Multimap<String, Object> parsedArgs;
+
+        Snapshot(Multimap<String, Object> parsedArgs) {
+            this.parsedArgs = parsedArgs;
```
ST-DDT: IMO the defensive copy should be created here instead to avoid potentially bad calls without copy.
Is there an immutable version of this as well?

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -180,12 +187,13 @@ private boolean parseShortFlags(CommandSource source, String shortFlags, Command
     @Override
     public Text getUsage(CommandSource src) {
         final List<Object> builder = new ArrayList<>();
-        for (Map.Entry<List<String>, CommandElement> arg : this.usageFlags.entrySet()) {
+        for (Iterator<Map.Entry<List<String>, CommandElement>> it = this.usageFlags.entrySet().iterator(); it.hasNext(); ) {
+            final Map.Entry<List<String>, CommandElement> arg = it.next();
             builder.add("[");
-            for (Iterator<String> it = arg.getKey().iterator(); it.hasNext();) {
+            for (Iterator<String> it1 = arg.getKey().iterator(); it1.hasNext();) {
```
ST-DDT: You should use better names for these iterators.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -438,17 +620,46 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
...
+                        elements.add(element);
+                        lastIndex = index;
+                    } else if (index == lastIndex) {
+                        elements.add(element);
+                    }
+                }
+                args.restoreSnapshot(argsSnapshot);
+            }
+            if (!elements.isEmpty()) {
+                final List<String> ret0 = new ArrayList<>();
```
ST-DDT: Should probably be called `result`.

bloodmc:
@Cybermaxke can you update this for 1.11 bleeding branch

zml2008:
This still has the issues I mentioned in my initial comment back in june -- get/setState should be entirely replaced by the snapshot system, not done in addition to.

Cybermaxke:
@dualspiral Can you also include the color command element (and maybe the relative vector3d one?) from this PR in your Command API PR?

parlough:
@Cybermaxke Which of these changes do you think are important to get in before API 7 is released? I think the snapshot changes are out of scope but do you have specific fixes in mind that need to/should get in? I could pick them out and work from there.

parlough:
Due to this going to be obsolete in API 8 once the command changes are introduced and there being many merge conflicts, going to close this for now. If you'd still like to get any non-breaking changes into 7.1, feel free to open a new PR with those.

PR #1285, opened 2016-06-26
Adding a method to create an entity naturally

rexbut:
API | [Common](https://github.com/SpongePowered/SpongeCommon/pull/818)

Issue : https://github.com/SpongePowered/SpongeAPI/issues/1107

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java
+++ b/src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java
@@ -180,7 +180,7 @@ public interface EntityUniverse {
      * @return An entity, if one was created
      */
     default Optional<Entity> createEntityNaturally(EntityType type, Vector3i position) {
-        return createNaturallyEntity(type, position.toDouble());
+        return createEntityNaturally(type, position.toDouble());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kenzierocks:
`spawnEntityNaturally`?

rexbut:
You want to rename functions "createEntityNaturally" ?

src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java
```diff
@@ -150,6 +150,40 @@
...
+     * {@link Optional#empty()} was returned, the entity is not able to spawn at
+     * the given location. Furthermore, this allows for the {@link Entity} to be
+     * customized further prior to traditional "ticking" and processing by core
+     * systems.</p>
+     *
+     * @param type The type
+     * @param position The position
+     * @return An entity, if one was created
+     */
+    Optional<Entity> createNaturallyEntity(EntityType type, Vector3d position);
```
kashike: `createEntityNaturally` sounds better, in my opinion

ZephireNZ:
Correct me if I'm mistaken, but shouldn't this target master? There's no breaking change involved, only an API addition.

rexbut:
Yes I think that could be committed to the master branch

gabizou:
I'd rather add it in API 5.0.0 (bleeding) since there will not likely be any more releases of API 4.x. I'm of the opinion this can be added to bleeding very easily, but @Zidane may have differing opinions.

rexbut:
To fix conflicts, I recreated a new branch : https://github.com/SpongePowered/SpongeAPI/pull/1350

ZephireNZ:
FYI, in the future you can just push to an existing branch to update a PR.

PR #1295, opened 2016-07-04
Dialogue API

### LINKED ISSUE #432: Conversation API
Bukkit had a decent API called "Conversations" where a [`Conversable`](https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/conversations/Conversable.java) would be able to converse with a plugin and vice versa. 

The core of the [`Conversation`](https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/conversations/Conversation.java) was being able to use a `Conversation` to maintain a quasi transient [`ConversationContext`](https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/conversations/ConversationContext.java) that basically was a `Map<String, Object>` to contain information pertinent to the conversation in the context between the player and the plugin.

The API had quite a bit to use, and there are some plugins that have been able to use them more robustly for interactive chat-based menus, and to be honest, I'd like to see some form of this API in SpongeAPI. 

Ideally, as the [`getConversationData(Object)`](https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/conversations/ConversationContext.java#L56) is rather abstract, I'd prefer to see some use of [`DataQuery`](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/persistence/data/DataQuery.java) as keys. Of course, all of this is up for debate and discussion.

pie-flavor:
Nobody was doing this, and it seemed fun. So here is the Sponge Dialogue API, which is Sponge's equivalent of Bukkit's Conversations API. Fixes #432.

For those who don't know, the purpose of this sort of API is to allow plugins to initialize a back-and-forth communication in the chat. This can be used as a replacement for complicated commands, or for navigating menus.

A `DialogueArchetype` object is constructed, which defines a few rules such as whether user input is swallowed as opposed to put in chat like normal, as well as an initial `Question` to use. It also contains a `DialogueChatHandler` and an arbitrary number of `DialogueConclusionHandler`s.

A `Speaker` is somethiing that can take part in a dialogue. It has a few basic methods regarding current dialogues.

Where a `DialogueArchetype` represents the structure, a `Dialogue` represents an actual dialogue happening, and can retrieve things such as the list of `Speaker`s and what `Question` they're on.

A `Question` is comprised of text shown to the `Speaker`, as well as an `AnswerProcessor` and an arbitrary number of `AdditionalAnswerProcessor`s. It also contains rules similar to those of the `DialogueArchetype`, and if defined, override them.

An `AnswerProcessor` is a functional interface for interpreting an `Answer` and returning the next `Question`. `AdditionalAnswerProcessor`s can do the same thing, minus being able to return the next `Question`. The purpose of these is to allow multiple plugins to take part in the same `Dialogue`.

An `Answer` represents a response to a `Question`. It is mostly a container class. 

A `DialogueChatHandler` is a class which determines how to handle chat sent to a `Speaker` in the middle of a `Dialogue`. It can perform handling whenever a message is sent, and decide whether or not to allow the message through, as well as being notified when the `Dialogue` has been concluded. However, a `DialogueConclusionHandler` is only for when the `Dialogue` has been concluded.

Each `Dialogue` also contains a `DataContainer`, which all processors can use. Upon creation of the `Dialogue`, the container is populated from the defaults set by the `DialogueArchetype`, which can also be specified when building it.

Input is greatly appreciated.


src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,102 @@
...
+        Builder suppressAllMessages(boolean suppress);
+
+        /**
+         * Sets specific {@link MessageChannel}s to not suppress. This is
+         * ignored if {@link #suppressAllMessages(boolean)} is set to false.
+         *
+         * @param channels The channels to allow
+         * @return This builder, for chaining
+         */
+        Builder allowedChannels(Iterable<MessageChannel> channels);
```
ryantheleach: I'd like to see the possibility of suppressing based on ChatType That would allow you to allow action bars through, or system messages.

src/main/java/org/spongepowered/api/text/dialogue/Answer.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    default Text toText() {
+        return Text.of(getAnswer());
+    }
+
+    /**
+     * Gets the response that was sent.
+     * @return The response
+     */
+    String getAnswer();
```
kashike: This should just return `Text`.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,141 @@
...
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.Optional;
+
+/**
+ * Represents a structure for a back-and-forth communication between one or
+ * more {@link Speaker}s and one or more plugiins.
+ */
+public interface Dialogue {
```
kashike: New line between here and methods

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,141 @@
...
+         */
+        Dialogue build();
+    }
+
+    /**
+     * An instance of a {@link Dialogue} that is in progress.
+     * While {@link Dialogue} is a structure, this is an actual instance of the
+     * back-and-forth that it structures.
+     */
+    interface Instance {
```
kashike: This should just be merged with Dialogue. and the start method currently there removed.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,149 @@
...
+public interface Question extends TextRepresentable {
+    /**
+     * Creates a {@link Builder} to return an {@link Question}.
+     *
+     * @return The new builder
+     */
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+    interface Builder extends ResettableBuilder<Question, Builder> {
```
kashike: Missing empty line

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,149 @@
...
+import org.spongepowered.api.util.Tristate;
+
+import java.util.Collection;
+
+/**
+ * Each {@link Dialogue} is made up of a dynamic tree of {@link Question}s,
+ * meaning that each {@link Question} chooses the next one. A {@link Question}
+ * is responded to by a {@link Speaker} with an {@link Answer}.
+ */
+public interface Question extends TextRepresentable {
```
kashike: Missing new line

src/main/java/org/spongepowered/api/text/dialogue/AdditionalAnswerProcessor.java
```diff
@@ -0,0 +1,21 @@
+package org.spongepowered.api.text.dialogue;
```
kashike: Missing license header.

pie-flavor:
@kashike 
The point of the Instance class is so that a Dialogue can be reused many times, and the definition is separate from the instance. As for getAnswer(), I cannot fathom a way that a Speaker can send anything more complex than a String, nor how it could be then parsed.

src/main/java/org/spongepowered/api/text/dialogue/Answer.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Gets who sent the response.
+     * @return The speaker
```
kashike: New line between body and `@param`/`@return`:
```java
    /**
     * Gets who sent the response.
     *
     * @return The speaker
     */

**Comment by kashike** [comment]
@pie-flavor Then create a `DialogueArchetype` and use that as the base, and move `Dialogue.Instance` to `Dialogue`.

File: src/main/java/org/spongepowered/api/text/dialogue/package-info.java [hunk 9]
```diff
@@ -0,0 +1,26 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault
+package org.spongepowered.api.text.dialogue;
```
kashike: Missing new line at end of file.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,175 @@
...
+    interface Builder extends ResettableBuilder<Question, Builder> {
+
+        /**
+         * Sets the {@link Text} to show the {@link Speaker} when this
+         * {@link Question} is shown.
+         *
+         * @param text The text to show
+         * @return This builder, for chaining
+         */
+        Builder setText(Text text);
```
kashike: Builders don't have a `set` prefix - this should be `text(Text text)`.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,175 @@
...
+
+        /**
+         * Sets the {@link AnswerProcessor} to use when processing {@link Answer}s.
+         *
+         * @param processor The processor to use
+         * @return This builder, for chaining
+         */
+        Builder setProcessor(AnswerProcessor processor);
```
kashike: Builders don't have a `set` prefix - this should be `processor(AnswerProcessor processor)`.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    AnswerProcessor getMainProcessor();
+
+    /**
+     * Gets the {@link AdditionalAnswerProcessor}s to use when processing
+     * {@link Answer}s.
+     *
+     * @return The processor
+     */
+    Collection<AdditionalAnswerProcessor> getAdditionalProcessors();
```
kashike: Is this collection immutable or mutable?

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Gets all {@link MessageChannel}s that are allowed through suppression.
+     *
+     * @return The channels
+     */
+    Collection<MessageChannel> getAllowedChannels();
```
kashike: Is this collection immutable or mutable?

src/main/java/org/spongepowered/api/text/dialogue/AnswerProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Processes an {@link Answer} and returns the next {@link Question} in the {@link Dialogue}.
+     *
+     * @param dialogue The current dialogue
+     * @param data The current dialogue-persistent data
+     * @param question The question being answered
+     * @param answer The answer to the question
+     * @return The next {@link Question} in the {@link Dialogue}
+     */
+    Question process(Dialogue dialogue, DataContainer data, Question question, Answer answer);
```
ST-DDT: What happens if there is no next question?
Maybe this should return Optional.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,68 @@
...
+ * back-and-forth that it structures.
+ */
+public interface Dialogue {
+
+    /**
+     * Gets the {@link DialogueArchetype} object that this is an instance of.
+     *
+     * @return The dialogue object
+     */
+    DialogueArchetype getDialogue();
```
ST-DDT: This should be getArchetype.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Creates a new {@link Dialogue} for the specified {@link Speaker}(s).
+     *
+     * @param speakers The speakers to send the initial {@link Question} to
+     * @return The newly created {@link Dialogue}
+     * @throws IllegalArgumentException If {@code speakers} is empty
+     */
+    Dialogue start(Speaker... speakers);
```
ST-DDT: Please also add another version of this method for iterables or Collections.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,176 @@
...
+        Builder processor(AnswerProcessor processor);
+
+        /**
+         * Adds extra processors to use when processing {@link Answer}s. They
+         * cannot determine the next {@link Question}.
+         *
+         * @param processors The processors to use.
+         * @return This builder, for chaining
+         */
+        Builder additionalProcessors(Iterable<AdditionalAnswerProcessor> processors);
```
ST-DDT: This method can be default implemented.
Also you could change the param to `Iterable<? extends AAP>` to make it usable for custom AAP lists.

src/main/java/org/spongepowered/api/text/dialogue/Speaker.java
```diff
@@ -0,0 +1,71 @@
...
+public interface Speaker extends MessageReceiver {
+
+    /**
+     * Gets whether this speaker is currently in a {@link Dialogue}.
+     * This is equivalent to calling {@link #getCurrentDialogue()
+     * getCurrentDialogue().isPresent()}, minus the {@link Optional} boxing.
+     *
+     * @return Whether or not the speaker is in a dialogue
+     */
+    boolean isSpeaking();
```
ST-DDT: Maybe you could default implement it, since you already specified the contract for this method.

src/main/java/org/spongepowered/api/text/dialogue/Speaker.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Adds this speaker to the provided {@link Dialogue}.
+     *
+     * @param instance The dialogue to add the speaker to
+     * @throws IllegalStateException If the speaker is already in a
+     * {@link Dialogue}
+     * @see Dialogue#addSpeaker(Speaker)
+     */
+    default void addToDialogue(Dialogue instance) {
+        instance.addSpeaker(this);
```
ST-DDT: Check for null input using Preconditions.

src/main/java/org/spongepowered/api/text/dialogue/Speaker.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Gets the {@link Question} that the speaker is currently in.
+     *
+     * @return The question if the speaker is in a dialogue, otherwise
+     * {@link Optional#empty()}
+     */
+    default Optional<Question> getCurrentQuestion() {
+        return this.getCurrentDialogue().flatMap(inst -> inst.getCurrentQuestionFor(this));
```
kashike: `inst` -> `dialogue`

src/main/java/org/spongepowered/api/text/dialogue/Answer.java
```diff
@@ -0,0 +1,67 @@
...
+    default Text toText() {
+        return Text.of(getAnswer());
+    }
+
+    /**
+     * Gets the response that was sent.
+     *
+     * @return The response
+     */
+    String getAnswer();
```
kashike: This should return `Text`.

pie-flavor:
@kashike Again, I cannot fathom a way that a Speaker can send a Text, or how an AnswerProcessor could reasonably interpret the texty aspects of it.

src/main/java/org/spongepowered/api/text/dialogue/AdditionalAnswerProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Processes an {@link Question}.
+     *
+     * @param dialogue The current dialogue.
+     * @param data The current dialogue-persistent data
+     * @param question The question being answered
+     * @param answer The answer to the question
+     */
+    void process(Dialogue dialogue, DataContainer data, Question question, Answer answer);
+}
```
ST-DDT: Please add an empty line after the last member of a class/interface.

src/main/java/org/spongepowered/api/text/dialogue/Speaker.java
```diff
@@ -0,0 +1,75 @@
...
+     * Adds this speaker to the provided {@link Dialogue}.
+     *
+     * @param instance The dialogue to add the speaker to
+     * @throws IllegalStateException If the speaker is already in a
+     * {@link Dialogue}
+     * @see Dialogue#addSpeaker(Speaker)
+     */
+    default void addToDialogue(Dialogue instance) {
+        checkNotNull(instance);
+        instance.addSpeaker(this);
```
ST-DDT: You can just convert it to a one-liner:
````java
checkNotNull(dialogue, "dialogue").addSpeaker(this);
````
Using a string reason helps tracking down the cause of the NPE, so you should always attach it.
Also please rename `instance` -> `dialogue`

**Comment by kashike** [comment]
@pie-flavor Message events use text, I don't see a reason why speaker can't either.

**Comment by ST-DDT** [comment]
> @kashike Again, I cannot fathom a way that a Speaker can send a Text, or how an AnswerProcessor could reasonably interpret the texty aspects of it.

Well the client sends `Text`s objects in chat. (See chat event for reference)
If it is just a plain string then it will be a simple `LiteralText`.

**Comment by pie-flavor** [comment]
Fair.

File: src/main/java/org/spongepowered/api/text/dialogue/AnswerProcessor.java [hunk 24]
```diff
@@ -0,0 +1,49 @@
...
+     * Processes an {@link Answer} and returns the next {@link Question} in
+     * the {@link Dialogue}.
+     *
+     * @param dialogue The current dialogue
+     * @param data The current dialogue-persistent data
+     * @param question The question being answered
+     * @param answer The answer to the question
+     * @return The next {@link Question} in the {@link Dialogue}
+     */
+    Optional<Question> process(Dialogue dialogue, DataContainer data, Question question, Answer answer);
```
ST-DDT: Isn't the `DataContainer` redundant here (`Dialogue#getData()`)? Or did you add it for convenience reasons?

pie-flavor:
@ST-DDT It was a mistake is what it was.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Creates a new {@link Dialogue} for the specified {@link Speaker}(s).
+     *
+     * @param speakers The speakers to send the initial {@link Question} to
+     * @return The newly created {@link Dialogue}
+     * @throws IllegalArgumentException If {@code speakers} is empty
+     */
+    Dialogue start(Iterable<? extends Speaker> speakers);
+
+    interface Builder extends ResettableBuilder<DialogueArchetype, Builder> {
```
ST-DDT: Maybe its just my superficial impression, but shouldn't be the `Builder` definition be the last member in a class? (At least that is what the `Text` classes do).

ST-DDT:
Nice work on the API specs, this looks quite easy to use. Thanks.

pie-flavor:
@ST-DDT The important methods are up at the top. The ones at the bottom aren't that all important; we already know they exist, as they are just getters for the existing setters.

Also, thank you. :)

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return The archetype
+     */
+    DialogueArchetype getArchetype();
+
+    /**
+     * Gets the current dialogue-persistent data object.
+     *
+     * @return The data
+     */
+    DataContainer getData();
```
Deamon5550: If you're going to have this make it extend `DataSerializable` rather than having this method.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
Deamon5550: Yes it should be, Its part of the checkstyle

ST-DDT:
@pie-flavor I'm not sure about that as well.

Just thing about it:
* Will you ever have a dialogue with the console ?
* Will you ever have a dialogue with a remote ?
* Will you ever have a dialogue with a command block (that will actually answer)?
* Will you ever have a dialogue with a random entity (that will actually answer)?

The console might be borderline but everything beyond that is just very unlikely.
(Console and remote usually are only able to send commands)

Maybe we can just start with a `Player` and support more (`CommandSource`/?) later if we uncovered an actual need for it.

pie-flavor:
@ST-DDT Command blocks: Heavily unlikely. Consoles / RCON: Bukkit did it.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Gets whether or not output is suppressed by default.
+     *
+     * @return Whether output is suppressed
+     * @see Builder#suppressesOutput(boolean)
+     */
+    boolean suppressesOutput();
+
+    /**
+     * Gets whether or not messages are suppressed by default.
```
ST-DDT: You should add what actually happens if messages are suppressed. Aka `the speakers will only recieve chat from X. Any other message will be deleted/delayed...`

The same goes for the other boolean methods.

Please add an option to just delay chat. So that it will be send after the dialog is done.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,205 @@
...
+     * @return The channels
+     */
+    Collection<MessageChannel> getAllowedChannels();
+
+    /**
+     * Gets whether or not this question suppresses messages.
+     *
+     * @return Whether or not messages are suppressed.
+     */
+    Tristate suppressesMessages();
```
ST-DDT: I'm not sure whether the Tristate is the real way to go maybe it should jsut return a primitive boolean based on the dialogues settings.
IMO it does not matter where the setting is set as long as it is set.

src/main/java/org/spongepowered/api/text/dialogue/AdditionalAnswerProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+ * A functional interface that processes answers without deciding the next
+ * {@link Question}.
+ */
+@FunctionalInterface
+public interface AdditionalAnswerProcessor {
+
+    /**
+     * Processes an {@link Question}.
+     *
+     * @param dialogue The current dialogue.
```
kashike: No period.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,198 @@
...
+     * in the same {@link Dialogue}.
+     *
+     * @return Whether or not output to speakers in the same conversation is
+     * suppressed
+     */
+    boolean suppressesOutputToPeers();
+
+    /**
+     * Gets the channels that are allowed through {@link
+     * #suppressesAllMessages()}. This {@link Collection} is immutable.
```
kashike:
```java
    /**
     * Gets a collection of message channels which are allowed
     * through {@link #suppressesAllMessages()}.
     *
     * <p>The returned{@link Collection} is immutable.</p>
     *
     * @return The allowed channels
     */
```

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,205 @@
...
+    default Text toText() {
+        return getText();
+    }
+
+    /**
+     * Gets the {@link AnswerProcessor} to use when processing {@link Answer}s.
+     *
+     * @return The processor
+     */
+    AnswerProcessor getMainProcessor();
```
kashike: -> `getProcessor`, "Main" is redundant since the other method has "Additional" in it.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Gets the {@link AdditionalAnswerProcessor}s to use when processing
+     * {@link Answer}s. This {@link Collection} is immutable.
```
kashike:
```java
    /**
     * Gets the {@link AdditionalAnswerProcessor}s to use when processing
     * {@link Answer}s.
     *
     * <p>The returned {@link Collection} is immutable.</p>
     *
     * @return The processor
     */
```

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
ryantheleach: I'd rather have a plugin supplied buffer / playback, with default options available in the API, Allowing the plugin creating the conversation the power to redirect chat, choose an appropriate buffer length, and whether replaying will happen, and at what speed, etc

src/main/java/org/spongepowered/api/text/dialogue/AdditionalAnswerProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A functional interface that processes answers without deciding the next
+ * {@link Question}.
+ */
+@FunctionalInterface
+public interface AdditionalAnswerProcessor {
+
+    /**
+     * Processes an {@link Question}.
```
12AwsomeMan34: An Question.

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Each {@link DialogueArchetype} is made up of a dynamic tree of {@link Question}s,
+ * meaning that each {@link Question} chooses the next one. A {@link Question}
+ * is responded to by a {@link Speaker} with an {@link Answer}.
+ */
+public interface Question extends TextRepresentable {
+
+    /**
+     * Creates a {@link Builder} to return an {@link Question}.
```
12AwsomeMan34: An Question. (again)

src/main/java/org/spongepowered/api/text/dialogue/Question.java
```diff
@@ -0,0 +1,205 @@
...
+         * @param processor The processor to use
+         * @return This builder, for chaining
+         */
+        Builder processor(AnswerProcessor processor);
+
+        /**
+         * Adds extra processors to use when processing {@link Answer}s. They
+         * cannot determine the next {@link Question}.
+         *
+         * @param processors The processors to use.
```
12AwsomeMan34: No period here either. :P

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
ST-DDT: @ryantheleach If I get your right your are proposing that we remove
* `suppressesOutput()`
* `suppressesAllMessages()`
* `suppressesOutputToPeers()`
* `getAllowedChannels()`

In favor of a new method

* `getOtherChatHandler() : DialogueChatHandler`

**DialogueChatHandler**
````java
public interface DialogueChatHandler {

    /**
     * Processes the given chat details. This method is called after the last chat event 
     * was processed but before the message is actually send to the client.
     *
     * <p>Implementations may queue the given (some) messages to send it after the dialogue is finished.</p>
     * 
     * @param dialogue The dialogue that is currently open.
     * @param details The details of the chat that has been send to a Speaker and has been intercepted.
     * @returns True, if the intercepted message should be forwarded to the client. 
     *             False, if the message was or will be handled by this instance or should be discarded.
     */
    public boolean process(Dialogue dialogue, ChatDetails details);

    /**
     * Notifies this handler of the end of the dialogue.
     * 
     * <p>This method can be used to send queued messages after the dialogue has been completed.</p>
     */
    public void notifyEnd(Dialogue dialogue);

}
````
**ChatDetails**
````java
public interface ChatDetails {

    public Text getMessage();

    public ChatType getType();

    public MessageChannel getChannel();

    [...]

}
````

File: src/main/java/org/spongepowered/api/text/dialogue/Question.java (hunk 28 above)
Comment (pie-flavor) [inline src/main/java/org/spongepowered/api/text/dialogue/Question.java]: @ST-DDT Questions are not necessarily aware of their Dialogue. Tristate is used to allow for the DialogueArchetype to define one; note that the identical methods in DialogueArchetype are primitive boolean. It was either Tristate or @Nullable Boolean.

**Comment by pie-flavor** [comment]
@ryantheleach I like it. Let's do it.

File: src/main/java/org/spongepowered/api/text/dialogue/AnswerProcessor.java [hunk 36]
```diff
@@ -0,0 +1,47 @@
...
+     * Processes an {@link Answer} and returns the next {@link Question} in
+     * the {@link Dialogue}.
+     *
+     * @param dialogue The current dialogue
+     * @param question The question being answered
+     * @param answer The answer to the question
+     * @return The next {@link Question} in the {@link Dialogue}. If {@link
+     * Optional#empty()}, the dialogue is concluded.
+     */
+    Optional<Question> process(Dialogue dialogue, Question question, Answer answer);
```
evilmidget38: The answer already contains references to both the Dialogue and the Question.  IMO this would be better simplified to just accepting an Answer.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,157 @@
...
+    Dialogue start(Speaker... speakers);
+
+    /**
+     * Creates a new {@link Dialogue} for the specified {@link Speaker}(s).
+     *
+     * @param speakers The speakers to send the initial {@link Question} to
+     * @return The newly created {@link Dialogue}
+     * @throws IllegalArgumentException If {@code speakers} is empty
+     */
+    Dialogue start(Iterable<? extends Speaker> speakers);
```
evilmidget38: I'm a little confused as to the behavior of Dialogues that contain multiple Speakers.  Each Speaker receives the initial question, but what happens after that?  When a user responds to a question is the next question then sent to every user?  This is further confused by Dialogue#getCurrentQuestionFor which makes it seem like speakers could potentially each have their own question within a dialogue.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Adds a new {@link Speaker} to this instance. The speaker will
+     * subsequently receive all {@link Question}s.
+     *
+     * @param speaker The speaker to add
+     */
+    void addSpeaker(Speaker speaker);
```
evilmidget38: Can a Speaker be removed from a dialogue other than through the Dialogue ending naturally?

src/main/java/org/spongepowered/api/text/dialogue/AnswerProcessor.java
pie-flavor: @evilmidget38 It's for convenience, mostly.

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
pie-flavor: When a Speaker responds to a question, the answer is sent to the AnswerProcessor. The next question returned by that AnswerProcessor goes to the Speaker. The difference between one Dialogue with many Speakers and many dialogues with one speaker each is the DataContainer inside the Dialogue object.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
pie-flavor: If an AnswerProcessor returns Optional.empty(), then the person is removed from the Dialogue. If there are no speakers left, the Dialogue is then concluded.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -66,4 +69,24 @@
...
     Optional<Question> getCurrentQuestionFor(Speaker speaker);
 
+    /**
+     * Stops the dialogue, removing all {@link Speaker}s.
+     *
+     * @param message The message to send as notification to the speakers
+     * @param suppress Whether or not to suppress the next chat message
+     * @throws IllegalStateException If the dialogue is already concluded
+     */
+    void conclude(Text message, boolean suppress);
```
ryantheleach: Shouldn't conclude restore whatever suppression was in place before the first dialogue started?

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Adds a new {@link Speaker} to this instance. The speaker will
+     * subsequently receive all {@link Question}s.
+     *
+     * @param speaker The speaker to add
+     * @throws IllegalStateException If the speaker is already part of the
+     * dialogue
+     */
+    void addSpeaker(Speaker speaker);
```
ST-DDT: If this throws an Exception then there needs to be a containsSpeaker method as well.
Or is `Speaker#isSpeaking()` enough?

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return The data
+     */
+    DataContainer getData();
+
+    /**
+     * Adds a new {@link Speaker} to this instance. The speaker will
+     * subsequently receive all {@link Question}s.
+     *
+     * @param speaker The speaker to add
+     * @throws IllegalStateException If the speaker is already part of the
```
ST-DDT: What happens if the Speaker is already part of a different dialogue?

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
ST-DDT: Also this method should return the DataContainer.
Because the data is all that is left with use after the dialogue is done.

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandler.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * <p>This method can be used to send queued messages after the dialogue
+     * has been completed. Please ensure that they are sent to the correct
+     * {@link MessageChannel}.</p>
+     *
+     * @param dialogue The dialogue that was concluded
+     */
+    void notifyEnd(Dialogue dialogue);
+
+}
```
ST-DDT: Missing newline at EOF.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
pie-flavor: Speaker has a getCurrentDialogue(), and they are such that an == comparison would work.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
pie-flavor: The DataContainer can be gotten from the Dialogue. It hasn't been destroyed or anything.

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandlers.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static class QueueingChatHandler implements DialogueChatHandler {
+
+        final ListMultimap<Dialogue, ChatDetails> map = ArrayListMultimap.create();
+        final int max;
+        final boolean hasLimit;
+
+        public QueueingChatHandler() {
+            this.max = 0;
+            this.hasLimit = false;
```
ST-DDT: IMO this flag is redundant as it is basically just:
`this.max == 0` although `this.max == Interger.MAX_VALUE` would be even better.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
ST-DDT: The suppress option should at least be better documented.
* The message of the former speakers is suppressed. First only? Per speaker? Time bound? Maybe there should be a 5 second timeframe where the server asks whether the client really eants to send the message.

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return The channel
+     */
+    MessageChannel getChannel();
+
+    /**
+     * Gets whatever sent the message.
+     *
+     * @return The sender, if there was one
+     */
+    Optional<Object> getSender();
```
kashike: Why `Object`? Why not make `ChatDetails` generic?

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
```diff
@@ -0,0 +1,66 @@
...
+package org.spongepowered.api.text.dialogue;
+
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.text.chat.ChatType;
+
+import java.util.Optional;
+
+/**
+ * A container class for use in {@link DialogueChatHandler}s.
```
kashike: This could be useful in cases outside of dialogues, you could bump this up to the `org.spongepowered.api.text.dialogue` package.

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public interface ChatDetails {
+
+    /**
+     * Gets the message that was sent in chat.
+     *
+     * @return The message
+     */
+    Text getMessage();
```
kashike: `getText` imo

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return The type of chat
+     */
+    ChatType getType();
+
+    /**
+     * Gets the channel that the chat was sent to.
+     *
+     * @return The channel
+     */
+    MessageChannel getChannel();
```
kashike: I'd make this `Optional`, for the above comment regarding package.

src/main/java/org/spongepowered/api/text/dialogue/Dialogue.java
```diff
@@ -0,0 +1,92 @@
...
+    void conclude(Text message, boolean suppress);
+
+    /**
+     * Removes a speaker from this dialogue.
+     *
+     * @param message The message to send as notification to the speakers
+     * @param suppress Whether or not to suppress the next chat message
+     * @param speaker The speaker to remove
+     * @throws IllegalStateException If the dialogue is already concluded, or
+     * if the speaker is not part of the dialogue
```
kashike: It'd be nice if you indent continuation lines by four spaces (all cases of this in the PR):
```java
    /**
     * Removes a speaker from this dialogue.
     *
     * @param message The message to send as notification to the speakers
     * @param suppress Whether or not to suppress the next chat message
     * @param speaker The speaker to remove
     * @throws IllegalStateException If the dialogue is already concluded, or
     *     if the speaker is not part of the dialogue
     */
```

src/main/java/org/spongepowered/api/text/dialogue/DialogueArchetype.java
```diff
@@ -0,0 +1,188 @@
...
+
+        /**
+         * Adds multiple handlers for when the dialogue concludes. This is
+         * repeatable (but likely unnecessarily so).
+         *
+         * @param handlers The handlers to add
+         * @return This builder, for chaining
+         */
+        default Builder conclusionHandlers(Iterable<DialogueConclusionHandler> handlers) {
+            checkNotNull(handlers).forEach(this::conclusionHandler);
```
kashike: `checkNotNull(handlers, "handlers")`

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandler.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Notifies this handler of the end of the dialogue.
+     *
+     * <p>This method can be used to send queued messages after the dialogue
+     * has been completed. Please ensure that they are sent to the correct
+     * {@link MessageChannel}.</p>
+     *
+     * @param dialogue The dialogue that was concluded
+     */
+    void notifyEnd(Dialogue dialogue);
```
kashike: Is `finish` or `complete` a better name here?

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandlers.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Creates a new {@link DialogueChatHandler} to queue all messages up to a
+     * specified maximum, and send them all at the {@link Dialogue}'s
+     * conclusion.
+     *
+     * @param max The maximum number of messages to queue
+     * @return The new handler
+     */
+    public static DialogueChatHandler queue(int max) {
+        checkArgument(max > 0, "max > 0");
```
kashike: `checkArgument(max > 0, "max must be greater than 0");`

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandlers.java
```diff
@@ -0,0 +1,116 @@
...
+        return new QueueingChatHandler(max);
+    }
+
+    private static class DiscardAllChatHandler implements DialogueChatHandler {
```
kashike: Why make this `private`? Someone might wish to `instanceof`, or extend it.

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandlers.java
```diff
@@ -0,0 +1,116 @@
...
+        public boolean process(Dialogue dialogue, ChatDetails details) {
+            return false;
+        }
+
+        @Override
+        public void notifyEnd(Dialogue dialogue) {}
+
+    }
+
+    private static class QueueingChatHandler implements DialogueChatHandler {
```
kashike: Why make this `private`? Someone might wish to `instanceof`, or extend it.

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandlers.java
```diff
@@ -0,0 +1,116 @@
...
+
+        final ListMultimap<Dialogue, ChatDetails> map = ArrayListMultimap.create();
```
kashike: `protected` for above comment (as well as other fields here)

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
pie-flavor: @kashike I'm not sure what you mean.

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
pie-flavor: @kashike Let's save that one for a rainy day.

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
kashike: Make `ChanDetails` generic, use the generic here instead of `Object`.
https://docs.oracle.com/javase/tutorial/java/generics/

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
kashike: Best to do it now instead of after this PR is pulled, if at all, as it is one less breaking change in future.

src/main/java/org/spongepowered/api/text/dialogue/ChatDetails.java
me4502: @pie-flavor, To give you an example on this, 

`public` interface ChatDetails<T> extends TextRepresentable {

and...

`Optional<T> getSender();

If you wanted to ensure that T extends another class, such as CommandSource, you'd use:
`<T extends CommandSource>
only in the class declaration.

20zinnm:
So, what do we need to move forward with this? Very excited.

src/main/java/org/spongepowered/api/text/dialogue/DialogueChatHandler.java
20zinnm: @kashike I'd say completed. There's some [funny stories](http://theanthrotorian.com/culture/2012/11/29/the-difference-between-complete-and-finished) about that, but generally completed gives the feel that all goals have been accomplished. By definition, they are exactly the same.

> When you marry a right woman you are complete. When you marry a wrong woman you are finished. When the right woman sees you with the wrong woman, you are completely finished.

In the above quote, there is a rather stark difference between the two, but in these sentences:

> The boy finished his homework.
The boy completed his homework.

they lend the exact same meaning.

20zinnm: Basically, finished means brought to an end whereas completed means all goals were accomplished. You can be finished without being complete.

gabizou:
Implementation would be required. As I've said before on various occasions, we're not at the point where we're going to accept new features unless the implementation is PR'ed and working.

20zinnm:
Ah. I see. I'll look into doing that.

src/main/java/org/spongepowered/api/text/dialogue/AdditionalAnswerProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+public interface AdditionalAnswerProcessor {
+
+    /**
+     * Processes a {@link Question}.
+     *
+     * @param dialogue The current dialogue
+     * @param question The question being answered
+     * @param answer The answer to the question
+     */
+    void process(Dialogue dialogue, Question question, Answer answer);
```
evilmidget38: I feel like providing the dialogue and Question here is a little redundant and useless.  Answer exists as a container type containing both the Dialogue and Question, and is referenced only in the AnswerProcessors.  AdditionalAnswerProcessor could be replaced by a Consumer<Answer>, and AnswerProcessor by a Function<Answer, Optional<Question>>.

You mentioned before that you provide them for convenience, but I find that excessive arguments is generally an inconvenience when it comes to functional interfaces.  If you wanted to use a method reference as an AnswerProcessor, you'd need every one of your methods to accept all three arguments, rather than just the one actually needed argument.  You're limited purely to your own implementations of AnswerProcessor.  By providing only the Answer, you allow interaction with anything that's a generic Consumer.

Deamon5550:

caseif:
@pie-flavor Do you still intend to complete this PR?

pie-flavor:
@caseif Whatever sort of creative thread I had going with this, I've lost it. Feel free to take what's there and continue it yourself.

PR #1301, opened 2016-07-04
Added GravityData and removed gravity from ArmorStandData

### LINKED ISSUE #1300: The NoGravity tag applies to all entites in 1.10
The NoGravity tag applies to all entities in 1.10. The tag used to only apply to armor stands, hence the data key `ARMOR_STAND_HAS_GRAVITY`. I propose that this should be changed to something like `ENTITY_HAS_GRAVITY`, and function on all entities.

Flibio:
**[API](https://github.com/SpongePowered/SpongeAPI/pull/1301)** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/828)

Adds `GravityData` and removes gravity from `ArmorStandData`. This is due to a change in Minecraft  1.10 that allows all entities, not just armor stands, to be affected by the `NoGravity` tag.

Resolves #1300. 

```java
@Listener
public void onPlayerJoin(ClientConnectionEvent.Join event) {
    Player player = event.getTargetEntity();
    // Prints the current value
    System.out.println(player.get(Keys.ENTITY_HAS_GRAVITY).get());
    // Turns off gravity for the player
    player.offer(Keys.ENTITY_HAS_GRAVITY, false);

    // Test the DataManipulator
    Optional<GravityData> optional = player.get(GravityData.class);
    if(optional.isPresent()) {
        Value<Boolean> value = optional.get().gravity();
        // Prints the current value (false)
        System.out.println(value.get());
        // Turns gravity back on for the player
        player.offer(optional.get().set(value.set(true)));
    }
}
```

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
@@ -60,6 +60,7 @@ import org.spongepowered.api.data.manipulator.mutable.entity.FoodData;
 import org.spongepowered.api.data.manipulator.mutable.entity.FuseData;
 import org.spongepowered.api.data.manipulator.mutable.entity.GameModeData;
 import org.spongepowered.api.data.manipulator.mutable.entity.GlowingData;
+import org.spongepowered.api.data.manipulator.mutable.entity.GravityData;
 import org.spongepowered.api.data.manipulator.mutable.entity.GriefingData;
 import org.spongepowered.api.data.manipulator.mutable.entity.HealthData;
 import org.spongepowered.api.data.manipulator.mutable.entity.HorseData;
@@ -307,6 +308,11 @@ public final class CatalogEntityData {
      * <!-- TODO: Find all non-effected entities -->
      */
     public static final Class<GlowingData> GLOWING_DATA = GlowingData.class;
+    /**
+     * Signifies that an entity will ignore gravity. Usually applies to all
+     * known types of entities.
+     */
+    public static final Class<GravityData> GRAVITY_DATA = GravityData.class;
     /**
      * Signifies that an entity can modify blocks in the world. Usually applies
      * to {@link Enderman} and {@link Humanoid}s.
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableGravityData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableGravityData.java
@@ -25,33 +25,22 @@
 package org.spongepowered.api.data.manipulator.immutable.entity;
 
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
-import org.spongepowered.api.data.manipulator.mutable.entity.ArmorStandData;
+import org.spongepowered.api.data.manipulator.mutable.entity.GravityData;
 import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.entity.Entity;
 
-public interface ImmutableArmorStandData extends ImmutableDataManipulator<ImmutableArmorStandData, ArmorStandData> {
-
-    ImmutableValue<Boolean> marker();
-
-    /**
-     * Returns whether this armor stand is a small armor stand or not.
-     *
-     * @return Whether this is a small armor stand
-     */
-    ImmutableValue<Boolean> small();
+/**
+ * An {@link ImmutableDataManipulator} for the gravity state. If the value is
+ * true, the {@link Entity} will have gravity.
+ */
+public interface ImmutableGravityData extends ImmutableDataManipulator<ImmutableGravityData, GravityData> {
 
     /**
-     * Returns whether this armor stand shows arms or not.
-     * <p>Arms that do not show may also not show an item in hand.</p>
+     * Gets the {@link ImmutableValue} of the gravity of an {@link Entity}.
+     * Returns true when the {@link Entity} has gravity.
      *
-     * @return Whether this armor stand shows its arms
+     * @return The immutable value of the gravity
      */
-    ImmutableValue<Boolean> arms();
-
-    /**
-     * Gets whether this armor stand has a visible base plate or not.
-     *
-     * @return Whether this armor stand has a visible base plate
-     */
-    ImmutableValue<Boolean> basePlate();
+    ImmutableValue<Boolean> gravity();
 
 }
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/GravityData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/GravityData.java
@@ -22,36 +22,25 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.data.manipulator.immutable.entity;
+package org.spongepowered.api.data.manipulator.mutable.entity;
 
-import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
-import org.spongepowered.api.data.manipulator.mutable.entity.ArmorStandData;
-import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableGravityData;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.Entity;
 
-public interface ImmutableArmorStandData extends ImmutableDataManipulator<ImmutableArmorStandData, ArmorStandData> {
-
-    ImmutableValue<Boolean> marker();
-
-    /**
-     * Returns whether this armor stand is a small armor stand or not.
-     *
-     * @return Whether this is a small armor stand
-     */
-    ImmutableValue<Boolean> small();
+/**
+ * A {@link DataManipulator} for gravity. If the value is true, the
+ * {@link Entity} will have gravity.
+ */
+public interface GravityData extends DataManipulator<GravityData, ImmutableGravityData> {
 
     /**
-     * Returns whether this armor stand shows arms or not.
-     * <p>Arms that do not show may also not show an item in hand.</p>
+     * Gets the {@link Value} of the gravity of an {@link Entity}. Returns true
+     * when the {@link Entity} has gravity.
      *
-     * @return Whether this armor stand shows its arms
+     * @return The value of the gravity
      */
-    ImmutableValue<Boolean> arms();
-
-    /**
-     * Gets whether this armor stand has a visible base plate or not.
-     *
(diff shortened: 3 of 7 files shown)


ST-DDT:
Well this is just my personal opinion, but for compatibility reasons, you should just mark the methods and `Key` constant as deprecated.
In fact you could "bind" them to the same value.

Flibio:
@ST-DDT Since this is going into bleeding, and will be apart of API 5, I though it would make sense to just remove the methods.

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableGravityData.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface ImmutableGravityData extends ImmutableDataManipulator<ImmutableGravityData, GravityData> {
+
+    /**
+     * Gets the {@link ImmutableValue} of the gravity state of an
+     * {@link Entity}.
+     *
+     * @return The immutable value of the gravity state
+     */
+    ImmutableValue<Boolean> ignoreGravity();
```
kashike: should be `gravity`,  and return `true` when the Entity has gravity

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/GravityData.java
```diff
@@ -0,0 +1,45 @@
...
+ * {@link Entity} will ignore gravity.
+ */
+public interface GravityData extends DataManipulator<GravityData, ImmutableGravityData> {
+
+    /**
+     * Gets the {@link Value} of the gravity state of an {@link Entity}.
+     *
+     * @return The value of the gravity state
+     */
+    Value<Boolean> ignoreGravity();
```
kashike: see above comment

Flibio:
I'm ready to squash the commits, if everything looks good.

kashike:
@Flibio see my comments

Flibio:
@kashike I've resolved both in b3cad6e.

Flibio:
@kashike They were both resolved, is everything good now?

ZephireNZ:
You changed the javadocs but didn't change the name.

Should be `gravity()`, not `ignoreGravity()`

Flibio:
@kashike  @ZephireNZ 
My apologies, I thought the comments were both referencing the Javadocs. The method names have been changed.

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableGravityData.java
```diff
@@ -0,0 +1,46 @@
...
+package org.spongepowered.api.data.manipulator.immutable.entity;
+
+import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.manipulator.mutable.entity.GravityData;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.entity.Entity;
+
+/**
+ * An {@link ImmutableDataManipulator} for the gravity state. If the value is
+ * true, the {@link Entity} is will ignore gravity.
```
Aaron1011: This needs to be updated

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/GravityData.java
```diff
@@ -0,0 +1,46 @@
...
+package org.spongepowered.api.data.manipulator.mutable.entity;
+
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableGravityData;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.Entity;
+
+/**
+ * A {@link DataManipulator} for gravity. If the value is true, the
+ * {@link Entity} will ignore gravity.
```
Aaron1011: Same

Flibio:
@Aaron1011 What is wrong with it? I'm unsure what I must change.

Aaron1011:
@Flibio: The comment says "If the value is true, the * {@link Entity} will ignore gravity" - but if it's true, gravity is turned on.

Flibio:
@Aaron1011: I'll get that fixed right away.

Flibio:
All the comments have been resolved. I'm ready to squash the commits, if everything looks good.

Flibio:
Is there anything holding this PR back? I could use this for a plugin I am working on.

gabizou:
Since all `Entity` instances can have this flag usable, could you add a default getter on `Entity` and then override in the implementation (as exemplified before) for ease of use? After that's done, squash these commits.

Flibio:
@gabizou I've added the method `hasGravity` to `Entity`.

src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogEntityData.java
```diff
@@ -312,6 +313,11 @@
      */
     public static final Class<GlowingData> GLOWING_DATA = GlowingData.class;
     /**
+     * Signifies that an entity will ignore gravity. Usually applies to all
```
kenzierocks: What's the "Usually" for, it does apply to all entities in all cases, right?

gabizou: In the event a modded entity doesn't follow normal MC mechanics.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -531,4 +531,11 @@ default boolean canSee(Entity entity) {
         Optional<Boolean> optional = entity.get(Keys.VANISH);
         return !optional.isPresent() || !optional.get();
     }
+
+    /**
+     * Returns whether this entity has gravity.
+     *
+     * @return True if this entity has gravity
+     */
+    boolean hasGravity();
```
gabizou: Default implement this method with a getter through Data API.

Flibio:
@gabizou I've got that changed.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -531,4 +531,14 @@ default boolean canSee(Entity entity) {
...
         return !optional.isPresent() || !optional.get();
     }
+
+    /**
+     * Returns whether this entity has gravity.
+     *
+     * @return True if this entity has gravity
+     */
+    default boolean hasGravity() {
+        Optional<Boolean> optional = this.get(Keys.ENTITY_HAS_GRAVITY);
```
gabizou: Just do `return get(Keys.ENTITY_HAS_GRAVITY).get()`

Flibio:
@gabizou I've got that changed.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -531,4 +531,13 @@ default boolean canSee(Entity entity) {
         Optional<Boolean> optional = entity.get(Keys.VANISH);
         return !optional.isPresent() || !optional.get();
     }
+
+    /**
+     * Returns whether this entity has gravity.
+     *
+     * @return True if this entity has gravity
+     */
+    default boolean hasGravity() {
```
ZephireNZ: @gabizou `hasGravity` or just `gravity`?

ZephireNZ: Also, pretty sure it's meant to follow the Value<> pattern - https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/entity/living/monster/Creeper.java#L36

gabizou: I'd keep it as `gravity()`. Refer to other Entity based interfaces where the default methods are included and such.

Flibio:
@ZephireNZ @gabizou I've updated the name of the method.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -531,4 +531,13 @@ default boolean canSee(Entity entity) {
         Optional<Boolean> optional = entity.get(Keys.VANISH);
         return !optional.isPresent() || !optional.get();
     }
+
+    /**
+     * Returns whether this entity has gravity.
+     *
+     * @return True if this entity has gravity
+     */
+    default boolean gravity() {
```
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1301/files#diff-2e9625646d911e5b538c4035decadb9eR92

Flibio:
@ZephireNZ @gabizou @kashike I've fixed the return value.

Flibio:
@gabizou Is there anything else that should be changed?

ZephireNZ:
If you could add a test, that would make testing and pulling easier:

- Tests that the datamanipulator can be fetched from a relevant entity, modified with keys, and then given.
- Tests that keys can be fetched from a relevant, and a new value can be passed.
- Test should be easily observable in the game, and/or be printed (can just use System.out.println, no need to be fancy)

You don't have to bother with making it a unit test, just a simple plugin class with command and/or event listeners will do.

ZephireNZ:
Oh, and don't add it to the pull commit, just copy/paste into the description.

Flibio:
@ZephireNZ I've added a small test.

ZephireNZ:
That tests the key getter/setter, but not the data.

Flibio:
@ZephireNZ: Sorry, I missed that. I've updated the example.

liach:
Why do you just call it `ENTITY_HAS_GRAVITY` instead of `GRAVITY`?

Flibio:
The key this is replacing is called `ARMOR_STAND_HAS_GRAVITY`, so I
followed that style.

Flibio:
@liach: The key this is replacing is called `ARMOR_STAND_HAS_GRAVITY`, so I followed that style.

ZephireNZ:
The data should be as general as possible. While the vanilla implementation of gravity may only work on entities, mods may add gravity for blocks, etc. Therefore you should avoid using the data's target in the name when possible. Or, as we've seen with this, newer versions may add or remove targets for data.

In hindsight the armour stand data could have just used `NO_GRAVITY` and avoided needing to change the key name with a change. 

correct, @gabizou ?

ryantheleach:
If gravity on blocks ever becomes a thing, I think that it's a different enough concept to be named differently, block gravity only updates on physics updates for example. I also suspect that it would be implemented as a blockstate, unless a future tile entity has support to have gravity.

gabizou:
Keep it as `GRAVITY`.

Flibio:
@gabizou: I've update the key name, let me know if this is ready to be squashed and merged.

Flibio:
@gabizou: Does anything else need to be changed?

Aaron1011:
@Flibio: Can you rebase this against `bleeding`?

Flibio:
@Aaron1011: I've rebased the branch, should I squash the commits now?

Aaron1011:
@Flibio: Yes, please.

Flibio:
@Aaron1011: Everything is squashed and ready to go.

Flibio:
Is there anything else I must do or is this ready? This would be useful for a plugin I am working on.

PR #1315, opened 2016-07-21
Refactor Block Change API

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/851) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/770)

This is a more focused PR at dealing with several issues at once, but for the sake of having not had a decent PR description in a while, I'll rehash the conversational description that we used to have a while back.

#### So what's changing?

In short, if a plugin developer is calling any API method that changes something, an event should be thrown. Traditionally, we used to have the inability to throw events for virtually everything, however, at this point in time, it's well within our means to throw events for all `setBlock`, `createEntity`, `spawnEntity`, etc. 

#### But why force plugin developers to supply a `Cause`? Don't developers know to throw events to check for things before they make a change?

Yes and no. It's been useful for developers to throw "dummy" events to check for permissions or block modifications before finalizing a change, and for other cases, to throw dummy spawn events before an entity is indeed spawned into the world. The problem with all of this is that it's entirely up to the developer whether to throw the events, or not throw the events, and other plugins are none the wiser. This is especially complicated by the fact that Sponge has a very complex tracking system that performs various actions based on pre-defined expectations. The change resolves a great many event cause tracking determining aspects, such as a synchronous task set up by plugin X to make a tree due to player Y. 

#### Could you explain why this wasn't an issue in API 4.X?

The issue lies in the API and the implementation. Originally planned, the CauseTracker in 1.8.9 does a very good job at determining what is causing X, Y, and Z happening, when it's *expected*. Plugins that call our methods to spawn entities are bypassing the CauseTracker, but only to a limited extent, there's no `Cause` required or provided, and other plugins have no knowledge of why an entity was being spawned. Or for that matter, the CauseTracker can flip out if a multitude of block changes take place in a synchronized task, but this isn't the fault of the CauseTracker, more so our contracts we set forth in the API. If I had to pinpoint why the change is necessary, it's that the CauseTracker is more of a mechanism to determine causes for which things happen.

#### So what actually is changing API wise?

In short, any method that changes a block in a `World` (or `Extent`) will require a `Cause`. Virtually anything that touches changing blocks will now provide a `Cause` to the system, and the system will be able to apply that provided `Cause` to the ensuing `ChangeBlockEvent` (unless this is a populator, which can simply provide a recycled `Cause` as shown in the skylands populators and generators). Data API has gained an overloaded method that will provide a `Cause` for the cases that the `#offer(Key,Object,Cause)` will change a block in a `World`, however, normal offers that simply change data on a `TileEntity` can still use the traditional `#offer(Key,Object)` method. It is understandable at this point as well that the other `DataHolder` implementations may throw Data related events with the provided `Cause`.

#### What else? What's this `BlockChangeFlag`?

Ah, so that is a brainchild of the various `boolean` flags that we used to require plugin developers to provide, namely to show that there are various *options* to perform a block change. The reasoning for it is that previously, we could only allow developers to *NOT* notify neighboring blocks on a block change, but we didn't allow it to *NOT* perform block physics on placement. This caused issues with some `BlockSnapshot#restore`s, since the restore process would always perform block physics due to the nature of the implementation. At this point, instead of adding yet another `boolean` flag and set of overloads, it was simpler to use an object to determine whether neighbor notifications should take place, and if block physics should take place. In the future, more flags may be added as necessary, and this time without breaking the API ;).

#### With all of the `Cause` usage, why does this look like the implementation is leaking into the API?

It is and isn't. Due to the nature of our tracking system, this is the best way that we can achieve a proper API contract of "all your changes throw events". Likewise, this eliminates the issues of being able to deterministically resolve what is performing changes to the world, without some obscure `Cause` containing a `Task` and or a `PluginContainer`. This also eliminates any sort of tracking required on our part to *determine* what caused those changes. This ends up with the API being more contractually sound with the end result that plugins no longer have to throw dummy events and the implementation has no requirement to track and determine what is going on throughout different phases.

Any further questions can be asked below.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockSnapshot.java
+++ b/src/main/java/org/spongepowered/api/block/BlockSnapshot.java
@@ -32,6 +32,7 @@ import org.spongepowered.api.data.DataView;
 import org.spongepowered.api.data.ImmutableDataBuilder;
 import org.spongepowered.api.data.LocatableSnapshot;
 import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+import org.spongepowered.api.world.BlockChangeFlag;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.storage.WorldProperties;
@@ -116,11 +117,11 @@ public interface BlockSnapshot extends LocatableSnapshot<BlockSnapshot> {
      *
      * @param force If true, forces block state to be set even if the
      *     {@link BlockType} does not match the snapshot one.
-     * @param notifyNeighbors If true, notifies neighboring blocks to update
-     *     physics
-     * @return true if the restore was successful, false otherwise
+     * @param flag The block change flags to determine whether neighbors are
+     *     notified, block physics performed, etc.
+     * @return True if the restore was successful, false otherwise
      */
-    boolean restore(boolean force, boolean notifyNeighbors);
+    boolean restore(boolean force, BlockChangeFlag flag);
 
     /**
      * Gets the {@link UUID}, if available, of the user who created this
--- a/src/main/java/org/spongepowered/api/world/BlockChangeFlag.java
+++ b/src/main/java/org/spongepowered/api/world/BlockChangeFlag.java
@@ -1,0 +1,141 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world;
+
+import com.google.common.base.Objects;
+
+/**
+ * A flag of sorts that determines whether a block change will perform various
+ * interactions, such as notifying neighboring blocks, performing block phsyics
+ * on placement, etc.
+ */
+public enum BlockChangeFlag {
+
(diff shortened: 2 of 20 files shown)


src/main/java/org/spongepowered/api/block/BlockSnapshot.java
```diff
@@ -116,11 +117,11 @@ static Builder builder() {
      *
      * @param force If true, forces block state to be set even if the
      *     {@link BlockType} does not match the snapshot one.
-     * @param notifyNeighbors If true, notifies neighboring blocks to update
+     * @param flag If true, notifies neighboring blocks to update
```
Zidane: Docs need adjustments

src/main/java/org/spongepowered/api/world/extent/worker/BlockVolumeWorker.java
```diff
@@ -53,21 +54,21 @@
...
      * Applies a mapping operation to all the blocks in the volume and saves the
      * results to the destination volume.
-     *
-     * @param mapper The mapping operation
+     *  @param mapper The mapping operation
      * @param destination The destination volume
+     * @param cause
      */
-    void map(BlockVolumeMapper mapper, MutableBlockVolume destination);
+    void map(BlockVolumeMapper mapper, MutableBlockVolume destination, Cause cause);
```
gabizou: @DDoS I had to change some of these methods to use the `Cause` as I'm planning on implementation to use a customary phase to enter for all the block changes until the worker has completed applying the changes. I'm not 100% sure that I did it all right but, please do comment to clear it up for me.

src/main/java/org/spongepowered/api/block/BlockSnapshot.java
```diff
@@ -116,11 +117,11 @@ static Builder builder() {
      *
      * @param force If true, forces block state to be set even if the
      *     {@link BlockType} does not match the snapshot one.
-     * @param notifyNeighbors If true, notifies neighboring blocks to update
-     *     physics
-     * @return true if the restore was successful, false otherwise
+     * @param flag The block change flags to determine whether neighbors are
```
kashike: "The block change flags" -> "The block change flag", there is only one accepted here

gabizou: There are multiple flags in the object itself.

src/main/java/org/spongepowered/api/world/extent/worker/BlockVolumeWorker.java
DDoS: Shouldn't the cause be passed when getting the worker instead? That would leave the actual worker methods unchanged.

src/main/java/org/spongepowered/api/world/BlockChangeFlag.java
```diff
@@ -0,0 +1,141 @@
...
+ * A flag of sorts that determines whether a block change will perform various
+ * interactions, such as notifying neighboring blocks, performing block phsyics
+ * on placement, etc.
+ */
+public enum BlockChangeFlag {
+
+    ALL(Flags.NEIGHBOR_MASK | Flags.PHYSICS_MASK),
+    NEIGHBOR(Flags.NEIGHBOR_MASK),
+    PHYSICS(Flags.PHYSICS_MASK),
+    NEIGHBOR_PHYSICS(Flags.NEIGHBOR_MASK | Flags.PHYSICS_MASK),
```
liach: What's the difference of `NEIGHBOR_PHYSICS` and `ALL`?

kashike: There can be more flags added in the future, `NEIGHBOR_PHYSICS` includes both `NEIGHBOR` and `PHYSICS`.

liach: Then I think you should probably move `ALL` option to the last, or if you call `setUpdateNeighbors` on `PHYSICS` instance, it may return `ALL` instead of `NEIGHBOR_PHYSICS`.

gabizou: That really doesn't matter since when a new flag is added, it'll still work the exact same way, except `ALL` won't be matched because of the obvious bit masked flags.

PR #1323, opened 2016-07-31
Add ability to get/set state of CommandContext, update GenericArguments.firstParsing to discard parsed arguments

dualspiral:
Pinging @zml2008 as the commands guru!

When using `GenericArguments.firstParsing()` with a sequence of arguments, for example:

```java
GenericArguments.firstParsing(
    GenericArguments.seq(
        GenericArguments.string(Text.of("test1")),
        GenericArguments.string(Text.of("test2"))
    ),

    GenericArguments.string(Text.of("test3"))
)
```

if we run the command `/test test`, both variables "test1" and "test3" will be added to the `CommandContext` as "test", which is incorrect. This is because the state of the `CommandArgs` object is saved, but the `CommandContext` is not.

This PR adds the ability to save the state of the `CommandContext` in the same way that the `CommandArgs` object is saved, so that using the first parsing argument with a sequence of arguments will not have their parsed arguments bleed into each other.

A unit test has been added to verify this behaviour, and the code used to test this behaviour, along with output, is in the included gist.

https://gist.github.com/dualspiral/0ac16d62f265e45222c1a22784cc2831

I have targeted master/API 4.2.x as this is a bugfix that does not break any API methods.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandContext.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandContext.java
@@ -158,4 +158,29 @@ public final class CommandContext {
     public boolean hasAny(Text key) {
         return hasAny(ArgUtils.textToArgKey(key));
     }
+
+    /**
+     * Return this arguments object's current state. Can be used to reset with the {@link #setState(Object)} method.
+     *
+     * @return The current state
+     */
+    public Object getState() {
+        return ArrayListMultimap.create(this.parsedArgs);
+    }
+
+    /**
+     * Restore the arguments object's state to a state previously used.
+     *
+     * @param state the previous state
+     */
+    @SuppressWarnings("unchecked")
+    public void setState(Object state) {
+        // Check that the object we get back also only has String keys.
+        if (!(state instanceof Multimap) || !((Multimap<?, ?>) state).keySet().stream().allMatch(String.class::isInstance)) {
+            throw new IllegalArgumentException("Provided state was not of appropriate format returned by getState!");
+        }
+
+        this.parsedArgs.clear();
+        this.parsedArgs.putAll((Multimap)state);
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -158,4 +158,29 @@ public boolean hasAny(String key) {
     public boolean hasAny(Text key) {
         return hasAny(ArgUtils.textToArgKey(key));
     }
+
+    /**
+     * Return this arguments object's current state. Can be used to reset with the {@link #setState(Object)} method.
+     *
+     * @return The current state
+     */
+    public Object getState() {
```
kashike: Why return `Object` here? Why not `Multimap<String, Object>`?

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -158,4 +158,29 @@ public boolean hasAny(String key) {
...
+        return ArrayListMultimap.create(this.parsedArgs);
+    }
+
+    /**
+     * Restore the arguments object's state to a state previously used.
+     *
+     * @param state the previous state
+     */
+    @SuppressWarnings("unchecked")
+    public void setState(Object state) {
```
kashike: Why accept `Object` here? Why not `Multimap<String, Object>`?

src/main/java/org/spongepowered/api/command/args/CommandContext.java
dualspiral: It's a matter of the intention of the function. I'm mirroring what is done in the `CommandArgs` class, which has methods that get and set objects.

The point of these methods is to capture the state of the object so that it can be reveresed. The consumer of this API does not need to know what it is, only that it holds some state, likewise, they only need to know that they give it back to this object to reset it. By making it an object, we are kind of saying that, "you don't need to know what this, only that it holds the current state." It also means that if changes to this state are made, it's not necessarily an API breaking change.

src/main/java/org/spongepowered/api/command/args/CommandContext.java
dualspiral: Same reason as for `getState`

dualspiral:
I've just realised that #1284 also resolves this, but only for API 5. So, if you decide on pulling that PR, this will only be useful for API 4. Won't be offended if you feel this doesn't need to be pulled, should have checked other PRs before doing this!

Deamon5550:

PR #1325, opened 2016-08-10
Added SmeltableProperty

Redrield:
Added a boolean item property, which represents whether an item can be smelted in a furnace.

Implemented by [SpongeCommon #877](https://github.com/SpongePowered/SpongeCommon/pull/877)

### DIFF
--- a/src/main/java/org/spongepowered/api/data/property/item/SmeltableProperty.java
+++ b/src/main/java/org/spongepowered/api/data/property/item/SmeltableProperty.java
@@ -1,3 +1,27 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/data/property/item/SmeltableProperty.java
```diff
@@ -0,0 +1,38 @@
+package org.spongepowered.api.data.property.item;
```
kashike: Missing license header.

src/main/java/org/spongepowered/api/data/property/item/SmeltableProperty.java
```diff
@@ -0,0 +1,62 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.property.item;
+
+import org.spongepowered.api.data.property.BooleanProperty;
+
+/**
+ * Represents an item that can be smelted in a furnace
```
gabizou: Whether or not the item can be smelted in a furnace. Remember, this is a `BooleanProperty`, so the value itself can be true or false.

PR #1326, opened 2016-08-12
Intersected Message Channels

Wundero:
This is a simple pull request that adds support for message channels that, when created, combine messages channels in such a way that only members present in every channel are sent messages. This is quite useful for plugins interacting. I couldn't properly import this into my IDE so I had to write this in GitHub, hopefully it works properly.


zml2008:
Please test locally before making a PR -- this doesn't even compile. Also squash your commits.

imo this would be a useful addition if done efficiently -- it adds another set operation to MessageChannels.

Wundero:
I would have done it locally if I could, but I couldn't get it to clone to my IDE on this computer. I am currently in the process of moving so I can't properly do this, so I apologize.

zml2008:
uhhh it's just `git clone git@github.com:Wundero/SpongeAPI.git` then `./gradlew` not that hard

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -106,6 +107,30 @@ static MessageChannel combined(Collection<MessageChannel> channels) {
...
     /**
+     * Creates a message channel that targets subjects contained within every given
+     * channels and applies the message transformations of each channel in order.
+     *
+     * @param channels The channels to combine
+     * @return The channel
+     * @see IntersectedMessageChannel
+     */
+    static MessageChannel intersected(MessageChannel... channels) {
+        return new IntersectedMessageChannel(channels);
```
ST-DDT: Maybe this should check if there is only one element in the channel list and return that in this case.

src/main/java/org/spongepowered/api/text/channel/type/IntersectedMessageChannel.java
```diff
@@ -0,0 +1,96 @@
...
+        List<MessageChannel> channels = new ArrayList<>();
+        channels.addAll(this.channels);
+        if (channels.size() == 1) {
+            return channels.get(0).getMembers();
+        }
+        List<MessageReceiver> sr = new ArrayList<>();
+        List<MessageReceiver> tr = new ArrayList<>();
+        sr.addAll(channels.get(0).getMembers());
+        for (int i = 1; i < channels.size(); i++) {
+            for (MessageReceiver r : sr) {
```
ST-DDT: There is a wonderful method in [`Collection#retainAll(Collection)`](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#retainAll-java.util.Collection-) that does exactly what you need.

src/main/java/org/spongepowered/api/text/channel/type/IntersectedMessageChannel.java
```diff
@@ -0,0 +1,96 @@
...
+            text = channel.transformMessage(sender, recipient, text, type).orElse(text);
+        }
+
+        return Optional.ofNullable(text);
+    }
+
+    @Override
+    public Collection<MessageReceiver> getMembers() {
+        // Collects members present in all channels. (Intersecting members).
+        // Not sure how to do this with streams.
```
ST-DDT: Stream solution.
But I'm not sure whether this is the best solution:
````java
Predicate<MessageReceiver> filter = channels.stream()
    .skip(1)
    .map(Collection::contains)
    .collect(Collectors.reducing(Predicate::and)).orElseGet(() -> value -> true);
return channels.get(0).stream()
    .filter(filter)
    .collect(GuavaCollectors.toImmutableSet());
````

File: src/main/java/org/spongepowered/api/text/channel/type/IntersectedMessageChannel.java [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+
+    public IntersectedMessageChannel(Collection<MessageChannel> channels) {
+        this.channels = ImmutableSet.copyOf(channels);
+    }
+
+    @Override
+    public Optional<Text> transformMessage(@Nullable Object sender, MessageReceiver recipient, Text original, ChatType type) {
+        Text text = original;
+        for (MessageChannel channel : this.channels) {
+            text = channel.transformMessage(sender, recipient, text, type).orElse(text);
```
ST-DDT: I'm not sure whether the intersection MessageChannel should also transform the messages with all those other message channels.

src/main/java/org/spongepowered/api/text/channel/type/IntersectedMessageChannel.java
```diff
@@ -0,0 +1,96 @@
...
+ * channels,
+ * {@code channels[n-1].transformMessage(channels[n-2].transformMessage(channels[...]
+ * .transformMessage(channels[0].transformMessage(input))))} would occur)
+ */
+public class IntersectedMessageChannel implements MessageChannel {
+
+    protected final Collection<MessageChannel> channels;
+
+    public IntersectedMessageChannel(MessageChannel... channels) {
+        this(Arrays.asList(channels));
```
ST-DDT: I guess you need `checkNotNull(channels, "channels")` checks here.

src/main/java/org/spongepowered/api/text/channel/type/IntersectedMessageChannel.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public Optional<Text> transformMessage(@Nullable Object sender, MessageReceiver recipient, Text original, ChatType type) {
+        Text text = original;
+        for (MessageChannel channel : this.channels) {
+            text = channel.transformMessage(sender, recipient, text, type).orElse(text);
+        }
+
+        return Optional.ofNullable(text);
```
kashike: Should just be `Optional.of(text)`, it won't be `null` in this case.

Wundero: Well, this is how it was done in the CombinedMessageChannel, so I figured I would follow the same style.

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
JBYoshi: While we're at it, maybe return `MessageChannel.TO_NONE` if it's empty?

gabizou:
@kashike Is this PR ok to merge or are there changes that are needing to be made?

Zidane:

Issue #1137, opened 2016-03-12
SnapshotVolume for Schematic API usage


gabizou:
Currently, we have a pending issue for working on a proper Schematics API where by an interface is needed to extend `BlockVolume` and have the ability represent non-live `Entity` objects (possibly as alternates of `EntitySnapshot`s) and non-live `TileEntity` objects.

The limitation for both is that the `SnapshotVolume` (working name for now) cannot extend `Extent`. This is partially related to the #546 issue that has been sitting for a while now.

Some of the objects proposed needed for Schematics in general (And overall usability for plugin developers):

- `SnapshotVolume`: A volume that has blocks at coordinates, has a pseudo mutable `BlockSnapshot` like object at each position (optionally, as air is not represented)
- `FabricatedBlock`: A mutable `DataHolder` that has a `BlockState` and possibly `DataManipulator`s and a high possibility of directly supporting *current* NBT compound access and modification of `BlockEntity` information
- `FabricatedEntity`: A mutable `DataHolder` that has an `EntityType` and likely all related optional data to set on the entity (including inventory).

More to come as it is discussed.

20zinnm:
Can you reference the pending issue with the Schematic design? I'd like to get a feel for the format (if a sample is provided).

20zinnm:
Any update on this? It's critical to a plugin I'm writing. Might put a bounty on it.

kenzierocks:
There's a spec if you look at the SpongePowered repos.

20zinnm:
I don’t need a spec, I need an implementation.

gabizou:
> I don’t need a spec, I need an implementation.



> Any update on this? It's critical to a plugin I'm writing. Might put a bounty on it.

Patience is a virtue. It'll get done when it gets done.

gabizou:
@20zinnm If you're up for it, give it a shot of writing up the `SnapshotVolume`, the premise would be of course that it has a bounds, adds a new sort of `BlockBlueprint` of sorts that has various info, such as the raw `DataContainer` for the tile entity information, the `BlockType`, and any relative `DataManipulator`s can be used (or optionally `BlockState would suffice)`. As it cannot be using a `Location` object, it must use a customized `Vector3i` for the position *relative to a certain point*. Likewise, the same treatment would have to be applied for `Entity` objects as they cannot be `EntitySnapshot`s. It may be possible to explore the `Inventory` suffice that there's a way to implement custom inventories for such things. 

[...]

PR #1330, opened 2016-08-14
Schematics API, Closes #1142, #1137

### LINKED ISSUE #1142: Support for serializing sections of the world to disk

Deamon5550:
Implemented in spongepowered/SpongeCommon#882 Closes #1142, #1137 

This adds an api for creating `ArchetypeVolume`s which are copies of regions of the world and may be serialized to disk in one of two schematic formats (The MCEdit format here called the legacy format, and an updated format for better mod compatability).

An example plugin using this can be found [In the cookbook](https://github.com/SpongePowered/Cookbook/blob/schematics/Plugin/CopyPasta/src/main/java/org/spongepowered/cookbook/plugin/CopyPasta.java#L47).

The sponge schematic format specification draft can be found [here](https://github.com/SpongePowered/Schematic-Specification/blob/master/versions/schematic-1.md). Once this is merged that specification will become version 1.

The gist of the usage is
```java
// copying
ArchetypeVolume volume = player.getWorld().createArchetypeVolume(min, max, player.getLocation().getPosition().toInt());

// pasting
volume.apply(player.getLocation(), BlockChangeFlag.ALL, Cause.of(NamedCause.of("plugin", this.plugin), NamedCause.source(player)));

// saving
Schematic schematic = Schematic.builder().volume(data.getClipboard()).metaValue(Schematic.METADATA_AUTHOR, player.getName()).metaValue(Schematic.METADATA_NAME, name).paletteType(PaletteTypes.LOCAL).build();
schematicData = DataTranslators.SCHEMATIC.translate(schematic);
DataFormats.NBT.writeTo(new GZIPOutputStream(new FileOutputStream(outputFile)), schematicData);
```


ZephireNZ:
I'm guessing that `EntityArchetype`s were removed from `Schematic`s as the implementation was a bit of a pain for the time being?

kenzierocks:
Will a standalone implementation be created for things like WorldEdit (for Forge) to depend upon?

Deamon5550:
@ZephireNZ Yeah and they've traditionally never been included and I wasn't happy with how I was allowing the plugin to specify if they were included in the archetype volumes or not. Needs some more thought basically and it can be added later pretty easily.

@kenzierocks I could write a one file library based on flow-nbt or jnbt that people could include but its pretty simple to implement. SpongeCommon's specific impl isn't very useful for people because it depends on our DataContainers.

ZephireNZ:
Also, I added this to my PR, but really it should be here:

From EntityArchetype (can probably be added to BlockArchetype too)
```java
    /**
     * Creates a new {@link EntityArchetype} with the specified {@link EntityType}.
     * @param type Type of the entity
     * @return An archetype of the given entity type
     */
    static EntityArchetype of(EntityType type) {
        return builder().type(type).build();
    }
```

src/main/java/org/spongepowered/api/data/DataHolder.java
```diff
@@ -65,5 +65,5 @@
      * @throws InvalidDataException If the container is missing or has invalid
      *     data that this holder will refuse
      */
-    void setRawData(DataContainer container) throws InvalidDataException;
+    void setRawData(DataView container) throws InvalidDataException;
```
ZephireNZ: Btw this change breaks Forge's MixinFluidStack

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -526,4 +526,16 @@ default void setNotifier(Vector3i pos, @Nullable UUID uuid) {
...
     Set<AABB> getIntersectingCollisionBoxes(Entity owner, AABB box);
 
+    /**
+     * Creates a new archetype volume from the specified section of this extent.
+     * The archetype's volume will be shifted such that the position given in
+     * the origin will be the origin of the volume.
+     *
+     * @param min The minimum point of the area to copy
+     * @param max The maximum point of the area to copy
+     * @param origin The eventual origin on the new archetype volume
```
DDoS: I think the origin shift should be moved to an ArchetypeVolume method. It's not strictly necessary for creating it, and also someone might want to change it latter. An other option is just to offset it so that the `min` corner is equivalent to `0`.

Deamon5550: It feels like a bad idea to have the min and max of the archetype volume be mutable and if you change the offset you are also changing the min and max. And for ease of use we want the offset to always be applied to them such that iterating over the volume is always relative to the offset.

src/main/java/org/spongepowered/api/data/DataHolder.java
Deamon5550: Fixed it in a [branch of spongeforge](https://github.com/SpongePowered/SpongeForge/compare/feature/schematics)

src/main/java/org/spongepowered/api/data/DataQuery.java
```diff
@@ -192,6 +209,24 @@ public DataQuery pop() {
...
+     * returned.
+     *
+     * @return The next level query
+     */
+    public DataQuery popFirst() {
+        if (this.parts.size() <= 1) {
+            return of();
+        }
+        ImmutableList.Builder<String> builder = ImmutableList.builder();
+        for (int i = 1; i < this.parts.size(); i++) {
```
ST-DDT: Wouldn't be sublist() good enough?

src/main/java/org/spongepowered/api/data/persistence/DataTranslators.java
```diff
@@ -0,0 +1,89 @@
...
+    public static final DataTranslator<LocalDateTime> LOCAL_DATE_TIME = DummyObjectProvider.createExtendedFor(DataTranslator.class, "LOCAL_DATE_TIME");
+    public static final DataTranslator<Instant> INSTANT = DummyObjectProvider.createExtendedFor(DataTranslator.class, "INSTANT");
+    public static final DataTranslator<ZonedDateTime> ZONED_DATE_TIME = DummyObjectProvider.createExtendedFor(DataTranslator.class, "ZONED_DATE_TIME");
+
+    // sortfields:OFF
+
+    private DataTranslators() {
+    }
+
+}
```
ST-DDT: Missing newline.

src/main/java/org/spongepowered/api/data/DataQuery.java
kenzierocks: `sublist` might lead to a memory leak, but copying a sublist might work.

src/main/java/org/spongepowered/api/world/schematic/Palette.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return If the blockstate existed in the mapping
+     */
+    boolean remove(BlockState state);
+
+    /**
+     * Gets all {@link BlockState}s contained in this palette.
+     * 
+     * @return All contained block states
+     */
+    Collection<BlockState> getEntries();
```
ST-DDT: Maybe add a method getAsMap().
You can use it to default implement some of the methods then.

src/main/java/org/spongepowered/api/world/schematic/Palette.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    Optional<BlockState> get(int id);
+
+    /**
+     * Removes the given blockstate from the mapping.
+     * 
+     * @param state The blockstate to remove
+     * @return If the blockstate existed in the mapping
+     */
+    boolean remove(BlockState state);
```
ST-DDT: What happens if you remove a state from the GLOBAL palette type list(which by definition contains all)?

DDoS: Unsupported operation exception?

ST-DDT: Then this behaviour should be documented.
As an alternative this could change its type to local.

src/main/java/org/spongepowered/api/entity/EntityArchetype.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Gets the {@link EntityType} of the entity contained in this archetype.
+     * 
+     * @return The entity type
+     */
+    EntityType getType();
+
+    /**
+     * Gets the raw {@link Entity} data that would be applied to the generated
+     * tile entity. Note that this is a copied container.
```
simon816: entity, not tile entity

src/main/java/org/spongepowered/api/world/schematic/Palette.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.spongepowered.api.block.BlockState;
+
+import java.util.Collection;
+import java.util.Optional;
+
+/**
+ * Represents a mapping for blockstates to a local identifier.
+ */
+public interface Palette {
```
simon816: The name palette seems a bit too general, perhaps call it `BlockPalette`

PR #1344, opened 2016-08-28
Remove custom asset directories

Minecrell:
The asset manager allows changing the directory where all plugin assets are stored. This was originally added for flexibility, but I can personally not see any usecases to change the asset directory. On the other hand, there are two reasons why it should **not** be changed:

- Assets in Vanilla Minecraft and Forge are always stored in the `assets/id` directory
- Changing the asset directory could cause problems with resource packs in future client-side plugins (since that requires assets to be stored in the default directory)

This PR removes the option to set custom asset directories. Unless someone has a usecase for the option I don't see any reason to keep it, since the default directory is always the best choice.

### DIFF
--- a/src/main/java/org/spongepowered/api/plugin/PluginContainer.java
+++ b/src/main/java/org/spongepowered/api/plugin/PluginContainer.java
@@ -114,15 +114,6 @@ public interface PluginContainer {
     }
 
     /**
-     * Gets the directory that contains this {@link Plugin}'s assets.
-     *
-     * @return Asset directory, or empty if none
-     */
-    default Optional<Path> getAssetDirectory() {
-        return Optional.empty();
-    }
-
-    /**
      * Retrieves the {@link Asset} of the specified name from the
      * {@link AssetManager} for this {@link Plugin}.
      *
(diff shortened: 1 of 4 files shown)


windy1:
Like I said when I spoke to @Minecrell earlier, I am also inclined to agree that this should be removed. The purpose of this PR is to raise a discussion and possibly discover any use cases for this feature that others may have found that we have overlooked.

Zidane:
@Minecrell @windy1 

With that said, is the Asset API as a whole...useless? All plugins need is something akin to ResourceLocation, a way to register a custom resource loader.

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
```diff
@@ -114,15 +114,6 @@ default String getName() {
     }
 
     /**
-     * Gets the directory that contains this {@link Plugin}'s assets.
```
kashike: Why remove this? A plugin would still have an asset directory, no?

windy1:
No I don't think the whole Asset API is useless, it makes the handling of resources as you might use them in a plugin development environment easier and also for exchanging resources between plugins. Keep in mind, the entire purpose of the Asset API was for convenience in the first place and it's also not even really that much code when it comes down to it.

I agree with @Minecrell for removing this one feature of it because, as he brought to my attention, I can't even really think of a use case for it.

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
Minecrell: Well yes, but since it is always `assets/<plugin_id>` I don't think this method is still needed.

Minecrell:
@Zidane The `Asset` interface in the Asset API is essentially just a simplified version of `ResourceLocation` (it uses plugin containers as domain, and it is a special resource location because the URL is already known).

PR #1352, opened 2016-09-09
Add support for virtual custom biomes

Deamon5550:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/928) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/900)

This adds support for creating custom biomes which exist only during generation, afterwards they are stored as an existing non-virtual biome. This can be used to support more varied generation easier without creating full blown custom biomes which would require client modifications as well.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
+++ b/src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
@@ -1,0 +1,115 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
```diff
@@ -0,0 +1,64 @@
...
+
+    interface Builder extends ResettableBuilder<VirtualBiomeType, Builder> {
+
+        Builder name(String name);
+
+        Builder temperature(double temp);
+
+        Builder humidity(double humidity);
+
+        Builder persistedBiome(BiomeType biome);
```
kashike: Should be `persistedType` to be consistent with the getter in `VirtualBiomeType`.

src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
```diff
@@ -0,0 +1,64 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.biome;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A virtual biome is one which exists purely for generation and therefore
+ * requires no modifications to clients in order to use it. It after generation
```
gabizou: I think some bad wording: `It after generation the biome is persisted as another non-virtual biome.`

src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
```diff
@@ -0,0 +1,64 @@
...
+ * the biome is persisted as another non-virtual biome.
+ */
+public interface VirtualBiomeType extends BiomeType {
+
+    public static Builder builder() {
```
kashike: Missing javadocs here, `getDefaultGenerationSettings`, and for all of `Builder`.

Deamon5550:
I'll be merging this tonight if no issues.

src/main/java/org/spongepowered/api/world/biome/VirtualBiomeType.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * A virtual biome is one which exists purely for generation and therefore
+ * requires no modifications to clients in order to use it. After generation for
+ * a chunk is complete the biome is persisted as its specified persisted biome
+ * type.
+ */
+public interface VirtualBiomeType extends BiomeType {
+
+    public static Builder builder() {
```
kashike: Javadocs?

PR #1354, opened 2016-09-13
Add methods to ChannelRegistrar for getting existing channels

simon816:
You can now query for existing channels.
A convenience method `getOrCreate` and `getOrCreateRaw` have also been added

Example use for bungeecord
```java
RawDataChannel bungeeChannel = Sponge.getChannelRegistrar().getOrCreateRaw(this, "BungeeCord");
```

Side node: modifications to SpongeForge/SpongeVanilla are trivial so I'll update them if this PR is merged

### DIFF
--- a/src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
+++ b/src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
@@ -26,6 +26,7 @@ package org.spongepowered.api.network;
 
 import org.spongepowered.api.Platform;
 
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
```diff
@@ -60,6 +61,62 @@
...
+     * {@link Optional#empty()} is returned. Otherwise, a new channel is
+     * created.
+     *
+     * @param plugin The plugin to register the channel if it doesn't exist
+     * @param channel The channel name
+     * @return A new or existing indexed message channel binding
+     * @throws ChannelRegistrationException for same reasons as
+     *         {@link #createChannel}.
+     */
+    default Optional<ChannelBinding.IndexedMessageChannel> getOrCreate(Object plugin, String channel) throws ChannelRegistrationException {
```
kashike: In what case would this return `Optional.empty()`?  It's creating if not present.

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
```diff
@@ -60,6 +61,62 @@
...
+     * {@link Optional#empty()} is returned. Otherwise, a new channel is
+     * created.
+     *
+     * @param plugin The plugin to register the channel if it doesn't exist
+     * @param channel The channel name
+     * @return A new or existing raw data channel binding
+     * @throws ChannelRegistrationException for same reasons as
+     *         {@link #createRawChannel}.
+     */
+    default Optional<ChannelBinding.RawDataChannel> getOrCreateRaw(Object plugin, String channel) throws ChannelRegistrationException {
```
kashike: See above comment.

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
Mumfrey: You can see the implementation below. It returns empty if the channel exists but isn't an indexed channel.

simon816: See above javadoc
> If the channel is not an indexed message channel, then {@link Optional#empty()} is returned

It returns empty if a channel by that name exists but is not an `IndexedMessageChannel`

kashike: I missed that line somehow, apologies.
`ChannelRegistrationException` doesn't make sense since it's not being registered at that point.

Mumfrey: There's an argument to be made for throwing a different exception at that point, since if the channel is registered but of a different type, then it would indicate some kind of conflict (maybe another plugin using the same channel, maybe an older/newer version which uses raw channels instead), so maybe it's worth throwing a `UnexpectedChannelTypeException` or something which indicates the conflict state?

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
```diff
@@ -60,6 +61,62 @@
...
+     * @param channel The channel name
+     * @return A new or existing indexed message channel binding
+     * @throws ChannelRegistrationException for same reasons as
+     *         {@link #createChannel}.
+     */
+    default Optional<ChannelBinding.IndexedMessageChannel> getOrCreate(Object plugin, String channel) throws ChannelRegistrationException {
+        Optional<ChannelBinding> existing = getChannel(channel);
+        if (existing.isPresent()) {
+            if (existing.get() instanceof ChannelBinding.IndexedMessageChannel) {
+                return Optional.of((ChannelBinding.IndexedMessageChannel) existing.get());
```
Cybermaxke: You can also just cast the `Optional` ->`return (Optional) existing;`

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
simon816: I'll go ahead and throw an `IllegalStateException`

PR #1356, opened 2016-09-15
Expose RemainingJoinedStrings raw mode.

ryantheleach:
@zml2008 @mcmonkey4eva 

@mcmonkey4eva mentioned in IRC that remainingJoinedStrings still had some form of processing being applied, @zml2008 said that there was already a raw option exposed.

I looked, saw that it was limited to private visibility, presumably because it was added later and not thought about.


mcmonkey4eva:
That'd be me! Need a way to get raw command input easily. <3

PR #1360, opened 2016-09-17
Improve the block ray filters

DDoS:
In response to: https://github.com/SpongePowered/SpongeAPI/pull/1342

This adds a `skipFilter` to ignore blocks that match a predicate. The previous filter functionality gets renamed to `stopFilter`.

Also, based off of feedback from @Deamon5550, I changed the block limit into a distance limit. Since the filter is now built-in it was removed as a stand-alone.

Finally I fixed the `GenericArguments` `#target` functionality not returning the target block, but the air block just before it. 

I've fixed Cookbook/Smite and I'll merge those changes if the PR is accepted.

@m0pt0pmatt

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -1228,7 +1228,11 @@ public final class GenericArguments {
                 yStr = split[1];
                 zStr = split[2];
             } else if (xStr.equals("#target") && source instanceof Entity) {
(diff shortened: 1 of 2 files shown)


zml2008:
Looks good, this is what I intended to do -- thanks for fixing it.

ryantheleach:
As for the command part I'm not sure if it works 100% but the current #target behaviour is useful for setting blocks / teleporting entities. 

Wouldn't changing it to the new behaviour cause it to teleport entities into blocks?

e.g. /makelootcrate #target currently would place a block on top of the block being aimed at, vs the new behaviour which would "bury" the block into the ground?

PR #1362, opened 2016-09-19
Rework KickPlayerEvent

jamierocks:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/946)

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/KickPlayerEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/KickPlayerEvent.java
@@ -25,11 +25,25 @@
 package org.spongepowered.api.event.entity.living.humanoid.player;
 
 import org.spongepowered.api.entity.living.player.Player;
-import org.spongepowered.api.event.message.MessageChannelEvent;
+import org.spongepowered.api.text.Text;
 
 /**
  * Fired when a {@link Player} is kicked.
  */
-public interface KickPlayerEvent extends TargetPlayerEvent, MessageChannelEvent {
+public interface KickPlayerEvent extends TargetPlayerEvent {
+
+    /**
+     * Gets the reason given to the player.
+     *
+     * @return The reason
+     */
+    Text getReason();
+
+    /**
+     * Sets the reason given to the player.
+     *
+     * @param reason The reason
+     */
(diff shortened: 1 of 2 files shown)


windy1:
This removes the ability to set the reason, does it not?

jamierocks:
Yes. If anything MessageEvent is more suited, but even that isn't a perfect match. Well we'd have to ask the question on if a kick reason can be canceled?

src/main/java/org/spongepowered/api/event/entity/living/humanoid/player/KickPlayerEvent.java
```diff
@@ -25,11 +25,25 @@
...
 
 import org.spongepowered.api.entity.living.player.Player;
-import org.spongepowered.api.event.message.MessageChannelEvent;
+import org.spongepowered.api.text.Text;
 
 /**
  * Fired when a {@link Player} is kicked.
  */
-public interface KickPlayerEvent extends TargetPlayerEvent, MessageChannelEvent {
+public interface KickPlayerEvent extends TargetPlayerEvent {
```
windy1: Yeah like I said this can still be a `MessageChannelEvent`. For the default, you can create a new channel for the recipients of kick messages (this is console only in vanilla?). The main pros for having it extend `MessageChannelEvent` are as followed

* Always be able to retrieve the original reason
* Empty text != no message, you can send empty strings through chat (this is why we have `setMessageCancelled`)
* Modify the recipients of said message

kashike: If console is the default in vanilla: `MessageChannel.TO_CONSOLE`

ST-DDT: Ops don't recieve kick messages in vanilla?

jamierocks: Alright. I'll get that done when I get back.

kashike:
@jamierocks This pull request needs updating.

parlough:
Going to close this for now as it has not been touched since September of 2016 with conflicts and changes requested. If you'd like to update it feel free to open a new PR or ping me to re-open it.

PR #1364, opened 2016-09-23
Add SoundType builder

ZephireNZ:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/955) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/923)

Currently if a developer wants to send a custom `SoundType` (such as one from a mod or resource pack) they must implement their own `SoundType`. While small, this changes it to instead be a builder so developers can rely on common to create an instance of `SoundEvent` (the NMS impl) for them.

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/sound/SoundType.java
+++ b/src/main/java/org/spongepowered/api/effect/sound/SoundType.java
@@ -25,6 +25,9 @@
 package org.spongepowered.api.effect.sound;
 
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.GameRegistry;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/effect/sound/SoundType.java
```diff
@@ -33,4 +35,38 @@
 @CatalogedBy(SoundTypes.class)
 public interface SoundType extends CatalogType {
 
+    static Builder builder() {
```
liach: Docs?

ZephireNZ: Done :)

src/main/java/org/spongepowered/api/effect/sound/SoundType.java
```diff
@@ -33,4 +36,48 @@
 @CatalogedBy(SoundTypes.class)
 public interface SoundType extends CatalogType {
 
+    /**
+     * Creates a new {@link Builder} for building SoundTypes.
+     * @return A new builder
```
liach: Empty line before this tag. Same below.

ZephireNZ: Also done.

src/main/java/org/spongepowered/api/effect/sound/SoundType.java
```diff
@@ -33,4 +36,50 @@
 @CatalogedBy(SoundTypes.class)
 public interface SoundType extends CatalogType {
 
+    /**
+     * Creates a new {@link Builder} for building SoundTypes.
+     *
+     * @return A new builder
+     */
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(SoundType.Builder.class);
```
liach: The one below will be more brief:
```java
return Sponge.getRegistry().createBuilder(Builder.class);
```

src/main/java/org/spongepowered/api/effect/sound/SoundType.java
```diff
@@ -33,4 +36,50 @@
...
+
+        /**
+         * Resource location for the sound. If no domain (indicated by the
+         * string before ':') is present, the default "minecraft" domain will
+         * be used.
+         *
+         * @param id ID of the sound
+         * @return This builder, for chaining
+         */
+        Builder id(String id);
```
Zidane: This should become a parameter to the build method.

ST-DDT: Why? Afaict most of the other builders take it as seperate methods.

Deamon5550: @ST-DDT All builders creating cataloged types should take the id as a parameter to the build method (and looking around a bit all the ones I found either did).

src/main/java/org/spongepowered/api/effect/sound/SoundType.java
ZephireNZ: Done

src/main/java/org/spongepowered/api/effect/sound/SoundType.java
ZephireNZ: Done :)

PR #1376, opened 2016-10-01
Add BlockSoundGroup

Maxqia:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/963)
One of Many for #1365

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
+++ b/src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
@@ -39,14 +39,14 @@ public interface BlockSoundGroup {
      *
      * @return The pitch
      */
-    float getVolume();
+    double getVolume();
 
     /**
      * Gets the volume used when playing sounds.
      *
      * @return The volume
      */
-    float getPitch();
+    double getPitch();
 
     /**
      * Gets the sound played when the block is breaking.
--- a/src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
+++ b/src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
@@ -32,7 +32,7 @@ import com.google.common.base.Optional;
 /**
  * Represents sounds for a block.
  */
-public interface BlockSounds {
+public interface BlockSoundGroup {
 
     /**
      * Gets the pitch used when playing sounds.
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -107,9 +107,9 @@ public interface BlockType extends CatalogType, Translatable, PropertyHolder {
     Optional<BlockTrait<?>> getTrait(String blockTrait);
 
     /**
-     * Gets the {@link BlockSounds} for this block.
+     * Gets the {@link BlockSoundGroup} for this block.
      *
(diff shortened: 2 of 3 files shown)


Zidane:
I dislike this name. Perhaps SoundInformation? SoundContainer?

Maxqia:
Well, it's very specifically for blocks..., How about BlockSoundTypes?

Zidane:
Except it isn't a SoundType...it just happens to be a container for a few of them. Like a store? SoundStore? I want to avoid saying type of any sort (as it wouldn't be one).

Zidane:
Or perhaps put these on BlockType itself.

kashike:
@Zidane This represents the vanilla `SoundType` used by `Block`s. It represents a "sound group" of a block, which has the pitch, volume, and the many `SoundEvent`s used for that `Block`s sounds.

kashike:
`BlockSoundGroup`, perhaps?

Zidane:
BlockSoundGroup would be fine to me...but perhaps just SoundGroup.

Maxqia:
block.SoundGroup is fine with me but it might conflict with other SoundGroups in the future

Zidane:
Good point. BlockSoundGroup it is.

src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -105,4 +105,11 @@
      * @return The block trait, if available
      */
     Optional<BlockTrait<?>> getTrait(String blockTrait);
+
+    /**
+     * Gets the {@link BlockSounds} for this block.
+     *
+     * @return The block sounds
+     */
+    BlockSoundGroup getSounds();
```
Zidane: getSoundGroup please.

src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
```diff
@@ -0,0 +1,85 @@
...
+ * Represents sounds for a block.
+ */
+public interface BlockSoundGroup {
+
+    /**
+     * Gets the pitch used when playing sounds.
+     *
+     * @return The pitch
+     */
+    float getVolume();
```
Minecrell: Should use `double` because the `playSound` methods also use `double`.

src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -105,4 +105,11 @@
      * @return The block trait, if available
      */
     Optional<BlockTrait<?>> getTrait(String blockTrait);
+
+    /**
+     * Gets the {@link BlockSoundGroup} for this block.
+     *
+     * @return This block's sound group.
```
kashike: No period at end of line.

src/main/java/org/spongepowered/api/block/BlockSoundGroup.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package org.spongepowered.api.block;
+
+import org.spongepowered.api.effect.sound.SoundType;
+import org.spongepowered.api.entity.living.Living;
+
+import com.google.common.base.Optional;
+
+/**
+ * Represents sounds for a block.
```
kashike: `A group of sounds used by a {@link BlockType}.`

PR #1379, opened 2016-10-04
Add default method to see if a vector is inside a world border.

m0pt0pmatt:
This adds 2 default methods to check if a vector/location are inside a world border.

I wrote this because I would like to test when things like ender pearls, arrows, and eggs are thrown outside of the world border.

I added a test method which has the same code as the default method (The version of mockito we use doesn't allow for testing default methods).

Let me know if there's anything that needs changing or more testing.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
+++ b/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
@@ -30,6 +30,7 @@ import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.ListMultimap;
+import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.entity.Entity;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/test/java/org/spongepowered/api/world/WorldBorderTest.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package org.spongepowered.api.world;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.junit.Test;
+
+public class WorldBorderTest {
```
Deamon5550: This unit test is completely useless, sure it helps you validate that your initial logic is correct but it does nothing to test the actual functionality of the class to serve any purpose in the future. All this test will do is get out of date.

mattysweeps: I agree, but do to limitations with mockito, i can't actually test the method with mocks.

Would you prefer no unit test at all?
Or I could create a large anonymous WorldBorder class and instance?

Deamon5550: Don't have the unit test at all.

mbroomfield-r7: @Deamon5550 Unit test has been removed.

ryantheleach:
This works with nether scaling right?

m0pt0pmatt:
@ryantheleach I haven't tested that. I'll do that and get back to you

kashike:
@m0pt0pmatt Do you still wish for this to be pulled? There were no commits in this PR.

mattysweeps:
@kashike This is so old I forgot I had it open. I destroyed the commits by force updating. 

No worries though, if I still need this functionality, I'll make another PR.
I never tested nether scaling anyways.

PR #1380, opened 2016-10-05
BiomeArea is now BiomeVolume and all that this implies

DDoS:
Needs a thorough review.

See: https://github.com/SpongePowered/SpongeCommon/pull/967
See: https://github.com/SpongePowered/SpongeForge/pull/938

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/BiomeVolume.java
+++ b/src/main/java/org/spongepowered/api/world/extent/BiomeVolume.java
@@ -25,67 +25,66 @@
 package org.spongepowered.api.world.extent;
 
 import com.flowpowered.math.vector.Vector2i;
-import org.spongepowered.api.util.DiscreteTransform2;
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.util.DiscreteTransform3;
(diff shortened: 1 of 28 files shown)


src/main/java/org/spongepowered/api/world/extent/BiomeVolume.java
```diff
@@ -25,165 +25,164 @@
...
     }
 
     /**
      * Gets the {@link BiomeType} at the given location.
      *
      * @param x The X position
-     * @param z The Z position
-     * @return The biome
+     * @param y The Y position
+     * @param z The Z position  @return The biome
```
Deamon5550: need a newline

DDoS:
This can be added to the small issue list.

src/main/java/org/spongepowered/api/world/extent/BiomeVolume.java
DDoS: This can be added to the small issue list.

kashike:
https://github.com/SpongePowered/SpongeAPI/commit/9e7975e3f7173b4600bd1a10795743827e70a86d

Issue #1375, opened 2016-10-01
Get Highest Block at Specific Location


kmecpp:
In Bukkit there was a `World.getHighestBlockAt(Location)` and `World.getHighestBlockYAt(Location)` method, which returned the highest non air block at the given location and the Y position of that block respectively. 

Unless there is some other really easy way to do this, I think these methods should be added to API

Maxqia:
PS for any devs looking at this, Internally it is net.minecraft.world.getHeight(BlockPos)

Zidane:
@kmecpp 

Out of curiosity, what were these generally used for?

ewized:
I used them for optimizing things that needed to check for specific actions in blocks that may be intensive other wise.

me4502:
I used it in CraftBook for Bukkit in the Snow feature, to determine where the snow should form. It's basically a convenience method.

ryantheleach:
I would use them to check whether skylight is accessible for our old servers teleport system, it would "beam up" players as a balance check against people teleporting out of mineshafts.

It's also useful for working out where to teleport people to when given 2d co-ordinates.

me4502:
I'm not at a computer to check right now, but what does this do internally? Does it just iterate over the column?

It does make sense to mirror this if it does anything different like caching.

Deamon5550:
There is no caching it simply iterates.

ST-DDT:
> There is no caching it simply iterates.

Then there is no need to expose this.

Well there a too many too easy to implement features connected to this. 

* Highest block below height X,
* Also ignore leaves or grass
* ignore floating blocks
* ...

So it should be the plugins job to do this. If the plugin needs help it can use the block ray for help. Maybe we can add a helper method for column iterating there if necessary.

Minecrell:
`World.getHeight(BlockPos)` does not iterate, it returns the current value from an internal height map of the chunk.

me4502:
In that case, it makes sense to expose this.

Deamon5550:
@Minecrell that value is useless almost by the time that the populators run

ryantheleach:
Honestly most of the time getting the highest skylight block is what people want, they are just not thinking about it.

If the two methods were in and documented, it would make people think twice about choosing which one.

Additionally if you make people iterate themselves, you have the problem of cubic chunks, whilst Minecraft might not make any optimizations for highest block, cubic chunks would be able to skip any chunks not generated.

theboomer:
[...]

PR #1381, opened 2016-10-05
Added World#getHighestLocationAt(Vector2i). Fixes SpongeAPI #1375

me4502:
As per https://github.com/SpongePowered/SpongeAPI/issues/1375, this PR adds a method to retrieve the highest block in a column, using the internal chunk heightmaps. This retrieves the highest opaque block, which is also the lowest block sunlight can reach.

The method name is up for debate.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/World.java
+++ b/src/main/java/org/spongepowered/api/world/World.java
@@ -84,14 +84,14 @@ public interface World extends Extent, WeatherUniverse, Viewer, ContextSource, M
     }
 
     /**
-     * Get the {@link Location} of the lowest block that sunlight can reach in the given column.
+     * Get the {@link Location} where sunlight reaches in the given column.
      *
      * <p>This method ignores all transparent blocks, providing the highest opaque block.</p>
      *
      * @param position The column position
      * @return The highest opaque block
      */
-    Location<World> getHighestLocationAt(Vector2i position);
+    Location<World> getHighestBlockAt(Vector2i position);
 
     /**
      * Get the loaded chunk at the given block coordinate position.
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -83,6 +84,16 @@
     }
 
     /**
+     * Get the {@link Location} of the lowest block that sunlight can reach in the given column.
```
kashike: Weird javadocs.

me4502: How would you suggest improving them?

Maxqia:
Vector2i -> 3i?

windy1:
Err the highest location at any x/y will always be at the map height :P

This should probably be called get highest block at or something like that.

me4502:
I chose 2i, because adding a third dimension is kind of redundant. It'd be ignored anyway

src/main/java/org/spongepowered/api/world/World.java
Maxqia: lowest -> highest

Maxqia:
Well it makes it easier to do something like this world.getHighestLocationAt(location.getBlockPosition()). Also can you change the name to getHighestBlockAt?

me4502:
What do people think of the new javadocs? I felt 'highest' was wrong, as the highest block would always be 255 as that's the highest block sunlight reaches.

Maxqia:
Maybe just the highest Opaque block for the java docs?

me4502:
Is that accurate?

Maxqia:
I think so, well also it wouldn't be 255 because it ignores transparent blocks :P

Deamon5550:
Add a version that takes two ints and default implement the vector one to call that one.

PR #1385, opened 2016-10-08
Provide implementations for mutable and immutable CatalogRegistryModules

ST-DDT:
Description is WIP. The implementation is ready for review.

# The issue
Currently there are multiple issues with more than a few existing implementations of `CatalogRegistryModule`s.

Currently there are 100 implemented `CatalogRegistryModule`s. 
From those 33 are not visible in `CatalogTypes` fields.

Issue | Affected
------------ | -------------
Exception instead of `Optional#empty()` in `#getById(String)` | 2
`#getById(String)` returns `Optional#empty()` for registered ids | 31
`#getById(String)` does not handle case properly | 2
Inconsistent handling of "sponge:" and "minecraft:" prefixes in `#getById()` | 30
Expensive deep copy in `getAll()` | 95
Duplicates in `getAll()` | 6
Incomplete/lacking `#registerAdditionalCatalog(CatalogType)` implemenations | 24 / 28 (86%)

Some of these issues are also mentioned in https://github.com/SpongePowered/SpongeCommon/issues/739.

# TestPlugin

* [Plugin](https://dl.dropboxusercontent.com/u/16999313/Sponge/CatalogRegistryModuleTest.jar)
* [Source](https://dl.dropboxusercontent.com/u/16999313/Sponge/CatalogRegistryModuleTest.zip)
* [Result](http://pastebin.com/ASr7D8LD)

# The solution
Provide simple to use implementations that reduce the implementation work to just the default registration and thus reduce the error potential.

(This implementation alone cannot fix the inconsistent prefix handling though)

# Usage

````java
@RegisterCatalog(GoalTypes.class)
public final class GoalTypeModule extends MutableCatalogRegistryModule<GoalType> {

    @Override
    protected Collection<String> aliasIdsForType(GoalType type) {
        // Also register without prefix
        if (type.getId().startsWith("minecraft:")) {
            return Collections.singletonList(type.getId().substring(10));
        }
        return Collections.emptyList();
    }

    @Override
    public void registerDefaults() {
        createGoalType(SpongeImpl.getMinecraftPlugin(), "normal", "Normal");
        createGoalType(SpongeImpl.getMinecraftPlugin(), "target", "Target");
    }

    public GoalType createGoalType(PluginContainer pluginContainer , String id, String name) {
        String combinedId = pluginContainer.getId().toLowerCase(Locale.ENGLISH) + ":" + id;
        GoalType newType = new SpongeGoalType(combinedId, name, EntityAITasks.class);
        registerAdditionalCatalog(newType);
        return newType;
    }

}
````

------------------------

I also offer my help to rewrite those modules to use these implemenations.

### DIFF
--- a/src/main/java/org/spongepowered/api/registry/AdditionalCatalogRegistryModule.java
+++ b/src/main/java/org/spongepowered/api/registry/AdditionalCatalogRegistryModule.java
@@ -26,8 +26,23 @@ package org.spongepowered.api.registry;
 
 import org.spongepowered.api.CatalogType;
 
+/**
+ * Registry module that is used to maintain a list of catalog types of a certain
+ * type. This module also accepts catalog type instances that are created by
+ * plugins.
+ *
+ * @param <T> The type of catalog type that is maintained by this module
+ */
 public interface AdditionalCatalogRegistryModule<T extends CatalogType> extends CatalogRegistryModule<T> {
 
-    void registerAdditionalCatalog(T extraCatalog);
+    /**
+     * Registers the given catalog type instance in this registry module. The
+     * ids that can be used to access this instance are chosen automatically.
+     *
+     * @param extraCatalog The catalog type to register
+     * @throws IllegalArgumentException If there is an id conflict with the
+     *         given type and an existing type
+     */
+    void registerAdditionalCatalog(T extraCatalog) throws IllegalArgumentException;
 
 }
--- a/src/main/java/org/spongepowered/api/registry/AlternateCatalogRegistryModule.java
+++ b/src/main/java/org/spongepowered/api/registry/AlternateCatalogRegistryModule.java
@@ -25,10 +25,28 @@
 package org.spongepowered.api.registry;
 
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.registry.util.RegisterCatalog;
 
 import java.util.Map;
 
+/**
+ * Registry module that is used to maintain a list of catalog types of a certain
+ * type. Implementations must also annotate the class or a field in the class
+ * with the {@link RegisterCatalog} annotation, to indicate which catalog type
+ * pseudo enum class should be populated with the given values. The
+ * {@link #provideCatalogMap()} method takes precedence before any fields
+ * annotated with the {@link RegisterCatalog} annotation.
+ *
+ * @param <T> The type of catalog type that is maintained by this module
+ */
 public interface AlternateCatalogRegistryModule<T extends CatalogType> extends CatalogRegistryModule<T> {
 
+    /**
+     * Returns a view of the internal id to instance mapping that can be used to
+     * populate the catalog type pseudo enums.
+     *
+     * @return The map that should be used to populate the pseudo enums.
+     */
     Map<String, T> provideCatalogMap();
+
 }
--- a/src/main/java/org/spongepowered/api/registry/CatalogRegistryModule.java
+++ b/src/main/java/org/spongepowered/api/registry/CatalogRegistryModule.java
@@ -29,10 +29,29 @@ import org.spongepowered.api.CatalogType;
 import java.util.Collection;
 import java.util.Optional;
 
+/**
+ * Registry module that is used to maintain a list of catalog types of a certain
+ * type.
+ *
+ * @param <T> The type of catalog type that is maintained by this module
+ */
 public interface CatalogRegistryModule<T extends CatalogType> extends RegistryModule {
 
+    /**
+     * Gets the registered catalog type with the given id or an equivalent.
+     *
+     * @param id The id of the catalog type to search
+     * @return The catalog type if a matching one is found
+     */
     Optional<T> getById(String id);
 
+    /**
+     * Gets a duplicate free collection of all registered catalog type instances
+     * of this type. The order of the contained elements should be considered
+     * random.
+     *
+     * @return A collection containing all registered catalog types
+     */
     Collection<T> getAll();
 
 }
--- a/src/main/java/org/spongepowered/api/registry/ImmutableCatalogRegistryModule.java
+++ b/src/main/java/org/spongepowered/api/registry/ImmutableCatalogRegistryModule.java
@@ -1,0 +1,97 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+        return this.types;
+    }
+
+    @Override
+    public Map<String, T> provideCatalogMap() {
+        return this.typesById;
+    }
+
+    @Override
+    // Registration needs to take place in the constructor
```
kashike: Move this comment above the annotation.

src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Creates a new instance with the given id to type mappings. All keys will
+     * be stored in lower case.
+     *
+     * @param typesById The map containing all registered ids for all catalog
+     *        types
+     */
+    protected ImmutableRegistryModule(Map<String, T> typesById) {
+        checkNotNull(typesById, "typeByKey");
```
kashike: Wrong second parameter.

src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Checks and converts the given key to its lower case equivalent, that can
+     * be used to get a type from the key map.
+     *
+     * @param id The key to check and convert
+     * @return The checked lower case key
+     * @see #getKeyMap()
+     */
+    protected final String toLowerCase(String id) {
```
kashike: Can be `static`.

src/main/java/org/spongepowered/api/registry/MutableRegistryModule.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Simple implementation of the {@link CatalogRegistryModule} that allows plugin
+ * created types to be registered.
+ *
+ * @param <T> The type of catalog type that is maintained by this module
+ */
+public abstract class MutableRegistryModule<T extends CatalogType> implements AdditionalCatalogRegistryModule<T>, AlternateCatalogRegistryModule<T> {
+
+    private final Map<String, T> typesById = new LinkedHashMap<>();
+    private Set<T> types;
```
kashike: `@Nullable`

src/main/java/org/spongepowered/api/registry/MutableRegistryModule.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Checks and converts the given id to its lower case equivalent, that can
+     * be used to get a type from the catalog map.
+     *
+     * @param id The id to check and convert
+     * @return The checked lower case id
+     * @see #provideCatalogMap()
+     */
+    protected final String toLowerCase(String id) {
```
kashike: Can be `static`.

src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java
ST-DDT: Fixed

src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java
ST-DDT: Fixed

src/main/java/org/spongepowered/api/registry/ImmutableRegistryModule.java
ST-DDT: Fixed

src/main/java/org/spongepowered/api/registry/MutableRegistryModule.java
ST-DDT: Fixed

src/main/java/org/spongepowered/api/registry/MutableRegistryModule.java
ST-DDT: Fixed

src/main/java/org/spongepowered/api/registry/MutableRegistryModule.java
```diff
@@ -23,7 +49,7 @@
 public abstract class MutableRegistryModule<T extends CatalogType> implements AdditionalCatalogRegistryModule<T>, AlternateCatalogRegistryModule<T> {
 
     private final Map<String, T> typesById = new LinkedHashMap<>();
-    private Set<T> types;
+    private @Nullable Set<T> types;
```
kashike: `@Nullable private Set<T> types;`

ST-DDT: Fixed

ST-DDT:
Updated PR description with actual numbers. Based on those numbers 30% of the existing `CatalogTypeRegistry`s would benefit from this.

Cybermaxke:
I think that it would be better that you can assign a `RegisterCatalog` to fields and methods that return a `Map`. Then you could also remove the `AlternateRegistryModule` and this could also allow multiple catalog mappings (and private provide methods?).

```java
@RegisterCatalog(ItemTypes.class)
private Map<String, ItemType> getItemTypeMappings() {
    final Map<String, ItemType> map = new HashMap<>();
    // ...
    return map;
}
```

ST-DDT:
Currently `@RegisterCatalog` is only supported once per Module. If that gets ever changed than it this variant would be just one more source.

The provideCatalog method is only for the use in the impl so there is no harm exposing it. Afaict its not possible to get the module instance directly from the API. And even if it is it wont be of any harm because its an unmodifyable view. Its keys might be useful for tab autocompletion though. Anything that cannot be found in the keys is unsuitable to get a catalog instance. I even considered creating a different PR that exposes all keys via the GameRegistry for the autocompletion to use.

gabizou:
So this doesn't do squat for the API aside from adding suggestive implementations that can't be used with the `GameRegistry` unless changes are made to it's registry management. Where does this actually fix the problems identified in the description, because as far as I can tell, those are specific to the *implementation*, not the API.

gabizou:
Closing as the rebuttal for a reason to add this to the API hasn't been provided.

PR #1387, opened 2016-10-09
Clarify some parts of the PermissionService documentation

lucko:
This PR attempts to clarify some of the ambiguities in PermissionService, as raised in #1383.

As far as I know, the added javadocs do not conflict with existing intended behaviours, rather just clarifies them. 

Possibly relevant: some of the issues were also briefly discussed here: https://forums.spongepowered.org/t/luckperms-an-advanced-permissions-system/14274/118

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/PermissionService.java
+++ b/src/main/java/org/spongepowered/api/service/permission/PermissionService.java
@@ -64,6 +64,11 @@ public interface PermissionService extends ContextualService<Subject> {
      * subject is at the root of all inheritance trees, above even subject type-specific
      * defaults, meaning it has the lowest priority when all other weighting is equal.
      *
+     * <p>Note: This data should be persisted, so plugins that add permissions to this subject
+     * must take care to not override permissions already set or modified. It is also
+     * recommended to use {@link Subject#getTransientSubjectData()} where possible to
+     * avoid persisting unnecessary data.
+     *
      * @return The default subject data
      */
     Subject getDefaults();
--- a/src/main/java/org/spongepowered/api/service/permission/Subject.java
+++ b/src/main/java/org/spongepowered/api/service/permission/Subject.java
@@ -48,6 +48,11 @@ import java.util.Set;
  * child levels would granted but {@code example.access} would denied (for
  * example).
  *
+ * <p>It is the responsibility of the {@link PermissionService} implementation
+ * to provide this behavior, and resolve the implicit permission node inheritance
+ * explained above when a Subject is queried for permissions. Use of a
+ * {@link NodeTree} is recommended.
+ *
  * <p>Plugins may opt to implement "dynamic" permissions such as {@code
  * example.region.define.&lt;region&gt;} where {@code region} would be a
  * dynamically added level based on the context, though some attention should be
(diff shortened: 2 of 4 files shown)


bloodmc:
@zml2008 Please review

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
```diff
@@ -64,6 +64,11 @@
      * subject is at the root of all inheritance trees, above even subject type-specific
      * defaults, meaning it has the lowest priority when all other weighting is equal.
      *
+     * <p>Note: This data should be persisted, so plugins that add permissions to this subject
+     * must take care to not override permissions already set or modified. It is also
+     * recommended to use {@link Subject#getTransientSubjectData()} where possible to
+     * avoid persisting unnecessary data.
```
ST-DDT: I'm not so sure about the persisting part.
If you modify defaults then its customized and no longer a default. IMO its better to specify a `user` group that inherits from the `default` group and overwrite things there.
In addition to that on one hand you force the permission plugin to persist the data, but on the other hand you ask the plugin owner to set transient data only.

Missing `</p>`tag.

src/main/java/org/spongepowered/api/service/permission/Subject.java
```diff
@@ -101,14 +114,17 @@
...
+     * Test whether the subject is permitted to perform an action corresponding to the
+     * given permission string. This must return the same boolean equivalent as
+     * {@link #getPermissionValue(Set, String)}.
      *
      * @param contexts The set of contexts that represents the subject's current environment
      * @param permission The permission string
      * @return True if permission is granted
      */
-    boolean hasPermission(Set<Context> contexts, String permission);
+    default boolean hasPermission(Set<Context> contexts, String permission) {
```
ST-DDT: This should also mention the "format" permission strings are supposed to follow. 
(See `PermissionDescription` for more details)
And it should mention that there are no wildcard.* permissions.

The same applies for `getPermissionValue`

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
lucko: I agree with you - however, the current intended behaviour is that defaults are persisted. 

(at least, thats what I'm assuming, judging from the responses I've had from blood, and from looking at the PEX implementation. I'm all for saying "nope, defaults should be transient", however this PR was about improving documentation, not changing intended behaviours.) 

There's missing `</p>` tags throughout PermissionService. You'll notice they don't appear in the existing JavaDocs either, so I just followed the same convention.

src/main/java/org/spongepowered/api/service/permission/Subject.java
lucko: Plugins wouldn't normally check if users had wildcard permissions, so I don't think those methods are the correct place to document that.

I added a reminder that implementations should consider "wildcards" in those checks. For the actual explanation, they can read the description in the main Subject class documentation.

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
zml2008: Yes, the intention is that persistent defaults and transient defaults are exposed.

Generally, the persistent defaults are primarily set by the user and transient defaults are primarily set by plugins, but the permissions API exposes both.

lucko: Yep - that's fine, it just wasn't documented.

zml2008: mk good, it would probably be good to say more about the purposes of persistent vs transient (and make sure to update the javadocs in SubjectCollection as well)

lucko:
Any update on this @zml2008?

PR #1391, opened 2016-10-18
Add method to determine if the current thread is the main thread

### LINKED ISSUE #897: Method to determine if calling from "main" thread
A method added to `Game` or `Server` to determine if the current thread is the main thread will be useful, as well as not requiring the plugin dev to figure it out by thread name (thread names are an implementation detail).
similar: https://github.com/Bukkit/Bukkit/blob/master/src/main/java/org/bukkit/Server.java#L805-L817

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1391) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/983)
Fixes #897

### DIFF
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -31,6 +31,7 @@ import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.scoreboard.Scoreboard;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.util.Threaded;
 import org.spongepowered.api.world.ChunkTicketManager;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.WorldArchetype;
@@ -47,7 +48,7 @@ import java.util.concurrent.CompletableFuture;
 /**
  * Represents a typical Minecraft Server.
  */
-public interface Server {
+public interface Server extends Threaded {
 
     /**
      * Gets the {@link Player}s currently online.
@@ -428,5 +429,4 @@ public interface Server {
      * @param timeout The player idle timeout
      */
     void setPlayerIdleTimeout(int timeout);
-
 }
--- a/src/main/java/org/spongepowered/api/util/Threaded.java
+++ b/src/main/java/org/spongepowered/api/util/Threaded.java
@@ -1,0 +1,41 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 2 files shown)


liach:
This would be a bad idea for per-world threads, I guess.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -429,4 +429,13 @@
...
     void setPlayerIdleTimeout(int timeout);
 
+    /**
+     * Tests if the {@link Thread#currentThread current thread} is the
+     * main thread of the server.
+     *
+     * @return {@code true} if current thread is the main
+     *     thread, {@code false} if not
+     */
+    boolean isMainThread();
```
simon816: I think `isOnServerThread` would be a better name

Zidane: @kashike 

Something like what @simon816 is proposing. When someone reads "main thread", that inherently implies only one and should client API ever to be added, the Client interface would have a method just like this one.

Deamon5550:
@liach The amount of things plugins have to code differently, not to mention a major mindset shift, for per-world threads kind of precludes doing it without re-writing every plugin you run anyway. Its not worth bringing up here.

aikar:
@liach also can add a isWorldThread in the future too.
The goal here is to give plugins the ability to know "Where am I" at any point.

This is really needed for me to add Sponge support to https://github.com/aikar/TaskChain - an extremely useful framework for plugin developers.
I took the time to refactor it to be platform agnostic, but Sponge is missing the critical piece of "am I on a thread I don't want to block"

src/main/java/org/spongepowered/api/Server.java
aikar: then really need to add client one at same time to support for both.

liach:
@aikar I would suggest to add a `NonThreadSafe` interface to represent objects like per thread world or server thread.

aikar:
@liach where would that even be used? The whole system is built on "run this code on the games main thread" or else "in a thread pool"

The goal being you post to main for any API's you need to call that need to be on main.  It doesn't care what your code in the task is.

src/main/java/org/spongepowered/api/util/Threaded.java
```diff
@@ -0,0 +1,41 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+/**
+ * Represents something which has a {@link Thread}.
+ */
+public interface Threaded {
```
liach, line 30: I would suggest add method like scheduleOn for task

Deamon5550: Putting the cart before the horse there. IF there were any api exposed alternate threads then first the scheduler would need a drastic overhaul, and only after that could you argue for a method doing this.

liach: This pull targets bleeding and thread check's introduction is already an overhaul.

aikar: @liach the main "Driver" for this API is my TaskChain system, which works completely fine as an external library, but we are in discussion on providing the TaskChain API as part of the Sponge API.

It honestly wouldn't even need to refactor the current scheduler, but simply add an access point to the scheduler like 

```
TaskChainFactory factory = Sponge.getScheduler().createTaskChainFactory(plugin);
factory.newChain()./* foo */.execute();
```

aikar:
I'm honestly not a fan of the "Threaded" Interface. Everything is bound to some form of thread... And the server does have multiple threads, just one is considered "Main"

I still am a fan of something like Sponge.isCurrentlyOnMainThread() that will identify client or server main thread.

kashike:
`Server extends Threaded`
-> `server.isCurrentThread()`
`Client extends Threaded`
-> `client.isCurrentThread()`

If you have a better name for the method, I'm all for it.

kashike:
https://github.com/SpongePowered/SpongeAPI/commit/f247c803524f44f2dc1431c976dca87c5789d2e9

PR #1393, opened 2016-10-20
Add "simple" custom data API

ZephireNZ:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/988)

I'll put it bluntly - implementing the current data API system is tedious. Say you want to save a string to an entity. You need to implement two `DataManipulator`s with keys, values, register everything, and a `DataManipulatorBuilder` too if you want it persisted.

That's a tonne of work for just a string, especially for someone new to plugin development or to sponge.

At the heart of this PR, there is still the bells and whistles of the data API. You use keys to access values, but the rest of the implementation work is done behind the scenes - all you need to do is register and you're good to go.

It still supports the forwards compatibility with `DataContentUpdater`s too, but that's an optional feature that can be left out if desired - the version will just default to `1`.

We'll still keep around the current system for those that want more control over their data, or providing a better interface to other plugins. But this will really ease the learning curve for those that don't care for the details (yet).

So instead of implementing and registering multiple classes to access your string data, you just do this:

``` java
SimpleCustomData<String> custom = SimpleCustomData.builder()
            .id("my-plugin:my_string")
            .name("My String")
            .build();

myEntity.offer(custom.getKey(), "Hello World!");
```

**Super simple to register, super simple to use.**

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/SimpleCustomData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/SimpleCustomData.java
@@ -1,0 +1,43 @@
+package org.spongepowered.api.data.manipulator;
+
+import com.google.common.reflect.TypeToken;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.persistence.DataContentUpdater;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.util.ResettableBuilder;
+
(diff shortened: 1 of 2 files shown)


Deamon5550:
Duplicate #1198 ???

We've been planning this for ages its just held up because I need to write some code for generating the manipulator classes with ASM similar to the functioning of our event system.

gabizou:
This is a duplicate of #1198. No amount of builder classes will simplify the creation of custom data except if we resort to class generation as #1198 is already proposing. There's quite a few changes that I still need to make to Data API in general to fully support #1198 to ensure a smooth transition to class generation and reliability of the keys generated.

Issue #1392, opened 2016-10-18
ItemStackBuilder does not accept negative quantities


ZephireNZ:
`ItemStack`s are allowed to have negative quantities, and `Inventory#query(ItemStack)` relies on quantity being -1 to ignore stack size differences.

`ItemStack` itself does not set a lower limit, however its builder does. This should be change to make like a bit easier when querying inventories.

Mumfrey:
Actually, given that "item stack with negative size" is actually a bit of a weird concept, and - as you identified - only certain negative values have meaning, effectively making those values semaphores. It would seem to make more sense for the builder to support setting the semaphore via an `ItemStackBuilder::anySize()` or similar setter which _internally_ sets the quantity to -1, but decouples this detail from the consumer code.

simon816:
An issue was made on spongecommon regarding negative stack sizes
SpongePowered/SpongeCommon#150

PR #1397, opened 2016-10-23
Negative one item stack quantities

hfreeb:
API | [Common](https://github.com/SpongePowered/SpongeCommon/pull/996)

As discussed at https://github.com/SpongePowered/SpongeAPI/issues/1392, a -1 quantity would be useful for ignoring stack size differences with `Inventory#query(ItemStack)`.

In my implementation I have added `ItemStack.Builder#anySize` which can be used for querying inventories.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
@@ -163,14 +163,12 @@ public interface ItemStack extends DataHolder, Translatable {
         Builder quantity(int quantity) throws IllegalArgumentException;
 
         /**
-         * Sets the quantity of the item stack without doing validation on the
-         * minimum stack size. Allows a quantity of zero and negative one but further
-         * negative numbers are not allowed.
+         * Sets the ItemStack's quantity to represent any size to be used for
+         * querying inventories.
          *
-         * @param quantity The quantity of the item stack
          * @return This builder, for chaining
          */
-        Builder unsafeQuantity(int quantity);
+        Builder anySize();
 
         /**
          * Adds a {@link Key} and related {@link Object} value to apply to the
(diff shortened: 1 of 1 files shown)


Deamon5550:
And as mentioned [here](https://github.com/SpongePowered/SpongeAPI/issues/1392#issuecomment-254449839) it would be far better to decouple these features from the quantity into a different solution.

hfreeb:
> And as mentioned here it would be far better to decouple these features from the quantity into a different solution.

What do you think about having a zero quantity?

Deamon5550:
Api features relying on glitches in the base game should generally be avoided because they because very subject to breakages in a future game update.

kashike:
Zero-quantity is helpful with inventories, as mentioned [here](https://github.com/SpongePowered/SpongeCommon/issues/150#issuecomment-136318289).

Faithcaio:
> What do you think about having a zero quantity?

I managed to crash the client when trying to collect a zero quantity stack onto cursor (double click). Not great.

hfreeb:
> And as mentioned here it would be far better to decouple these features from the quantity into a different solution.

I have changed it to the suggested solution.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -164,6 +163,14 @@ static ItemStack of(ItemType itemType, int quantity) {
         Builder quantity(int quantity) throws IllegalArgumentException;
 
         /**
+         * Sets the ItemStack's quantity to represent any size to be used for
+         * querying inventories.
+         *
+         * @return This builder, for chaining
+         */
+        Builder anySize();
```
liach, line 171: Should there be a method called `empty` or `invalid` since Minecraft introduced invalid item stack concepts in 1.11?

gabizou:
I'm pretty sure that negative item stack quantities is no longer valid in 1.11.

Lunaphied:
@gabizou I poked into the chain of issues connected to the -1/generally negative item quantities and found that people have may have been merging two separate issues. One issue is about the need for a generic ItemStack that matches any size for a generic ```Inventory#query(ItemStack)```, which would use a -1 quantity except it can't be built that way. The other is about 0 sized stacks and -1 sized stacks for shops. 

The first should be able to be handled with an internal attribute entirely disconnected from quantity right? (That seems the proper solution regardless, unless an abstraction of a fully spec'd item would replace this use of `ItemStack` That appears to be what this issue is about and the title (and implementation) should likely be updated accordingly.

The other issue may be invalidated by the changes you believe 1.11 has. Are zero quantity stacks still valid?

kashike:
Zero quantities are no longer valid.

PR #1411, opened 2016-11-14
Updates to the permissions API to better match with what permissions plugins are doing

zml2008:
Based on discussions with some of the Sponge permissions plugin developers, we found that the current permissions API is inadequate for some of the more complicated features of existing permissions plugins.

To fill in some of the gaps, I've put together these changes which:

- Add a distinction between cached subjects, which are immediately available, and uncached subjects, which are loaded asynchronously.
- Add events for permissions changes
- Add the concept of a subject type definition which allows specifying metadata about a certain subject type.


src/main/java/org/spongepowered/api/service/permission/SubjectTypeDefinition.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.function.Predicate;
+
+/**
+ * A class defining attributes about how a permissions plugin should interpret a certain subject type.
+ */
+public final class SubjectTypeDefinition {
+    private static final SubjectTypeDefinition DEFAULT = builder().build();
+
+    private final boolean transientHasPriority;
+    public final Predicate<String> identifierValidityPredicate;
```
kashike: Why `public`? It has a getter.

src/main/java/org/spongepowered/api/service/permission/SubjectTypeDefinition.java
```diff
@@ -0,0 +1,132 @@
...
+        this.nameIdentifierMapping = builder.getNameIdentifierMapping();
+    }
+
+    /**
+     * Return whether transient data has priority over persistent data for this subject type for lookups
+     *
+     * @return The priority of transient data
+     */
+    public boolean transientHasPriority() {
+        return transientHasPriority;
```
kashike: `this.` (applies to all missing `this.`)

src/main/java/org/spongepowered/api/service/permission/SubjectTypeDefinition.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * A class defining attributes about how a permissions plugin should interpret a certain subject type.
+ */
+public final class SubjectTypeDefinition {
+    private static final SubjectTypeDefinition DEFAULT = builder().build();
```
kashike, line 38: New line after class declaration.

kashike (changes requested):

src/main/java/org/spongepowered/api/service/permission/change/OptionChange.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Represents a change in option values
+ */
+public class OptionChange {
+    private final String option;
+    private final String oldValue;
+    private final String newValue;
+
+    public OptionChange(String option, String oldValue, String newValue) {
```
liach: Mark `@Nullable` in params that can be null.

src/main/java/org/spongepowered/api/event/service/permission/TargetSubjectEvent.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface TargetSubjectEvent extends Event {
+
+    /**
+     * The targeted subject
+     *
+     * @return The subject affected by this event
+     */
+    Subject getTargetSubject();
+}
```
liach, line 41: Newline before this line

src/main/java/org/spongepowered/api/event/service/permission/TargetSubjectEvent.java
```diff
@@ -0,0 +1,41 @@
...
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.service.permission.Subject;
+
+/**
+ * An event that targets a {@link Subject}
+ */
+public interface TargetSubjectEvent extends Event {
+
+    /**
+     * The targeted subject
```
liach: Docs should be full sentence. Applies to all others.

src/main/java/org/spongepowered/api/event/service/permission/ChangeSubjectEvent.java
```diff
@@ -0,0 +1,92 @@
...
+         */
+        List<PermissionChange> getChangedPermissions();
+
+        default void forEach(Consumer<PermissionChange> func) {
+            getChangedPermissions().forEach(func);
+        }
+    }
+
+    interface Option extends ChangeSubjectEvent {
+        /**
```
liach, line 69: Newline in the beginning

src/main/java/org/spongepowered/api/event/service/permission/ChangeSubjectEvent.java
```diff
@@ -0,0 +1,92 @@
...
+         * Get the list of changed options. This list is immutable.
+         *
+         * @return The changed options
+         */
+        List<OptionChange> getChangedOptions();
+
+        default void forEach(Consumer<OptionChange> func) {
```
liach, line 76: Do we need docs for this one?

src/main/java/org/spongepowered/api/event/service/permission/ChangeSubjectEvent.java
```diff
@@ -0,0 +1,92 @@
...
+public interface ChangeSubjectEvent extends TargetSubjectEvent {
+
+    /**
+     * The set of {@link Context}s that this value was changed in
+     *
+     * @return The contexts changed
+     */
+    Set<Context> getContexts();
+
+    interface Permission extends ChangeSubjectEvent {
```
liach, line 51: Docs for sub interfaces?

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -29,9 +29,15 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
 
+import com.google.common.collect.Maps;
```
liach, line 33: ??

liach (changes requested):

src/main/java/org/spongepowered/api/event/service/permission/ChangeSubjectEvent.java
```diff
@@ -0,0 +1,101 @@
...
+
+        /**
+         * Get the list of changed permissions. This list is immutable.
+         *
+         * @return The changed permissions
+         */
+        List<PermissionChange> getChangedPermissions();
+
+        default void forEach(Consumer<PermissionChange> func) {
```
ST-DDT, line 60: If you name the method like this you cannot implement `ChangeSubjectEvent.Option` and `ChangeSubjectEvent.Permission` at the same time.
I'm not sure whether this is necessary, but it might be useful for bulk/transactional updates.

liach: They are implemented by SpongeEventFactory generated classes. They are never supposed to be implemented by the same class.

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
liach: I mean this import should not be grouped like this.

src/main/java/org/spongepowered/api/service/permission/NodeTree.java
```diff
@@ -107,6 +107,26 @@ public Tristate get(String node) {
             }
         }
         return lastUndefinedVal;
+    }
+
+    /**
+     * Returns the value assigned to a specific node, <em>without</em> returning parents
```
liach, line 113: Please use complete sentence.

src/main/java/org/spongepowered/api/service/permission/NodeTree.java
```diff
@@ -107,6 +107,26 @@ public Tristate get(String node) {
...
+        Node currentNode = this.rootNode;
+        Tristate nearestValue = Tristate.UNDEFINED;
+        for (String str : parts) {
+            if (!currentNode.children.containsKey(str)) {
+                return Tristate.UNDEFINED;
+            }
+            currentNode = currentNode.children.get(str);
+            nearestValue = currentNode.value;
+        }
+        return nearestValue;
```
liach, line 130: Why to leave a blank line here?

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -26,11 +26,19 @@
...
+import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Object that manages subjects of a certain type (user, group, etc).
+ *
+ * A distinction is made between subjects that are stored and subjects that are currently cached. The cached methods must return quickly, but the
```
liach, line 38: `<p></p>`?

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -42,14 +50,39 @@
     String getIdentifier();
 
     /**
+     * Get the type definition controlling how this subject functions.
```
liach, line 53: `Gets`. And use `Sets` for below.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -42,14 +50,39 @@
...
+    /**
+     * Returns the subject specified. Will only return an empty optional if the subject has not been cached.
+     *
+     * @param identifier The identifier to look up a subject by.
+     *                   Case-insensitive
+     * @return A stored subject if present, otherwise a subject that may be
+     *         stored if data is changed from defaults
+     */
+    Optional<Subject> getIfCached(String identifier);
+
```
liach, line 85: Extra blank line.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -59,7 +92,16 @@
...
      */
-    boolean hasRegistered(String identifier);
+    CompletableFuture<Boolean> hasRegistered(String identifier);
+
+    /**
+     * Returns whether or not the permissions provider has any data for the
+     * given subject currently in its cache.
+     *
+     * @param identifier The identifier of the given subject
+     * @return whether any data is stored
```
liach, line 102: `Whether`

src/main/java/org/spongepowered/api/service/permission/SubjectTypeDefinition.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private final boolean transientHasPriority;
+    private final Predicate<String> identifierValidityPredicate;
+    private final Function<String, Optional<String>> nameIdentifierMapping;
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static SubjectTypeDefinition defaults() {
```
liach, line 48: Static utility methods need javadocs.

liach (changes requested):

Deamon5550:
On a related note I've just been told by @lucko that [SubjectData#setPermission](https://github.com/SpongePowered/SpongeAPI/blob/feature/perms-update/src/main/java/org/spongepowered/api/service/permission/SubjectData.java#L67-L78) is allowed to perform the change asynchronously such that a call immediately after to `hasPermission` might not reflect the change.

I disagree with this behaviour and think that it is not unreasonable to have the requirement to update any cached storage of the set permission to reflect the change immediately as that is what people will expect. Otherwise the javadocs of that method will have to be updated that you cannot rely on the change being visible immediately.

Letting the change not be immediately visible means that there is **no way** to know when a change will go through and if you have logic that requires knowing this you are shit out of luck (for example, granting someone a permission then performing an action on their behalf before removing the permission from them again).

gabizou:
I'd like to get some of the issues noted above resolved. @zml2008 Let me know whether you need someone to take over this PR to refine it.

Zidane:
@lucko 

With https://github.com/SpongePowered/SpongeAPI/pull/1524 being merged very soon, do you want to salvage some aspects of this one so it can be merged as well? I know events would be nice here.

lucko:
I'm happy with the content of the other PR. 

Quoted from #1524 

> This PR addresses no. 2 above, but depending on the success of this, I may also attempt to address the others over time.

I think events need a bit more thought before they're added. Same goes for the other aspects imho.

Zidane:
@lucko Can you take this over and salvage what you can + other additions you want?

lucko:
Uhh, it's really just events missing now. I personally would opt for something very simple.

zml was trying to add events for every time a single permission/option/parent was added/removed. 

This becomes a very clunky event API, and also in my experience, very difficult to implement for the service provider.

In my opinion, a single `SubjectDataUpdateEvent` would suffice.

Effectively, "hey plugin, fyi, something about this Subjects data has changed. you might wanna update any internal caches you have".

(the only really discussion would be whether that should be called async or not. it would obviously need to be consistent across providers, and also probably need to be optionally called)

This covers the most common usage, which is chat formatting plugins, and plugins which sync "group data" with external applications. (specifically things like Enjin)

I'd be happy to PR this, but I don't think I'd do a good job of anything more complex, if it was decided that Sponge wanted something more than just a simple update event.

Zidane:
@lucko I think that event would be fine and if someone asks later, we can address then.

lucko:
@Zidane Does this seem ok?

https://github.com/lucko/SpongeAPI/commit/6d73959cc772e8c0619d170c894a230a6277479e

Zidane:
@lucko Can you open a PR so we can get some comments going on it? Closing this in favor of that.

PR #1414, opened 2016-11-17
Keys Docs Fix

Maxqia:


kashike:
Please add this to the [OCD issue](https://github.com/SpongePowered/SpongeAPI/issues/1000) if you don't plan on fixing more than this.

PR #1416, opened 2016-11-19
Add Extent.getHighestYAt

### LINKED ISSUE #1375: Get Highest Block at Specific Location

Maxqia:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1015)
This replaces #1381 and fixes #1375.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Location.java
+++ b/src/main/java/org/spongepowered/api/world/Location.java
@@ -346,6 +346,20 @@ public final class Location<E extends Extent> implements DataHolder {
     }
 
     /**
+     * Create a new instance with a new block position.
+     *
+     * @param position The new position
+     * @return A new instance
+     */
+    public Location<E> setBlockPosition(Vector3i position) {
+        checkNotNull(position, "position");
+        if (position == getBlockPosition()) {
+            return this;
+        }
+        return new Location<>(getExtent(), position);
+    }
+
+    /**
      * Subtract another Vector3d to the position on this instance, returning
      * a new Location instance.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -329,6 +329,25 @@ default void setSerializationBehavior(SerializationBehavior behavior) {
...
+    int getHighestYAt(int x, int z);
+
+    /**
+     * Get the {@link Location} of the highest block that sunlight can reach in the given column.
+     * <p>This method ignores all transparent blocks, providing the highest opaque block.</p>
+     *
+     * @param position The column position
+     * @return The highest opaque block
+     */
+    default Location<World> getHighestBlockAt(Vector3i position) {
```
kashike: Why `Vector3i` and not `Vector2i`?

kashike (changes requested):

src/main/java/org/spongepowered/api/world/World.java
Maxqia: "Well it makes it easier to do something like this world.getHighestBlockAt(location.getBlockPosition()). "

JBYoshi: Maybe you could add a method for both?

Maxqia: That'd be pointless, nothing with locations uses Vector2i

src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -680,6 +694,15 @@ public boolean spawnEntities(Iterable<? extends Entity> entities, Cause cause) {
         return this.getExtent().spawnEntities(entities, cause);
     }
 
+    /**
+     * Gets the highest Location at this Location
+     * @return the highest Location at this Location
```
gabizou, line 699: New line between description and `@return`

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -106,6 +106,25 @@
     }
 
     /**
+     * Get the y value of the highest block that sunlight can reach in the given column.
```
gabizou, line 109: Line length is longer than 80 characters. Please run `gradle checkstyle` to fix your warnings.

gabizou (changes requested):
Checkstyle issues. Documentation style issues.

kashike:
@Maxqia Able to get this updated?

PR #1417, opened 2016-11-19
Add methods for getting entity sounds for an entity

Maxqia:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1026)
I had this lying around for a month or so..., so here you go!
There might be some sounds missing and if there are, please tell me where I can find them.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -35,6 +35,7 @@ import org.spongepowered.api.data.DataTransactionResult;
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.mutable.TargetedLocationData;
 import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.cause.entity.damage.source.DamageSource; 
 import org.spongepowered.api.text.translation.Translatable;
@@ -560,4 +561,17 @@ public interface Entity extends Identifiable, Locatable, DataHolder, DataSeriali
         return getValue(Keys.HAS_GRAVITY).get();
     }
 
+    /**
+     * Gets the {@link SoundType} that is played when this entity moves in water.
+     *
+     * @return The SoundType for noise this entity makes when it swims
+     */
+    SoundType getSwimSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity jumps in water.
+     *
+     * @return The SoundType for the noise this entity makes when it enters water
+     */
+    SoundType getSplashSound();
 }
--- a/src/main/java/org/spongepowered/api/entity/living/Living.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Living.java
@@ -33,6 +33,7 @@ import org.spongepowered.api.data.property.entity.EyeLocationProperty;
 import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
 import org.spongepowered.api.data.value.mutable.OptionalValue;
 import org.spongepowered.api.effect.potion.PotionEffect;
+import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.scoreboard.TeamMember;
 
@@ -108,6 +109,43 @@ public interface Living extends Entity, TeamMember {
     default OptionalValue<Double> lastDamage() {
         return getValue(Keys.LAST_DAMAGE).get();
     }
+
+    /**
+     * Gets the volume that the sounds for this entity are played
+     *
+     * @return The volume of the sounds for this entity
+     */
+    double getVolume();
+
+    /**
+     * Gets the pitch for the sounds for this entity
+     * (e.g for if it's a baby)
+     *
+     * @return The pitch of the sounds for this entity
+     */
+    double getPitch();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity gets hurt.
+     *
+     * @return The {@link SoundType} that is played when this entity gets hurt
+     */
+    SoundType getHurtSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity dies.
+     *
+     * @return The {@link SoundType} that is played when this entity dies
+     */
+    SoundType getDeathSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity falls
+     * This return a big or small fall sound based on how far the entity is specified to fall.
+     * @param distance How far the entity fell
+     * @return The {@link SoundType} that is played when this entity falls that distance
+     */
+    SoundType getFallSound(int distance);
 
     /**
      * Returns this entity's head rotation.


src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -560,4 +561,17 @@ default boolean canSee(Entity entity) {
...
+     */
+    SoundType getSwimSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity jumps in water.
+     *
+     * @return The SoundType for the noise this entity makes when it enters water
+     */
+    SoundType getSplashSound();
 }
```
liach, line 577: You need an extra line before this line.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     * Gets the {@link SoundType} that is played when this entity dies.
+     *
+     * @return The {@link SoundType} that is played when this entity dies
+     */
+    SoundType getDeathSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity falls
+     * This return a big or small fall sound based on how far the entity is specified to fall.
+     * @param distance How far the entity fell
```
liach, line 145: You need an extra empty line of Javadoc before this line.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     * @return The {@link SoundType} that is played when this entity falls that distance
+     */
+    SoundType getFallSound(int distance);
```
liach, line 148: You should probably use a `double` for such distance.

liach (changes requested):

liach:
In general, the best way to retrieve a sound is to provide a context.

ST-DDT:
In https://github.com/SpongePowered/SpongeAPI/pull/1376 / https://github.com/SpongePowered/SpongeAPI/commit/4ae202fa5cfb99824177851c45352863640d3704 we went for `BlockSoundGroup` maybe we should do the same here for `EntitySoundGroup`?

Maybe we could also structure it a little bit:

soundgroup
[Picture/DOT-SourceCode](http://pastebin.com/keUkcCCe)

How about screaming/growling/angry pigmen...
Also there is the "idle" sound. Afaict its called `getAmbientSound()`

[Here](http://pastebin.com/LqrDygjD) is a list of remaining mob sounds for 1.10 (without death, hit and hurt)

ST-DDT:
> In general, the best way to retrieve a sound is to provide a context.

@liach What context are you referring to?

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
     }
 
     /**
+     * Gets the volume that the sounds for this entity are played
```
ST-DDT, line 114: Missing dot at the end of sentence.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     *
+     * @return The volume of the sounds for this entity
+     */
+    double getVolume();
+
+    /**
+     * Gets the pitch for the sounds for this entity
+     * (e.g for if it's a baby)
```
ST-DDT, line 122: Missing dot at the end of sentence.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     *
+     * @return The pitch of the sounds for this entity
+     */
+    double getPitch();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity gets hurt.
+     *
+     * @return The {@link SoundType} that is played when this entity gets hurt
```
ST-DDT, line 131: I guess its not necessary to link to the returned class here again.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     */
+    SoundType getHurtSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity dies.
+     *
+     * @return The {@link SoundType} that is played when this entity dies
```
ST-DDT, line 138: I guess its not necessary to link to the returned class here again.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -110,6 +111,43 @@ default DamageableData getMortalData() {
...
+     */
+    SoundType getDeathSound();
+
+    /**
+     * Gets the {@link SoundType} that is played when this entity falls
+     * This return a big or small fall sound based on how far the entity is specified to fall.
+     * @param distance How far the entity fell
+     * @return The {@link SoundType} that is played when this entity falls that distance
```
ST-DDT, line 146: I guess its not necessary to link to the returned class here again.

src/main/java/org/spongepowered/api/entity/living/Living.java
ST-DDT: Just curious: Is there a difference of sound depending on the fall distance? (I haven't noticed it myself yet, but I just may be deaf.^^) Maybe we can expose the short and long sounds directly.

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -560,4 +561,17 @@ default boolean canSee(Entity entity) {
         return getValue(Keys.HAS_GRAVITY).get();
     }
 
+    /**
+     * Gets the {@link SoundType} that is played when this entity moves in water.
+     *
+     * @return The SoundType for noise this entity makes when it swims
```
ST-DDT, line 567: IMO these sentences could be simplified to:

```
The sound that the entity emits when it ...
```

Same goes for the other methods. Not 100% sure though.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/entity/living/Living.java
liach: Yes, if you have noticed vanilla sound types. There is a small fall and a big fall.

Maxqia:
I did it mainly for simplicity of the implementation, because the methods and fields for getting sounds for entities are in the entity class, while block sounds are in a different class.

stephan-gh:
@Maxqia 
In my opinion the structure @ST-DDT suggested makes sense and is actually quite simple to implement:

It shouldn't take long to create another class in the implementation that links to the internal Entity class and just returns the sounds. If you really have to you could even make the internal Entity class implement `EntitySoundGroup` directly and just return `this` when `getSoundGroup()` is called on `Entity`.

The purpose of SpongeAPI is not to represent the internal classes exactly but to hide most of the internal structure and provide it in a logical way to plugin developers.

Maxqia:
I might make it actually so that BlockSoundGroup is integrated into BlockType, because making a sound group for (practically) every entity is going to make me lose my sanity :P
EDIT : it also makes a little more sense logically

kashike:
@Maxqia Status on the comments above?

kashike:
@Maxqia Status on the comments above?

Zidane:
@kashike is taking over this PR.

PR #1419, opened 2016-11-19
Recipe Refactor

Maxqia:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1027)
Test Code : https://gist.github.com/Maxqia/5579f49f9cfd3f4b08d6912d899d259d


src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
+package org.spongepowered.api.item.recipe;
```
kashike: Missing licence header.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+ * A RecipeRegistry holds all registered smelting recipes for a given game.
+ */
+public interface SmeltingRegistry {
+    /**
+     * Adds a smelting recipe to this registry.
+     * @param input the input of the recipe
+     * @param output the {@link ItemStack} the recipe gives
+     * @param exp the amount of exp the recipe gives
+     */
+    void addSmeltingRecipe(ItemStack input, ItemStack output, float exp);
```
liach: Use `ItemStackSnapshot`.

kashike (changes requested):

Cybermaxke:
I would like to see a `FurnaceRecipe` and a `Builder` similar to the recipe system for crafting grids. I made the following draft to give a general idea, you can also checkout the package: `https://github.com/SpongePowered/SpongeAPI/tree/bleeding/src/main/java/org/spongepowered/api/item/recipe`

``` java
public interface SmeltingRecipeRegistry {

    void register(SmeltingRecipe recipe);

    void remove(SmeltingRecipe recipe);

    Set<SmeltingRecipe> getRecipes();

    Optional<SmeltingRecipe> match(ItemStack itemStack);

   // Move these to their own registry???

    void registerFuel(Fuel fuel);

    void removeFuel(Fuel fuel);

    Set<Fuel> getFuels();

    Optional<Fuel> matchFuel(ItemStack itemStack);
}
```

``` java
public interface SmeltingRecipe {

    ItemStackSnapshot getInput();

    ItemStackSnapshot getOutput();

    double getExperience();

    int getDuration(); // Duration may also be nice to have

    public interface Builder {

         // Builder stuff
    }
}
```

And maybe also custom fuel?

``` java
public interface Fuel {

     ItemStackSnapshot getInput();

    // The result item when the fuel is consumed, like a bucket from a lava bucket
     Optional<ItemStackSnapshot> getReturnItem();

     int getDuration();

     public interface Builder {

         // Builder stuff
    }
}
```

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.spongepowered.api.item.inventory.ItemStack;
+
+/**
+ * A RecipeRegistry holds all registered smelting recipes for a given game.
+ */
+public interface SmeltingRegistry {
+    /**
+     * Adds a smelting recipe to this registry.
+     * @param input the input of the recipe
```
ST-DDT: Needs an empty javadoc line before this line. Same goes for all methods in this class.
Also the param descriptions should start with a capital letter after the param name.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    ItemStack getMapping(ItemStack stack);
+
+    /**
+     * Gets the experience the {@link ItemStack} drops when it is made
+     * @param the {@link ItemStack}
+     * @return the experience of the {@link ItemStack}
+     */
+    float getExperience(ItemStack stack);
+}
```
ST-DDT: Please add an empty line before this line.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+     * @return if it removed a recipe
+     */
+    boolean removeSmeltingRecipe(ItemStack input);
+
+    /**
+     * Gets what this {@link ItemStack} makes
+     * @param the {@link ItemStack}
+     * @return the {@link ItemStack} it makes
+     */
+    ItemStack getMapping(ItemStack stack);
```
ST-DDT: This should return Optionals for materials that are not registered/smeltable.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,37 @@
+package org.spongepowered.api.item.recipe;
+
+import org.spongepowered.api.item.inventory.ItemStack;
+
+/**
+ * A RecipeRegistry holds all registered smelting recipes for a given game.
+ */
+public interface SmeltingRegistry {
+    /**
```
ST-DDT: Missing newline before this line.

ST-DDT (changes requested):

Maxqia:
@Cybermaxke I was going to do that, until I realized that it would just be alot of = statements ...

liach:
@Maxqia What are a lot of `=` statements? The suggested one is way more logical.

liach:
@Cybermaxke Minecraft Forge manages fuel registry with a list of `IFuelHandler`s which returns the maximum burn time for a specific item stack, calling all registered handlers each time. Your `Fuel` interface probably cannot represent that.

Maxqia:
eg. 
SpongeSmeltingRecipe :

setInput(ItemStack input) {
    this.input = input;
}
build() {
    return new SpongeSmeltingStack(input,output,exp);
}

SpongeSmeltingStack :

ItemStack input,output;
float exp;
SpongeSmeltingStack(ItemStack input, ItemStack output, float exp) {
    this.input = input; etc...
}

Maxqia:
Also SmeltingRegistry.register(SmeltingRecipe.builder().setInput(input).setOutput(output).setExp(exp).build()) -> SmeltingRegistry.addSmeltingRecipe(input,output,exp)

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    boolean removeSmeltingRecipe(ItemStack input);
+
+    /**
+     * Gets what this {@link ItemStack} makes.
+     *
+     * @param stack the {@link ItemStack}
+     * @return the {@link ItemStack} it makes
+     */
+    ItemStack getMapping(ItemStack stack);
```
kashike: This should return `Optional<ItemStack>`.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
Maxqia: Oops, forgot about that ...

Cybermaxke:
@Maxqia The sponge environment is built around `Builder`s, this should't be an exception. You can always add helper methods to the `SmeltingRecipe` to simplify the creation.

@liach The implementation should be possible with the exception of the `getFuels` method, it would be a bit tricky though.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -313,11 +314,18 @@
...
+     */
+    CraftingRegistry getCraftingRegistry();
+
+    /**
+     * Retrieves the SmeltingRegistry for this GameRegistry.
      *
-     * @return The recipe registry
+     * @return The smelting registry
      */
-    RecipeRegistry getRecipeRegistry();
```
gabizou, line 299: This needs to be left in with `@Deprecated` to just redirect to the `CraftingRegistry`.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -47,29 +44,10 @@
...
 
     /**
-     * Returns the list of item types that result when successful crafting of
-     * this Recipe is completed.
+     * Gets the result of this {@link Recipe}.
      *
-     * @return The resultant list of item types
+     * @return the results of this {@link Recipe}
      */
-    List<ItemType> getResultTypes();
```
gabizou, line 55: Likewise with the previous deprecated method, this should be left in with `@Deprecated` to use the `getResults()` list.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -47,29 +44,10 @@
...
-
-    /**
-     * Checks if the given {@link GridInventory} fits the required constraints
-     * to craft this Recipe.
-     *
-     * @param grid The ItemGrid to check for validity
-     * @return True if the given input matches this recipe's requirements
-     */
-    boolean isValid(GridInventory grid);
```
gabizou, line 64: Should be left in as deprecated, potentially default implemented to return true? I'm not sure about it, but try to have as minimal of a breakage as possible.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -27,29 +27,32 @@
...
-    void remove(Recipe recipe);
+    void remove(Recipe recipe) throws IllegalArgumentException;
 
     /**
      * Retrieves all recipes registered in this registry.
      *
      * @return All registered recipes
      */
     Set<Recipe> getRecipes();
-
```
gabizou, line 54: Unnecessary whitespace removal.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -27,29 +27,32 @@
...
 
     /**
      * Registers the given Recipe to make it available to craft.
      *
      * @param recipe The Recipe to register
+     * @throws IllegalArgumentException if it's the wrong
+     *     type of recipe for this registry
      */
-    void register(Recipe recipe);
+    void register(Recipe recipe) throws IllegalArgumentException;
```
gabizou: This is screaming about how it should use generics instead of just a raw `Recipe` type.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the experience given when this recipe is made.
+     *
+     * @return the experience given when this recipe is made
+     */
+    float getExperience();
```
gabizou: Never use `float`, you gain nothing from it. Use `double`.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Gets the item needed to make this recipe.
+     *
+     * @return item needed to make this recipe
+     */
+    ItemStack getIngredient();
```
gabizou: I'm curious why you're using an `ItemStack` instead of an `ItemStackSnapshot`.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Gets the result of this {@link Recipe}.
+     *
+     * @return the result of this {@link Recipe}
+     */
+    ItemStack getResult();
```
gabizou: Likewise here.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface SmeltingRegistry extends RecipeRegistry {
+
+    /**
+     * Gets the experience an {@link ItemStack} drops when it is made.
+     *
+     * @param stack the {@link ItemStack}
+     * @return the experience of the {@link ItemStack}
+     */
+    Optional<Float> getExperience(ItemStack stack);
```
gabizou: Should use `Optional<Double>` or if you want, `java.util.OptionalDouble`.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Returns the results for running this SmeltingRegistry over an {@link ItemStack}
+     *
+     * @param stack An ItemStack as input
+     * @return A ItemStack or {@link Optional#empty()} if the given
+     *          ItemStack does not match any recipes.
+     */
+    Optional<ItemStack> getResult(ItemStack stack);
```
gabizou: Likewise, here, `Optional<ItemStackSnapshot>` instead.

gabizou (changes requested):
Some discussion is necessary on the use of `ItemStack`s instead of `ItemStackSnapshot`s. Since there's a big difference between the two (recipes in the underlying implementation aren't meant to change, but seeing as there's no immutable implementation aside from Sponge's, that's a different issue). To the API, we should be marking the objects as immutable as much as possible, when necessary.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    interface Builder extends ResettableBuilder<SmeltingRecipe, Builder> {
+        /**
+         * Sets the ingredient for the requirements of this {@link SmeltingRecipe}.
+         *
+         * @param ingredient The ingredient
+         * @return fluent interface
+         */
+        Builder setIngredient(ItemStack ingredient);
```
Cybermaxke: You don't have to prefix the builder methods with `set`, just use for example `ingredient`.

kashike: You *should not* prefix builder methods with `set`.

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+    ItemStack getIngredient();
+
+    /**
+     * Gets the result of this {@link Recipe}.
+     *
+     * @return the result of this {@link Recipe}
+     */
+    ItemStack getResult();
+
+    default List<ItemStack> getResults() {
```
liach: `@Override`?

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+package org.spongepowered.api.item.recipe;
+
+import com.google.common.collect.ImmutableList;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.List;
+
+public interface SmeltingRecipe extends Recipe {
```
liach: javadoc

liach (changes requested):

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
Maxqia: Was just about to say that to myself :P

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
```diff
@@ -0,0 +1,104 @@
...
+    default List<ItemStack> getResults() {
+        return ImmutableList.of(getResult());
+    }
+
+    interface Builder extends ResettableBuilder<SmeltingRecipe, Builder> {
+        /**
+         * Sets the ingredient for the requirements of this {@link SmeltingRecipe}.
+         *
+         * @param ingredient The ingredient
+         * @return fluent interface
```
ST-DDT: Maybe change this to `@return This instance for chaining` (see other builders)

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
Maxqia: I mean there's nothing to break... it was always broken

src/main/java/org/spongepowered/api/item/recipe/SmeltingRecipe.java
Maxqia: Oh, I just copied that part of the comment from the other recipe builders, I'll change them too while I'm at it

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public interface SmeltingRegistry extends RecipeRegistry<SmeltingRecipe> {
+
+    /**
+     * Gets the experience an {@link ItemStack} drops when it is made.
+     *
+     * @param stack the {@link ItemStack}
+     * @return the experience of the {@link ItemStack}
+     */
+    Optional<Double> getExperience(ItemStack stack);
```
liach: Use `java.util.OptionalDouble`

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Returns the results for running this SmeltingRegistry over an {@link ItemStack}
+     *
+     * @param stack An ItemStack as input
+     * @return A ItemStack or {@link Optional#empty()} if the given
+     *          ItemStack does not match any recipes.
+     */
+    Optional<ItemStack> getResult(ItemStack stack);
+}
```
liach, line 55: line

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,121 @@
...
+         */
+        Builder experience(float exp);
+
+        /**
+         * Builds a new {@link SmeltingRecipe} from this builder.
+         *
+         * @return A new {@link SmeltingRecipe}
+         */
+        SmeltingRecipe build();
+    }
```
liach, line 119: line

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,121 @@
...
+         */
+        Builder result(ItemStack result);
+
+        /**
+         * Sets the experience given when this {@link SmeltingRecipe} is crafted.
+         *
+         * @param exp the experience
+         * @return This builder, for chaining
+         */
+        Builder experience(float exp);
```
liach: accept double

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,121 @@
...
+     * @return item needed to make this recipe
+     */
+    ItemStack getIngredient();
+
+    /**
+     * Gets the result of this {@link Recipe}.
+     *
+     * @return the result of this {@link Recipe}
+     */
+    ItemStack getResult();
```
liach: `ItemStackSnapshot`

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+    boolean isValid(GridInventory grid);
+
+    /**
+     * Returns the results for running this CraftingRegistry over an {@link GridInventory}
+     *
+     * @param grid An ItemGrid as input
+     * @return A list of ItemStacks or {@link Optional#empty()} if the given
+     *          {@link GridInventory} does not match any recipes.
+     */
+    Optional<List<ItemStack>> getResults(GridInventory grid);
```
liach: Can it return an empty list for nothing?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+}
```
liach, line 48: line

liach (changes requested):

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
gabizou: @Maxqia There's a difference between binary compatibility and something being inherently broken. This is a binary breakage, anything relying on these methods (while they should be deprecated) would ultimately fail to work between API builds. I don't see the reason why this wouldn't be possible to retain for the sake of deprecation and uselessness (you can even print out a pretty printed log if you want)

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -24,32 +24,36 @@
...
      * @param recipe The Recipe to register
      */
-    void register(Recipe recipe);
+    void register(T recipe);
 
     /**
      * Removes the given Recipe from registration in this registry.
      *
      * @param recipe The Recipe to unregister
+     * @return if it was successful or not
```
gabizou: Capitalize the `if`

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
gabizou: Yeah, have to agree here, why `Optional` instead of just an empty list?

Maxqia: IDK, just moved it from Recipe....

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
Maxqia: No, it doesn't have a bunch of functions that Optional has ...

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
Maxqia: missed that, will do

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
kashike: https://github.com/SpongePowered/SpongeAPI/pull/1098/files#diff-533288ff4558c56e3fb98c90df746cafR66

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -47,29 +44,18 @@
...
-
-    /**
-     * Returns the results for running this Recipe over an {@link GridInventory}
-     *
-     * @param grid An ItemGrid as input
-     * @return A list of ItemStacks or {@link Optional#empty()} if the given
-     *          ItemGrid does not match this recipe's requirements.
-     */
-    Optional<List<ItemStack>> getResults(GridInventory grid);
+    RecipeRegistry<?> getRegistry();
```
ST-DDT, line 58: If this recipe is restricted to a single registry you should throw an IllegalArgumentException if you try to register it in a different one.

Is there a builder method for this one?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public interface CraftingRegistry extends RecipeRegistry<CraftingRecipe> {
+    /**
+     * Checks if the given {@link GridInventory} fits the required constraints
+     * to craft this Recipe.
+     *
+     * @param grid The ItemGrid to check for validity
+     * @return True if the given input matches this recipe's requirements
+     */
+    boolean isValid(GridInventory grid);
```
ST-DDT: ImO this method would be better returning the matching recipe. The javadoc are strange currently.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -45,37 +45,36 @@ static Builder builder() {
...
     }
 
-
     /**
      * Gets the ingredients for this ShapelessRecipe.
      *
      * @return The ingredients
      */
-    Collection<ItemStack> getIngredients();
+    Collection<ItemStackSnapshot> getIngredients();
```
Zidane: Getting ingredients as snapshots makes sense but I wouldn't require them when adding to the builder. Makes working with the API painful (when it could turn them into snapshots itself).

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -48,7 +47,7 @@
      *
      * @return the results of this {@link Recipe}
      */
-    Collection<ItemStackSnapshot> getResults();
+    ImmutableCollection<ItemStackSnapshot> getResults();
```
ST-DDT, line 50: I would prefer if this returns any kind of `Collection` instead of this specialized one.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -24,32 +24,37 @@
...
-import java.util.Set;
+import com.google.common.collect.ImmutableCollection;
+import org.spongepowered.api.item.recipe.crafting.CraftingRegistry;
+import org.spongepowered.api.item.recipe.smelting.SmeltingRegistry;
+
+import java.util.function.Predicate;
 
 /**
- * A RecipeRegistry holds all registered recipes for a given game.
+ * Used for {@link SmeltingRegistry} and {@link CraftingRegistry}
```
jamierocks, line 34: Missing full stop at end of sentence.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -47,29 +43,18 @@
...
-     * Checks if the given {@link GridInventory} fits the required constraints
-     * to craft this Recipe.
+     * Gets the result of this {@link Recipe}.
      *
-     * @param grid The ItemGrid to check for validity
-     * @return True if the given input matches this recipe's requirements
+     * @return the results of this {@link Recipe}
      */
-    boolean isValid(GridInventory grid);
+    ImmutableCollection<ItemStackSnapshot> getResults();
```
liach, line 50: `Collection`

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -24,32 +24,37 @@
...
-    void remove(Recipe recipe);
+    boolean remove(Predicate<? super T> predicate);
 
     /**
      * Retrieves all recipes registered in this registry.
      *
      * @return All registered recipes
      */
-    Set<Recipe> getRecipes();
+    ImmutableCollection<T> getRecipes();
```
liach, line 58: `Collection`

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+public interface CraftingRegistry extends RecipeRegistry<CraftingRecipe> {
+
+    /**
+     * Returns the results for running this CraftingRegistry over an {@link GridInventory}
+     *
+     * @param grid An ItemGrid as input
+     * @return A list of ItemStacks or {@link Optional#empty()} if the given
+     *          {@link GridInventory} does not match any recipes.
+     */
+    Optional<ImmutableCollection<ItemStackSnapshot>> getResults(GridInventory grid);
```
liach, line 46: `Collection<ItemStackSnapshot>`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,161 @@
...
+     * @return The aisle
+     */
+    ImmutableCollection<String> getShape();
+
+    /**
+     * Gets the ingredients.
+     *
+     * @return The ingredients
+     */
+    ImmutableMap<Character, ItemStackSnapshot> getIngredients();
```
liach, line 68: `Map`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * Gets the shape of the recipe.
+     *
+     * <p>This returns an unmodifiable copy of the original aisle.</p>
+     *
+     * @return The aisle
+     */
+    ImmutableCollection<String> getShape();
```
liach, line 61: `List`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,128 @@
...
+
+        /**
+         * Sets the ingredients of this shapeless recipe.
+         * (Minecraft codes it so that it only takes one item of a type)
+         *
+         * @param ingredients The ingredients
+         * @return The builder
+         */
+        default Builder ingredients(ItemStack... ingredients) {
+            return ingredients(ingredients);
```
liach, line 88: Meh, doesn't even compile

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+    double getExperience(ItemStackSnapshot stack);
+
+    /**
+     * Returns the results for running this SmeltingRegistry over an {@link ItemType}.
+     *
+     * @param type the type to search for
+     * @return A {@link ItemStackSnapshot} or {@link Optional#empty()} if the given
+     *          {@link ItemType} does not match any recipes.
+     */
+    Optional<ItemStackSnapshot> getResult(ItemType type);
```
liach, line 54: Take `ItemStackSnapshot` as argument instead

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -24,32 +24,37 @@
...
     /**
-     * Removes the given Recipe from registration in this registry.
+     * Removes Recipes from this Registry using the given Predicate.
      *
-     * @param recipe The Recipe to unregister
+     * @param predicate The check to remove the Recipe or not
+     * @return If it removed any recipes or not
      */
-    void remove(Recipe recipe);
+    boolean remove(Predicate<? super T> predicate);
```
kashike, line 51: Don't remove the previous version of this method, add a new one.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
Maxqia: ? It does...

jamierocks: Will cause StackOverflowException

Maxqia: 0-0 oh...

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /**
+         * Sets the ingredients of this shapeless recipe.
+         * (Minecraft codes it so that it only takes one item of a type)
+         *
+         * @param ingredients The ingredients
+         * @return The builder
+         */
+        Builder ingredients(Collection<ItemStack> ingredients);
```
ST-DDT, line 98: No `ItemStackSnapshot` support? (Sorry for nagging)

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -47,29 +43,18 @@
...
+     * Gets the {@link RecipeRegistry} that
+     * corresponds with this Recipe.
      *
-     * @param grid An ItemGrid as input
-     * @return A list of ItemStacks or {@link Optional#empty()} if the given
-     *          ItemGrid does not match this recipe's requirements.
+     * @return the {@link RecipeRegistry} that corresponds with this Recipe
      */
-    Optional<List<ItemStack>> getResults(GridInventory grid);
+    RecipeRegistry<?> getRegistry();
```
ST-DDT, line 58: What is the reason for this method? 

I'm not really against it being here, but IMO its simply not-required overhead.

* If this is not yet/no longer registered, this would return null
* (if ever supported) custom impls might return the wrong registry
* I don't see a usecase

What are your thoughts adding this method?

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
Maxqia: Requested by Zidane to be an ItemStack :P https://github.com/SpongePowered/SpongeAPI/pull/1419/files/109bca54aacdfe4112c7a2b6ae536b5d86b62abb#diff-a6badc7df0748d1ae0caa7f85876c1c2

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
liach: Or `ItemStack`

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
Maxqia: It's immutable because we don't want people changing it...

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
Zidane: @Maxqia 

You can do both lol.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
Maxqia: I put ItemType there because it ignores stack size. I'll probably be changing getExperience to that too

liach: No, meta is super important. Stack size can simply be ignored.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
gabizou: I don't see the point in having this method. Just gut it.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
gabizou: Seconded, the removal of the recipe can still be retained.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
gabizou: Don't use `ImmutableCollection`, just `Collection` is fine (you can even say it's unmodifiable in the javadoc).

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,161 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the shape of the recipe.
+     *
+     * <p>This returns an unmodifiable copy of the original aisle.</p>
+     *
+     * @return The aisle
```
gabizou, line 59: The aisle or shape?

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /**
+         * Sets the ingredients of this shapeless recipe.
+         * (Minecraft codes it so that it only takes one item of a type)
+         *
+         * @param ingredients The ingredients
+         * @return The builder
+         */
+        default Builder ingredients(ItemStack... ingredients) {
```
gabizou, line 87: In general, avoid only adding vararg methods like this, if possible, you should have single argument methods and then two argument methods with the second argument being a vararg. This would simplify some logic.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,128 @@
...
+
+        /**
+         * Sets the ingredients of this shapeless recipe.
+         * (Minecraft codes it so that it only takes one item of a type)
+         *
+         * @param ingredients The ingredients
+         * @return The builder
+         */
+        default Builder ingredients(ItemType... ingredients) {
+            return ingredients(Lists.newArrayList(ingredients).stream().map(
```
gabizou, line 76: Fix this formatting, it looks bad. Refer to other areas where streams are used.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,121 @@
...
+    ItemStackSnapshot getIngredient();
+
+    /**
+     * Gets the result of this {@link Recipe}.
+     *
+     * @return the result of this {@link Recipe}
+     */
+    ItemStackSnapshot getResult();
+
+    default ImmutableCollection<ItemStackSnapshot> getResults() {
```
gabizou, line 73: You javadoc everything else but not this method? I'm confused as to why there's a single result and then a collection of other results. What other results can there be?

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRegistry.java
gabizou: The data on the `ItemStack` is 100% very important, not to just the `ItemType`.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
jamierocks: I'm fairly certain there are a number of mods which give out more than one result in the furnace, can't think of any off the top of my head though.

Maxqia: Oh, It's a defaulted method from Recipe :P. not sure if @Override works here

gabizou: @Maxqia Yes, `@Override` works here. You should always annotate the methods appropriately so it's plain and clear that the method doesn't originate from this interface/class and is an overridden method. Everything in Sponge follows this rule.

kashike:
This is more than likely going to need changes with 1.12 arrives.

gabizou:
@kashike While that is true, would it be worth having the changes being made so that recipes can be managed in API 6 and maybe 7?

kashike:
I'm going to close this in favour of #1582.

Issue #1395, opened 2016-10-22
Expose method to make a player spectate an entity.


ryantheleach:
Kane_Hart on IRC was requesting a plugin to allow admins to be able to spectate and scroll through players.

For best results a plugin should probably use the vanilla spectator mechanic as opposed to vanishing people and constantly updating their position.

```
/**
Attempts to make player spectate entity.
returns whether the method was successful.
this method will fail if the player is not in the correct Gamemode.
*/
bool player.setSpectating(@Nullable Entity entity);

Optional<Entity> player.getSpectating();
```

gravityfox:
I think there could be use cases for this beyond just for a spy plugin. Minigames could make use of this.

Like how in some games when you die, you can spectate your teammates.

parlough:
Is this ready to be closed with https://github.com/SpongePowered/SpongeAPI/pull/1422 in or are you looking for more/other functionality @ryantheleach?

ryantheleach:
The only other potential thing is events? if a player is in spectator, is there currently any way of preventing them from settings their own spectate target, or filtering the targets shown to them, if that's unwanted?

PR #1422, opened 2016-11-29
Spectator camera view

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1422) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1046)
Adds #1395


ryantheleach:
Does this work for *any* entity? or are only some Spectatable?

Also your common link, links to the API,

kashike:
You can spectate any `Entity` as far as I'm aware. There might be some limitations for stupid ones, like fish hooks, but that'd be done in vanilla.

Issue #1312, opened 2016-07-19
Stupidly Difficult Idea (TM) - Text Data Smart Trimming


gravityfox:
Text these days is really rather complicated, supporting numerous options and features. However, this also means that the JSON text can get really large really fast, even if the visible/interactable content is actually rather minimal.

This became really apparent in #1309 which has a bunch of seemingly excessive information.

While working on a plugin that generated a list, I have managed to break the 32k byte limit and disconnect clients with messages that are just too long. Pagination relieves the problem, but only by so much.

My idea was that it could be possible to take a complicated Text object, and simplify it without removing any of its content. This would be done by combining like but separate elements, removing empty ones, and collapsing any unnecessary hierarchy. Ideally we could evaluate a Text object and compress it down to the smallest size possible that still represents the same thing.

While this isn't a large or important optimization, it could be helpful to those plugins that just want to dump a lump of text into chat without disconnecting anyone.

Thoughts anyone?

theboomer:
would this require the CLIENT to have mods to deconvolute the modified datastream though?

JBYoshi:
No. Here's an example of what he's talking about:
```
{text: "Spo", children=[{text: "ngePo"}, {text: "wered"}]}
```
can be simplified to:
```
{text: "SpongePowered"}
```
without breaking protocol compatibility.

theboomer:
Ah, sweet example

ST-DDT:
Here an example implemententation: 
https://gist.github.com/ST-DDT/b296398917891345e8d4c160916bed48

**Output**
From @JBYoshi 's example

````java
test(Text.builder("Spon")
		.append(Text.of("gePo"))
		.append(Text.of("wered"))
		.build());
````

````text
Text{children=[Text{gePo}, Text{wered}], Spon}
Text{SpongePowered}
````
more complex example
````java
test(Text.of(FakeColor.C1,
		Text.of("Spon"),
		"ge",
		Text.of(FakeColor.C2),
		Text.of(FakeColor.C1, "Pow",
				Text.of("ered"))));
````

````text
[...]

PR #1424, opened 2016-12-04
Add simplified append and of methods to Text

Katrix:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2161)

Adds `Text#ofCompacted` and `Text#compact` which will try to compact the text when creating it, and when called.

See #1312 for discussion.

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -194,6 +194,26 @@ public interface Text extends Comparable<Text>, DataSerializable, TextRepresenta
     }
 
     /**
+     * Builds a compacted {@link Text} from a given array of objects.
+     *
+     * <p>Like the normal {@link #of(Object...)} method, but will try to
+     * compact the passed in text objects if possible. No guarantees are given
+     * regarding the structure of the {@link Text}, only that it will
+     * display the same.</p>
+     *
+     * <p></p>The result of this might be, but does not need to be equivalent to
+     * {@code Text.of(arrayOfObjects).compact()}.</p>
+     *
+     * @see #of(Object...) For more information on the normal of factory.
+     * @param objects The object array
+     * @return The compacted built text object
+     */
+    @SuppressWarnings("deprecation")
+    static Text ofCompacted(Object... objects) {
+        return Sponge.getRegistry().getTextFactory().ofCompacted(objects);
+    }
+
+    /**
      * Creates a {@link Text.Builder} with empty text.
      *
      * @return A new text builder with empty text
@@ -534,6 +554,17 @@ public interface Text extends Comparable<Text>, DataSerializable, TextRepresenta
      */
     Text trim();
 
+    /**
+     * Removes unneeded children from this {@link Text} and it's children
+     * while keeping the appearance the same.
+     *
+     * <p>No guarantees are given regarding the structure of the {@link Text},
+     * only that it will display the same.</p>
+     *
+     * @return Text result
+     */
+    Text compact();
+
     @Override
     default Text toText() {
         return this;
@@ -665,6 +696,18 @@ public interface Text extends Comparable<Text>, DataSerializable, TextRepresenta
          * @see Text#getChildren()
          */
         List<Text> getChildren();
+
+		/**
+		 * Gets if this builder should try to compact when appending text.
+		 *
+		 * @return If this builder will compact when appended text
+		 */
+		boolean shouldCompact();
+
+		/**
+		 * Sets if this builder should compact when appending text.
+		 */
+		Builder setCompact(boolean compact);
 
         /**
          * Appends the specified {@link Text} to the end of this text.
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -697,17 +697,17 @@ public interface Text extends Comparable<Text>, DataSerializable, TextRepresenta
          */
         List<Text> getChildren();
 
-		/**
-		 * Gets if this builder should try to compact when appending text.
-		 *
-		 * @return If this builder will compact when appended text
-		 */
-		boolean shouldCompact();
+        /**
+         * Gets if this builder should try to compact when appending text.
+         *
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -535,14 +545,66 @@ public Builder onShiftClick(@Nullable ShiftClickAction<?> shiftClickAction) {
         }
 
         /**
+         * Gets if this builder should try to simplify when appended text.
+         *
+         * @return If this builder will simplify when appended text.
+         */
+        public boolean getSimplify() {
```
kashike: `shouldSimplify`

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -535,14 +545,66 @@ public Builder onShiftClick(@Nullable ShiftClickAction<?> shiftClickAction) {
...
+        /**
+         * Appends the specified {@link Text} to the end of this Text,
+         * making sure to simplify if possible.
+         *
+         * @param child The text to append
+         * @return This text builder
+         */
+        public Builder appendSimplified(Text child) {
+            if (!child.isEmpty()) {
+                int lastIndex = this.children.size() - 1;
```
kashike: Does this code account for an empty `children` list?

kashike (changes requested):

src/main/java/org/spongepowered/api/text/Text.java
Katrix: Yes next line is a check for if `children` is empty or not. I guess I just forgot to move `lastIndex` into that block.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -250,6 +250,15 @@ public final boolean isEmpty() {
     public abstract Builder toBuilder();
 
     /**
+     * Check if this text can be converted to plain and back without loosing any information.
+     *
+     * @return {@code true} if this text is simple.
```
parlough: I don't think a code tag is needed here, could just say `@return If this text is simple`

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -250,6 +250,15 @@ public final boolean isEmpty() {
     public abstract Builder toBuilder();
 
     /**
+     * Check if this text can be converted to plain and back without loosing any information.
```
parlough: I think this is longer than 80 characters.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -535,14 +545,66 @@ public Builder onShiftClick(@Nullable ShiftClickAction<?> shiftClickAction) {
         }
 
         /**
+         * Gets if this builder should try to simplify when appended text.
+         *
+         * @return If this builder will simplify when appended text.
```
parlough: No period needed

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -535,14 +545,66 @@ public Builder onShiftClick(@Nullable ShiftClickAction<?> shiftClickAction) {
...
+         */
+        public boolean shouldSimplify() {
+            return this.simplify;
+        }
+
+        /**
+         * Sets if this builder should simply when appending text.
```
parlough: Did you mean "simplify" instead of "simply"?

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -1020,6 +1077,45 @@ public static Text of(Object... objects) {
     }
 
     /**
+     * Builds a simplified {@link Text} from a given array of objects.
+     * This method is identical in every way with {@link #of(Object...)}
```
parlough: Put everything starting with "this" in a new paragraph.

parlough (changes requested):
I suppose I see some use for this. Here's some comments relating to Javadocs.

ryantheleach:
On the assumption that this is ready and passes review for the implementation, Does the design as it stands match what was wanted in #1312 ? 

I still think there are some design questions to answer.


Correct me if I'm wrong, but 
As it stands in this PR, it seems that a 'simple' text is only literals, that do not have color, and do not have any mouse actions.

However, it also allows a text to have children that also match this.

In my mind this is not a 'simple' text. I disagree with the design and do not think it matches the expectations of the people in the ticket of #1312.

I see 4 potential designs of simple text.

1. Minimally nested, has the minimum 'depth' of the Text tree to accomplish the format, and hovers the user wants. This means compacting where necessary to flatten, if that makes it shorter (note that this isn't straight forward, due to inheriting default text colors)

2. Text that is just text, whose length can be calculated, and has no hovers or variables like selectors or scoreboards, or translations.

3. Completely flat text. has a single color, and can be split easily, whose length can be calculated.

4. Flat literaltext without color.

I believe 1. is the definition that people in the linked issue wanted.

I don't know whose definition of simple this PR is referring to, or how it is useful to users, I'd like to see proof or a user story for it's inclusion into API7 which it is currently milestoned for.

Katrix:
Yeah, I want to rework this into something better myself too. I've since managed to get at least somewhat better implementation than what I have here.

Katrix:
So to shed some light on where this is at the moment. I've been sitting on a working implementation for a while now. I cleaned it up some today and added further optimizations. However, I don't feel like I have a good test set to make sure nothing breaks horribly. Anyone mind sending me some busy serialized texts and check that they still look good when displayed?

Here's a current sample of before and after going through the optimizer. As you can see, it will absolutely destroy any form of structure. However, everything I've tried so far still displays correctly.

Before: http://jsoneditoronline.org/?id=1fdee1b7259c4cb51c557ffec11a9a20
After: http://jsoneditoronline.org/?id=c1be09111e2c7d9a8b17f08c5908b4e4

gabizou:
Needs to update for Text API v4 changes.

gabizou (approved):

JBYoshi:

PR #1428, opened 2016-12-09
Attributes

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1428) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1064)

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/attribute/Attribute.java
+++ b/src/main/java/org/spongepowered/api/entity/attribute/Attribute.java
@@ -1,0 +1,118 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.attribute;
+
+import org.spongepowered.api.entity.attribute.operation.AttributeOperation;
+import org.spongepowered.api.entity.attribute.type.AttributeType;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * An attribute instance.
+ */
+public interface Attribute {
+
+    /**
+     * Gets the type of this attribute.
+     *
+     * @return The attribute type
+     */
+    AttributeType getType();
+
+    /**
+     * Gets the base value of this attribute.
+     *
+     * <p>This is the value <i>before</i> modifiers are applied.</p>
+     *
+     * @return The base value
+     */
+    double getBaseValue();
+
+    /**
+     * Sets the base value of this attribute.
+     *
+     * <p>This is the value <i>before</i> modifiers are applied.</p>
+     *
+     * @param baseValue The base value
+     */
+    void setBaseValue(double baseValue);
+
+    /**
+     * Gets the value of this attribute.
+     *
+     * <p>This is the value with modifiers applied.</p>
+     *
+     * @return The value
+     */
+    double getValue();
+
+    /**
+     * Gets a collection of all applied modifiers.
+     *
+     * @return A collection of applied modifiers
+     */
+    Collection<AttributeModifier> getModifiers();
+
+    /**
+     * Gets a collection of applied modifiers with the provided operation.
+     *
+     * @return A collection of modifiers
+     */
+    Collection<AttributeModifier> getModifiers(AttributeOperation operation);
+
+    /**
+     * Checks if this attribute has the provided modifier.
+     *
+     * @param modifier The modifier
+     * @return {@code true} if this attribute has the modifier, {@code false}
+     *     otherwise
+     */
+    boolean hasModifier(AttributeModifier modifier);
+
+    /**
+     * Gets an attribute modifier by its unique id.
+     *
+     * @param uniqueId The unique id
+     * @return The attribute modifier, if present, {@link Optional#empty()}
+     *     otherwise
+     */
+    Optional<AttributeModifier> getModifier(UUID uniqueId);
+
+    /**
+     * Adds a modifier to this attribute.
+     *
+     * @param modifier The modifier
+     */
+    void addModifier(AttributeModifier modifier);
+
+    /**
+     * Removes a modifier from this attribute.
+     *
+     * @param modifier The modifier
+     */
+    void removeModifier(AttributeModifier modifier);
+}
--- a/src/main/java/org/spongepowered/api/entity/attribute/AttributeModifier.java
+++ b/src/main/java/org/spongepowered/api/entity/attribute/AttributeModifier.java
@@ -1,0 +1,116 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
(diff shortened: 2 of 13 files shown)


simon816:
Aren't most of these things available (or can be made available) through the data API?

kashike:
No.
http://minecraft.gamepedia.com/Attribute

src/main/java/org/spongepowered/api/entity/attribute/AttributeModifier.java
```diff
@@ -0,0 +1,33 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.attribute;
+
+import org.spongepowered.api.entity.attribute.operation.AttributeOperation;
+import org.spongepowered.api.util.Identifiable;
+
+public interface AttributeModifier extends Identifiable {
```
gabizou, line 37: Javadocs please ;)

Maxqia:
ooooh, I don't have to use NMS anymore for this : https://github.com/Maxqia/MiscStuff/blob/master/src/main/java/com/maxqia/miscstuff/DisableAttackTime.java, btw your Common link is accidentally referencing SpongeAPI

src/main/java/org/spongepowered/api/entity/attribute/AttributeModifier.java
```diff
@@ -0,0 +1,33 @@
...
+
```
Cybermaxke, line 38: The `getAmount` and `getName` methods are missing?

Cybermaxke:
The `Common` refers to a wrong pr.

src/main/java/org/spongepowered/api/entity/attribute/operation/AttributeOperation.java
```diff
@@ -0,0 +1,32 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.attribute.operation;
+
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(AttributeOperations.class)
```
gabizou, line 33: Could use some Javadocs.

gabizou (changes requested):
Javadocs please.

src/main/java/org/spongepowered/api/entity/attribute/Attribute.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.spongepowered.api.entity.attribute.operation.AttributeOperation;
+import org.spongepowered.api.entity.attribute.type.AttributeType;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * An attribute instance.
```
gabizou, line 35: Clarify? Is it just a container for some values? What does it represent? Who can own one?

src/main/java/org/spongepowered/api/entity/attribute/AttributeModifier.java
```diff
@@ -0,0 +1,116 @@
...
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.attribute.operation.AttributeOperation;
+import org.spongepowered.api.util.Identifiable;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.UUID;
+
+/**
+ * An attribute modifier.
```
gabizou, line 35: What are these for? What do they modify on an `Attribute`? How do they interact with each other?

src/main/java/org/spongepowered/api/entity/attribute/operation/AttributeOperation.java
```diff
@@ -0,0 +1,36 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.attribute.operation;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * An attribute operation.
```
gabizou, line 31: An operation that does "what"?

gabizou (changes requested):
Clarification in javadocs needed. Also, wouldn't this PR fit to have some events? Or usability with Data API being able to perform more modifications than just getting an `Attribute` instance from `Living`?

randombyte-developer:
What's the status of this?

HaedHutner:
Also interested in a status update on this. Will it make it into API7?

parlough:
@HaedHutner @randombyte-developer I do not think so unfortunately, but I will see what I can do as this has been a popular request recently.

HaedHutner:
@Meronat What else needs to be done on this? I'd be willing to help out

ryantheleach:
Kashike has some changes locally on this that need to be finished, If anyone picks this up, I'd recommend contacting them first to see what direction it was heading in.

BrainStone:
@kashike what's the status on this? I'd love to help out, as I really want to see this implemented!

Also knowing what needs to be done would be helpful. I'd highly apprechiate if you get in contact with me. If you don't mind I'd apprechiate getting write access to your branches, so I don't have to fork it myself and create another PR.

gabizou:
@BrainStone I'd recommend forking and PR'ing due to the time difference since the original commit from kashike and the current "status" of the API. As the API has evolved since 2016, there are some concepts that need to be changed with how registrations are handled, and other things (look at the registration event for example, build new attributes and register them there). I'd say that it would be safe to start from there, and the implementation as well (can assist with updating the implementation to bring it "up to speed" with other changes, but definitely a "test" would be to write a test plugin that takes advantage of the API and operates with the basis of how the Attribute API can be used by plugins.

Faithcaio:
@kashike status?

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeTypes.java
```diff
@@ -0,0 +1,84 @@
...
+        public static final RangedAttributeType FOLLOW_RANGE = DummyObjectProvider.createFor(RangedAttributeType.class, "FOLLOW_RANGE");
+
+        public static final RangedAttributeType KNOCKBACK_RESISTANCE = DummyObjectProvider.createFor(RangedAttributeType.class, "KNOCKBACK_RESISTANCE");
+
+        public static final RangedAttributeType LUCK = DummyObjectProvider.createFor(RangedAttributeType.class, "LUCK");
+
+        public static final RangedAttributeType MAX_HEALTH = DummyObjectProvider.createFor(RangedAttributeType.class, "MAX_HEALTH");
+
+        public static final RangedAttributeType MOVEMENT_SPEED = DummyObjectProvider.createFor(RangedAttributeType.class, "MOVEMENT_SPEED");
+
```
Cybermaxke, line 51: Should we also support the `generic.reachDistance` attribute which is added by forge?

Cybermaxke:
How can you get/set attribute modifiers on `ItemStack`s?

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeTypes.java
parlough: Is this something we can implement in vanilla? Does it need to be synced with the client?

Cybermaxke: It will work nicer that way yes, but you can also make it server side only by blocking interaction if the reach distance is below the client distance, or handle custom interaction if it's the other way around (could also just limit this to forge).

src/main/java/org/spongepowered/api/entity/attribute/type/RangedAttributeType.java
```diff
@@ -0,0 +1,53 @@
...
+ * An attribute type with a ranged value.
+ */
+public interface RangedAttributeType extends AttributeType {
+
+    /**
+     * Gets the description of the attribute.
+     *
+     * @return The description
+     */
+    String getDescription();
```
ST-DDT: Why is this method here? Can't we move it to `AttributeType`?
Maybe add a separate interface for cases where an description is present.

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -184,4 +188,12 @@ default void lookAt(Vector3d targetPos) {
         setRotation(new Vector3d(pitchDeg, yawDeg, getRotation().getZ()));
     }
 
+    /**
+     * Gets the attribute by its type.
+     *
+     * @param type The attribute type
+     * @return The attribute, if present, {@link Optional#empty()} otherwise
+     */
+    Optional<Attribute> getAttribute(AttributeType type);
```
Cybermaxke: Move this to a `AttributeHolder` interface and use this in `Living`.

JBYoshi:
How's this going? I saw a few people asking about this on the forums.

Sago92:
I am comming from the Sponge forum and I am looking forward for the completion of this implementation. This would be an essential part in my RPG pluing (for example manipulating attack speed of entities and so on ...) Any plan for the finish line?

SirFancyBacon:
I also am looking for a way to adjust attack speed on entities.

Sago92:
Closed without any final statement? What is the future plan for the ItemStack attributes like attack speed?

gabizou:
The branch needs to be rebased/recreated based on `api-8`. Bleeding and 1.13 are severely outdated, so they were deleted. Unfortunately, GitHub doesn't have a way to automatically re-target the PR's that were targeting those branches, so it auto-closes them. It'll be looked at after some deployments get working to build `api-8`, but for now this PR needs to be rebased and rebuilt.

PR #1430, opened 2016-12-11
Add context weight

lucko:
Adds a weight value to Contexts. 

Intended use is that plugins implementing Contextual services can determine which contexts to check for first.

The current (afaik) way to determine context weight is to simply look at the number of entries in the `Set<Context>`. This change allows you to determine an overall weight based upon the combined weights of the contained contexts.

Supposing the following set is passed in a calculator:
`[A, B, C]`

If an exact match was not found for those contexts, then the implementation (should) will randomly check either `[A, B]` or `[B, C]`. Ordering is important for some plugins. (cc: @bloodmc )

Whilst this elimination system isn't a particularly effective way to flatten context sets, it's the easiest way to explain it. :)

The hashcode and equals implementations were taken from Guava's AbstractMapEntry. (what was being used in the previous implementation of those methods)

### DIFF
--- a/src/main/java/org/spongepowered/api/service/context/Context.java
+++ b/src/main/java/org/spongepowered/api/service/context/Context.java
@@ -24,9 +24,10 @@
  */
 package org.spongepowered.api.service.context;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.collect.Maps;
+import com.google.common.base.Objects;
 
 import java.util.Map;
 
@@ -44,18 +45,32 @@ public final class Context implements Map.Entry<String, String> {
     public static final String LOCAL_IP_KEY = "localip";
     public static final String LOCAL_PORT_KEY = "localport"; // This portkey is an old boot
 
-    private final Map.Entry<String, String> wrapped;
+    private final String key;
+    private final String value;
+    private final int weight;
 
     /**
      * Create a new context instance
      *
      * @param type Context type. Must not be null.
      * @param name Context name. Must not be null.
+     * @param weight Context weight. Must be greater than or equal to 1. Default is 1.
+     */
+    public Context(String type, String name, int weight) {
+        this.key = checkNotNull(type, "type");
+        this.value = checkNotNull(name, "name");
+        checkArgument(weight >= 1, "weight must be >= 1");
+        this.weight = weight;
+    }
+
+    /**
+     * Create a new context instance with the default weight
+     *
+     * @param type Context type. Must not be null.
+     * @param name Context name. Must not be null.
      */
     public Context(String type, String name) {
-        checkNotNull(type, "type");
-        checkNotNull(name, "name");
-        this.wrapped = Maps.immutableEntry(type, name);
+        this(type, name, 1);
     }
 
     /**
@@ -74,20 +89,32 @@ public final class Context implements Map.Entry<String, String> {
      * @return the specific name of the item involved in this context, for
      *         example if the type were {@code world} this would be the name of the
      *         world.
-     * 
      */
     public String getName() {
         return getValue();
     }
 
+    /**
+     * Get the weight.
+     *
+     * <p>This value should be used to calculate the total weight of a Context set. A value of
+     * 1 is considered default. The weight of a set is calculated by adding the weights of each
+     * contained Context. Higher value = higher/heavier priority.</p>
+     *
+     * @return the weight of the context pair.
+     */
+    public int getWeight() {
+        return this.weight;
+    }
+
     @Override
     public String getKey() {
-        return this.wrapped.getKey();
+        return this.key;
     }
 
     @Override
     public String getValue() {
-        return this.wrapped.getValue();
+        return this.value;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


bloodmc:
@zml2008 Please review as I need this functionality for GP in order to guarantee order of contexts being checked.

src/main/java/org/spongepowered/api/service/context/Context.java
```diff
@@ -44,18 +45,32 @@
     public static final String LOCAL_IP_KEY = "localip";
     public static final String LOCAL_PORT_KEY = "localport"; // This portkey is an old boot
 
-    private final Map.Entry<String, String> wrapped;
+    private final String key;
+    private final String value;
+    private final byte weight;
```
liach: Why byte?

lucko: changed to int

src/main/java/org/spongepowered/api/service/context/Context.java
```diff
@@ -44,18 +45,32 @@
...
      * Create a new context instance
      *
      * @param type Context type. Must not be null.
      * @param name Context name. Must not be null.
+     * @param weight Context weight. Must be greater than or equal to 1. Default is 1.
+     */
+    public Context(String type, String name, int weight) {
+        this.key = checkNotNull(type, "type");
+        this.value = checkNotNull(name, "name");
+        checkArgument(weight >= 1, "weight must be >= 1");
```
ST-DDT, line 62: Why is there such a limitation? What if I want to have my context less than default weight?

src/main/java/org/spongepowered/api/service/context/Context.java
```diff
@@ -96,21 +122,21 @@ public String setValue(String value) {
...
+        return this.key.hashCode() ^ this.value.hashCode();
     }
 
     @Override
     public String toString() {
-        return this.wrapped.toString();
+        return "Context(key=" + this.key + ", value=" + this.value + ", weight=" + this.weight + ")";
     }
-}
+}
```
ST-DDT: Missing newline at end of file.

src/main/java/org/spongepowered/api/service/context/Context.java
```diff
@@ -74,20 +89,31 @@ public String getType() {
...
-     * 
      */
     public String getName() {
         return getValue();
     }
 
+    /**
+     * Get the weight.
+     *
+     * This value should be used to calculate the total weight of a Context set. A value of
```
ST-DDT: The second paragraph should be encapsuled in `<p>`.

Also it should be documented what this weight actually does. Are higher values "better" or lower values? What happens if both have the same weight?

Last but not least, this method should be below the `getValue()` method.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/service/context/Context.java
lucko: You calculate the weight of a set by adding the weights of all contained contexts together. 

With the previous standard being to use the #size method, it makes sense that the default is 1. (it would produce the same result) It doesn't make sense to have a negative weight, or a weight of zero, since the only use case for this is making one of your contexts take priority over another.

src/main/java/org/spongepowered/api/service/context/Context.java
lucko: Will fix the first two bits. Any reason for ordering under getValue? Those methods are just for compatibility with the Map.Entry interface, or rather, just aliases for the other methods.

src/main/java/org/spongepowered/api/service/context/Context.java
ST-DDT: Okay I must have misunderstood this feature. All good now.

src/main/java/org/spongepowered/api/service/context/Context.java
ST-DDT: > Any reason for ordering under getValue?

Well its the (IMO) least important and you used that ordering for the fields as well.

EDIT: Oh I haven't seen the get type and name methods during my initial review. I guess its okay as it is.

src/main/java/org/spongepowered/api/service/context/Context.java
```diff
@@ -44,18 +45,32 @@
     public static final String LOCAL_IP_KEY = "localip";
     public static final String LOCAL_PORT_KEY = "localport"; // This portkey is an old boot
 
-    private final Map.Entry<String, String> wrapped;
+    private final String key;
+    private final String value;
```
ST-DDT, line 49: How about renaming the fields to type and name, as this is their actual purpose. The key and value stuff is just for Map.Entry compatibility, but contains less meaning.

zml2008:
The weight field should be split out into a WeightedContext subclass -- in some cases where Contexts are used, a weight doesn't make sense -- like in getAllPermissions/getAllOptions/getAllParents the permissions plugin has to create contexts itself and wont know weights -- and the original plugin providing the calculator for the contexts will not necessarily be present, so a method can't be added there.

The weights are only used when calculating, not storing permissions, so the distinction has to be made somehow -- and since there are limitations in using a List a WeightedContext is another way to go -- unless you have another idea.

src/main/java/org/spongepowered/api/service/context/Context.java
liach: @lucko It just can be any value. It does not affect.

lucko:
Following discussion on IRC, I think we all somewhat agree that this would be best resolved in API 6, meaning this PR is useless. 

Closing for that reason.

PR #1434, opened 2016-12-15
Tweak Statistics API for implementation

windy1:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1434) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1076)

This PR modifies the existing Statistics API to comply with my implementation.

#### `StatisticGroup` has been replaced with `StatisticType`
The old system didn't really make a lot of sense for the implementation and was rather arbitrary. `StatisticType` is for special statistics on blocks, items, and entities. All other statistics have the `BASIC` type.

----

#### Builders and API registration has been removed
Registering a statistic or achievement from the API makes about as much sense as doing so for an `ItemType` or `BlockType` so I decided to remove API registration altogether.

----

#### Removed `StatisticFormat`
Instead I just decided to expose the `NumberFormat` the `Statistic` uses.


src/main/java/org/spongepowered/api/statistic/Statistic.java
```diff
@@ -25,103 +25,30 @@
...
+     * Returns the objective {@link Criterion} for this statistic.
      *
-     * @return The statistic group this statistic belongs to
+     * @return Objective criterion
      */
-    StatisticGroup getGroup();
+    Optional<Criterion> getCriterion();
 
-    interface Builder extends StatisticBuilder<Statistic, Builder> {
+    NumberFormat getFormat();
```
Deamon5550, line 54: Javadocs?

src/main/java/org/spongepowered/api/statistic/Statistics.java
```diff
@@ -33,101 +33,109 @@
 
     // SORTFIELDS:ON
 
-    public static final Statistic ANIMALS_BRED = DummyObjectProvider.createFor(Statistic.class, "ANIMALS_BRED");
+    public static final Statistic OPEN_SHULKER_BOX = createFor(Statistic.class, "OPEN_SHULKER_BOX");
```
Deamon5550: Looks like you need to run the sort fields task.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -24,34 +24,19 @@
...
-    /**
-     * Creates a new {@link Builder} to build a {@link Achievement}.
-     *
-     * @return The new builder
-     */
-    static Builder builder() {
-        return Sponge.getRegistry().createBuilder(Builder.class);
-    }
-
+public interface Achievement extends Statistic {
```
Deamon5550, line 41: I originally decided against this as it for a number of reasons. Primarily that the actual statistics value is pretty useless for achievements and representing it as such is a bit of a hack. I'm guessing you made this change for ease of implementation?

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -75,89 +60,17 @@ static Builder builder() {
...
     /**
-     * If this achievement is backed by a statistic (eg. if
-     * {@link #getSourceStatistic()} is not absent) then this will return the
-     * target value which must be reached for this achievement to be awarded to
-     * a player.
+     * Returns true if this is a "special" achievement.
      *
-     * @return The target value, if available
-     */
-    Optional<Long> getStatisticTargetValue();
```
Deamon5550, line 92: Was this moved anywhere else? I don't see it.

Deamon5550 (changes requested):
I can live with not having the registration although I'm not sure why you choose to remove it?

liach:
@Deamon5550 When I commented on the commit several hours ago, the pull was not yet opened. Will comment here.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -59,6 +59,7 @@
 import org.spongepowered.api.entity.living.monster.Skeleton;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.projectile.arrow.Arrow;
+import org.spongepowered.api.statistic.Statistic;
```
liach, line 62: Import order

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -75,89 +60,17 @@ static Builder builder() {
...
 
     /**
-     * Gets the {@link Statistic} which this achievement is backed by if it is available.
+     * Returns the {@link ItemStack} that this achievement is represented by.
      *
-     * @return The source statistic, if available
+     * @return ItemStack representation
      */
-    Optional<Statistic> getSourceStatistic();
+    Optional<ItemStack> getItemStack();
```
liach: I suggest the use of `ItemStackSnapshot` here.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
windy1: Essentially yes. However, I did implement it in such a way that they are always separated by the API. `StatData` will not contain `Achievements` and `AchievementData` will not contain `Statistics`. I think this is fine since there is no actual value in a `Statistic` and all the other implemented methods still make sense for an `Achievement`.

windy1:
Just going to post what I said to @Deamon5550 in IRC for context.

I chose to remove API registration because it makes about as much sense as allowing API registration for say ItemType or BlockType. Statistics and Achievements require a client side mod to be a "full statistic" and while it would be possible to register stats only on the server, this would cause the server to start sending junk stat IDs to the client.

I do understand the possible use case of server-only stats but I came to the conclusion that it wasn't worth it since a plugin could easily simulate their own stats. Statistics are just an Object->Int map anyways.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
Deamon5550: Yeah that sound fine.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
windy1: As far as I could tell there is no relationship between achievements and "source" statistics? I could be wrong though. Also an achievements value is 0 for locked and >0 for unlocked although this is not exposed by the API.

src/main/java/org/spongepowered/api/statistic/Statistics.java
```diff
@@ -24,7 +24,7 @@
  */
 package org.spongepowered.api.statistic;
 
-import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+import static org.spongepowered.api.util.generator.dummy.DummyObjectProvider.createFor;
```
gabizou: Don't static import the method `createFor` please. Yes, it does add to line length, but it's clearer when reading where it's coming from.

src/main/java/org/spongepowered/api/statistic/Statistic.java
ST-DDT: How would this formatter be used for the time format?

Expected result: 5d 4h...

src/main/java/org/spongepowered/api/statistic/Statistics.java
ST-DDT: IMO its quite obvious where this method comes from (especially since its a CatalogType list)

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
ST-DDT: If its impl side connected i don't care, but having this connected from the API side does not make sense IMO.

>  StatData will not contain Achievements and AchievementData will not contain Statistics. 

This sentences alone is a reason to separate it API wise. (They don't have anything in common)

The format method and statistic type do not make sense for Achievements at all.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
```diff
@@ -75,89 +60,17 @@ static Builder builder() {
...
-        Builder targetValue(long value);
-
-        /**
-         * Builds and registers an instance of an {@link Achievement}.
-         *
-         * @return A new instance of a achievement
-         * @throws IllegalStateException If the achievement is not completed
-         */
-        Achievement buildAndRegister() throws IllegalStateException;
+    boolean isSpecial();
```
ST-DDT, line 75: What is a "special" achievement? Is this pertaining the color of the achievement?

src/main/java/org/spongepowered/api/statistic/Statistic.java
```diff
@@ -25,103 +25,30 @@
...
 
+    default StatisticType getType() {
+        return StatisticTypes.BASIC;
     }
 
-    /**
-     * Represents a builder to create new and custom instances of {@link Statistic}
-     * s.
-     */
-    interface StatisticBuilder<T extends Statistic, B extends StatisticBuilder<T, B>> extends ResettableBuilder<T, B> {
```
ST-DDT, line 78: Why did you remove the builders?

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/statistic/Statistic.java
windy1: Implement it yourself. The format really has nothing to do with the server so providing an implementation to "mock" the client is entirely arbitrary.

src/main/java/org/spongepowered/api/statistic/Statistic.java
windy1: I've already explained this in the OP and again in the comments.

src/main/java/org/spongepowered/api/CatalogTypes.java
```diff
@@ -253,7 +252,6 @@
     public static final Class<HorseStyle> HORSE_STYLE = HorseStyle.class;
 
     @Deprecated public static final Class<HorseVariant> HORSE_VARIANT = HorseVariant.class;
-
```
gabizou, line 256: Shouldn't be deleting the empty line between these.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
Deamon5550: Its displayed as a change of the border around them. example

src/main/java/org/spongepowered/api/statistic/BlockStatistic.java
```diff
@@ -25,35 +25,19 @@
...
-        /**
-         * Sets the {@link BlockType} of this {@link BlockStatistic}.
-         *
-         * @param block The block
-         * @return This builder, for chaining
-         */
-        Builder block(BlockType block);
-
+    default BlockType getBlockType() {
+        return getItemType().getBlock().get();
```
gabizou, line 40: I'm guessing this is always guaranteed? If not, in any case, I'd recommend doing an `orElseThrow` with an appropriate exception based on the `ItemType`, perhaps something to do with the implementation.

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
gabizou: Likewise, yes, I'd recommend using `ItemStackSnapshot` to avoid having to constantly copy the `ItemStack` defensively when returning it.

gabizou (changes requested):
Minor changes, good job so far :)

src/main/java/org/spongepowered/api/CatalogTypes.java
windy1: Huh weird, I think `sortClassFields` might have done this for some reason. Maybe because of the annotation?

src/main/java/org/spongepowered/api/statistic/BlockStatistic.java
windy1: Yes the implementing class `SpongeBlockStatistic` will always have a valid `BlockType`.

gabizou: Gotcha.

src/main/java/org/spongepowered/api/CatalogTypes.java
gabizou: Perhaps. Might just need to manually add it and re-run to test.

windy1: Can confirm the gradle task was causing this.

windy1:
@gabizou Added and tested your requested changes on API and Common PR

src/main/java/org/spongepowered/api/statistic/Statistic.java
ST-DDT: Then you can also drop this NumberFormat for the same reasons.
Especially since its not very usable for all instances (time + distance).

src/main/java/org/spongepowered/api/statistic/achievement/Achievement.java
ST-DDT: Isn't this a client only thing? Or is there something special for the server as well? If not then this should probably be dropped until we add client support. (to avoid API breakage)

ST-DDT (approved):

windy1:
Anything else before these get merged?

gabizou (approved):
Good job.

PR #1437, opened 2016-12-21
Clarify Inventory#query/contains behaviour + new versions for any amount

Faithcaio:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1437) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1095)

With Minecraft 1.11 it is no longer possible to create -1 sized ItemStacks.
Additionally the current query method never actually cared about stack sizes.

This claryfies the javadocs on the existing query/contains methods
and adds a new one to query for ItemStacks matching any size
and one to check for inventories containing any amount.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -271,13 +271,27 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
     boolean hasChildren();
 
     /**
-     * Checks for whether the given stack is contained in this Inventory. This
-     * is equivalent to calling <code>!inv.query(stack).hasChildren();</code>
+     * Checks whether the stacks quantity or more of given stack is
+     * contained in this Inventory. This is equivalent to calling
+     * <code>!inv.query(stack).hasChildren();</code> To check if an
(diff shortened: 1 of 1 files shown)


Cybermaxke:
This pr will resolve #1392

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -271,15 +271,28 @@
...
-     * Checks for whether the given stack is contained in this Inventory. This
-     * is equivalent to calling <code>!inv.query(stack).hasChildren();</code>
+     * Checks whether the stacks quantity or more of given stack is
+     * contained in this Inventory. This is equivalent to calling
+     * <code>!inv.query(stack).hasChildren();</code> To check if an
+     * inventory contains any amount use {@link #containsAny(ItemStack)}.
      *
      * @param stack The stack to check for
-     * @return True if the stack is present in this list
+     * @return True if the at least the stack stack is present in this inventory
```
ST-DDT: Something is wrong with this sentence,
Maybe something like this is intended.
`@return true if there are at least the given stack's amount of items present in this inventory.`

ST-DDT (changes requested):

PR #1438, opened 2016-12-22
Add some more command arguments

pie-flavor:
I figured the things in MoreArguments would be better used in the standard library.

[Test plugin as requested by](https://github.com/pie-flavor/SpongeTestPlugin/releases/tag/more-arguments%2F1.0.1) @kashike


src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,247 @@ public Text getUsage(CommandSource src) {
...
+
+    /**
+     * Expect an argument to be a valid {@link ConfigurationNode}.
+     *
+     * @param key The key to store under
+     * @param forceJson If true, requires JSON; if false, allows the more
+     *                  lenient HOCON
+     * @return the argument
+     */
+    public static CommandElement configNode(Text key, boolean forceJson) {
```
me4502: Configurate supports numerous configuration types, even YAML. Also I don't see the appeal of creating a ConfigurationNode. It'd make more sense to just have a generic JSON deserializing argument.

pie-flavor: The appeal is it's already there, with wide use already, and there's also the bit about object mapping. And if not a node, what would it return?

And the idea here is that "Mojangson" is unintentionally a subset of HOCON, and other commands already use strict JSON.

I wouldn't know how to actually make a JSON/Mojangson deserializing argument; I just know how to use the Configurate library. If it shouldn't use Configurate, I'd just remove it.

me4502: I'd probably ask @zml2008 what he thinks should work.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,247 @@ public Text getUsage(CommandSource src) {
...
+            } catch (MalformedURLException ex) {
+                throw new ArgumentParseException(Text.of("Invalid URL!"), ex, str, 0);
+            }
+            URI uri;
+            try {
+                uri = url.toURI();
+            } catch (URISyntaxException ex) {
+                throw new ArgumentParseException(Text.of("Invalid URL!"), ex, str, 0);
+            }
+            if (this.returnURI) {
```
ST-DDT: Why do you place the if here? Do you want to verify that it matches the URI format even if it is an URL?
I guess you could move the if before the uri part or wrap the toURI part in the if. Thats just my thought though.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,247 @@ public Text getUsage(CommandSource src) {
...
+                if (this.self) {
+                    if (source instanceof RemoteSource) {
+                        args.setState(state);
+                        return ((RemoteSource) source).getConnection().getAddress().getAddress();
+                    } else {
+                        throw args.createError(Text.of("Invalid IP address, and source was not a player!"));
+                    }
+                }
+                throw args.createError(Text.of("Invalid IP address!"));
+            }
```
ST-DDT: How about specifing a player name here to obtain the ip:
`/banip badGuy`

ryantheleach: UUID would be nice too.

pie-flavor: That sounds fun.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: Yeah, the goal here is to verify the structure either way but return what they're asking for. I wasn't about to pull in all of Apache Validators just for the UrlValidator.

windy1: From the JavaDocs:
> Note, any URL instance that complies with RFC 2396 can be converted to a URI. However, some URLs that are not strictly in compliance can not be converted to a URI.

I think it's best to maybe validate that the URL is compliant by calling `toURI` but then ultimately returning the `URL`. Returning one or the other is too ambiguous in my eyes here.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,277 @@ public Text getUsage(CommandSource src) {
...
+
+    /**
+     * Expect an argument to return an IP address, in the form of an
+     * {@link InetAddress}, or if nothing matches and the source is a
+     * {@link RemoteSource}, return the source's address.
+     *
+     * @param key The key to store under
+     * @return the argument
+     */
+    public static CommandElement ipOrSource(Text key) {
```
windy1, line 1824: I like this idea and I think it could be applied elsewhere as well (e.g. `entityOrTarget`).

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,277 @@ public Text getUsage(CommandSource src) {
...
+        @Override
+        protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            if (!args.hasNext() && this.self) {
+                if (source instanceof Player) {
+                    return ((Player) source).getConnection().getAddress().getAddress();
+                } else {
+                    throw args.createError(Text.of("No IP address was specified!"));
+                }
+            }
+            Object state = args.getState();
```
windy1, line 1849: Correct me if I'm wrong here but wouldn't this get the index of the argument you are parsing, therefore making your call to `args.setState` reset the index to before the argument you've already parsed? Also is dealing with the state even necessary for single argument `CommandElement`s? I was under the impression that it is used for multi-arg elements such as `remainingJoinedStrings`.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,277 @@ public Text getUsage(CommandSource src) {
...
+
+    }
+
+    /**
+     * Expect an argument to return a {@link BigDecimal}.
+     *
+     * @param key The key to store under
+     * @return the argument
+     */
+    public static CommandElement bigDecimal(Text key) {
```
windy1, line 1879: Hmm, I don't know about this one, can you describe a use case where a regular `double` is insufficient for user input? I understand the implications of float point numbers in calculation, but for user input?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1604,277 @@ public Text getUsage(CommandSource src) {
...
+        }
+    }
+
+    /**
+     * Expect an argument to return a {@link BigInteger}.
+     *
+     * @param key The key to store under
+     * @return the argument
+     */
+    public static CommandElement bigInteger(Text key) {
```
windy1, line 1907: This one is a bit more understandable I suppose, still a little questionable, IMO.

windy1 (changes requested):
Looks good overall, definitely justified in most cases IMO.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: Hmm? We already have `entityOrSource`.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: Actually, it's `bigInteger` that is somewhat unnecessary - the point of `bigDecimal` is that the Economy API is entirely structured around them, and I find myself constantly using casts and checks (not to mention the fact that an arbitrary economy plugin may use enormous numbers).

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: That is in fact the point; otherwise this argument would be swallowed without parsing if the "or source" part is executed.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
windy1: Right, but `BigDecimal` has a `double` value constructor and the point of `BigDecimal` is more about it's increased precision which I don't think is really applicable to humans in general and is more meant for lossless floating point calculations.

pie-flavor: Well, an arbitrary economy API might use incredibly large precise values. Humans are humans, but command blocks exist as well. And the point here is abstraction; why shouldn't you be able to request the one thing that half of your plugin is structured around?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: I suppose.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
windy1: Oh did not know that, I still think an `entityOrTarget` (by directional vector) would be cool.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
sibomots: Generics, man. Generics.   fooTypeX, fooTypeY, fooTypeZ is not generic.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
sibomots: the name of this function tells me nothing specific.  The name says "configNode". Does it config _a_ node? which node, a parameter.. nope the object's node, well it's static method of a "GenericArguments" class.

sibomots:
The name of this PR is unclear.  Is there a style guide floating around for how to name bugs/PR ?

I'll name my next PR "Stuff that works better."

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1585,4 +1628,259 @@ public Text getUsage(CommandSource src) {
...
+            super(key);
+            this.self = self;
+            this.possiblePlayer = new PlayerCommandElement(key, false);
+        }
+
+        @Nullable
+        @Override
+        protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            if (!args.hasNext() && this.self) {
+                if (source instanceof Player) {
```
phroa: According to the javadoc for `ipOrSource` you meant `RemoteSource` here and in the cast below.

phroa (changes requested):
...I'm not sure "start a review" was what I should have clicked.  I guess I'll request changes?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: That is a good point - it's not exactly clear that it's a 'config object'. Any input on new names?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: That's actually a very good idea, I'll look into that. That's one that would get used plenty (and by Reveries, no less :P) 

Edit: Sorta used the exact code

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: If only `GenericArguments.of(BigInteger.class)` was feasible.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
ryantheleach: There is a #target hard coded entity name for selecting entity targets I think. Running off memory due to xmas

pie-flavor: Not for entities, just for vectors/locations.

pie-flavor:
@windy1 Are you still requesting changes? Because it seems like it was resolved but the change request is still open.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1508,31 +1534,61 @@ public Text getUsage(CommandSource src) {
...
-        return new EntityCommandElement(key, false);
+        return new EntityCommandElement(key, false, false, null);
     }
 
     public static CommandElement entityOrSource(Text key) {
-        return new EntityCommandElement(key, true);
+        return new EntityCommandElement(key, true, false, null);
+    }
+
+    public static CommandElement entity(Text key, Class<? extends Entity> type) {
```
ST-DDT: Shouldn't this method be above `entityOrSource()`? In order to have similar methods are next to each other.
In addition to that the `entityOrSource(Text key, Class<? extends Entity> type)` variant is missing.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1579,9 +1645,395 @@ private static Entity tryReturnSource(CommandSource source, CommandArgs args) th
...
+
+    }
+
+    /**
+     * Expect an argument to be a {@link UUID}.
+     *
+     * @param key The key to store under
+     * @return the argument
+     */
+    public static CommandElement uuid(Text key) {
```
ST-DDT, line 1969: Maybe there should be a helper method, that allows some uuid suggestions (online players/entities etc).

````
public static CommandElement uuid(Text key, Collection<UUID> suggetedUUIDs);
````

Or is there a CommandElement extension that allows adding custom suggestions? <- Maybe we should add one.

**Review by ST-DDT — CHANGES_REQUESTED** [review]

**Comment by ST-DDT** [comment]
@pie-flavor The system will not recognize it as resolved until @windy1 reviews your PR again and presses the accept button.

**Comment by pie-flavor** [comment]
@ST-DDT I know, that's why I poked him.

File: src/main/java/org/spongepowered/api/command/args/GenericArguments.java (hunk 10 above)
Comment (pie-flavor) [reply src/main/java/org/spongepowered/api/command/args/GenericArguments.java:1969]: The point here is that it's a plain UUID, not related to anything in particular. `entity()` and `player()` methods already exist should they want it to fit a pattern.

File: src/main/java/org/spongepowered/api/command/args/GenericArguments.java (hunk 9 above)
Comment (pie-flavor) [reply src/main/java/org/spongepowered/api/command/args/GenericArguments.java]: I appended to the bottom; if this is an issue I'll move it.

Intentional; if `orSource` is used, the target type is a player.

Comment (ST-DDT) [reply src/main/java/org/spongepowered/api/command/args/GenericArguments.java]: > I appended to the bottom; if this is an issue I'll move it.

I just mentioned it as my own preference. Whether you should change it or not is up to you/others.

> Intentional; if `orSource` is used, the target type is a player.

Well I thought the class limitation would be `Entity <= X <= Player`.

This would allow things such as:
`/heal [Living/Player]`

But I guess there are other ways as well.

**Comment by pie-flavor** [comment]
@windy1 Still waiting on that clarification :P

File: src/main/java/org/spongepowered/api/command/args/GenericArguments.java [hunk 11]
```diff
@@ -1584,27 +1609,57 @@ public static CommandElement entity(Text key) {
      * @return The command element
      */
     public static CommandElement entityOrSource(Text key) {
-        return new EntityCommandElement(key, true);
+        return new EntityCommandElement(key, true, false, null);
+    }
+
+    public static CommandElement entity(Text key, Class<? extends Entity> type) {
```
parlough: Probably should document this and the two methods below(`entity(Text key, Class<? extends Entity> type)`, `entityOrTarget(Text key)`, and `entityOrTarget(Text key, Class<? extends Entity> type)` , explaining what they accomplish.

As a whole this PR seems good and I would love to see it merged.

kashike:
@pie-flavor status?

pie-flavor:
@kashike status is if you can't find anything wrong then it's ready, but still waiting for windy to approve changes

kashike:
@pie-flavor Could you provide a test plugin in the description?

pie-flavor:
@windy1 any updates?

pie-flavor:
Um.
Is that commit going to create problems? I ask because I still don't fully understand the merge conflict thing, and I used GitHub's tool this time.

Deamon5550:
Its usually better to rebase than merge when updating feature branches

pie-flavor:
um

approximately 0.01% of that was intentional.

pie-flavor:
Right, I give up. GG git.

PR #1159, opened 2016-03-26
Update AgeableData

hsyyid:
This PR makes https://github.com/SpongePowered/SpongeCommon/pull/562 possible. 

**Changes:**

* [x] - Added the key for `IS_ADULT`
* [x] - Removed redundant methods, which were discussed with @gabizou a while back.

PR #1440, opened 2016-12-23
Update AgeableData

Maxqia:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1092)
See #1159
Edited for 1.11 and added an adult() method for Ageable

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableAgeableData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableAgeableData.java
@@ -45,15 +45,6 @@ public interface ImmutableAgeableData extends ImmutableDataManipulator<Immutable
 
     /**
      * Gets the {@link ImmutableValue} for whether the {@link #age()} is
-     * considered to be a "baby". If the value is false, {@link #adult()}
-     * should be {@code true} and vice versa.
-     *
-     * @return The immutable value for the "baby" state
-     */
-    ImmutableValue<Boolean> baby();
-
-    /**
-     * Gets the {@link ImmutableValue} for whether the {@link #age()} is
      * considered to be an "adult". If the value is false, {@link #baby()}
      * should be {@code true} and vice versa.
      *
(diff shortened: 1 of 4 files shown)


me4502 (approved):

me4502:
Is this something we'd want to deprecate before removing?

Deamon5550:
@me4502 All things where it is convenient should be deprecated before removing, as I'd say this case is.

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/AgeableData.java
```diff
@@ -44,15 +44,6 @@
 
     /**
      * Gets the {@link Value} for whether the {@link #age()} is
-     * considered to be a "baby". If the value is false, {@link #adult()}
-     * should be {@code true} and vice versa.
-     *
-     * @return The immutable value for the "baby" state
-     */
-    Value<Boolean> baby();
```
gabizou: Since this can be deprecated and still supported, just have it deprecated.

src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableAgeableData.java
```diff
@@ -45,15 +45,6 @@
 
     /**
      * Gets the {@link ImmutableValue} for whether the {@link #age()} is
-     * considered to be a "baby". If the value is false, {@link #adult()}
-     * should be {@code true} and vice versa.
-     *
-     * @return The immutable value for the "baby" state
-     */
-    ImmutableValue<Boolean> baby();
```
gabizou, line 53: Since this can be deprecated and still supported, just have it deprecated.

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/AgeableData.java
gabizou, line 52: Since this can be deprecated and still supported, just have it deprecated.

gabizou (changes requested):
Deprecate methods that we can still support. Ease code removals between API versions.

Maxqia:
It'd still do nothing, because to implement it, I would have to add another key and this would just encourage bad coding practices.

gabizou:
> It'd still do nothing, because to implement it, I would have to add another key and this would just encourage bad coding practices.

I'm not asking.

PR #1449, opened 2017-01-02
Generalize recipes and implement them

Limeth:
*SpongeAPI*|[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1119)|[SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1188)

This is a WIP continuation of [kashike's PR](https://github.com/SpongePowered/SpongeAPI/pull/1098), with his consent.
It aims to generalize how recipes are registered and handled. Instead of checking the equality of the ingredients, it works by testing ingredient predicates. It also aims to let the user implement the recipe interfaces by themselves, making it, for example, possible to transfer NBT data from the ingredients to the result.
In this PR, `ItemStackSnapshot`s are used extensively. This is to prevent the user from modifying the original `ItemStack`. I could instead provide an `ItemStack#copy`, but that doesn't make it visually obvious the original item won't change.


Limeth:
The SpongeAPI part is now in the state of what I'd like to propose. Comments and reviews very welcome. Continuing by attempting to implement the SpongeCommon part.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,229 @@
...
+         * Sets an ingredient predicate based on the aisle pattern for the shaped recipe.
+         * This mimics the vanilla checking behavior.
+         *
+         * @param symbol The ingredient symbol
+         * @param ingredient The ingredient to set, or remove if null
+         * @return The builder
+         * @throws IllegalArgumentException If the aisle does not contain
+         *     the specified character symbol
+         */
+        Builder where(char symbol, @Nullable ItemStackSnapshot ingredient) throws IllegalArgumentException;
```
ST-DDT, line 112: * [x] I guess this can be default implemented as well.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,93 @@
...
+     *         when the recipe has been fulfilled (possibly empty),
+     *         or {@link Optional#empty()} if the given
+     *         {@link GridInventory} does not match this recipe's requirements.
+     */
+    Optional<List<ItemStack>> getRemainingItems(GridInventory grid);
+
+    /**
+     * @return {@code width * height} for shaped recipes, or the number of ingredients for shapeless recipes
+     */
+    int getRecipeSize();
```
ST-DDT: When would i/the impl need this method?

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
```diff
@@ -24,32 +24,32 @@
...
      */
-    void register(Recipe recipe);
+    void register(T recipe);
 
     /**
-     * Removes the given Recipe from registration in this registry.
+     * Removes the given CraftingRecipe from registration in this registry.
      *
-     * @param recipe The Recipe to unregister
+     * @param recipe The CraftingRecipe to unregister
```
ST-DDT: The javadocs mention crafting recipes but the generic type allows plain recipes as well.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,93 @@
...
+     * as it customizes the result further depending on the specified
+     * input {@param grid}.
+     * It is advised to use the output of {@link #getExemplaryResult()},
+     * modify it accordingly, and {@code return} it.
+     *
+     * @param grid The crafting input, typically 3x3 or 2x2
+     * @return An {@link ItemStack} or {@link Optional#empty()} if the given
+     *         {@link GridInventory} does not match this recipe's requirements.
+     */
+    Optional<ItemStack> getResult(GridInventory grid);
```
ST-DDT: Here you get the results but how do you actually consume the items? Is it required to reduce the number by one? How about Blood Magic's magicians orbs that act as a mere catalyst and thus is not consumed?  Or recipes where a tool is used and the tool only takes some damage?

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,141 @@
...
+     * return The {@link ItemStackSnapshot} as explained above
+     */
+    ItemStackSnapshot getExemplaryIngredient();
+
+    /**
+     * @param ingredient The ingredient to check against
+     * @return Whether this {@param ingredient} can be used to craft the result
+     */
+    default boolean isValid(ItemStackSnapshot ingredient) {
+        // ItemStackSnapshot#NONE gets replaced at runtime
```
ST-DDT: Missing null check.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Gets the ingredient predicates for this ShapelessCraftingRecipe.
+     *
+     * @return An unmodifiable list of the ingredient predicates
+     */
+    List<Predicate<ItemStackSnapshot>> getIngredientPredicates();
+
+    @Override
+    default boolean isValid(GridInventory grid, World world) {
+        List<Predicate<ItemStackSnapshot>> predicates = Lists.newLinkedList(getIngredientPredicates());
```
ST-DDT: Missing null checks.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,229 @@
...
+    default Optional<Predicate<ItemStackSnapshot>> getIngredientPredicate(char symbol) {
+        return Optional.ofNullable(getIngredientPredicates().get(symbol));
+    }
+
+    /**
+     * @param x The x coordinate counted from the left side
+     * @param y The y coordinate counted from the top
+     * @return The ingredient predicate at this position defined by the aisle
+     */
+    default Optional<Predicate<ItemStackSnapshot>> getIngredientPredicate(int x, int y) {
```
Deamon5550: Why are these methods default implemented?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,93 @@
...
+ * a 2x2 or 3x3 crafting matrix.</p>
+ *
+ * <p>The requirements of a CraftingRecipe can be general, they just have to
+ * eventually return a boolean given an itemgrid.</p>
+ *
+ * <p>You can implement it manually to suit your creative needs,
+ * or you can simply use the {@link ShapelessCraftingRecipe.Builder} or {@link ShapedCraftingRecipe.Builder}.</p>
+ */
+public interface CraftingRecipe extends Recipe {
+    /**
```
kashike: Missing new line after body start.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * A ShapedCraftingRecipe is a CraftingRecipe that has shape and fits into a grid.
+ */
+public interface ShapedCraftingRecipe extends CraftingRecipe {
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * @return An unmodifiable copy of the original aisle
```
kashike: Needs full javadocs.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,229 @@
...
+        return getIngredientPredicate(symbol);
+    }
+
+    @Override
+    default boolean isValid(GridInventory grid, World world) {
+        int gapWidth = grid.getColumns() - getWidth();
+        int gapHeight = grid.getRows() - getHeight();
+
+        // Shift the aisle along the grid wherever possible
+        for(int offsetX = 0; offsetX <= gapWidth; offsetX++) {
```
kashike: `for (` (space)

kashike (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Limeth: It cannot, the default item matching behavior in vanilla Minecraft is defined in NMS code and I needed to mimic it. Can't do that with SpongeAPI. Maybe I should create a method in SpongeCommon that mimics the behavior and expose it via SpongeAPI?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: The removal of items from the crafting grid is handled by the CraftingManager in NMS. To preserve items in the crafting grid, provide them via `#getRemainingItems(GridInventory)`.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: The implementation on the NMS side assigns priority to recipes depending on this value. You would use this to change that priority.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
Limeth: The argument is non-null by default, as defined in the `package-info.java` file in the package.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Limeth: Do I have my checkstyle plugin set up incorrectly? I am not getting the warnings here.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Limeth: So people who want to implement this interface themselves don't have to implement this method.

Deamon5550: Is that not what builders are for? if you're cluttering up the API to make it a little easier for the tiny percent of people why have to implement this in a plugin then you can remove all of these default implementations. 

The only things that should be default implemented is when its a trivial helper method.

Limeth: I'd find it incredibly useful to not have to write these myself, what is the issue exactly? All ShapedCraftingRecipes should behave like this, this is why they are default-implemented. And in corner cases, it also makes it possible to override them. It sounds like a win-win scenario to me. Note that all CraftingRecipes have an `isValid` method.

gabizou: I'd much rather have the least amount of actual *implementation* in the *API*. That being said, this is implementation, not API.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistry.java
Limeth: Thanks, didn't catch this one after auto-refactoring it.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
Limeth: Ah, turns out `@Nonnull` is only used for static analysis.

Maxqia:
-_- https://github.com/SpongePowered/SpongeAPI/pull/1419

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
ST-DDT: Please document this behaviour in the javadocs.

Limeth:
Began implementing the [SpongeCommon part](https://github.com/SpongePowered/SpongeCommon/pull/1119).

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -24,52 +24,14 @@
...
-    /**
-     * Checks if the given {@link GridInventory} fits the required constraints
-     * to craft this Recipe.
-     *
-     * @param grid The ItemGrid to check for validity
-     * @return True if the given input matches this recipe's requirements
-     */
-    boolean isValid(GridInventory grid);
-
     /**
```
Deamon5550: Missing javadoc

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,100 @@
...
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.type.GridInventory;
+import org.spongepowered.api.item.recipe.Recipe;
+import org.spongepowered.api.world.World;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * <p>A CraftingRecipe represents some craftable recipe in the game.</p>
```
Deamon5550: Don't need paragraph tags for the first paragraph.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,100 @@
...
+ * as a function from a crafting matrix to a list of {@link ItemStack}
+ * (the crafting result), therefore making it an immutable interface.</p>
+ *
+ * <p>The passed in ItemGrid is usually a crafting inventory, e.g.
+ * a 2x2 or 3x3 crafting matrix.</p>
+ *
+ * <p>The requirements of a CraftingRecipe can be general, they just have to
+ * eventually return a boolean given an itemgrid.</p>
+ *
+ * <p>You can implement it manually to suit your creative needs,
```
Deamon5550: Skip this bit, its confusing and its stating the obvious.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param world The world this recipe would be used in
+     * @return True if the given input matches this recipe's requirements
+     */
+    boolean isValid(GridInventory grid, World world);
+
+    /**
+     * This method is preferred over the {@link #getExemplaryResult()} method,
+     * as it customizes the result further depending on the specified
+     * input {@param grid}.
+     * It is advised to use the output of {@link #getExemplaryResult()},
```
Deamon5550: Either join this to the previous line or make it a new paragraph.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,167 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * This returns an unmodifiable copy of the original aisle.
+     *
+     * @return An unmodifiable copy of the original aisle
+     */
+    List<String> getAisle();
```
Deamon5550: The aisle stuff is great for the builder but it doesn't have a place in the actual recipe. You already have width and height information and a method for getting the predicate for a position, keeping the aisle stuff around is useless as translating the characters is way more work than just using the methods that take an x, y position.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,167 @@
...
+    Map<Character, Predicate<ItemStackSnapshot>> getIngredientPredicates();
+
+    /**
+     * @param symbol The symbol paired with the ingredient predicate in the aisle
+     * @return The paired ingredient predicate
+     */
+    Optional<Predicate<ItemStackSnapshot>> getIngredientPredicate(char symbol);
+
+    /**
+     * @param x The x coordinate counted from the left side
```
Deamon5550: Missing javadoc

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Gets the ingredient predicates for this ShapelessCraftingRecipe.
+     *
+     * @return An unmodifiable list of the ingredient predicates
+     */
+    List<Predicate<ItemStackSnapshot>> getIngredientPredicates();
+
+    @Override
+    default boolean isValid(GridInventory grid, World world) {
+        Preconditions.checkNotNull(grid, "The grid must not be null");
```
Deamon5550: Remove this implementation.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+     * return The {@link ItemStackSnapshot} as explained above
+     */
+    ItemStackSnapshot getExemplaryIngredient();
+
+    /**
+     * @param ingredient The ingredient to check against
+     * @return Whether this {@param ingredient} can be used to craft the result
+     */
+    default boolean isValid(ItemStackSnapshot ingredient) {
+        Preconditions.checkNotNull(ingredient, "The ingredient must not be null");
```
Deamon5550: Remove this implementation

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+     * as it customizes the result further depending on the specified
+     * {@param ingredient} {@link ItemStackSnapshot}.
+     * It is advised to use the output of {@link #getExemplaryResult()},
+     * modify it accordingly, and {@code return} it.
+     *
+     * @param ingredient The {@link ItemStackSnapshot} currently being smelted
+     * @return The result of smelting the {@param ingredient} {@link ItemStack}, if the {@param ingredient} is valid
+     */
+    default Optional<ItemStack> getResult(ItemStackSnapshot ingredient) {
+        Preconditions.checkNotNull(ingredient, "The ingredient must not be null");
```
Deamon5550: Remove this implementation

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.item.recipe.Recipe;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.Optional;
+
+/**
+ * A general interface for furnace recipes.
```
Deamon5550: Whatever you are wrapping the javadocs in this whole file to it isn't 80 characters.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * An exemplary {@link ItemStackSnapshot},
+     * which will always make {@link #isValid(ItemStackSnapshot)} return {@code true}.
+     *
+     * return The {@link ItemStackSnapshot} as explained above
+     */
+    ItemStackSnapshot getExemplaryIngredient();
+
+    /**
```
Deamon5550: Missing javadoc

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+        ItemStack generalIngredient = getExemplaryIngredient().createStack();
+        ItemStack ingredientStack = ingredient.createStack();
+
+        generalIngredient.setQuantity(ingredientStack.getQuantity());
+
+        return generalIngredient.equalTo(ingredientStack);
+    }
+
+    /**
+     * This method is preferred over the {@link #getExemplaryResult()} method,
```
Deamon5550: Shouldn't say preferred, this is the method to use.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,146 @@
...
+        Preconditions.checkNotNull(ingredient, "The ingredient must not be null");
+
+        if(!isValid(ingredient))
+            return Optional.empty();
+
+        return Optional.of(getExemplaryResult().createStack());
+    }
+
+    /**
+     * @param ingredient The ingredient being smelted
```
Deamon5550: Missing javadoc

Deamon5550 (changes requested):
Still some implementation in here, a few missing or misworded javadocs, and you need to go through with your IDE and fix the line wrapping of your javadocs its all over the place.

Limeth:
@Deamon5550 Thanks for the review. Should this be highlighted by the checkstyle plugin? I am not getting any warnings like those you named.

Deamon5550:
The checkstyle and formatter do their best to help you but common sense is always the best.

Limeth:
@Deamon5550 Fixed.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,132 @@
...
+     * @return The amount of experience released after completing this recipe,
+     *         if the {@param ingredient} is valid
+     */
+    Optional<Double> getExperience(ItemStackSnapshot ingredient);
+
+    /**
+     * Builds a simple furnace recipe
+     */
+    interface Builder extends ResettableBuilder<SmeltingRecipe, Builder> {
```
ST-DDT, line 89: * [x] Missing javadocs for all methods inside this builder.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,132 @@
...
+        default Builder ingredient(ItemStack ingredient) {
+            return ingredient(ingredient.createSnapshot());
+        }
+
+        /**
+         * @param experience The amount of experience released when this recipe
+         *                   is completed
+         * @return This builder, for chaining
+         */
+        Builder experience(double experience);
```
ST-DDT: Maybe add a `@Nullable` variant to reset the value, if you reuse the builder.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,100 @@
...
+     * inventory.
+     *
+     * @param grid The crafting input, typically 3x3 or 2x2
+     * @param world The world this recipe would be used in
+     * @return The list of items to be added to the inventory of the player
+     *         when the recipe has been fulfilled (possibly empty),
+     *         or {@link Optional#empty()} if the given
+     *         {@link GridInventory} does not match this recipe's requirements.
+     */
+    Optional<List<ItemStack>> getRemainingItems(GridInventory grid, World world);
```
ST-DDT: I quite unsure about this one. 

Wouldn't be a `Optional<CraftingResult>` better here?
Because currently you have to check the validity of the recipe twice (one for the result stack and one time for the remaining stacks).

If you use a `CraftingResult` which would store both of them at once you could skip one validity execution.

ST-DDT (changes requested):
Nice API design. I guess this will become quite handy.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: Thanks for the suggestion. Originally, the methods used to have different parameters, I didn't notice they could be merged together after changing them, thanks for pointing this out!

Limeth:
`CraftingRecipe#getResult(GridInventory grid, World world)` and `getRemainingItems(GridInventory grid, World world)` cannot be combined, because there are separate methods calling them in `net.minecraft.item.crafting.CraftingManager`.
I had to make a choice whether to make these two API methods return `Optional<T>`, where `Optional#empty()` would be returned when the recipe is not fulfilled, or just `T` without checking the recipe requirements. The downside of returning `T` directly is, that the API relies on that the user has checked the `isValid` method; on the other hand, if they decide to call both API methods, they only have to check the validity once.
~~Minecraft actually checks the recipe validity for both methods, so I decided to go with the `Optional<T>` return type.~~

Edit: Since users would have to implement `IRecipe`, the methods must basically be the same.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipeRegistry.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.recipe.smelting;
+
+public interface SmeltingRecipeRegistry {
```
Cybermaxke: You forgot to extend the `RecipeRegistry`.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+    Optional<ItemStack> getResult(ItemStackSnapshot ingredient);
+
+    /**
+     * Returns the amount of experience released after completing this recipe.
+     *
+     * @param ingredient The ingredient being smelted
+     * @return The amount of experience released after completing this recipe,
+     *         if the {@param ingredient} is valid
+     */
+    Optional<Double> getExperience(ItemStackSnapshot ingredient);
```
liach: Use `java.util.OptionalDouble`

Limeth: https://www.youtube.com/watch?v=PEZWYXPvmS8

kashike: @Limeth http://stackoverflow.com/a/33266445

Limeth: But there are no `map`/`flatMap` functions for chaining in `OptionalDouble`. Is performance really at stake here?

liach: Why do you have 2 wrappers if you can just have one @Limeth?

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
liach: Don't accept `null` as ingredient. Ask users to use `ItemStackSnapshot.NONE`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        /**
+         * Sets the resultant {@link ItemStackSnapshot} for when this shaped
+         * recipe is correctly crafted.
+         *
+         * @param result The resultant snapshot
+         * @return The builder
+         */
+        Builder result(@Nullable ItemStackSnapshot result);
```
liach: Same as above.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        /**
+         * Builds a {@link ShapedCraftingRecipe} from this builder.
+         *
+         * @return A new {@link ShapedCraftingRecipe}
+         */
+        ShapedCraftingRecipe build();
+    }
+}
```
liach: formatting

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Creates a new {@link Builder} to build a {@link ShapelessCraftingRecipe}.
+     *
+     * @return The new builder
+     */
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+
```
liach: Extra empty lines

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
gabizou: @Limeth What part of item matching is not possible in SpongeAPI?

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
Limeth: Exactly for the reason I stated above.

liach: You don't really need to map for doubles.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Limeth: @gabizou [This part](https://github.com/SpongePowered/SpongeCommon/pull/1119/files#diff-2ca3fc329aa6380a03d7f5253b5db9d9R77), as far as I'm aware.

me4502:
This is more a question of curiosity, but does this support information such as Lore, Names, Enchantments, etc, on recipe items?

Limeth:
@me4502 Yes, you can implement that ingredient predicate yourself and only return `true` if correct lore is present. I'm currently stuck implementing the SpongeCommon part of this PR, so if you anyone would like to help, that would be awesome.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+        /**
+         * Changes the experience and returns this builder. It is the
+         * required amount of experience the user must possess in order to be
+         * able to fulfill the recipe.
+         *
+         * @param experience The amount of experience released when this recipe
+         *                   is completed
+         * @return This builder, for chaining
+         */
+        Builder experience(@Nullable Double experience);
```
liach: What does `null` stand for here? Use default?

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+
+        /**
+         * Changes the result and returns this builder. The result is the
+         * {@link ItemStack} created when the recipe is fulfilled.
+         *
+         * @param result The output of this recipe
+         * @return This builder, for chaining
+         */
+        default Builder result(@Nullable ItemStack result) {
+            return result(result != null ? result.createSnapshot() : null);
```
liach: Seems like you missed this one by not using `ItemStackSnapshot.NONE`.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -24,52 +24,17 @@
...
-    /**
-     * Returns the results for running this Recipe over an {@link GridInventory}
-     *
-     * @param grid An ItemGrid as input
-     * @return A list of ItemStacks or {@link Optional#empty()} if the given
-     *          ItemGrid does not match this recipe's requirements.
-     */
-    Optional<List<ItemStack>> getResults(GridInventory grid);
-
+    ItemStackSnapshot getExemplaryResult();
```
ST-DDT: Add an empty line after the last member in a class/interface.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+
+    interface Builder extends ResettableBuilder<ShapedCraftingRecipe, Builder> {
+
+        /**
+         * Sets the aisle pattern for the shaped recipe.
+         *
+         * @param aisle A string array of ingredients
+         * @return The builder
+         */
+        Builder aisle(@Nullable String... aisle);
```
ST-DDT, line 86: * [x] This method should mention what happens to the previously registered ingredient `Predicate`s. (Especially those, that are no loner needed for this aisle.)

Ref: `where(char, Predicate)` <- `@throws IllegalArgumentException If the aisle does not contain the specified character symbol`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+         * Sets an ingredient predicate based on the aisle pattern for the
+         * shaped recipe. This mimics the vanilla checking behavior.
+         *
+         * @param symbol The ingredient symbol
+         * @param ingredient The ingredient to set, or remove if null
+         * @return The builder
+         * @throws IllegalArgumentException If the aisle does not contain
+         *     the specified character symbol
+         */
+        Builder where(char symbol, ItemStackSnapshot ingredient) throws IllegalArgumentException;
```
ST-DDT: I recommend adding `@Nullable` here to keep the remove if null part. This should also document what happens if the `ItemStackSnapshot.NONE` is set (IMO its the same as null => remove ingredient )

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+        /**
+         * Sets the resultant {@link ItemStack} for when this shaped recipe
+         * is correctly crafted.
+         *
+         * @param result The resultant stack
+         * @return The builder
+         */
+        @SuppressWarnings("ConstantConditions")
+        default Builder result(@Nullable ItemStack result) {
+            return result(result != null ? result.createSnapshot() : ItemStackSnapshot.NONE);
```
ST-DDT: This method should document the implications of this `ItemStackSnapshot.NONE` here.
IMO an empty stack is not a valid result.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+         *
+         * @param symbol The ingredient symbol
+         * @param ingredient The predicate to set, or remove if null
+         * @return The builder
+         * @throws IllegalArgumentException If the aisle does not contain
+         *     the specified character symbol
+         */
+        @SuppressWarnings("ConstantConditions")
+        default Builder where(char symbol, @Nullable ItemStack ingredient) throws IllegalArgumentException {
+            return where(symbol, ingredient != null ? ingredient.createSnapshot() : ItemStackSnapshot.NONE);
```
ST-DDT: I'm not quite sure about the `NONE` part. because its effectifly an invalid ingredient.
Otherwise you have to document the implications of using this. Especially since the implications are slightly different than the other methods.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
```diff
@@ -0,0 +1,150 @@
...
+        default Builder result(@Nullable ItemStack result) {
+            return result(result != null ? result.createSnapshot() : ItemStackSnapshot.NONE);
+        }
+
+        /**
+         * Builds a {@link ShapedCraftingRecipe} from this builder.
+         *
+         * @return A new {@link ShapedCraftingRecipe}
+         */
+        ShapedCraftingRecipe build();
```
ST-DDT: Maybe add a `throws IllegalStateException` for cases where the aisle is not set or the predicates are not set.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,123 @@
...
+    interface Builder extends ResettableBuilder<ShapelessCraftingRecipe, Builder> {
+
+        /**
+         * Adds an ingredient predicate to the requirements of this
+         * {@link ShapelessCraftingRecipe}.
+         *
+         * @param ingredient The ingredient predicate
+         * @return This builder, for chaining
+         */
+        Builder addIngredientPredicate(@Nullable Predicate<ItemStackSnapshot> ingredient);
```
ST-DDT: The effect of passing null should be documented.
IMO passing null here should not be allowed, otherwise the behaviour "remove if null" from the other builder does not match here.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,123 @@
...
+         * Adds an ingredient predicate to the requirements of this
+         * {@link ShapelessCraftingRecipe}. This mimics the vanilla equality
+         * checking behavior.
+         *
+         * @param ingredient The ingredient
+         * @return This builder, for chaining
+         */
+        @SuppressWarnings("ConstantConditions")
+        default Builder addIngredientPredicate(@Nullable ItemStack ingredient) {
+            return addIngredientPredicate(ingredient != null ? ingredient.createSnapshot() : ItemStackSnapshot.NONE);
```
ST-DDT: I'm not quite sure about the NONE part. because its effectifly an invalid ingredient.
Otherwise you have to document the implications of using this. Especially since the implications are slightly different than the other methods.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,123 @@
...
+        /**
+         * Changes the result and returns this builder. The result is the
+         * {@link ItemStack} created when the recipe is fulfilled.
+         *
+         * @param result The result
+         * @return This builder, for chaining
+         */
+        @SuppressWarnings("ConstantConditions")
+        default Builder result(@Nullable ItemStack result) {
+            return result(result != null ? result.createSnapshot() : ItemStackSnapshot.NONE);
```
ST-DDT: This method should document the implications of this ItemStackSnapshot.NONE here.
IMO an empty stack is not a valid result.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        /**
+         * Builds a new {@link ShapelessCraftingRecipe} from this builder.
+         *
+         * @return A new {@link ShapelessCraftingRecipe}
+         */
+        ShapelessCraftingRecipe build();
```
ST-DDT: Add `throws IllegalStateException`

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
+ * A general interface for furnace recipes. You can implement it manually to
+ * suit your creative needs, or you can simply use the
+ * {@link SmeltingRecipe.Builder}.
+ */
+public interface SmeltingRecipe extends Recipe {
+    /**
```
ST-DDT: Missing newline

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+     * @return Whether this {@param ingredient} can be used to craft the result
+     */
+    boolean isValid(ItemStackSnapshot ingredient);
+
+    /**
+     * This method should be used instead of the {@link #getExemplaryResult()}
+     * method, as it customizes the result further depending on the specified
+     * {@param ingredient} {@link ItemStackSnapshot}. It is advised to use
+     * the output of {@link #getExemplaryResult()}, modify it accordingly,
+     * and {@code return} it.
```
ST-DDT: I'm not sure whether the main javadoc part should be about the implementation. I would recommend explaining when it is used.

Out of thin air example:
````
This method is called at the beginning of the smelting process to check whether there is space 
for the result and at the end of the smelting to calculate to actually add it to the results.

<p>Note: It is advised that the result from getExemplaryResult and this method are strictly 
related to each other. 
/ ... the results from this method are a derived result from getExemplaryResult </p>
````

File: src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java [hunk 46]
```diff
@@ -0,0 +1,153 @@
...
+     */
+    interface Builder extends ResettableBuilder<SmeltingRecipe, Builder> {
+        /**
+         * Changes the result and returns this builder. The result is the
+         * {@link ItemStack} created when the recipe is fulfilled.
+         *
+         * @param result The output of this recipe
+         * @return This builder, for chaining
+         */
+        Builder result(@Nullable ItemStackSnapshot result);
```
ST-DDT: What are the implications of setting null here?
IMO setting it to null is useless since you have to set it to something none null anyway.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+
+        /**
+         * Changes the ingredient and returns this builder. The ingredient is
+         * the {@link ItemStack} required in order for the recipe to be
+         * fulfilled.
+         *
+         * @param ingredient The required ingredient
+         * @return This builder, for chaining
+         */
+        Builder ingredient(@Nullable ItemStackSnapshot ingredient);
```
ST-DDT: Document null behaviour. See above.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+
+        /**
+         * Changes the ingredient and returns this builder. The ingredient is
+         * the {@link ItemStack} required in order for the recipe to be
+         * fulfilled.
+         *
+         * @param ingredient The required ingredient
+         * @return This builder, for chaining
+         */
+        default Builder ingredient(@Nullable ItemStack ingredient) {
```
ST-DDT: Document null behaviour. See above.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+        Builder experience(@Nullable Double experience);
+
+        /**
+         * Builds the recipe and returns it.
+         *
+         * @return The built recipe
+         */
+        SmeltingRecipe build();
+    }
+}
```
ST-DDT: Missing newline + one a line above

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
ST-DDT: liach is referring to `.map(Function<T,R>)`.

~~Well if the absent result result is equivalent to "does not grant experience" then it could just return primitive `double` zero~~

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,153 @@
...
+     * @return The amount of experience released after completing this recipe,
+     *         if the {@param ingredient} is valid
+     */
+    Optional<Double> getExperience(ItemStackSnapshot ingredient);
+
+    /**
+     * Builds a simple furnace recipe
+     */
+    interface Builder extends ResettableBuilder<SmeltingRecipe, Builder> {
+        /**
```
ST-DDT: Missing newline

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+     * This method should only be called if
+     * {@link #isValid(GridInventory, World)} returns {@code true}.
+     *
+     * <p>This method is preferred over the
+     * {@link CraftingRecipe#getExemplaryResult()} method, as it customizes
+     * the result further depending on the context.</p>
+     *
+     * <p>It is advised to use the output of
+     * {@link CraftingRecipe#getExemplaryResult()}, modify it accordingly,
+     * and {@code return} it.</p>
```
ST-DDT: There should be a hint that this comment is only for implementing classes.

````
Implementing classes are advised...optionally cache it and {@code return} it.
````

File: src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java [hunk 52]
```diff
@@ -0,0 +1,167 @@
...
+
+        /**
+         * Sets the resultant {@link ItemStack} for when this shaped recipe
+         * is correctly crafted.
+         *
+         * @param result The resultant stack
+         * @return The builder
+         */
+        default Builder result(ItemStack result) {
+            return result(result.createSnapshot());
```
ST-DDT: There should be a null check here `checkNotNull(result, "result")`

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapelessCraftingRecipe.java
```diff
@@ -0,0 +1,132 @@
...
+
+        /**
+         * Changes the result and returns this builder. The result is the
+         * {@link ItemStack} created when the recipe is fulfilled.
+         *
+         * @param result The result
+         * @return This builder, for chaining
+         */
+        default Builder result(ItemStack result) {
+            return result(result.createSnapshot());
```
ST-DDT: Missing null check 
`return result(checkNotNull(result, "result").createSnapshot());`

ST-DDT (changes requested):
Good job. Almost done.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+    List<ItemStackSnapshot> getRemainingItems(GridInventory grid);
+
+    /**
+     * The recipe size is used to determine the priority of this recipe compared
+     * to other recipes.
+     *
+     * @return {@code width * height} for shaped recipes, or the number of
+     *         ingredients for shapeless recipes
+     */
+    int getSize();
```
Maxqia, line 119: Invalid Description, It wouldn't do that if you added it after. Honestly this function isn't really necessary at all.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Checks if the given {@link GridInventory} fits the required constraints
+     * to craft this {@link CraftingRecipe}.
+     *
+     * @param grid The ItemGrid to check for validity
+     * @param world The world this recipe would be used in
+     * @return True if the given input matches this recipe's requirements
+     */
+    boolean isValid(GridInventory grid, World world);
```
Maxqia: I'd say you should just combine this with getResult and default implement it to return false if it returns ItemTypes.AIR or null...

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
```diff
@@ -0,0 +1,181 @@
...
+    ItemStackSnapshot getExemplaryIngredient();
+
+    /**
+     * Checks if the given {@link ItemStackSnapshot} fits the required
+     * constraints to craft this {@link SmeltingRecipe}.
+     *
+     * @param ingredient The ingredient to check against
+     * @return Whether this {@param ingredient} can be used to craft the result
+     */
+    boolean isValid(ItemStackSnapshot ingredient);
```
Maxqia, line 67: Duplicate Function? and see above...

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,52 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.recipe.smelting;
+
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+
+public interface SmeltingResult {
```
Maxqia: Why is this a separate interface?

Maxqia (commented):
Ehhhhhh

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: In order to support listing modded recipes, I had to make `CraftingRecipe` symmetric to `IRecipe`. Making this return the `ItemStackSnapshot` would also unnecessarily create instances in cases they aren't needed.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: See above.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipe.java
Limeth: Not a duplicate. Extends `Recipe`, not `CraftingRecipe`.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
Limeth: In this case, we can afford to do the optimization you suggested for `CraftingRecipe#isValid(GridInventory, World)`, because there is no special interface for smelting recipes in Minecraft code. This way, users don't get access to the resulting `ItemStackSnapshot` and the amount of experience released, unless the `SmeltingRecipe#isValid(ItemStackSnapshot)` returns `true`.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+ * <p>The requirements of a CraftingRecipe can be general, they just have to
+ * eventually return a boolean given an itemgrid.</p>
+ */
+public interface CraftingRecipe extends Recipe {
+
+    /**
+     * Checks if the given {@link GridInventory} fits the required constraints
+     * to craft this {@link CraftingRecipe}.
+     *
+     * @param grid The ItemGrid to check for validity
```
simon816: `ItemGrid` was an old name, should now be `GridInventory`

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+     * the result further depending on the context.</p>
+     *
+     * <p>It is advised to use the output of
+     * {@link CraftingRecipe#getExemplaryResult()}, modify it accordingly,
+     * and {@code return} it.</p>
+     *
+     * @param grid The crafting input, typically 3x3 or 2x2
+     * @return An {@link ItemStackSnapshot}
+     */
+    ItemStackSnapshot getResult(GridInventory grid);
```
simon816, line 77: Recipes can output multiple items
See this comment: https://github.com/SpongePowered/SpongeAPI/pull/242#issuecomment-64930768

Also, it's not defined what would happen if `isValid` returns `false`, does it throw an exception? Maybe it should return  an `Optional`?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * This method should only be called if
+     * {@link #isValid(GridInventory, World)} returns {@code true}.
```
simon816, line 81: Again, need to define what happens if `isValid` returns `false`

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,52 @@
...
+     * {@link SmeltingRecipe#getExemplaryResult()} method, as it customizes the
+     * result further depending on the specified {@param ingredient}
+     * {@link ItemStackSnapshot}. It is advised to use the output of
+     * {@link SmeltingRecipe#getExemplaryResult()}, modify it accordingly, and
+     * {@code return} it.
+     *
+     * @return The result of fulfilling the requirements of a
+     *         {@link SmeltingRecipe}
+     */
+    ItemStackSnapshot getResult();
```
simon816: I think this should be a list of items. Modded recipes may return more than one item (e.g. a furnace that can smelt double at once)

simon816 (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
simon816: I believe the order defined in the implementation is so that the crafting manager can prefer large recipes (3x3) before considering smaller 2x2 ones.
This is somewhat implementation detail that need not be exposed.
For shaped recipes the implementation just needs to do `getWidth() * getHeight()` and for shapeless do `getIngredientPredicates().size()`

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: Crafting recipes cannot. That comment talks about [pulverizer recipes](https://github.com/CoFH/ThermalExpansion/blob/master/src/main/java/cofh/thermalexpansion/api/crafting/recipes/IPulverizerRecipe.java).

Limeth: What happens is undefined behavior, no way around that.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Limeth: See above.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
Limeth: Yes, but only in "modern" GUIs. Those recipes are not supported for standard vanilla GUIs as there is only one output slot.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipe.java
Deamon5550: its not undefined it should throw an exception

Limeth: Well, yes, I could add a line that the method might throw a specific exception, but it does not necessarily check the validity of the input crafting grid.

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,74 @@
...
+
+import javax.annotation.Nonnull;
+
+public final class SmeltingResult {
+
+    private final ItemStackSnapshot result;
+    private final double experience;
+
+    @SuppressWarnings("ConstantConditions")
+    public SmeltingResult(@Nonnull ItemStackSnapshot result, double experience) {
```
Deamon5550: `@Nonnull` is not needed as [things are nonnull by default](https://github.com/SpongePowered/SpongeAPI/pull/1449/files#diff-3c5a576f501e9e5919cc697a5f5aa30fR25)

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,74 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.recipe.smelting;
+
+import static com.google.common.base.Preconditions.*;
```
kashike: No star imports.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,74 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.recipe.smelting;
+
+
```
stephan-gh: Unneeded empty line here and below the imports

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingResult.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Returns the amount of experience released after completing a recipe.
+     *
+     * @return The amount of experience released after fulfilling the
+     *         requirements of a {@link SmeltingRecipe}
+     */
+    public double getExperience() {
+        return this.experience;
+    }
+
```
stephan-gh, line 75: It would be useful to define a `toString()` and possibly `equals`+`hashCode` methods for this class (using Guava's ToStringHelper)

src/main/java/org/spongepowered/api/item/recipe/smelting/SmeltingRecipeRegistry.java
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.recipe.smelting;
+
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.item.recipe.RecipeRegistry;
+
+import java.util.Optional;
+
+public interface SmeltingRecipeRegistry extends RecipeRegistry<SmeltingRecipe> {
```
ST-DDT: Missing javadoc

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingResult.java
```diff
@@ -0,0 +1,84 @@
...
+     * {@link GridInventory}, contains {@link ItemStackSnapshot#NONE}s for
+     * slots which should be cleared.
+     *
+     * @return A list of {@link ItemStackSnapshot}s to be set in the input
+     *         {@link GridInventory}
+     */
+    public List<ItemStackSnapshot> getRemainingItems() {
+        return remainingItems;
+    }
+
```
Deamon5550, line 83: Should add the typical equals/toString/hashCode

Limeth:
Currently, `CraftingRecipe` has these two methods:
```java
ItemStackSnapshot getResult(GridInventory grid)
```
and
```java
Optional<CraftingResult> getResult(GridInventory grid, World world)
```
I think it might be wise to rename the first one, any suggestion on what the new name should be?

liach:
@Limeth I simply suggest to remove the first method.

Limeth:
Let me clarify once again:
`CraftingRecipe` has Sponge methods equivalent to native methods from `IRecipe`. This is to make it possible to implement a custom `CraftingRecipe` without having to use `ShapedCraftingRecipe` and `ShapelessCraftingRecipe`.

liach:
is furnace recipes crafting recipes?

Limeth:
No, `SmeltingRecipe`s do not inherit from `CraftingRecipe`s.

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
```diff
@@ -24,52 +24,19 @@
...
+     * A general result of this recipe. This result may be customized depending
+     * on the context.
      *
-     * @param grid An ItemGrid as input
-     * @return A list of ItemStacks or {@link Optional#empty()} if the given
-     *          ItemGrid does not match this recipe's requirements.
+     * @return The exemplary result of this recipe
      */
-    Optional<List<ItemStack>> getResults(GridInventory grid);
+    ItemStackSnapshot getExemplaryResult();
```
Aaron1011, line 40: I'm not sure if 'exemplary' makes sense here. What does this represent?

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Aaron1011: @Limeth: That link seems to be broken. Can the implementation take care of passing in `ItemStackSnapshot.NONE` to the predicate, if that's what the issue is?

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingRecipeRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+     * @return The {@link CraftingResult} if a recipe was found, or
+     *         {@link Optional#empty()} if not
+     */
+    default Optional<CraftingResult> getResult(GridInventory grid, World world) {
+        return findMatchingRecipe(grid, world)
+                .flatMap(recipe -> recipe.getResult(grid, world));
+    }
+
+    /**
+     * Creates a predicate with vanilla matching behavior.
```
Aaron1011, line 66: Can you clarify what this means? Otherwise, plugins may not find it very useful.

src/main/java/org/spongepowered/api/item/recipe/crafting/ShapedCraftingRecipe.java
Aaron1011: We should clarify what an 'asile' means here. If I'm correct in my understanding of it, each 'asile' corresponds to a row of the crafting grid.

Aaron1011 (changes requested):

src/main/java/org/spongepowered/api/item/recipe/Recipe.java
Faithcaio: I think its the item you see in the CraftingOutput before you take it out.

Limeth: The native `IRecipe` class contains an `ItemStack getRecipeOutput()`, where this value is used as the return value, and the `FurnaceRecipes` class contains a `Map<ItemStack, ItemStack> smeltingList`, where this value is used as a value in the map. I think these might be used in TooManyItems/NotEnoughItems to display the recipes, that's why I chose the word 'exemplary'.

Limeth:
@Aaron1011: You said, that _That link seems to be broken. Can the implementation take care of passing in ItemStackSnapshot.NONE to the predicate, if that's what the issue is?_
I couldn't reply directly to that comment for some reason. Anyway, I don't quite understand what you're pointing out here, could you elaborate?

Aaron1011:
@Limeth: What I'm wondering is this: Why do we need the overload of `where` that takes an `ItemStackSnapshot` (i.e. `where(char symbol, @Nullable ItemStackSnapshot ingredient)`). Is there something that can't be accomplished through passing in a Predicate?

Limeth:
@Aaron1011 No, this is just a convenience method which uses the `getVanillaIngredientPredicate` you pointed out above. We don't _need_ to overload it, but I'd find it useful.

src/main/java/org/spongepowered/api/item/recipe/crafting/CraftingResult.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        CraftingResult that = (CraftingResult) o;
+
+        return this.result.equals(that.result) && this.remainingItems.equals(that.remainingItems);
+    }
+
+    @Override
+    public int hashCode() {
+        int result1 = this.result.hashCode();
```
stephan-gh, line 101: You can use `Objects.hash(this.result, this.remainingItems)` here.

parlough:
It seems like recipes and similar will change heavily in 1.12 with some changes Dinnerbone is working on, so it may be best to revisit this with that update.

gabizou:
> It seems like recipes and similar will change heavily in 1.12(or maybe 1.13) with some changes Dinnerbone is working on, so it may be best to revisit this with that update.

I'd rather have recipes partially implemented for API 6 and 7 while we still can, if not fully implemented, so that when MC 1.12 comes around, we can easily update the API as necessary, along with the implementation.

Faithcaio:
MC 1.12 is out and Sponge still has no working Recipe API.
What is holding this back?

Zidane:
@Faithcaio @gabizou Not really interested in adding work back for older versions. Now that Forge 1.12 is in the wild, mods will skip 1.11 and head straight for that and leave 1.10 behind. This API needs to be updated for 1.12's changes and proceed from there.

parlough:
It seems he has a few changes to make here still but especially in the implementation with there being suggestions, questions, merge conflicts, and I'm assuming issues with 1.12 changes. @Limeth do you have the time and/or interest to continue these pull requests?

Limeth:
I got discouraged by the PR process and am no longer interested in contributing. If anyone wants to take over this PR, feel free to do so.

liach:
Yeah Sponge devs often respond slowly.

kashike:
@Limeth May I ask what discouraged you, and why you're no longer interested in contributing?

Limeth:
@kashike It was difficult for me to adapt to the required code style, most of the responses on my PRs were only addressing the "distracting" parts of code that didn't follow the code style. This lead to much longer periods of the iterative development process (code/compile/test/debug), as I felt like I didn't get enough feedback on the actual logic I was implementing.
Along with this, I got busy with school and when I returned, I realized programming Minecraft plugins wasn't what I'd like to continue on doing, at least for now.

jamierocks:
As it has been brought up, I thought I would share https://dev.to/terrehbyte/my-pull-request-etiquette - I thought it was a rather good read, and although not all of it would be easy for Sponge to implement, I'm sure it'd be worth some though ;)

There are a number of other good reads over on dev.to regarding code reviews, such as: https://dev.to/vaidehijoshi/crafting-better-code-reviews and https://dev.to/cyberomin/a-pull-request-is-submitted-what-next

kashike:
Superseded by #1582

PR #1467, opened 2017-01-21
Overhaul chunk pre-generation

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1149)

This PR is designed to open up the World PreGeneration code to allow plugins to effectively monitor and manipulate the pregeneration. This PR contains breaking changes that allow this to happen, so should not target anything but bleeding. This is based off my experience with NucleusMixins, and should allow all plugins to control the pre generation more effectively - and I'm offering this after being asked by @bloodmc to do so in https://github.com/NucleusPowered/Nucleus/issues/528

Discussion of the internals are discussed on the Common PR.

**Movement of API**

Prior to this, the pre-gen builder code was a sub interface to `WorldBorder.ChunkPreGenerate`. This seems to indicate that the pre-generation is related to the world border - it is not. Further, as the builder now returns a custom `ChunkPreGenerate` object, it made sense to move this all to it's own class. This does not affect how to get a builder.

The builder is subinterface to the `ChunkPreGenerate` interface, in line with other builders.

**Introduction of `ChunkPreGenerate`**

This interface is the base of the pre-gen system, and contains information about how the pre-generation is going - number of chunks generated, number skipped (that is, not required to be generated), etc. This is now returned from the `ChunkPreGenerate.Builder`, rather than just a `Task` - and plugins should now cancel pre-generation through this interface rather than the `Task`.

(It might be nice to add cancellation routines to the `Task` system, where code could be run if a task is cancelled, but that's out of the scope of this PR)

**Events**

There are now four events in the lifecycle of pre-generation, all in the `ChunkPreGenerationEvent` interface. They all return the `WorldProperties` and `ChunkPreGenerate` objects for proper progress reporting.

* **Pre**: fired before a pregeneration step is taken. If cancelled, cancels the entire pre generation. If `setSkipStep` is set to true, then that pre-generation step is skipped, but will try again when it is next scheduled to do so.
* **Post**: fired after a pregeneration step is taken. Contains stats about what has just been generated. If cancelled, cancels the entire pre generation.
* **Cancelled**: fired if the pregeneration is cancelled using `ChunkPreGenerate#cancel`.
* **Complete**: fired if the pregeneration is completed.

**Builder changes and additions**

* `Builder#addListener` adds a event listener for `ChunkPreGenerationEvent` to this specific task. This works in broadly the same way as the Custom Inventory listener methods.
* `Builder#logger` now adds the logger by adding a listener that fires on `ChunkPreGenerationEvent.Post` and `ChunkPreGenerationEvent.Complete`
* `from` now accepts a `ChunkPreGenerate` object, rather than a `Task` object.
* `start` returns the `ChunkPreGenerate` object now.

**Thoughts and notes**

* The `Pre/Post` events, if cancelled, will cancel the entire generation. Is this right? Should it be limited to `Pre`, and then only skip a step?
* Docmentation might need a few improvements!
* I think I caught everything in the style, but I've probably missed something!

**Test code & log results**

I will try to add more later - for now, this tests the listeners work as expected - both single task listeners, and global listeners.

https://gist.github.com/dualspiral/2664950914056f5a3ce005fba78811c8

### DIFF
--- a/src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
@@ -1,0 +1,115 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.world;
+
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.world.ChunkPreGenerate;
+
+import java.time.Duration;
+
+/**
+ * Base event for when a {@link ChunkPreGenerate} task
+ * is underway.
+ */
+public interface ChunkPreGenerationEvent extends TargetWorldEvent {
+
+    /**
+     * The object that contains the progress information for the
+     * current {@link ChunkPreGenerate}
+     *
+     * @return The {@link ChunkPreGenerate}
+     */
+    ChunkPreGenerate getChunkPreGenerate();
+
+    /**
+     * Event fired when chunks are about to be generated. Cancelling this task
+     * will cancel the entire pre-generation.
+     */
+    interface Pre extends ChunkPreGenerationEvent, Cancellable {
+
+        /**
+         * Returns whether the chunk generator will skip generation this time
+         * around. This only prevents the next step from being performed.
+         *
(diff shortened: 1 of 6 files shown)


bloodmc:
@Deamon5550 @DDoS please review.

Deamon5550:
Doesn't touch worldgen so if the other stuff is fine its fine by me.

src/main/java/org/spongepowered/api/world/ChunkPreGenerate.java
```diff
@@ -0,0 +1,200 @@
...
+ * Represents an ongoing chunk pre-generation.
+ */
+public interface ChunkPreGenerate {
+
+    /**
+     * The World that this task is operating on.
+     *
+     * @return The {@link WorldProperties}
+     */
+    WorldProperties getWorld();
```
DDoS: World or WorldProperties? This should be named accordingly. Also the doc needs to be updated.

DDoS (approved):

ST-DDT (approved):

src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
```diff
@@ -0,0 +1,115 @@
...
+         * Sets whether the next step should be skipped.
+         *
+         * <p>Use {@link #setCancelled(boolean)} to cancel the entire pre
+         * generation task.</p>
+         *
+         * @param skipStep If set to {@code true}, the next step will be
+         *                 skipped.
+         */
+        void setSkipStep(boolean skipStep);
+    }
```
liach, line 74: line

src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
```diff
@@ -0,0 +1,115 @@
...
+         */
+        int getChunksSkippedThisStep();
+
+        /**
+         * The {@link Duration} of the previous step.
+         *
+         * @return The {@link Duration}
+         */
+        Duration getTimeTakenForStep();
+    }
```
liach, line 104: line

src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Event fired when the pre generation has been completed.
+     */
+    interface Complete extends ChunkPreGenerationEvent {
+    }
```
liach: line

src/main/java/org/spongepowered/api/event/world/ChunkPreGenerationEvent.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Event fired when the pre generation has been cancelled.
+     */
+    interface Cancelled extends ChunkPreGenerationEvent {
+    }
```
liach: line

src/main/java/org/spongepowered/api/world/ChunkPreGenerate.java
```diff
@@ -0,0 +1,200 @@
...
+
+        /**
+         * Schedules the task with the {@link Game#getScheduler()}.
+         *
+         * @return The resulting {@link ChunkPreGenerate} that can be used
+         *         obtain progress
+         */
+        ChunkPreGenerate start();
+    }
+}
```
liach, line 201: lines

liach (approved):

PR #1476, opened 2017-02-02
Absorption data

kashike:
[**SpongeAPI**](https://github.com/SpongePowered/SpongeAPI/pull/1476) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1182)
Adds #1268 

Test plugin: https://gist.github.com/kashike/09d174f61c9fbb1c40a24096346a3179

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -91,6 +91,8 @@ public final class Keys {
 
     // SORTFIELDS:ON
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -88,6 +88,8 @@
 
     // SORTFIELDS:ON
 
+    public static final Key<Value<Float>> ABSORPTION = KeyFactory.fake("ABSORPTION");
```
gabizou: Needs to be double.

PR #1477, opened 2017-02-05
Move text actions to the implementation

kashike:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1186)

### DIFF
--- a/src/main/java/org/spongepowered/api/text/action/ClickAction.java
+++ b/src/main/java/org/spongepowered/api/text/action/ClickAction.java
@@ -24,8 +24,10 @@
  */
 package org.spongepowered.api.text.action;
 
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.ResettableBuilder;
 
 import java.net.URL;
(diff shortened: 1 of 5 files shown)


ST-DDT:
Is there a special reason for this?
IIRC the entire Text API is SpongeAPI implemented.

kashike:
It makes `HoverAction`s implement `TextRepresentable`. There's no way in the API to get the `Text` value of `HoverAction`s, since it comes from the implementation. Not all parts of the text API are implemented in the API.

src/main/java/org/spongepowered/api/text/action/TextActionFactory.java
```diff
@@ -0,0 +1,131 @@
...
+import org.spongepowered.api.statistic.achievement.Achievement;
+import org.spongepowered.api.text.Text;
+
+import java.net.URL;
+import java.util.function.Consumer;
+
+/**
+ * A factory to create {@link TextAction}s.
+ */
+public interface TextActionFactory {
```
Zidane: Not sure I like having factory classes. We moved away from this for the large part to a builder system which seems like it could do the below?

kashike:
#1635

PR #1479, opened 2017-02-06
Feature/loadingticket world

adpking:
The API currently does not provide a way for the ticket to provide the world from the ticket, despite having provided it during the creation of the ticket. Forge supports this, so this is essentially just adding to the completion of the Forge Chunk Manager wrapper.

This is paired with my SpongeForge pull request [#1248](https://github.com/SpongePowered/SpongeForge/pull/1248)

### DIFF
--- a/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
+++ b/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
@@ -189,7 +189,7 @@ public interface ChunkTicketManager {
         int getMaxNumChunks();
 
         /**
-         * Gets the world associated with this ticket during it's creation
+         * Gets the world associated with this ticket during its creation
          *
          * @return The world associated with this ticket
          */
--- a/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
+++ b/src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
@@ -188,12 +188,12 @@ public interface ChunkTicketManager {
          */
         int getMaxNumChunks();
 
-		/**
-		 * Gets the world associated with this ticket during it's creation
-		 *
-		 * @return The world associated with this ticket
-		 */
-		World getWorld();
+        /**
+         * Gets the world associated with this ticket during it's creation
+         *
+         * @return The world associated with this ticket
+         */
+        World getWorld();
 
         // TODO: NBTTag getCustomData(); (Is saved with ticket information)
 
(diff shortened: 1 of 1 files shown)


liach:
How about spongevanilla?

src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
```diff
@@ -188,6 +188,13 @@
          */
         int getMaxNumChunks();
 
+        /**
+         * Gets the world associated with this ticket during it's creation
+         *
+         * @return The world associated with this ticket
+         */
+        World getWorld();
```
liach: A weak reference may be better.

adpking: SpongeVanilla doesn't have any of the chunk loading wrapper as it's a wrapper for Forge. It shouldn't be used in vanilla at all, right?

liach: Then does SpongeVanilla still compile with this change?

liach: You cannot make sure that sponge ticket get recycled when the forge one did.

src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
```diff
@@ -188,6 +188,13 @@
          */
         int getMaxNumChunks();
 
+        /**
+         * Gets the world associated with this ticket during it's creation
```
liach: `its creation`

adpking: Good spot, thanks

src/main/java/org/spongepowered/api/world/ChunkTicketManager.java
adpking: Recycled how? My issue is that the LoadingTicket is useless at the moment unless you arbitrarily apply a ticket to chunks when the callback is running since there's no NBT alternative for tickets yet and there's no way to check the world. I don't want to assign a ticket created with a World instance like the nether to a chunk that exists in the overworld.

adpking: Also, I'm testing SpongeVanilla compilation now.

adpking: @liach It does compile with SpongeVanilla

Deamon5550:
See #1480 ...

PR #1481, opened 2017-02-08
Add methods to require data from a CompositeValueStore/ValueContainer

kashike:
This adds the following methods:
- `org.spongepowered.api.data.value.ValueContainer#require`
- `org.spongepowered.api.data.value.mutable.CompositeValueStore#require`

which simply call the associated `get` method and, if the value is not present, throws `NoSuchElementException`.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/value/ValueContainer.java
+++ b/src/main/java/org/spongepowered/api/data/value/ValueContainer.java
@@ -34,6 +34,7 @@ import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.data.value.immutable.ImmutableValueStore;
 import org.spongepowered.api.data.value.mutable.CompositeValueStore;
 
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


gabizou:
One word question: "Why?"

 I know what this adds, but I'm not following why this is being added.

Deamon5550:
Convenience, same reason we have the builder static methods and other such usability changes. Its a simple addition that makes the api easier to use.

kashike:
@Deamon5550 stole my words.

Zidane:
@gabizou 

This is actually a very elegant fix to the concerns brought up by a few peeps outside our community that you should know I'm talking about.

Honestly @kashike , bravo...sometimes the simplest things are best.

src/main/java/org/spongepowered/api/data/value/ValueContainer.java
```diff
@@ -70,6 +71,26 @@
...
+     * @param <E> The type of value
+     * @return The value
+     * @throws NoSuchElementException If the value is not supported or present
+     */
+    default <E> E require(Key<? extends BaseValue<E>> key) {
+        final Optional<E> optional = this.get(key);
+        if (optional.isPresent()) {
+            return optional.get();
+        }
+        throw new NoSuchElementException(String.format("Could not retrieve value for key '%s'", key.getId()));
```
liach: Why not use `orElseThrow`?

Issue #1277, opened 2016-06-20
Add JavaDocs for TOTAL_EXPERIENCE and HELD_EXPERIENCE


randombyte-developer:
I can imagine that Keys.HELD_EXPERIENCE is the experience the player currently has. But what is Keys.TOTAL_EXPERIENCE? Does it even belong to the player? The explanation should be added to the JavaDocs.

JBYoshi:
`TOTAL_EXPERIENCE` is for players. `HELD_EXPERIENCE` is for XP orbs.

randombyte-developer:
Thanks. Now this should be added to the JavaDocs.

Bammerbom:
And add a javadoc CONTAINED_EXPERIENCE too, I have no idea what it is for.

stephan-gh:
Fixed in https://github.com/SpongePowered/SpongeAPI/pull/1484

PR #1483, opened 2017-02-10
Remove duplicate experience key, add javadoc

ghost:
* Add Javadoc for the Keys mentioned in #1277 

Apparently `HELD_EXPERIENCE` and `CONTAINED_EXPERIENCE` were pointing to the same thing. `HELD_EXPERIENCE` was used in the API only whereas `CONTAINED_EXPERIENCE` [is actually implemented](https://github.com/SpongePowered/SpongeCommon/blob/6eec2f857519660061cd02f5ddd6f593f343fc11/src/main/java/org/spongepowered/common/data/processor/data/entity/ExpOrbDataProcessor.java#L46),

Therefore this PR removes the `HELD_EXPERIENCE` key and changes ExperienceOrb to use `CONTAINED_EXPERIENCE` instead.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -370,10 +370,26 @@ public final class Keys {
      */
     public static final Key<Value<Vector3i>> EXIT_POSITION = KeyFactory.fake("EXIT_PORTAL");
 
+    /**
+     * Represents the {@link Key} for representing the total amount of xp needed
+     * to advance from the current level to the next one.
+     *
+     * @see #EXPERIENCE_SINCE_LEVEL
+     */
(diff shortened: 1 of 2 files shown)


ghost:
On a side node, I'm not entirely convinced that `EXPERIENCE_FROM_START_OF_LEVEL` is a good name, but changing it would require not only a better idea but also changes to SpongeCommon, which is why I did not include a change in this PR.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -483,8 +499,6 @@
 
     public static final Key<MutableBoundedValue<Float>> HEIGHT = KeyFactory.fake("HEIGHT");
 
-    public static final Key<MutableBoundedValue<Integer>> HELD_EXPERIENCE = KeyFactory.fake("HELD_EXPERIENCE");
```
Deamon5550, line 486: HELD_EXPERIENCE should be deprecated not removed, and then it'll be removed in the next major version bump.

Deamon5550 (changes requested):

ghost:
Superseded by #1484

PR #1487, opened 2017-02-14
ItemStack isEmpty

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1487) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1205)

Adds an isEmpty Method to ItemStacks

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
@@ -40,6 +40,7 @@ import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.data.value.BaseValue;
 import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.item.ItemTypes;
 import org.spongepowered.api.text.translation.Translatable;
 
 import java.util.Map;
@@ -78,6 +79,14 @@ public interface ItemStack extends DataHolder, Translatable {
         return builder().itemType(itemType).quantity(quantity).build();
     }
 
+    /**
+     * Returns an empty {@link ItemStack}
+     *
+     * @return The empty ItemStack
+     */
+    static ItemStack empty() {
+        return builder().itemType(ItemTypes.NONE).build();
+    }
 
     /**
      * Gets the {@link ItemType} of this {@link ItemStack}.
@@ -134,6 +143,16 @@ public interface ItemStack extends DataHolder, Translatable {
      * @return True if this equals the ItemStack
      */
     boolean equalTo(ItemStack that);
+
+    /**
+     * Returns true if {@link #getQuantity()} is zero and therefore this
+     * ItemStack is empty.
+     *
+     * <p>In Vanilla empty ItemStacks are not rendered by the client.</p>
+     *
+     * @return True if this ItemStack is empty
+     */
+    boolean isEmpty();
 
     @Override
     ItemStack copy();
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStackSnapshot.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStackSnapshot.java
@@ -57,6 +57,14 @@ public interface ItemStackSnapshot extends ImmutableDataHolder<ItemStackSnapshot
     int getCount();
 
     /**
+     * Returns true if {@link #getCount()} is zero and therefore this
+     * ItemStackSnapshot is empty.
+     *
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -79,6 +79,13 @@ static ItemStack of(ItemType itemType, int quantity) {
         return builder().itemType(itemType).quantity(quantity).build();
     }
 
+    /**
+     * Returns an empty {@link ItemStack}
+     *
+     * @return The empty ItemStack
+     */
+    static ItemStack empty() { return builder().itemType(ItemTypes.AIR).build(); }
```
stephan-gh: The method body should be on an extra line.

kashike: `return of(ItemTypes.AIR, 1);`

Faithcaio:
I'm not sure what to do with those NONE and AIR ItemTypes.
NONE seems redundant as Vanilla already uses AIR for empty ItemStacks. `Item.getItemFromBlock(Blocks.AIR)`
imo we should use only one of the two everywhere in the API.

It is currently possible to create NONE ItemStacks with a non-zero quantity. :S

src/main/java/org/spongepowered/api/item/inventory/ItemStackSnapshot.java
```diff
@@ -57,6 +57,13 @@
     int getCount();
 
     /**
+     * Returns true if this ItemStackSnapshot is of Type air or has 0 or less quantity.
```
Zidane: Don't mention that it is of type air.

Simply mention that the implementation defines what empty means.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -136,6 +144,14 @@ static ItemStack of(ItemType itemType, int quantity) {
      */
     boolean equalTo(ItemStack that);
 
+    /**
+     * Returns true if this ItemStack is empty. Empty ItemStacks always
```
Zidane: Same as the comment below.

src/main/java/org/spongepowered/api/item/inventory/ItemStackSnapshot.java
Faithcaio: So not even mention 0 quantity?
Not sure how to put this into words best.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -135,6 +144,13 @@ static ItemStack of(ItemType itemType, int quantity) {
      */
     boolean equalTo(ItemStack that);
 
+    /**
+     * Returns true if this ItemStack is empty.
```
gabizou: Could you elaborate what empty means?

Faithcaio: like that?

src/main/java/org/spongepowered/api/item/inventory/ItemStackSnapshot.java
```diff
@@ -57,6 +57,13 @@
     int getCount();
 
     /**
+     * Returns true if this ItemStackSnapshot is empty.
+     *
+     * @return True if this ItemStackSnapshot is empty
+     */
+    boolean isEmpty();
```
ST-DDT: You need to add the quantity in the javadocs here as well.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -135,6 +144,14 @@ static ItemStack of(ItemType itemType, int quantity) {
      */
     boolean equalTo(ItemStack that);
 
+    /**
+     * Returns true if {@link #getQuantity()} is zero and therefore this
+     * ItemStack is empty.
```
gabizou, line 149: I meant more of: "What does an empty item stack differ from a 'non-empty' item stack? Does the client still render the item? Is it even an item to be considered by inventory logic? Can it be changed? etc."

Faithcaio: Wouldn't that be implementation detail?
In vanilla it won't render it because empty ItemStacks are always of type AIR.
You CAN change the quantity back up to a positive number.
Then the Item is no longer of type AIR. (if it was not before).
Offering an empty Itemstack will do nothing.
But setting it is a valid thing to do.

Faithcaio:
ping @gabizou

PR #1490, opened 2017-02-16
Overall API code cleanup

gabizou:
This is to finish/fix all current check style warnings within the API and potentially all OCD issues towards an API 5.1.0 release.

Comments regarding javadoc enhancements are welcome.


me4502:
Is this something that can be easily ported forwards to 1.11? Or is it meant just for stable-5.1.0

Mumfrey:
@me4502 the reason I've done my bits as very small commits is to help with merging it, so should be able to get most of this in bleeding.

ghost:
Will new PRs have to fix all checkstyle errors now in order to be merged to keep the code base clean?

stephan-gh:
I'm not entirely sure why, but it seems like there are much more Javadoc warnings with broken references now: https://travis-ci.org/SpongePowered/SpongeAPI/jobs/202222669#L353-L398
It seems like both classes are inner classes so it might be related to that.

Mumfrey:
> Will new PRs have to fix all checkstyle errors now in order to be merged to keep the code base clean?

@saladoc to be honest with you, the only reaons this has gotten so bad is because people (both core devs and PR authors) have just been selectively ignoring checkstyle rules and obviously don't actually run checkstyle in their IDE (which would make all these issues blindingly obvious), so I don't think it's unreasonable to expect checkstyle to pass going forward.

PR #1494, opened 2017-02-19
BlockCarriers

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1494) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1640)

Adds Interfaces for Carrier that are Blocks or multiple Blocks
also adds access for inventories that can be accessed from a particular direction.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
@@ -22,22 +22,36 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.block.tileentity.carrier;
+package org.spongepowered.api.item.inventory;
 
-import org.spongepowered.api.block.tileentity.TileEntity;
-import org.spongepowered.api.item.inventory.BlockCarrier;
-import org.spongepowered.api.item.inventory.Carrier;
-import org.spongepowered.api.item.inventory.type.TileEntityInventory;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.World;
+
+import java.util.Optional;
 
 /**
- * Represents a {@link TileEntity} that is a carrier of 
- * {@link TileEntityInventory}.
- *
- * <p>Additional information regarding the {@link TileEntityCarrier} is
- * dependent on the {@link TileEntity} itself.</p>
+ * Represents a Block that is a carrier of an Inventory
+ * e.g. An Anvil
  */
-public interface TileEntityCarrier extends TileEntity, BlockCarrier {
+public interface BlockCarrier extends Carrier {
 
-    @Override
-    TileEntityInventory<TileEntityCarrier> getInventory();
+    /**
+     * Returns the Location of the Block.
+     *
+     * @return The Location of the Block
+     */
+    Location<World> getLocation();
+
+    /**
+     * Returns the Inventory owned by this Carrier and accessible from given direction.
+     *
+     * <p>e.g. A Furnace accessed from {@link Direction#DOWN} will return its Fuel Slot</p>
+     *
+     * <p>When no inventory is accessible a {@link EmptyInventory} is returned</p>
+     *
+     * @param from The Direction
+     * @return The inventory at the location when accessed from given direction
+     */
+    Inventory getInventory(Direction from);
 }
--- a/src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
@@ -22,22 +22,46 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.block.tileentity.carrier;
+package org.spongepowered.api.item.inventory;
 
-import org.spongepowered.api.block.tileentity.TileEntity;
-import org.spongepowered.api.item.inventory.BlockCarrier;
-import org.spongepowered.api.item.inventory.Carrier;
-import org.spongepowered.api.item.inventory.type.TileEntityInventory;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.api.world.Location;
+import org.spongepowered.api.world.World;
+
+import java.util.List;
+import java.util.Optional;
 
 /**
- * Represents a {@link TileEntity} that is a carrier of 
- * {@link TileEntityInventory}.
- *
- * <p>Additional information regarding the {@link TileEntityCarrier} is
- * dependent on the {@link TileEntity} itself.</p>
+ * Represents Multiple Blocks that are the Carrier for an Inventory.
+ * e.g. A DoubleChest Inventory carried by two blocks
  */
-public interface TileEntityCarrier extends TileEntity, BlockCarrier {
+public interface MultiBlockCarrier extends BlockCarrier {
 
-    @Override
-    TileEntityInventory<TileEntityCarrier> getInventory();
+    /**
+     * Returns the Locations of the Blocks
+     *
+     * @return The Locations of the Blocks
+     */
+    List<Location<World>> getLocations();
+
+    /**
+     * Returns the Inventory at given location if owned by this Carrier
+     *
+     * @param at The location
+     * @return The inventory at given location
+     */
+    Optional<Inventory> getInventory(Location<World> at);
+
+    /**
+     * Returns the Inventory at given location if owned by this Carrier and
+     * accessible from given direction.
+     *
+     * e.g. A Furnace accessed from {@link Direction#DOWN} will return its Fuel Slot
+     *
+     * @param at The Location
+     * @param from The Direction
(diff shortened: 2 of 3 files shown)


src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
```diff
@@ -0,0 +1,20 @@
+package org.spongepowered.api.item.inventory;
```
kashike, line 25: Missing licence header.

src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
```diff
@@ -0,0 +1,27 @@
+package org.spongepowered.api.item.inventory;
```
kashike: Missing licence header.

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * Returns the Location of the Block
+     *
+     * @return The Location of the Block
+     */
+    Location<World> getLocation();
+}
+
+// TODO Implementation: Could me mixed in into the Containers
```
kashike: Missing new line at end of file.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
```diff
@@ -0,0 +1,27 @@
...
+import org.spongepowered.api.world.World;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents Multiple Blocks that are the Carrier for an Inventory.
+ * e.g. A DoubleChest Inventory carried by two blocks
+ */
+public interface MultiBlockCarrier extends Carrier {
```
kashike: Is it possible to inherit `BlockCarrier`? rename it to `SingleBlockCarrier` and add a parent `BlockCarrier`?

Faithcaio: like that?

gabizou:
@Faithcaio I think there's something to be had here from the extra Fluid Data API stuff I added to work in conjunction with SpongeAPI (since Fluids is something specific to Forge, and not a Vanilla based concept). Basically, some mods have direction based inventories, which of course pull different items tacks depending on the direction used, instead of having an upper or lower inventory.

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
liach: Seems fixed

src/main/java/org/spongepowered/api/item/inventory/SingleBlockCarrier.java
```diff
@@ -0,0 +1,44 @@
...
+ * e.g. An Anvil
+ */
+public interface SingleBlockCarrier extends BlockCarrier {
+
+    /**
+     * Returns the Location of the Block
+     *
+     * @return The Location of the Block
+     */
+    Location<World> getLocation();
```
liach: You can default out this method to use the first location provided in that list

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
```diff
@@ -0,0 +1,69 @@
...
+ * e.g. An Anvil
+ */
+public interface BlockCarrier extends Carrier {
+
+    /**
+     * Returns the Locations of the Blocks
+     *
+     * @return The Locations of the Blocks
+     */
+    List<Location<World>> getLocations();
```
Zidane: You've got this backwards. A BlockCarrier should be considered a single one by default. The sub interface MultiBlockCarrier should only be the one with multiple locations.

ryantheleach: Plugins wishing to handle BlockCarriers generically would then be out of luck? How do you generalize a sub-interface having a larger list then a parent interface? Or would one of them be the 'primary' location?

Zidane: One would be primary.

JBYoshi: I'd prefer the original way; it makes it harder for plugin developers to write (forgive the wording) dumb code. With the multiple block type as a subtype, it's very easy for plugins to forget there are multiple blocks; the correct code to get all blocks (to ensure correct behavior) would be this:

```java
Collection<Location> locations;
if (carrier instanceof MultiBlockCarrier) {
    locations = ((MultiBlockCarrier) carrier).getLocations();
} else {
    locations = Collection.singleton(carrier.getLocation());
}
```
On the other hand, it's easy to forget to use all this - using simply `carrier.getLocation()` could result in incorrect behavior.

On the other hand, in the original method, plugin developers would have to either handle multiple locations or cast to force it to only be one location. It's still possible to ignore it, but it's an extra step to nudge developers in the right direction.

src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory;
+
+/**
+ * Represents Multiple Blocks that are the Carrier for an Inventory.
+ * e.g. A DoubleChest Inventory carried by two blocks
+ */
+public interface MultiBlockCarrier extends BlockCarrier {
```
liach, line 38: You should probably add a method `List<Location<World>> getAssociatedLocations()`. which the locations are guaranteed to vary from that of the carrier whose method is called.

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
liach: Some of the mod multiblocks may have different inventories for different blocks in the multiblock structure. In this case, Zidane's idea prevails. You should also move the two methods below to `MultiBlockCarrier`.

src/main/java/org/spongepowered/api/item/inventory/Carrier.java
```diff
@@ -36,6 +34,6 @@
      *
      * @return The inventory this Carrier is holding
      */
-    CarriedInventory<? extends Carrier> getInventory();
+    Inventory getInventory();
```
liach: You shouldn't really change this because you may return an inventory from a random block entity

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
Faithcaio: I'm not sure we can support mod multiblocks at all.
I don't see any concept of multiblock things in forge.

src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
Faithcaio: So you mean getLocations() minus getLocation()?

liach: Yes. I think this way, the method would be more useful in plugins, and the rest of those locations don't need to be in a list necessarily.

src/main/java/org/spongepowered/api/item/inventory/BlockCarrier.java
```diff
@@ -0,0 +1,66 @@
...
+     * Returns the Inventory at given location if owned by this Carrier and
+     * accessible from given direction.
+     *
+     * e.g. A Furnace accessed from {@link Direction#DOWN} will return its Fuel Slot
+     *
+     * @param at The Location
+     * @param from The Direction
+     * @return The inventory at the location when accessed from given direction
+     */
+    Optional<Inventory> getInventory(Location<World> at, Direction from);
```
liach: I would suggest these methods:
`Optional<Inventory> getInventory(Direction from);`
This can be in block carrier, with the location implied to be the called carrier's location.
(I suggest leaving my method in this class and moving two methods above to `MultiBlockCarrier`)

src/main/java/org/spongepowered/api/item/inventory/MultiBlockCarrier.java
```diff
@@ -0,0 +1,67 @@
...
+import org.spongepowered.api.world.World;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents Multiple Blocks that are the Carrier for an Inventory.
+ * e.g. A DoubleChest Inventory carried by two blocks
+ */
+public interface MultiBlockCarrier extends BlockCarrier {
```
liach, line 38: Are there one instance of this carrier at each position?

Faithcaio: This is mixed in to InventoryLargeChest.
So the same carrier for both positions. But the InventoryLargeChest is recreated every time you access it. So it probably wont stay the same across multiple events

liach: That sounds great! Then it looks all good.

PR #1502, opened 2017-02-26
GuiIDProperty and ContainerProvider

Faithcaio:
*API* | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1221)

- [x] GUIID Property
     - provides the GUI-ID for displaying the inventory(container) to the client

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/property/GuiIdProperty.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/property/GuiIdProperty.java
@@ -1,0 +1,50 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.property;
+
+import org.spongepowered.api.data.Property;
+
+/**
+ * A property to define which GUI should be opened on the client for the inventory.
+ */
+public class GuiIdProperty extends AbstractInventoryProperty<String, GuiId> {
+
+    public GuiIdProperty(GuiId value) {
+        super(value);
+    }
+
+    @Override
+    public int compareTo(Property<?, ?> other) {
+        if (other == null) {
+            return 1;
+        }
+
+        if (other instanceof GuiIdProperty) {
+            return this.getValue().getId().compareTo(((GuiId) other.getValue()).getId());
+        }
+
+        return 1;
+    }
+}
--- a/src/main/java/org/spongepowered/api/item/inventory/property/GuiIds.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/property/GuiIds.java
@@ -1,0 +1,88 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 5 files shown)


Faithcaio:
@gabizou, @Mumfrey, @Zidane could I get your opinion on this API?

A lot of people are asking for custom inventories that include the vanilla container logic.

src/main/java/org/spongepowered/api/item/inventory/property/GuiIDProperty.java
```diff
@@ -0,0 +1,35 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.property;
+
+/**
+ * A property to define which GUI should be opened on the client for the inventory
+ */
+public class GuiIDProperty extends StringProperty {
```
kashike: `GuiIdProperty`

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetype.java
```diff
@@ -130,6 +143,15 @@ default Builder title(Text title) {
         Builder property(InventoryProperty<String, ?> property);
 
         /**
+         * Sets the container provider. Use null to fallback to a basic container.
+         * TODO describe more / example? e.g. AnvilContainer can Repair basic container cannot.
+         *
+         * @param containerProvider the container provider
+         * @return Fluent pattern
+         */
+        Builder container(@Nullable BiFunction<Inventory, Player, Container> containerProvider);
```
gabizou: Could I ask about when the function could possibly change to provide more information (arguments?) I would rather a plain object interface to avoid having to royally break the method in the future (allows for future deprecation and such).

src/main/java/org/spongepowered/api/item/inventory/property/GuiIdProperty.java
```diff
@@ -0,0 +1,35 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.property;
+
+/**
+ * A property to define which GUI should be opened on the client for the inventory
```
gabizou: Missing period at the end.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetype.java
```diff
@@ -92,6 +97,14 @@ static Builder builder() {
     <T extends InventoryProperty<String, ?>> Optional<T> getProperty(Class<T> type, String key);
 
     /**
+     * Returns a Container for given inventory and player
+     * @param inventory The Player
```
gabizou: space between the parameters and method documentation.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetype.java
```diff
@@ -92,6 +97,14 @@ static Builder builder() {
     <T extends InventoryProperty<String, ?>> Optional<T> getProperty(Class<T> type, String key);
 
     /**
+     * Returns a Container for given inventory and player
```
gabizou: Missing period.

gabizou (changes requested):
Minor style changes.

src/main/java/org/spongepowered/api/item/inventory/InventoryArchetype.java
Faithcaio: its based on the IInteractionObject interface from mc:
`Container createContainer(InventoryPlayer playerInventory, EntityPlayer playerIn);`
InventoryPlayer is not needed because it is obtainable from EntityPlayer.

*added Interface

Faithcaio:
ping @gabizou

JBYoshi:
I was initially quite confused on the name; I thought it was referring to an ID unique to the inventory or the container. I'd prefer if it was reworded to something like "display type" (though that might not be the best wording either). Also, I don't think it's a good idea to expose the raw `String`; should this be a `CatalogType` or something similar?

Faithcaio:
There is no registration for those. Not sure how we could make cataloguetypes work with gui from mods.

Zidane:
@Faithcaio 

At least catalog the Vanilla ones for sure. Get help or insight from @bloodmc on how to do something about mods.

src/main/java/org/spongepowered/api/item/inventory/property/GuiIds.java
```diff
@@ -0,0 +1,51 @@
...
+    public static final GuiId DISPENSER = DummyObjectProvider.createFor(GuiId.class, "DISPENSER"); // "minecraft:dispenser"
+    public static final GuiId CRAFTING_TABLE = DummyObjectProvider.createFor(GuiId.class, "CRAFTING_TABLE"); // "minecraft:crafting_table"
+    public static final GuiId BREWING_STAND = DummyObjectProvider.createFor(GuiId.class, "BREWING_STAND"); // "minecraft:brewing_stand"
+    public static final GuiId HOPPER = DummyObjectProvider.createFor(GuiId.class, "HOPPER"); // "minecraft:hopper"
+    public static final GuiId BEACON = DummyObjectProvider.createFor(GuiId.class, "BEACON"); // "minecraft:beacon"
+    public static final GuiId ENCHANTING_TABLE = DummyObjectProvider.createFor(GuiId.class, "ENCHANTING_TABLE"); // "minecraft:enchanting_table"
+    public static final GuiId ANVIL = DummyObjectProvider.createFor(GuiId.class, "ANVIL"); // "minecraft:anvil"
+    public static final GuiId VILLAGER = DummyObjectProvider.createFor(GuiId.class, "VILLAGER"); // "minecraft:villager"
+    public static final GuiId HORSE = DummyObjectProvider.createFor(GuiId.class, "HORSE"); // "EntityHorse"
```
Cybermaxke: Can't we make the id like `minecraft:horse` and map it internally to `EntityHorse`? It would be nice for consistency and including the plugin id that provides the `GuiId`.

Faithcaio: Guiids are not really provided. Its just a String.
Mods using them probably **should** prefix with their id but its not enforced anywhere afaik.

Cybermaxke: @Faithcaio In any case, `minecraft:horse` would be preferrable. Since uppercase characters aren't allowed in identifiers and this should be consistent across the API.

Issue #1396, opened 2016-10-23
GenericArguments.enumValue assumes capitalized names.


ryantheleach:
Using an enum that goes against Java naming conventions creates the following exception.

```
ava.lang.IllegalArgumentException: No enum constant au.id.rleach.pokeparticles.EffectType.FIESTA
    at java.lang.Enum.valueOf(Enum.java:238) ~[?:1.8.0_91]
    at org.spongepowered.api.command.args.GenericArguments$EnumValueElement.getValue(GenericArguments.java:857) ~[GenericArguments$EnumValueElement.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.args.PatternMatchingCommandElement.parseValue(PatternMatchingCommandElement.java:59) ~[PatternMatchingCommandElement.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.args.CommandElement.parse(CommandElement.java:83) ~[CommandElement.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.args.GenericArguments$SequenceCommandElement.parse(GenericArguments.java:273) ~[GenericArguments$SequenceCommandElement.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.spec.CommandSpec.populateContext(CommandSpec.java:284) ~[CommandSpec.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.spec.CommandSpec.process(CommandSpec.java:330) ~[CommandSpec.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.api.command.dispatcher.SimpleDispatcher.process(SimpleDispatcher.java:331) ~[SimpleDispatcher.class:1.8.9-1890-4.2.0-BETA-1762]
    at org.spongepowered.common.command.SpongeCommandManager.process(SpongeCommandManager.java:279) [SpongeCommandManager.class:1.8.9-1890-4.2.0-BETA-1762]
    at net.minecraft.command.ServerCommandManager.func_71556_a(SourceFile:80) [bd.class:?]
    at net.minecraft.network.NetHandlerPlayServer.func_147361_d(NetHandlerPlayServer.java:815) [lm.class:?]
    at net.minecraft.network.NetHandlerPlayServer.func_147354_a(NetHandlerPlayServer.java:794) [lm.class:?]
[...]

PR #1508, opened 2017-03-01
EnumValueElement#getValue case fix

hfreeb:
This is a fix to https://github.com/SpongePowered/SpongeAPI/issues/1396

At the moment, using EnumValueElement#getValue(String) only works when the enum field is all uppercase, this fixes that, making the search case sensitive.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -940,9 +940,7 @@ public final class GenericArguments {
 
         @Override
         protected Iterable<String> getChoices(CommandSource source) {
-            return Arrays.stream(this.type.getEnumConstants())
-                .map(Enum::name)
-                .collect(Collectors.toList());
+            return this.values.keySet();
         }
 
         @Override
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -929,14 +929,14 @@ public final class GenericArguments {
 
         @Override
         protected Iterable<String> getChoices(CommandSource source) {
-            return Arrays.asList(this.type.getEnumConstants()).stream()
-                .map(input -> input == null ? null : input.name())
+            return Arrays.stream(this.type.getEnumConstants())
+                .map(Enum::name)
                 .collect(Collectors.toList());
         }
 
         @Override
         protected Object getValue(String choice) throws IllegalArgumentException {
-            return Enum.valueOf(this.type, choice.toUpperCase());
+            return Enum.valueOf(this.type, choice);
         }
     }
 
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -930,7 +930,7 @@ public final class GenericArguments {
             this.values = Arrays.stream(type.getEnumConstants())
                     .collect(Collectors.toMap(
                             value -> value.name().toLowerCase(),
-                            value -> value,
+                            Function.identity(),
                             (value, value2) -> {
                                 throw new UnsupportedOperationException(type.getCanonicalName() + " contains more than one enum constant " +
                                         "with the same name, only differing by capitalization, which is unsupported.");
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -57,6 +57,7 @@ import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.storage.WorldProperties;
 
+import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -929,14 +929,14 @@ public static CommandElement bool(Text key) {
 
         @Override
         protected Iterable<String> getChoices(CommandSource source) {
-            return Arrays.asList(this.type.getEnumConstants()).stream()
+            return Arrays.stream(this.type.getEnumConstants())
                 .map(input -> input == null ? null : input.name())
```
liach: You can optimize this line to 
```java
.map(input::name)
```
since the elements of the array is guaranteed to be non-null.

liach (approved):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -929,14 +929,14 @@ public static CommandElement bool(Text key) {
...
-                .map(input -> input == null ? null : input.name())
+            return Arrays.stream(this.type.getEnumConstants())
+                .map(Enum::name)
                 .collect(Collectors.toList());
         }
 
         @Override
         protected Object getValue(String choice) throws IllegalArgumentException {
-            return Enum.valueOf(this.type, choice.toUpperCase());
+            return Enum.valueOf(this.type, choice);
```
kashike: This is going to make things fail if user input does not match the enum constant name. A better choice would be to populate a `Map<String, T>` during argument creation, with the `String` key *lowercased*, and simply get the `keySet` for choices, and a simple `get` here.

kashike (changes requested):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
hfreeb: Should I worry about cases when there are two different values with the same value but different capitilisation? As this is possible with enums. Obviously this is terrible practice, but with the implementation you suggested, there would be no way to differentiate between different values only differing by capitialisation.

Deamon5550: No I wouldn't worry about that case, usability is more important here.

kashike: You could warn during population of the map, if you want.

Desetude: This may be a stupid question, but how should I warn?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -921,22 +922,38 @@ public static CommandElement bool(Text key) {
 
     private static class EnumValueElement<T extends Enum<T>> extends PatternMatchingCommandElement {
         private final Class<T> type;
+        private final Map<String, T> values;
 
         EnumValueElement(Text key, Class<T> type) {
             super(key);
             this.type = type;
+            this.values = Arrays.stream(type.getFields())
```
kashike: Why can't you use `getEnumConstants` here?

kashike (changes requested):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -921,22 +922,37 @@ public static CommandElement bool(Text key) {
...
+                                        "with the same name, only differing by capitalization, which is unsupported.");
+                            }
+                    ));
         }
 
         @Override
         protected Iterable<String> getChoices(CommandSource source) {
-            return Arrays.asList(this.type.getEnumConstants()).stream()
-                .map(input -> input == null ? null : input.name())
+            return Arrays.stream(this.type.getEnumConstants())
```
kashike: You can simply do `return this.values.keySet();` here.

kashike (changes requested):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -921,22 +922,37 @@ public static CommandElement bool(Text key) {
...
         private final Class<T> type;
+        private final Map<String, T> values;
 
         EnumValueElement(Text key, Class<T> type) {
             super(key);
             this.type = type;
+            this.values = Arrays.stream(type.getEnumConstants())
+                    .collect(Collectors.toMap(
+                            value -> value.name().toLowerCase(),
+                            value -> value,
```
liach: Use `Function::identity`

liach (changes requested):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -921,22 +922,37 @@ public static CommandElement bool(Text key) {
...
     private static class EnumValueElement<T extends Enum<T>> extends PatternMatchingCommandElement {
         private final Class<T> type;
+        private final Map<String, T> values;
 
         EnumValueElement(Text key, Class<T> type) {
             super(key);
             this.type = type;
+            this.values = Arrays.stream(type.getEnumConstants())
+                    .collect(Collectors.toMap(
+                            value -> value.name().toLowerCase(),
```
liach, line 932: Can you use `value.name::toLowerCase`?

hfreeb: Don't think that's possible in this situation, unless I'm missing something.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -921,22 +922,35 @@ public static CommandElement bool(Text key) {
 
     private static class EnumValueElement<T extends Enum<T>> extends PatternMatchingCommandElement {
         private final Class<T> type;
+        private final Map<String, T> values;
```
ST-DDT, line 925: Should this map be made immutable? Or should we add a feature to add custom mappings? (Some well known aliases).

liach: It is already immutable since it is created by a collector.

liach (approved):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
ST-DDT: Nope, its not. `Collectors` don't make any  assertions regarding its (im)mutability. Thats why you should use your own collection type if you want to be sure.
[`Collectors#toMap(...)`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/stream/Collectors.java#1267)

Maybe this can be added to the Sponge's `GuavaCollectors` util class.

liach: Why should it be immutable? The code never modifies it in this pull, and the map is already private.

kashike (approved):

PR #1513, opened 2017-03-06
Add error message for @Listener methods.

JBYoshi:
See https://github.com/SpongePowered/SpongeCommon/pull/1233#issuecomment-284154794.

### DIFF
--- a/src/ap/resources/META-INF/services/javax.annotation.processing.Processor
+++ b/src/ap/resources/META-INF/services/javax.annotation.processing.Processor
@@ -1,1 +1,2 @@
 org.spongepowered.plugin.processor.PluginProcessor
+org.spongepowered.plugin.processor.ListenerProcessor
--- a/src/ap/java/org/spongepowered/plugin/processor/ListenerProcessor.java
+++ b/src/ap/java/org/spongepowered/plugin/processor/ListenerProcessor.java
@@ -1,0 +1,104 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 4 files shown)


src/ap/java/org/spongepowered/processor/event/ListenerProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+                    msg.printMessage(Diagnostic.Kind.ERROR, "method must not be abstract", method);
+                }
+                if (method.getEnclosingElement().getKind().isInterface()) {
+                    msg.printMessage(Diagnostic.Kind.ERROR, "interfaces cannot declare listeners", method);
+                }
+                if (method.getReturnType().getKind() != TypeKind.VOID) {
+                    msg.printMessage(Diagnostic.Kind.ERROR, "method must return void", method);
+                }
+                List<? extends VariableElement> parameters = method.getParameters();
+                if (parameters.isEmpty() || isTypeSubclass(parameters.get(0), EVENT_CLASS)) {
```
stephan-gh: This should be `|| !isTypeSubClass(...)` because you want to show the error if the method doesn't have an `Event` as first parameter

src/ap/java/org/spongepowered/processor/event/ListenerProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+
+        return false;
+    }
+
+    private boolean isTypeSubclass(Element typedElement, String subclass) {
+        Elements elements = processingEnv.getElementUtils();
+        Types types = processingEnv.getTypeUtils();
+
+        TypeMirror event = types.getDeclaredType(elements.getTypeElement(subclass));
+        return types.isAssignable(event, typedElement.asType());
```
stephan-gh: The order of the arguments for `isAssignable` is wrong here. From the Javadocs:

> returns true if and only if the **first type is assignable to the second**

You want to check if the parameter type extends `Event` which means that you need to check if the parameter is assignable to `Event`. `Event` isn't directly assignable to a subclass of it.

stephan-gh (changes requested):
:+1:

src/ap/resources/META-INF/services/javax.annotation.processing.Processor
```diff
@@ -1 +1,2 @@
-org.spongepowered.plugin.processor.PluginProcessor
+org.spongepowered.processor.plugin.PluginProcessor
```
stephan-gh: While I understand why you move it, I would prefer if it stays in its original package so [we don't break older versions of SpongeGradle](https://github.com/SpongePowered/SpongeGradle/blob/master/src/main/groovy/org/spongepowered/gradle/plugin/SpongePluginBasePlugin.groovy#L37)

src/ap/java/org/spongepowered/processor/ProcessorUtils.java
```diff
@@ -0,0 +1,50 @@
...
+        final String name = clazz.getName();
+        for (TypeElement element : elements) {
+            if (element.getQualifiedName().contentEquals(name)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+}
```
liach: Need a new line

src/ap/java/org/spongepowered/processor/event/ListenerProcessor.java
liach: I think it is fine here.

src/ap/resources/META-INF/services/javax.annotation.processing.Processor
liach: We can just ask plugin devs to update SpongeGradle.

src/ap/java/org/spongepowered/processor/ProcessorUtils.java
JBYoshi: @liach Fixed.

src/ap/resources/META-INF/services/javax.annotation.processing.Processor
JBYoshi: @Minecrell Moved back to the original package.

src/ap/java/org/spongepowered/processor/event/ListenerProcessor.java
JBYoshi: @Minecrell Fixed.

stephan-gh (approved):

liach (approved):

ST-DDT (approved):

kashike (approved):

me4502 (approved):

PR #1514, opened 2017-03-09
Pattern.quote input strings in PatternMatchingCommandElement

dags-:
Some input strings can cause PatternSyntaxException's to be thrown when compiling the regex pattern due to them containing certain characters.

For example, compiling the user input for a blockstate `minecraft:stone[` fails due to the open bracket: [log snippet](https://gist.github.com/dags-/a6cdb24a4b1605c013d456f33fed154e)

Pattern.quote()'ing the input string resolves this in my tests.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
+++ b/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
@@ -80,7 +80,7 @@ public abstract class PatternMatchingCommandElement extends CommandElement {
 
     Pattern getFormattedPattern(String input) {
         if (!input.startsWith("^")) { // Anchor matches to the beginning -- this lets us use find()
-            input = "^" + input;
+            input = "^" + Pattern.quote(input); // Quote input to avoid syntax errors caused by characters such as '['
         }
         return Pattern.compile(input, Pattern.CASE_INSENSITIVE);
 


src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
```diff
@@ -80,7 +80,7 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
 
     Pattern getFormattedPattern(String input) {
         if (!input.startsWith("^")) { // Anchor matches to the beginning -- this lets us use find()
-            input = "^" + input;
+            input = "^" + Pattern.quote(input); // Quote input to avoid syntax errors caused by characters such as '['
```
dualspiral, line 83: This is only going to affect some (admittedly, the vast majority of) inputs - if I try to parse an argument input as `^minecraft:stone[`, it'll still fail, because of the if statement above.

The better option here would be to treat the input as a literal regardless of what it starts with, quote that, and then _always_ add `^` to the beginning - that way, any argument that uses this element as a base can _also_ start with `^`.

(edit: I had a look at the arguments that use this, wondering if there is ever any need to pass a regex into here, rather than a literal - it doesn't seem so and as this parses the input which is always going to be a literal, fixing it here should suffice),

dualspiral (changes requested):

zml2008:
Wrapping the input in `Pattern.quote` defeats the whole point of the PatternMatchingCommandElement which is to allow matching results by regex. A better solution would be to try to parse the input as regex, and only if that fails `Pattern.quote` the input and parse the quoted input -- that way a valid regex can still be interpreted as a regex.

dualspiral:
The problem arises in edge cases when a valid regex is also a valid literal, but the regex would give a different answer. Taking the example above with block states, one example is `minecraft:stone[variant=granite]`. Trying to parse it as a regex means that the user in question can't get the entry in question unless they know the argument is a regex and know to escape the first `[`.

Worse, let's consider `minecraft:stone[variant=smooth_granite]`. Again - this is a valid regex, but I imagine it will match (after addition of the initial caret), `minecraft:stone_slab` because of the underscore - and would require a user to know that the argument is a regex and that they again, have to escape the first `[`. It might match other things, right now, just working from the top of my head!

My concern is that, given the above, most server owners do not expect that the argument parses like this and getting a literal should be easier than inputting the regex. I don't think we can expect players to have to catch these edge cases themselves, telling them to type in `minecraft:stone\[variant=smooth_granite\]` in order to avoid getting stone steps. The majority of server owners simply expect that they can put in the start of some ID, or name, and that it'll match the literal.

I'm honestly not sure how to solve this in the best way. My initial thoughts were to parse the literal first, then parse the regex if that turned up nothing, and then allow for an `r:` token at the beginning of the argument to indicate you wish to parse a regex straight off the bat. It's probably not the best way, but parsing the regex first when many players (and developers) will not expect the argument to be a regex (or know what a regex is) is just going to end up in confusion.

dags-:
How about something like this:

```java
private static Pattern getFormattedPattern(String input) {
    String literal = ".*?" + Pattern.quote(input);

    if (!input.startsWith("^")) {
        input = "^" + input;
    }
        
    try {
        return Pattern.compile(input + "|" + literal, Pattern.CASE_INSENSITIVE);
    } catch (PatternSyntaxException e) {
        return Pattern.compile(literal, Pattern.CASE_INSENSITIVE);
    }
}
```

So if the input is valid regex, match using the expression _or_ by finding the literal somewhere within the string.

Otherwise only match if the literal is somewhere within the string.

dualspiral:
The problem with that is that you aren't prioritising one over the other, which means you get the results from the regex _and_ the literal. Like my previous example above, take an argument that selects block states. Again, let's select `minecraft:stone[variant=smooth_granite]`, that'll get us this block as expected (from the literal), and _also_ all the `minecraft:stone_slab` block state variants (from the regex) - probably not expected in this case.

I also don't think being able to select the literal from _anywhere_ in the string is a particularly good idea. The expectation is currently that whatever you type into the argument forms the _start_ of the string, we should keep it that way.

This issue is more complex than it is on the surface, because we're moving into the realms of potentially breaking the current behaviour - however unexpected it might be for some. Currently, the argument is set up to accept a regex _only_, but ideally, I think we should be doing the simpler check of comparing literal strings by default, and _optionally_ turning on regex (somehow) for those that need/want it.

dags-:
If the unformatted string matches an element exactly then only that element is returned:
https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java#L58

I must've been stuck tab-completing with vanilla commands for too long, thought that was the norm in Sponge as well.

Maybe the CommandElement for CatalogTypes just needs to _not_ extend this class and handle parsing/tab-completions itself.

dualspiral:
I had missed that for exact matches, good point, thanks. It certainly allays most of my concerns - I should have tested it, I was mostly commenting from work!

> I must've been stuck tab-completing with vanilla commands for too long, thought that was the norm in Sponge as well.

Not with this argument type, as it currently _always_ anchors the pattern to the start, and I think that needs to continue. 

For the most part, I think extending this argument type for catalog types is OK because I see why regex can be useful, it's mostly block states that cause the issue as they contain square brackets.

ST-DDT:
I'm not sure about the regex thing at all. First of all most players habe no clue about it especially its syntax requirements.

Lets disassemble it a bit:
Who would type something like `.*stone.*\[variant=mossy\]` in chat just to find a mossy cobblestone wall? I hope nobody.
If i use `skele` i would like to get skeleton, wither skeleton and skeleton horse. Dont just skip the wither because it doesnt start with skeleton.
If i use `variant=stone, direction=NORTH` i would like to also find entries where the order is reversed or there is no space in between both.
Regex can DDos your server.

Can somebody give me a usecase where a regex is useful for players to specify a single element/value?

ryantheleach:
Essentials on Bukkit had this behavior, and it was confusing for some users, others relied entirely upon it.

e.g. banning a user that's currently online, but has a hard to type username that ends in 64? /ban *64

but this was also able to target multiple people simultaneously.

Noting that this was pre selectors, nowadays I think you would struggle to find someone who expects a command to affect multiple targets without explicitly targeting them using some form of selector.

gabizou:
Just as a reminder, this is blocked pending our 1.13 update.

Zidane:
Closing this PR as the Commands Re-write PR will change all of this. Regardless, I want to thank @dags- for his contribution here.

PR #1516, opened 2017-03-11
Added EconomyService methods to get all unique and virtual accounts

### LINKED ISSUE #1398: Can't get list of accounts from EconomyService
The service lets me get/create accounts and check if accounts exist, but doesn't let me get a list of all unique and virtual accounts.

It should.

I need access to arbitrary virtual accounts, but i don't know what they are, which means i can't tab complete their identifiers.

zachoooo:
Fixes #1398

### DIFF
--- a/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
+++ b/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
@@ -115,4 +115,18 @@ public interface EconomyService extends ContextualService<Account> {
      * @return The {@link Account}, if available.
      */
     Optional<Account> getOrCreateAccount(String identifier);
+
+    /**
+     * Returns the {@link Set} of all stored {@link UniqueAccount} objects. 
+     *
+     * @return The {@link Set} of all stored {@link UniqueAccount} objects.
+     */
+    Set<UniqueAccount> getAllUniqueAccounts();
+
+    /**
+     * Returns the {@link Set} of all stored {@link VirtualAccount} objects.
+     *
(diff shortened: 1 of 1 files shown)


DenWav:
This looks good, thanks for the PR and for contributing to Sponge. :smile:

However, I think it's important to bring up the conversation that was mentioned in IRC, so the Sponge developers can address whether they are relevant and you can make whatever changes are needed.

1. This wasn't directly brought up in the discussions, but based on what was said I think the JavaDocs should be updated to reflect the relevant failure states which may occur depending on the implementation. Having standard failure responses is generally better than leaving it up to the implementation, that way users of the API can handle all failures in a more general manner (and more appropriately than just `try { ... } catch (Exception e) {}`).
2. There may be some distinction between an empty set (no accounts) and a failure state (either the implementation can't get all accounts because they are dynamically allocated, or there was a database error, or who knows what other reason). Depending on the severity, wrapping the `Set` in an `Optional` and/or throwing relevant exceptions may be a better alternative.
3. Being an Economy service, it's very likely some plugins implementing this interface will require database interactions to actually fully implement it. In this case, it may make more sense to return some type of `Future` instead of the bare `Set` (or `Optional` if the above point were used). `Future`s also provide set methods of handling failure states which would address point two as well. If this is the case, then this actually also applies to both `hasAccount` methods and both `getOrCreateAccount` methods, however neither of those cases are relevant to this particular PR.

Zidane:
Closing as I've manually put this into a branch that will be merged for 1.13.

Thanks a lot for the work and help!

ItsDoot:
@Zidane Looks to be missing from the new `api-8` branch now

PR #1517, opened 2017-03-11
Return created type from Archetype#apply, Fixes #1423

### LINKED ISSUE #1423: EntityArchetype's apply() does not return a reference
Though I know that it is a limitation of extending `Archetype<EntitySnapshot>`, `EntityArchetype`'s `apply()` method does not return a reference to the created `Entity`, and it is then impossible to actually know which one was spawned. I am then forced to use `toSnapshot(location).restore()`, which is bad for a couple of reasons. First of all, it creates a useless object along the way - the snapshot is never used for anything but creating an entity. Second, it kind of abuses the point of snapshots - they're supposed to be a snapshot of live data, and this represents an entity which never existed at all.

Deamon5550:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1246)

A small change to allow a plugin to access the newly created type from applying an archetype.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/tileentity/TileEntityArchetype.java
+++ b/src/main/java/org/spongepowered/api/block/tileentity/TileEntityArchetype.java
@@ -42,7 +42,7 @@ import org.spongepowered.api.world.World;
  * Represents the data of a {@link TileEntity} which does not exist in the world
  * and may be used to create new {@link TileEntity}s with the same data.
  */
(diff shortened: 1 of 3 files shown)


liach:
Small but yet breaking...

Deamon5550:
@liach Yes, thats why this is bleeding and not stable.

liach (approved):

PR #1524, opened 2017-03-27
Add CompletableFutures to PermissionService

lucko:
This PR attempts to address some of the issues raised in #1411.

There were a number of conversations that took place back in October/November about what could be done to the Permissions API to better reflect what permission plugins were doing.

It was a while ago now, but as far as I can recall, the following was brought up & agreed on.

1.  The API was lacking events. People really wanted to be able to listen for changes to permissions, and react to those changes in their own plugins.
2.  The API didn't properly address the situations where data existed, but wasn't currently loaded in memory.
3. There was no way to retrieve data about how inheritance should be done for a given SubjectCollection, specifically regarding whether transient data was to be given priority.

This PR addresses no. 2 above, but depending on the success of this, I may also attempt to address the others over time.

#### What's the problem?
For the purposes of explaining, assume the following setup:
* Permissions plugin is storing data in MySQL
* MySQL server is perhaps half way across the world, or hosted on some crappy machine somewhere else. (as most shared hosts tend to.)
* User has a fairly big server, lots of permissions data being stored.

Any reasonable permissions implementation wouldn't have *all* data loaded at all times. Most just load in group data & info about online players.

The issue comes when a plugin comes along, and wants to, for example, add a permission to an offline user.

* Command gets executed on the main server thread.
* Plugin goes to the PermissionService and requests the Subject for the user, or just uses the `User` instance, assuming everything is loaded and ready.
* The Subject isn't actually loaded, and the PermissionService has to query it's backend for the Subjects data.
* The MySQL query to load the subjects data takes 700ms, and the server thread has to pause for ~15 ticks as the data gets loaded. 👎 

#### The fix
It was pretty much agreed that the fix for this was to use `CompletableFuture`s where necessary, and clearly distinguish between calls which should return quickly, and ones which may take a while. 

The PR makes the following changes:

* Changes method signature to make the difference between **load** and **get** calls very clear. The former will return a Future, and the latter will return an Optional, depending of if the Subject is cached already.
* Methods in SubjectData which mutate state return futures to indicate when the change has been persisted/applied.
* Calls to get ALL subjects in a collection are removed. It is quite reasonable that over time, servers could accumulate hundreds of thousands of unique player logins, and therefore hundreds and thousands of Subjects in one collection. Loading them all at the same time is a bad idea.
* Methods to apply a change to a number of Subjects is provided, to allow users to still modify or collect data from an entire collection, but allow the implementation do execute the action (possibly) in parallel, and allow the unused instances to be GCed quickly. 
* Parent related methods in Subject and SubjectData refer to a `SubjectReference`, and not a complete Subject. This removes the requirement that all "parents" of a Subject have to be loaded when the child is.

Additionally, the PR adds two other small things, which were also covered in zmls PR. 
* "Identifier Validity Predicates", which indicate whether an identifier is allowed for a Subject/SubjectCollection. (think user subjects, only UUIDs should be allowed, etc.)
* "Friendly Identifiers" - again, think user subjects, in this case, they would return the username.

That's pretty much it. I'm happy to PR the SpongeCommon side of this when/if it gets the go ahead.


bloodmc:
@zml2008 please review

lucko:
Also, @djxy any thoughts on this?

ryantheleach:
I'm not going to claim to have enough knowledge on what type of future we should use for this, but it looks pretty good to me.

One thing I'm not sure of, API wise, is 

>Parent related methods in Subject and SubjectData refer to a SubjectReference, and not a complete Subject. This removes the requirement that all "parents" of a Subject have to be loaded when the child is.

Could do with a little more explanation on how this is different from a complete subject vs an identifier.

Should the old methods be deprecated for a release or 2?

lucko:
> Should the old methods be deprecated for a release or 2?

Could say the same thing about all of the Future methods too. I know it's a huge backwards incompatible change, but there's not really a clean way to make these changes and keep the old methods there too. 

You obviously can't have two methods with the same name + parameters, but with different return types. To get around this, you can just pick other method names, but it results in a confusing API in the long run, because you have to pick slightly twisted method names.

For example, `addParent` would have to become something like `giveParent` or `applyParent`. Doesn't really make much sense IMO.

The "important", or most frequently used methods (hasPermission & getPermissionValue) have no changes, so most plugins won't have any issues with this PR. It only affects people trying to retrieve specific subjects, or modifying SubjectData.

> Could do with a little more explanation on how this is different from a complete subject vs an identifier.

Identifier: Just a String of the Subjects identifier   
Reference: An object containing the subject identifier, the name of the containing collection, and a method to resolve to a Subject (load all of the data)   
Subject: The identifier, plus all data data about the Subject loaded.

> I'm not going to claim to have enough knowledge on what type of future we should use for this, but it looks pretty good to me.

Yeah, initially when this was being discussed, I was kinda keen to keep it to just the plain Future interface, however, zml wanted to use CompletableFutures, and having used the class since then, I agree with him on that one.

src/main/java/org/spongepowered/api/service/permission/Subject.java
```diff
@@ -80,6 +80,17 @@
 public interface Subject extends Contextual {
 
     /**
+     * Returns the friendly identifier associated with this Subject.
```
ST-DDT: What is considered friendly? Human readable? A player's name? An identifier stripped of all special characters? Log/Console friendly?

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -42,54 +58,197 @@
...
+    CompletableFuture<Subject> loadSubject(String identifier);
 
     /**
-     * Returns whether there is any data stored for the given subject. The
-     * return value of this function does not influence whether or not the
-     * result from {@link #get(String)} should be queried.
+     * Returns a subject with the given identifier, if the subject is already
+     * loaded within this collection.
+     *
+     * <p>It is important to note that a subject collection with the given
```
ST-DDT: AFAICT this method returns subjects, why is it important to note, that a subject collection might be present or not?

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -42,54 +58,197 @@
...
+        Preconditions.checkNotNull(action, "action");
+        Preconditions.checkNotNull(identifiers, "identifiers");
+        return CompletableFuture.runAsync(() -> {
+            for (String id : identifiers) {
+                try {
+                    Subject s = loadSubject(id).get();
+                    action.accept(s);
+                    suggestUnload(s.getIdentifier());
+                } catch (Exception e) {
+                    e.printStackTrace();
```
ST-DDT: I'm not sure whether it is wise to just print the stack to the console. It would probably be better to log to slf4j or a Logging framework or at least cancel the execution. I would be fine if you keep a reference to the first exception that happened and add the others as suppressed and after the invocation propagate that exception up.

ST-DDT (changes requested):

ST-DDT:
With the change to `SubjectReference` parents it is no longer guaranteed that the complete `SubjectData` hierarchy is loaded, How can we ensure that `hasPermission()` returns without loading/waiting for the missing parent subjects?

src/main/java/org/spongepowered/api/service/permission/Subject.java
lucko: Yes, player name. Will add that to the JDs.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
lucko: My bad, that should be `subject`, not `subject reference`.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
lucko: Yeah, I hadn't considered that. Is it worth just not catching the exception? I guess if it's going to throw for the first Subject, it will probably throw for all of them.

Idk, what do you think is best?

lucko:
That's down to the implementation to handle. The hierarchy should be loaded, yes, as you say, to allow for quick getPermissionValue checks. 

However, it might be the case that implementations flatten these values down somehow when the Subject is loaded, meaning actual instances of the parent Subjects are not needed. Switching to SubjectReference here allows that flexibility.

ST-DDT:
Oh yes. You are right.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
ST-DDT: That is what i would expect. What happens if the exception is not caught will it make it to the logs even if it is not explicitly handled?

lucko: Yes, it would cause the Future to `#completeExceptionally`.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completeExceptionally-java.lang.Throwable-

ST-DDT:
@lucko If you don't instantly squash your commits it is easier to see the changes between each version.
Squaschin should usually only be done if the commit history is too long due to a lot of small changes or right before a merge.

ST-DDT (approved):

zml2008 (commented):
Looks like a good impovement. The only change I think could help clean things up is making the SubjectData instances immutable and have references -- that's more like how PEX handles things and it means that you don't have to deal with a whole ton of CompletableFutures if you want to do multiple changes to a single subject.

PEX's SubjectDataReference implementation: https://github.com/PEXPlugins/PermissionsEx/blob/master/permissionsex-core/src/main/java/ninja/leaping/permissionsex/data/SubjectDataReference.java

lucko:
Seems like a pretty drastic design change. I don't really like it as an idea. It means that whenever a change is applied, the implementation has to compare the two and work out what changes were actually made, so those changes can be pushed to the file/SQL. 

Also a pretty huge risk of creating race conditions.

```
plugin1 requests copy of data
plugin2 requests copy of data

both plugins make their changes, assuming everything went ok.

plugin1 "sets" their copy back.
plugin2 "sets" their copy back.

plugin1's changes are completely lost, as they have been overridden by plugin2.
```

I think this sort of condition would become fairly common if this were to be implemented.

I see the PEX implementation uses a function based update, which avoids this. However, it makes it hard to change behaviour as a result of success return values, and then do things based upon this. The function would also still need to be posted via a Future, and you may want to return values as a result. (if it was a success or not.) The function can only return one value, so this would get pretty messy.

> means that you don't have to deal with a whole ton of CompletableFutures if you want to do multiple changes to a single subject.

Really don't think that's an issue at all. See `CompletableFuture#allOf`.

ryantheleach:
It sounds like Zml wants some way of performing a transaction of many changes.

lucko:
Yes, I understand exactly what's being asked for, but just trying to (perhaps poorly) point out that it makes the API significantly more difficult & messy to use, for no significant gain in functionality. 

All of the methods in SubjectData return booleans to let you know if the change being applied was successful or not. With a transactional style API (like the one used in PEX), it is not possible to cleanly return these values out of the future.

For example:   
https://gist.github.com/lucko/188b83acd118c9b9f69cbf80b55f5629

If the only reason for making the change is
> means that you don't have to deal with a whole ton of CompletableFutures if you want to do multiple changes to a single subject.

Then I don't really think it applies. You don't have to deal with a "whole ton" of Futures at all.
```java
return CompletableFuture.allOf(
        data.setPermission(SubjectData.GLOBAL_CONTEXT, "some.perm", Tristate.TRUE),
        data.setPermission(SubjectData.GLOBAL_CONTEXT, "some.perm.thing", Tristate.TRUE),
        data.setPermission(SubjectData.GLOBAL_CONTEXT, "some.other.perm", Tristate.TRUE),
        data.setPermission(SubjectData.GLOBAL_CONTEXT, "some.thing", Tristate.TRUE)
);
```

The `#allOf` method effectively IS your transaction, and returns when all of the changes have been made. If you want to use the results of the methods, then you can just wrap the calls in your own future or scheduler task, and use `#join`.

ST-DDT:
How about adding a method using a map to do multiple changes. That way you can update your database in one go instead of x atomic operations? If you set permissions you usually do it in bulk.

Either via `setPermissions(Set<Context> context, Map<String,Tristate> permissions)`
Or via `setPermissions(Map<Set<Context>, Map<String,Tristate>> permissions)`

lucko:
That would only really be a convenience method for the end user, and I'm not sure how useful it would actually be, to be honest.

If you want to bulk set a whole load of permissions, they're more likely to be in a list form, in the form Map<String, Boolean>, or just something you write out in your code. Basically, I don't think padding these interfaces out with loads of alias methods would be beneficial.

You could go on forever with it, on the basis of "what if" people decide to store their data in this way.
```java
setPermission(Set<Set<Context>> contexts, String perm, Tristate ts);
setPermission(Set<Set<Context>> contexts, Map<String, Tristate> map);
setPermission(Set<Set<Context>> contexts, Map<String, Boolean> map);
removeAllPermissions(Map<Set<Context>, Set<String>> map)
... and so on.
```

Implementations will still need to be able to process multiple changes, so adding more methods really doesn't make a difference. You can't rely on people to actually the "bulk change" methods. I personally just use a buffer, which only commits the changes to the file/db after a period of inactivity.

It **doesn't** make the interface easier to use/understand, and it makes no difference to how easy it is to implement.

zml2008:
>  it is not possible to cleanly return these values out of the future.

It still is -- the future can return the old ImmutableSubjectData and the new ImmutableSubjectData, and can contain an exception if an exception was thrown, and the changes with that system will only result in one scheduler task and one future object being created.

In PEX's case, I've been moving towards only exposing the update method for subjects, rather than exposing get/set directly which removes the room for error there.

lucko:
> the future can return the old ImmutableSubjectData and the new ImmutableSubjectData, and can contain an exception if an exception was thrown

And that's clean? What if you want to return your own result? Functions can only return one object. (I pointed this out in the examples I gave, which I guess you must've missed.)

> and the changes with that system will only result in one scheduler task and one future object being created.

If the changes are only memory based, no threads/tasks are involved. The methods can just return completed futures. Instance creation is negligible, and would apply either way. Even if threads are needed, ForkJoinPool is very fast and designed specifically to handle this sort of thing.


The PR is meant to fix the poor design choices made with regard to i/o calls on the main thread. A transactional rewrite of these interfaces has nothing to do with this. And, frankly, more complexity here is the last thing needed. It's confusing enough to use / implement as it is.

lucko:
I've added an optional 'transaction' method to SubjectData, which effectively implements @zml2008 's suggestion, but retains the previous design of the interface. (no added complexity for those implementing the service)

To achieve this, a `SubjectDataModel` interface is added, (not sure on the name) which just contains the mutation methods from `SubjectData`, except returning a raw `boolean` instead of `CompletableFuture<Boolean>`. A default implementation of the methods are also there, implemented using `ProxySubjectDataModel`, to simply proxy the calls back to the original subject.

This means the "transactional" style API is there, but there is no requirement that the implementations of PermissionService do anything special to implement it. Those who do wish to use it can though. 

I'm not too sure about the new method/class names, so any better suggestions for those would be appreciated.

@zml2008 can you please re-review.

src/main/java/org/spongepowered/api/service/permission/Subject.java
ryantheleach: Are they expected to be namespaced at all?

ryantheleach:
With the blocking methods and many complete-able futures vs 1 complete-able future argument, is there any difference in the expected time before the result returns?

One of the concerns I had when there wasn't a transactional interface, was that it may need to block for each permission being changed, rather then batching and having the whole lot implemented, or fail, with results for each change.

src/main/java/org/spongepowered/api/service/permission/ProxySubjectDataModel.java
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param delegate The delegate SubjectData instance
+     */
+    public ProxySubjectDataModel(SubjectData delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public boolean setPermission(Set<Context> contexts, String permission, Tristate value) {
+        return delegate.setPermission(contexts, permission, value).join();
```
kashike: `this.`

src/main/java/org/spongepowered/api/service/permission/ProxySubjectDataModel.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public boolean clearPermissions() {
+        return delegate.clearPermissions().join();
```
kashike: `this.`

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -42,54 +58,193 @@
...
+     * @param action The action to perform on each subject
+     * @param identifiers a set of identifiers to apply the action to
+     * @return A future which will complete when the operation has finished
+     */
+    default CompletableFuture<Void> applyToAll(Consumer<Subject> action, Set<String> identifiers) {
+        Preconditions.checkNotNull(action, "action");
+        Preconditions.checkNotNull(identifiers, "identifiers");
+        return CompletableFuture.runAsync(() -> {
+            for (String id : identifiers) {
+                Subject s = loadSubject(id).join();
```
kashike: Please use something more than `s`.

kashike (changes requested):

lucko:
> With the blocking methods and many complete-able futures vs 1 complete-able future argument, is there any difference in the expected time before the result returns?

> One of the concerns I had when there wasn't a transactional interface, was that it may need to block for each permission being changed, rather then batching and having the whole lot implemented, or fail, with results for each change.

Yes, the default implementation will return at exactly the same speed. Implementations can override this method if they wish to provide the transactional style "properly". There's really no way around it. If implementations want to provide perfect atomic writes, an SQL query might need to run each time one of those setPermission methods are called. They **need** futures, and no transactional API would ever solve that issue of "need to block for each permission being changed, rather then batching".

src/main/java/org/spongepowered/api/service/permission/ProxySubjectDataModel.java
lucko: Fixed

src/main/java/org/spongepowered/api/service/permission/ProxySubjectDataModel.java
lucko: Fixed

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
lucko: Fixed

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -111,18 +112,30 @@
...
+     * Searches for and returns all subjects that have this permission set.
+     *
+     * <p>If you want to know to which role-templates this permission is
+     * assigned use {@link PermissionService#SUBJECTS_ROLE_TEMPLATE}.
+     *
+     * @param type The subject type identifier to use
+     * @return An immutable map of subjects that have this permission set
+     * @see SubjectCollection#getAllWithPermission(String)
+     */
+    CompletableFuture<Map<SubjectReference, Boolean>> findAssignedSubjects(String type);
```
Zidane: This isn't the goal of this PR but should you be willing to do further permissions work @lucko then maybe look into changing this parameter to take something akin to a catalog? Having arbitrary string values doesn't make for good API as without reading the docs I'd have no idea what to pass in here.

lucko: Perhaps I'm mistaken here, but as far as I know, that "type" parameter simply refers to the id of the SubjectCollection to be queried.

The API has no real reference to "subject types", other than a few static collection names in the PermissionService class, which suggest standard identifiers for collections which store x type of subject.

I'm not sure if a catalog would be appropriate for these, as they are only default or "standard" names for those collections.
https://github.com/lucko/SpongeAPI/blob/feature/perms-futures/src/main/java/org/spongepowered/api/service/permission/PermissionService.java#L45-L85

It's quite normal that a permission service would hold more than just the standard collections.

ST-DDT: There is nothing that prevents you from adding a register method that allows adding new catalog entries. But somehow you are right as we use that parameter in the previous/current version in a lot of places as well including the prototype method just a few lines above.

ST-DDT:
>> With the blocking methods and many complete-able futures vs 1 complete-able future argument, is there any difference in the expected time before the result returns?

>> One of the concerns I had when there wasn't a transactional interface, was that it may need to block for each permission being changed, rather then batching and having the whole lot implemented, or fail, with results for each change.

> Yes, the default implementation will return at exactly the same speed. Implementations can override this method if they wish to provide the transactional style "properly". There's really no way around it. If implementations want to provide perfect atomic writes, an SQL query might need to run each time one of those setPermission methods are called. They need futures, and no transactional API would ever solve that issue of "need to block for each permission being changed, rather then batching".

Then wouldn't it be better if we remove the boolean result from the `SubjectDataModel` interface? Because which reason would cause a transactional `setPermission()` call to fail, if executed on the perm thread? If we remove the boolean return value we can execute the changes in `O(1)` instead of `O(n)`. If an exception happens during the function we can rollback the changes completely like a single transaction. If you want to execute atomic operations, then why don't you use the atomic methods (with the CompletableFutures) in the first place?

lucko:
> Then wouldn't it be better if we remove the boolean result from the SubjectDataModel interface? Because which reason would cause a transactional setPermission() call to fail, if executed on the perm thread?

The result indicates if the Subject had the permission already. It *should* ideally be atomic, but in most implementations it won't be. What do you mean by the "perm thread"? There's no special thread running just for the permission service.

>  If we remove the boolean return value we can execute the changes in O(1) instead of O(n).
 Then what's the point in the transaction? You might as well just use the future methods and ignore the result.

> If an exception happens during the function we can rollback the changes completely like a single transaction.

That's not what the booleans being returned represent.

> If you want to execute atomic operations, then why don't you use the atomic methods (with the CompletableFutures) in the first place?

Not what I meant by atomic. Transactions would imply that the entire subject data becomes locked whilst the transaction takes place. Totally unnecessary imo and really hard to implement if you want everything atomic right back to the datastore.

bloodmc:
@zml2008 please review so we can get this merged.

lucko:
Since it's been a couple of weeks since there was any activity here, to attempt to summarise what's happening: 

As far as I know, the initial changes are all ok, regarding adding futures to blocking methods. There weren't any objections to doing that.

However, since then, two other things were brought up:

One is **transactions**. There are various discussions about that change in particular above. My opinions are that it's a pretty unnecessary change & makes the API harder and more confusing to use/implement. 

A kind of "compromise" implementation was added in, which I'm not too happy with. I don't think it really adds anything. @ST-DDT made a good point about how the default implementation of it doesn't really work well, so, before this gets merged, that whole thing needs to be sorted.

My personal opinion is that the whole idea of transactions should be scrapped here. I don't really think it adds anything. (I guess just see above. I've explained it already.)

The second thing was about **Catalog Types**. The suggestion was that SubjectCollection identifiers should be replaced with those. I'm not all that familiar with how that system works, but as far as I know they're more for constants, which identifiers are definitely not. Need some advice RE that.

zml2008:
I don't think trying to mash transactions and not-transactions together would work well -- we have to go either one way or the other (but I haven't looked at transactions since one of the first proposals here)
yeah, catalog types do not fit subject collections at all -- they should remain separate things.
I'm in the middle of exams and moving right now, so it might take me a bit to get to a more detailed review, but hopefully I'll be able to put together a response by the end of this week.

bloodmc:
@zml2008 Please finalize the review for this PR.

lucko:
I've removed the transaction stuff which was added previously.

lucko:
hello @zml2008

lucko:
@zml2008 if you're not working on the project anymore, could you try to arrange for another member of the team to look into this. 

I'm happy to make adjustments and implement the SpongeCommon aspects, all I'd like is a reply.

zml2008 (approved):
I'm ready to kill PEX, let's go

kashike (approved):

kashike:
This will most likely be pulled on, or after, this coming Sunday.

lucko:
Awesome, glad to finally see some movement. 

There's a couple of Javadocs I've just noticed which still need to be tidied up. I will try to get that done before Saturday. 

I'll also give implementing the common side a bash. I started a branch a couple of months ago, but I think those commits will conflict with changes since then. 

It's mostly just a case of adjusting for changes in method signature. (zml, it should be relatively painless for you to make PEX compatible. It'd be a shame to loose another perms plugin option 😢 )

lucko:
I've made a couple of final changes, and implemented the changes to the `SpongePermissionService` in SpongeCommon.

https://github.com/SpongePowered/SpongeCommon/pull/1445

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -103,70 +122,113 @@
...
     /**
-     * Gets all subjects that have this permission set.
+     * Gets the owning plugin the permission belongs to.
+     *
+     * <p>Will return an empty optional for descriptions which have been
+     * automatically generated.</p>
+     *
+     * @return The owning plugin the permission belongs to
+     */
+    Optional<PluginContainer> getOwner();
```
Zidane, line 146: Just noticed this.

Can you elaborate on how this can be Optional? No matter what is executing code, a PluginContainer is involved (even for the base game).

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
```diff
@@ -29,82 +29,238 @@
...
+ * run a callback once the method has completed. Alternatively,
+ * {@link CompletableFuture#join()} can be used, but care should be taken to
+ * avoid blocking the server thread.</p>
  */
 public interface PermissionService extends ContextualService<Subject> {
 
+    /**
+     * The standard identifier for the collection which stores users.
+     */
     String SUBJECTS_USER = "user";
```
Zidane, line 69: Not necessary for this PR but did we discuss making these into a Catalog? Its been a while so that discussion, if happened, is lost to me.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
```diff
@@ -24,86 +24,315 @@
...
+
+    /**
+     * Indicate that a certain subject may be unloaded by the implementation.
+     *
+     * <p>This is only a hint to the implementation, and does not guarantee that
+     * the subject will be unloaded.</p>
+     *
+     * @param identifier The subject to be unloaded
+     */
+    void suggestUnload(String identifier);
```
Zidane, line 336: Would it be of any value to have a return of some Future that indicates if it succeeded?

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
lucko: PermissionDescriptions can be automatically generated by the permission plugin.

As all checks are forwarded and handled by the perms plugin, they can maintain their own registry of "known" permission. (I do this in LuckPerms for tab completion, etc)

I did consider that in this case, the plugin container for the permission plugin could be used, however, the permission plugin wouldn't really be the "owner" of the permission. Also, using an Optional allows you to distinguish between PermissionDescriptions where were automatically generated, and those which were registered and "official".

The builder for PermissionDescriptions will requires a plugin instance. That remains the same.

Hopefully that logic makes sense, I'm happy to change it though if you disagree. 😄

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
lucko: Yeah, it was mentioned.

https://github.com/SpongePowered/SpongeAPI/pull/1524#discussion_r110904194
https://github.com/SpongePowered/SpongeAPI/pull/1524#issuecomment-297090378

> yeah, catalog types do not fit subject collections at all -- they should remain separate things.

I think the general agreement was that they weren't appropriate, but again, I'm sure you know more than I about where they should and should not be used.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
Faithcaio: maybe you should document this in the API then, why the container may be missing

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
lucko: Possibly, but I don't see when you'd ever want to wait for an unload to occur.

As the javadoc says, it's more just a "friendly nudge" to the implementation.

Personally, in LuckPerms, the calls are just ignored. I handle caching myself, and unload after a period of inactivity.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
Zidane: Hmmmm you make a good point. Whereas the perm plugin could return it, it would be a lie in certain cases. Keep it as it is.

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
Zidane: @lucko 

I see, that has refreshed my memory. In this case, I'd argue for being cataloged simply because it makes for better API visibility (as it stands, one has to know to go to PermissionService for these constants). As well as it gives a registry a plugin could perform a lookup in for identifiers being used by others (and not needing to have the dependency on that plugin for utilizing that identifier).

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
Zidane: Fair point.

Personally I'd say add it simply because it doesn't hurt to have it. I'm not seeing a consequence here (even if it remains, largely, un-used).

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
Zidane: I agree with @Faithcaio 's suggestion of changing the docs here just so people understand what Optional means in this case.

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
lucko: The consequence of the method is intentionally left open. (to make it flexible for implementations, this is really important in the PS API.)

Some possible results:

* Implementation just ignores the call (returns false??)
* Implementation always listens to the call, and unloads the subject no matter what. (returns true)
* Implementation normally listens to the call, but for a certain subject, doesn't want to unload, say, if the related user was online. (returns false?)

Lots of different possible outcomes, with very vague information being presented to the user, which I would argue, has no use, and couldn't really be depended on.

Idk, like:
```java
SubjectCollection collection;
Subject subj;

boolean ret = collection.suggestUnload(subj.getIdentifier());

// then do what with the ret? how would it ever be useful!?
// even if ret was true/false, what does that actually mean?
```

I think that by adding a return, you might encourage users to act upon the result, which would be totally arbitrary.

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
lucko: I'm not sure on the purpose of catalogs, but I understand they're similar to enums?

In which case, it's important to know that the "standard" identifiers expressed as static vars in the PermissionService class are by no means a definitive list of collections. Merely, they are the ones which are used by Sponge.

Other plugins are free to "create" or load their own SubjectCollections for their own purposes.

Also, not sure how they keying system works, but each collection would still need a String identifier.

As long as it's possible to **easily** register and retrieve new catalog entries at runtime, I guess this would be ok, but I still don't really see the benefit. (loaded collections are already exposed as a map)

(it also breaks away from the same method signature for loading Subjects, currently loading a SubjectCollection & Subject is a very similar process, and the methods are named and behave in similar ways)

src/main/java/org/spongepowered/api/service/permission/SubjectCollection.java
Zidane: I see. Well I guess the question then is simply: Is it ever imperative to know if the implementation actually did it? I'll be honest and state that nothing is coming to mind but I wanted to be sure to throw up the point in-case something was being forgotten. Should you have no reason either, ignore this situation and stay the course.

lucko: > Is it ever imperative to know if the implementation actually did it?

I don't think so. I'd extend that to say it would never even be useful, let alone imperative.

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
Zidane: After further thought, lets keep this to a future PR. I have some plans for some GameRegistry refactoring as registering additional catalogs strictly from the API isn't handled well currently.

As far as doing this in-general, dynamic identifiers make for good catalog types if and only if it is at all useful for plugin B to use plugin A's identifiers. If not then it doesn't make for a good candidate.

lucko: > As far as doing this in-general, dynamic identifiers make for good catalog types if and only if it is at all useful for plugin B to use plugin A's identifiers. If not then it doesn't make for a good candidate.

It might be.

I'd say that plugin A using plugin B's collections is very rare. I don't even know of any plugins which create their own collections. (at this point, might become more common as Sponge gets more plugins, idk)

However, plugin A using Sponge's collections is very common. (it happens all the time.)

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
lucko: It already did. :wink:

```java
    /**
     * Gets the owning plugin the permission belongs to.
     *
     * <p>Will return an empty optional for descriptions which have been
     * automatically generated.</p>
     *
     * @return The owning plugin the permission belongs to
     */
    Optional<PluginContainer> getOwner();
```

lucko:
Just fixed some other checkstyle issues with the javadocs. No other changes made.

src/main/java/org/spongepowered/api/service/permission/PermissionService.java
ryantheleach: ~~@lucko It probably would be appropriate for them to be catalog typed.~~

~~CatalogTypes are essentially extensible enums, that are usually registered once per game? server? during the startup lifecycle events, and can not be unregistered or changed apart from adding new ones.~~

~~They seem appropriate for something that has a few Sponge / Minecraft specific static things, but more can be added on a per mod/plugin basis.~~

~~I suspect the main reason why there isn't more cross plugin compatibility is that it's generally been hard to do correctly without having dependencies on plugins implementing services.~~

~~Catalog types are a way to facilitate that based on a mod prefix and id. for example if luck perms didn't support having timed permissions on a subectType a catalogued type would allow people to retrieve them from Sponge using a string. It would then be up to the admin to understand what the subject type would mean whilst the plugin handling the timed permissions could have no idea.~~

Nevermind, rereviewed, I don't know enough about PermissionCollections to know if it's appropriate.

lucko:
Any update on this @Zidane ?

PR #1531, opened 2017-04-06
Refactor Data Registration

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1287) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/tree/refactor/custom-data)
This PR intends to do a lot for Data Registration. Ideally the registration transition for developers should be easy, as the object is now a consumable registration object that is registered upon creation.

#### *Wait, so what's going on now? I thought Data API was solid? What about #1198?*

Well, to answer that, I have to disclose an implementation detail of how custom data is serialized. We all already know that custom data can be registered with Sponge's Data API, and it's served some plugins well, but the issue with it all was that the registered custom data would serialize in a lazy way, suffering from some of the same issues that [Serializable](http://stackoverflow.com/questions/5305473/how-to-deal-with-a-java-serialized-object-whose-package-changed) suffers: Class and package renaming.

#### *Huh? Package renaming? I thought it just stored the custom data itself?*

Well, give me a moment! I said I need to disclose the implementation detail, right? Right. So, the strategy for serializing custom data in a way where it'd use the [class name as an id](https://github.com/SpongePowered/SpongeCommon/blob/6220b444dc665bd159174bb66612735a3a5dbf9b/src/main/java/org/spongepowered/common/data/util/DataUtil.java#L106-L175), resulting in NBT looking something akin to this:
```
creeper {
  "Sponge": {
    "CustomManipulators": [
      - {
        "DataClass": "com.gabizou.simpleplugin.data.MyCustomData"
        "ManipulatorData": {
          "foo":1
          "bar": "baz"
        }
      }
    ]
  }
}

```

#### *Ok... what's wrong with that? It deserializes fine, doesn't it?*

Well, yes and no. Remember how I mentioned it uses the class name as an id? If I went and refactored my plugin where `MyCustomData` is now located in package: `com.gabizou.simpleplugin.data.mutable.MyCustomData`, Sponge's system would ultimately fail because the class `com.gabizou.simpleplugin.data.MyCustomData` cannot be found. And currently, the system is designed in such a way that if a class cannot be found, well, the data is dropped from the `DataHolder` and then lost forever.

#### *Oh.... that does sound bad... But can't you use a string id or something? Why go through and change everything?*

You're somewhat on the right track. We will be using a developer provided string id, but, because there's also the monstrosity of [the registration method](https://github.com/SpongePowered/SpongeAPI/blob/09d03bf4d8d40a660273b9712df36978bee60e19/src/main/java/org/spongepowered/api/data/DataManager.java#L130-L145) for custom data, I elected to use a builder immediate registration pattern, as you may have seen in other cases in the API. So, instead of using the following:
```java

@Listener
public void onInit(GameInitializationEvent event) {
  Sponge.getDataManager().register(MyCustomData.class, ImmutableCustomData.class, new CustomDataBuilder());
}
```
It's going to be like so:
```java
@Listener
public void onInit(GameInitializationEvent event) {
  DataRegistration.<MyCustomData, ImmutableCustomData>builder()
      .setDataClass(MyCustomData.class)
      .setImmutableDataClass(ImmutableCustomData.class)
      .setBuilder(new CustomDataBuilder())
      .setManipulatorId("my_custom")
      .buildAndRegister(this);
}
```

#### *But that takes more lines! It feels clunkier! I don't like it!*

What about it bothers you? The line additions? It's a builder pattern, so if anything happens where you need to change a specific line, or an exception is thrown for a specific method call, you know which line it is off the bat. The extra casting? Unfortunately, due to the nature of how Java inflects generics, it's not possible to ignore the generic types when calling `builder()`, it is possible if it was in a non-generically typed class. 

The requirement for the `String` id is that it is now a constant that can be utilized by plugin developers to safely and carefully identify their data.

#### ***Sigh..** I guess I can deal with it... What else does it change?*

Nothing much else. There's some additions for legacy id's that can be used, such as previously used class names. Likewise, the implementation will automatically detect a plugin using the older methods such that we'll be able to determine and generate a new string id for the custom data and assign it within sponge itself. Of course, this will require developers to upgrade at some point, but until then, there will be warning system in place where the server will emit a warning for plugin developers to update to the new system. It will look something like so:
```
/************************************************************************************/
/*                             Legacy Data Registration                             */
/************************************************************************************/
/* Sponge's Custom Data Registration is changing! Plugin developers are strongly    */
/* advised to update their data registrations!However, because of this, Sponge is   */
/* automatically applying an upgrade policy for their old registrations that will   */
/* persistuntil an unknown future time. Likely for the next few Minecraft versions. */
/************************************************************************************/
/* Plugin registering custom data: demotest                                         */
/* Custom Data Class: DemoTestData                                                  */
/* Generated Manipulator Id: demotest:demo_test                                     */
/* The generated Id is already registered with Sponge, and is likely advisable for  */
/* the plugin developer to utilizethe id for future registrations.                  */
/************************************************************************************/
```
And note that this will only print out once per data registration. Likewise, the nice thing about this is that **Sponge even goes as far as suggesting a new registration id for the developer**. So if the developer does take note of this id, some method changes and that's all that has to happen!

#### *Well, I guess this isn't to bad then, what about custom data generation? Do I still have to make those custom classes?*
For the mean time, yes. However, this is something that will be easier to change in the very near future, since the data serialization strategy has been updated, it will be easier for plugin developers to update between API 5 and API 6.

*"Come to the dark side, we have cookies"*
*- Unknown*


src/main/java/org/spongepowered/api/data/DataAlreadyRegisteredException.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.annotation.Nullable;
+
+public class DataAlreadyRegisteredException extends DataException {
+
+    @Nullable
+    private final Class<? extends DataManipulator<?, ?>> manipulatorClass;
+    @Nullable
+    private final Class<? extends ImmutableDataManipulator<?, ?>> immutableManipulatorClass;
+    @Nullable
+    private final DataManipulatorBuilder<?, ?> builder;
```
kashike: Getters for these fields?

src/main/java/org/spongepowered/api/data/DataManager.java
```diff
@@ -134,17 +137,34 @@
      * data representation, and mass application of a {@link DataManipulator}
      * to multiple {@link DataHolder}s.
      *
+     * <p>Due to the addition of {@link DataRegistration}</p>
```
kashike: Incomplete

src/main/java/org/spongepowered/api/data/DataRegistrationNotFoundException.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+
+import javax.annotation.Nullable;
+
+public class DataRegistrationNotFoundException extends DataException {
+
+    @Nullable private String registrationQuery;
+    @Nullable private Class<? extends DataManipulator<?, ?>> manipulatorClass;
+    @Nullable private Class<? extends ImmutableDataManipulator<?, ?>> immutableClass;
```
kashike, line 12: Getters?

kashike (changes requested):

src/main/java/org/spongepowered/api/util/TypeTokens.java
```diff
@@ -0,0 +1,485 @@
...
+    public static final TypeToken<Value<WireAttachmentType>> WIRE_ATTACHMENT_TYPE_VALUE_TOKEN = new TypeToken<Value<WireAttachmentType>>() {};
+
+    public static final TypeToken<org.spongepowered.api.data.type.ZombieType> ZOMBIE_TYPE_TOKEN = new TypeToken<org.spongepowered.api.data.type.ZombieType>() {};
+
+    public static final TypeToken<Value<org.spongepowered.api.data.type.ZombieType>> ZOMBIE_TYPE_VALUE_TOKEN = new TypeToken<Value<org.spongepowered.api.data.type.ZombieType>>() {};
+
+    // SORTFIELDS:OFF
+
+    private TypeTokens() {
+        throw new RuntimeException("Lolno");
```
stephan-gh, line 447: We have this in none our classes, so I don't think throwing an exception is needed here.

src/main/java/org/spongepowered/api/util/TypeTokens.java
```diff
@@ -0,0 +1,450 @@
...
+package org.spongepowered.api.util;
+
+import com.flowpowered.math.vector.Vector3d;
+import com.flowpowered.math.vector.Vector3i;
+import com.google.common.reflect.TypeToken;
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.data.meta.PatternLayer;
+import org.spongepowered.api.data.type.*;
```
Zidane, line 34: Tsk

Deamon5550: Do we really need >100 imports for when we really are importing everything in the package.

kinggoesgaming:
kudos to @gabizou  :smile:

PR #1532, opened 2017-04-09
[WIP] Conversation API

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1294) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/310) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1427)

**Note:** This write-up is outdated at the moment. I am getting much closer and will update this soon once I get a more stable API and properly working implementation to show it off. We're getting there my friends :)

So this is something I have been dabbling around with and it is still **work in progress**, but I would love to get some feedback from you while I continue this process. As a whole, I have looked at other similar APIs from the past and other proposals such as @pie-flavor's [Dialogue API](https://github.com/SpongePowered/SpongeAPI/pull/1295) which I especially found a lot of initial inspiration from as it seemed to match with many of Sponge's conventions and be fairly easy to use.

So far I have tested this on SpongeVanilla and it works relatively well so far, outside of a few issues, meeting all of my initial goals and more.

A `Conversation` is simply a form of back and forth between two or more people. With this API, it allows plugin developers to create one of their own, with any amount of `Conversant`s if they so desire. A common use case is to collect information over a plethora of questions, to create something such as a new world. This can replace a long command with a lot of different arguments and flags. Each argument with a conversation would simply be a new `Question`. 

A `Conversation` is an active conversation or one that has recently ended and is still being manipulated. A `ConversationArchetype` is what these are started from and can easily be built by calling `ConversationArchetype.builder()` with many of the values having a default. 

As plugins handle answers to questions, they can either store the data they collect themselves or store it with data queries within the `DataContainer` present in the `Conversation`, by using `Conversation#getContext`.

One goal of this PR is to allow the use of `GenericArguments` and `CommandElement`s as a whole, as that system is already pretty great and extensive, instead of having a separate idea of different kinds of `Prompt`s. This will also allow me to add support for tab completion easily soon and makes the whole idea of conversations similar to what developers are already used to with commands.

Much of it branches off there and the Javadocs should do it justice until I have some more time to add to this PR description, but if you have questions feel free to ask.

New Goals -

- [ ] Allow other types of `Conversant`s, maybe such as `ConsoleSource` and `RconSource`.

- [x] Possibly keep weak references to `Conversant`s, so if they die or their object is recreated the conversation can remove them. Looking for input on a good solution here.

- [x] Utilize the ability of the `CommandElement` of the current question to allow for tab completion.

Here is a small example I threw together quickly:
```java
final Question two = Question.builder()
            .id("seed")
            .argument(GenericArguments.longNum(Text.of("seed")))
            .prompt(Text.of("Enter desired seed:"))
            .handler((convo, conversant, quest, context) -> {
                Optional<Long> optionalLong = context.getOne("seed");

                if (optionalLong.isPresent()) {
                    convo.getContext().set(DataQuery.of("seed"), optionalLong.get());
                    return QuestionResult.end();
                } else {
                    return QuestionResult.repeat();
                }
            }).build();

        final Question question = Question.builder()
            .id("name")
            .argument(GenericArguments.string(Text.of("name")))
            .prompt(Text.of("Type in the desired world name:"))
            .handler((convo, conversant, quest, context) -> {
                Optional<String> optionalName = context.getOne("name");

                if (optionalName.isPresent()) {
                    convo.getContext().set(DataQuery.of("name"), optionalName.get());
                    return QuestionResult.nextQuestion(two);
                } else {
                    return QuestionResult.repeat();
                }
            }).build();

        ConversationArchetype.builder()
            .id("world-builder")
            .title(Text.of(TextColors.GREEN, "World Creation"))
            .padding(Text.of("-"))
            .startingMessage(Text.of("You will be creating a world within this conversation."))
            .allowCommands(false)
            .catchesOutput(true)
            .exitString("cancel")
            .defaultChatHandler(ExternalChatHandlers.sendOnFinish())
            .firstQuestion(question)
            .endingHandler(((conversation, dataContainer, conversationEndType) -> {
                if (conversationEndType.equals(ConversationEndTypes.FINISHED)) {
                    final String name = dataContainer.getString(DataQuery.of("name")).get();
                    final long seed = dataContainer.getLong(DataQuery.of("seed")).get();

                    conversation.getConversants().forEach(c -> c.sendThroughMessage(Text.of(
                        "Your world will be called: " + name + " and have the seed: " + seed)));

                    // In reality you could ask for all of the arguments for this
                    WorldArchetype.builder()
                        .seed(seed)
                        .build(name.toLowerCase(), name);
                }

                conversation.getConversants().forEach(c -> c.sendThroughMessage(Text.of("Thanks for playing!")));
            })).build().start(this.pluginContainer, player);
```


src/main/java/org/spongepowered/api/Game.java
```diff
@@ -204,4 +205,12 @@
      */
     ChannelRegistrar getChannelRegistrar();
 
+    /**
+     * Gets the {@link ConversationManager} for managing and
+     * handling {@link org.spongepowered.api.conversation.Conversation}s.
```
vectrixdevelops: Import this link.

src/main/java/org/spongepowered/api/conversation/Conversation.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    boolean removeEndingHandler(EndingHandler endingHandler);
+
+    /**
+     * A convenience method to get the id of the conversation archetype.
+     *
+     * @return The id of the conversation archetype
+     */
+    default String getId() {
+        return getArchetype().getId();
```
vectrixdevelops: Add `this.` to keep consistent with others in this interface.

vectrixdevelops (commented):
Keep it up nice work!

src/main/java/org/spongepowered/api/conversation/Conversation.java
ST-DDT: Well I would do it the other wa round. Remove the `this.` from method calls. You only need it for field access.

src/main/java/org/spongepowered/api/conversation/ResponseContext.java
```diff
@@ -0,0 +1,108 @@
...
+    <T> Collection<T> getAll(Text key);
+
+    /**
+     * Gets the value for the given key if the key has only one value.
+     *
+     * @param key the key to get
+     * @param <T> the expected type of the argument
+     * @return the argument
+     */
+    <T> Optional<T> getOne(String key);
```
ST-DDT: What happens if there are more that one value? Does it return absent, cause an exception or return a random/the first value?

parlough: So, I just needed this class to abstract away from `CommandContext` as it has methods I did not really want accessible as they did not make sense in relation to conversations.  The implementation of this is here: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/command/args/CommandContext.java#L94

So if there is more than one value it will return an empty Optional.

The whole command system is useful for conversations and I feel could be less implemented in the API as a whole, but that is outside of the scope of this PR.

I did this, so `GenericArguments` and `CommandElement` could be used so that whole system does not need to be rewritten and is already familiar to users. Doing this also allows me to add support for tab completion once I get around to that.

src/main/java/org/spongepowered/api/Game.java
parlough: Is this standard throughout the API? It's the only place used in this class so I was not sure.

vectrixdevelops: Other links for Javadocs in that class import. So up to you. :P

src/main/java/org/spongepowered/api/conversation/Conversation.java
vectrixdevelops: Very true, that would be better.

src/main/java/org/spongepowered/api/conversation/ExternalChatHandlers.java
```diff
@@ -0,0 +1,123 @@
...
+
+import org.spongepowered.api.text.Text;
+
+import java.util.ArrayDeque;
+import java.util.Queue;
+
+/**
+ * The main external chat handlers you would need.
+ */
+public class ExternalChatHandlers {
```
liach: Make this class final and add private constructor.

src/main/java/org/spongepowered/api/conversation/ExternalChatHandlers.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Creates an external chat handler which lets through all messages so
+     * the {@link Conversant} sees all the messages even within the conversation.
+     *
+     * @return An external chat handler which lets through all messages
+     */
+    public static ExternalChatHandler letThrough() {
+        return new LetThroughExternalChatHandler();
```
liach: Keep instance fields instead of creating new instances each time.

src/main/java/org/spongepowered/api/conversation/ExternalChatHandlers.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public static ExternalChatHandler sendOnFinish() {
+        return new SendOnFinishChatHandler();
+    }
+
+    /**
+     * An external chat handler which lets through all messages so the
+     * {@link Conversant} sees all the messages even within the conversation.
+     */
+    public static class LetThroughExternalChatHandler implements ExternalChatHandler {
```
liach: Make these private.

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndCause.java
```diff
@@ -0,0 +1,79 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets how the conversation ended.
+     *
+     * @return The type of of conversation end
+     */
+    @Nonnull
```
liach: ?!

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCause.java
```diff
@@ -0,0 +1,43 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.cause.conversation.common;
+
+import org.spongepowered.api.event.cause.conversation.ConversationEndCause;
+import org.spongepowered.api.event.cause.conversation.ConversationEndType;
+
+public class AbstractConversationEndCause implements ConversationEndCause {
```
liach: declare abstract

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+
+@SuppressWarnings("unchecked")
+public abstract class AbstractConversationEndCauseBuilder<T extends ConversationEndCause,
+B extends ConversationEndCause.ConversationEndCauseBuilder<T, B>> implements ConversationEndCause.ConversationEndCauseBuilder<T, B> {
+
+    protected AbstractConversationEndCauseBuilder() {
+
+    }
+
+    @Nullable protected ConversationEndType endType;
```
liach: fields should come before the constructor

src/main/java/org/spongepowered/api/event/cause/conversation/common/package-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault package org.spongepowered.api.event.cause.conversation.common;
```
liach: newline

pie-flavor:
Love it. Thanks for finishing this idea; I couldn't finish Dialogue because I had no idea how I was going to implement it and a lot of stuff came up IRL. Hope to see this in the API!

src/main/java/org/spongepowered/api/conversation/ExternalChatHandlers.java
parlough: This is pretty smart, I didn't think of that, but I will only be able to do so for the `letThrough` and `deleteAll` handlers as the other needs to be unique per player.

src/main/java/org/spongepowered/api/Game.java
kashike: Import.

src/main/java/org/spongepowered/api/conversation/AnswerHandler.java
```diff
@@ -0,0 +1,61 @@
...
+ * A functional interface, which a {@link Question} utilizes to handle responses.
+ */
+@FunctionalInterface
+public interface AnswerHandler {
+
+    /**
+     * Handles a {@link Conversant}'s answer to a {@link Question}.
+     *
+     * <p>This is where you should collect all information from individual
+     * {@link Question}s or prompts, to the{@link org.spongepowered.api.data.DataContainer}
```
kashike: Import.

src/main/java/org/spongepowered/api/conversation/AnswerHandler.java
```diff
@@ -0,0 +1,61 @@
...
+     * accessible through {@link Conversation#getContext()} with
+     * {@link org.spongepowered.api.data.DataQuery} or store it yourself.</p>
```
kashike: Import.

src/main/java/org/spongepowered/api/conversation/Conversant.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.spongepowered.api.command.CommandSource;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
+
+/**
+ * Represents someone that can take part in a {@link Conversation}.
+ */
+public interface Conversant extends CommandSource {
```
kashike: This name is weird. Why does it extend `CommandSource`? Something doesn't need to know how to deal with commands to deal with conversations - if anything, extend `MessageReceiver`.

src/main/java/org/spongepowered/api/conversation/Conversation.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    DataContainer getContext();
+
+    /**
+     * An ease of access method for the string which a {@link Conversant} can
+     * use to exit a conversation.
+     *
+     * @return The exit string
+     */
+    default String getExitString() {
```
kashike: `Text`?

src/main/java/org/spongepowered/api/conversation/Conversation.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Adds a {@link Conversant} to the conversation.
+     *
+     * @param conversant The conversant to add to the conversation
+     * @param externalChatHandler The chat handler to give the conversant
+     */
+    void addConversant(Conversant conversant,
+        ExternalChatHandler externalChatHandler);
```
kashike: Wrong formatting.

src/main/java/org/spongepowered/api/conversation/Conversation.java
```diff
@@ -0,0 +1,242 @@
...
+     * @return The handler of the specified {@link Conversant}, if present
+     */
+    Optional<ExternalChatHandler> getChatHandler(Conversant conversant);
+
+    /**
+     * Whether or not this conversation has ended.
+     *
+     * @return Whether or not this conversation has ended
+     */
+    boolean ended();
```
kashike: `hasEnded`

src/main/java/org/spongepowered/api/conversation/ConversationArchetype.java
```diff
@@ -0,0 +1,273 @@
...
+     */
+    Optional<Text> getStartingMessage();
+
+    /**
+     * Gets the string that if the conversant enters will cause them to leave
+     * the conversation.
+     *
+     * @return The string which allows the conversant to leave the conversation
+     */
+    String getExitString();
```
kashike: `Text`?

src/main/java/org/spongepowered/api/conversation/EndingHandler.java
```diff
@@ -0,0 +1,53 @@
...
+     * {@link Conversant#sendThroughMessage(Text)} to send conversants a message
+     * if the conversation is set to catch messages.</p>
+     *
+     * @param conversation The involved conversation
+     * @param context The context which has been added to over the course of
+     *     the conversation
+     * @param endType How the conversation ended
+     */
+    void handle(Conversation conversation, DataContainer context,
+        ConversationEndType endType);
```
kashike: Wrong formatting.

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+        return (B) this;
+    }
+
+    @Override
+    public B reset() {
+        this.endType = null;
+        return (B) this;
+    }
+
+}
```
kashike: Missing newline.

src/main/java/org/spongepowered/api/event/conversation/ConversationCloseEvent.java
```diff
@@ -0,0 +1,64 @@
...
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+
+/**
+ * An event called when a conversation is either ending or has already ended.
+ *
+ * <p>It is highly recommended to utilize the event's sub-interfaces rather
+ * as that is how you will know if the event has ended or not yet.</p>
+ */
+public interface ConversationCloseEvent extends Event {
```
kashike: Should extend a `ConversationEvent`.

src/main/java/org/spongepowered/api/event/conversation/ConversationOpenEvent.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Set;
+
+/**
+ * An event called when a {@link Conversation} is either starting or has started.
+ *
+ * <p>You really should use the sub-interfaces as they include the most access
+ * to information and let you know if the conversation is starting or has
+ * started already.</p>
+ */
+public interface ConversationOpenEvent extends Event {
```
kashike: Should extend a `ConversationEvent`.

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+    protected AbstractConversationEndCauseBuilder() {
+
+    }
+
+    @Nullable protected ConversationEndType endType;
+
+    @Override
+    public B type(ConversationEndType endType) {
+        this.endType = checkNotNull(endType,
+            "ConversationEndType cannot be null!");
```
kashike: Wrong formatting.

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndTypes.java
```diff
@@ -0,0 +1,58 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.cause.conversation;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * Provided conversation end types that you should attempt to stick to.
+ */
+public class ConversationEndTypes {
```
kashike: `final`

kashike (changes requested):

src/main/java/org/spongepowered/api/conversation/Conversant.java
parlough: Any suggestions for a better name?

src/main/java/org/spongepowered/api/conversation/Conversation.java
parlough: No need. This is a string the conversant types to leave the conversation.

src/main/java/org/spongepowered/api/conversation/ConversationArchetype.java
parlough: No need. This is a string the conversant types to leave the conversation.

src/main/java/org/spongepowered/api/conversation/Conversation.java
kashike: We use `Text` to represent the plain-text message sent from the client to the server, so we may as well use `Text` here too.

src/main/java/org/spongepowered/api/conversation/Conversant.java
parlough: The reason I use `CommandSource` is because I utilize much of command's argument classes and capabilities to not reuse a lot of that. For example the `CommandElement#parse` method requires a CommandSource. Like I mentioned earlier a lot of what is used for commands could be abstracted more and improved in some ways but I felt that was out of the scope of this PR.

src/main/java/org/spongepowered/api/conversation/Conversation.java
parlough: I used the raw message(String s) from the implementations though before even the events are called for the time I check this.  No reason to really turn a string into a piece of text when it will just be turned right back into a string. I suppose it also takes more memory to store a text in the archetype too, but that's pretty insignificant. 

I am not completely against changing this to Text, but I don't feel it's necessary.

pie-flavor: You say "user", but a Conversant could be anything. For example, something that could send Text.

src/main/java/org/spongepowered/api/event/cause/conversation/common/package-info.java
liach: good, fixed!

src/main/java/org/spongepowered/api/conversation/Question.java
```diff
@@ -0,0 +1,142 @@
...
+     * @return The answer handler for this question
+     */
+    AnswerHandler getHandler();
+
+    /**
+     * Gets the prompt for this question.
+     *
+     * @return The prompt
+     */
+    Text getPrompt();
```
dags-: It'd be neat if prompt messages could make use of the current context. For example, question no.1 might ask for the conversant's name, then question no.2 could be like: "`hi <name>, how old are you?`".

Maybe prompts could be a Function<Context,Text>-like interface rather than preset Texts?

gabizou: OR simply using a `TextTemplate`?

liach:
@Meronat Can you move the package under extra?

kashike:
This would fit under the `text` package.

src/main/java/org/spongepowered/api/conversation/Question.java
dags-: Not all command elements produce textrepresentable results. A question might ask for a Player, and then the next prompt might want to include info about that player - I suppose the dev could add those details to the context manually in the answerhandler or something

parlough: So at the moment a prompt is constant and you could accomplish this within the previous answer handler. Instead of having to specify a function every time with information that you already have access to in the answer handler, maybe the next prompt could be specified in the QuestionResult? If it present it will override the prompt specified within the question. 


Also, should I include a reference to the data container (the context) within answer handlers?

parlough: So thinking about this, I was thinking we should keep the prompt as is to allow for a simple, generic option, but allow you to specify an optional replacement `Text` within the `QuestionResult` you return. If it is specified, it takes the place of the prompt.

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
parlough: Should be good now :)

parlough:
For starting a conversation, I accept a `PluginContainer`, should I accept a plugin object instead?

src/main/java/org/spongepowered/api/text/conversation/QuestionResult.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Gets the next question, if it is present.
+     *
+     * <p>Generally should only be present if the type is NEXT.</p>
+     *
+     * @return The next question, if available.
+     */
+    public Optional<Question> getNextQuestion() {
+        return Optional.ofNullable(this.nextQuestion);
```
ST-DDT: Sine it is a constant i would store it as an Optional directly.

src/main/java/org/spongepowered/api/text/conversation/QuestionResult.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+public class QuestionResult {
+
+    /**
+     * Gets a question result which will end the question.
+     *
+     * @return A question result which ends the question
+     */
+    public static QuestionResult end() {
+        return new QuestionResult(QuestionResultType.END);
```
ST-DDT: These ones look unmodifyable. Maybe store them in a constant?

src/main/java/org/spongepowered/api/text/conversation/QuestionResult.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Gets the question result which will move the conversation to the
+     * next question.
+     *
+     * @param nextQuestion The desired question to move on to
+     * @return A question result which will move on to the next question
+     */
+    public static QuestionResult nextQuestion(Question nextQuestion) {
+        return new QuestionResult(QuestionResultType.NEXT, nextQuestion);
```
ST-DDT: Missing null check for question.

ST-DDT (changes requested):

ST-DDT:
> For starting a conversation, I accept a PluginContainer, should I accept a plugin object instead?

That is what we usually do

src/main/java/org/spongepowered/api/conversation/Question.java
ST-DDT: No. If you add it to the question result then you have a strong coupling between questions, because one question('s AnswerHandler) needs to know how the next question looks like. What if i want fto implement a quiz game. The order of questions is random, but i would like to show the current question number.

There is also the possibility that the context less question promt does not make sense. Like a confiemation dialog that does not show what you confirm. And if you add a helper method for constant prompt there is not much complexity overhead.

src/main/java/org/spongepowered/api/text/conversation/ExternalChatHandler.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.spongepowered.api.text.conversation;
+
+import org.spongepowered.api.text.Text;
+
+/**
+ * Each {@link Conversant} has one of these to handle the chat messages being
+ * sent to them.
+ */
+public interface ExternalChatHandler {
```
dualspiral: I feel like this could/should be a `CatalogType` - you did it for the `ConversationEndType` type and so you should probably do the same here. 

Your implementation in `ExternalChatHandlers` should then be moved into Common, just leaving keys for the types you've defined.

src/main/java/org/spongepowered/api/text/conversation/ResponseContext.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.conversation;
```
dualspiral: Is this the right place to put something common to Commands and Conversations? I think this belongs in `command`, as the response is, in effect, responding to a well defined action and has nothing to do with formatting of text.

To be honest - question for the developers here, given what the Conversation API does (essentially it's a glorified set of command executors which works like a wizard to execute actions on the server, rather than just display & formatting of text), should this whole thing be under the `command` package? I know kashike suggested this `text` package, but I'm not sure it's as good a fit as `command`.

src/main/java/org/spongepowered/api/text/conversation/QuestionResult.java
```diff
@@ -0,0 +1,124 @@
...
+
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * What to return from your answer handler on a question, to allow moving on to
+ * a next question, the same question, or to end the conversation.
+ */
+public class QuestionResult {
```
dualspiral: Most of this is implementation that should be in Common.

That said, I can only see that there are _two_ options for returning a result, ask a question, or end the conversation. Is there value in having a `REPEAT` type - instead just setting the question again (which you already have from the `AnswerHandler`)? If not, then I suggest stripping the types out.

I understand having this result object (though maybe as an interface), in case we want to add more information in the future should the need arise. For example, this could probably extend (or at least use a subset of) `CommandResult` though the usefulness of that is probably debatable.

dualspiral (dismissed):

src/main/java/org/spongepowered/api/text/conversation/ExternalChatHandler.java
ST-DDT: IMO it is not a catalog type. Think about a custom impl that let through messages that contains red text or a specific keyword. Such as a server shutdown notification.

dualspiral: Sure, but allowing custom implementations shouldn't prevent something becoming a `CatalogType`. [Plugins would be able to register their own `ExternalChatHandler`s](https://jd.spongepowered.org/5.1.0/org/spongepowered/api/GameRegistry.html#register-java.lang.Class-T-) and should probably be encouraged to do so. 

A plugin doesn't _really_ need to specifically register it - in which case, all this will do is make them implement `getId()` and `getName()` with some throwaway string. They shouldn't, but I guess it's an option some would take.

ST-DDT: Are you suggesting that he should add a ECHType that is a factory like think that can create ECH instances? Because ECH itself cannot be CatalogTypes because they exists as 1/conversant and not 1/game. The id needs to be unique and must contain a meaning, just iterating some counters and throwaway names is not the purpose of CatalogTypes.

However a factory like thing is required anyway because the ECH is not target aware but is returned in the archetpye for all conversants.

dualspiral: Yep, you're right, this can't be a `CatalogType` - I didn't think about the `SendOnFinishChatHandler` there which as you correctly note is `1/conversant`, and I had my head in "this is implementation" mode and only properly saw the `1/game` ones. My mistake.

So yes, some sort of factory like thing is the way to go.

parlough:
So there's double of each commit. Just ignore that I guess since I don't know how to fix it.

parlough:
I'll be getting back to this soon, probably this weekend. Will be makes changes according to feedback received and hopefully extending this to ConsoleSources as well.

src/main/java/org/spongepowered/api/text/conversation/ExternalChatHandler.java
parlough: Can I have clarification on what a "factory" like thing would look like?

src/main/java/org/spongepowered/api/text/conversation/ResponseContext.java
parlough: I agree this is better in the command package for now due to a lot of common usage. I have moved it to its own sub-package.

src/main/java/org/spongepowered/api/text/conversation/QuestionResult.java
parlough: I'm not really sure how to move this implementation to common. For builders we have a registry which can get create the implementation of the builder. I am not disagreeing it should be moved, I just don't know how to properly.

src/main/java/org/spongepowered/api/text/conversation/ResponseContext.java
parlough: Relating to the repeat one it is sort of just a convenience method I suppose, I think it's a common use and wouldn't hurt to keep.

src/main/java/org/spongepowered/api/text/conversation/ExternalChatHandler.java
ST-DDT: Sure.

````java
interface ExternalChatHandler {

    boolean process(Text text);
    void finish();

}
````

````java
@CatalogedBy(ExternalChatHandlerTypes.class)
interface ExternalChatHandlerType extends CatalogType {

     ExternalChatHandler createFor(Conversation conversion, Conversant conversant);

}
````

````java
final class ExternalChatHandlerTypes {

     public static final ExternalChatHandlerType PASSTHROUGH = DummyCatalogType.create(...);
     public static final ExternalChatHandlerType SEND_AFTER_COMPLETION = ...;
     public static final ExternalChatHandlerType DISCARD = ...;

}
````

**Comment by parlough** [comment]
So I went ahead and added a `CancellingHandler` which can be specified to handle input to see if a `Conversant` should quit the conversation. This replaces the exit keyword. 

I also have moved to allow for contextual prompts through a `PromptHandler`. 

Next step is moving `ExternalChatHandler` to some sort of `CatalogType`/Factory solution.

Then finally is supporting consoles as a `Conversant`.

Will see you all soon with an updated example and some documentation.

File: src/main/java/org/spongepowered/api/command/conversation/Conversation.java [hunk 31]
```diff
@@ -0,0 +1,268 @@
...
+    /**
+     * Adds the specified {@link Conversant} to the conversation,
+     * with the generic delete all {@link ExternalChatHandler}.
+     *
+     * <p>This sends them the prompt of the current active question.</p>
+     *
+     * @param conversant The conversant to add to the conversation
+     */
+    default void addConversant(Conversant conversant) {
+        addConversant(conversant, ExternalChatHandlers.deleteAll());
```
ST-DDT: Shouldn't this use some kind of default specified on the archetype?

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,268 @@
...
+     * conversation will just continue without any conversants.</p>
+     */
+    void removeAllConversants();
+
+    /**
+     * If the specified {@link Conversant} is in the conversation it changes
+     * their {@link ExternalChatHandler} to the one you specify. If they are
+     * not in the conversation, instead use {@link #addConversant(Conversant)}.
+     *
+     * @param conversant The conversant which you want to modify
```
ST-DDT: What happens ro the previous ECH?

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,268 @@
...
+     * @return Whether or not the specified ending handler was removed
+     */
+    boolean removeEndingHandler(EndingHandler endingHandler);
+
+    /**
+     * A convenience method to get the id of the conversation archetype.
+     *
+     * @return The id of the conversation archetype
+     */
+    default String getId() {
```
ST-DDT: I would move it to the top of the class since its the id.

src/main/java/org/spongepowered/api/command/conversation/EndingHandler.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * <p>Also consider the fact that there may be no conversants left
+     * within the conversation.</p>
+     *
+     * @param conversation The involved conversation
+     * @param context The context which has been added to over the course of
+     *     the conversation
+     * @param endType How the conversation ended
+     */
+    void handle(Conversation conversation, DataContainer context, ConversationEndType endType);
```
ST-DDT: Missing Cause?

src/main/java/org/spongepowered/api/command/conversation/PromptHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.api.text.Text;
+
+/**
+ * A functional interface, which a {@link Question} utilizes
+ * to generate its prompt.
+ */
+@FunctionalInterface
+public interface PromptHandler {
+
+    Text handle(Conversation conversation, DataContainer context);
```
ST-DDT: Javadocs and i would rename the method to something like getPromptText .

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
parlough: Good question. Right now nothing, but do you think it's process should be called? Maybe I can add a boolean to the method to determine that. What's your opinion?

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
parlough: You are correct. This has been addressed.

src/main/java/org/spongepowered/api/command/conversation/PromptHandler.java
parlough: `getPrompt` :)

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
parlough: Done.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
ST-DDT: Maybe add a new method to ECHs for this

void drainTo(Consumer<Text> text) and add the Consumer interface if it is not already there.

That way you could also send it to the player mid conversation...

ST-DDT:
Also i suggest adding a new ECHT that uses a scheduler to send the messages after completion.
Ex One message per second otherwise the user might get spammed with the messages and is unable to read/see them.

parlough:
@ST-DDT That sounds like a great idea actually. I'm going to refactor ECHs with the other suggested changes tomorrow so I'll add that to my list. Thanks for the consistent feedback.

parlough:
So at the moment, due to there being the handlers to catch chat, I have to add a `sendThroughMessage` method to conversants for our messaging uses and plugin creators to use if they want to send something extra during their answer handlers. I'm not super happy with this solution if anyone has any input in this area.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,268 @@
...
+     */
+    Optional<Question> getCurrentQuestion();
+
+    /**
+     * Sets the new question, sending them the header, as well as the prompt
+     * of the question.
+     *
+     * @param question The question to set to
+     */
+    void setQuestion(@Nullable Question question);
```
ST-DDT: Could you please explain in the javadocs what the implications of setting the questions to null are? Does it terminate the convetsation or does  it just disable the processing of the send messages. What would be a usecase?

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,268 @@
...
+     * <p>This does not end the conversation, so keep in mind
+     * you should use {@link #end(ConversationEndType, Cause)} or else this
+     * conversation will just continue without any conversants.</p>
+     */
+    void removeAllConversants();
+
+    /**
+     * If the specified {@link Conversant} is in the conversation it changes
+     * their {@link ExternalChatHandler} to the one you specify. If they are
+     * not in the conversation, instead use {@link #addConversant(Conversant)}.
```
ST-DDT: I would change this a bit to throw an IllegalArgumentException or ISE if called for an Conversant that does not participate in the conversation.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
```diff
@@ -0,0 +1,349 @@
...
+     *     within the conversation
+     */
+    ExternalChatHandlerType getDefaultChatHandlerType();
+
+    /**
+     * Gets the handlers that handles the end of the {@link Conversation}.
+     *
+     * @return All ending handlers
+     */
+    ImmutableSet<EndingHandler> getEndingHandlers();
```
ST-DDT: I would just return a Set here and state in the javadocs that the list is immutable. That way the impl isnt limited to the given impl class.

ST-DDT: Or maybe just a Collection because some impls might use a List to retain an execution order.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.spongepowered.api.text.Text;
+
+import java.util.function.Consumer;
+
+/**
+ * Each {@link Conversant} has one of these to handle the chat messages being
+ * sent to them.
+ */
+public interface ExternalChatHandler extends Consumer<Text> {
```
ST-DDT, line 35: The Consumer's apply method should be default implemented.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return Whether or not to send the message normally to the player
+     */
+    boolean process(Text text);
+
+    /**
+     * Notify this handler that the conversation has come to an end.
+     */
+    void finish();
+
+    void drainTo(Consumer<Text> consumer);
```
ST-DDT: Missing Javadoc

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+    ExternalChatHandlerType getType();
```
ST-DDT: Missing javadoc

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandlerType.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.spongepowered.api.command.conversation;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(ExternalChatHandlerTypes.class)
+public interface ExternalChatHandlerType extends CatalogType {
+
+    ExternalChatHandler createFor(Conversation conversation, Conversant conversant);
```
ST-DDT: Missing Javadocs.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandlerTypes.java
```diff
@@ -0,0 +1,39 @@
...
+
+public final class ExternalChatHandlerTypes {
+
+    public static final ExternalChatHandlerType PASS_THROUGH = DummyObjectProvider.createFor(ExternalChatHandlerType.class, "PASS_THROUGH");
+
+    public static final ExternalChatHandlerType DISCARD = DummyObjectProvider.createFor(ExternalChatHandlerType.class, "DISCARD");
+
+    public static final ExternalChatHandlerType SEND_ON_COMPLETION = DummyObjectProvider.createFor(ExternalChatHandlerType.class, "SEND_ON_COMPLETION");
+
+    public static final ExternalChatHandlerType SEND_SLOWLY_ON_COMPLETION = DummyObjectProvider.createFor(ExternalChatHandlerType.class, "SEND_SLOWLY_ON_COMPLETION");
```
ST-DDT: Missing javadocs and iirc the catalog types should be sorted shouldnt they?

src/main/java/org/spongepowered/api/command/conversation/QuestionResult.java
```diff
@@ -0,0 +1,126 @@
...
+ */
+public class QuestionResult {
+
+    /**
+     * Gets a question result which will end the question.
+     *
+     * @return A question result which ends the question
+     */
+    public static QuestionResult end() {
+        return new QuestionResult(QuestionResultType.END);
```
ST-DDT: Why dont you use static/cached instances for this?

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndCause.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import javax.annotation.Nonnull;
+
+/**
+ * The cause behind a conversation ending.
+ */
+public interface ConversationEndCause {
```
ST-DDT, line 33: Where is this interface used? I havent found a method using it?!

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
parlough: To do what?

src/main/java/org/spongepowered/api/command/conversation/QuestionResult.java
Meronat: image
in the `SpongeConversation` class.

It's essentially to include the type in the cause, but I have had doubts on if that is necessary. Maybe I should just put the ConversationEndType in the cause, I'm not sure.

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndCause.java
parlough: image
in the `SpongeConversation` class.

It's essentially to include the type in the cause, but I have had doubts on if that is necessary. Maybe I should just put the ConversationEndType in the cause, I'm not sure.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
parlough: Sounds good.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
ST-DDT: Just call `process()` in there and discard the boolean result.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
parlough: Good idea.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
parlough: Thinking about it now, plugins shouldn't really need to set a question to null, I'll think more about this.

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndCause.java
parlough: Definitely looking for feedback here, as looking back I'm just not sure if the separate cause is warranted.

src/main/java/org/spongepowered/api/command/conversation/QuestionResult.java
parlough: @dualspiral Wanted me to move QuestionResult's more to the implementation I think and I was looking for some more input on that.

I wasn't sure of a great way as there was nothing that clean I could come up with, and I was trying to make it familiar to usage of `CommandResult`. 

For the mean-time I will move this to a static instance, thanks.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
parlough: Done :)

src/main/java/org/spongepowered/api/event/cause/conversation/ConversationEndCause.java
dualspiral: With the Cause and Contexts PR well underway, I'd look into what's going on there and how it's being dealt with - what the process is there, as the earliest this will be is API 7, as will that. So, I'd worry less about this unless/until you know what the general pattern will be going forward.

parlough: You have a great point there. I'll look more into cause and context and see how I'd like to handle this with that new system.

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandlerTypes.java
parlough: Yeah you're right, should be good now unless I messed up the alphabet..

src/main/java/org/spongepowered/api/command/conversation/ExternalChatHandler.java
dualspiral: Does it really need to extend `Consumer<Text>`? What use does that have?

parlough: Provides a common way to transfer messages from one `ExternalChatHandler` to another if the developer switches the one that the `Conversant` is in.

src/main/java/org/spongepowered/api/event/conversation/ConversationOpenEvent.java
```diff
@@ -0,0 +1,100 @@
...
+ * started already.</p>
+ */
+public interface ConversationOpenEvent extends ConversationEvent {
+
+    /**
+     * Called before the {@link Conversation} is created, giving you access to
+     * the archetype as well as who will be in the conversation.
+     *
+     * <p>If you are removing all conversants, it is better to simply cancel the
+     * conversation creation.</p>
```
dualspiral: Rather than this and leave a zero conversant conversation, why not state that you'll detect this and just not create the conversation if it is empty?

src/main/java/org/spongepowered/api/event/conversation/ConversationOpenEvent.java
```diff
@@ -0,0 +1,100 @@
...
+import org.spongepowered.api.command.conversation.Conversation;
+import org.spongepowered.api.command.conversation.ConversationArchetype;
+import org.spongepowered.api.event.Cancellable;
+
+import java.util.Set;
+
+/**
+ * An event called when a {@link Conversation} is either starting or has started.
+ *
+ * <p>You really should use the sub-interfaces as they include the most access
```
dualspiral: I don't think it's necessary to state this. Developers for Sponge should be used to the fact that subinterfaces are generally a better option, they are used everywhere.

src/main/java/org/spongepowered/api/event/conversation/ConversationCloseEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Gets the conversation.
+     *
+     * @return The conversation
+     */
+    Conversation getConversation();
+
+    /**
+     * Called when the conversation is in the process of ending, so it can
+     * still be cancelled.
```
dualspiral, line 48: What happens when you are cancelling a conversation which is ending?

src/main/java/org/spongepowered/api/event/conversation/ConversationOpenEvent.java
parlough: Good point, simpler for everyone I suppose.

src/main/java/org/spongepowered/api/event/conversation/ConversationCloseEvent.java
parlough: At the moment it does not call the `EndingHandler`s, remove the conversants, etc. The plugin listening to the event, would isolate what type of end it is through the ConversationEndType/Cause, and if it is the appropriate kind with the appropriate information they could cancel it. The only type of end where it does not completely make sense is when the conversation is truly finished, but that is up to them to handle. Some may want to set the question again or whatever.

dualspiral: I think this needs to be documented then. You need to say what cancellation does and what is expected after, because while cancelling _might_ stop a conversation from ending, it just leaves it in limbo with no question in play. 

To be honest, having this cancellable without the ability to set a question at the same time seems off to me, but with the way the `Cancellable` interface has a `setCancelled` method, it makes it difficult to do it any other way. The problem is that a plugin cancelling this event might not find out if the ending was actually cancelled or not (something could uncancel it), could put it into a weird state. So, I'd at least add a method to set the next question, which will cancel the event at the same time.

src/main/java/org/spongepowered/api/command/conversation/Conversant.java
```diff
@@ -0,0 +1,85 @@
...
+    void removeFromConversation();
+
+    /**
+     * Due to conversations often catching messages, this allows messages while
+     * they're in a conversation to go through to them no matter what
+     * and bypass their {@link ExternalChatHandler}.
+     *
+     * @param message The message to send
+     */
+    void sendThroughMessage(Text message);
```
dualspiral: What about an overload to `sendMessage`, `sendMessage(Text message, boolean ignoreConversation)`? Again, not entirely sure whether this is the _best_ way to go, but I feel like the method name should remain the same, `sendThroughMessage`, at least at a glance, isn't clear about what it does.

vectrixdevelops: I think that's a very appropriate idea.

src/main/java/org/spongepowered/api/event/conversation/ConversationCloseEvent.java
parlough: So I have a sort of solution partially implemented atm(_note to self still need to use question they set_), but it's not that great and might not even work. I am going to continue to think about this.

Meronat:
I've still got some cleaning up to do related to the closing event, messages, and the console as a conversant. Getting closer though :)

parlough:
I plan to get back to this soon. At the moment I've got some cleaning up to do as well as figuring out messaging better, and of course full console functionality.

bloodmc:
@Meronat any update on this PR?

parlough:
@bloodmc I was going to wait until dual's command PR was finished since conversations rely on that backbone for quite a lot(being quite similar to commands), and will experience API changes due to changes in his.

If that is okay, I could work on getting this updated and finished, with the expectation of quite a few breakages in bleeding once his PRs are merged as well as the cause and context refactor.

src/main/java/org/spongepowered/api/command/conversation/Conversant.java
```diff
@@ -0,0 +1,88 @@
...
+     * if they are in one.
+     *
+     * @return This conversant's current {@link Conversation}, if available
+     */
+    Optional<Conversation> getConversation();
+
+    /**
+     * A convenience method to see if the player is currently in a conversation.
+     *
+     * @return Whether or not they are a conversation
```
Zerthick, line 48: I think this should read "@return Whether or not they are **in** a conversation"

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents a created conversation with one or more {@link Conversant}s.
+ *
+ * <p>These are checked every five minutes to see if there are no players.
+ * If this is true, the conversation is ended with </p>
```
Zerthick, line 43: Ended with?

Zerthick (commented):
Just some javadoc nitpicks 😛

Zerthick:
By and large I think this is pretty cool. One idea that came to mind while looking it over was to incorporate additional ways to receive the answer to a question besides direct typing.  One of the biggest benefits in my mind of having a conversation API is to spare the user from having to type out long and complicated commands, however it also has the potential to remove the need to type entirely.  One use case that comes to mind is that the question presents the user choices in the form of clickable links in chat, the user only has to click a choice to provide an answer to a question.  This could streamline the conversation process even more.  Consideration would have to be made for `Conversant`s without a mouse to click though, like the console for example.  The same effect could also be achieved by simply asking the user to type a choice out, though it introduces more potential for input errors. Just food for thought.

dualspiral:
We discussed this after SoS, but to confirm, I’ll co-ordinate with @Meronat to ensure that the command API is structured in a way that exposes what Conversations need easily.

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
```diff
@@ -0,0 +1,127 @@
...
+public interface ConversationManager {
+
+    /**
+     * Gets an active {@link Conversation} with the specified id and creator.
+     *
+     * @param plugin The plugin who created the conversation
+     * @param id The id of the conversation
+     * @return The matching conversation, if present
+     */
+    Collection<Conversation> getConversation(PluginContainer plugin, String id);
```
Wundero, line 54: Should this not be an optional?

liach: This can be empty

src/main/java/org/spongepowered/api/command/conversation/Conversant.java
```diff
@@ -0,0 +1,88 @@
...
+     *     one they were originally in
+     */
+    Conversation setConversation(Conversation conversation);
+
+    /**
+     * Sets their conversation to null. Should only be used by Sponge as it does
+     * not remove them from the conversation. Instead use the
+     * {@link Conversation#removeConversant(Conversant)} method.
+     */
+    void removeFromConversation();
```
liach, line 74: Should we expose this method?

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
Wundero: But the idea is that it returns either one conversation, with the given id, or nothing. Having the collection implies returning multiple conversations is supported, however multiple conversations under the same plugin and id should not exist.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * Sets the new question, sending them the header, as well as the prompt
+     * of the question.
+     *
+     * <p>This overrides the current question, so be careful with
+     * its use.</p>
+     *
+     * @param question The question to set to
+     */
+    void setQuestion(Question question);
```
liach, line 133: Is this nullable?

Wundero: Why would it be? If you are providing a null question, you likely want to forcefully end the conversation, which is what the `end` method is for.

src/main/java/org/spongepowered/api/command/conversation/Conversant.java
Wundero: All it does is internally set the conversation to null in the conversant; the conversation that they are in still holds reference to them, and thus this method does not effectively remove them. I do think a convenience method in `Conversant` should exist though, which calls `getConversation().ifPresent(c->c.removeConversant(this));` as well as the method. (Or, if the`removeConversant` method calls this method, just the previous code.)

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
parlough: Multiple conversations under the same plugin and id can certainly exist. The id is specified in the archetype, which can be reused for multiple active conversations which have the same functionality/actions.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
liach: The question is optional. It is not necessary for a conversation.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * Gets the handlers that handles the end of the {@link Conversation}.
+     *
+     * <p>The collection is implementation dependent, but is
+     * generally immutable or a copy.</p>
+     *
+     * @return All ending handlers
+     */
+    Collection<EndingHandler> getEndingHandlers();
```
liach, line 104: Should this be a simple collection or a set?

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
```diff
@@ -0,0 +1,351 @@
...
+     *     a plugin instance
+     */
+    Optional<Conversation> start(Object plugin, Conversant... conversants);
+
+    /**
+     * Gets a builder based off of this archetype's values.
+     *
+     * @return The builder
+     */
+    Builder toBuilder();
```
liach, line 177: No... Add a method `from(ConversationArchetype archetype)` in the builder instead.

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
liach: Should this be a collection or a set?

src/main/java/org/spongepowered/api/command/conversation/Question.java
```diff
@@ -0,0 +1,168 @@
...
+ * allowing for handling, and identification.
+ */
+public interface Question {
+
+    /**
+     * Creates a new {@link Question.Builder} to build a {@link Question}.
+     *
+     * @return The new builder
+     */
+    static Question.Builder builder() {
```
liach, line 45: You don't need to qualify the name of `Question.Builder`, `Builder` would be fine

src/main/java/org/spongepowered/api/command/conversation/Question.java
```diff
@@ -0,0 +1,168 @@
...
+     * @return This question's input tokenizer
+     */
+    InputTokenizer getInputTokenizer();
+
+    /**
+     * Gets a builder based on the values present in the question.
+     *
+     * @return The new builder
+     */
+    Builder toBuilder();
```
liach, line 95: Same. Change it to `Builder from(Question question)` in the builder

src/main/java/org/spongepowered/api/command/conversation/QuestionResult.java
```diff
@@ -0,0 +1,152 @@
...
+     * @return The next question, if available.
+     */
+    Optional<Question> getNextQuestion();
+
+    /**
+     * Gets a builder based off of this result's values.
+     *
+     * @return The builder
+     */
+    Builder toBuilder();
```
liach, line 90: Same

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+import org.spongepowered.api.event.cause.conversation.ConversationEndCause;
+import org.spongepowered.api.event.cause.conversation.ConversationEndType;
+
+import javax.annotation.Nullable;
+
+@SuppressWarnings("unchecked")
+public abstract class AbstractConversationEndCauseBuilder<T extends ConversationEndCause, B extends ConversationEndCause.ConversationEndCauseBuilder<T, B>>
+    implements ConversationEndCause.ConversationEndCauseBuilder<T, B> {
+
+    @Nullable protected ConversationEndType endType;
```
liach, line 38: Annotation and field should be on different lines

src/main/java/org/spongepowered/api/event/conversation/ConversationOpenEvent.java
```diff
@@ -0,0 +1,96 @@
...
+
+        /**
+         * Gets a copy of all conversants.
+         *
+         * <p>If you want to add or remove a conversant, use the provided
+         * matching methods instead.</p>
+         *
+         * @return A copy of all conversants
+         */
+        Set<Conversant> getConversants();
```
liach, line 63: How about exposing this and the archetype to the ConversationEvent?

src/main/java/org/spongepowered/api/event/conversation/ConversationCloseEvent.java
```diff
@@ -0,0 +1,93 @@
...
+ * An event called when a conversation is either ending or has already ended.
+ */
+public interface ConversationCloseEvent extends ConversationEvent {
+
+    /**
+     * Gets the conversation.
+     *
+     * @return The conversation
+     */
+    Conversation getConversation();
```
liach, line 44: Should you create a sub-interface for the majority of conversation events which has a conversation? (Except the before begin one)

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
Wundero: How is it optional? `Question` handles the majority of the actual conversation, from prompting the user to directing the flow. If a conversation has no current question it is a dead conversation. I can see no useful example where setting question as null would be used instead of ending the conversation.

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
Wundero: @Meronat Okay that makes more sense then, I didn't realize id's weren't supposed to be unique.

@liach It should probably be a collection, given that (knowing that id's are not unique) sets require unique entries.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
Wundero: What benefit does having a set use?

liach: Stating that it does not have duplicate entries

liach: Collections can return a list, so I am asking here

Wundero: I was more wondering why duplicate entry prevention mattered though. I don't think the uses of this collection will be enough to notice a performance difference between set and list, considering speed vs duplicates, so is there another reason to force distinct entries?

parlough:
Thanks for the continued feedback, keep it coming. I plan to rebase on-top of dual's command PR not longer after I return from a trip and address feedback then as well :)

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
kashike: No, this is fine.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
ST-DDT: In arenas/battles you might want to stop chat or handle chat differently.
(Only teammates can see each others messages and outside spam is blocked...)
Thus you use a conversation to use the external chat handler feature and some special answer handlers without ever asking a question.

src/main/java/org/spongepowered/api/command/conversation/ConversationManager.java
ST-DDT: Then please add at least a hint to the javadocs thats its the id of the archetype and not the id of the conversion. Currently its confusing

parlough: Well I mean the id of its archetype is the conversation's id as well because the id specifies what kind of conversation it essentially is, but I can try to add some clarification.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
liach: @Meronat Is it intended to have duplicate entries or not?

src/main/java/org/spongepowered/api/event/cause/conversation/common/AbstractConversationEndCauseBuilder.java
liach: Hmm the code style updates too frequently to keep up.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
dualspiral: Given `Set` implements `Collection`, this is a moot point. Why would a plugin care? They would just want all the ending handlers. The _implementation_ should be what decides this, not the API.

src/main/java/org/spongepowered/api/command/conversation/Conversation.java
Wundero: You might as well use chat channels for that though. Besides, the answer handlers are in the question not the conversation, so even in that situation you would need a question. I don't believe setting this to null offers any real benefits. If you want to pause the conversation you can do it with a prompt handler rather that having null values that may cause issues.

src/main/java/org/spongepowered/api/command/conversation/ConversationArchetype.java
ryantheleach: Because it exposes the contract of whether duplicates can exist or not.

dualspiral: Yes, I am aware of that, but I’m still unsure as to why a plugin would care in this situation. Collections are returned in other parts of the API that you wouldn’t expect duplicates from - unless this is a modifiable set, I don’t see why ‘Set‘ vs ‘Collection‘ matters as the return type here and by specifying ‘Collection‘, we become free to alter what is returned later if we need to without breaking the contract of the API.

Zidane: Collection is fine. The docs can state that the implementation shouldn't have duplicates in the collection but we don't enforce that type. The implementation could decide to use another data type that prevents duplicates and is NOT a set.

liach: agrees with Zidane for performance's sake

dualspiral:
@Meronat Please have a look at #1587 - particularly the last comment and about the `Parameters` part. Also have a look at the `CommandContext` and how the implementation works with arguments, and whether this is reasonable for this, or whether you might need a different tack. Does this seem reasonable in the context of the conversations?

Note that any implementation between commands and conversations DO NOT have to be shared once commands are merged, if there needs to be impl differences, that's not unreasonable and we will be able to handle that now. Thus, I'm more interested about whether the API is suitable, rather than impl.

We can discuss this in a more informal setting when you and I get a bit more time, but for now, it's worth starting to make sure that things should work fairly simply from merge.

ImMorpheus:

Issue #1533, opened 2017-04-09
Fire an event when a ServerBossBar is created/modified


VapidLinus:
It would be great with an event that fires when a ServerBossBar is created or modified, as well for when a player is added or removed from one.

See the first and last posts in this thread:
https://forums.spongepowered.org/t/solved-serverbossbar-retrieval-help/18268/5

mosemister:
The original post suggests that you also want a way so you can retrieve bossBars from the player such as one of the following ways

    List<BossBar> list = Player.getBossBars();

Or 

    List<ServerBossBar> list = Sponge.getServer().getBossBars();
    //From there filter bars with the player

VapidLinus:
@mosemister Yes that's correct.

I was thinking though if the Sponge team didn't want to do that for whatever reason, then having events fired when they're created/joined/left would be enough for a plugin to track that by itself.

However, having both the events and `player.getBossBars()` and `server.getBossBars()` would be the best.

ryantheleach:
@VapidLinus we have been discussing what the possible API would look like for this, and were wondering what the usecase was, so we can make sure that whatever gets proposed would suit.

Some brainstorming we came up with:

Renaming withers to say who spawned them.
editing the text to have a countdown timer for how long the wither will last.
Stopping the bars being displayed for riding enderdragons.

I havn't tested the forum usecase, but presumably if you are giving the player a bossbar, your own plugin could cache / keep a collection of the bossbars it has issued and manage them that way.

kashike:
You can obtain the `ServerBossBar` for a `Wither` now: https://github.com/SpongePowered/SpongeAPI/pull/1534/files#diff-96e1f3b357e27049ddf52fd33571cadfR63
Access to the dragon bossbar is coming later.

PR #1534, opened 2017-04-10
Add some helper methods for working with ServerBossBars

kashike:
[**API**](https://github.com/SpongePowered/SpongeAPI/pull/1534) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1296)
Adds #1533

Test plugin: https://gist.github.com/kashike/bb89aee16738a68f078a6e0c82432583

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/complex/EnderDragon.java
+++ b/src/main/java/org/spongepowered/api/entity/living/complex/EnderDragon.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.entity.living.complex;
 
+import org.spongepowered.api.boss.ServerBossBar;
 import org.spongepowered.api.entity.EnderCrystal;
 import org.spongepowered.api.entity.living.Aerial;
 import org.spongepowered.api.entity.living.monster.Boss;
@@ -47,5 +48,12 @@ public interface EnderDragon extends ComplexLiving, Boss, Aerial, ProjectileLaun
      * @return The ender crystal
      */
     Optional<EnderCrystal> getHealingCrystal();
+
+    /**
+     * Gets the boss bar this dragon uses.
+     *
+     * @return The boss bar
+     */
+    ServerBossBar getBossBar();
 
 }
--- a/src/main/java/org/spongepowered/api/entity/living/monster/Wither.java
+++ b/src/main/java/org/spongepowered/api/entity/living/monster/Wither.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.entity.living.monster;
 
+import org.spongepowered.api.boss.ServerBossBar;
 import org.spongepowered.api.entity.explosive.FusedExplosive;
 import org.spongepowered.api.entity.living.Aerial;
 import org.spongepowered.api.entity.living.Agent;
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -339,4 +343,42 @@ default GameModeData getGameModeData() {
...
+     *
+     * @param bar The boss bar
+     * @see ServerBossBar#removePlayer(Player)
+     */
+    default void removeBossBar(final ServerBossBar bar) {
+        checkNotNull(bar, "bar").removePlayer(this);
+    }
+
+    /**
+     * Gets an unmodifiable collection of boss bars this player is currently in.
```
ST-DDT: While technically correct, this sounds odd.

ST-DDT (approved):
I like the last method. The other ones just look like inversion of controll, but for usability reasons i guess its ok.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
kashike: What would you recommend?

ST-DDT: `... that are shown to this player` or `... that apply to this player`. Sorry i should have proposed these directly, but I'm not sure about them either.

parlough: `Gets an unmodifiable collection of the boss bars this player is currently viewing.`

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -339,4 +343,42 @@ default GameModeData getGameModeData() {
...
+     *
+     * @param bar The boss bar
+     * @see ServerBossBar#addPlayer(Player)
+     */
+    default void addBossBar(final ServerBossBar bar) {
+        checkNotNull(bar, "bar").addPlayer(this);
+    }
+
+    /**
+     * Removes this player to the boss bar.
```
JonathanBrouwer: Is this correct english?
I think it should be 'Removes this player from the boss bar.'

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -339,4 +343,42 @@ default GameModeData getGameModeData() {
...
+     * @see ServerBossBar#removePlayer(Player)
+     */
+    default void removeBossBar(final ServerBossBar bar) {
+        checkNotNull(bar, "bar").removePlayer(this);
+    }
+
+    /**
+     * Gets an unmodifiable collection of boss bars this player is currently in.
+     *
+     * @return A collection of boss bars this player is currently in
```
parlough: Change this to match whatever you choose for ST-DDT's suggestion.

parlough (approved):
I think this is a good addition and simply makes sense to have.

Zidane:
@kashike can you update the comments above for the suggestions so I can look into merging this.

ryantheleach:
Talked to kashike, Current status is a slight rework in the relationship between players and bossbars vs their helper methods. I don't know exact details, but have heard there will be work on this soon.

gabizou:
@kashike @Zidane would we want this in API 7 at all?

PR #1537, opened 2017-04-13
Basic client support

mattmess1221:
[**SpongeAPI**](https://github.com/SpongePowered/SpongeAPI/pull/1537) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1304)

### DIFF
--- a/src/main/java/org/spongepowered/api/Client.java
+++ b/src/main/java/org/spongepowered/api/Client.java
@@ -1,0 +1,101 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
+
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.util.ThreadContext;
+import org.spongepowered.api.world.World;
+
+import java.net.InetSocketAddress;
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Represents the Minecraft client.
+ */
+public interface Client extends ThreadContext {
+
+    /**
+     * Gets the currently loaded world, if any.
+     *
+     * @return The current world
+     */
+    Optional<World> getLoadedWorld();
+
+    /**
+     * Gets the current client player entity, if it exists.
+     *
+     * @return The client player
+     */
+    Optional<Player> getClientPlayer();
+
+    /**
+     * Gets the current integrated server instance if it is running.
+     *
+     * @return The integrated server
+     */
+    Optional<Server> getIntegratedServer();
+
+    /**
+     * Loads a world save with the given name.
+     *
+     * @param saveLocation The location of the save
+     */
+    CompletableFuture<Server> loadWorldSave(Path saveLocation);
+
+    /**
+     * Joins a server.
+     *
+     * TODO return a multiplayer session object
+     *
+     * @param server The server to join
+     */
+    CompletableFuture<?> joinServer(InetSocketAddress server);
+
+    /**
+     * Gets the current pause state of the game. In singleplayer, pausing the game will also pause
+     * the world. In multiplayer, it will only bring up the pause menu.
+     *
+     * @return The current pause state
+     */
+    Value<Boolean> pause();
+
+    /**
+     * Gets the directory resource packs are stored in.
+     *
+     * @return The directory for resource packs
+     */
+    Path getResourcePacksDirectory();
+
+    /**
+     * Gets the directory game assets are located and should be stored in.
+     *
+     * @return The directory for assets
+     */
+    Path getAssetsDirectory();
+
+}
--- a/src/main/java/org/spongepowered/api/Game.java
+++ b/src/main/java/org/spongepowered/api/Game.java
@@ -72,6 +72,23 @@ public interface Game {
     Server getServer();
 
     /**
+     * Returns if the {@link Client} is available for use. The result of this method is entirely
+     * dependent on the implementation.
+     *
+     * @return True if the Client is available, false if not
+     */
+    boolean isClientAvailable();
+
+    /**
+     * Gets the {@link Client} if the game is client-side and is available. If it is not, an error
+     * is thrown.
+     *
+     * @return The client
+     * @throws IllegalStateException If the Client isn't currently available.
+     */
+    Client getClient();
+
+    /**
      * Gets the {@link PluginManager}.
      *
      * @return The plugin manager
--- a/src/main/java/org/spongepowered/api/Server.java
+++ b/src/main/java/org/spongepowered/api/Server.java
@@ -31,6 +31,7 @@ import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.scoreboard.Scoreboard;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageChannel;
+import org.spongepowered.api.util.ThreadContext;
 import org.spongepowered.api.world.ChunkTicketManager;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.WorldArchetype;
@@ -47,7 +48,7 @@ import java.util.concurrent.CompletableFuture;
 /**
  * Represents a typical Minecraft Server.
  */
-public interface Server {
+public interface Server extends ThreadContext {
 
     /**
      * Gets the {@link Player}s currently online.
@@ -433,10 +434,4 @@ public interface Server {
      */
     void setPlayerIdleTimeout(int timeout);
 
-    /**
-     * Checks if the current thread matches the main thread of the server.
-     *
-     * @return True if main thread, false if not
-     */
-    boolean isMainThread();
 }
--- a/src/main/java/org/spongepowered/api/Sponge.java
+++ b/src/main/java/org/spongepowered/api/Sponge.java
@@ -174,6 +174,36 @@ public final class Sponge {
     }
 
     /**
+     * Gets whether a {@link Client} instance is available without throwing an
+     * exception from calling {@link #getServer()}.
+     *
+     * @see Game#isClientAvailable()
+     * @return True if the client instance is available
+     */
+    public static boolean isClientAvailable() {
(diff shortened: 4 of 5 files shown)


src/main/java/org/spongepowered/api/ThreadContext.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
```
kashike: `.util`

src/main/java/org/spongepowered/api/multiplayer/ServerEntry.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Created by Matthew on 4/8/2017.
+ */
+public interface ServerEntry {
+
+    /**
+     * Gets the address used to connect to this server. It is guarenteed to be a valid address.
+     *
+     * TODO return {@link InetAddress}?
```
kashike: :+1:

Cybermaxke:
I also made a PR for a key binding API: https://github.com/SpongePowered/SpongeAPI/pull/1210
But this allows the server to bind key bindings on the client and the client will notify the server if anything changes, so no client side API.

You may also want to separate this PR into smaller ones? Because this PR will keep growing as you add new or missing things. You should try to target a specific feature that you want.

src/main/java/org/spongepowered/api/multiplayer/ServerStatusResponse.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Returns true if this server is hosted on a client which has been opened
+     * to LAN.
+     *
+     * @return True if this is on a local client, false if not
+     */
+    boolean isLocal();
+
+    /**
+     * Gets the brand this server is running on.
```
kashike:
```java
    /**
     * Gets the brand this server is running on.
     *
     * <p>e.g. "forge,fml,sponge", "Spigot"</p>
     */

File: src/main/java/org/spongepowered/api/multiplayer/Session.java [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+import org.spongepowered.api.util.Identifiable;
+
+/**
+ * The session data for a user.
+ */
+public interface Session extends Identifiable {
+
+    String getName();
+
+    String getSessionID();
```
kashike: `getSessionId`

kashike (changes requested):

mattmess1221:
I'm already removing the inputs api, so this will just be for generic client, resources, and textures.

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,80 @@
...
+
+    Optional<Session> getSession();
+
+    Optional<World> getLoadedWorld();
+
+    Optional<Player> getClientPlayer();
+
+    Optional<Server> getIntegratedServer();
+
+    void loadWorld(String world);
```
ST-DDT: I would rename this method to `loadSavegame()` otherwise it might cause confusion.

ST-DDT: It might also return `CompleteableFuture<Server>`

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,67 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.resource;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface ResourceManager {
```
Cybermaxke: You should expand the existing `AssetManager` instead of creating a new interface used for the same purpose.

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,80 @@
...
+
+    void joinServer(ServerEntry server);
+
+    Value<Boolean> pause();
```
ST-DDT: I'm neither sure about the name nor the location of this method. Maybe wr could add an interface for IntegratedServers and put that method there.

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,80 @@
...
+
+    Optional<World> getLoadedWorld();
+
+    Optional<Player> getClientPlayer();
+
+    Optional<Server> getIntegratedServer();
+
+    void loadWorld(String world);
+
+    void joinServer(ServerEntry server);
```
ST-DDT: Is there nothing to get the currently connected remote servers details?

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -72,6 +73,23 @@
...
+    boolean isClientAvailable();
+
+    /**
+     * Gets the {@link Client} wrapped in an {@link Optional} if the game is client-side.
+     * If it is not, it returns empty.
+     *
+     * @return The client
+     * @throws IllegalStateException If the Client isn't currently available.
+     */
+    Client getClient();
```
ST-DDT, line 89: Javadocs and return type dont match.

src/main/java/org/spongepowered/api/Client.java
ST-DDT: Nvm. Session might be that information. However returning a `CompleteableFuture<Session>` or some kind of progress indivator might be useful.

src/main/java/org/spongepowered/api/multiplayer/Session.java
```diff
@@ -0,0 +1,72 @@
...
+     * <p>Authenticates this session against mojang's session server.</p>
+     *
+     * <p>
+     * Note: This may use network bandwidth, so it may lock up the thread
+     * while it waits.
+     * </p>
+     *
+     * @return The auth level of the user
+     */
+    Auth authenticate();
```
ST-DDT: If this might block then it should probably return a CompleteableFuture.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/Client.java
mattmess1221: Updated with javadocs

src/main/java/org/spongepowered/api/Client.java
mattmess1221: `Session` may be a confusing name. It actually represents the current session with Mojang in terms of authentication.

src/main/java/org/spongepowered/api/resource/ResourceManager.java
mattmess1221: Do assets support resource packs? I explain this in the javadocs for `ResourceLocation`

src/main/java/org/spongepowered/api/resource/ResourceLocation.java
```diff
@@ -0,0 +1,104 @@
...
+     * The spec must include a domain and a path in the format of {@code domain:path/to/resource.png}.
+     * </p>
+     *
+     * @param spec fully qualified resource path, including domain
+     * @return A resource location
+     * @throws IllegalArgumentException If the spec is invalid
+     */
+    static ResourceLocation of(String spec) {
+        int indx = spec.indexOf(':');
+        checkArgument(indx < 0, "Resource has no domain: " + spec);
```
kashike: It isn't required to have a domain, `"minecraft"` should be assumed if no domain is specified.

src/main/java/org/spongepowered/api/resource/ResourceManager.java
Cybermaxke: It could be supported, but sponge scans currently only the class loader for resources because it's currently only server side. A server side resouce pack system would also be nice, to create new advancements and recipes in 1.12

mattmess1221: That's one of the reasons why I was thinking about implementing the resource manager in common. And changing the assets system like that will likely break things.

src/main/java/org/spongepowered/api/multiplayer/ServerStatusResponse.java
mattmess1221: Done

src/main/java/org/spongepowered/api/Game.java
mattmess1221: Fixed

src/main/java/org/spongepowered/api/multiplayer/PlayerInformation.java
```diff
@@ -0,0 +1,65 @@
...
+     * @return The profile
+     */
+    GameProfile getProfile();
+
+    /**
+     * Gets the location of this player's skin if it exists
+     *
+     * @return The skin location
+     */
+    Optional<ResourceLocation> getSkinLocation();
```
kashike: Might be good to have a `CatalogType` of parts (`SKIN`, `CAPE`, `ELYTRA`).

src/main/java/org/spongepowered/api/resource/ResourceManager.java
Cybermaxke: Existing methods can still be used and how will this break things? The implementation will just allow resources/assets to be overriden through a resource pack system, but will fall back to the class loader if there is no file present. It only expands the functionality.

mattmess1221: I think each has a different use-case. `Asset`s should be used for things like config defaults which should not need to be overridden by the user.

Resources, which **can** be overridden, would be for game content. Things like recipes, blockstates, models, loot tables, or even structures.

Cybermaxke: Ok yes, that seems a valid point. But I would still like to see a common API for both overridable and unoverridable assets. A developer can choose the class loader asset manager, or the asset manager that supports resource packs, which falls back to the class loader asset manager.

killjoy1221:
How would you suggest I go about doing that?

src/main/java/org/spongepowered/api/resource/ResourceManager.java
mattmess1221: How would you suggest I do this?

Cybermaxke: Something like this: https://gist.github.com/Cybermaxke/24c015876742730b9b1407b96e463506
There are more methods present in the current API but this will give a basic idea. In the Server/Client class are there two methods, one for the system asset manager and one that supports resource packs and custom child asset managers.

You can also look at a implementation I did a while ago: https://github.com/LanternPowered/LanternServer/tree/master/src/main/java/org/lanternpowered/server/asset
Although it can be cleaned up, since I created a custom Asset interface to add the id (resource location). And I like your idea about having an event to check when a asset manager is reloaded.

mattmess1221: The event is actually a vanilla feature. See `IReloadableResourceManager.registerReloadListener(IResourceManagerReloadListener)`

Cybermaxke: Yeah, but I referred to throwing a event through the sponge event manager and listening to them with `@Listener`, and not registering a listener directly for a asset manager.

src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageModifierTypes.java
```diff
@@ -85,11 +85,6 @@
     public static final DamageModifierType BLOCKING = DummyObjectProvider.createFor(DamageModifierType.class, "SHIELD");
 
     /**
-     * Represents a {@link DamageModifier} that will reduce damage due to using a shield.
-     */
-    public static final DamageModifierType SHIELD = DummyObjectProvider.createFor(DamageModifierType.class, "SHIELD");
```
Zidane: Out of scope of this PR and just creates noise.

Zidane:
Not a bad first step in looking at the client. This needs to be broken up across multiple, distinctive, PRs. There is simply too much going on that getting changes to all the sections will just have this end up in PR hell.

src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageModifierTypes.java
mattmess1221: Sorry, that was done automatically by gradle's sortClassFields

mattmess1221:
Going to reopen this

PR #1539, opened 2017-04-14
Expose sending a specific page of the Pagination

ryantheleach:
This exposes a method to send a specific page of a pagination.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
+++ b/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
@@ -196,7 +196,7 @@ public final class CommandSpec implements CommandCallable {
 
         /**
          * Sets an extended description to use in longer help listings for this
-         * command. Will be appended to the short description and the command's
+         * command. Should be appended to the short description and the command's
          * usage.
          *
          * @param extendedDescription The description to set
@@ -399,9 +399,7 @@ public final class CommandSpec implements CommandCallable {
     }
 
     /**
-     * Return a longer description for this command. This description is
-     * composed of at least all present of the short description, the usage
-     * statement, and the extended description
+     * Returns the extended description
      *
      * @param source The source to get the extended description for
      * @return the extended description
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -103,8 +103,18 @@ static Builder builder() {
      * Send the constructed pagination list to the given source.
      *
      * @param source The source to send to
+     * @param page The page to send
      */
-    void sendTo(MessageReceiver source);
+    void sendTo(MessageReceiver source, int page);
+
+    /**
+     * Send the constructed pagination list to the given source.
```
ST-DDT, line 113: Maybe we should state explicitly that it sends the first page. Just to avoid confusion with a random `current` page.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -103,8 +103,18 @@ static Builder builder() {
      * Send the constructed pagination list to the given source.
      *
      * @param source The source to send to
+     * @param page The page to send
```
ST-DDT: We should mention that the first page is one vs zero as developers count indexes. We should also mention what happens if we use a page after the last one.

ST-DDT (changes requested):
Thats a really nice addition to pagination.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
ST-DDT: Plus maybe a `throws IllegalArgumentException` if the page is zero or negative.

src/main/java/org/spongepowered/api/command/spec/CommandSpec.java
```diff
@@ -399,26 +399,15 @@ public Text getUsage(CommandSource source) {
...
-        Optional<Text> desc = getShortDescription(source);
-        if (desc.isPresent()) {
-            builder.append(desc.get(), Text.NEW_LINE);
-        }
-        builder.append(getUsage(source));
-        if (this.extendedDescription != null) {
-            builder.append(Text.NEW_LINE, this.extendedDescription);
-        }
-        return Optional.of(builder.build());
+        return Optional.ofNullable(extendedDescription);
```
kashike, line 410: What are these changes doing in here?

Zidane:
Closed in favor of: https://github.com/SpongePowered/SpongeAPI/pull/1622

PR #1540, opened 2017-04-15
Basic client support

mattmess1221:
[**SpongeAPI**](#) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1445)

I'm going to start with some basic client support right now. Currently there is getting the client world, player, loading a save, joining a world, as well as fetching the integrated server (which can also be done from Game.


src/main/java/org/spongepowered/api/util/ThreadContext.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.spongepowered.api.util;
+
+/**
+ * Interface for checking the current thread context.
+ */
+public interface ThreadContext {
+
+    /**
+     * Checks if the current thread matches the main thread of the server.
```
kashike: Javadocs need updating.

src/main/java/org/spongepowered/api/Sponge.java
```diff
@@ -174,6 +174,35 @@ public static Server getServer() {
     }
 
     /**
+     * Gets whether a {@link Client} instance is available without throwing an
+     * exception from calling {@link #getServer()}.
```
kashike: Javadocs need updating.

kashike (changes requested):

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param server The server to join
+     */
+    void joinServer(InetSocketAddress server);
+
+    /**
+     * Gets the {@link ServerConnection} to the current server.
+     *
+     * @return The server connection
+     */
+    Optional<ServerConnection> getServerConnection();
```
ST-DDT: Does it support both local and remote servers? If (not) it should be specified.

mattmess1221: Yes. A `RemoteConnection` can point to a local channel.

ST-DDT (approved):

src/main/java/org/spongepowered/api/Client.java
ST-DDT: Then please add this to the javadocs so you can be sure about that without knowing anything about the API and just doing some API discovery. (Much info => good documentation)

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -69,7 +66,7 @@
  * <p>Any methods called on Player that are not on User do not store any data
  * that persists across server restarts.</p>
  */
-public interface Player extends Humanoid, User, RemoteSource, Viewer, ChatTypeMessageReceiver {
+public interface Player extends BasePlayer, RemoteSource, Viewer, ChatTypeMessageReceiver {
```
kashike: `ServerPlayer extends Player`

src/main/java/org/spongepowered/api/entity/living/player/BasePlayer.java
```diff
@@ -0,0 +1,60 @@
...
+import org.spongepowered.api.entity.living.Human;
+import org.spongepowered.api.scoreboard.Scoreboard;
+
+import java.util.Set;
+
+/**
+ * Base interface for {@link Player server} and {@link ClientPlayer client}
+ * players.
+ */
+public interface BasePlayer extends Human, User {
```
kashike: `Player`

src/main/java/org/spongepowered/api/entity/living/player/tab/PlayerTabList.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+import javax.annotation.Nullable;
+
+/**
+ * A mutable version of {@link TabList}. It also contains a reference to the
+ * {@link Player} entity.
+ */
+public interface PlayerTabList extends TabList {
```
kashike: `ServerTabList`

kashike (changes requested):

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,100 @@
...
+ * Represents the Minecraft client.
+ */
+public interface Client extends ThreadContext {
+
+    /**
+     * Gets the currently loaded world, if any.
+     *
+     * @return The current world
+     */
+    Optional<World> getLoadedWorld();
```
ST-DDT: I'm not sure about this method.
First of all it is not entirely sure which world this is. Is this the world the client is currently viewing or the main world of the savegame and does it include the local copy of the remote world.

Then the location i would rather move it to the LocalServer interface because it is a substate of that.

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    Optional<LocalServer> getLocalServer();
+
+    /**
+     * Loads a world save with the given name.
+     *
+     * @param saveLocation The location of the save
+     * @see Game#getSavesDirectory()
+     */
+    void loadWorldSave(Path saveLocation);
```
ST-DDT: There should be a way to know whether the loafing succeeded and the server/game is running

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Joins a server.
+     *
+     * @param server The server to join
+     */
+    void joinServer(InetSocketAddress server);
```
ST-DDT: Same here

src/main/java/org/spongepowered/api/Client.java
ST-DDT: Or note that the method blocks.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -24,136 +24,17 @@
...
-     *
-     * <p>If text formatting is not supported in the implementation
-     * it will be displayed as plain text.</p>
-     *
-     * @param message The message to send
-     * @param cause The cause for the message
-     * @return The event that was thrown from sending the message
-     */
-    MessageChannelEvent.Chat simulateChat(Text message, Cause cause);
+public interface Player extends Human, User {
```
JBYoshi: Should be `Humanoid` instead of `Human`.

JBYoshi (changes requested):

mattmess1221:
Some changes in my latest push:
Earlier, I had completely butchered the Player class. No more of that. Instead, I extracted a couple sub-interfaces. One is a `BasePlayer`, and the other is a `PlayerController`. (I'm not married to those names yet)

Anyway, BasePlayer seems pretty straightforward, so I'll focus on PlayerController.

Using the `PlayerController` and subsequent `ClientPlayerController`, you can control certain things about the player and make it do things. This is basically where I stuck all the shared methods which the client player and all the server players have access to.

The server can open a container on a player, and so can the client to itself. The client is not able to do this to other players.

src/main/java/org/spongepowered/api/Client.java
liach: the world the client is currently viewing

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Gets the current client player controller, if it exists.
+     *
+     * @return The client player
+     */
+    Optional<ClientPlayerController> getPlayerController();
+
+    @Override
+    default void sendMessage(ChatType type, Text message) {
```
ST-DDT: Javadocs

ST-DDT:
Some default methods have the Override on an newline and some have them in the same line. IMo it should be always in a newline.

mattmess1221:
Recent changes:
1. Add connection events for the client. Named `ServerConnectionEvent`
2. Deprecated the chat event. It's been replaced by 4 platform specific events. client -> server, server -> client, send, receive.
3. Added logical side to `Extent`
4. Added a loopback `MessageChannel` for either the console or client.

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -101,6 +101,23 @@
...
+    boolean isClientAvailable();
+
+    /**
+     * Gets the {@link Client} if the game is client-side and is available. If
+     * it is not, an error is thrown.
+     *
+     * @return The client
+     * @throws IllegalStateException If the Client isn't currently available.
+     */
+    Client getClient();
```
parlough, line 120: Why not just have this return `Optional<Client>` instead? Errors are a pain and an optional fits well here I think. If it's empty that means there is no client available.

mattmess1221: I've been keeping it consistent with Server. If someone else agrees, I'll change it.

src/main/java/org/spongepowered/api/Client.java
```diff
@@ -53,6 +56,11 @@
      */
     Optional<ClientPlayerController> getPlayerController();
 
+    @Override
+    default void sendMessage(ChatType type, Text message) {
```
ryantheleach: Is this plugin to gui? or client to server?

mattmess1221: I removed this method earlier. It delegated to the PlayerController, so it would send a message to the client.

src/main/java/org/spongepowered/api/Game.java
Zidane: Keep this consistent with getServer, we did it so that everyone's code wouldn't be littered with TONS of getServer().get() calls.

src/main/java/org/spongepowered/api/network/NetworkPlayer.java
```diff
@@ -0,0 +1,43 @@
...
+import org.spongepowered.api.entity.living.player.tab.TabListEntry;
+import org.spongepowered.api.profile.GameProfile;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
+
+/**
+ * @see TabListEntry
+ */
+public interface NetworkPlayer {
```
Zidane, line 13: Maybe NetworkPlayerEntry? NetworkPlayer itself may imply something different to others.

src/main/java/org/spongepowered/api/entity/living/player/ClientPlayerController.java
```diff
@@ -0,0 +1,68 @@
...
+ * Player controller for the client's player
+ */
+public interface ClientPlayerController extends PlayerController {
+
+    /**
+     * Gets the {@link Humanoid} player entity.
+     *
+     * @return The player entity
+     */
+    Humanoid getPlayer();
```
JBYoshi, line 40: Should this be in `PlayerController` as well? It can be default-implemented in `Player` to `return this`.

src/main/java/org/spongepowered/api/entity/living/player/ClientPlayerController.java
```diff
@@ -0,0 +1,68 @@
...
+     * sign (&#x00a7)</p>
+     *
+     * <p>Note: Calling this method too often may result in being kicked for
+     * chat spam. Different server implementations have different thresholds
+     * and cooldown times for chat messages.</p>
+     *
+     * @param message The message to send.
+     * @throws IllegalArgumentException If the message has illegal characters
+     */
+    void sendChatMessage(String message) throws IllegalArgumentException;
```
JBYoshi, line 66: Should this be part of `PlayerController`?

src/main/java/org/spongepowered/api/entity/living/player/PlayerController.java
```diff
@@ -0,0 +1,126 @@
...
+ * itself. e.g. The client can show itself a container, but it cannot do the
+ * same to someone else.</p>
+ *
+ * <p>Other methods which could be located here include things the player knows
+ * about itself, but cannot possibly know about another. If the server is not
+ * likely to know it, look elsewhere.</p>
+ *
+ * TODO I don't like the name 'controller'
+ */
+public interface PlayerController extends RemoteSource, ChatTypeMessageReceiver {
```
JBYoshi, line 50: Not sure if this should extend `RemoteSource` - the `ClientPlayerController` isn't technically "remote".

src/main/java/org/spongepowered/api/entity/living/player/ClientPlayerController.java
mattmess1221: No. If I put it in `PlayerController`, people will see it in `Player` and think it sends a message to the player. In actuality, this method sends a method to the server, which `Player` already has a similar method for. (`simulateChat(Text, Cause)`)

src/main/java/org/spongepowered/api/entity/living/player/ClientPlayerController.java
mattmess1221: This actually shouldn't be here. I want to abstractify the player's entity.

src/main/java/org/spongepowered/api/entity/living/player/ClientPlayerController.java
JBYoshi: How about making that the shared method? That name seems to work for both client and server.

gabizou:
Is this being superseded by https://github.com/SpongePowered/SpongeAPI/pull/1804? Genuine question because I do feel both targeting the same idea, but as I mentioned in the other one, we'll be exposing the client in a more fuller context for API 8 along with some more breakages.

mattmess1221:
I guess it is being superceded. IMO, this request got too big for 1 thing. I would like to focus on my Resources PR in api 8 for now anyway.

Zidane:
@killjoy1221 

Would you be interested in re-opening this PR? This was a step in the right direction and I've already merged some client stuff into 1.13. Namely that other PR.

mattmess1221:
Alright, I'll work on it a bit. I will also create a new PR targetting 1.13.

PR #1544, opened 2017-04-22
Add StringDataFormat that allows reading/writing data from text I/O

stephan-gh:
Adds an extended `StringDataFormat` interface for `DataFormat` that allows reading/writing data for text formats like HOCON or JSON much more efficiently directly from/to a string or a reader/writer instead of always going through a binary `InputStream`.

Not entirely sure about the naming, open for alternative suggestions for the interface and/or the new methods.

Implementation PR: https://github.com/SpongePowered/SpongeCommon/pull/1318

### DIFF
--- a/src/main/java/org/spongepowered/api/data/persistence/DataFormats.java
+++ b/src/main/java/org/spongepowered/api/data/persistence/DataFormats.java
@@ -33,9 +33,9 @@ public final class DataFormats {
 
     // SORTFIELDS:ON
 
-    public static final DataFormat HOCON = DummyObjectProvider.createFor(DataFormat.class, "HOCON");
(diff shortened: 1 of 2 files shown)


Deamon5550:
Maybe just `StringDataFormat`. TextDataFormat makes me think its something to do with our `Text` stuff.

stephan-gh:
@Deamon5550 Yeah, I was thinking about the same when I wrote this. The reason I chose text is because you usually call them "text files" and not "string files". I agree with you though, `StringDataFormat` is probably less confusing.

gabizou:
If it counts, I prefer `StringDataFormat` over `TextDataFormat`, at first glance, I thought it had to do with our Text API.

stephan-gh:
@Deamon5550 @gabizou Changed it to `StringDataFormat`.

PR #1553, opened 2017-05-08
Expose slot ordinal via interface

NeumimTo:
No need to update sponge common 
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/item/inventory/adapter/impl/slots/SlotAdapter.java

Slot adapter already has the member, this pr just adds needed method into the Slot interface which allows us to access the ordinal number of a slot.

If anyone want to access slot's id, or whatever you call it, now the only possible way is to add spongevanilla/forge into your dependencies and typecast to slotadapter. Which is annoing to do.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Slot.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Slot.java
@@ -38,4 +38,11 @@ public interface Slot extends Inventory {
      */
     int getStackSize();
 
+	/**
+     * Gets the ordinal of a slot.
+     *
+     * @return the slot's ordinal
+     */
+    int getOrdinal();
+
 }


Mumfrey:
> If anyone want to access slot's id, or whatever you call it, now the only possible way is to add spongevanilla/forge into your dependencies and typecast to slotadapter. Which is annoing to do.

Why do you want the slot ID in the first place, this is an implementation detail which is *deliberately* hidden from consumers, exposing it has absolutely zero value.

NeumimTo:
>Why do you want the slot ID

From the simple fact i want to know which exact slots were affected by the slot transaction.

>implementation detail which is deliberately hidden from consumers

Why?

>exposing it has absolutely zero value.

The existence of this pr refutes this claim

DenWav:
Now now child, no need to get an attitude.

gabizou:
@NeumimTo If you were to explain why the ordinal id is *needed* and what it is to be used for, maybe that might be better. Otherwise I have to agree with @Mumfrey that this PR serves no purpose and the implementation detail it exposes is just that, an implementation detail.

NeumimTo:
@DemonWav  ?

@gabizou
Currently we dont know (just with the api) in which inventory slot SlotTransaction occured.

gabizou:
> Currently we dont know (just with the api) in which inventory slot SlotTransaction occured.

So you want to expose an internal magical number to associate with an ordinal? How does that help you figure out the slot position or any other information?

dualspiral:
@NeumimTo Perhaps it's worth showing us what you're trying to do, possibly along with an example, explaining why exposing this is the only way to go. Explaining the _why_ it's needed (showing us there is no other way to do what you want) along with the workflow, rather than the proposed _what_ you want doing would help.

me4502:
So I actually have a use case for this. In CraftBook, I used to use the ordinal to work out which way the user scrolled for the painting switcher. I am not aware of another way of doing it.

Deamon5550:
@me4502 It could make sense to get an ordinal from a`Hotbar` as it does have a numbered index relating to the number keys used to select them.

NeumimTo:
@gabizou some examples

- i would like to know from side (right or left) a player scrolled to an item, basically what @me4502 wrote

- when player picks up an item i would like to know into which slot of hotbar it was assigned.

- trying to do some item exchange between two players. In a nutshell shared grid inventory, where i want to know if a player moved an item from left side to right side, or reversed. Yes possible workaround with custom data manipulators, but again too much work, where slotid could be used without too much code being written.

JBYoshi:
For a bunch of that, there's a 2D slot position you can grab from a `GridInventory` with `gridInventory.getProperty(slot, SlotPos.class, "slotpos")`. And in `OrderedInventory`s, you can grab a pseudo-ordinal with `orderedInventory.getProperty(slot, SlotIndex.class, "slotindex")`. Although I don't know if they're implemented or not; I'll look into it.

JBYoshi:
Just checked, and it should work.

Note SpongePowered/SpongeCommon#1344 - though I'm looking into it right now.

**EDIT:** I put in the wrong code above. Corrected.

gabizou:
As @JBYoshi has demonstrated, you can retrieve the slot position for both grid and indexed based inventories through the API.

PR #1554, opened 2017-05-10
Advancements API

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1645) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/1913) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/363)

This PR adds the advancements API, with additions extra to the features provided in vanilla Minecraft. Implementations will follow when 1.12 is released.

The `AdvancementTree` represents a advancement Tab, it can be build with a root advancement. `Advancement`s that are attached to the root advancement will be present in the tree. The layout of the tree can also be customized, which means that you can move the advancements around in the tab though a `AdvancementTreeEvent.GenerateLayout` event.

`AdvancementCondition`s are the conditions that should be met before an advancement is unlocked. I provided AND and OR operations that can be applied. There is not limit in the complexity like in the json files.

There is also `Trigger`s that can be applied to the `AdvancementCondition`, the details of vanilla triggers are not exposed in the API (the configuration). However you can create your own `Trigger`s which will be parsed in the advancement json files.
http://minecraft.gamepedia.com/Advancements#Triggers

Additionally there is a `ScoreAdvancementCriterion` which requires a specific goal value to be achieved before it gets done, this is also supported by triggers. The trigger will need to trigger the criteria multiple times until the goal is reached, this is also supported in the json files by adding a `trigger_times` property to the trigger.

Extras:
* Generic Events - Used for the registrations of advancements and trees.
* `ScoreAdvancementCriterion` - Allows you to provide a goal value and when the score reaches its goal the advancement will be unlocked. Also supported in json files.
* Complex AND and OR condition operations
* Custom triggers (configurable) and supported in json
* Customize the advancement tree layout.


src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
```diff
@@ -0,0 +1,202 @@
...
+     */
+    AdvancementCriterion EMPTY = DummyObjectProvider.createFor(AdvancementCriterion.class, "EMPTY");
+
+    /**
+     * Attempts to build a new AND operation with the given {@link AdvancementCriterion}s.
+     *
+     * @param criteria The other criteria
+     * @return The and operation
+     */
+    static AdvancementCriterion buildAnd(AdvancementCriterion... criteria) {
```
stephan-gh: I don't think the `build` prefix is needed here

kashike:
This should target the `1.12` branch.

src/main/java/org/spongepowered/api/advancement/Styled.java
```diff
@@ -0,0 +1,155 @@
...
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.TextRepresentable;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * Some that has a style within a advancement tree. Either an
+ * {@link AdvancementTree} or an {@link Advancement}.
+ */
+public interface Styled extends CatalogType, TextRepresentable {
```
kashike: This name should be more specific.

kashike (changes requested):

src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
Cybermaxke: Otherwise will it conflict with the non-static `and` method.

Cybermaxke:
@kashike I don't think that this PR will be finished before the 1.12 branch is merged. Or is there going to be separate development between 1.11 and 1.12 for a while?

kashike:
@Cybermaxke I'd say it's better to target `1.12` - there are changes there for 1.12, not in bleeding, and this PR targets 1.12.

stephan-gh:
@Cybermaxke You can easily change the target branch again once we merge 1.12. Until it is merged it should target 1.12 so you get all the updated dependencies etc. (you will also not be able to implement it by targeting bleeding)

src/main/java/org/spongepowered/api/advancement/AdvancementType.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.spongepowered.api.advancement;
+
+import org.spongepowered.api.CatalogType;
+
+/**
+ * Changes the frame around the {@link Advancement} icon and
+ * also the appearance in the notifications.
+ */
+public interface AdvancementType extends CatalogType {
```
mattmess1221: @ CatalogedBy(AdvancementTypes.class)

src/main/java/org/spongepowered/api/advancement/AdvancementTypes.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * An enumeration of all the available {@link AdvancementType}s in minecraft.
+ */
+public final class AdvancementTypes {
+
+    // SORTFIELDS:ON
+
+    /**
+     * Notification appearance:
+     *      Challenge Complete!
```
mattmess1221, line 38: Add a method for this in `AdvancementType`

Something like `List<Text> format(Advancement adv);`

src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
mattmess1221: You can make it `static AdvancementCriterion and(AdvancementCriterion adv1, AdvancementCriterion... therest)`

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerType.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.spongepowered.api.advancement.criteria.trigger;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.advancement.criteria.AdvancementCriterion;
+
+/**
+ * Represents a source that can trigger a {@link AdvancementCriterion}.
+ */
+public interface TriggerType extends CatalogType {
```
mattmess1221: @ CatalogedBy(TriggerTypes.class)

src/main/java/org/spongepowered/api/advancement/AdvancementProgress.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    Advancement getAdvancement();
+
+    /**
+     * Gets the {@link ScoreCriterionProgress} for the given {@link ScoreAdvancementCriterion}.
+     *
+     * @param criterion The score criterion
+     * @return The score criterion progress
+     */
+    default ScoreCriterionProgress tryGet(ScoreAdvancementCriterion criterion) {
```
liach: `getUnchecked`?

src/main/java/org/spongepowered/api/advancement/AdvancementTypes.java
Cybermaxke: Should it be a `List`? Or maybe just a `Text` object where the lines are separated with `\n`.

src/main/java/org/spongepowered/api/advancement/AdvancementProgress.java
Cybermaxke: Maybe `require` similar to the data methods?
https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/data/value/ValueContainer.java
`getUnchecked` isn't used anywhere else as far I know,

stephan-gh:
Please rebase your PR on the latest bleeding branch.

kashike:
@Cybermaxke Are you working on rebasing this?

Cybermaxke:
I will try to get back on this in a week or so, then I will have more time. I will also start implementing at that time.

randombyte-developer:
What's the status on this?

Cybermaxke:
@randombyte-developer I am going to start with the implementation, I already updated the API. ;)

parlough:
@Cybermaxke The hero we all want but don't deserve.

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public static final TriggerType BRED_ANIMALS = DummyObjectProvider.createFor(TriggerType.class, "BRED_ANIMALS");
+
+    public static final TriggerType BREWED_POTION = DummyObjectProvider.createFor(TriggerType.class, "BREWED_POTION");
+
+    public static final TriggerType CONSTRUCT_BEACON = DummyObjectProvider.createFor(TriggerType.class, "CONSTRUCT_BEACON");
+
+    public static final TriggerType CONSUME_ITEM = DummyObjectProvider.createFor(TriggerType.class, "CONSUME_ITEM");
+
+    public static final TriggerType CURSE_ZOMBIE_VILLAGER = DummyObjectProvider.createFor(TriggerType.class, "CONSUME_ITEM");
```
ImMorpheus: "CONSUME_ITEM" again ?

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public static final TriggerType TAME_ANIMAL = DummyObjectProvider.createFor(TriggerType.class, "TAME_ANIMAL");
+
+    public static final TriggerType TICK = DummyObjectProvider.createFor(TriggerType.class, "TICK");
+
+    public static final TriggerType USED_ENDER_EYE = DummyObjectProvider.createFor(TriggerType.class, "USED_ENDER_EYE");
+
+    public static final TriggerType VILLAGER_TRADE = DummyObjectProvider.createFor(TriggerType.class, "VILLAGER_TRADE");
+
+    public static final TriggerType CHANGED_DIMENSION = DummyObjectProvider.createFor(TriggerType.class, "CHANGED_DIMENSION");
```
ImMorpheus: This should be before CONSTRUCT_BEACON.

ImMorpheus:
Some triggers are missing, is that on purpose ? (effects_changed, levitation, ecc.)

Cybermaxke:
@ImMorpheus No, that's not intentional, they probably added more triggers after I designed the API.

sibomots:
AND, OR.. but not NOT ?

It's also remarkable how many bugs you might find and defects uncover by elaborating the JavaDocs.  Please attempt to flesh those out and scratch off the work item "java docs can be improved."  

Ie., writing the docs will uncover some defects and deliver aid for those who want to understand the PR.

src/main/java/org/spongepowered/api/advancement/AdvancementTypes.java
sibomots: Do not use embedded special characters.    Collection<E> where Collection is List, Set, etc.. is the right approach. Please, no magic format characters in the data.  Use the Library.

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
sibomots: Typically, we use labels like Dummy to mean something that does absolutely nothing.  But here you're using it to do something meaningful.  Why not maker ObjectProvider generic   FooObjectProvider seems to be a factory.  See also, Design Patterns, GoF.  etc..

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
sibomots: All of these string labels "VILLAGER_TRADE" should be moved into a class holding static final strings.  I am loathe to see embedded literal strings in code as much as possible.

Cybermaxke:
@sibomots AND and OR operations are the only ones supported in vanilla minecraft. I already added a custom one for a "score" criterion. Do you know a usecase where a NOT operation could be used, this would complicate the internals a lot more so I will only added it if it's worth adding.

I will also document the javadocs a bit more when the implementation is finished, since things can still change.

src/main/java/org/spongepowered/api/advancement/Advancer.java
```diff
@@ -49,5 +49,5 @@
      *
      * @return The advancement trees
      */
-    Set<AdvancementTree> getAdvancementTrees();
+    Set<AdvancementTree> getUnlockedAdvancementTrees();
```
liach: Guess a collection may be more efficient

src/main/java/org/spongepowered/api/advancement/criteria/AndCriterion.java
```diff
@@ -43,4 +43,18 @@
...
+     * Attempts to build a new AND operation with the given {@link AdvancementCriterion}s.
+     * <p>
+     * There is no guarantee that the returned extends {@link AndCriterion}, this depends
+     * on if there are duplicate criteria, {@link #EMPTY} is present or when no
+     * extra criteria are provided.
+     *
+     * @param criteria The other criteria
+     * @return The AND operation
+     */
+    static AdvancementCriterion build(Iterable<AdvancementCriterion> criteria) {
```
liach: Should you name them `of` to improve efficiency

liach: I mean for shorter method names

src/main/java/org/spongepowered/api/advancement/Advancement.java
```diff
@@ -0,0 +1,133 @@
...
+     * @return The advancement tree
+     */
+    Optional<AdvancementTree> getTree();
+
+    /**
+     * Gets all the children {@link Advancement}s.
+     *
+     * @return The children advancements
+     */
+    List<Advancement> getChildren();
```
liach: Do they have a particular order? If yes, can you specify in the javadoc? Otherwise, we may use a collection.

src/main/java/org/spongepowered/api/advancement/Progressable.java
```diff
@@ -0,0 +1,68 @@
...
+ * the process of being achieved.
+ */
+public interface Progressable {
+
+    /**
+     * Gets whether this {@link Progressable} is achieved.
+     *
+     * @return Is achieved
+     */
+    default boolean achieved() {
```
liach, line 41: You should leave this one. Implementations may check so that the method can return without creating extra objects.

src/main/java/org/spongepowered/api/advancement/TreeLayoutElement.java
```diff
@@ -0,0 +1,59 @@
...
+     * @return The position
+     */
+    Vector2d getPosition();
+
+    /**
+     * Sets the position of the {@link Advancement}.
+     *
+     * @param position The position
+     */
+    void setPosition(Vector2d position);
```
liach: I find it okay to redirect this one to `#setPosition(double, double)` like those methods in `Chunk` and `Entity`.

src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
```diff
@@ -0,0 +1,145 @@
...
+ * Represents a criterion that should be acquired
+ * to unlock a {@link Advancement}.
+ */
+public interface AdvancementCriterion {
+
+    /**
+     * Represents an empty criterion, this means that nothing has to be achieved to
+     * unlock a {@link Advancement}.
+     */
+    AdvancementCriterion EMPTY = DummyObjectProvider.createFor(AdvancementCriterion.class, "EMPTY");
```
liach, line 45: I don't know if interface constants can be treated with access transformers. Can anyone enlighten me?

src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    AdvancementCriterion or(Iterable<AdvancementCriterion> criteria);
+
+    /**
+     * Gets the {@link Trigger}s of this {@link AdvancementCriterion}. The
+     * {@link Set} can be empty.
+     *
+     * @return The triggers
+     */
+    Set<Trigger> getTriggers();
```
liach: I don't think we should use a set here. We don't need fast `#contains` operation, and we can indicate that there is no duplication in the javadoc. Triggers instances might not be kept very often; they only have their types exposed.

src/main/java/org/spongepowered/api/advancement/criteria/OperatorCriterion.java
```diff
@@ -0,0 +1,68 @@
...
+    Optional<AdvancementCriterion> findFirst(String name);
+
+    /**
+     * Gets the children {@link AdvancementCriterion}s of this
+     * {@link OperatorCriterion}. May contain
+     * {@link OperatorCriterion}s.
+     *
+     * @return The criteria
+     */
+    Set<AdvancementCriterion> getCriteria();
```
liach: Why sets? I would suggest using naviagable sets and sort the criteria by their name to enable fast lookup and iteration (both)

src/main/java/org/spongepowered/api/advancement/criteria/ScoreCriterionProgress.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Adds the target score value, the score cannot exceed
+     * the goal value ({@link #getGoal()}) or be under zero.
+     * The achieved time will be returned if the goal is met.
+     *
+     * @param score The score to set
+     * @return The achieving time, if achieved
+     */
+    Optional<Instant> set(int score);
```
liach, line 62: Does it mean setting the current score, and return the instant is the set score is the goal?

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
liach: just run sortfield once more and it will be fine.

src/main/java/org/spongepowered/api/advancement/criteria/AdvancementCriterion.java
Cybermaxke: It is done by setting a private static field, not through a access transformer (which is impossible for interfaces).
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/RegistryHelper.java#L96

src/main/java/org/spongepowered/api/advancement/criteria/ScoreCriterionProgress.java
Cybermaxke: If the goal was achieved before, that time will be returned, if you set a score lower than the goal, empty is returned, and equal to goal and not set before will return the current time.

src/main/java/org/spongepowered/api/event/advancement/CriterionRevokeEvent.java
```diff
@@ -0,0 +1,31 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.advancement;
+
+import org.spongepowered.api.event.Cancellable;
+
+public interface CriterionRevokeEvent extends CriterionEvent, Cancellable {
```
parlough: I feel like elsewhere in the API we keep these as nested sub-terfaces to the original since they're so similar and it seems generally the `CriterionEvent` wouldn't be used on its own. Is there a particular reason we shouldn't here? 

For example look at `TabCompleteEvent` or `RconConnectionEvent` and its nested sub-interfaces. 

So if we make this change you'd listen to `CriterionEvent.Revoke` for this one. 

This would also apply to the advancement events.

src/main/java/org/spongepowered/api/advancement/criteria/trigger/TriggerTypes.java
ryantheleach: @sibomots The strings here are used only for debugging. the dummy objects are nearly useless and as far as I'm aware don't even necessarily need correct names outside of logging in the case that something goes terribly wrong.

If Java had a way of copying the variable name or line number into a parameter that'd be the ideal solution.

Yes the dummy objects are a gross hack, but it's a gross hack that was decided on by the core team a long while ago, and isn't the subject of this PR outside correcting the incorrect debug text.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -222,7 +223,10 @@
      *      is not supported
      * @throws CatalogTypeAlreadyRegisteredException if the type cannot be
      *      registered because a matching type was already registered
+     * @deprecated Is scheduled to be removed in API 8, the
+     *      {@link GameRegistryEvent.Register} should be used instead
```
gabizou, line 227: Nested classes will not link properly in Java 8's javadoc generator. They need to be fully qualified to work.

tedstardev:
Do you think it would be possible to add a way to register and call custom triggers (likely namespaced by plugin ID) and criteria sets with this API?  I've simulated this in a vanilla MCP server but I'm not sure if it would be suitable/possible for Sponge?

For example, say I add a new trigger with a plugin: `banking:money_changed`
Then the server owner could add advancements that have the same trigger and a criteria (defined by the plugin) of say `"min": "500"` and it awards whatever advancement... maybe it would be useful?

Edit: Unless I'm totally missing something and you've already done that, I'm not too familiar with github or sponge api yet.. :/

Cybermaxke:
@tedstardev You can create custom criteria (with id) in the API and you can check for available trigger types. There is currently no API for custom triggers, but I will look into adding it.

The following test plugin demonstrates how you can use the API:
https://github.com/Cybermaxke/SpongeCommon/blob/e055c6e6c6f3c3b0b7e768e6a25bf1bfbbcd7b82/testplugins/src/main/java/org/spongepowered/test/AdvancementTest.java

XakepSDK:
How does api works?
Action item depends on advancement icon? Is it possible to set **stone** icon to advancement "smelt **iron**"
What is first arg of `build` method in test plugin?(in api i see only one string arg)

Also, make DUMMY criterion default.

Cybermaxke:
@XakepSDK You cannot modify other advancements with the API, if you really want to do that modify the vanilla resource pack, owever you can create new ones. And look at the javadocs if you want to know what each argument means, in the `build` methods case is it the identifier, without the namespace, this will be retrieved from the plugin container that is handling the event. Which means that you cannot register advancements with a different namespace.

XakepSDK:
Ah, I mean make my own advancement "smelt iron" with stone icon.

Cybermaxke:
@XakepSDK Ah yes, that is possible. The example should give a good idea how to do it.

Cybermaxke:
https://github.com/Cybermaxke/SpongeCommon/blob/e055c6e6c6f3c3b0b7e768e6a25bf1bfbbcd7b82/testplugins/src/main/java/org/spongepowered/test/AdvancementTest.java

XakepSDK:
For example - dig dirt. Where do you say, that player should "dig" dirt?
Only icon differs. Does advancement icon means action? E.g. shovel = dig,
furnace = smelt?

Cybermaxke:
@XakepSDK You trigger the advancements by listening to events for specific actions, for example within the break block event, join event, etc. You have to trigger them by manually. The icon is just visual, it has nothing to do with the triggering

src/main/java/org/spongepowered/api/advancement/AdvancementTree.java
```diff
@@ -0,0 +1,98 @@
...
+        Builder background(String background);
+
+        /**
+         * Builds the {@link AdvancementTree} with the specified
+         * identifier (without the namespace).
+         *
+         * @param id The identifier
+         * @return The advancement tree
+         */
+        AdvancementTree build(String id);
```
Cybermaxke: @gabizou Should I move to `id` to a separate method, to be consistent with the keys builder you added?

gabizou (approved):

PR #1563, opened 2017-05-20
Update TeleportHelper to allow plugin controlled safe teleports

dualspiral:
**Sponge API** | [Sponge Common](https://github.com/SpongePowered/SpongeCommon/pull/1362) | [Original Issue](https://github.com/SpongePowered/SpongeCommon/issues/848)

This PR opens up parts of the Sponge `TeleportHelper` to plugins and mods that support Sponge, allowing plugins reap the benefits of the TeleportHelper, whilst controlling what they consider to be a safe location for a player to warp to.

A <s>`TelportHelperKernel`</s> `TeleportHelperFilter` can be provided and registered with Sponge to control the following:

* Whether a `BlockState` is a suitable floor material.
* Whether a `BlockState` is a safe, passable material, that is, a player can travel through such a material and not get hurt.
* The above checks can be overriden for specific locations, for example, if you want to check to see if a location can see the sky (I use this in the surface only kernel, so an example of this can be seen in the SpongeCommon PR).
* Multiple filters can be specified, _all_ must return a positive result for a location to be accepted as a safe location.

I've also added two parameters to the `TeleportHelper#getSafeLocation` methods, one that takes an integer that specifies how far below a point the floor can be, and one that accepts the kernel.

One thing I haven't worked out yet, and is more one for SpongeCommon, configurable default blacklist for blocks, but as that is configuration and implementation, that does not come into play here.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/TeleportHelper.java
+++ b/src/main/java/org/spongepowered/api/world/TeleportHelper.java
@@ -25,13 +25,21 @@
 package org.spongepowered.api.world;
 
 import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
+import org.spongepowered.api.world.teleport.TeleportHelperFilter;
+import org.spongepowered.api.world.teleport.TeleportHelperFilters;
 
 import java.util.Optional;
 
 /**
  * Finds safe {@link Location}s for {@link Entity}s (typically ones that won't
  * hurt them).
+ *
+ * <p>Typically, the teleport helper will first determine whether the requested
+ * location is a safe one. If not, it will investigate locations close by,
+ * favouring spots closer, and favouring a location above or below over the
+ * x-z plane if two locations are equidistant.</p>
  */
 @NonnullByDefault
 public interface TeleportHelper {
@@ -40,13 +48,19 @@ public interface TeleportHelper {
     int DEFAULT_HEIGHT = 3;
     /** The default width radius to scan for safe locations. */
     int DEFAULT_WIDTH = 9;
+    /**
+     * The default distance to check for a suitable floor below any candidate
+     * location
+     * */
+    int DEFAULT_FLOOR_CHECK_DISTANCE = 2;
 
     /**
      * Gets the next safe {@link Location} around the given location.
      *
      * <p>Safe entails that the returned location will not be somewhere that
      * would harm an {@link Entity}. This method will use the default height and
-     * width for a search area.</p>
+     * width for a search area, and will check for a suitable floor up to two
+     * blocks below any selected block.</p>
      *
      * <p>It's possible the same location will be returned that was passed in.
      * This means it was safe.</p>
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/world/TeleportHelper.java
```diff
@@ -71,9 +132,16 @@
...
+     *                        be. If this is zero or negative, a floor check
+     *                        will not be performed.
+     * @param params The {@link TeleportHelperKernel} to use to determine if a
+     *               location is safe.
      * @return A safe location near the original location or the original
      *         location if it is deemed safe. If no safe location can be found,
      *         {@link Optional#empty()} will be returned
      */
-    Optional<Location<World>> getSafeLocation(Location<World> location, int height, int width);
+    Optional<Location<World>> getSafeLocation(Location<World> location, int height, int width, int maxDropDistance, TeleportHelperKernel params);
```
kashike: Why is it called "params"?

src/main/java/org/spongepowered/api/world/teleport/TeleportHelperKernel.java
```diff
@@ -0,0 +1,80 @@
...
+    Tristate isValidLocation(World world, Vector3i position);
+
+    /**
+     * Tests whether a {@link BlockState} should be considered a safe block
+     * to land on.
+     *
+     * @param blockState The {@link BlockState} to check
+     * @return <code>true</code> if the material should be safe to land on.
+     */
+    boolean isSafeFloorMaterial(BlockState blockState);
```
kashike: `boolean` vs `Tristate`?

src/main/java/org/spongepowered/api/world/teleport/TeleportHelperKernel.java
```diff
@@ -0,0 +1,80 @@
...
+     * Tests whether a {@link BlockState} should be considered a safe block
+     * for the body to be inside of.
+     *
+     * <p>Generally, you want this to be a passable block!</p>
+     *
+     * @param blockState The {@link BlockState} to check
+     * @return <code>true</code> if the material should be safe for the body to
+     *         be inside of.
+     */
+    boolean isSafeBodyMaterial(BlockState blockState);
```
kashike: `boolean` vs `Tristate`?

src/main/java/org/spongepowered/api/world/TeleportHelper.java
dualspiral: Simply from a previous iteration when I was cobbling it together, that function signature was something else. Forgot to change it, will get done.

src/main/java/org/spongepowered/api/world/teleport/TeleportHelperKernel.java
dualspiral: `boolean` here. What would undefined do? I only care for yes or no, unlike the location method, where undefined means "This location has no special significance, continue with material checks."

src/main/java/org/spongepowered/api/world/teleport/TeleportHelperKernel.java
dualspiral: As above.

src/main/java/org/spongepowered/api/world/teleport/TeleportHelperKernel.java
kashike: I was thinking more for using more than one kernel at a time (a kernel which checks with other kernels) - `UNDEFINED` would mean "I don't know what to do, maybe a different kernel does".

dualspiral: Ah OK. I see where you're coming from, but I don't know what a kernel would suggest is unknown, and in the case of one kernel, then there needs to be a resolution. I'd think it's up to the wrapping kernel to decide what to do, so in the case of a modded block kernel that works with our `DEFAULT`, I would expect it would bypass our kernel and just use it's own checks for the modded blocks. A kernel that adds extra restrictions would just `AND` it's own and the wrapped checks, similarly, one that loosens protections to `OR` or selective bypass the wrapped checks.

dualspiral:
Notes to self to do:

- [x] Improve documentation on kernel about the location check. I missed out the consequence of `Tristate.UNDEFINED`.
- [x] Add more documentation to the TeleportHelper to explain how a reasonable implementation should interact with the kernels.

PR #1565, opened 2017-05-21
Correct 'ifSucessful' to DataTransactionResult#ifSuccessful

widd:
Simple spelling mistake fix.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/DataTransactionResult.java
+++ b/src/main/java/org/spongepowered/api/data/DataTransactionResult.java
@@ -330,7 +330,7 @@ public final class DataTransactionResult {
      *
      * @param consumer The consumer to call
      */
-    public void ifSucessful(Consumer<List<ImmutableValue<?>>> consumer) {
+    public void ifSuccessful(Consumer<List<ImmutableValue<?>>> consumer) {
         if (isSuccessful()) {
             try {
                 consumer.accept(this.success);


kashike:
Rename this method to `ifSuccessful`, but add a new method with the old name and deprecate it.

```java


    /**
     * If this result of {@link #isSuccessful()} returns {@code true},
     * the provided {@link Consumer} is called provided a list of all
     * "successful" data as retrieved from {@link #getSuccessfulData()}.
     *
     * @param consumer The consumer to call
     * @deprecated use {@link #ifSuccessful(Consumer)}
     */
    @Deprecated
    public void ifSucessful(Consumer<List<ImmutableValue<?>>> consumer) {
        this.ifSuccessful(consumer);
    }
```

liach:
Report it in #1000 instead, please!

Deamon5550:
No seeing as this is a breaking change it is not suited to the minor issue list.

liach:
eh, why not merge it then?

src/main/java/org/spongepowered/api/data/DataTransactionResult.java
```diff
@@ -330,7 +330,7 @@ public boolean isSuccessful() {
      *
      * @param consumer The consumer to call
      */
-    public void ifSucessful(Consumer<List<ImmutableValue<?>>> consumer) {
```
dualspiral, line 333: This method should be deprecated (with a `@deprecated` tag explaining why in the javadocs), rather than outright removed, and call the correct method - [a bit like what was done here](https://github.com/SpongePowered/SpongeAPI/blob/0477faf48f47f93b83d02a9c9a8370d22d685206/src/main/java/org/spongepowered/api/item/inventory/property/InventoryDimension.java#L39-L44). We'd then remove it two API versions up the road (so, if it was pulled now, we would want to deprecate for API 7, which we'd then remove in API 8)

dualspiral (changes requested):

PR #1566, opened 2017-05-26
Add a message channel which posts MessageChannelEvent

kashike:
Adds https://github.com/SpongePowered/SpongeCommon/issues/1374.

### DIFF
--- a/src/main/java/org/spongepowered/api/text/channel/impl/EventPostingMessageChannel.java
+++ b/src/main/java/org/spongepowered/api/text/channel/impl/EventPostingMessageChannel.java
@@ -1,0 +1,64 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/text/channel/impl/EventPostingMessageChannel.java
```diff
@@ -0,0 +1,64 @@
...
+public class EventPostingMessageChannel extends DelegateMessageChannel {
+
+    public EventPostingMessageChannel(final MessageChannel delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public void send(@Nullable final Object sender, final Text original, final ChatType type) {
+        final MessageChannelEvent event = SpongeEventFactory.createMessageChannelEvent(
+                Cause.of(NamedCause.source(sender), NamedCause.of("Type", type)),
```
kashike, line 53: `sender` is nullable, what should be used as the source when this is the case?

Aquazus: Nothing?

kashike: Exclude a `source` from the `Cause`?

Wundero: You could handle the send method without firing the event if the sender is null, or you could just `return` out of the method without sending the text. I think calling `super.send(...)` is best if the sender is null just to avoid having messages disappear.

Zidane: @kashike when can the sender ever be null?

kashike: It is `null` unless someone passes a sender - a sender isn't required.

LJack2k:
Could we merge this so I can finish my plugin? Would be awesome.

src/main/java/org/spongepowered/api/text/channel/impl/EventPostingMessageChannel.java
Zidane: @kashike that is what I am confused on. How can someone be sent from a sender that doesn't exist? This just reeks bad API imo. I would far prefer having a sender ALWAYS (even for plugin sent messages, the PluginContainer could be the sender itself).

LJack2k: That would be very helpful actually because you then know what plugin send a message.

ryantheleach: I believe you could tell that anyway due to causes?

Zidane: @ryantheleach Not when the sender may be excluded from the cause if a null sender is provided. Which will confuse admins and devs all to hell when messages are being sent by another plugin who passed us a null sender and the event doesn't get fired.

Honestly, enforce the sender. It should have never been optional. This is elementary to fix.

kashike:
Coming with text refactor.

Issue #1560, opened 2017-05-13
Request: Rain-reachable Y-location  


theboomer:
There is no clean way to determine if a player is being rained on (or any location is recieving any precipitation on it) 

The Extent#getHighestYAt() will return the highest solid block that is recieving light, aka, grass blocks under tree canopies, or anything under a glass roof just above or high in the sky.   {At least on SpongeVanilla 6.0.0-Beta-1 this is the implimentation...}

Yet minecraft must know if a location is capable of recieving precipitation or not, since the client will or wont render drops (and splashes on the first block encountered down from the sky) - which may just be a client side thing - BUT, one of the conditions for ICE is that water only freezes if the entire column of air above it is empty and it is able to see precipitation - put tree leaves or glass blocks over, it wont freeze.
As well, the rain-based hydration of soil requires knowing if it is rain-exposed, soil under glass wont hydrate in a rainstorm, soil in the open will.

Please add a Y-position reporter for the block in the column which will recieve rain // is the topmost block in the column solid or transparent, which blocks precipitation from forming in the column below.

parlough:
You should be good now with https://github.com/SpongePowered/SpongeAPI/pull/1572 merged :)

PR #1572, opened 2017-06-04
Add method to get the precipitation level of a column

hfreeb:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1388)

This is a response to the feature request #1560, adding a method to Extent to get the y level that precipitation ends.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/Extent.java
+++ b/src/main/java/org/spongepowered/api/world/extent/Extent.java
@@ -26,6 +26,7 @@ package org.spongepowered.api.world.extent;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.flowpowered.math.vector.Vector2i;
 import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3i;
 import org.spongepowered.api.block.BlockSnapshot;
@@ -117,15 +118,6 @@ public interface Extent extends EntityUniverse, TileEntityVolume, InteractableVo
     int getHighestYAt(int x, int z);
 
     /**
-     * Returns the y level that precipitation ends falling in the given column.
-     *
-     * <p>A value is still returned for columns in biomes which do not receive precipitation.</p>
-     *
-     * @return the y level that precipitation ends
-     */
-    int getPrecipitationLevel(int x, int z);
-
-    /**
      * Get the {@link Location} of the highest block that sunlight can reach in
      * the given column.
      *
@@ -137,6 +129,26 @@ public interface Extent extends EntityUniverse, TileEntityVolume, InteractableVo
      */
     default Vector3i getHighestPositionAt(Vector3i position) {
         return new Vector3i(position.getX(), getHighestYAt(position.getX(), position.getZ()), position.getZ());
+    }
+
+    /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not receive precipitation.</p>
+     *
+     * @return the y level that precipitation ends
+     */
+    int getPrecipitationLevel(int x, int z);
+
+    /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not receive precipitation.</p>
+     *
+     * @return the y level that precipitation ends
+     */
+    default int getPrecipitationLevel(Vector2i column) {
+        return this.getPrecipitationLevel(column.getX(), column.getY());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -117,6 +117,15 @@
     int getHighestYAt(int x, int z);
 
     /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not receive precipitation.</p>
+     *
+     * @return the y level that precipitation ends
+     */
+    int getPrecipitationLevel(int x, int z);
```
kashike: Add another method which accepts `Vector2i` - the new method can be `default`-implemented and call this method.

kashike (changes requested):

src/main/java/org/spongepowered/api/world/extent/Extent.java
hfreeb: Done.

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -131,6 +132,26 @@ default Vector3i getHighestPositionAt(Vector3i position) {
     }
 
     /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not receive precipitation.</p>
```
stephan-gh: Wrap Javadocs after 80 characters. This looks too long.

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -131,6 +132,28 @@ default Vector3i getHighestPositionAt(Vector3i position) {
     }
 
     /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not
+     * receive precipitation.</p>
+     *
+     * @return the y level that precipitation ends
```
parlough: In both methods, wouldn't hurt to change the "the" to have a capital T. Seems to fit better with usual style across Sponge.

parlough (approved):
Looks like a good, simple addition to me.

kashike (approved):

src/main/java/org/spongepowered/api/world/extent/Extent.java
```diff
@@ -131,6 +132,28 @@ default Vector3i getHighestPositionAt(Vector3i position) {
...
+
+    /**
+     * Returns the y level that precipitation ends falling in the given column.
+     *
+     * <p>A value is still returned for columns in biomes which do not
+     * receive precipitation.</p>
+     *
+     * @return The y level that precipitation ends
+     */
+    default int getPrecipitationLevel(Vector2i column) {
```
dualspiral: Should there also be a `default Vector3i getPrecipitationLevelAt(Vector3i position)` method, for consistency with `getHighestYAt` and `getHighestPositionAt`? Similarly, if we're going to have this Vector2i method, should `getHighestYAt` also have a similar `Vector2i` method? It seems odd that we have two similar actions don't match up with each others' signatures.

parlough: I think both of your points are valid, and the first one is a useful addition for sure.

hfreeb: The method isn't anything to do with 3d locations, only 2d columns and it isn't hard to convert a vec3 location to a vec2 of the column: `vec3.toVector2(true)`. Although, for consistency sake, I'm not sure. Opinions @kashike?

parlough: Well it is nice for consistency and people generally work with Vector3i when working with positions like they would be for this. Sure a Vector2i represents a column well, but it's not really what people are using when they would use this method. I suppose they're generally used for chunks.

dualspiral: [The same could be said of the methods I referenced](https://github.com/Desetude/SpongeAPI/blob/16c731482f87594ec4f5ef9b0fd654a51da2fc31/src/main/java/org/spongepowered/api/world/extent/Extent.java#L109-L132) - they aren't really anything to do with 3d locations, but it has a `Vector3i` method for sake of ease. Most developers will have a `Vector3i` to hand, rather than a `Vector2i` when wanting this location. It's also useful to return a `Vector3i` so a developer can use the position immediately without having to construct the vector themselves.

I don't disagree with your point in that the y co-ordinate is irrelavent, certainly, but I'd like to see the method signatures be consistent with one another.

Desetude: Alright, I'll add a vec2 method to getHighestYAt and a vec3 method for getPrecipitationLevel.

hfreeb: If we're going for that level of consistency between the two methods, do you think I should rename `getPrecipitationLevel` to `getPrecipitationLevelAt`?

theboomer:
Thank you Desetude for tackling this :)

parlough (approved):
Looks good to me, would be nice to see this added.

PR #1577, opened 2017-06-11
Preserve the rotation and scale in addTranslation().

### LINKED ISSUE #1557: Transform#addTranslation() doesn't save Scall and Rotation
Hi,
The function Transform#addTranslation return Transform without Scall and Rotation.
[Transform.java](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/entity/Transform.java#L348)

JBYoshi:
Fixes #1557.


kashike (approved):

kashike:
Thanks.

PR #1578, opened 2017-06-13
Add extent factory methods taking the extent min point

Deamon5550:
This adds some methods to the extent factory that allow setting the minimum point of the buffers. Currently you can only create buffers that are based at the origin of the world which makes them more annoying to work with when mirroring some part of the world without applying a transformation at when going between the world and the buffer.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/extent/ExtentBufferFactory.java
+++ b/src/main/java/org/spongepowered/api/world/extent/ExtentBufferFactory.java
@@ -34,10 +34,21 @@ public interface ExtentBufferFactory {
     /**
      * Returns a new biome buffer of the desired size.
      *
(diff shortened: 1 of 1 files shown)


parlough:
Seems like a good addition, would be nice to see it implemented.

kashike:
@Deamon5550

Deamon5550:
@Meronat It is implemented see [SpongeCommin/feature/event-factory-mins](https://github.com/SpongePowered/SpongeCommon/compare/feature/extent-factory-mins) if someone wants to rebase and merge it go ahead, otherwise I'll get it when I have some time.

Deamon5550:
Merged in 6cb5c97d2512a650384981e1e3333c2a46e7b2e7

PR #1591, opened 2017-06-25
Inventory Transfer Events

Faithcaio:
*SpongeAPI* | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1410) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1603) | [Issue](https://github.com/SpongePowered/SpongeCommon/issues/1402)

Adding a PreTransfer event as Hopper are VERY spammy.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
@@ -25,7 +25,10 @@
 package org.spongepowered.api.event.item.inventory;
 
 import org.spongepowered.api.entity.living.Living;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.entity.item.TargetItemEvent;
+import org.spongepowered.api.item.inventory.Inventory;
 import org.spongepowered.api.item.inventory.ItemStack;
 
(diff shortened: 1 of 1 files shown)


bloodmc:
@Faithcaio You could add a Pre/Post as sub interfaces under Transfer instead of Transfer/PreTransfer. I think this would follow our other events better.

src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
```diff
@@ -40,7 +43,45 @@
...
+         * Gets the source {@link Inventory} of this {@link Event}.
+         *
+         * @return The source {@link Inventory}
+         */
+        Inventory getSourceInventory();
+
+        /**
+         * Fired before an {@link Inventory} attempts to transfer items.
+         */
+        interface Pre extends TargetInventoryEvent, Cancellable {
```
Zidane, line 61: This should extend Transfer, Cancellable and remove the methods inside.

Faithcaio: Pre has no SlotTransactions

Zidane: oh, my mistake.

parlough:
Cheers 🍡

PR #1598, opened 2017-07-02
Allow checking if an enchantment is a treasure enchantment or not - Closes #1596

### LINKED ISSUE #1596: Cannot check if an enchantment is a "treasure" enchantment
There are some Minecraft enchantments that cannot be obtained from enchanting in an enchantment table. Internally there is a method to check this and I think it would be good to expose this in the Enchantment interface, so plugin developers can check if vanilla or forge added enchantments are of this type.

I'd PR this, just looking for some input to see if it should be exposed. It would probably be just Enchantment#isTreasureEnchantment

parlough:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1416)

Closes #1596

### DIFF
--- a/src/main/java/org/spongepowered/api/item/Enchantment.java
+++ b/src/main/java/org/spongepowered/api/item/Enchantment.java
@@ -112,4 +112,15 @@ public interface Enchantment extends CatalogType, Translatable {
      * @return Whether these enchantments are compatible
      */
     boolean isCompatibleWith(Enchantment ench);
+
+    /**
+     * Gets whether or not this enchantment is considered a "treasure"
+     * enchantment.
+     *
+     * <p>These do not occur naturally in enchantment tables.</p>
+     *
+     * @return If this is a treasure enchantment
+     */
+    boolean isTreasureEnchantment();
+
 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/item/Enchantment.java
```diff
@@ -112,4 +112,15 @@
...
+
+    /**
+     * Gets whether or not this enchantment is considered a "treasure"
+     * enchantment.
+     *
+     * <p>These do not occur naturally in enchantment tables.</p>
+     *
+     * @return If this is a treasure enchantment
+     */
+    boolean isTreasureEnchantment();
```
kashike: Remove `Enchantment` from the method name.

kashike (changes requested):

Meronat:
I'd be open to a better method name if anyone has one.

src/main/java/org/spongepowered/api/item/Enchantment.java
parlough: I was trying to think of a better name since enchantment seemed redundant, but "isTreasure" doesn't really convey what it is that well.

kashike: `isTreasure` works fine, javadocs exist for a reason.
`enchantment.isTreasure()`.

parlough: Eh it works for me, changing now.

kashike (approved):

PR #1587, opened 2017-06-18
Add new high-level command API

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1400) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1667)

This is a new API and implementation of a high-level command system to bring the design inline with Sponge's general design and to try to resolve some of the perceived weakness of the system. It was one of the things I said I'd look into when I joined the Sponge team... so here we go!

## Motivation & Goals

The command system is out of place in the current ecosystem and parts of it can be difficult to use. It doesn't have much in the way of abstraction, and this is hindering potential development of it. I've also seen quite a few plugins eschew the argument system, preferring to just split the string and doing the parsing themselves. Consider the following points:

* It was built early in Sponge's development (API 2, I believe), and hasn't really seen much daylight since (unlike Texts, which is effectively on version three).
* The argument (parameter) building is just a load of wrapped elements, and can be difficult to follow (one example of this is [Nucleus' TP command](https://github.com/NucleusPowered/Nucleus/blob/sponge-api/5/src/main/java/io/github/nucleuspowered/nucleus/modules/teleport/commands/TeleportCommand.java#L72-L105))
* Many many objects can be created for each command, most of them duplicates of one another - we can try to reuse some simpler objects and potentially reduce Sponge's footprint a little bit.
* It's all API implemented, and cannot take advantage of optimisations through Sponge implementation, nor is it easy to make some changes without causing a major breakage (though this may be by design, see the consideration below)
* There are few interfaces, mostly final classes, and this makes it hard to make a slow transition to newer structures, effectively making it more difficult to make positive changes in a none breaking manner.

So, as a result, I've been playing about with creating a new high level system to try to make it easier to resolve, if not resolve, some of the issues that currently plague the system.

**It's not finished, and it certainly doesn't work right now**, but with community help, we can flesh out a system that is hopefully more flexible and more usable than before.

It's also possible that this ends up being a very bad idea and/or _more_ complex than before and thus doesn't get merged. A major goal is to make the command system, particularly parameters/arguments a little easier to work with, if we can't do that, this PR will become worthless.

Another goal is to modularise the system enough so that we can build other systems around it with ease, for example, some annotation based system could make use of the value parameters independently, and modifiers can be specified independently, without worrying about how the wrapping might otherwise work. However, that's further down the road.

## Consideration: do we _want_ some of this to be put into Common?

[...]

PR #1605, opened 2017-07-12
Add some more arguments to GenericArguments

pie-flavor:
Somehow I managed to completely ruin the git log on #1438, so I'm starting fresh.
Changelog as requested by @dualspiral:

* Change `entity` to allow for `Class` or `EntityType` specifier 
* Add `entityOrTarget` (which can be specified with `Class` or `EntityType`) [`Entity`]
* Add `url` [`URL`]
* Add `ip` [`InetAddress`]
* Add `ipOrSource` [`InetAddress`]
* Add `bigDecimal` [`BigDecimal`]
* Add `bigInteger` [`BigInteger`]
* Add `dataContainer` [`DataContainer`]
* Add `uuid` [`UUID`]
* Add `text` [`Text`]
* Add `dateTime` [`LocalDateTime`]
* Add `dateTimeOrNow` [`LocalDateTime`]
* Add `duration` [`Duration`]
* Add `withSuggestions` (replaces the existing suggestions with new ones) [dynamic]
* Add `withConstrainedSuggestions` (constrains the existing suggestions) [dynamic]


parlough (approved):
Seems like a good addition to me, I will definitely find some of these pretty useful. 

Small nitpick but you forgot a newline at the end of the file.

Not necessarily worth adding a test plugin to common for but could still link to your old test plugin.

pie-flavor:
@Meronat No nitpick is too small.

dualspiral:
Some thoughts: a lot of your javadocs are rather brief and really could do with explaining what format the parameter is expecting. Take your `Duration` parameter as an example, that's a useful parameter, but looking at the javadocs alone, I do not know what format it needs to be in. Same with the time one.

Also, your thinking behind the parameters that output `BigDecimal`, `BigInteger` and `DataContainer`, are these primarily intended for command blocks and other command scripts?

What would be handy is if you could list _all_ of the parameter types you've added, so we don't have to try to pick them out of an already messy class. For me, I can then cross reference and add appropriate methods to #1587, for anyone else interested in reviewing, we can see the intent more clearly.

I see no reason why this, or parts of this, can't go into the next API release subject to review and comments (including those from #1438), because it would be relatively easy for me to transplant this into my commands refactor, which will not be ready for API 7.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1662,10 +1753,532 @@ private static Entity tryReturnSource(CommandSource source, CommandArgs args) th
...
+     * Expect an argument to be valid {@link Text}. Can use either JSON or color codes.
+     *
+     * @param key The key to store under
+     * @param complex If true, parsed as JSON text; if false, parsed as
+     *                color-coded text
+     * @param allRemaining If true, consumes all remaining arguments; if false,
+     *                     uses a single argument
+     * @return the argument
+     */
+    public static CommandElement text(Text key, boolean complex, boolean allRemaining) {
```
dualspiral: Instead of a boolean for the formatting, I think you should just pass in a `TextSerializer` to give maximum flexibility.

Zidane:
@pie-flavor 

Can you add an entity command element that takes an EntityType :)?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
Zidane: @pie-flavor 

Can you cover @dualspiral 's request?

pie-flavor:
Sorry, been away for a while.
@dualspiral I'll fix the javadocs for those posthaste; are there any others you feel that way about? And no, the original motivation for BigInteger and BigDecimal was simply being able to represent arbitrary size/precision numbers (the economy API uses BigDecimal willy-nilly, why can't you get it from commands?). And DataContainer is there because vanilla uses NBT in commands; this enables the same behavior. A good goal for any command API is the ability to fully reimplement all vanilla commands solely using API features.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,125 @@ public Text getUsage(CommandSource src) {
...
+    public static CommandElement entityOrTarget(Text key, EntityType type) {
+        return new EntityCommandElement(key, false, true, type.getEntityClass());
     }
 
     private static class EntityCommandElement extends SelectorCommandElement {
 
-        private final boolean returnSource;
+        protected final boolean returnTarget;
+        protected final boolean returnSource;
+        protected final Class<? extends Entity> type;
```
ST-DDT: This field should be annotated as nullable.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,125 @@ public Text getUsage(CommandSource src) {
...
 
-        protected EntityCommandElement(Text key, boolean returnSource) {
+        protected EntityCommandElement(Text key, boolean returnSource, boolean returnTarget,
+                                       @Nullable Class<? extends Entity> type) {
             super(key);
             this.returnSource = returnSource;
+            this.returnTarget = returnTarget;
+            this.type = type;
```
ST-DDT: It will probably better to automatically fall back to Entity.class

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,125 @@ public Text getUsage(CommandSource src) {
...
+                if (this.returnTarget) {
+                    return tryReturnTarget(source, args, this.type == null ? Entity.class : this.type);
+                }
             }
 
             Object state = args.getState();
             try {
-                return super.parseValue(source, args);
+                Entity entity = (Entity) super.parseValue(source, args);
+                if (!type.isAssignableFrom(entity.getClass())) {
```
ST-DDT: Missing `this.` and potential NPE on `type`.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,125 @@ public Text getUsage(CommandSource src) {
...
+     * @return the argument
+     */
+    public static CommandElement entityOrTarget(Text key, EntityType type) {
+        return new EntityCommandElement(key, false, true, type.getEntityClass());
     }
 
     private static class EntityCommandElement extends SelectorCommandElement {
 
-        private final boolean returnSource;
+        protected final boolean returnTarget;
```
ST-DDT: Why is this field protected? Should be private.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: No extra method call (reflection, no less) if it's just nullable.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1662,10 +1780,547 @@ private static Entity tryReturnSource(CommandSource source, CommandArgs args) th
             }
         }
 
+        private static Entity tryReturnTarget(CommandSource source, CommandArgs args, Class<? extends Entity> type) throws ArgumentParseException {
+            Entity entity = tryReturnSource(source, args);
+            return entity.getWorld().getIntersectingEntities(entity, 10).stream()
+                    .filter(e -> !e.getEntity().equals(entity)).map(EntityUniverse.EntityHit::getEntity)
+                    .filter(e -> !type.isAssignableFrom(e.getClass())).findFirst()
```
ST-DDT: NPE and `this.`

ST-DDT (changes requested):
`EntityCommandElement#tryReturnSource()` needs to filter its result as well.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1643,29 +1754,581 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
...
-                    return ret.get();
+                    Entity entity = ret.get();
+                    if (this.checkEntity(entity)) {
+                        return entity;
+                    }
+                    found = true;
                 }
             }
+            if (found) {
+                throw new IllegalArgumentException("Input value " + choice + " was not an entity of the required type!");
```
dualspiral, line 1791: Why not just do away with the `found` boolean and throw this where you set `found` to `true`?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1643,29 +1754,581 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
...
+            super(key);
+        }
+
+        @Nullable
+        @Override
+        protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
+            String str = args.next();
+            URL url;
+            try {
+                url = new URL(str);
```
dualspiral: Nitpicking really (avoiding duplicate code!): but you could just write `return new URL(str).toURI();` and use a multicatch on the catch block given they give the same exception message.

Not stupidly important, just me being me here.

dualspiral (commented):
A couple of minor, probably very nitpicky things I've spotted whilst on my lunch break!

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
Mumfrey: Because that has different behaviour, setting the semaphore means the loop continues and may find the correct entity in another world. Though not sure if that's possible because it implies there's an entity with a matching uuid but incorrect type in one world, and matching uuid and correct type in another so...

dualspiral: Yeah, I was under the impression that the UUID was unique across the server, not per world. If two different entities with the same UUID can exist in different worlds, then my point can and should be ignored.

JBYoshi: Interesting. Never knew that.

pie-flavor: It's theoretically possible by random chance, much more likely when mapmakers use specific ones on purpose. For instance, imagine a server responsible for managing several instances of a parkour map, which used UUID'd armor stands for fancy effects.

ryantheleach: Also common when you backup and load a world as a copy.

ryantheleach:
This looks like it has been reviewed to death.

We need a decision on if it makes API7 or rolled into the API8 Big Commando Update.

parlough:
I agree. As it stands from my look over it, there is nothing breaking, and this will all be replaced soon enough, so why not allow the extra functionality while we can.

@pie-flavor The one thing I do see is that you unnecessarily messed up the import order.

image

parlough (approved):

dualspiral:
I think this adds value to API 7 and should be included. Linking #1587 to remind me to add these over there.

parlough:
Thanks @pie-flavor I will test these out tonight!

dualspiral (approved):
There is no point holding back a feature from API 7 on the premise that it'll change in API 8. This looks reasonable, let's get this in.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,126 @@ public Text getUsage(CommandSource src) {
...
+    /**
+     * Expect an argument to represent an {@link Entity} of the specified
+     * {@link EntityType}.
+     *
+     * @param key The key to store under
+     * @param type The type which the entity must be
+     * @return the argument
+     */
+    public static CommandElement entity(Text key, EntityType type) {
+        return new EntityCommandElement(key, false, false, type.getEntityClass());
```
JBYoshi: This might not work as expected. Passing `ZOMBIE` will include `PIG_ZOMBIE` as well since `EntityPigZombie` extends `EntityZombie`.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1578,45 +1605,126 @@ public Text getUsage(CommandSource src) {
...
+     * {@link EntityType}, or if the argument is not present and the
+     * {@link CommandSource} is looking at an applicable entity, return that
+     * entity.
+     *
+     * @param key The key to store under
+     * @param type The type which the entity must be
+     * @return the argument
+     */
+    public static CommandElement entityOrTarget(Text key, EntityType type) {
+        return new EntityCommandElement(key, false, true, type.getEntityClass());
```
JBYoshi: See above.

JBYoshi (changes requested):
Looks good to me, except for one thing.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
pie-flavor: TIL. I'll get right on that. If you ask me, though, zombie pigmen extending zombies is the real unexpected behavior here.

parlough:
Tested these and they all seemed to work as intended, had some issues but pie worked through the problems.

PR #1607, opened 2017-07-13
Expose a new InjectionPoint that can be used in Guice providers.

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1435)

This adds a new `InjectionPoint` type that can be injected by Guice, it comes with a fix that is implemented in the Common.

The API side of the issue will be on hold for now, once https://github.com/google/guice/issues/1121 is resolved can we safely include this.

### DIFF
--- a/src/main/java/org/spongepowered/api/inject/InjectionPoint.java
+++ b/src/main/java/org/spongepowered/api/inject/InjectionPoint.java
@@ -30,7 +30,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 
 /**
- * A {@link InjectionPoint} is a method to get extra information when a
+ * An {@link InjectionPoint} is a method to get extra information when a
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/inject/InjectionPoint.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.spongepowered.api.inject;
+
+import com.google.common.reflect.TypeToken;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
+
+/**
+ * A {@link InjectionPoint} is a method to get extra information when a
```
kashike: `An`

kashike (changes requested):

Zidane:
If this API PR still on hold?

PR #1609, opened 2017-07-14
Added a Record API

### LINKED ISSUE #1365: Missing Effects
While working on PoreRT, I noticed that the Sponge API is missing some effects that Bukkit has...
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/Effect.html
- [x] MOBSPAWNER_FLAMES
- [ ] ENDER_SIGNAL
- [x] VILLAGER_PLANT_GROW
- [x] DRAGON_BREATH
- [x] END_GATEWAY_SPAWN
- [x] STEP_SOUND (No easy way of getting the correct sounds for breaking/placing/stepping from a BlockType)
- [ ] RECORD_PLAY (Same issue as STEP_SOUND, but not as much of an issue)

I'll try to implement this this weekend, but we'll see.
Also the Sponge API's effect files seem badly documented ...

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1438)

Fixes #1365

Additions:

- Play/stop a `RecordType` at a block position.
- Retrieve the `RecordType` from a `ItemStack` or `ItemType`.

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/sound/record/RecordType.java
+++ b/src/main/java/org/spongepowered/api/effect/sound/record/RecordType.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/effect/sound/record/RecordType.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.api.block.tileentity.Jukebox;
+import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents the type of record that can be
+ * played by a {@link Jukebox}.
+ */
+@CatalogedBy(RecordTypes.class)
+public interface RecordType extends CatalogType, Translatable {
```
simon816, line 38: This could extend `SoundType`, or provide a method to get the corresponding `SoundType` for the record

kashike: I'd prefer a method to get the `SoundType` rather than extending it.

ST-DDT (approved):

parlough (approved):
Great additions.

VapidLinus:
Would a method to check if a record is playing at a specific position be possible?

Cybermaxke:
@VapidLinus That is sadly enough not possible, when you could get the length of record it would be possible. But resource packs could break this.

PR #1610, opened 2017-07-15
Rewrite inventory queries.

JBYoshi:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1440)

When inventories were first added, they had support for querying for sub-inventories based on various attributes about them. One of those attributes was the type of `ItemStack` they contained. However, sometimes, developers want to use the quantity of the `ItemStack`, and sometimes they want to ignore it but keep the data on it. Originally, the way to specify this was through a quantity of `-1`. Then Minecraft dropped support for negative quantities. Sponge left in its wake a `queryAny()` method that ignored the quantity. In a discussion earlier today, Mumfrey decided this system needed to change, and I decided to be the one to write it. This version exposes a `QueryOperation` that tells the query how to treat each argument. It should also be more flexible than the current system in case Minecraft breaks some other behavior in the future.

Closes https://github.com/SpongePowered/SpongeCommon/issues/1431.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -28,12 +28,16 @@ import org.spongepowered.api.Nameable;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.event.item.inventory.InteractInventoryEvent;
 import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.item.inventory.query.QueryOperation;
+import org.spongepowered.api.item.inventory.query.QueryOperationTypes;
 import org.spongepowered.api.item.inventory.transaction.InventoryTransactionResult;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.text.translation.Translation;
 import org.spongepowered.api.util.ResettableBuilder;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.function.Consumer;
@@ -422,8 +426,16 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
      * @param <T> expected inventory type, specified as generic to allow easy
      *      pseudo-duck-typing
      * @return the query result
+     * @deprecated use {@link #query(QueryOperation...)} instead
      */
-    <T extends Inventory> T query(Class<?>... types);
+    @Deprecated
+    default <T extends Inventory> T query(Class<?>... types) {
+        QueryOperation<?>[] operations = new QueryOperation<?>[types.length];
+        for (int i = 0; i < types.length; i++) {
+            operations[i] = QueryOperationTypes.INVENTORY_TYPE.of(types[i].asSubclass(Inventory.class));
+        }
+        return query(operations);
+    }
 
     /**
      * Query this inventory for inventories containing any of the supplied item
@@ -436,8 +448,16 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
      * @param <T> expected inventory type, specified as generic to allow easy
      *      pseudo-duck-typing
      * @return the query result
+     * @deprecated use {@link #query(QueryOperation...)} instead
      */
-    <T extends Inventory> T query(ItemType... types);
+    @Deprecated
+    default <T extends Inventory> T query(ItemType... types) {
+        QueryOperation<?>[] operations = new QueryOperation<?>[types.length];
+        for (int i = 0; i < types.length; i++) {
+            operations[i] = QueryOperationTypes.ITEM_TYPE.of(types[i]);
+        }
+        return query(operations);
+    }
 
     /**
      * Query this inventory for inventories containing stacks which match the
@@ -451,8 +471,16 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
      * @param <T> expected inventory type, specified as generic to allow easy
      *     pseudo-duck-typing
      * @return the query result
+     * @deprecated use {@link #query(QueryOperation...)} instead
      */
-    <T extends Inventory> T query(ItemStack... types);
+    @Deprecated
+    default <T extends Inventory> T query(ItemStack... types) {
+        QueryOperation<?>[] operations = new QueryOperation<?>[types.length];
+        for (int i = 0; i < types.length; i++) {
+            operations[i] = QueryOperationTypes.ITEM_STACK_EXACT.of(types[i]);
+        }
+        return query(operations);
+    }
 
     /**
      * Query this inventory for inventories which match any of the supplied
@@ -467,8 +495,16 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
      * @param <T> expected inventory type, specified as generic to allow easy
      *      pseudo-duck-typing
      * @return the query result
+     * @deprecated use {@link #query(QueryOperation...)} instead
      */
-    <T extends Inventory> T query(InventoryProperty<?, ?>... props);
+    @Deprecated
+    default <T extends Inventory> T query(InventoryProperty<?, ?>... props) {
+        QueryOperation<?>[] operations = new QueryOperation<?>[props.length];
+        for (int i = 0; i < props.length; i++) {
+            operations[i] = QueryOperationTypes.INVENTORY_PROPERTY.of(props[i]);
+        }
+        return query(operations);
+    }
 
     /**
      * Query this inventory for inventories matching any of the supplied titles.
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -377,118 +377,15 @@
...
-     * <code>OR</code> is applied between operands.
-     * This ignores stack sizes. To query for stacks of a specific size use
-     * {@link #query(ItemStack...)}.
-     *
-     * @param types items to query for, the size of the stacks is always ignored
-     * @param <T> expected inventory type, specified as generic to allow easy
-     *      pseudo-duck-typing
-     * @return the query result
-     */
-    <T extends Inventory> T queryAny(ItemStack... types);
```
gabizou: All of these methods could technically be `@Deprecated` with redirects to the new query operation method, with maybe the difference of the `query(Object...)` method.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
```diff
@@ -0,0 +1,28 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.query;
+
+public interface QueryOperation {
```
gabizou: missing some information? or am I looking at a template?

gabizou (changes requested):
Deprecation policy needs to be abided here.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
JBYoshi: I got most of them, except for two: the `Translation` one and the `String` one. The `String` one wasn't even implemented, but I'll get the `Translation` one tomorrow.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
JBYoshi: I added some Javadoc to the class. That should help explain it.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
JBYoshi: I re-added `query(Translation...)`.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperationTypes.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.function.Predicate;
+
+public final class QueryOperationTypes {
+
+    // SORTFIELDS:ON
+
+    public static final QueryOperationType<InventoryProperty<?, ?>> INVENTORY_PROPERTY = DummyObjectProvider.createFor(QueryOperationType.class,
+            "INVENTORY_PROPERTY");
+
+    public static final QueryOperationType<Translation> INVENTORY_TITLE = DummyObjectProvider.createFor(QueryOperationType.class, "INVENTORY_TITLE");
```
kashike: The [inventory title](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/item/inventory/property/InventoryTitle.java#L44) is a `Text`.

kashike: The [inventory title](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/item/inventory/property/InventoryTitle.java#L44) is a `Text`.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperationTypes.java
JBYoshi: Actually, if it's an `InventoryProperty`, this field isn't necessary - `INVENTORY_PROPERY` would work just fine.

kashike: This shouldn't be called `INVENTORY_TITLE`, then. It should be called `INVENTORY_TRANSLATION`.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -452,27 +475,16 @@
      * @param <T> expected inventory type, specified as generic to allow easy
      *      pseudo-duck-typing
      * @return the query result
-     */
-    <T extends Inventory> T query(String... names);
```
gabizou, line 493: I'm confused, why is this going from `String` to `Translation`? The point of deprecating the old methods is to provide an easier upgrade path.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -452,27 +475,16 @@
...
-     * individual inventory. A naive match may be obtained by calling .equals()
-     * against the child inventory passing the unknown operand as an argument.
-     * </p>
-     *
-     * @param args search parameters
-     * @param <T> expected inventory type, specified as generic to allow easy
-     *      pseudo-duck-typing
-     * @return the query result
-     */
-    <T extends Inventory> T query(Object... args);
```
gabizou: Where did this go?

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory.query;
+
+/**
+ * Represents an operation that is part of a query. The details of the
+ * filtering are implementation-specific and are not exposed in this
+ * interface.
+ */
+public interface QueryOperation {
```
gabizou: What's the point of exposing the interface if there's nothing in it?

gabizou (commented):
I might be misreading some changes, but some things are disappearing from what I can see.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
kashike: Better than just accepting `query(Object...`, limits it to query operations. Also prevents breaking the return type if something *is* added.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
JBYoshi: That would be a diff bug. There was a `Translation` method originally; it was posted just above this method, but the diff thinks the original method was part of the `InventoryProperty` entry.

I removed the `String` version both because it was never implemented and because it was inconsistently referred to: the API referred to it as names, but the implementation referred to it as expressions.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
JBYoshi: I would put something in it, except that "something" would end up referring to lenses and fabrics and other implementation details like that.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
JBYoshi: Probably to `/dev/null`, which seems to be a very broken peripheral hidden inside my computer somewhere; I can't even read from it. Let me see if I can get the code replaced with something.

Zidane:
@JBYoshi Is this ready to go?

@Faithcaio @Mumfrey How is this looking?

JBYoshi:
This should be ready, unless anyone requests changes.

src/main/java/org/spongepowered/api/item/inventory/query/QueryOperation.java
ST-DDT: Maybe add an `getType()`?

ST-DDT (approved):
I would add a newline after the last member/method in every class/interface.

I would also add a brief description/javadocs to each of the `QueryOperationType`s. (ie: `Exact ItemStack match including all data such as Enchantments and AttackModifiers`).

With the new structure it should be possible to create `OR`, `AND` and similar operator types. (ie: Search for Gold ignots OR Gold nuggets. Or search for all armor that is not damaged.) Without the need to fallback to the/a complex itemstack predicate.

Faithcaio:
As query(Class...) is used very often I think it could be worth keeping it in the API. (maybe others too?)

query(Object...) should go in favor of the QueryOperations

Issue #1615, opened 2017-07-19
Check if inventory contains slot


Faithcaio:
It is a pain to try and find out if a slot in inside an inventory in an event.
E.g. Check if clicked slot is in player inventory.

Therefore I propose adding `Inventory#containsSlot(Slot)` (or Inventory?) to the API

@Mumfrey @Zidane

Mumfrey:
I'd personally prefer `hasChild(Inventory)` because it's consistent with the general contract of inventories (eg. that everything is an inventory) but also has utility in case you have any other kind of subinventory.

Faithcaio:
hasChild would only be for direct subinventories?

Mumfrey:
My interpretation would be that it means the child is anywhere in the descendent tree of the inventory, and that it could be used to determine whether a Slot appears in an Inventory, a Row appears in a Grid, etc. a Row appears in an Inventory, etc.

Your stated goal is "is Slot in Inventory", but the general case of that is "is Inventory in Inventory" is what I was saying. If you prefer `Inventory::containsInventory` rather than `Inventory::containsChild` then I'm ok with that. It's the contract I was more interested in than the name.

Mumfrey:
I suppose the question at that points is whether `containsInventory` respects the identity function. Eg. does `someInventory.containsInventory(someInventory)` return true?

Faithcaio:
so that should work too?
`inventory.query(firstrow).containsInventory(inventory.query(stoneitems))`
basicly checking if each slot is contained in the other inventory

Mumfrey:
The contract of query states that if the query matches every child of a particular inventory, then the inventory is returned. So in your example above the `inventory.query(stoneItems)` were to return a single slot or multiple slots there are two possibilities for the contract of `containsInventory`:

1. That `containsInventory` returns `true` if the supplied `Inventory` is anywhere in the inventory's heirarchy. This would return `true` for a single slot, or `true` if the query matched all slots in the row and thus returned the row itself, but `false` for multiple slots since they'd be returned in a `ResultInventory` which isn't a child of the originally queries inventory, unless

2. The `containsInventory` returns `true` if the supplied `Inventory` **or** all of its children. In which case it would return `true` in the scenarios above but also in the situation that all the slots in the result were children of the row.

Personally I think the first, simpler contract is better because it's more obvious what it does and satisfies the main use-case of checking whether a particular inventory is a child of another.

Faithcaio:
Why not both?
[...]

PR #1616, opened 2017-07-19
Inventory set operations

Faithcaio:
*SpongeAPI* | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1451)

Adds `intersect` `union` and `containsInventory` to the Inventory API

Issue: https://github.com/SpongePowered/SpongeAPI/issues/1615

@Mumfrey

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -505,8 +505,9 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
     InventoryArchetype getArchetype();
 
     /**
-     * Intersects the slots of both inventories
-     * The resulting inventory will only contain slots.
+     * Intersects the slots of both inventories.
+     * The resulting inventory will only contain slots
+     * that are present in both inventories.
      *
      * @param inventory the other inventory
      * @return an inventory wrapping all slots that are present in both inventories
@@ -515,7 +516,8 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
 
     /**
      * Constructs a union of the slots in both inventories.
-     * The resulting inventory will only contain slots.
+     * The resulting inventory will contain all slots
+     * from both inventories.
      *
      * @param inventory the other inventory
      * @return an inventory wrapping all slots of both inventories.
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -505,6 +505,33 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
     InventoryArchetype getArchetype();
 
     /**
+     * Intersects the slots of both inventories
+     * The resulting inventory will only contain slots.
+     *
+     * @param inventory the other inventory
+     * @return an inventory wrapping all slots that are present in both inventories
+     */
+    Inventory intersect(Inventory inventory);
+
+    /**
+     * Constructs a union of the slots in both inventories.
+     * The resulting inventory will only contain slots.
+     *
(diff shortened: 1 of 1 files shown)


ryantheleach:
@Mumfrey What about hotbar.except(first) ? where hotbar is the hotbar inventory, and first is the first hotbar slot?

(vs using the term complement, which is well known to Set theorists, but except is more known in SQL among programmers)

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -505,6 +505,33 @@
...
+    Inventory union(Inventory inventory);
+
+    /**
+     * Returns true if the given inventory is a descendant of this one.
+     * This includes deeply nested ones but not usually not query-results.
+     *
+     * @param inventory the other inventory
+     * @return whether the given inventory is contained in this one.
+     */
+    boolean containsInventory(Inventory inventory);
```
ST-DDT, line 544: I would rephrase or extend the javadoc by something like:
* `an inventory is contained in another if all slots are contained in the parent`
* `an inventory is a decendant of an inventory if all slots of the decendant are contained in the parent`

This method can also be default implemented using:
`this.union(inventory).capacity() == this.capacity()`

Faithcaio: ContainsInventory will not operate on each slot. 
So you usually using queryresults as parameter will return false.
Except if you check e.g. for a row to be contained in a gridinventory
or a single slot in an ordered inventory

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -505,6 +505,33 @@
     InventoryArchetype getArchetype();
 
     /**
+     * Intersects the slots of both inventories
```
Zidane: This needs more documentation as it isn't clear what is going on whatsoever here.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -505,6 +505,35 @@
...
+     * from both inventories.
+     *
+     * @param inventory the other inventory
+     * @return an inventory wrapping all slots of both inventories.
+     */
+    Inventory union(Inventory inventory);
+
+    /**
+     * Returns true if the given inventory is a descendant of this one.
+     * This includes deeply nested ones but not usually not query-results.
```
Zidane: This needs further explanation...

It doesn't affect queries or?

Faithcaio: Maybe @Mumfrey can find a good way to describe it?

ryantheleach: Query Results are essentially a new Inventory View created to solve the query.

The Query Result returned, may, or may not be, a sub inventory of an existing inventory depending on the query, or it may be an amalgamation from various sources.

If you have a reference to a sub-inventory, this will work in contains.

However, after sleeping on this after we talked @Faithcaio I'm not sure this is entirely expected from user-developers.

Maybe @ST-DDT Is correct, in that we should essentially run contains for all of the smallest inventory views contained within the query result, as I think there are situations where you could end up with a custom view of a compound inventory made of multiple slots, that all belong to a single parent.

Unless you want to change the signature of query results to be a collection like class of inventories anyway, to make this explicit.

Faithcaio: @ryantheleach if you wanted to operate on slots you would just use the intersect method

kashike:
https://github.com/SpongePowered/SpongeAPI/commit/4e8d9e92cb4ff4a4c3a877dfcd454a64db8d3f92

PR #1620, opened 2017-07-28
Custom Data Generation

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1772)

This is a continuation of the PR #1198, this is still WIP

The basic concept is still there, but I redesigned the builder pattern to allow more effective usage of generics. This builder system also supports the generation of `VariantData`, `ListData` and `MappedData` on top of the default `DataManipulator`, altough these only support one `Key`.

A example plugin:
```java
@Plugin(id = "data_test", name = "Data Test", description = "A plugin to custom data.")
public class DataTest {

    public static Key<ListValue<String>> LINES_KEY = DummyObjectProvider.createFor(Key.class, "LINES_KEY");
    public static Key<Value<String>> AUTHOR_KEY = DummyObjectProvider.createFor(Key.class, "AUTHOR_KEY");
    public static Key<OptionalValue<String>> PUBLISHER_KEY = DummyObjectProvider.createFor(Key.class, "PUBLISHER_KEY");
    public static Key<Value<Integer>> REVISIONS_KEY = DummyObjectProvider.createFor(Key.class, "REVISIONS_KEY");
    public static Key<Value<Boolean>> MY_BOOLEAN_KEY = DummyObjectProvider.createFor(Key.class, "MY_BOOLEAN_KEY");

    public static DataRegistration<BookData, ImmutableBookData> MY_BOOK_DATA;
    public static DataRegistration<? extends VariantData<Boolean, ?, ?>, ? extends ImmutableVariantData<Boolean, ?, ?>> MY_BOOLEAN_DATA;

    @Listener
    public void onRegisterKeys(GameRegistryEvent.Register<Key<?>> event) {
        LINES_KEY = Key.builder()
                .type(new TypeToken<ListValue<String>>() {})
                .query(DataQuery.of("Lines"))
                .name("Lines")
                .id("lines")
                .build();
        event.register(LINES_KEY);
        AUTHOR_KEY = Key.builder()
                .type(new TypeToken<Value<String>>() {})
                .query(DataQuery.of("Author"))
                .name("Author")
                .id("author")
                .build();
        event.register(AUTHOR_KEY);
        PUBLISHER_KEY = Key.builder()
                .type(new TypeToken<OptionalValue<String>>() {})
                .query(DataQuery.of("Publisher"))
                .name("Publisher")
                .id("publisher")
                .build();
        event.register(PUBLISHER_KEY);
        REVISIONS_KEY = Key.builder()
                .type(new TypeToken<Value<Integer>>() {})
                .query(DataQuery.of("Revisions"))
                .name("Revisions")
                .id("revisions")
                .build();
        event.register(REVISIONS_KEY);
        MY_BOOLEAN_KEY = Key.builder()
                .type(new TypeToken<Value<Boolean>>() {})
                .query(DataQuery.of("MyBoolean"))
                .name("MyBoolean")
                .id("my_boolean")
                .build();
        event.register(MY_BOOLEAN_KEY);
    }

    @Listener
    public void onRegisterData(GameRegistryEvent.Register<DataRegistration<?,?>> event) {
        MY_BOOK_DATA = GenericDataGenerator.builder()
                .key(LINES_KEY, new ArrayList<>())
                .key(AUTHOR_KEY, "Unknown")
                .key(PUBLISHER_KEY, Optional.empty())
                .key(REVISIONS_KEY, 1)
                .interfaces(BookData.class, ImmutableBookData.class)
                .id("book_data")
                .name("Book Data")
                .build();
        event.register(MY_BOOK_DATA);

        final BookData bookData = MY_BOOK_DATA.getDataManipulatorBuilder().create();
        System.out.println("DEFAULT AUTHOR: " + bookData.getAuthor());
        System.out.println("DEFAULT PUBLISHER: " + bookData.getPublisher());
        bookData.setAuthor("Cybermaxke");
        bookData.setPublisher("Cybermaxke");
        System.out.println("NEW AUTHOR: " + bookData.getAuthor());
        System.out.println("NEW PUBLISHER: " + bookData.getPublisher());

        MY_BOOLEAN_DATA = VariantDataGenerator.builder()
                .key(MY_BOOLEAN_KEY)
                .defaultValue(false)
                .id("my_boolean_data")
                .name("My Boolean Data")
                .build();
        event.register(MY_BOOLEAN_DATA);
    }
}
```
```java
public interface BookData extends DataManipulator<BookData, ImmutableBookData> {

    @KeyValue("lines")
    List<String> getLines();

    @KeyValue("author")
    String getAuthor();

    @KeyValue("author")
    void setAuthor(String author);

    @KeyValue("publisher")
    Optional<String> getPublisher();

    @KeyValue("publisher")
    void setPublisher(@Nullable String publisher);
}
```
```java
public interface ImmutableBookData extends ImmutableDataManipulator<ImmutableBookData, BookData> {

}
```


src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -61,6 +61,22 @@
     Class<I> getImmutableManipulatorClass();
 
     /**
+     * Gets the {@link DataManipulator} class for this registration
+     * that implements the {@link #getManipulatorClass()}.
+     *
+     * @return The manipulator implementation class
+     */
+    Class<T> getImplementationManipulatorClass();
```
kashike: `getManipulatorImplementationClass`?

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -61,6 +61,22 @@
...
+
+    /**
+     * Gets the {@link ImmutableDataManipulator} class for this registration
+     * that implements the {@link #getImmutableManipulatorClass()}.
+     *
+     * @return The immutable manipulator implementation class
+     */
+    Class<I> getImplementationImmutableManipulatorClass();
```
kashike: `getImmutableManipulatorImplementationClass`?

kashike (changes requested):

src/main/java/org/spongepowered/api/data/DataRegistration.java
Mumfrey: `getTotallyImmutableImplementationClassFactoryProviderModelDescriptorCriterionScope` ?

Faithcaio: `mixinRedirectGetTotallyImmutableImplementationClassFactoryProviderModelDescriptorCriterionScopeInvocationReturnableOriginMethodInterface` !

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -181,6 +181,11 @@ default ItemType getItem() {
          */
         Builder itemType(ItemType itemType);
 
+        /**
```
Zidane: Remove from this PR, just adds noise.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,405 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+
+/**
+ * A generator that allows a {@link DataRegistration} to be constructed with
+ * generated {@link DataManipulator} and {@link ImmutableDataManipulator} classes.
+ */
+public interface DataGenerator extends ResettableBuilder<DataRegistration<?, ?>, DataGenerator.TypeBuilder> {
```
Zidane: @gabizou 

In keeping with the themes found elsewhere, should this go on to DataManipulator? Not sure I like having the builder separate like this..

Zidane: On second thought, fair bit of stuff is happening in here. Hmm..

src/main/java/org/spongepowered/api/data/generator/testing/dummy/DummyManipulatorExample.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.api.data.key.Key;
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.util.TypeTokens;
+
+import java.util.ArrayList;
+
+public class DummyManipulatorExample {
```
gabizou: All the example usages can be moved to implementation test now.

src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractListData.java
```diff
@@ -61,10 +67,10 @@ protected AbstractListData(List<E> value, Key<? extends BaseValue<List<E>>> used
     }
 
     @Override
-    public <V> Optional<V> get(Key<? extends BaseValue<V>> key) {
+    public <E> Optional<E> get(Key<? extends BaseValue<E>> key) {
```
gabizou: The reason why the generic was using `V` was so it wouldn't be hiding the class generic of `E`.

bloodmc:
@gabizou Can you review this?

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -40,8 +41,12 @@
      * the desired {@link DataManipulator} type to be registered.
      *
      * @return The new builder instance
+     * @deprecated Use the {@link DataGenerator} instead, this is the new
+     *    method of generating {@link DataManipulator} implementations and
```
gabizou: Why would this be removed from the API? The point of dynamic class generation is not to wholeheartedly replace its method of building (since sponge implementation is likely to build with it).

src/main/java/org/spongepowered/api/data/generator/testing/SimpleExamples.java
```diff
@@ -0,0 +1,94 @@
...
+import org.spongepowered.api.data.manipulator.mutable.VariantData;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.MapValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.util.TypeTokens;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class SimpleExamples {
```
gabizou: This should be added to the implementation test classes/plug-ins once the implementation branch is created

src/main/java/org/spongepowered/api/data/generator/testing/dummy/DummyManipulator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.spongepowered.api.data.generator.KeyValue;
+import org.spongepowered.api.data.meta.ItemEnchantment;
+import org.spongepowered.api.data.value.BoundedValue;
+import org.spongepowered.api.data.value.mutable.ListValue;
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
+import org.spongepowered.api.data.value.mutable.Value;
+
+import java.util.List;
+
+public interface DummyManipulator extends DataManipulator<DummyManipulator, ImmutableDummyManipulator> {
```
gabizou: Likewise this.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
```diff
@@ -114,16 +119,45 @@ public int hashCode() {
...
         };
     }
 
+    public static <E> Key<Value<E>> makeValueKey(TypeToken<E> elementToken,
+            DataQuery query, String id, String name) {
+        final TypeToken<Value<E>> valueToken = new TypeToken<Value<E>>() {}
+                .where(new TypeParameter<E>() {}, elementToken);
+        return makeSingleKey(elementToken, valueToken, query, id, name);
+    }
+
```
gabizou: Need some java docs to differentiate what the different variations of the methods are.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
```diff
@@ -196,53 +196,37 @@ public String toString() {
...
+     * @param name The name for the new key
+     * @param <K> The type of keys
+     * @param <V> The type of values
+     * @return The generated key
+     */
+    public static <K, V> Key<MapValue<K, V>> makeMapKeyWithKeyAndValue(TypeToken<K> keyToken, TypeToken<V> valueToken,
+            DataQuery query, String id, String name) {
+        final TypeToken<Map<K, V>> elementToken = new TypeToken<Map<K, V>>() {}
+                .where(new TypeParameter<K>() {}, keyToken)
+                .where(new TypeParameter<V>() {}, valueToken);
```
gabizou: This won’t save your generics as they were initially intended. It’s going to get raw generics at runtime and compile time due to the lack of the creation at the compilation level.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
```diff
@@ -259,124 +243,55 @@ public String toString() {
...
-            public String toString() {
-                if (this.string == null) {
-                    this.string = "Key{Value:" + "OptionalValue<" + elementToken.toString() + ">, Query: " + query.toString() + "}";
-                }
-                return this.string;
-            }
-        };
+    public static <E> Key<OptionalValue<E>> makeOptionalKey(TypeToken<E> elementToken,
+            DataQuery query, String id, String name) {
+        final TypeToken<Optional<E>> elementToken0 = new TypeToken<Optional<E>>() {}
```
gabizou: Likewise here.

src/main/java/org/spongepowered/api/data/manipulator/immutable/common/AbstractImmutableData.java
```diff
@@ -49,12 +49,17 @@
  *
  * @param <I> The immutable data manipulator type
  * @param <M> The mutable manipulator type
+ * @deprecated These classes are only to be used for easing the compatibility requirements
+ * for plugin developers moving to the new system introduced by
```
gabizou: Import the data generator class to eliminate the long line.

gabizou (changes requested):
Minor changes, some of the TypeToken class generation wont maintain the generic typing as intended due to the creation within type erased compiled classes.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
Zidane: @gabizou ^ ?

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
Cybermaxke: @gabizou This is done the way it's explained on the guava wiki, and I tested it and it works at runtime. https://github.com/google/guava/wiki/ReflectionExplained

src/main/java/org/spongepowered/api/data/DataRegistration.java
Cybermaxke: @gabizou If it's just going to be used by SpongeCommon, etc., why not move away from the builder and put it in the implementation. I thought you wanted to get rid of all the AbstractData classes, unless you still want to support custom implemented `DataManipulator`'s?

gabizou: Maybe not specifically in implementation, but I don't see why the actual builder method is being removed if it still can be used, whether developers choose to implement custom data manipulators themselves, or using dynamic class generation through the API.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
gabizou: Oh good, I was under the belief that generic typed type tokens like these wouldn't work very well with type loss.

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -90,6 +90,20 @@
      */
     PluginContainer getPluginContainer();
 
+    /**
+     * Constructs a default {@link DataManipulator} of type {@link T}.
+     *
+     * @return The constructed data manipulator
+     */
+    T constructManipulator();
```
gabizou: maybe `createDefault()`

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -90,6 +90,20 @@
      */
     PluginContainer getPluginContainer();
 
+    /**
+     * Constructs a default {@link DataManipulator} of type {@link T}.
```
gabizou: `@link T` will not work, use `@code T`

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -90,6 +90,20 @@
...
+     * @return The constructed data manipulator
+     */
+    T constructManipulator();
+
+    /**
+     * Constructs a default {@link ImmutableDataManipulator} of type {@link I}.
+     *
+     * @return The constructed immutable data manipulator
+     */
+    I constructImmutableManipulator();
```
gabizou: Maybe `createDefaultImmutable()`

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -90,6 +90,20 @@
...
 
+    /**
+     * Constructs a default {@link DataManipulator} of type {@link T}.
+     *
+     * @return The constructed data manipulator
+     */
+    T constructManipulator();
+
+    /**
+     * Constructs a default {@link ImmutableDataManipulator} of type {@link I}.
```
gabizou: Same here.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * The nature of the builder doesn't allow values of data registration to be
+     * reused. It is also unlikely that {@link Key}s names and manipulator types
+     * will be reused in multiple {@link DataRegistration}s. Use {@link #reset()}
+     * and resupply the builder instead.
+     *
+     * @param value The built object
+     * @return This builder, for chaining
+     */
+    @Deprecated
```
gabizou: If it is deprecated, should have the `@deprecated` java doc in the description to explain why it's deprecated.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+    G version(int contentVersion);
+
+    /**
+     * Defines a {@link Predicate} that checks whether the supplied
+     * {@link DataManipulator} type is supported by the {@link DataHolder}.
+     *
+     * @param predicate The predicate
+     * @return This builder, for chaining
+     */
+    G predicate(Predicate<? extends DataHolder> predicate);
```
Cybermaxke: @gabizou Is this still something you want from your original PR? It would also require support for non generated `DataRegistration`s.

Cybermaxke:
@gabizou There should also be a way to move over from the abstract data classes to the data generators without losing any data. Since the serialization method is internal, what could be a good way to do this?
Maybe a "legacy" builder interface where they can apply their old deserialization methods?

Cybermaxke:
Diving further into the upgradeability, how could we handle changing the value type of a key? This would also break serialization, and devs cannot update manually because serialization isn't exposed.

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -226,18 +269,28 @@
...
+         * @deprecated Use {@link #build()} and the {@link Register} event instead
          */
+        @Deprecated
         DataRegistration<T, I> buildAndRegister(PluginContainer container) throws IllegalStateException, IllegalArgumentException,
                                                                                   DataAlreadyRegisteredException;
 
+        /**
+         * Builds the {@link DataRegistration} object, this registration must be
+         * registered in the {@link Register} event.
+         * <p>The {@link PluginContainer} that is currently in the
```
gabizou: Line before `<p>` and you're missing the closing `</p>`

src/main/java/org/spongepowered/api/data/DataView.java
```diff
@@ -172,13 +173,24 @@ default boolean contains(Key<?> key, Key<?>... keys) {
...
+     * {@link Optional#empty()} is returned.
      *
      * @param path The path to the Object
      * @return The Object, if available
      */
     Optional<Object> get(DataQuery path);
 
+    /**
+     * Gets an object from the desired path and deserializes the data using
+     * the given {@link TypeToken}. If the path is not defined,
```
gabizou, line 185: How does the serialization rule work? Is it in the case of it works by attempting to find it as a `DataSerializable`, something that has a registered `DataTranslator`?

src/main/java/org/spongepowered/api/data/DataView.java
```diff
@@ -189,6 +201,18 @@ default boolean contains(Key<?> key, Key<?>... keys) {
      */
     DataView set(DataQuery path, Object value);
 
+    /**
+     * Sets the given Object value according to the given path relative to
+     * this {@link DataView}'s path and serializes the data using the
+     * given {@link TypeToken}.
```
gabizou, line 207: How does this differ from what the above set method does? Ideally the set method should be the only used method since it can derive almost all information on the object.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * A generator that allows a {@link DataRegistration} to be constructed with
+ * generated {@link DataManipulator} and {@link ImmutableDataManipulator} classes.
+ */
+public interface DataGenerator<
+        M extends DataManipulator<M, I>,
+        I extends ImmutableDataManipulator<I, M>,
+        G extends DataGenerator<M, I, G, R>,
+        R extends DataGenerator<?, ?, ?, R>>
```
gabizou, line 48: Need some explanation for some of these generics.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
gabizou: Ideally yes, since this can be a functional way to allow plugin developers to determine what data holders they can be added to.

src/main/java/org/spongepowered/api/data/DataView.java
gabizou: Personally this doesn't make much sense otherwise.

src/main/java/org/spongepowered/api/data/DataView.java
gabizou: This shouldn't exist, Either the same logic has to apply in `set(DataQuery,Object)` or it should be split up to support different types. I'm not seeing the benefit of the type token personally.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+        extends ResettableBuilder<DataRegistration<?,?>, R> {
```
gabizou, line 49: Like specifically why the `R` random type.

src/main/java/org/spongepowered/api/data/generator/KeyValue.java
```diff
@@ -0,0 +1,144 @@
...
+ * <code>E</code></p>. Primitive values can also be automatically boxed.
+ * In a {@link ImmutableDataManipulator} is a return type required, this will return the newly
+ * constructed immutable manipulator.
+ * You can also return a manipulator in a{@link DataManipulator}, this will always return the
+ * object itself and allows chaining operations. The return type isn't required in this case.
+ * <p>For example:
+ * <pre>
+ * {@code
+ * // The key
+ * Key<Value<Integer>> MY_INT_KEY = Key.builder()
```
gabizou: This sort of formatting isn't well supported by java docs, if you can write the html output for it correctly, it should work. Somewhat like so: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/entity/DamageEntityEvent.java#L114
which produces nice code blocked code.

src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractData.java
```diff
@@ -93,6 +99,7 @@ protected AbstractData() {
      * @param key The key for the value return type
      * @param function The function for getting the value
      */
+    @Deprecated
```
gabizou: No sense in deprecating the methods. The entire class is already deprecated.

src/main/java/org/spongepowered/api/util/Copyable.java
```diff
@@ -0,0 +1,40 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+/**
+ * Represents something that can be copied.
+ *
+ * @param <T> The type
+ */
+public interface Copyable<T extends Copyable<T>> {
```
gabizou, line 32: The point of this being? I understand making `DataManipulator` copyable, but not the base `ValueContainer`? It seems like a duplicated method handle.

gabizou (changes requested):

gabizou:
Also, if we are making Key builders not register on build, we don't need to dummy object provider them in the example, we can simply register them at the appropriate time.

src/main/java/org/spongepowered/api/util/Copyable.java
Cybermaxke: This is to allow other objects to be also copyable, for example objects provided by plugins or `ItemStack`s. These objects will be copied when copying a `DataManipulator` (or immutable version).

src/main/java/org/spongepowered/api/data/DataView.java
Cybermaxke: It will be mainly used to support generic types in `Map`s, `List`s, etc. Similar how it is serialized/deserialized in configurate.

src/main/java/org/spongepowered/api/data/generator/DataGenerator.java
Cybermaxke: This one is used when resetting the builder, to reset the generic types as well.

src/main/java/org/spongepowered/api/data/DataView.java
gabizou: How is it different? It honestly shouldn’t because it is telling me that it is already duplicated code now.

XakepSDK:
How content version migration is handled?

gabizou:
@xakepsdk what I’m planning on doing is having a semi functional interface by registering upgrades as you would do otherwise normally.

gabizou:
So, thoughts on data version upgrades:
Extract the ContentUpdater with a DataViewModifier which would be a functional interface. Then, the interface could be registered with integer version numbers as either a builder or just a simple multi-arg method like so:
```java
  @Listener
  public void onDataRegister(GameRegistryEvent.Register<DataRegistration<?>> event) {
     MY_BOOK_DATA = GenericDataGenerator.builder()
                .key(LINES_KEY, new ArrayList<>())
                .key(AUTHOR_KEY, "Unknown")
                .key(PUBLISHER_KEY, Optional.empty())
                .key(REVISIONS_KEY, 1)
                .interfaces(BookData.class, ImmutableBookData.class)
                .id("book_data")
                .name("Book Data")
                .build();
        event.register(MY_BOOK_DATA);
  }

  @Listener
  public void onContentUpdaterRegister(DataRegistryEvent.RegisterUpdaters event) {
    event.createUpdater(MY_BOOK_DATA, Versions.LEGACY, Versions.CURRENT, (view) -> {
      final int copyCount = view.getInt(Queries.LEGACY_COPY_COUNT).orElse(0);
      view.remove(Queries.LEGACY_COPY_COUNT);
      view.put(Queries.COPY_COUNT, copyCount);
      return view;
    }
  }
```

This way the content upgraders would be registered both through the DataRegistration object, since the manipulator classes are provided through that, or it can be made through a ContentUpdater.Builder object that would do the same thing here. I have to give the API a look over again and do some usability tests (not really that hard) and *eventually* get the implementation working.

gabizou:
Closing since DataManipulators are nuked in api-8 for much better simplification.

PR #1622, opened 2017-08-05
Allow sending a specific page of a pagination list

parlough:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1475)

Replaces https://github.com/SpongePowered/SpongeAPI/pull/1539

Also cleans up some Javadocs, adds some convenience methods, and organizes the class.

There is a breaking change in that the sendTo method that accepts a `MessageChannel` in the builder was removed as it makes it seem like it sends it to the channel as one, but really it just sends it to each member separately. I doubt there is any usage of this method, and they can easily switch to the method which accepts an `Iterable` of `MessageReceiver`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
+++ b/src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.api.service.pagination;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.channel.MessageChannel;
@@ -36,7 +38,8 @@ import java.util.Optional;
 import javax.annotation.Nullable;
 
 /**
- * Represents an immutable iterable of {@link Text}s, which can be sent to a {@link MessageReceiver}.
+ * Represents an immutable iterable of {@link Text}s, which can be sent to
+ * a {@link MessageReceiver}.
  *
  * <p>An instance of this class may be obtained using {@link Builder}.</p>
  */
@@ -69,8 +72,8 @@ public interface PaginationList {
      * Gets the header to be displayed for this output on all pages after the
      * title bar but before the contents, if available.
      *
-     * <p>Header and footer will use this Text's style and color for formatting.
-     * </p>
+     * <p>Header and footer will use this Text's style and color for
+     * formatting.</p>
      *
      * @return The header to be displayed
      */
@@ -92,29 +95,62 @@ public interface PaginationList {
     Text getPadding();
 
     /**
-     * Gets the maxinum number of lines that can be displayed on a source's
-     * screen.
+     * Gets the maximum amount of lines that will be sent per page.
      *
-     * @return The maximum number of lines per page
+     * <p>This defaults to the maximum amount of lines that can be displayed
+     * on a source's screen at one time if not specified.</p>
+     *
+     * @return The maximum amount of lines that will be sent per page
      */
     int getLinesPerPage();
 
     /**
-     * Send the constructed pagination list to the given source.
+     * Sends the first page of the constructed pagination list
+     * to the specified message receiver.
      *
-     * @param source The source to send to
+     * @param receiver The receiver to send the first page to
+     * @see PaginationList#sendTo(MessageReceiver, int) to send a specific page
      */
-    void sendTo(MessageReceiver source);
+    default void sendTo(MessageReceiver receiver) {
+        sendTo(receiver, 1);
+    }
 
     /**
-     * Send the constructed pagination list to the specified
-     * {@link MessageChannel}.
+     * Send the specified page of the constructed pagination list
+     * to the specified message receiver.
      *
-     * @param channel channel to send to
+     * <p>A page that is out of bounds will result in a friendly
+     * error message being sent to the receiver.</p>
+     *
+     * <p>Pages start at an index of 1.</p>
+     *
+     * @param receiver The receiver to send the page to
+     * @param page The page to send, starting at an index of 1
      */
-    default void sendTo(MessageChannel channel) {
-        for (MessageReceiver receiver : channel.getMembers()) {
-            this.sendTo(receiver);
+    void sendTo(MessageReceiver receiver, int page);
+
+    /**
+     * Sends the first page of the constructed pagination list to
+     * all {@link MessageReceiver}s within an {@link Iterable}.
+     *
+     * @param receivers The message receivers to send the first page to
+     * @see PaginationList#sendTo(Iterable, int) to send a specific page
+     */
+    default void sendTo(Iterable<MessageReceiver> receivers) {
+        sendTo(receivers, 1);
+    }
+
+    /**
+     * Sends the specified page of the constructed pagination list
+     * all {@link MessageReceiver}s within an {@link Iterable}.
+     *
+     * @param receivers The message receivers to send the page to
+     * @param page The page to send
+     */
+    default void sendTo(Iterable<MessageReceiver> receivers, int page) {
+        checkNotNull(receivers, "The iterable of receivers cannot be null!");
+        for (MessageReceiver receiver : receivers) {
+            sendTo(receiver, page);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -182,15 +223,18 @@ default void sendTo(MessageChannel channel) {
...
         /**
          * Sets the maximum number of lines that can be displayed per page.
          *
-         * @param linesPerPage The number of lines
-         * @return this
+         * <p>This defaults to the maximum amount of lines that can be displayed
+         * on a source's screen at one time if not specified.</p>
+         *
+         * @param linesPerPage The maximum number of lines to display per page
+         * @return tThis builder
```
vectrixdevelops: Small typo here.

vectrixdevelops (changes requested):

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -202,24 +246,30 @@ default void sendTo(MessageChannel channel) {
...
         }
 
+        /**
+         * Sends the constructed pagination list to an iterable of
+         * {@link MessageReceiver}s.
+         *
+         * @param receivers The message receivers to send the list to
+         */
         default void sendTo(Iterable<MessageReceiver> receivers) {
             receivers.forEach(this::sendTo);
```
liach: I think this foreach will make you build the same pagitation multiple times, which is a waste.

parlough: This is a good point since the pagination list is immutable and what is different between receivers is the `ActivePagination`. I will make this change :)

parlough: Should be good now.

JBYoshi (approved):

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
ST-DDT: Why did you remove the multi reciever variant? IMO its more likely (for me) to have a list of recievers than a message channel.

AFAICT there wont be a difference in the impl. You could even simplyfy it like this:
channel -> recievers -> single reciever

ST-DDT (changes requested):
Dont forget the usual not null checks.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
ST-DDT: Oh you did not remove it from the main object it was missing in the first place. If we have the method on the builder it should be on the main as well.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -202,26 +246,34 @@ default void sendTo(MessageChannel channel) {
...
+         * Sends the constructed pagination list to the members of the
+         * specified {@link MessageChannel}.
          *
-         * @param channel channel to send to
+         * @param channel The channel to send the list to
          */
         default void sendTo(MessageChannel channel) {
-            channel.getMembers().forEach(this::sendTo);
+            final PaginationList list = build();
+            channel.getMembers().forEach(list::sendTo);
```
ST-DDT: Maybe change this to `sendTo(channel.getMembers())` to simplyfy the impl/move it to a single method.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -202,26 +246,34 @@ default void sendTo(MessageChannel channel) {
...
+        /**
+         * Sends the constructed pagination list to an iterable of
+         * {@link MessageReceiver}s.
+         *
+         * @param receivers The message receivers to send the list to
+         */
         default void sendTo(Iterable<MessageReceiver> receivers) {
-            receivers.forEach(this::sendTo);
+            final PaginationList list = build();
+            receivers.forEach(list::sendTo);
```
ST-DDT: Maybe change it to the plain for each loop as that is usually faster (no need to create the lambda object).

ST-DDT: This would also be more consistent with the above impls.

ST-DDT (changes requested):
How about returing the PaginationList from the builders sendTo methods?

Meronat:
@ST-DDT Thanks for the great points, will make these changes shortly :)

parlough:
@ST-DDT Thanks for the suggestions, I think I addressed them all, mind looking over the API and common changes one last time?

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -202,26 +273,43 @@ default void sendTo(MessageChannel channel) {
...
+         * specified {@link MessageChannel}.
          *
-         * @param channel channel to send to
+         * @param channel The channel to send the list to
+         * @return The constructed pagination list
          */
-        default void sendTo(MessageChannel channel) {
-            channel.getMembers().forEach(this::sendTo);
+        default PaginationList sendTo(MessageChannel channel) {
+            checkNotNull(channel, "The message cannot be null!");
```
ST-DDT: You forgot the `channel` in the message.

ST-DDT (changes requested):
Looks good to me. As a minor thing I would order the methods like this:

* sendToReciever
* sendToRecieverWithPage
* sendToRecievers
* sendToRecieversWithPage
* sendToChannel
* sendToChannelWithPage

And in the builder like this:
* sendToReciever
* sendToRecievers
* sendToChannel

ST-DDT:
Also after noticing that the `contents` may not be null you have to throw an `IllegalStateException` in the build method and document it properly.

parlough:
@ST-DDT I am thinking at the moment that maybe we should remove this check entirely, and default to an empty list if it null. I'm open to discussion here, but for now this is what I will do.

...

I believe I have made all desired changes now in the both this and the common PR. Should be good to go unless anyone has any objections or further suggestions.

ST-DDT (approved):
Looks good to me.

vectrixdevelops (approved):

parlough:
These PRs are ready from my point of view and should be good to merge if everything looks good to everyone.

@kashike and @ryantheleach mind giving them a look?

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -92,76 +95,141 @@ static Builder builder() {
...
     }
 
     /**
+     * Sends the first page of the constructed pagination list to
+     * the members of the specified {@link MessageChannel}.
+     *
+     * @param channel The message channel to send the first page to
+     * @see PaginationList#sendTo(MessageChannel, int) to send a specific page
+     */
+    default void sendTo(MessageChannel channel) {
```
kashike: I don't like these methods that accept `MessageChannel`. You aren't sending it to the channel, you're sending it to the members of the channel - remove these two methods and keep the ones above that accept an `Iterable`.

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
```diff
@@ -182,46 +250,67 @@ default void sendTo(MessageChannel channel) {
...
-        default void sendTo(Iterable<MessageReceiver> receivers) {
-            receivers.forEach(this::sendTo);
+        /**
+         * Sends the constructed pagination list to the members of the
+         * specified {@link MessageChannel}.
+         *
+         * @param channel The channel to send the list to
+         * @return The constructed pagination list
+         */
+        default PaginationList sendTo(MessageChannel channel) {
```
kashike: See comment above.

kashike (changes requested):

src/main/java/org/spongepowered/api/service/pagination/PaginationList.java
parlough: That will cause this to be a breaking change as that method already existed. We could do that but since the Javadocs specify it's to the members I think it's okay.

kashike: I'd rather these just be removed.

parlough: Okay, will do.

PR #1627, opened 2017-08-11
Add SubjectDataUpdateEvent for permission changes

lucko:
Continues on from #1524 (and #1411)

### Why only an update event?
More specific update events are hard to (accurately) implement, and I personally don't think they'll get used. The PermissionService API is already pretty confusing to understand (especially to implement) and I think adding events into the mix will make things significantly more difficult.

Adding more specific stuff like `PermissionSetEvent` raises questions about what to do for bulk changes, changes which are made offline, etc.

In my opinion, a simple update event is all that is needed. It allows plugins which use permissions for data storage to update any caches they may have when data is updated.

#### Should the event be called async?
The API was recently updated to use CompletableFutures for SubjectData changes, so imo, it makes sense for the event to **always** be called async.

Consistency is obviously important, it can't be sometimes async and sometimes not. I'm not sure if async events would be a new addition for Sponge, but I assume they're supported? (in other words, is the EventManager thread safe?)

I think that's about it.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
+++ b/src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
@@ -30,11 +30,11 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import org.spongepowered.api.event.permission.SubjectDataUpdateEvent;
 import org.spongepowered.api.service.context.Context;
 import org.spongepowered.api.util.Tristate;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


ST-DDT:
I agree with Zidane here. Even if the perm plugin is used acros multiple servers as long as the change is syned to all servers (to update their perm cache) they should also give the oportunity to other plugins to update their caches.

As for changes made offline. The perm plugin could post the event when the entry is loaded or just rely on the other plugin to update its cache on join. If this is done during a perm plugin reload the plugin should throw this event for all online users/loaded subjects because things could have changed.

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -51,20 +52,61 @@
...
+    /**
+     * Called each time the data in this {@link MemorySubjectData} is changed.
+     *
+     * <p>The default behaviour is to notify the change listener, if one is
+     * present.</p>
+     */
+    protected void onUpdate() {
+        if (this.updateListener != null) {
+            this.updateListener.accept(this);
+        }
```
ST-DDT: IMO this should just post the change event. Otherwise this might be forgotten.

lucko: It gives the implementing plugin a chance to add any causes etc.

Also, since it is a requirement that the event is called async, a plugin instance is needed to schedule the event.

lucko:
I've removed the clause about the event being optional.

src/main/java/org/spongepowered/api/service/permission/MemorySubjectData.java
```diff
@@ -51,20 +52,61 @@
...
+     *
+     * @param updateListener The new update listener
+     */
+    public void setUpdateListener(@Nullable Consumer<MemorySubjectData> updateListener) {
+        this.updateListener = updateListener;
+    }
+
+    @Override
+    public Subject getSubject() {
+        return subject;
```
kashike: `this.`

lucko: Fixed

lucko:
Umm, I've kinda forgotten where this got to exactly, so err...

I think the issues remaining are:

* Is it ok to call events async?
* Should this event, in particular be called async? (I personally think yes)
* Is my kinda weird implementation in MemorySubjectData ok?

I think it might make more sense to either:

* Make MemorySubjectData abstract, and require that all uses of that class implement the `#onUpdate# method
* Define `#onUpdate` as a protected method, so implementations using the class can choose to implement it if they wish.

I'm not sure if this be preferred over the current Consumer (defined listener) implementation.

¯\\\_(ツ)_/¯

(tagging @Zidane because I think it was you who asked me to PR this, maybe)

bloodmc:
@lucko can you please fix conflicts?

lucko:
@bloodmc Done

I've swapped out the Consumer in favour of a protected update method.

bloodmc:
@zml2008 @Zidane This PR look good to pull?

PR #1629, opened 2017-08-13
Fix typo in BlockChangeFlag JavaDoc ("phsyics" => "physics")

vorburger:


ST-DDT:
Such small fixes should probably be reported there: https://github.com/SpongePowered/SpongeAPI/issues/1000

PR #1632, opened 2017-08-16
Split SleepingData into SleepingData and IgnoreSleepingData

ghost:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1488)

This pull request removes a method from `SleepingData` and moves it into its own `DataManipulator` called `IgnoreSleepingData`. It also adds a `Key` to represent it.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/IgnoreSleepingData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/IgnoreSleepingData.java
@@ -26,24 +26,19 @@ package org.spongepowered.api.data.manipulator.mutable.entity;
 
 import org.spongepowered.api.data.key.Keys;
 import org.spongepowered.api.data.manipulator.DataManipulator;
-import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableSleepingData;
+import org.spongepowered.api.data.manipulator.immutable.entity.ImmutableIgnoreSleepingData;
 import org.spongepowered.api.data.value.mutable.Value;
-import org.spongepowered.api.entity.living.Bat;
 import org.spongepowered.api.entity.living.player.Player;
 
-/**
- * An {@link DataManipulator} for the "sleeping" state of a
- * {@link Player}. Also applicable to {@link Bat}s where they can be
- * shown as "sleeping".
- */
-public interface SleepingData extends DataManipulator<SleepingData, ImmutableSleepingData> {
+public interface IgnoreSleepingData extends DataManipulator<IgnoreSleepingData, ImmutableIgnoreSleepingData> {
 
     /**
-     * Gets the {@link Value} for the "sleeping" state.
+     * Gets the {@link Value} for whether the owning {@link Player}'s
+     * "sleeping" state is ignored when vanilla mechanics dictate that all
+     * {@link Player}s must be sleeping to advance through the night.
      *
-     * @return The mutable value for the "sleeping" state
-     * @see Keys#IS_SLEEPING
+     * @return The mutable value for ignoring sleep
+     * @see Keys#IGNORE_SLEEPING
      */
-    Value<Boolean> sleeping();
-
+    Value<Boolean> ignoreSleeping();
 }
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/IgnoreSleepingData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/IgnoreSleepingData.java
@@ -41,4 +41,5 @@ public interface IgnoreSleepingData extends DataManipulator<IgnoreSleepingData, 
      * @see Keys#IGNORE_SLEEPING
      */
     Value<Boolean> ignoreSleeping();
+
 }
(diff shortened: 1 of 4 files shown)


Zidane:
1. Your SpongeCommon PR link points to the list of SpongeCommon PRs.
2. Might be better to deprecate and tell people to use this other way instead. Deprecated API gets removed after so much time has lapsed.

ghost:
@Zidane I am creating the SpongeCommon pull request now (hence why this pull request links to that list).

ryantheleach:
I have doubts that this is correct.

"the interface it was intended for already has a method on it for this purpose."

There are shortcut methods on a ton of interfaces that the data always being present is known for.

Whilst I can't think of a non Player entity that should be able to have this data, I don't see it as a valid reason for removal.

What surprises me more, is that the sleeping method on the mutable version, seems to have the javadocs for the ignoreSleeping method on the immutable version.

My suspicion, is that it's the mutable version that's incorrect, not the immutable version.

ghost:
@ryantheleach The reason why this method was deprecated was also because the value of it does not make sense for bats (and there is no way to store this value on them, but that is besides the point) and so this means that not all of the data is applicable to all of the targets (bats and players). I am not sure if there are other `DataManipulator`s whose data is not applicable to all targets hence why this method was deprecated.

ryantheleach:
Bat sleep is not the same data / mechanism as players sleeping. Apart from the two sharing an English word, there's no relation. A sleeping bat is more like a AI state. They should never have shared a data manipulator. That said, are they still sharing a manipulator?

ghost:
They would still be sharing a data manipulator in this pull request.

Deamon5550:
Then lets fix that instead and leave this method intact.

ghost:
I don't think that `SleepingData` should be split into two separate data manipulators for each entity type and that both bats and players should keep sharing this data manipulator. I think this as there is some evidence that the data manipulator is to be used for both bats and players as its Javadoc and that of the key `IS_SLEEPING` appears to say that they are able to be used with both of them.

ryantheleach:
The JavaDocs and Key for IS_SLEEPING date way back to the very early days of the DATA API. Back then there were many contributors, and not much consensus on how data should be separated. I think it's pretty clear that sleeping and hanging are two different concepts. 

The way I see it we have 

1. Bats hanging.
2. EntityPlayer rendered sideways in beds.
3. Players being counted as sleeping.

IgnoreSleeping covers 3. it's currently exposed on the interface as a shortcut method to what the DataAPI should be handling. https://github.com/SpongePowered/SpongeAPI/blob/80e588f0b94c74f6663ad1190eb907822a12055a/src/main/java/org/spongepowered/api/entity/living/player/Player.java#L297

2. and 1. are currently being handled by sleeping(). They could continue to share, but IMO ignoreSleeping should still be belonging to this data manipulator, and re-added to the mutable one.

ryantheleach:
The JavaDocs and Key for IS_SLEEPING date way back to the very early days of the DATA API. Back then there were many contributors, and not much consensus on how data should be separated. I think it's pretty clear that sleeping and hanging are two different concepts. 

The way I see it we have 

**1.** Bats hanging.
**2.** EntityPlayer rendered sideways in beds.
**3.** Players being counted as sleeping.

IgnoreSleeping covers 3. it's currently exposed on the interface as a shortcut method to what the DataAPI should be handling. https://github.com/SpongePowered/SpongeAPI/blob/80e588f0b94c74f6663ad1190eb907822a12055a/src/main/java/org/spongepowered/api/entity/living/player/Player.java#L297

2. and 1. are currently being handled by sleeping(). They could continue to share, but IMO ignoreSleeping should still be belonging to this data manipulator, and re-added to the mutable one. With a new DataManipulator introduced for 1.

Ignore Sleeping should definitely not be removed, or deprecated from the data manipulators without a viable alternative, the choices as I see them are

**a.** 1  & 2 Share a data manipulator, it's solely used for the rendering state of both bats, humans and players. 3. gets it's own data manipulator.

**b.** 2 & 3 share a data manipulator, 1 gets separated out as 'hanging'

**c.** 1 & 2 & 3 each have seperate data manipulators, 1 applies to animal/bat sleep states. 2. applies to players/humans 3. applies to players.

tldr; removing ignoreSleeping from the DataAPI without moving it is wrong. Let's fix the API instead.

src/main/java/org/spongepowered/api/data/manipulator/mutable/entity/IgnoreSleepingData.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Gets the {@link Value} for whether the owning {@link Player}'s
+     * "sleeping" state is ignored when vanilla mechanics dictate that all
+     * {@link Player}s must be sleeping to advance through the night.
+     *
+     * @return The mutable value for ignoring sleep
+     * @see Keys#IGNORE_SLEEPING
+     */
+    Value<Boolean> ignoreSleeping();
+}
```
ST-DDT: Add an empty line after the last member.

ST-DDT (approved):

gabizou (approved):

JBYoshi:
I was a bit too hasty with this merge. We may need some extra discussion on this, but I can't reopen it.

parlough:
I think ryan is right. This is not the correct solution. I think more discussion is needed, but keeping SleepingData as it is, to only refer to players makes the most sense to me. Bats really aren't sleeping to my understanding, they're just hanging, but either way it's quite different from player sleeping. So maybe introduce a HangingData to handle bats instead.

PR #1634, opened 2017-08-17
Move Titles to the implementation

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1491)

This moves the implementation of `Title` to SpongeCommon to follow what has been a continued goal of the API. 

There shouldn't be any breaking changes outside of the removing the static fields, but every method and static method should have a new version. 

If you believe there should be any other changes while I'm at it, breaking or not, please mention them.

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/Viewer.java
+++ b/src/main/java/org/spongepowered/api/effect/Viewer.java
@@ -182,14 +182,14 @@ public interface Viewer {
      * all settings back to default values.
      */
     default void resetTitle() {
-        sendTitle(Title.RESET);
+        sendTitle(Title.reset());
     }
 
     /**
      * Removes the currently displayed {@link Title} from the player's screen.
      */
     default void clearTitle() {
-        sendTitle(Title.CLEAR);
+        sendTitle(Title.clear());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


parlough:
I'm also looking for Javadoc improvements, especially since clear and reset functionality is kind of strange.

stephan-gh:
I appreciate your pull request, but I'd prefer if we move everything in one run to the implementation, and not only parts. It seems to be the best way going forward, but I still have some planning to do before we can do the switch.

For now, could you create a separate pull request with the Javadoc improvements so we can merge them earlier?

parlough:
The Javadocs are mostly the same with some slight changes, so there's not too much benefit to creating a separate pull request. 

I understanding wanting to do all of text in one go, so if you would like help with the rest I'd be willing to help where I can. I have some work locally moving other parts to the implemention but I haven't got far yet.

kashike:
#1635

Zidane:
I have to agree with @Meronat , I've fallen for this trap before where we state "we're moving Text to the impl!" and then it gets put on the back burner then shelved so we're going to try something different. It also gives us a chance to take chunks of it, evaluate potential flaws, and fix them.

You know, bite off what we CAN chew.

gabizou:
Since Title was moved to implementation when https://github.com/SpongePowered/SpongeAPI/pull/1744 was merged, this PR can be closed I think.

PR #1636, opened 2017-08-18
Make some changes for size data

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1492)

This PR makes some changes to SizeData as the keys were mutable bounded values but the data manipulator methods were just values. I also modified some documentation about how modifying these can not always work as intended. 

Please read the common PR for more information and provide as much feedback as possible on documentation and functionality.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableSizeData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/immutable/entity/ImmutableSizeData.java
@@ -27,7 +27,6 @@ package org.spongepowered.api.data.manipulator.immutable.entity;
 import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
 import org.spongepowered.api.data.manipulator.mutable.entity.SizeData;
 import org.spongepowered.api.data.value.immutable.ImmutableBoundedValue;
-import org.spongepowered.api.data.value.immutable.ImmutableValue;
 import org.spongepowered.api.entity.Entity;
 
 /**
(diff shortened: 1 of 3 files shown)


JBYoshi (approved):

kashike (approved):

ST-DDT (approved):

parlough:
Oh man this is embarrassing one second

parlough:
These should be good for a final review/test with the fixes I made last night.

parlough:
Looking for considerations on changing `base` to `width` as it is known internally? Might make more sense as well.

NeumimTo:
I had to guess that "base" is supposed to affect "width".

There seems to be an inconsistence with Entity#getBoundingBox where this method already returns AABB, but size data rely on "base" and "height" - two flaot params.

ryantheleach:
Base isn't width.

Width can be confused for width & depth, when base is used for both.

image

Docs could be written that base affects entities x and z components of the bounding box.

liach:
Can't entities just use any aabb for their shapes?

gabizou (approved):

Cybermaxke:
@parlough Can you change the use of `Float`s to `Double`s? This makes it more consistent with the rest of the API where almost everything uses `Double`.

Cybermaxke:
https://github.com/SpongePowered/SpongeAPI/commit/23d18da481d6c3e18177df067ac81e8e84ac1cdd

PR #1639, opened 2017-08-20
Add method to get hand that was used in UseItemStackEvent

### LINKED ISSUE #1464: Get the hand state
It would be nice if there was some way to get the hand state of a player. What I mean with this is to get if a hand is being used (rightclicking). In forge you have stuff like `player.isHandActive()` or `player.getActiveItemStack()`. One example of a use for this would be to do something special if a player is using a shield.

While it is possible to do something similar using `UseItemStackEvent` it feels cubersome, and also doesn't give you information about which hand the action happened in.

ghost:
**SpongeAPI** | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/343)

This pull request adds a method that, when a `UseItemStackEvent` gets fired, gets what hand (out of the main hand and the off-hand) the `ItemStack` that caused the event to be fired was on, as a `HandType` instance. This also fixes some typos in the Javadocs in that class.

This pull request fixes #1464.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/Player.java
@@ -45,7 +45,6 @@ import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.event.message.MessageChannelEvent;
 import org.spongepowered.api.item.inventory.Container;
 import org.spongepowered.api.item.inventory.Inventory;
-import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.network.PlayerConnection;
 import org.spongepowered.api.resourcepack.ResourcePack;
 import org.spongepowered.api.scoreboard.Scoreboard;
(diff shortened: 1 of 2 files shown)


ryantheleach:
Hand needs to return an optional for this to be mod compatible. Several mods allow you to use item stacks from slots other then MainHand and OffHand

ghost:
I think the API for this pull request is more or less done now so I have removed the "[WIP]" marker in the title.

ST-DDT (approved):

gabizou:
No idea how this got left behind for a year....

gabizou (approved):

ST-DDT (approved):

ImMorpheus:
This can't be merged, a SpongeCommon PR is missing.

However I'd prefer something like this: https://github.com/SpongePowered/SpongeAPI/issues/1650

Thoughts ?

ImMorpheus:
Superseded by https://github.com/SpongePowered/SpongeCommon/pull/2191

Issue #1348, opened 2016-09-02
Player#setTransformSafely


Bammerbom:
.setLocation has .setLocationSafely
.setLocationAndRotation has .setLocationAndRotationSafely
.setTransform has nothing.

I would like a .setTransformSafely() method.

parlough:
You have a good point, I'll look into adding this soon.

PR #1644, opened 2017-09-08
Adds a convenience method for setting the transformation of an entity safely

parlough:
A simple API addition to fix https://github.com/SpongePowered/SpongeAPI/issues/1348, allowing plugins to set transform safely similarly to how you can with location.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/Entity.java
+++ b/src/main/java/org/spongepowered/api/entity/Entity.java
@@ -270,6 +270,21 @@ public interface Entity extends Identifiable, Locatable, DataHolder, Translatabl
     boolean setTransform(Transform<World> transform);
 
     /**
+     * Sets the transformation of this entity using a safe location from
+     * {@link TeleportHelper#getSafeLocation(Location)}.
+     *
+     * @param transform The transform to set
+     * @return True if the transform was set successfully, or false if either
+     *     the transform couldn't be set as no safe location was found or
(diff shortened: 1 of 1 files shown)


parlough:
@dualspiral

src/main/java/org/spongepowered/api/entity/Entity.java
```diff
@@ -271,6 +271,24 @@ default boolean setLocationAndRotationSafely(Location<World> location, Vector3d
...
+     * {@link TeleportHelper#getSafeLocation(Location)}.
+     *
+     * @param transform The transform to set
+     * @return True if the transform was set successfully, or false if either
+     *     the transform couldn't be set as no safe location was found or
+     *     {@link org.spongepowered.api.event.entity.MoveEntityEvent.Teleport}
+     *     was cancelled
+     */
+    default boolean setTransformSafely(Transform<World> transform) {
+        checkNotNull(transform, "The transform cannot be null!");
```
dualspiral, line 283: Sure, this works, but why not just `return setLocationAndRotationSafely(transform.getLocation(), transform.getRotation())` instead? It's one less place to change if we change the teleport helper, and it means we don't create an extra `Transform` object.

parlough: Good points. Works for me, I'll make that change.

parlough:
Should be good now unless I missed something.

dualspiral (approved):
Looks reasonable to me.

kashike (approved):

PR #1645, opened 2017-09-08
Bugfixes and improvements for selectors

rexbut:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1519)

I am sorry to have recreated a PR but I have no news from @kenzierocks : https://github.com/SpongePowered/SpongeCommon/pull/793

Changes to accommodate the SpongeCommon PR. See that PR for details.


ryantheleach:
I realize it's code inherited from the current codebase that was left over from when selectors were first introduced, but what was the difference between forced and not forced resolution?

rexbut:
The forced résolution search entities in all worlds same if there are a location argument in argument list. 
I don't understand in what case, there are useful.

Deamon5550:
We need to deprecate the `resolveForce` methods before removing them.

rexbut:
I have deprecated the methods but the result is now an List <Entity> and there will be a break of compatibility. 
I think nobody currently uses the Selectors because the results is broken. Example (@p: return 2 players)

src/main/java/org/spongepowered/api/text/selector/Selector.java
```diff
@@ -325,4 +324,4 @@ static Selector parse(String selector) {
 
     }
 
-}
+}
```
kashike: newline at end of file please

kashike (changes requested):

Deamon5550:
Yeah thats unfortunate. Can you also add an @deprecated tag in the javadoc to point people to just use `resolve` instead now on the resolveForce methods.

parlough:
I'm sorry for this not receiving much attention, will definitely see more soon with API 8!

Katrix:
Thoughts on specifying that the parse method throws an `IllegalArgumentException` in the javadocs. Both so that others using it know what to look for as an error, and to make sure that other future implementations don't do something else?

dualspiral:
Superseded by https://github.com/SpongePowered/SpongeAPI/pull/1883

PR #1648, opened 2017-09-15
Fix ContextValue Annotation for Event Listeners

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1528)

Long story short: Since the Cause and Context PR merge, this annotation which *replaced* `@Named`, we've been unable to reproduce the feature set due to the language requirements that annotations are utilizing constants at *compile* time and not *runtime*. As such, it's not feasible to use for example:
```java
@Listener
public void onChange(ChangeBlockEvent.Place event, @ContextValue(EventContextKeys.OWNER.getName()) Player player) {
 // foo
}
```
This will result in a compile time error due to the lack of constants in this case.

This adds the constants to be used and some fuzzy string matching on the context to perform the job as expected.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/cause/EventContext.java
+++ b/src/main/java/org/spongepowered/api/event/cause/EventContext.java
@@ -97,6 +97,19 @@ public final class EventContext {
         return Optional.ofNullable((T) this.entries.get(key));
     }
 
+    @SuppressWarnings("unchecked")
+    public <T> Optional<T> findByName(String possibleName, Class<T> klass) {
+        checkNotNull(possibleName, "String names cannot be null");
+        checkNotNull(klass, "Type class cannot be null");
+        checkArgument(!possibleName.isEmpty(), "String names cannot be empty");
+        final String simplified = possibleName.toLowerCase().replace('_', ' ');
+        return (Optional<T>) this.entries.entrySet().stream()
+            .filter(entry -> entry.getKey().getName().toLowerCase().contains(simplified))
+            .map(Map.Entry::getValue)
+            .filter(klass::isInstance)
+            .findFirst();
+    }
+
     /**
      * Gets the value corresponding to the given key from the context.
      *
--- a/src/main/java/org/spongepowered/api/event/cause/EventContextKeys.java
+++ b/src/main/java/org/spongepowered/api/event/cause/EventContextKeys.java
@@ -106,6 +106,65 @@ public final class EventContextKeys {
 
     // SORTFIELDS:OFF
 
+    public static final class Named {
+        // SORTFIELDS:ON
+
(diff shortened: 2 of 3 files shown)


src/main/java/org/spongepowered/api/event/filter/cause/ContextValue.java
```diff
@@ -54,7 +58,7 @@
      * If specified the possible type for the returned object (normally
      * specified by the type of the annotated parameter) is restricted to only
      * the specified types.
-     *
+     * <p>
```
Deamon5550, line 61: Extra \<p\>

src/main/java/org/spongepowered/api/event/cause/EventContextKeys.java
```diff
@@ -106,6 +106,65 @@
 
     // SORTFIELDS:OFF
 
+    public static final class Named {
+        // SORTFIELDS:ON
+
+        public static final String BLOCK_HIT = "BLOCK_HIT";
```
Deamon5550, line 112: Should just make these names be the ids as in https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/type/event/EventContextKeysModule.java#L78

src/main/java/org/spongepowered/api/event/cause/EventContext.java
```diff
@@ -97,6 +97,19 @@ public static Builder builder() {
         return Optional.ofNullable((T) this.entries.get(key));
     }
 
+    @SuppressWarnings("unchecked")
+    public <T> Optional<T> findByName(String possibleName, Class<T> klass) {
+        checkNotNull(possibleName, "String names cannot be null");
+        checkNotNull(klass, "Type class cannot be null");
+        checkArgument(!possibleName.isEmpty(), "String names cannot be empty");
+        final String simplified = possibleName.toLowerCase().replace('_', ' ');
```
Deamon5550, line 105: Just have this get the EventContextKey from the registry by id, check that the key's type matches klass, and then return the get. We don't need fuzzy matching (and this isn't even good fuzzy matching).

Deamon5550 (changes requested):

Deamon5550:
Thinking further not sure why the new method on `EventContext` is even needed, is basically just a helper over getting the key from the registry, but without the caching that the proper way of doing this provided. Plugins should just be getting the key from the registry by id if they wish, allowing them to handle any error from the key not existing there. This method provides nothing new while simultaneously lowering the performance and masking any error from the key not existing.

Zidane:
After looking at this again over time, I agree with @Deamon5550 's thoughts.

PR #1655, opened 2017-10-02
Catalogue key

kashike:
[**SpongeAPI**](https://github.com/SpongePowered/SpongeAPI/pull/1655) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1554)

List of things needing to be replaced:
- [ ] Move all existing usages of `CatalogType#getId()` to `CatalogType#getKey()`


src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -0,0 +1,103 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
+
+/**
+ * A catalog key.
```
liach: You should specify the case requirements and illegal characters here.

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public interface CatalogKey {
```
liach: I suggest a default method for plugin containers like:
````java
/**
 * Creates a catalog key for this plugin.
 * 
 * @param value The value
 * @return A new catalog key
 */
static CatalogKey keyOf(String value) {
    return Sponge.getRegistry().createKey(getId(), value);
}
````

File: src/main/java/org/spongepowered/api/CatalogType.java [hunk 3]
```diff
@@ -50,7 +50,12 @@
      *
      * @return The unique identifier of this dummy type
      */
-    String getId();
+    @Deprecated
+    default String getId() {
+        return this.getKey().toString();
+    }
+
+    CatalogKey getKey();
```
liach: This method is lacking implementations, causing build failure. You should add a javadoc, too.

kashike: I'm well aware - see the `status: wip` label.

liach:
I suggest a default method for plugin containers.
````java
/**
 * Creates a catalog key for this plugin.
 * 
 * @param value The value
 * @return A new catalog key
 */
static CatalogKey keyOf(String value) {
    return Sponge.getRegistry().createKey(getId(), value);
}
````

File: src/main/java/org/spongepowered/api/CatalogKey.java [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Creates a catalog key.
+     *
+     * @param namespace The namespace
+     * @param value The value
+     * @return A new catalog key
+     */
+    static CatalogKey create(final String namespace, final String value) {
```
Zidane: Should be of()

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -85,6 +85,23 @@
 public interface GameRegistry {
 
     /**
+     * Creates a catalog key.
+     *
+     * @param namespace The namespace
+     * @param value The value
+     * @return A new catalog key
+     */
+    CatalogKey createKey(final String namespace, final String value);
```
Zidane: Frankly these methods offer little to nothing at all. CatalogKey has the create method (should be renamed of) which makes this method pointless and resolve can be put there as well. In what case would anyone ever reach out to the GameRegistry to create a CatalogKey...?

Zidane: Ignore what I said...need somewhere to point the static of methods in CatalogKey to. I guess GameRegistry will be our dumping ground :p.

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -96,11 +113,29 @@
      * @param id The case insensitive string id of the dummy type
      * @param <T> The type of dummy type
      * @return The found dummy type, if available
+     * @deprecated use {@link #getType(Class, CatalogKey)}
      * @see CatalogType
      */
+    @Deprecated
```
Zidane, line 119: I disagree with the deprecation. Sure we highly encourage usage of a CatalogKey but it shouldn't be a requirement: if someone passes a string that can't be mapped to anything of the type then they get back the empty optional.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
```diff
@@ -70,18 +72,43 @@ private KeyFactory() {}
      * @param id The id for the new key
      * @param name The name for the new key
      * @return The generated key
+     * @deprecated use {@link #makeSingleKey(TypeToken, TypeToken, DataQuery, CatalogKey, String)}
      */
+    @Deprecated
```
Zidane: Same here, you already create the key for them...I see no reason to just outright remove the method.

src/main/java/org/spongepowered/api/CatalogKey.java
liach: Addressed.

src/main/java/org/spongepowered/api/GameRegistry.java
gabizou: Boo!!! More clutter!! Booo!!!!

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -85,6 +85,23 @@
...
+     */
+    CatalogKey createKey(final String namespace, final String value);
+
+    /**
+     * Resolves a catalog key.
+     *
+     * @param value The value
+     * @return A new catalog key
+     */
+    CatalogKey resolveKey(final String value);
```
gabizou, line 102: Perhaps this should accept a particular `CatalogType` class in the case where we have different defaults depending on the type? Like `BlockType` versus `DataRegistration` etc.

src/main/java/org/spongepowered/api/GameRegistry.java
gabizou: Agreed with Zidane, it should be at the base level lowest level allowance to use Strings. We want to avoid the little issues we've had with usability of `DataQuery`s.

src/main/java/org/spongepowered/api/data/key/KeyFactory.java
gabizou: I personally don't like that `CatalogKey` has the registration id but we still have the name separate. Perhaps we could merge the two in this case? Still feels cluttered. We may want to move the KeyFActory to use builders instead at this point with the amount of arguments.

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
```diff
@@ -163,4 +164,16 @@ default Logger getLogger() {
...
 
+    /**
+     * Creates a catalog key with the {@link #getId() id} of
+     * this plugin as the namespace.
+     *
+     * @param value The value
+     * @return The catalog key
+     * @see CatalogKey#of(String, String)
+     */
+    default CatalogKey createCatalogKey(final String value) {
```
gabizou, line 175: The method name is funky, maybe `createNamespacedKey`?

src/main/java/org/spongepowered/api/text/format/TextColors.java
```diff
@@ -46,6 +47,11 @@ private TextColors() {
         private final Color color = Color.BLACK;
 
         @Override
+        public CatalogKey getKey() {
+            return CatalogKey.sponge("none");
```
gabizou: We really should get Text rewrite done to eliminate this implementation leakage.......

gabizou (changes requested):

src/main/java/org/spongepowered/api/GameRegistry.java
kashike: We're going to become like `ResourceLocation` in Forget-me-not, then. Forge is stuck with `String` ids everywhere.
`String` based ids are just plain silly. It results in us needing to guess what namespace to prepend to the string they provide, it has no documentation that can be seen during creation, it requires a bunch of spots throughout the implementation (which we already have a huge mess of, which this PR set will fix) changing the case to lowercase, etc.

From the Oracle documentation:
> When you design an API, carefully consider whether it supersedes an old API. If it does, and you wish to encourage developers (users of the API) to migrate to the new API, then deprecate the old API. Valid reasons to deprecate an API include:
>
> - It is insecure, buggy, or highly inefficient
> - It is going away in a future release
>- It encourages bad coding practices
>
>Deprecation is a reasonable choice in all these cases because it preserves "backward compatibility" while encouraging developers to change to the new API. Also, the deprecation comments help developers decide when to move to the new API, and so should briefly mention the technical reasons for deprecation.

It would be stupid to keep the `String`-based methods non-deprecated.

Zidane: After discussion with @kashike he convinced me to allow the deprecation here with a compromise that it'll be a bit before we remove it. I do agree with him on the fact that the Forge ecosystem cannot move past the usage of the strings for the keys because it was never deprecated correctly and now impossible to remove.

liach:
Any updates?

Cybermaxke:
Status?

liach:
time to summon the catman!

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Creates a catalog key.
+     *
+     * @param namespace The namespace
+     * @param value The value
+     * @return A new catalog key
+     */
+    static CatalogKey of(final String namespace, final String value) {
+        return Sponge.getRegistry().createKey(namespace, value);
```
gabizou: Why no builder?

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -49,8 +49,19 @@
      * default/vanilla minecraft types.</p>
      *
      * @return The unique identifier of this dummy type
+     * @deprecated use {@link #getKey()}
```
gabizou: Explain why the difference of naming, something like:
```
@deprecated Use {@link #getKey()} due to id's being represented now as a namespaced mod id and the catalog type's id. Avoids the ambiguity of what part of what makes the ID being the mod namespace.
```

src/main/java/org/spongepowered/api/CatalogType.java
```diff
@@ -49,8 +49,19 @@
...
      */
-    String getId();
+    @Deprecated
+    default String getId() {
+        return this.getKey().toString();
+    }
+
+    /**
+     * Gets the catalog key for this catalog type.
```
gabizou: Maybe explain that the catalog key is provided and should be used to retrieve this `CatalogType` from a `CatalogTypeRegistry`.

gabizou: And mention that this key should be used for serialization.

src/main/java/org/spongepowered/api/GameRegistry.java
gabizou: I still believe we should be providing a type class to better resolve the key for, since you can have ambiguity in some cases, like the format of the block state id's compared to the block type id's.

src/main/java/org/spongepowered/api/extra/modifier/empty/VoidWorldGeneratorModifier.java
```diff
@@ -54,8 +55,8 @@ public void modifyWorldGenerator(WorldProperties world, DataContainer settings,
     }
 
     @Override
-    public String getId() {
-        return "sponge:void";
+    public CatalogKey getKey() {
```
gabizou, line 58: Shouldn't this class be moved to implementation?

src/main/java/org/spongepowered/api/registry/CatalogRegistryModule.java
```diff
@@ -38,8 +39,22 @@
      *
      * @param id The id of the catalog type requested
      * @return The catalog type, if available
+     * @deprecated use {@link #get(CatalogKey)}
```
gabizou, line 42: Same javadoc description as above.

gabizou (changes requested):

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -24,10 +24,32 @@
...
- * A catalog key.
+ * An object representation of a location or pointer to a {@link CatalogType}
+ * that can be used to retrieve said type from the {@link GameRegistry}. The
+ * key can be represented as a {@link String} by {@link #toString()}. The key
+ * is built with two parts:
+ * <ol>
+ *     <li>The Namespace</li>
+ *     <li>The Value</li>
+ * </ol>
+ * Normally, the namespace is lowercased and likewise, so is the value. This
```
kashike, line 43: The namespace and value **must** be lowercase - we throw if they aren't. Valid characters for a namespace are `a-z0-9_.-`, valid characters for a value are `a-z0-9/._-`.

Cybermaxke: Do we want to allow the characters `/.` in ids? I am currently replacing all `/`'s for advancement ids with `_`'s. As for `.`'s, they are currently only used for sound ids (as far I know) and can also be replaced by `_`'s (to reflect `SoundTypes` field names)?
Or is this something that needs to change for 1.13?

Cybermaxke: But `/` and `.` are allowed to be a part of a `ResourcePath`, which will be added by a separate PR for 1.13.

kashike: Err, what? `CatalogKey` is equal to `ResourceLocation` internally - why do we need a separate `ResourcePath`?

gabizou: @Cybermaxke I've implemented `CatalogKey` over `ResourceLocation` currently as it works quite well. The resolution of id's is also defaulted on `ResourceLocation`, so having that handling already is much better than having yet another object mapping between the two classes.

Currently implementation is *almost passing*.

Cybermaxke: `ResourcePath`s will only be used related to data packs, not ids.
https://github.com/SpongePowered/SpongeAPI/pull/1706
I would like to keep ids and resource paths separate, but in impl they could be the same thing.

It's just the allowed characters in catalog keys that should be specified, I think that `/` could be useful for advancements that use a path. So that they keep a directory structure.
But I would get rid of `.` in keys, to be more consistent for ids. They could easily be replaced with `/` or `_`. For example sound ids: `entity.husk.ambient` -> `entity/husk/ambient`

liach:
yay!

Issue #1658, opened 2017-10-03
Per-World View Distance missing?


Dockter:
Per @kashike,

Unless I've missed it, there doesn't seem to be a way to set a per-world view distance.  Please advise.

gabizou:
Merged in with #1659 .

PR #1659, opened 2017-10-04
view distance

kashike:
[**SpongeAPI**](https://github.com/SpongePowered/SpongeAPI/pull/1659) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1557)
adds #1658

### DIFF
--- a/src/main/java/org/spongepowered/api/world/World.java
+++ b/src/main/java/org/spongepowered/api/world/World.java
@@ -416,7 +416,6 @@ public interface World extends Extent, WeatherUniverse, Viewer, ContextSource, M
      * Causes an {@link Explosion} in a world.
      *
      * @param explosion The explosion to cause
-     * @param cause The cause for the explosion
      */
     void triggerExplosion(Explosion explosion);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -447,4 +446,20 @@ default void setSerializationBehavior(SerializationBehavior behavior) {
      * @throws IOException If the save failed
      */
     boolean save() throws IOException;
+
+    /**
+     * Gets the view distance for this world.
```
Deamon5550: We should specify that the value is in chunks

Deamon5550 (changes requested):

src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -447,4 +446,20 @@ default void setSerializationBehavior(SerializationBehavior behavior) {
...
+    int getViewDistance();
+
+    /**
+     * Sets the view distance for this world.
+     *
+     * <p>The view distance must be greater than or equal to 3, andless than or equal to 32.</p>
+     *
+     * @param viewDistance The view distance
+     */
+    void setViewDistance(final int viewDistance);
```
Deamon5550, line 464: Also we have no way of unsetting a world's view distance (setting it to use the global setting rather).

ryantheleach: as per https://github.com/SpongePowered/SpongeCommon/pull/1557#discussion-diff-142562964R107 sending a value of -1 resets it. 

**if** this remains in the API, it needs to be documented.

Deamon5550: we can probably do better than that

Zidane:
@kashike 

This needs to be updated to address @Deamon5550 's concerns above.

PR #1660, opened 2017-10-05
Ensure that an argument that is an exact match always takes precedence

dualspiral:
For reference, some of the discussion in https://github.com/SpongePowered/SpongeAPI/pull/1514 is relevant, but this will fix a more immediate problem.

There are two things that this fixes - and I'm PRing because I haven't tested this yet and don't currently have the setup to do so:

### Some arguments require escaping for an exact match

Take the fake player `[Mekanism]`. If this is entered into the `UserCommandElement`, this will not match the fake player `[Mekanism]` because the argument is transformed into the regex `^[Mekanism]`, meaning it'll match any player that starts with any of the letters in `Mekanism`. The `PatternMatchingCommandElement` _does_ try to match a choice with an element, but only _after_ filtering choices with the regex.

This PR will fix that by ensuring the exact match test happens _before_ the regex is used to filter the elements. This should fix exactness checks for `BlockState` `CatalogTypes` too, which often use `[]` characters.

### The UserCommandElement tries to use a PlayerCommandElement first - sometimes erroneously matching a Player before a User

Assume that there are two players that have joined the server: `dualspiral` and `dualspiral_`. Say `dualspiral_` is online and `dualspiral` is offline. Due to the current setup of the `UserCommandElement`, it wraps a `PlayerCommandElement`, which picks out the online "dualspiral_" even if only `dualspiral` (or `dualspira` etc.) is typed - which is clearly not intended.

This moves the user exactness check to the beginning of the element parser.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
+++ b/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
@@ -31,7 +31,6 @@ import com.google.common.collect.Iterables;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.text.Text;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.regex.Pattern;
@@ -52,14 +51,17 @@ public abstract class PatternMatchingCommandElement extends CommandElement {
     @Override
     protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
         final String unformattedPattern = args.next();
-        Pattern pattern = getFormattedPattern(unformattedPattern);
-        Iterable<String> filteredChoices = Iterables.filter(getChoices(source), element -> pattern.matcher(element).find());
-        for (String el : filteredChoices) { // Match a single value
-            if (el.equalsIgnoreCase(unformattedPattern)) {
-                return Collections.singleton(getValue(el));
-            }
+        Iterable<String> choices = getChoices(source);
+
+        // Check for an exact match before we create the regex.
+        // We do this because anything with ^[abc] would not match [abc]
+        Optional<Object> exactMatch = getExactMatch(choices, unformattedPattern);
+        if (exactMatch.isPresent()) {
+            return exactMatch.get();
         }
-        Iterable<Object> ret = Iterables.transform(filteredChoices, this::getValue);
+
+        Pattern pattern = getFormattedPattern(unformattedPattern);
+        Iterable<Object> ret = Iterables.transform(Iterables.filter(choices, element -> pattern.matcher(element).find()), this::getValue);
 
         if (!ret.iterator().hasNext()) {
             throw args.createError(t("No values matching pattern '%s' present for %s!", unformattedPattern, getKey() == null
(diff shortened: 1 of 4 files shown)


ryantheleach:
So what's the old behavior vs the new behavior?

Old list of priority:

src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
```diff
@@ -87,6 +89,18 @@ Pattern getFormattedPattern(String input) {
...
+     * Tests a string against a set of valid choices to see if it is a
+     * case-insensitive match.
+     *
+     * @param choices The choices available to match against
+     * @param potentialChoice The potential choice
+     * @return If matched, an {@link Optional} containing the matched value
+     */
+    protected Optional<Object> getExactMatch(final Iterable<String> choices, final String potentialChoice) {
+        return Iterables.tryFind(choices, potentialChoice::equalsIgnoreCase).toJavaUtil().map(this::getValue);
+    }
```
ryantheleach, line 101: This forces case insensitivity for *anything* that uses patternmatchingcommandelement, I understand that it might be wanted for players / users, but I think this is overstepping somewhat, at least until 1.13

ryantheleach:
So what's the old behavior vs the new behavior?

Old list of priority:
1. Parse online players first (unfortunately this matches partial matches for online players first).
2. Pattern Match against all users in user storage.
3. (optionally return the source player, if return source was enabled)

New list of priority:
1. Use selectors to parse users, (matching all worlds, unless location parameters are set).
2. Pattern match users if selectors fail.
3. (optionally return the source player, if return source was enabled)

This would seem to remove the priority of online players matching before offline players for partial matches.

I'd define the 'correct' and wanted behavior by admins to be.

1. Use selectors to parse users.
2. Exact Match online users.
3. Exact Match offline users.
4. partial match online users.
5. partial match all users.
6. optionally return the current command source, if no others found.
??. Exact Match MojangService users??

The original goal of allowing regex patterns, was to provide wildcard matching to admins who have grown to depend on it coming from bukkit / essentials.

Remember, that it's always possible to have an argument that accepts multiple users, and asks for disambiguation in case of doubt, and have a -force flag or something on the command.


e.g. /op KratosKnox using a plugin that accepts multiple users could result in the following.

When only KratosKnox1 and KratosKnox2 are online.

(red) Multiple matching users found, did you mean KratosKnox1 or KratosKnox2 (use /op KratosKnox --force to op multiple people)

or

(green) opped KratosKnox1, KratosKnox2

Depending on whether the plugin believes the command to be dangerous or not.

Partial matches for *single* player elements should only accept the element, if there is a single user (in storage / cache) that matches the name.

src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java
dualspiral: It’s not - I’m mirroring what the original code intended, and it tries to do a case insensitivity check for exact matches now, just in the wrong place.

dualspiral:
@ryantheleach That list is what this argument does now, except selectors come third (but users don’t start with the `@` symbol, so it doesn’t matter). I had to move some logic into the `parse` method to do that, so not everything is in the `parseValue` method - and I’ve been told by @bloodmc that the `get` method in the `UserStorageService` will prioritise an online player (though in most cases, this shouldn’t matter because names are unique anyway).

As for prioritising an online player over an offline player, it doesn’t necessarily do that right now and I don’t necessarily agree there should be priority, it should return everything that matches the regex if it’s not an exact match. So, in my example, `dualspiral_` and `dualspiral`, if I enter `dual`, it should return both - `onlyOne()` is a modifier that should handle that, not this element.

ryantheleach:
At the very least it should prioritize online players for tab completion / order of the returned collection but that's less of an issue.

dualspiral:
I'm going to merge this to bleeding tomorrow then to stable-7 shortly thereafter unless anyone has any objections, it's been sitting around a while and I know at least @SimonFlash has been waiting on it. Wouldn't mind comment on the second part of this PR (on exact matching) if anyone else has comments.

kashike (approved):

PR #1664, opened 2017-10-11
Refactoring to use specialised functional interfaces

ameyaKetkar:
Hello,
I am a Graduate Student at Oregon State University, studying the impact of autoboxing. I have written a tool, which analyses the source code to find opportunities to refactor the code to a specialised implementation for primitives, thus eliminating autoboxing. All the changes, have been done manually, and have been verified by me.In this pull request I have specialised two instances of FunctionalInterfaces . I have many more such patches, I will be happy to contribute.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/biome/GroundCoverLayer.java
+++ b/src/main/java/org/spongepowered/api/world/biome/GroundCoverLayer.java
@@ -26,12 +26,13 @@ package org.spongepowered.api.world.biome;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.MoreObjects;
+import java.util.function.DoubleFunction;
+
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.util.weighted.SeededVariableAmount;
 import org.spongepowered.api.util.weighted.VariableAmount;
 
-import java.util.function.Function;
+import com.google.common.base.MoreObjects;
 
 /**
  * Represents a layer of BlockStates specific to a biome which may be placed in
@@ -39,7 +40,7 @@ import java.util.function.Function;
  */
 public class GroundCoverLayer {
 
-    private Function<Double, BlockState> block;
+    private DoubleFunction<BlockState> block;
     private SeededVariableAmount<Double> depth;
 
     /**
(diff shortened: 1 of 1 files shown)


Dockter:
I'm confused.  This literally changes nothing but breaks a bunch of existing plugins to pieces.

ameyaKetkar:
The reason for the change : to change functional interface to specialised ones to avoid autoboxing. 
YEs, this is my fault, I shouldn't have made these changes in the public methods(since its an API here). This wouldn't have been a problem, if every invocation used a lambda expressions. 

But I can generate an analysis for any class which uses generics. Example, I have instances where HashMap could be refactored to Int2IntHashMap(if you use fastUtils/gs-collections.) or other specialised implementations.I could also do it for any Util you use say Tuple, Pair, etc..  If you would like such patches I will be happy to contribute.

Deamon5550:
I too can find and replace, hardly ground breaking research... As for the changes themselves, this is the kind of thing where perhaps it would be nice to do if this was really hot code (which this isn't particularly) but if we were going to change it we would change it to wrap up the concepts better in a new type (the double in this is an implementation defined noise value which is kind of nasty to expose to the api).

ameyaKetkar:
I don't just do what you can do with find and replace.  I can get you all the instances that could be refactored to specialised implementation(with given mapping), all the other places which are affected by the change and also create the changes.With other analysis, like primitive patterns in methods accepting varargs(or object type), generic classes which could be specialised to specific primitives (like Tuple, Pair, or anything depending on their usage in the code base). 
If I can help improve your code with such analysis, I will be happy to contribute.

ryantheleach:
@ameyaKetkar With respect, unless you are being instructed to do this by a professor, or have a relationship with this project, I think you will "change the world" a lot faster by embedding these inspections into IDE's rather then PR all of github, one project at a time.

@Dockter I believe their point stands, It would improve performance slightly, But the larger questions are whether we want to expose specific versions of Guava, Specific collections libraries, etc considering we don't have full control of our dependencies.

ameyaKetkar:
Yes. That's the plan. I have made this a part of IDE(Eclipse plugin). I am just testing my code, and using this as an opportunity to contribute.

PR #1674, opened 2017-10-27
Add exhaustion to damage source and add some javadocs

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1573)

Adds the ability to set and retrieve the exhaustion of a damage source. Also adds some Javadocs while I'm at it. I'm open to changes of the defaults for exhaustion and where I set the exhaustion when building the damage sources(refer to common PR).

### DIFF
--- a/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
+++ b/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
@@ -61,8 +61,8 @@ public interface DamageSource {
     DamageType getType();
 
     /**
-     * Gets whether this {@link DamageSource} can not be modified and the
-     * damage is absolute.
+     * Gets whether this {@link DamageSource}'s damage is absolute and
+     * will ignore potion effects and enchantments.
      *
      * @return If this damage source deals absolute damage
      */
@@ -112,8 +112,9 @@ public interface DamageSource {
      * Gets the amount of exhaustion this {@link DamageSource} will
      * add to the entity, generally only to players.
      *
-     * <p>In vanilla this is generally set to 0.1 by default and
-     * overridden and set to 0 if the damage is set to be absolute.</p>
+     * <p>In vanilla gameplay this is set to 0.1 by default and
+     * overridden to 0 if the source is set to be absolute or
+     * as overriding armor.</p>
      *
      * @return The increase in exhaustion
      */
@@ -133,14 +134,18 @@ public interface DamageSource {
 
         /**
          * Sets this {@link DamageSource} as dealing damage that
-         * bypasses armor.
+         * bypasses armor modifiers.
+         *
+         * <p>This sets the exhaustion increase caused
+         * by this source to 0. You can override this
+         * with {@link #exhaustion(double)}.</p>
          *
          * @return This builder
          */
         B bypassesArmor();
 
         /**
-         * Sets whether this {@link DamageSource} as an explosion.
+         * Sets this {@link DamageSource} as an explosion.
          *
          * @return This builder
          */
@@ -148,8 +153,12 @@ public interface DamageSource {
 
 
         /**
-         * Sets this {@link DamageSource} as not being able to be modified
-         * with damage that is absolute.
+         * Sets whether this {@link DamageSource}'s damage is absolute and
+         * will ignore potion effects and enchantments.
+         *
+         * <p>This sets the exhaustion increase caused
+         * by this source to 0. You can override this
+         * with {@link #exhaustion(double)}.</p>
          *
          * @return This builder
          */
@@ -175,9 +184,14 @@ public interface DamageSource {
          * Sets the amount of exhaustion this {@link DamageSource} will
          * add to the entity, generally only to players.
          *
-         * <p>In vanilla this defaults .1 in vanilla, and turns to 0 if the
-         * damage is absolute. This builder generally defaults to it to 0.1
-         * and is not overridden if you set the damage as absolute.</p>
+         * <p>In vanilla gameplay, the default is 0.1, unless if the damage
+         * is absolute or bypasses armor, where the exhaustion gets set to 0.
+         * This builder follows this mechanic, but if you set the exhaustion
+         * through this method that will be overridden.</p>
+         *
+         * <p>If you don't set this exhaustion manually, calling
+         * {@link #absolute()} or {@link #bypassesArmor()} will
+         * set this 0 and if you don't this will default to 0.1.</p>
          *
          * @param exhaustion The amount of exhaustion to add to the entity
          * @return This builder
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
```diff
@@ -108,24 +108,99 @@ static Builder builder() {
...
+         *
+         * @param exhaustion The amount of exhaustion to add to the entity
+         * @return This builder
+         */
+        B exhaustion(double exhaustion);
+
+        /**
+         * Sets the {@link DamageType} of this source.
+         *
+         * <p>This is often required to be set.</p>
```
dualspiral: Often required? When is it not required?

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
```diff
@@ -108,24 +108,99 @@ static Builder builder() {
...
+         *
+         * @return This builder
+         */
         B creative();
 
+        /**
+         * Sets the amount of exhaustion this {@link DamageSource} will
+         * add to the entity, generally only to players.
+         *
+         * <p>In vanilla this defaults .1 in vanilla, and turns to 0 if the
```
dualspiral: The javadoc here needs rewording a bit, maybe something like:

```
In vanilla gameplay, the default is 0.1, unless the damage is absolute, 
where the exhaustion is zero. This may not be the case for this builder, 
in particular, a call to {@link #absolute()} is not guaranteed to set this to zero.
```

With that being said, I don’t know if that is the best behaviour - I feel like the implementation should follow vanilla mechanics if the exhaustion isn’t set.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
```diff
@@ -108,24 +108,99 @@ static Builder builder() {
...
+         * Sets whether this {@link DamageSource} as an explosion.
+         *
+         * @return This builder
+         */
         B explosion();
 
+
+        /**
+         * Sets this {@link DamageSource} as not being able to be modified
+         * with damage that is absolute.
```
dualspiral: This is a bit confusing - do you actually mean that absolute damage cannot modify this?

What I imagine you mean is that this damage source will be unaffacted by damage modifiers, right?

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/common/AbstractDamageSourceBuilder.java
```diff
@@ -39,9 +39,9 @@
     protected boolean absolute = false;
     protected boolean magical = false;
     protected boolean creative = false;
+    protected double exhaustion = 0.1;
```
dualspiral: ...and then I find the impl in the API.

What about making this a nullable `Double`, which indicates “follow vanilla”? Then, when building, this would indicate “do as Vanilla does”.

Just a thought, quite happy to be told this isn’t viable.

dualspiral (commented):
Few questions on the API, if you don’t mind! :)

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/common/AbstractDamageSourceBuilder.java
parlough: I like your suggestion, I really couldn't decide the best way to handle this. I would be happy to make this change soon/tomorrow. 

I'm definitely open for more feedback here from anyone as this was the hardest part of the change to decide.

_It is unfortunate a lot of this is in the API, but that's a story for later._

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: I thought I found one case where it was not required in some subclass or something, I will look again later though to confirm. I may change wording to "generally".

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: I like this suggestion as well. I would still use the "not guaranteed" wording though, as a source or builder implementing this is not guaranteed to act that way.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: I copied this from the get method. I will clean both of them up after looking at how we exactly handle a damage source being absolute in common.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
dualspiral: I would actually avoid words like “generally”, because the problem is that the onus is on the developer to find out when it is and isn’t required. Better to say that the type is required, except in the cases when - if the subclass is a subclass of this builder, the javadoc on the method _there_ should state that it’s not required.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
dualspiral: Noted, may as well take advantage of this PR to do that, I guess.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: Fair enough. Best to say when it is not true when it mostly is I suppose. Will work on something more clear.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: @dualspiral @gabizou So looking at it now(since we handle damage completely ourselves for the most part), it seems(not 100% sure) we only ignore armor and potion resistance if damage is absolute(in SpongeCommon), but not enchantments, shields, and absorption. Is this an oversight and should we bypass those protections as well, or should I document absolute as just bypassing armor and potion resistance?

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
Cybermaxke: Bypassing the armor (`bypassesArmor`) should also set the exhaustion to zero.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
parlough: Looks like we were ignoring armor if it was absolute instead of if it was bypassing armor, and were not bypassing enchantments if it was absolute. I will make these changes in my SpongeCommon PR.

To summarize: If a damage source is absolute, like the mcp Javadocs say, it seems to just bypass potions and enchantments.

parlough:
Issues should be addressed, still looking for feedback(and more 3rd-party testing) here and on the common PR.

PR #1675, opened 2017-10-28
Add cooldown tracker API

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1575)

Adds an API for the `CooldownTracker`, allowing you to set a cooldown on the use of items. In vanilla this is used for ender pearls, chorus fruit, and shields. An example you could add with this is putting a cooldown on water buckets so people don't spam water everywhere. 

Fixes https://github.com/SpongePowered/SpongeAPI/issues/1307

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
@@ -1,0 +1,85 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.living.player;
+
+import org.spongepowered.api.item.ItemType;
+
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+
+/**
+ * Provides access to the item cooldowns of a {@link Player}.
+ */
+public interface CooldownTracker {
+
+    /**
+     * Checks if the specified {@link ItemType} is currently on cooldown
+     * for the player.
+     *
+     * @param type The item type to check is on cooldown
+     * @return Whether or not the specified item type is cooldown
+     */
+    boolean hasCooldown(ItemType type);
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
```diff
@@ -0,0 +1,89 @@
...
+     * Gets the percentage of the specified {@link ItemType}'s cooldown that
+     * is remaining for the player, or empty if the the item type
+     * is currently not on cooldown.
+     *
+     * <p>If present, this value will be between 0.0 and 1.0.</p>
+     *
+     * @param type The item type to get the cooldown percentage remaining
+     * @return The percentage of cooldown remaining for the specified item type
+     */
+    OptionalDouble getPercentageRemaining(final ItemType type);
```
dualspiral: This isn’t returning a percentage - you’ve specified that it’s between 0 and 1, not 0 and 100. I suppose you could use “getFractionRemaining”.

ST-DDT (commented):
Remove the final modifier from the interface's method params (if the method is not implemented). It only adds noise and no benefits.

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Checks if the specified {@link ItemType} is currently on cooldown
+     * for the player.
+     *
+     * @param type The item type to check is on cooldown
+     * @return Whether or not the specified item type is cooldown
+     */
+    default boolean hasCooldown(final ItemType type) {
+        return !getCooldown(type).isPresent();
```
kashike: what? why `!`? this should return `true` if there *is* a cooldown

kashike (changes requested):

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
parlough: You're quite correct, oops. I overrode this in the implementation so I never noticed.

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
parlough: Good point, I'll make this change and add a little to the Javadocs as well.

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    OptionalInt getCooldown(final ItemType type);
+
+    /**
+     * Resets the cooldown of the specified {@link ItemType} for the
+     * player.
+     *
+     * @param type The item type to reset the cooldown for
+     */
+    default void resetCooldown(final ItemType type) {
```
Deamon5550: Don't default implement this.

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
Deamon5550: Just don't default implement this as every implementation is almost certainly going to have a more efficient implementation of it. Default implementing should only really be used for calling alternatives.

Deamon5550 (changes requested):

src/main/java/org/spongepowered/api/entity/living/player/CooldownTracker.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.spongepowered.api.item.ItemType;
+
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+
+/**
+ * Provides access to the item cooldowns of a {@link Player}.
+ */
+public interface CooldownTracker {
```
gabizou, line 35: Maybe perhaps even have the `Player` object available for retrieval? Since the implementation provides such a method (only case it doesn't is in single player).

Thoughts @Zidane ?

parlough: I'm not sure I see a need as this is gotten from the `Player` object.

gabizou: In certain circumstances where consumer code is passing a `CooldownTracker`, it would be useful being able to just use the simple getter, simply by passing a `CooldownTracker` instead of the pair of `Player` and `CooldownTracker` objects.

ST-DDT: If you need both why not just provide the player since you can get the CDT from the player?

kashike: :-1:, I don't see a need for it.

parlough:
Should be good for a final review.

parlough:
Works well for me and have fixed brought up issues. Will test one more time tonight then merge if there are no objections or other issues brought up.

PR #1677, opened 2017-10-30
updated javadoc for Event.getCause

sibomots:
Edit to suit.

Goal: To make a the JavaDoc a bit more expressive of the result of getCause.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/Event.java
+++ b/src/main/java/org/spongepowered/api/event/Event.java
@@ -41,7 +41,17 @@ import org.spongepowered.api.util.annotation.eventgen.PropertySettings;
 public interface Event {
 
     /**
-     * Gets the cause for the event.
+     * Gets the cause for the event.  The Cause of the event
+     * is an object which implements the Iterable interface.
+     * So, when investigating the Cause of the event a common
+     * idiom is to use operations (functions) on the result
+     * of getCause as follows:
+     *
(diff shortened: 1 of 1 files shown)


ST-DDT (commented):
Your additions start with a double space. Otherwise looks good to me.

PS: isn't it required to use `&lt;` instead of `<`? Or maybe use the code or literal block.

src/main/java/org/spongepowered/api/event/Event.java
```diff
@@ -41,7 +41,17 @@
...
     /**
-     * Gets the cause for the event.
+     * Gets the cause for the event.  The Cause of the event
+     * is an object which implements the Iterable interface.
+     * So, when investigating the Cause of the event a common
+     * idiom is to use operations (functions) on the result
+     * of getCause as follows:
+     *
+     * Use-case: Getting the Player (if any) responsible:
+     * Optional<Player> optPlayer = event.getCause().first(Player.class);
```
liach: `<code></code>` or `{@code }`

src/main/java/org/spongepowered/api/event/Event.java
```diff
@@ -41,7 +41,17 @@
...
+     *
+     * In general the event.getCause() exposes access to the
```
liach: This is already mentioned earlier.

PR #1678, opened 2017-10-31
Make StateMatcher serializable

### LINKED ISSUE #1528: Config Serializable State Matchers
https://github.com/SpongePowered/SpongeAPI/commit/abd6fa8fee5c0f86845e80dd4f9980395aaa0388#diff-c00b1420115c3cf9601763262c66ddbdR307

Awesome work guys, StateMatchers are a good step towards making better configs for blocks :)

Would you consider accepting a PR that made them config serializable in some form, so we had a standard way of specifying "loose blocks" in configs?

Katrix:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1795)

Makes StateMatcher serializable.

I'm a bit unsure about where to register the builder. Anyone mind pointing that out to me? Closes #1528.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
+++ b/src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
@@ -26,6 +26,8 @@ package org.spongepowered.api.block;
 
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.trait.BlockTrait;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.util.ResettableBuilder;
 
 import java.util.List;
@@ -39,7 +41,7 @@ import java.util.function.Predicate;
  * that contains 4 traits, and only 2 are wanting to be matched,
  * then the other two traits may be variable).
  */
-public interface BlockStateMatcher extends Predicate<BlockState> {
+public interface BlockStateMatcher extends Predicate<BlockState>, DataSerializable {
 
     /**
      * Constructs a new {@link Builder} to construct a {@link BlockStateMatcher}.
@@ -78,7 +80,7 @@ public interface BlockStateMatcher extends Predicate<BlockState> {
     /**
      * A builder for building {@link BlockStateMatcher}s.
      */
-    interface Builder extends ResettableBuilder<BlockStateMatcher, Builder> {
+    interface Builder extends DataBuilder<BlockStateMatcher> {
 
         /**
          * Sets the root {@link BlockType} for the {@link BlockStateMatcher}.


src/main/java/org/spongepowered/api/block/BlockState.java
```diff
@@ -308,6 +314,29 @@ public MatcherBuilder reset() {
...
+
+        @Override
+        protected Optional<StateMatcher> buildContent(DataView container) throws InvalidDataException {
+            reset();
+            Map<DataQuery, Object> entries = container.getValues(false);
+            BlockType blockType = container.getCatalogType(DataQuery.of("blockType"), BlockType.class).orElseThrow(
+                    () -> new InvalidDataException("Invalid BlockType"));
+            type(blockType);
+
+            for(Map.Entry<DataQuery, Object> entry : entries.entrySet()) {
```
liach: space between `for`, `if` and `(`

src/main/java/org/spongepowered/api/block/BlockState.java
```diff
@@ -308,6 +314,29 @@ public MatcherBuilder reset() {
...
+                String traitId = entry.getKey().last().asString("");
+                Object obj = entry.getValue();
+
+                if(obj instanceof Comparable) {
+                    blockType.getTrait(traitId).ifPresent(t -> trait((BlockTrait)t, (Comparable)obj));
```
liach: Suppress rawtypes warning?

Katrix:
Fixed.

gabizou (changes requested):
If we're making it serializable, I'd rather we move the implementation to implementation. Since this is no longer just a simple state matcher, I'd like it to exist where we don't expose what keys are being used, etc.

ryantheleach:
With the flattening coming in 1.13 I see us needing a matcher that fulfills this role, is human-serializable, but uses legacy block properties or Data Keys/Values instead.

Nearly all cases where the old State Matcher would have been useful, are changed for 1.13 to either just work with Blocks rather then blockstates, or no longer have implementation available to query.

e.g. there is no color property on wool anymore.

gabizou:
> With the flattening coming in 1.13 I see us needing a matcher that fulfills this role, is human-serializable, but uses legacy block properties or Data Keys/Values instead.
> 
> Nearly all cases where the old State Matcher would have been useful, are changed for 1.13 to either just work with Blocks rather then blockstates, or no longer have implementation available to query.

Hence why this should be moved to implementation. So that implementation can support whatever we need to support, and warn appropriately.

ryantheleach:
This is a good start, but I'd like to see a potential BlockMatcher as well, 

That can handle 
1. (1.13) tags e.g. `#minecraft:anvil`
2. BlockTypes `minecraft:anvil` (This is done I believe)
3. partialStates `minecraft:anvil[damage=0]` (This is done I believe)
4. DataKeys, Maybe interacting with https://github.com/SpongePowered/SpongeAPI/pull/1715

No matter what, the serialization format has to be relatively simple, and understandable to people who have dealt with minecraft data previously. e.g. /testblock

@gabizou can this be merged as is for now, (for 7.1)? (except maybe in impl) and work on a generalized BlockMatcher can come later?

Katrix:
I've finished most of the things I've been working on lately so I think I should be able to get some work on this and other stuff I started today and tomorrow.

gabizou:
This needs updating since `BlockStateMatcher` has moved and is now just an interface. I'll gladly merge once the update has been done, and there's a test plugin added to the implementation. I don't believe this needs 1.13 to be available prior to merging.

Katrix:
Don't think I ever mentioned that I updated this.

gabizou (approved):

PR #1679, opened 2017-11-01
add AITargetEvent

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1679) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1582)

Adds an Event for when the AITarget of an Agent changes.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/ai/SetAttackTargetEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/ai/SetAttackTargetEvent.java
@@ -1,0 +1,56 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.entity.ai;
+
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.entity.living.TargetAgentEvent;
+
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * Fires before the attack target of an {@link Agent} is set.
+ */
+public interface SetAttackTargetEvent extends TargetAgentEvent, Cancellable {
+
+    /**
+     * Gets the {@link Agent}s new target {@link Entity}.
+     *
+     * @return The target
+     */
+    Optional<Entity> getTarget();
+
+    /**
+     * Sets the {@link Agent}s new target {@link Entity}.
+     *
+     * <p>This target is set after the event when it is not cancelled.</p>
+     *
+     * @param target The target
+     */
+    void setTarget(@Nullable Entity target);
+}


src/main/java/org/spongepowered/api/event/entity/ai/AITargetEvent.java
```diff
@@ -0,0 +1,50 @@
...
+     * @return The target
+     */
+    Entity getAITarget();
+
+    /**
+     * Sets the {@link Agent}s new target {@link Entity}
+     *
+     * @param target The target
+     */
+    void setAITarget(Entity target);
```
liach: What's the difference between this and https://github.com/SpongePowered/SpongeAPI/blob/14f459d824e3b0216a2ec0c4683c2c06fc80404c/src/main/java/org/spongepowered/api/entity/living/Agent.java#L59

src/main/java/org/spongepowered/api/event/entity/ai/AITargetEvent.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.Agent;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.entity.living.TargetAgentEvent;
+
+/**
+ * Fires before the target of an {@link Agent} is changed.
+ */
+public interface AITargetEvent extends TargetAgentEvent, Cancellable
+{
```
liach: put it on last line, this is sponge not forge

src/main/java/org/spongepowered/api/event/entity/ai/AITargetEvent.java
Faithcaio: The target (when not cancelled) will be set after the event.

src/main/java/org/spongepowered/api/event/entity/ai/AITargetEvent.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.api.event.entity.living.TargetAgentEvent;
+
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+/**
+ * Fires before the target of an {@link Agent} is changed.
+ */
+public interface AITargetEvent extends TargetAgentEvent, Cancellable {
```
ryantheleach: What is the semantics for cancelling?

Will it maintain the current target, or set it to null?

If you maintain the current target you get the unintuitive behavior of an event being raised for something to *continue* to target the same player periodically, where cancelling has the same effect as not changing the target.

However, setting it to null when cancelled will end up possibly leaving the AI in a state where it's unable to target anything hostile, as it's always aiming for the closest.

I'm torn between whether I think cancelling should then target the next valid entity, or whether this should be an uncancellable event.

Either way, can we add a getOriginalTarget(); and getPreviousTarget() that can't be set, for informational purposes?

ryantheleach: To others: I've discussed this a little with faith, and he is interested in input.

dualspiral: The javadocs holds the key for me:

> Fires before the target of an {@link Agent} is changed.

Canceling an event means, to me, that the server acts as if the event had not taken place. So it should maintain the current target.

The javadocs for `setTarget` makes it clear that you can change the target if the event is _not_ canceled:

> This target is set after the event when it is not cancelled.

Doing otherwise is inconsistent with the rest of the API - for example, the `MessageChannelEvent`, canceling prevents a message appearing (no change to the current server), but there is an option to change the message (setMessage), where there is still a material change in the server state.

If you want to make it clear what is canceled, change the event name to `AIChangeTargetEvent` - that's a little more descriptive about what it's doing.

liach: ~~The current implementation sets the target to `null` on cancellation. It should probably keep the old target when cancelling instead.~~ By the way, I suggest the name `SetAITargetEvent` for this class.

liach: Now, I think we should only use this event for notification purposes; if plugins want to set the target,  they may call [`Agent#setTarget`](https://github.com/SpongePowered/SpongeAPI/blob/14f459d824e3b0216a2ec0c4683c2c06fc80404c/src/main/java/org/spongepowered/api/entity/living/Agent.java#L59) which fires the event again (this should be safe with cause manager)

ryantheleach: `AIChangeTargetEvent` I think would fix the confusion I was having, except for the 'retargetting'

e.g. an entity already has you as a target, and is randomly checking to retarget. It chooses you again and the event is thrown. Should it not throw an event if the target hasn't changed?

@dualspiral My questions wern't so much, "what is the current behavior as described (or implemented)" but "What **Should** the behavior be, and can we possibly JavaDoc it better" (faith asked me to contribute better JD once the definition was locked down, if I thought I could explain it better)

I guess my real issue is that in testing, it would seem this event gets called too late IMO. but I have doubts on how easy it would be to fix in a mod compatible way.

Usecases I can think of off the top of my head:

1. Make an entity untargettable by all(most? (unvanished)) entities. (Spectator, vanishTargetting, seems to cover this well enough, maybe needs a little tweaking)
2. Prevent entities targeting those on the same 'team' (can be covered by raw MC teams, or this event, (scoreboards are too inflexible for most complex uses))
3. Make an entity target a vehicle or pet before a player.
4. Make entities not target players in safe zones, so they stay out of protected regions.

The problem I'm trying to avoid is "paralysing" entities with indecision when there are multiple valid targets. e.g. they target nothing, because they are targeting a vanished player, when they could be targeting a real player.

Faithcaio: I don't think we can easily avoid this "paralysing".
The actual Entitytargeting is hidden away in multiple AI classes.
But a plugin listening to this event could do the targeting itself.

As for the name: `SetAttackTargetEvent`
or `SetAITargetEvent` ?
It is also called when the target is set but does not change.
Agent#setTarget does not cause this event to fire (currently)

liach: > It chooses you again and the event is thrown. 

Does Minecraft actually do that?

ryantheleach: " could do the targeting itself."

If this is to be the case, then we are going to try to need a lot more information about what circumstances it's being thrown under, otherwise how do plugins know if it was a hostile target, or someone holding wheat?

kashike (approved):

parlough (approved):

src/main/java/org/spongepowered/api/event/entity/ai/SetAttackTargetEvent.java
```diff
@@ -0,0 +1,56 @@
...
+    Optional<Entity> getTarget();
+
+    /**
+     * Sets the {@link Agent}s new target {@link Entity}.
+     *
+     * <p>This target is set after the event when it is not cancelled.</p>
+     *
+     * @param target The target
+     */
+    void setTarget(@Nullable Entity target);
```
dualspiral, line 55: Though it might be obvious to us, perhaps state what happens in the JD if null is passed (i.e. no target)?

PR #1686, opened 2017-11-10
Add missing ItemTypes and add some clarifying Javadocs

parlough:
I'm surprised we didn't notice this earlier, especially concrete and concrete powder.


kashike (approved):

liach:
Can we find an easy way to iterate through all the `GLAZED_TERRACOTTA`s? Mojang's removal of meta data is really silly indeed.

gabizou:
> Can we find an easy way to iterate through all the GLAZED_TERRACOTTAs? Mojang's removal of meta data is really silly indeed.

I anticipate that 1.13 massive changes in the pipeline are going to affect all of this altogether.

Issue #1684, opened 2017-11-07
Issues with Enchantment related classes/interfaces


parlough:
At the moment, `Enchantment` represents what would normally be called `EnchantmentType` in the API, and is catalogued by `Enchantments` when generally would be catalogued by `EnchantmentTypes`. Then the actual representation of an existing enchantment is `ItemEnchantment`. I may be mistaken and there may be a reason behind this, so if so, please point it out to me.

This goes against what is represented elsewhere, and an example of how it should be is `PotionEffect`, `PotionEffectType`, and `PotionEffectTypes`. So it would go:

`Enchantment` -> `EnchantmentType`
`Enchantments` -> `EnchantmentTypes`
`ItemEnchantment` -> `Enchantment`

This is semi-related to a similar issue discussed in https://github.com/SpongePowered/SpongeAPI/commit/dccaebccc2772ad485b61aec4728235830ff4140 where the implementation of `ItemEnchantment` is exposed when it necessarily should not be and is lacking a builder. 

Enchantments must have came before some API decisions were made and are just kind of out of place now.

I am looking for discussion on this and whether it would be worth to break(in API 8 or even still in API 7?) while also moving what is currently `ItemEnchantment` more to the implementation with a builder, while improving the Javadocs on all of the relevant interfaces and classes. @Zidane 

I plan to look into and possibly propose some other changes to enchantments to allow increased functionality of custom enchantments, allow more control of the enchanting table, etc. So these may be able to go along with these changes if they come to fruition.

parlough:
Will PR this later today(most likely).

PR #1687, opened 2017-11-12
Cleanup of enchantments

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1593) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/361) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1847)

I will finish what I need to tomorrow, but for now looking for any feedback and also where I undoubtedly made mistakes as I haven't really tested much and some stuff was done automatically through IntelliJ refactoring and I may have missed mistakes.

Closes https://github.com/SpongePowered/SpongeAPI/issues/1684


src/main/java/org/spongepowered/api/data/Queries.java
```diff
@@ -42,8 +42,8 @@
     public static final DataQuery WEIGHTED_SERIALIZABLE = of("DataSerializable");
     public static final DataQuery WEIGHTED_SERIALIZABLE_WEIGHT = of("Weight");
 
-    // Item Enchantment
-    public static final DataQuery ENCHANTMENT_ID = of("Enchantment");
+    // Item EnchantmentType
+    public static final DataQuery ENCHANTMENT_ID = of("EnchantmentType");
```
parlough: This was done automatically by IntelliJ. Should I keep I revert this change to not cause compatibility issues?

src/main/java/org/spongepowered/api/block/tileentity/EnchantmentTable.java
```diff
@@ -25,7 +25,7 @@
 package org.spongepowered.api.block.tileentity;
 
 /**
- * Represents an Enchantment Table.
+ * Represents an EnchantmentType Table.
```
liach: This one is absolutely wrong.

src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogItemData.java
```diff
@@ -114,7 +114,7 @@
      */
     public static final Class<DyeableData> DYEABLE_DATA = DyeableData.class;
     /**
-     * Signifies that an item has {@link Enchantment}s applied to it.
+     * Signifies that an item has {@link EnchantmentType}s applied to it.
```
liach: You probably should revert this one.

src/main/java/org/spongepowered/api/data/manipulator/catalog/CatalogItemData.java
```diff
@@ -169,7 +169,7 @@
      */
     public static final Class<SpawnableData> SPAWNABLE_DATA = SpawnableData.class;
     /**
-     * Signifies that there are stored {@link Enchantment}s available on the
+     * Signifies that there are stored {@link EnchantmentType}s available on the
```
liach: Same?

src/main/java/org/spongepowered/api/data/meta/Enchantment.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.meta;
```
liach: I think we should move the package of this class. It is not really a meta any more, since it has been converted into an interface.

src/main/java/org/spongepowered/api/data/meta/Enchantment.java
```diff
@@ -0,0 +1,114 @@
...
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.item.EnchantmentType;
+import org.spongepowered.api.item.inventory.ItemStack;
+
+/**
+ * Represents an {@link EnchantmentType} on an {@link ItemStack} that is paired
+ * with a level.
+ *
+ * <p>The contract of enchantments is that their level will
+ * never be below 1.</p>
```
liach:
````
<p>The contract of enchantments is that their level will
be in the range of {@code getType().getMinimumLevel()}
and {@code getType().getMinimumLevel()}.</p>
````

File: src/main/java/org/spongepowered/api/item/EnchantmentTypes.java [hunk 7]
```diff
@@ -0,0 +1,256 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * An enumeration of known {@link EnchantmentType}s.
```
liach: You should probably mention that all of them have a minimum level of 1.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Increases damages to arthropod mobs. In vanilla this includes spiders,
+     * cave spiders, silverfish, and endermites.
+     *
+     * <p>In vanilla the maximum level is 5.</p>
+     */
+    public static final EnchantmentType BANE_OF_ARTHROPODS = DummyObjectProvider.createFor(EnchantmentType.class, "BANE_OF_ARTHROPODS");
+
+    /**
+     * Prevents removal of the cursed items.
```
liach: You should probably state that this enchantment only target items in the 4 armor slots.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * <p>In vanilla the maximum level is 1.</p>
+     */
+    public static final EnchantmentType AQUA_AFFINITY = DummyObjectProvider.createFor(EnchantmentType.class, "AQUA_AFFINITY");
+
+    /**
+     * Increases damages to arthropod mobs. In vanilla this includes spiders,
+     * cave spiders, silverfish, and endermites.
+     *
+     * <p>In vanilla the maximum level is 5.</p>
```
liach: At high level, the enchantment will slow down the arthropod hit for few seconds.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * An enumeration of known {@link EnchantmentType}s.
+ */
+public final class EnchantmentTypes {
+
+    // SORTFIELDS:ON
+
+    /**
+     * Increases underwater mining rate.
```
liach: You should say it just increase the speed to the regular mining speed.

src/main/java/org/spongepowered/api/data/meta/Enchantment.java
liach: Uhh yes, it can be any natural number as long as the number can be stored in data.
Does this hurt items with empty enchantment tags? Those tags may give item stacks an effect while not having a valid enchantment.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
```diff
@@ -0,0 +1,256 @@
...
+     * Reduces projectile damage you take, for example from arrows, ghasts,
+     * blaze fire charges, and similar in vanilla.
+     *
+     * <p>In vanilla the maximum level is 4.</p>
+     */
+    public static final EnchantmentType PROJECTILE_PROTECTION = DummyObjectProvider.createFor(EnchantmentType.class, "PROJECTILE_PROTECTION");
+
+    /**
+     * Reduces all damage, outside of a few sources, such as the void,
+     * the kill command, and hunger damage in vanilla.
```
liach: Probably mention that it doesn't reduce damage that bypasses armor.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Attackers are damaged when they deal damage to the wearer.
+     *
+     * <p>In vanilla the maximum level is 3.</p>
+     */
+    public static final EnchantmentType THORNS = DummyObjectProvider.createFor(EnchantmentType.class, "THORNS");
+
+    /**
+     * RIncreases effective durability.
```
liach: Remove that R I guess

src/main/java/org/spongepowered/api/util/TypeTokens.java
```diff
@@ -213,9 +213,9 @@
 
     public static final TypeToken<List<FireworkEffect>> LIST_FIREWORK_TOKEN = new TypeToken<List<FireworkEffect>>() {private static final long serialVersionUID = -1;};
 
-    public static final TypeToken<List<ItemEnchantment>> LIST_ITEM_ENCHANTMENT_TOKEN = new TypeToken<List<ItemEnchantment>>() {private static final long serialVersionUID = -1;};
+    public static final TypeToken<List<Enchantment>> LIST_ITEM_ENCHANTMENT_TOKEN = new TypeToken<List<Enchantment>>() {private static final long serialVersionUID = -1;};
 
-    public static final TypeToken<ListValue<ItemEnchantment>> LIST_ITEM_ENCHANTMENT_VALUE_TOKEN = new TypeToken<ListValue<ItemEnchantment>>() {private static final long serialVersionUID = -1;};
+    public static final TypeToken<ListValue<Enchantment>> LIST_ITEM_ENCHANTMENT_VALUE_TOKEN = new TypeToken<ListValue<Enchantment>>() {private static final long serialVersionUID = -1;};
```
liach: You should rename these 2 fields.

src/main/java/org/spongepowered/api/data/meta/Enchantment.java
parlough: So we only check when building if it is above 0 or below Short.MAX_VALUE. There are issues if the level is below 0 if I recall correctly but in reality it can be 0(I think?). I just changed it to 1 since I didn't see a reason why an enchantment would ever need to be 0 and Minecraft doesn't really expect it with all of theirs having a minimum level of 0. Do you think I should switch back to 0 or greater?

Also to your suggestion. I don't think that is a great idea as we return the vanilla max and min which are not necessarily as high as they could be set.

I'm heavily open to feedback relating to our lower limit and how to specify it.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
parlough: Interesting, never knew this.

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
parlough: Makes sens, will do.

src/main/java/org/spongepowered/api/data/Queries.java
parlough: Anyone have input here?

kashike: undo

src/main/java/org/spongepowered/api/data/meta/Enchantment.java
parlough: How do you feel about moving `Enchantment`, `EnchantmentType`, and `EnchantmentTypes` to a new `enchantment` package inside of the `effect` package?

src/main/java/org/spongepowered/api/item/EnchantmentTypes.java
parlough: Done.

src/main/java/org/spongepowered/api/data/Queries.java
parlough: Will do. 👍

Cybermaxke:
Could we also use a custom ListValue and extra methods in the enchantment data manipulator to turn in some kind of mapped data? So you don't have to loop through all the enchantments to get the level, this could also be optimized internally.

```java
import org.spongepowered.api.data.value.mutable.ListValue;
import org.spongepowered.api.item.EnchantmentType;
import org.spongepowered.api.item.Enchantment;

import java.util.OptionalInt;

public interface EnchantmentListValue extends ListValue<Enchantment> {

    OptionalInt getLevel(EnchantmentType enchantment);

    void add(EnchantmentType enchantment, int level);
}
```

ryantheleach:
No we can not @Cybermaxke :( 

Vanilla very occasionally will generate with multiple enchants of the same type on an item.

https://www.youtube.com/watch?v=66zHMfi-bsI

parlough:
What are people's opinions on the minimum enchantment value? Should we allow it to be 0 or above, or 1 and above to match vanilla's enchantments and to avoid possible issues.

liach:
@Meronat We should probably allow only positive levels, no zero. The glowing effects some items have is from an empty enchantment tag list (so the list can exist and be empty, a different concept) so level 0 enchantments are useless

parlough:
These 4 PRs should be ready for more testing and review. 🥂

ryantheleach:
@Meronat are we exposing the 'natural maximum' enchantment level of each item anywhere? Is it even plausible?

parlough:
@ryantheleach I believe this is what you mean? https://github.com/Meronat/SpongeAPI/blob/ae6f6d92a545bdd9ff8fd40b08f77c5b724b87f8/src/main/java/org/spongepowered/api/item/enchantment/EnchantmentType.java#L62

ryantheleach:
https://gaming.stackexchange.com/questions/262999/what-is-the-highest-effective-enchantment-level/263013#263013 States that some out of band negative values may have some effect. Not sure if we should handle them, or reject them outright. How do other parts of the API behave with out of band values?

Also don't know if any mods do anything outrageous.

src/main/java/org/spongepowered/api/item/enchantment/Enchantment.java
```diff
@@ -0,0 +1,113 @@
...
+         * @return The modified builder, for chaining
+         */
+        Builder type(EnchantmentType enchantmentType);
+
+        /**
+         * Sets the level for this enchantment.
+         *
+         * @param level The desired level
+         * @return The modified builder for chaining
+         * @throws IllegalArgumentException If the level is below 1
```
gabizou: This contradicts the argument in the java doc for the class of `Enchantment` [here](https://github.com/SpongePowered/SpongeAPI/pull/1687/files#diff-5e7629a1c106a95db1ab90b7621fc617R37)

gabizou: And I'm somewhat concerned about mod compatibility here.

gabizou (changes requested):

src/main/java/org/spongepowered/api/item/enchantment/Enchantment.java
parlough: There I said their level will always be larger than 0, which equates to the same thing as never below 1, but I can clean up the wording and add the max as well. 

How do you propose I handle the concern? Should we simply allow them to use the full range of `Short.MIN_VALUE` to `Short.MAX_VALUE`, specifying that unintentional issues may arise if you go outside of `EnchantmentType#getMinumumLevel` and `EnchantmentType#getMaxmimumLevel`?

gabizou: > Should we simply allow them to use the full range of Short.MIN_VALUE to Short.MAX_VALUE, specifying that unintentional issues may arise if you go outside of EnchantmentType#getMinumumLevel and EnchantmentType#getMaxmimumLevel?

I am going to have to say "yes" because we cannot 100% control the possibilities of enchantments. If developers choose to make unworkable enchantments, that's more on them.

parlough: I will make this change and the necessary changes to the docs later tomorrow.

**Edit:** Sorry haven't had a chance to get around to this yet, will do soon!

parlough:
Any final reviews on API or implementation? Changes work for me.

gabizou (approved):

PR #1688, opened 2017-11-12
Adds a direct invulnerable to invulnerability data

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1595)

This adds control of direct invulnerability without timing to `InvulnerabilityData` alongside the ticks(being hurt). This allows entities to be impervious to damage **BESIDES damage from `DamageSources.VOID` and from a creative player.** 

**NEED INPUT:** Relating to the lack of protection of damage from creative players(due to this being used for armor stands I'm assuming?), I could change this check, removing the check if it is from a creative player from entity and moving that to armor stand instead. I'm open to feedback on this and please point out if it is used elsewhere in this manner.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -1113,6 +1113,17 @@ public final class Keys {
     public static final Key<Value<Boolean>> INVISIBLE = KeyFactory.fake("INVISIBLE");
 
     /**
+     * Represents the {@link Key} for representing if an {@link Entity}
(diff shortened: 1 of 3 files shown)


parlough:
I forgot to add this(will next commit), but the kill command bypasses it as well. This will remain the same though.

parlough:
Okay should be good on my end for API and implementation. Looking for feedback and if anyone wants to test, that'd be cool.

gabizou (approved):

PR #1693, opened 2017-11-14
Split up pickup event

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1693) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1602) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1863)

The current ChangeInventoryEvent.Pickup does not cover what really happens. 
It fires before the inventory change yet pretends to know what changes will happen in the inventory.
related issues: https://github.com/SpongePowered/SpongeCommon/issues/1597 |  https://github.com/SpongePowered/SpongeCommon/issues/1402

I propose splitting it up into a Pre-Event to handle the Item Entity pickup and ItemStacks to be added to the Inventory.
And the "normal" ChangeInventoryEvent.Pickup which will then contain the actual Slot changes.

Ideas & Comments are welcome.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
@@ -24,24 +24,34 @@
  */
 package org.spongepowered.api.event.item.inventory;
 
+import org.spongepowered.api.entity.Item;
 import org.spongepowered.api.entity.living.Living;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.entity.item.TargetItemEvent;
 import org.spongepowered.api.item.inventory.Inventory;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
+
+import java.util.List;
+import java.util.Optional;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/ChangeInventoryEvent.java
```diff
@@ -79,9 +88,58 @@
...
+     * Fires after an {@link Item} has been picked up.
+     */
+    interface Pickup extends ChangeInventoryEvent {
+
+        /**
+         * Fires before an {@link Item} is picked up.
+         *
+         * <p>Modifying the picked up items causes this event to be automatically canceled if the inventory does not fit the entire list.</p>
+         */
+        interface Pre extends TargetInventoryEvent, TargetItemEvent, Cancellable {
```
gabizou, line 109: Could we do something where it's much like equipment pickup?

Faithcaio: Equipement Pickup?

ryantheleach: I was wondering the same thing, my assumption was the Equipment Change events, like when a zombie directly equips armor off the floor.

Faithcaio: I think Zombie Pickup is not implemented yet

kashike (approved):

PR #1697, opened 2017-11-19
Adds ability to get the connected chest from a chest, if it is there

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1697) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1606)

Adds API to get the other double chest location.


parlough:
Maybe this would be better returning the tile entity instead, if they need the location they can get it from the tile entity. `Chest#getConnectedChest`?

parlough (approved):
A simple addition, and since it is isolated to chests and his a simple implementation, I don't see why to not add it.

ryantheleach:
Yeah, 1.13 when it comes allows you to place single chests side by side, and (I believe) removes the possibility of having 3 wide bug chests, but mods could always do something funky.

Looks good to me.

Issue #1236, opened 2016-05-09
Naming Colors


gabizou:
Current, `Color` is of various utility in the game, but Colors are not "named" or "searchable" for user input of `dark cyan` or `black` etc.

@Zidane brought up the point that `Color`s that are "standard" should be named and searchable, but considering the possibility of having 16 million colors and the ability to get various colors by RGB values, it is a complex issue.

So, we ask the community developers what possible solutions or opinions you may have about the above issue.

windy1:
Not quite the same idea but I think a system that could benefit developers greatly would be some kind of way to retrieve MC colors by RGB by finding the closest match.

As far as naming go. There could be another level of abstraction for MC colors that are used in multiple places and TextColor or DyeColor or whatever could just wrap those somehow.

Anyways just my initial thoughts. Shouldnt be considered fleshed out.

PR #1699, opened 2017-11-22
Orange you glad to see this a-peel-ling PR?

SimonFlash:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1699) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1625)

For just ~~one~~ two PR's to Sponge, you too can support orange and show him some love!

Orange

*All puns are 100% oranginal and totally weren't stolen from google. Or Parker.*


Meronat:
You are going to be citrisized for all of these puns for sure.

parlough:
**You are going to be _citrisized_ for all of these puns for sure.** 

Color needs some larger changes that I plan to work on later, but it does make sense for orange to be added for now.

ryantheleach:
Where are you pulling this value of Orange from?

Without having yet tested or investigated, my suspicion is that there is possibly corresponding PR's that need to be made to SpongeCommon.

SimonFlash:
Standard orange [web color](https://en.wikipedia.org/wiki/Web_colors).

The `Color` class is not a catalog type and constants are not registered. There should be no conflicts or other issues caused by this at all.

liach:
Field order!

Grinch:
Apple cider. (note: This was requested by Zidane)

JBYoshi:
So this is what 🔶 meant...

orange

I'd add to the puns if I could "concentrate" enough to think of one 😉

ryantheleach:
I'm defaulting to rejecting this PR outright, this whole class is a mess.

I get having our own Color interface could be handy for a few reasons, but having fields inside it that are tightly coupled with any one particular color palette (wool, dyes, sheep, fireworks, text, concrete, armor) seems beyond ridicule. (this isn't your fault it's ours, clearly)

You stated that you got it from the standard web color orange which whilst I understand when talking to you on discord is a useful color for your particle effects, doesn't really seem to belong in this class as is.

I could maybe understand pulling it if the color was related to say, dye colors, but then the dye colors class has that down pat, and you can do DyeColors.ORANGE.getColor which still doesn't match the color you want or need. Nor does `ColorUtil.fromColor(DyeColors.ORANGE.getColor())` translate correctly to the EnumDyeColor in implementation, but that's another story and possibly related to 1.12 world of color update. 

Yes this class needs work, but I don't think this is the right way to do it, and I don't think people want to accept this change for one person only to rip it out and break your plugin shortly afterwards.

I love the fun and puns, and I hope this causes someone to refactor the colors class sooner rather then later.

SimonFlash:
You're being mean 😢 

But in all serious, this is something I've known for a while now. While I've had some fun with this PR (as have some others, it seems), the fact of the matter is that this is overlooking the bigger issues.

@Meronat and I have discussed the possibility of rebuilding the `Color` class and standardizing the use of colors throughout Sponge. As Ryan and I discovered, common's `ColorUtils` is completely broke and needs one heck of a tune-up.

I'm glad that this 'simple' PR has brought forth some deeper questions about some of the things in Sponge that we might take for granted - even something as basic as colors is not safe.

I'm more than willing to proceed with refactoring these classes and trying to update them to a standard, however I'd like to get the approval of a higher-up first as these will undoubtedly be breaking changes for plugins using the `Color` class (or we can keep the mis-matched names, but I'd prefer not).

In the end, it appears this has been a fruit-full endeavor after all. 🍊

ryantheleach:
@SimonFlash If you want to take this on I'm sure everyone would appreciate it. But any breaking changes would either need to be in pretty quick, or delayed until API8.

If you need to take longer I'd like to see you try and tackle the potential problems in the color utils first, and confirm whether it's due to the World Of Color update or if they have been broken since before 1.12

Also creating a test plugin would be useful, one that runs through the different color palettes, and tests serialization, as well as converting between dyecolors to implementations and back, etc.

Of course, all of this is idealistic, I'd appreciate any improvement over the current situation.

JBYoshi:
Here's another way to get `Color` objects that already works with the current API:

`DyeColors.ORANGE.getColor()`

**EDIT:** Never mind, already posted above.

SimonFlash:
This is going to be a lot harder than it might look.

I started making some changes to the `Color` class and updating the tests and in the process seem to have found some type of bug or weird result with `java.awt.Color`s. White, for example, is `255, 255, 255` in rgb or `0xFFFFFF` in hex code (which is `16777215`). This results in the following expression in the constructor:

```java
// alpha, red, green, blue
value = ((255 & 0xFF) << 24) |
                ((255 & 0xFF) << 16) |
                ((255 & 0xFF) << 8)  |
                ((255 & 0xFF) << 0);
```

IntelliJ happily points out that this is a numeric overflow, and calling `java.awt.Color.WHITE.getRGB()` returns `-1`. Unless there are some *really* weird things going on behind the scenes that I'm not aware of, this is a huge issues as basically renders any of the `java.awt.Color` colors inaccurate. Please tell me I'm not loosing my sanity... at least from this.

Anyways, I'd like to get a discussion going about this with some of the core developers as it's not going to be a simple PR and will likely result in some breaking changes - and I haven't even looked at SpongeCommon yet.

All this for orange...

ryantheleach:
-1 seems like a normal result to me, considering that it's a packed int value. FFFFFFFF is how -1 is represented in unsigned hex (e.g. FFFFFFFF is -1 with 2's complement encoding).

related: https://github.com/SpongePowered/SpongeAPI/issues/1236

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -83,7 +87,7 @@
      * @return The color object
      */
     public static Color ofHex(int hex) {
-        checkArgument(0 <= hex && hex <= 0xFFFFFF, "hex (" + hex + ") must be in range 0 to 0xFFFFFF");
+        checkArgument(0 <= hex && hex <= 0xFFFFFF, "hex (%s) must be in range 0 to 0xFFFFFF", hex);
```
ST-DDT: Since you show the decimal value in the bracets, it would be a good idea to show the decimal value for `0xFFFFFF` as well.

`"hex (%s) must be in range 0 to 0xFFFFFF (16777215)"`

SimonFlash: Good point; I can also include the hex value `0xFF` with `ofRgb` errors for consistency.

Would it be beneficial to include a print out of the hex value as well using `Integer.toHexString(hex)` and likewise for the `ofRgb` messages?

SimonFlash:
One of the things I've been wondering about is how do we want to structure the static `Color` fields used throughout the API, and I believe this starts with the colors offered in the `Color` class.

The current API has a variety of colors seemingly without rhyme or reason - they match no palette I've found in my research (though similar in names to [dye colors](https://minecraft.gamepedia.com/Dye#Data_values), there's an extra magenta and no orange). I updated these fields to match the 16 [text colors](https://minecraft.gamepedia.com/Formatting_codes#Color_codes), but I think we should standardize how we obtain and store colors for various uses.

I'd like to propose the following changes:

* `Color`s for color codes and dyes will be stored in `TextColors` and `DyeColors` through the `#getColor` method. These colors would no longer be stored in the `Color` class.
* `TextColors` should receive a `#getBackgroundColor` for the darker color used behind each text.
* ~~`TextColors` should retrieve their colors from the `net.minecraft.client.gui.FontRenderer#colorCodes` array~~ We don't have access to this class as it's only on the client, so we will need to recreate it somewhere. `DyeColors` already gets the color from Minecraft through [MixinEnumDyeColor](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/data/types/MixinEnumDyeColor.java#L59-L65), but it might be a good idea to store it as a constant.
* The `Color` class should contain static constants for a palette of standard colors, such as either the [RGB or RYB color wheel](https://www.w3schools.com/colors/colors_wheels.asp) or the [HTML web colors](https://en.wikipedia.org/wiki/Web_colors#HTML_color_names), with [CSS orange](https://en.wikipedia.org/wiki/Web_colors#CSS_colors) of course.

Additionally, it might also be worth considering having `Color` cataloged by a `Colors` class which contains all text, dye, and common colors we use. I'm not sure what the implications of this change are, so if someone would like to weight in on that that'd be great.

Finally, a fair amount of this will be within SpongeCommon and I intended to open up a PR there for these color changes (and provide a well-needed update to our [SheepFactoryColorUtil](https://github.com/SpongePowered/SpongeCommon/blob/8bc321eaeadb16c347fae58426c069e551419be6/src/main/java/org/spongepowered/common/util/ColorUtil.java#L60) class). I'll be hoping to get to this by the end of the week.

SimonFlash:
With the PR to SpongeCommon finished (which was much faster than I expected as I didn't need to rewrite ColorUtil), this PR is complete pending feedback on the above questions and any other requested changes.

Additionally, I'd like to propose a few more notes as food for thought:

* Might we prefer using `#of` instead of `#ofHex`, `#ofRgb`, or other variants? Alternatively, rename the other `#of` methods to things like `#ofVector3i` for consistency.
* Have `TextColor#getBackgroundColor` be a default method in instead of implementing it in SpongeCommon
* Add a method for merging (dye) colors using the same formula as dying armor, shown [here](https://minecraft.gamepedia.com/Dye#Dyeing_armor).
* Would also like to reiterate the note above about having a `Colors` `CatelogType`.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -38,44 +36,68 @@
 import org.spongepowered.api.data.persistence.InvalidDataException;
 import org.spongepowered.api.data.type.DyeColor;
 
-import java.util.Objects;
 import java.util.Optional;
 
+import static com.google.common.base.Preconditions.checkArgument;
```
Deamon5550: Don't reorder this.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -38,44 +36,68 @@
...
+    public static final Color GRAY = ofHex(0x808080);
+
+    public static final Color BLACK = ofHex(0x000000);
+
+    public static final Color RED = ofHex(0xFF0000);
 
-    public static final Color BLACK = ofRgb(0x000000);
+    public static final Color MAROON = ofHex(0x800000);
 
-    public static final Color GRAY = ofRgb(0x808080);
```
Deamon5550: Just don't change these, its an absolutely pointless breaking change.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -85,33 +107,39 @@
      * @param hex The hexadecimal value of the color
      * @return The color object
      */
-    public static Color ofRgb(int hex) {
-        return ofRgb((hex >> 0x10) & MASK, (hex >> 0x8) & MASK, hex & MASK);
+    public static Color ofHex(int hex) {
+        checkArgument(0 <= hex && hex <= 0xFFFFFF, "hex value [%s] must be in range 0 to 0xFFFFFF (16777215)", hex);
```
Deamon5550: Not sure about this range check as its pretty common to have hexcodes with alpha values, better to just ignore the alpha than requiring it to be zero.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -85,33 +107,39 @@
...
      *
-     * @param red The red value
-     * @param green The green value
-     * @param blue The blue value
+     * @param hex The hexadecimal value of the color
      * @return The color object
      */
-    public static Color ofRgb(int red, int green, int blue) {
-        return new Color(red, green, blue);
+    @Deprecated
```
Deamon5550: If you're deprecating this you need to include a tag in the javadoc stating the alternative.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -191,39 +216,64 @@ public static Color mixColors(Color... colors) {
...
     /**
-     * Gets the {@code red} value of this {@link Color}.
+     * Gets the {@code hex} value of this {@link Color}.
      *
-     * @return The red value
+     * @return The color in hexadecimal
      */
-    public int getRed() {
-        return MASK & this.red;
+    public int getHex() {
```
Deamon5550: I'm not sure why we're renaming this as `getRGB` is a lot more descriptive than `getHex`. Especially since the javadocs don't even mention the ordering or what the values are (ARGB, ABGR, RGB, BGR, HSV, CMYK...),

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -191,39 +216,64 @@ public static Color mixColors(Color... colors) {
...
+    public int getRgb() {
+        return this.hex;
+    }
+
+    /**
+     * Gets the {@code red} value of this {@link Color}.
+     *
+     * @return The red value
+     */
+    public int getRed() {
```
Deamon5550: Don't reorder these methods.

Deamon5550 (changes requested):

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: This change is one of the largest reasons this PR exists. There is no reasoning behind the existing colors offered in the classes, and this attempts to standardize them to the common web colors.

It is also not a breaking change. Unused fields have been marked `@Deprecated` to be removed later, and the only field that now represents a different color is Purple, which is a shade change.

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: I felt that `getHex` more accurately described what was being returned at first, but having a better understanding of color theory now I agree with this.

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: I believe having all the getters together is much more organized than grouping by color. Because the `with` methods also create a new `Color` object, it makes more sense to me to have them separated.

ST-DDT:
IMO Color should not be a CatalogType as there are too many of them. I would agree with a named color map though.

src/main/java/org/spongepowered/api/util/Color.java
gabizou: Grouping shouldn’t change. It falls in line with much of the grouping in the rest of the api.

src/main/java/org/spongepowered/api/util/Color.java
Deamon5550: And when they are removed that is the breaking change. What is the point of removing a few colors from what is already a completely arbitrary set of colors?

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -38,135 +38,179 @@
...
+
+    /**
+     * Creates a new {@link Color} based on the red-green-blue (rgb) value for
+     * the color. Colors do not utilize an alpha modifier.
+     *
+     * @param rgb The red-green-blue value
+     * @return The color object
+     */
+    public static Color of(int rgb) {
+        checkArgument(0 <= rgb && rgb <= 0xFFFFFF, "rgb value %s (%s) must be in range 0 to 0xFFFFFF (16777215", Integer.toHexString(rgb), rgb);
```
Deamon5550: Still need to remove this, just ignore the alpha always.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -38,135 +38,179 @@
...
+     * values of the vector. Values are rounded to the nearest whole number.
      *
      * @param vector3f The vector of three floats representing color
      * @return The color object
+     * @deprecated Use {@code #of(vector3f.round().toInt())}
      */
+    @Deprecated
     public static Color of(Vector3f vector3f) {
-        return new Color(Math.round(vector3f.getX()), Math.round(vector3f.getY()), Math.round(vector3f.getZ()));
+        return of(vector3f.round().toInt());
```
Deamon5550: To be completely honest `round().toInt()` was/is completely the wrong way to handle this, if you are specifying a color as a set of floats then the floats should be in the range of 0-1 as is standard for float colors.

dualspiral:
I don't think `Color` should be catalogued (it shouldn't be forced upon plugin implementors to catalogue a colour), but there could be a `NamedColor` class that extends `Color` and `CatalogType` to facilitate this. That way, plugins could catalogue their own colours for use in other plugins - if a plugin wants to then provide a specific shade of orange and expose it - they can do using existing Sponge systems, allowing plugins to be weakly dependent even when it comes down to colour.

Without commenting on the substance of this PR - and what I've seen from a cursory glance looks reasonable - if we're pushing to remove implementation from the API, we should be doing it here too as much as possible.

I accept that the result of colour mixing should not really be different across implementations... and let's face it, the colours shouldn't at all (though, one man's orange is another man's slightly darker orange), but all these methods are implementation - different implementations might have better performing algorithms! So I believe that we might as well take this oppotunity to do this properly and put the impl into SC rather than kick the issue down the road for later.

I'll allow others to decide whether that's what is actually wanted - just my tuppence.

SimonFlash:
I like the idea of splitting this class into an API interface and Common implementation. I'm not sure how we would set this up, however, as we would need a system to reroute the default `Color#of` methods in the interface to our `SpongeColor` implementation - this is done through `GameRegistry#createBuilder`, but I don't think that would apply in the same way. Is there a system in place for this type of creation, or do we need to come up with something? (this problem is not exclusive to the `Color` class and is present in things like [`ItemEnchantment`](https://github.com/SpongePowered/SpongeAPI/blob/dccaebccc2772ad485b61aec4728235830ff4140/src/main/java/org/spongepowered/api/data/meta/ItemEnchantment.java).

With regards to mixing colors, I'm actually considering deprecating those methods - they are not used in SpongeAPI, SpongeCommon, SpongeVanilla, or SpongeForge. It might be worth it to keep a static `mixColors` method, but unless you're trying to mix colors from user input there's really no need from it, and I think that's few and far between. I attached the link in a todo a couple commits back, but I'm also in favor of changing this method to match Minecraft's calculation of [dyeing armor](https://minecraft.gamepedia.com/Dye#Dyeing_armor) instead.

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: To make it so the colors are *not* arbitrary. Though it is possible to retrieve colors from either `TextColor`s or `DyeColor`s, I think it's reasonable to have some standardized colors (especially those that aren't available above or used in multiple places through API and Common, like black) offered to developers. I've opted to use the 17 CSS named colors, which are the 16 HTML colors and orange.

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: I don't particularly have a preference on this, but I think it's probably best to throw an exception if alpha is defined to make it clear that it isn't supported. If we were to support alpha, we could either strip it away (which in my opinion masks the issue and could cause problems with converting to java colors) or keep it, which begs the question of what do we do with alpha values for `#getRgb`, `#equals`, and `#hashCode`?

dualspiral:
> as we would need a system to reroute the default Color#of methods in the interface to our SpongeColor implementation 

Not really because we can use deprecations. You'd do this in two stages:

* API N, Create `NamedColor` (extends `Color` and `CatalogedType`) and `NamedColors` (which contains `NamedColor`s), create `NamedColor` registry.
* API N, Deprecate any static field in `Color`, encourage the use of the `NamedColors` instead.
* API N+1, delete deprecated fields and methods.
* Maybe create a `Color.Builder` that can perform the function of the various utility methods (note that any method names given here are just for illustrative purposes - the idea being that each method will reset that value, hence the word `set`):

```java
Color.builder().setHex(hex).build();
Color.builder().setRed(red).setBlue(red).setGreen(red).build();
// etc
```

You could also potentially create a builder for mixing colours (or even use the same one if you are creative) `Color.mix(colour).with(colour).with(colour).build();` - but I would check to see whether other plugins use this (my guess is it's unlikely) before doing any extra work. We can always re-add it later, we have the commit history already so we can redo this.

Before you go ahead and do this, it's wise to seek other's opinions on namings, direction, etc. I just think it's wise to take a step back and do this properly, rather than just rearrange some methods on an API class and add impl to the API when we should be going in the opposite direction,

dualspiral:
Oh, just to add:

> default Color#of methods in the interface to our SpongeColor implementation 

You'd route them to the relavant builder, that's all, so in the example above, `Color.of(hex)` could be a shortcut for `Color.builder().setHex(hex).build();`

Deamon5550 (changes requested):

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: Noted. I'm going to wait for another person or two to weigh in on this. Ignoring it would probably be the best in terms of support for java colors that may half alpha, but it puts unnecessary pressure on the class to maintain it.

Copied from the earlier review:
> I don't particularly have a preference on this, but I think it's probably best to throw an exception if alpha is defined to make it clear that it isn't supported. If we were to support alpha, we could either strip it away (which in my opinion masks the issue and could cause problems with converting to java colors) or keep it, which begs the question of what do we do with alpha values for #getRgb, #equals, and #hashCode?

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: I agree, it's a ridiculous way of handling it which is why I deprecated the method. I like the idea of interpreting this as 0-1 values as I know of at least one place in Sponge where this is used and it's how Mojang stores colors as well. This would change the method to have a different result, but I can probably check the vector for values between 0 and 1 and handle it from there. Thoughts?

SimonFlash:
If we decide to go with a `CatelogType` approach, we can definitely use the builder registry to handle the connection from API to Common. That's a huge step to take though (especially for a relatively simple class) and I'm not sure if a builder is the correct design pattern here as each method would basically do it's own thing rather than stacking, which is generally what's expected from a builder. If you and the other developers are fine with going that direction, then I'm all in. While this originally started out as a couple simple changes, it's quickly become apparent that there's some fundamental issues going on here as well.

dualspiral:
Whether or not the builder _is_ the right design pattern here, it’s how we separate construction of our API elements from the implementation in a clean, non hacky manner, without resorting to performing either some hacky reflection or ASM to write portions of the class at runtime. Having static methods (such as `of`) that is expected to have construction code within it makes it incredibly difficult to do that. There are other places in the API where a builder isn’t quite right, but used because it’s the tool of choice and we wrap our preferred style for an API around that.

> each method would do its own thing rather than stacking

[Another builder that is the same can be found here](https://github.com/SpongePowered/SpongeAPI/blob/4fe3fffd10b2a7762a3b2f7ac00f670282a9301f/src/main/java/org/spongepowered/api/world/gen/populator/Fossil.java#L62). As an aside, Zidane suggested to me in my commands PR that such methods should have a `set` prefix, why I’ve done that up there - though it’s not done elsewhere in the API when perhaps it should have been. However, it is ultimately to aid clean separation.

That’s not to say that you _can’t_ have static `of` methods, but they should become convenience methods that call the builder. So, `of(hex)` would call `Sponge.getRegistry().createBuilder(Color.Builder.class).setHex(hex).build()`, it looks clean from an API consumer’s side and it gives us the separation that we need. There is potentially an extra penalty where we need to check the state of the builder on build or when each element is called, but ultimately, I don’t see that as a problem.

SimonFlash:
I've migrated the majority of the `Color` class to Common and have replaced it with the new `Color` interface and `Color.Builder` builder. There are still no breaking changes caused by this PR with the `Color` interface directly, but there may be from moving the data builder to Common.

Discussion about the implementation should be moved to the PR in [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1625).

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -387,65 +366,42 @@ public Color mixWithDyes(DyeColor... dyeColors) {
...
-    @Override
-    public int hashCode() {
-        return this.rgb;
-    }
-
-    @Override
-    public boolean equals(Object obj) {
-        return this == obj || obj != null && getClass() == obj.getClass() && this.rgb == ((Color) obj).rgb;
-    }
+    interface Builder extends ResettableBuilder<Color, Builder> {
```
liach: Individual red, blue, green setters?

src/main/java/org/spongepowered/api/text/format/TextColor.java
```diff
@@ -50,7 +50,9 @@
      *
      * @return the RGB background color of this text color
      */
-    Color getBackgroundColor();
+    default Color getBackgroundColor() {
+        return Color.of(getColor().toVector3i().div(4));
```
Deamon5550: Don't default implement this.

src/main/java/org/spongepowered/api/util/Color.java
liach: Since you changed a concrete class to an interface, you should simply remove all the deprecated methods scheduled for removal.

src/main/java/org/spongepowered/api/util/Color.java
```diff
@@ -387,65 +366,42 @@ public Color mixWithDyes(DyeColor... dyeColors) {
...
+         * Sets the red, green, and blue values for this color by taking the
+         * average of each for every color in the array.
+         *
+         * @param colors The colors to mix
+         * @return This builder
          */
-        public Builder() {
-            super(Color.class, 1);
-        }
+        Builder mix(Color... colors);
```
Deamon5550: A better pattern here is that mix takes one color and mixes with the current color of the builder

Deamon5550 (changes requested):
Please remember (whoever is eventually thinking of potentially merging this) that changing a class to an interface is a breaking change in that it represents a binary breakage with any compiled plugin calling any method on the type at all.

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: I did that in one attempt, but there isn't really a point to it. I'd say about 99% of use cases are going to be through the default methods, and it's both easier and faster to set them all at the same time. This also ensures that there's never a partial color in the builder, which will be 5x as important if we decide to ignore alpha values rather than removing them (which means we need to save a rgb value as well that would need to be updated for each of those method calls). It's an overhead that I don't think is worth it in the slightest even though it's a bit against the builder mentality - but so is the entire class anyways.

src/main/java/org/spongepowered/api/text/format/TextColor.java
SimonFlash: I understand it's basic, but it is a valid property that every `TextColor` has. Are you suggesting to revert it to a interface method or remove it entirely?

src/main/java/org/spongepowered/api/util/Color.java
SimonFlash: That would have default implementations mix with the color black. Any static method for mixing would need to pull out the first element of the array, build thay, then copy the remaining elements to a new array to be used with this method. The mix methods aren't used in Sponge (bar tests) and are probably used in few to no plugins. I'd be more in favor of removing them entirely.

SimonFlash:
The class to interface change being breaking at the binary level is news to me, so I withdraw my earlier statemement - this *is* a breaking change.

src/main/java/org/spongepowered/api/text/format/TextColor.java
Deamon5550: Just make it a regular interface method, the only things that should be default implemented is trivial calls to overloaded methods.

SimonFlash:
Small fix in the above commit for a requested change.

Here are the above potential projects remaining for this PR:

* Have `Color` be a `CatelogType`? (various options, +1 to `NamedColor` from dual)
* Remove `@Deprecated` methods (recompile to potential refactoring)

SimonFlash:
As API 7 is nearing a stable release, I'd like to move for this PR to be accepted by the end of the month. There are still two discussion points remaining with regards to API:

* Should we remove `@Deprecated` methods? While this doesn't provide developers with a lot of time to address any issues, this PR already requires a recompile and the effects of these changes are limited.
* I would also like to update the various `mixColor` methods, especially for dye colors. These methods aren't used in Sponge (and likely aren't used outside of Sponge), as well as being poor/inaccurate implementations.
  * Mixing RGB colors isn't an easy process, and while averaging all their components is generally close enough it isn't ideal and shouldn't be used in any type of environment. I suggest removing this from the API entirely. If someone would like this effect, they can convert the color to a Vector3i and use the methods from there (which is much more efficient for two values anyways).
  * Mixing dye colors has the same flaws as rgb colors (as the use the same method) but is much more obscure. Additionally, the formula used is different from the one used by Minecraft to calculate the result of blending dyes with armor. I suggest either the removal of these methods completely or converting them to be default methods of Color that use the right formula for mixing dyes.

Beyond that, if anyone has any remaining suggestions now's the time.

dualspiral:
The benefit of leaving the deprecated methods, even if a plugin requires a recompile, is that you can attach documentation to them, allowing you to point developers to the new, improved methods. So, leaving them is not a bad thing for that reason.

I still feel like we should have a named colour catalogue type for the common colours, if only to bring this in line with the rest of the API design, singular class name for the type, plural for the constants.

gabizou:
I'm going to have to pass on having this merged for API 7. API 8 (Minecraft 1.13's compatibility update) is more likely where it will have a more fitting merge. There's more about this PR that I'd develop a bit further, as @dualspiral has mentioned, that I can't anticipate enough time to dedicate before API 7's release date.

SimonFlash:
So trying to work this down to a resolution, have we decided on making colors a `CatelogType` or is that still up in the air? With regards to the `mixColor` methods as explained above, what are your thoughts on the different options presented?

Additionally, what are the other things with this PR that you'd develop further? Adding a `CatelogType` is a bit of a project, but isn't something unique that needs to be developed specifically for this PR. From the sound of it there seems to be something I'm missing, so what other components are there?

dualspiral:
I would like to see convergence on a consistent API with the broader Sponge ecosystem. My thoughts are:

* `Color` as a basic interface probably should _not_ extend `CatalogType`.
* A new `NamedColor` _should_ extend `CatalogType` for our... well, named colours.
* A `NamedColors` or `Colors` (though probably the former) class with static references to all _current_ colours and any others that make sense to have.
* We might not want to allow `NamedColor`s to be registerable. There might be a way to hide the `NamedColor` interface in that case, but it's not something I've looked into.

I'm not happy with the deprecation of the colours we currently have in order to favour the 16 html colours and orange. Why the 16 standards and orange? I get that this PR was originally about adding the colour orange, but we have significantly expanded the goal here and removing the traditional Minecraft colours that a lot of plugins might still rely on shouldn't be removed. Prime example - our `TextColor`s, the ones that were originally the minecraft colour codes, are backed by some of this colours, and removing the deprecated colours would break that. We should closely match what Minecraft does with colours and _all_ pallets, not just suggest 16+1 colours from an arbitrary standard which do not align with Minecraft.

On mixing colours - I don't know if it's used. I don't know why it's in there and who put it in - I'm not going to pass judgment on whether it should be there or not. Note that my comment on the Common PR was on the basis that it would be there.

SimonFlash:
To start with mixing colors, I completely get what you're going for there. The actual methods aren't used anywhere apart from Sponge, and as a user I don't think I'll even be in the situation where I'll want to mix colors like that as opposed to using my own with the sole exception of mixing dyes for armor (which uses a different formula anyways). This is really a gabizou call on whether to keep this or not, but I don't see any reason why it's necessary.

***

The colors Minecraft uses are very much all over the place, so it's near impossible to find a standard. While dyes themselves have a color, dyed blocks each have their own color schemes so it doesn't make sense to attempt to catalog those (you'd have over 100 colors and they'd have to be named `WOOL_RED` and such). This isn't feasible or helpful to us, so I don't think we should try documenting all colors 

The only color codes that are reasonably standardized in Minecraft are the chat colors, which are almost identical to the CGA colors. The CGA colors using a particular formula based on a colors number with one exception - color 6, which is brown instead of an olive color. This is also the sole exception in Minecraft, with brown being replaced with orange instead.

If we were to use text colors, this wouldn't cover some of the main colors and are slightly off from the standard color codes. Additionally, note that these colors are technically a catalog type through `TextColors` as they have a `getColor` method (and now a `getBackgroundColor` method). 

Prior to this PR, the `Color` class contained a variety of colors that didn't match up with any standard color palette - it was close to a couple, but never exact. The old class had `MAGNETA`, `PURPLE`, and `DARK_MAGENTA` (which matches no set in Minecraft at all) and the color code choices more closely matched the web colors as opposed to Minecrafts (such as `GREY` being `0x808080` instead of `0xAAAAAA` (text grey) or `0x555555` (text dark grey), or `DARK_CYAN` being `0x008B8B` (which is somewhat close to cyan, presumably) over `0x00AAAA` (text cyan) and so on). Simply put, there was no rhyme or reason to the colors present, their names, and their actual color values.

I wanted to standardize this to common colors that developers might want to use, so I spent a lot of time researching the different possibilities - color wheels, the aforementioned CGA colors, and HTML web colors - before decided on a superset of the web colors and CSS orange. This covers the main colors that a user might want and, while not a 'perfect' selection it is much more complete and orderly than what was existing. I'd also support included some other colors like brown and pink, but at some point we have to draw a line. This is something can isn't too difficult to change, but I felt this was a reasonable selection for our use case.

***

I agree that `Color` should not be a `CatalogType` for the same reasons positions are not cataloged to a `World` - though technically finite, it's not what the class represents and would be unfeasible to support.

Adding in a `NamedColor` class to handle registered colors would work, but I personally see this as a lot of additionally boilerplate for next to no benefits, *especially* if new colors can't be registered (which even if allowed wouldn't have much of a significant benefit - it's not like this is a `Key` or `Item` with a lot of unique properties that needs to be tracked and retrieved frequently and potentially across multiple plugins; it's literally just a color. I see the largest benefits of the registry being the ability to document objects that don't exist within Sponge and aren't available until Minecraft begins and register it's own type and retrieve it - as we're not using either of those, I just see having a `CatelogType` as being unnecessary additional code to have static fields. I would also prefer `NamedColors` over `Colors` if this would be the case, as this matches the Sponge style more and `Colors` implies `Color` is the implementing `CatalogType`.

An alternative is to create a `NamedColors` class that simply contains static fields for any colors we'd like to offer. This avoids jumping through hoops with the registry while maintaining the same organization effects. Nothing else in the API is in the same situation as colors, so I don't see a reason to make it structured the same as something it doesn't represent.

gabizou:
> rior to this PR, the Color class contained a variety of colors that didn't match up with any standard color palette - it was close to a couple, but never exact. The old class had MAGNETA, PURPLE, and DARK_MAGENTA (which matches no set in Minecraft at all) and the color code choices more closely matched the web colors as opposed to Minecrafts (such as GREY being 0x808080 instead of 0xAAAAAA (text grey) or 0x555555 (text dark grey), or DARK_CYAN being 0x008B8B (which is somewhat close to cyan, presumably) over 0x00AAAA (text cyan) and so on). Simply put, there was no rhyme or reason to the colors present, their names, and their actual color values.

All of these colors (most all of them at least) were reverse engineered based on the set of Dye colors available in Minecraft, as originally we had to design the Color class for use with DyeColors and interchangeably with leather armor color mixing (which is what the mixing brings to play). As far as representing the *possible* list of colors available, it is very well possible to have the possible color palettes for the variety of use cases, being DyeColors as they are represented, Text colors as they are represented, and maybe some other collection of standard color palette, such as the HTML palette. 

I haven't quite built up a proper response for the rest of the PR, but my response for colors and their mixing is quite clear in my opinion.

SimonFlash:
I'm sure there was a reason for having the previous colors offered at the time, but we're moving past that now. If the goal of having `mixColors` was for mixing dyes with armor, then it should use the [correct formula](https://minecraft.gamepedia.com/Dye#Dyeing_armor). As the text and dye color palettes are already contained within `TextColor` and `DyeColor`, there isn't a reason to replicate them elsewhere. That's why I opted to include only a standard palette of commonly used colors.

I might have missed something, but I haven't seen a response from you regarding mixing colors that presents a clear resolution, so I'd have to ask you to clarify on that.

ryantheleach:
Rather then longer paragraphs, can someone make an actionable summary (like in bullet points / check marks) so we know where this PR stands?

SimonFlash:
Here's the remaining discussion points that I am aware of:

- [x] Depreciation of `mixColor` methods
- [ ] Method for dye color mixing for armor (with correct formula)
- [ ] A `CatalogType` for provided colors instead of static fields

My thoughts and related responses can be found above.

Additionally, there is still a point about Forge support with `ColoredDataProcessor` for determining what things support colors. Still waiting for feedback on this as I'm not familiar with Forge.

SimonFlash:
Latest commit depreciates the mix methods and removes the `Builder#mix` method in Common. I've also added in brown as a color which I apparently forgot and fixed a bug where `from` and `reset` returned `DataBuilder` instead of `Builder`.

Additionally, the following is the correct method for mixing dye colors (see `net.minecraft.item.crafting.RecipiesArmorDyes#getCraftingResult` for comparison).
```java
static Color mixAsDyes(Color... colors) {
    int red = 0, green = 0, blue = 0, max = 0;
    for (Color color : colors) {
        red += color.getRed();
        green += color.getGreen();
        blue += color.getBlue();
        max += Math.max(color.getRed(), Math.max(color.getGreen(), color.getBlue()));
    }
    float mul = max / (Math.max(red, Math.max(green, blue))) / colors.length;
    return builder().rgb((int) (red * mul), (int) (green * mul), (int) (blue * mul)).build();
}
```

dualspiral:
> As the text and dye color palettes are already contained within TextColor and DyeColor, there isn't a reason to replicate them elsewhere

No - the backing `Color`s should be catalogued. Anything catalogued that uses a colour should use a catalogued named colour too. For `TextColor`, as you propose adding a background colour, the backing colours should be catalogued. `DyeColor`, on the other hand, is just a wrapper for a colour, but if we have a wrapped colour and colours should be catalogued, I'd argue that those colours should be catalogued too.

Catalogue the standard colours (that's fine), but catalogue the backing text and dye colours too. As @gabizou said, there are reasons as to why the colours were chosen. Just because it is also represented in the `TextColor` class (for example) doesn't mean we should hide it there.

Also, I read the response as _keeping_ the mix colour methods. Now knowing what I know (I've never cared for mixing colours for armour and whatnot), I can see the value in the methods. Redo it to use the method you pinpointed, sure, but there are uses for mixing colours ahead of time.

I think the following needs to happen:

* Keep the `mix` methods on colours - at least provide a method that allows this
* Correct the mix method to use the standard method you've discovered
* Create a named colour catalogue type
  * Ensure that all standard/expected text colours and dye colours are catalouged (could prefix colours with `TEXT_` or `DYE_` to remove ambiguity if we really need to.)
  * Add any appropriate standard colours (maybe this could be `STANDARD_RED` etc.)

It might be worth waiting for gabi's input before continuing as he is ultimately the person who signs off on the API design.

ImMorpheus:
> The only color constants to be in the API will be the ones that exist in the base game.
> This is a Minecraft API

PR #1700, opened 2017-11-23
Deprecate BedData in favor of DyeableData.

JBYoshi:
`BedData` only contains a key for dye colors. We already have that in `DyeableData`.

Related: SpongePowered/SpongeCommon#1611

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/tileentity/BedData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/tileentity/BedData.java
@@ -27,13 +27,17 @@ package org.spongepowered.api.data.manipulator.mutable.tileentity;
 import org.spongepowered.api.block.tileentity.Bed;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.manipulator.immutable.tileentity.ImmutableBedData;
+import org.spongepowered.api.data.manipulator.mutable.DyeableData;
 import org.spongepowered.api.data.type.DyeColor;
 import org.spongepowered.api.data.value.mutable.Value;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/data/manipulator/mutable/tileentity/BedData.java
```diff
@@ -1,46 +0,0 @@
...
-import org.spongepowered.api.data.manipulator.DataManipulator;
-import org.spongepowered.api.data.manipulator.immutable.tileentity.ImmutableBedData;
-import org.spongepowered.api.data.type.DyeColor;
-import org.spongepowered.api.data.value.mutable.Value;
-
-/**
- * A {@link DataManipulator} handling the various data of a
- * {@link Bed}.
- */
-public interface BedData extends DataManipulator<BedData, ImmutableBedData> {
```
gabizou: Since we can technically support it, this should be deprecated in use instead of flat out removed.

gabizou (changes requested):
Deprecation first, removal in next version.

ryantheleach:
BedData either probably needs to have a data value whether it is occupied / being slept in, or a new DataManipulator found to cover this purpose.

Cybermaxke:
If this will be added in API 8, it will very likely be removed anyway, due the 1.13 blockstate changes. Beds will no longer be tile entities.

parlough:
@Cybermaxke I think this is targeted for API 7 just BedData will be deprecated instead of removed.

src/main/java/org/spongepowered/api/data/manipulator/mutable/tileentity/BedData.java
JBYoshi: Done.

PR #1702, opened 2017-11-24
Merge duplicate ModifierFunctions.

JBYoshi:
Fixes SpongePowered/SpongeCommon#1614.

Test:
1. Run the command `/give <your username> diamond_sword 1 0 {ench:[{id:0,lvl:3},{id:0,lvl:3}]}`
2. Attack any entity with the sword.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
+++ b/src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
@@ -56,20 +56,20 @@ public abstract class AbstractModifierEvent<T extends ModifierFunction<M>, M> ex
     protected ImmutableList<T> init(double originalValue, List<T> originalFunctions) {
         final ImmutableList.Builder<Tuple<M, Double>> modifierMapBuilder = ImmutableList.builder();
         final ImmutableList.Builder<T> functionListBuilder = ImmutableList.builder();
-        final ImmutableMap.Builder<M, Double> mapBuilder = ImmutableMap.builder();
+        final Map<M, Double> mapBuilder = new LinkedHashMap<>();
         double finalDamage = originalValue;
         for (T tuple : originalFunctions) {
             this.modifierFunctions.add(convertTuple(tuple.getModifier(), tuple.getFunction()));
             double tempDamage = checkNotNull(tuple.getFunction().applyAsDouble(finalDamage));
             finalDamage += tempDamage;
             modifierMapBuilder.add(new Tuple<>(tuple.getModifier(), tempDamage));
-            mapBuilder.put(tuple.getModifier(), tempDamage);
+            mapBuilder.merge(tuple.getModifier(), tempDamage, Double::sum);
             this.modifiers.put(tuple.getModifier(), tempDamage);
             functionListBuilder.add(convertTuple(tuple.getModifier(), tuple.getFunction()));
         }
         this.originalFinalAmount = finalDamage;
         this.originalModifiers = modifierMapBuilder.build();
-        this.originalModifierMap = mapBuilder.build();
+        this.originalModifierMap = ImmutableMap.copyOf(mapBuilder);
         return functionListBuilder.build();
     }
 


src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
```diff
@@ -56,20 +56,24 @@
...
             this.modifierFunctions.add(convertTuple(tuple.getModifier(), tuple.getFunction()));
             double tempDamage = checkNotNull(tuple.getFunction().applyAsDouble(finalDamage));
             finalDamage += tempDamage;
             modifierMapBuilder.add(new Tuple<>(tuple.getModifier(), tempDamage));
-            mapBuilder.put(tuple.getModifier(), tempDamage);
+            if (mapBuilder.containsKey(tuple.getModifier())) {
+                mapBuilder.put(tuple.getModifier(), mapBuilder.get(tuple.getModifier()) + tempDamage);
+            } else {
+                mapBuilder.put(tuple.getModifier(), tempDamage);
+            }
```
ST-DDT: Isn't it possible to simplyfy this like:

`mapBuilder.merge(tuple.getModifier(), tempDamage, (a,b) -> a + b);`

[`Map#merge(K key, V value, BiFunction<? super V,? super V,? extends V> remappingFunction)`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#merge-K-V-java.util.function.BiFunction-)

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
JBYoshi: I'll do that.

phit:
requesting a backport for this when it gets merged, as this causes a lot of random crashes due to skeleton traps quite often having duplicate enchants

ST-DDT (approved):

ryantheleach:
I havn't tested this, but LGTM on the surface.

1. We need to verify whether sum is appropriate.
2. We need to verify this fixes the crash.
3. Tested with multiple armor protection enchants.

I trust JB tested 2. given the sword command, and that previously caused the problem without this change?

Sum makes it look like it's impossible to disambiguate the sources, but would fix the issue for what's a pretty clear edgecase, I'm just mildly concerned at how it acts with armor.

I'm happy as is. Just might want to run it past @Gabizou.

ryantheleach (approved):

src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
```diff
@@ -56,20 +56,20 @@
...
-        final ImmutableMap.Builder<M, Double> mapBuilder = ImmutableMap.builder();
+        final Map<M, Double> mapBuilder = new LinkedHashMap<>();
         double finalDamage = originalValue;
         for (T tuple : originalFunctions) {
             this.modifierFunctions.add(convertTuple(tuple.getModifier(), tuple.getFunction()));
             double tempDamage = checkNotNull(tuple.getFunction().applyAsDouble(finalDamage));
             finalDamage += tempDamage;
             modifierMapBuilder.add(new Tuple<>(tuple.getModifier(), tempDamage));
-            mapBuilder.put(tuple.getModifier(), tempDamage);
+            mapBuilder.merge(tuple.getModifier(), tempDamage, Double::sum);
```
gabizou, line 66: May want to comment why we merge here.

gabizou (commented):
Just add a comment and it should be ready to merge!

src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
```diff
@@ -56,20 +56,20 @@
...
         final ImmutableList.Builder<Tuple<M, Double>> modifierMapBuilder = ImmutableList.builder();
         final ImmutableList.Builder<T> functionListBuilder = ImmutableList.builder();
-        final ImmutableMap.Builder<M, Double> mapBuilder = ImmutableMap.builder();
+        final Map<M, Double> mapBuilder = new LinkedHashMap<>();
         double finalDamage = originalValue;
         for (T tuple : originalFunctions) {
             this.modifierFunctions.add(convertTuple(tuple.getModifier(), tuple.getFunction()));
             double tempDamage = checkNotNull(tuple.getFunction().applyAsDouble(finalDamage));
             finalDamage += tempDamage;
             modifierMapBuilder.add(new Tuple<>(tuple.getModifier(), tempDamage));
```
ryantheleach, line 65: Does this line need to be changed as well?

src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
```diff
@@ -56,20 +56,20 @@
...
-            mapBuilder.put(tuple.getModifier(), tempDamage);
+            mapBuilder.merge(tuple.getModifier(), tempDamage, Double::sum);
             this.modifiers.put(tuple.getModifier(), tempDamage);
```
ryantheleach, line 67: and this one?

ImMorpheus:
Regarding ryan's comment (plus a few tests):

First, let's use a slightly different sword:
`/give <player> diamond_sword 1 0 {ench:[{id:17,lvl:3},{id:17,lvl:2}]}`
(Smite III + Smite II)

Now, take this test plugin:

```
    @Listener
    public void onA(AttackEntityEvent event) {
        System.out.println("OriginalDamages");
        for (Map.Entry<DamageModifier, Double> a : event.getOriginalDamages().entrySet()) {
            a.getKey().getCause().first(Enchantment.class).ifPresent(enchantment -> System.out.println(enchantment.getId() + " - " + a.getValue()));
        }

        System.out.println("\nModifiers");
        for (DamageFunction ab : event.getModifiers()) {
            ab.getModifier().getCause().first(Enchantment.class).ifPresent(enchantment -> System.out.println(enchantment.getId() + " - " + ab.getFunction().applyAsDouble(7)));
        }

        System.out.println("\nOriginalFunctions");
        for (ModifierFunction<DamageModifier> ac : event.getOriginalFunctions()) {
            ac.getModifier().getCause().first(Enchantment.class).ifPresent(enchantment -> System.out.println(enchantment.getId() + " - " + ac.getFunction().applyAsDouble(7)));
        }

        System.out.println("\n----------------------------------\n");
        System.out.println("OriginalDamage: " + event.getOriginalDamage());
        System.out.println("BaseAmountDamage: " + event.getBaseOutputDamage());
        System.out.println("FinalOutputDamage: " + event.getFinalOutputDamage());
        System.out.println("OriginalFinalDamage: " + event.getOriginalFinalDamage());
        System.out.println("__________________________");

        DamageModifier test = event.getModifiers().get(1).getModifier(); //ignore this for now
        System.out.println(event.getOriginalModifierDamage(test)); 
    }
```
Result:
```
 OriginalDamages
 minecraft:smite - 12.5
 
 Modifiers
 minecraft:smite - 7.5
 minecraft:smite - 5.0
 
 OriginalFunctions
 minecraft:smite - 7.5
 minecraft:smite - 5.0
 
 ----------------------------------
 
 OriginalDamage: 7.0
 BaseAmountDamage: 7.0
 FinalOutputDamage: 19.5
 OriginalFinalDamage: 19.5
 __________________________
 7.5 
```

> Sum makes it look like it's impossible to disambiguate the sources, but would fix the issue for what's a pretty clear edgecase, I'm just mildly concerned at how it acts with armor.

Not exactly, since `protected List<Tuple<M, Double>> originalModifiers;` would contain duplicate entries because you've added duplicate modifiers while iterating in that method. 
Cool, does that fix the issue ? No, It introduces new ones :D 

Classes which extend AbstractModifierEvent like AbstractAttackEntityEvent contain methods which iterate on the assumption that that list does not contain duplicate entries (ex. [getOriginalModifierDamage](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/event/impl/AbstractAttackEntityEvent.java#L59) )
This leads to inconsistency like: `System.out.println(event.getOriginalModifierDamage(test)); ` printing 7.5 (should be 12.5).

src/main/java/org/spongepowered/api/event/impl/AbstractModifierEvent.java
liach: No, they are lists

JBYoshi:
Seems like the better option would be to not send duplicate enchantments in common then. There's probably plugin code that depends on not having duplicates as well.

I'm doing some testing right now, and I'll submit a PR for that once I'm done.

ryantheleach:
I think we have no choice but to send the duplicate enchantments? Can the math be made to work?

JBYoshi:
https://github.com/SpongePowered/SpongeCommon/pull/1627

ImMorpheus:
IMO it's ok to send duplicate enchantments. Another dirty solution would be to change method that rely on originalModifiers to use originalModifierMap.
For example, using getOriginalDamages().get(damageModifier); inside getOriginalModifierDamage.

Issue #1667, opened 2017-10-14
Get shulker box Item/BlockType from DyeColor


VapidLinus:
I have a `DyeColor` and want to get the `ItemType` or `BlockType` of a shulker box that represents that colour.

The different colours of the shulker box are split into different types and there doesn't seem to be any API methods for getting a type of a specific "block" that represents a certain color.

At the moment I came up with a pretty dirty util method that does this for me:

```java
public class ShulkerBoxTypeProvider implements DyeBasedTypeProvider {
	@Override
	public Optional<ItemType> getItem(DyeColor dyeColor) {
		return Sponge.getRegistry().getAllOf(ItemType.class)
				.stream()
				.filter(itemType -> itemType.getName().endsWith("_shulker_box") &&
						itemType.getName().contains(dyeColor.getName()))
				.findFirst();
	}

	@Override
	public Optional<BlockType> getBlock(DyeColor dyeColor) {
		return Sponge.getRegistry().getAllOf(BlockType.class)
				.stream()
				.filter(blockType -> blockType.getName().endsWith("_shulker_box") &&
						blockType.getName().contains(dyeColor.getName()))
				.findFirst();
	}
}
```

Then I can do `TypeProviders.SHULKER_BOX.getItem(DyeColors.GREEN)` to get `ItemTypes.GREEN_SHULKER_BOX`.

This is pretty ugly though and is something I feel should be in the API. The upcoming [1.13 Flattening](https://minecraft.gamepedia.com/1.13/Flattening) will contribute to this issue. I would have to do something similar to get a specific wool colour, or getting a sapling or wooden plank from a "`WoodType`", etc.

I was recommended by @dualspiral to post this as an issue to document my use case of getting a shulkerbox from a `DyeColor` and I believe an API that solves my use case could also be made to solve upcoming use cases in 1.13 to get item/blocktypes from certain properties such as "WoodType", "DyeColor" or even "SlabType", etc.

Some problems I can quickly think of:
- Getting a shulkerbox from a `DyeColor`. Glass from color. Getting other types from other properties? Imagine "SlabType", "MusicDiscType", "WoodType", etc.
- How to check if something is a wooden plank? Without checking if the block name's ends with "_planks". Same with beds, glass, saplings, leaves, pots, player heads, etc.
[...]

PR #1704, opened 2017-11-27
1.13 - The Technical and Aquatic Update

Cybermaxke:
Very WIP, I am opening this PR now to start a discussion on what can be reused/needs to be removed. Also designing a system to easily get variants of certain block types. For example getting a slab with the material oak.

Related to:
https://github.com/SpongePowered/SpongeAPI/pull/1715 - The variant system
https://github.com/LanternPowered/LanternServer/pull/48
https://github.com/SpongePowered/SpongeAPI/pull/1742

We have to decide what to do with the following data types, could it be reused for the variant system or should be deleted. I already removed a few types, feel free to comment if you don't agree with me removing them.

#### Modify/remove:

Catalog types and data:
- [x] BrickTypes - _Removed_
- [x] BrickData - _Removed_
- [x] CoalTypes - _Removed_
- [x] CoalData - _Removed_
- [x] CookedFishes - _Removed_
- [x] CookedFishData - _Removed_
- [x] DirtTypes - _Removed_
- [x] DirtData - _Removed_
- [x] DisguisedBlockTypes - _Removed_
- [x] DisguisedBlockData - _Removed_
- [x] DoublePlantTypes - _Removed_
- [x] DoublePlantData - _Removed_
- [x] DyeColors - _Still used by other data types, used in variant system_
- [x] DyeableData - _Still used by other data types_
- [x] Fishes - _Removed_
- [x] FishData - _Removed_
- [x] LogAxes _- Removed, bark got separated from log, log uses now the Axis_
- [x] LogAxisData - _Removed_
- [x] GoldenApples - _Removed_
- [x] GoldenAppleData - _Removed_
- [x] PistonTypes - _Removed_
- [x] PistonData - _Removed_
- [x] PlantTypes - _Removed_
- [x] PlantData - _Removed_
- [x] PrismarineTypes - _Removed_
- [x] PrismarineData - _Removed_
- [x] SandstoneTypes - _Removed_
- [x] SandstoneData - _Removed_
- [x] SandTypes - _Removed_
- [x] SandData - _Removed_
- [x] ShrubTypes - _Removed_
- [x] ShrubData - _Removed_
- [x] SkullTypes - _Removed_
- [x] SkullData - _Removed_
- [x] SlabTypes - _Removed_
- [x] SlabData - _Removed_
- [x] StoneTypes - _Removed_
- [x] StoneData - _Removed_
- [x] TreeTypes - _Will be used in the variant system_
- [x] TreeData - _Removed_
- [x] WallTypes - _Removed_
- [x] WallData - _Removed_
- [x] QuartzTypes - _Removed, separate blocks, pillar variant now uses the Axis._
- [x] QuertzData - _Removed_
- [x] BigMushroomTypes - _Removed, replaced with BigMushroomPoresData_
- [x] BigMushroomData - _Removed, replaced with BigMushroomPoresData_

Other data (and keys):
- [x] SeamlessData - _Removed_
- [ ] SpawnableData

Tile entities:
- [x] Note - _Removed_
- [x] Skull - _Removed_
- [x] FlowerPot- _Removed_

#### New/Modified

Data (probably separate PRs):
- [x] LitData - _New_ (furnaces, redstone torch, redstone)
- [x] InstrumentData - _New_ (note block)
- [x] SlabPortionData (and SlabPortion) - _New_
- [x] ChestAttachmentData (and ChestAttachmentType) - _New_
- [x] InvertibleData - _New_ (daylight sensor)
- [x] SurfaceAttachmentData (floor, ceiling, wall) - _New_ (buttons, lever), is this a good name?
- [x] BigMushroomPoresData - _New_ (replaces BigMushroomData)
- [ ] UnstableData - _New_ (TNT)

Tile entities:
- [x] Player head - _New_

Populators:
- [x] Merged Shrub and Flower into Plant

Other:
- [x] Update statistics, moved to https://github.com/SpongePowered/SpongeAPI/pull/1742
- [x] Update block traits
- [x] Update particles
- [x] RecordType -> MusicDisc - In 1.13, the name 'record' is no longer used in any translation keys, item ids, etc. (item names were already Music Discs)
- [ ] ~~Remove extended block states? Everything is now controlled server side.~~ - Forge is still going to use them.


VapidLinus:
Just leaving a quick note that this PR seems to relate to my issue regarding how to get a coloured ShulkerBox from a DyeColor: https://github.com/SpongePowered/SpongeAPI/issues/1667

**edit:** music disc type, maybe?

me4502:
I’m of the opinion it’s still nice to have groupings of obvious variants, such as wool colours. Idk how others feel tho

ryantheleach:
Before you get too far into this, remember that the Data API is an abstraction over the base game.

Having this data available, even if not directly represented in the Mojang block properties was entirely the point of abstracting it in the first place.

BlockTrait API on the other hand, was supposed to be a closer to metal abstraction, that was implementation dependent.

Cybermaxke:
@ryantheleach That's why I am opening a discussion on this. And lot (block) based data is still tied quite closely to the MC internals, making it not possible to update without breaking changes (or ugly workarounds). Since there are already a lot BlockType changes, I don't see any issues with breaking the data types as well.

Cybermaxke:
@VapidLinus There is a `RecordType` catalog type.

src/main/java/org/spongepowered/api/data/type/Material.java
```diff
@@ -22,17 +22,19 @@
...
-import org.spongepowered.api.data.manipulator.mutable.block.QuartzData;
-import org.spongepowered.api.data.type.QuartzType;
+import org.spongepowered.api.CatalogType;
 
-/**
- * An {@link ImmutableVariantData} for the {@link QuartzType}. Usually
- * applicable to {@link BlockTypes#QUARTZ_BLOCK}.
- */
-public interface ImmutableQuartzData extends ImmutableVariantData<QuartzType, ImmutableQuartzData, QuartzData> {
+public interface Material extends CatalogType {
```
gabizou: After thinking about it more today, I don't like the idea of naming it `Material`. Maybe call it `Variety` and then `Varieties.OAK` and `Varieties.QUARTZ` could then work. This way you can also have `Varieties.STAIR` or `PLANK` even. This way the "type" is not defined as a Material (since stair is not a material, but a variety of a block). This also allows for us to make a combination of varieties for a specific object to "describe" it.

gabizou: Thinking on it more, `Variety` could even be introduced in API 7 (given some time this week maybe to solidify it as a possibility), in which a "moving" towards it is somewhat available before the drastic change of some of this data removal.

gabizou: After thinking about it more today, I don't like the idea of naming it `Material`. Maybe call it `Variety` and then `Varieties.OAK` and `Varieties.QUARTZ` could then work. This way you can also have `Varieties.STAIR` or `PLANK` even. This way the "type" is not defined as a Material (since stair is not a material, but a variety of a block). This also allows for us to make a combination of varieties for a specific object to "describe" it.

Thinking on it more, Variety could even be introduced in API 7 (given some time this week maybe to solidify it as a possibility), in which a "moving" towards it is somewhat available before the drastic change of some of this data removal.

Thoughts @Zidane ?

src/main/java/org/spongepowered/api/effect/sound/SoundTypes.java
```diff
@@ -207,6 +207,8 @@
 
     public static final SoundType BLOCK_PORTAL_TRIGGER = DummyObjectProvider.createFor(SoundType.class, "BLOCK_PORTAL_TRIGGER");
 
+    public static final SoundType BLOCK_PUMPKIN_CARVE = DummyObjectProvider.createFor(SoundType.class, "BLOCK_PUMPKIN_CARVE");
```
gabizou, line 250: Carve or Carved?

src/main/java/org/spongepowered/api/variant/VariantQueryType.java
```diff
@@ -22,13 +22,18 @@
...
-package org.spongepowered.api.data.type;
+package org.spongepowered.api.variant;
 
 import org.spongepowered.api.CatalogType;
-import org.spongepowered.api.util.Cycleable;
-import org.spongepowered.api.util.annotation.CatalogedBy;
 
-@CatalogedBy(LogAxes.class)
-public interface LogAxis extends CatalogType, Cycleable<LogAxis> {
+public interface VariantQueryType<T> extends CatalogType {
```
gabizou: What's the point of this? It's feeling very much either like a Property or a Key in some aspects.

src/main/java/org/spongepowered/api/variant/VariantQueryTypes.java
```diff
@@ -22,24 +22,20 @@
...
-    private LogAxes() {}
+@SuppressWarnings("unchecked")
+public final class VariantQueryTypes {
 
     // SORTFIELDS:ON
 
-    public static final LogAxis NONE = DummyObjectProvider.createFor(LogAxis.class, "NONE");
-
-    public static final LogAxis X = DummyObjectProvider.createFor(LogAxis.class, "X");
+    public static final VariantQueryType<Material> MATERIAL = DummyObjectProvider.createFor(VariantQueryType.class, "MATERIAL");
```
gabizou: With my comment above about Variety vs Material, this is an example where "Variants" will win overall as you can use different varieties to describe something unique (almost mergeable with properties).

src/main/java/org/spongepowered/api/variant/VariantQueryableCatalogType.java
```diff
@@ -0,0 +1,80 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.variant;
+
+import org.spongepowered.api.CatalogType;
+
+import java.util.Optional;
+
+public interface VariantQueryableCatalogType extends CatalogType {
```
gabizou: Description of what this type of catalog type does?

src/main/java/org/spongepowered/api/variant/VariantQueryableCatalogType.java
```diff
@@ -0,0 +1,80 @@
...
+     * Gets the query argument that can be used to match this
+     * {@link CatalogType} with the specified {@link VariantQueryType}.
+     * {@link Optional#empty()} will be returned if the query type
+     * isn't supported by this catalog type.
+     *
+     * @param queryType The query type
+     * @param <T> The argument type
+     * @return The argument type, if the query type is supported
+     */
+    <T> Optional<T> getArgument(VariantQueryType<T> queryType);
```
gabizou: Why is it a query type instead of just `VariantQuery`? We don't call `Key`s `KeyType`.

gabizou (changes requested):
Some good ideas, but I think I've found a direction I'd like to take this towards. 

Overall, I'm going to still have a think about `Variety` and `Varieties` to define what you're calling `Material`. Effectively what it should solve for is defining a Wool Carpet Block as having `Varieties.WOOL`, `Varieties.CARPET`, and `Varieties.BLUE` to describe the block. This way you can determine already that it is a wool block, it's a carpet (so you can gather all blocks that are matching `Varieties.CARPET` and `WOOL`, or simply attempt to match all blocks/items with `WOOL` varieties.

It would be interesting to try and resolve the awkwardness of showing DyeColor as a value instead of a Property of sorts. One way this is retrievable should be through the existing Property API system in place with Data API. Since it is not an immutable data (can't have mutable properties), you can retrieve the whole collection of Varieties in this fashion. 

Would love some input from @SpongePowered/developers .

src/main/java/org/spongepowered/api/variant/VariantQueryableCatalogType.java
Cybermaxke: It was to return the argument you would pass in to create the `VariantQuery` from the `VariantQueryType`.

Cybermaxke:
@gabizou I will open a new PR that targets API 7 for the Variety system. I will try to reuse things from the Data API that are already present to replace the VariantQuery, like using properties to query for specific things.
`TreeTypeProperty`, `DyeColorProperty`, etc.
Which other properties would you think that would be worth adding, based from the data that will be removed, for example:
`GoldenAppleTypeProperty`, `CoalTypeProperty`, etc.

kashike:
Please rebase against `1.13`.

src/main/java/org/spongepowered/api/effect/sound/SoundTypes.java
JBYoshi: Carve. This is the sound that plays when an uncarved pumpkin turns into a carved pumpkin.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -1358,6 +1358,14 @@
      */
     public static final Key<Value<Vector3d>> LEFT_LEG_ROTATION = KeyFactory.fake("LEFT_LEG_ROTATION");
 
+    /**
+     * Represents the {@link Key} for the state that something is "lit",
+     * for example a {@link BlockTypes#FURNACE} or {@link BlockTypes#REDSTONE_TORCH}.
+     *
+     * @see LitData#lit()
+     */
+    public static final Key<Value<Boolean>> LIT = KeyFactory.fake("LIT");
```
ryantheleach: 🚬🔥

gabizou (changes requested):
I’m willing to allow keeping the removed data, only to support retrieving the data, and then removing it later. This way plugins still will have an upgrade understanding. I know that this api is supposed to break the world, but it’d be nice to retain it for support like we did with some data types in api 6.

Cybermaxke:
@gabizou I don't feel like keeping something as half working is a good idea. Maybe it would be better to provide a spongedocs page dedicated to updating from API7 to API8, related to big changes like item types/block types?

parlough:
@Cybermaxke I think I share the same opinion, keeping it with half functionality will unnecessarily clutter the API and only confuse developers. You are right though we need to document all of the changes well, not only updating outdated docs but a migration guide will probably be highly beneficial. In general we should keep a more detailed changelog of larger changes as well.

liach:
Bump: @Cybermaxke Would you add contents from the aquatic update, aside technical changes?

Cybermaxke:
@liach Yes, I am planning on doing that.

gabizou:
Can you update the PR description as to what is going on? (I believe we opted to support legacy data, on the basis that it was read only, much like the option we took with `HorseVariant`s when horses got split up.

gabizou (changes requested):
There's a metric ton of formatting changes that I'm not able to see what is actually added/changed in the API without scrolling and reading every line diff (I feel that the first 10th of the PR changes are just formatting....)

Cybermaxke:
@gabizou I rebased against bleeding but didn't change the target branch, is fixed now.

I still think that keeping half broken API isn't a good thing, and that it would be better to provide good documentation to transition between API 7 and API 8 (once 1.13 is merged).

Would love some input on this from some other developers

parlough:
@Cybermaxke I share your opinion, leaving half functional API would be a pain for maintenance and it will just cause confusion for users. They will have to update and make changes anyway for their plugin to be functional and they can update their removed data retrievals while they're at it. Leaving it will have them expecting API to work that won't work and/or even can't be made to work. 

Our best solution is documenting all the necessary changes quite well, through written documentation, but also through video or stream. I am planning to work on this quite a bit and I know others will be happy to help.

src/main/java/org/spongepowered/api/data/manipulator/immutable/block/ImmutableDecayableData.java
```diff
@@ -26,22 +26,35 @@
...
+    ImmutableBoundedValue<Integer> distance();
+
+    /**
+     * Gets the {@link Value} for whether the {@link BlockState} state is
+     * persistent and will be prevented from decaying.
+     *
+     * @return The immutable value for the persistent state
+     * @see Keys#PERSISTS
+     */
+    ImmutableValue<Boolean> persists();
```
kashike: `persistent`

kashike (changes requested):

src/main/java/org/spongepowered/api/data/manipulator/immutable/block/ImmutableDecayableData.java
Cybermaxke: @kashike I named it `persists` to match the `Key`

kashike: name the key `persistent`, then! :P

Cybermaxke:
Moved to https://github.com/SpongePowered/SpongeAPI/pull/1949

PR #1705, opened 2017-11-27
Change entity spawning return types, breaking change - Fixes #1701

### LINKED ISSUE #1701: EntityUniverse::spawnEntities should return a list of entities spawned.
As per https://github.com/SpongePowered/SpongeCommon/issues/1365

> The return is actually useless regardless of what it is changed to return. It should instead return a list of entities that were actually spawned.

https://github.com/SpongePowered/SpongeAPI/blob/9aeb46721b8bcaa45c96e1b73ef0ddd18c4d78ff/src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java#L268

Needs to be modified to return a list of entities.

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1622)

This is a breaking change and people likely spawn entities quite a bit, so I would like discussion on whether or not we should make this change during API 7 or API 8 and whether all of this is even necessary. 

I am also looking for heavy review on the implementation changes in SpongeCommon.

Closes https://github.com/SpongePowered/SpongeAPI/issues/1701

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Location.java
+++ b/src/main/java/org/spongepowered/api/world/Location.java
@@ -674,16 +674,16 @@ public final class Location<E extends Extent> implements DataHolder {
     }
 
     /**
-     * Similar to {@link #spawnEntity(Entity, Cause)} except where multiple
+     * Similar to {@link #spawnEntity(Entity)} except where multiple
      * entities can be attempted to be spawned with a customary {@link Cause}.
      * The recommended use is to easily process the entity spawns without
      * interference with the cause tracking system.
      *
-     * @param entities The entities to be spawned
+     * @param entities The entities which spawned correctly, or empty if none
      * @return True if any of the entities were successfully spawned
-     * @see EntityUniverse#spawnEntities(Iterable, Cause)
+     * @see EntityUniverse#spawnEntities(Iterable)
      */
-    public boolean spawnEntities(Iterable<? extends Entity> entities) {
+    public Collection<Entity> spawnEntities(Iterable<? extends Entity> entities) {
         return this.getExtent().spawnEntities(entities);
     }
 
--- a/src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java
+++ b/src/main/java/org/spongepowered/api/world/extent/EntityUniverse.java
@@ -262,10 +262,10 @@ public interface EntityUniverse {
      * Similar to {@link #spawnEntity(Entity)} except where multiple entities
      * can be attempted to be spawned.
      *
-     * @param entities The entities to be spawned
+     * @param entities The entities which spawned correctly, or empty if none
      * @return True if any of the entities were successfully spawned
      */
-    boolean spawnEntities(Iterable<? extends Entity> entities);
+    Collection<Entity> spawnEntities(Iterable<? extends Entity> entities);
 
     /**
      * Gets all the entities that intersect the bounding box, in no particular


src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -666,24 +666,24 @@ public Entity createEntity(EntityType type) {
      * should be taken note that there can be many reasons for a failure.</p>
      *
      * @param entity The entity to spawn
-     * @return True if successful, false if not
+     * @return The entity, if it was spawned successfully
      * @see EntityUniverse#spawnEntity(Entity)
      */
-    public boolean spawnEntity(Entity entity) {
+    public Optional<Entity> spawnEntity(Entity entity) {
```
Zidane: I'm not sure I can allow a break for sugaring in this case.

src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -666,24 +666,24 @@ public Entity createEntity(EntityType type) {
...
      * interference with the cause tracking system.
      *
      * @param entities The entities to be spawned
-     * @return True if any of the entities were successfully spawned
-     * @see EntityUniverse#spawnEntities(Iterable, Cause)
+     * @return The entities which spawned correctly, or empty if none
+     * @see EntityUniverse#spawnEntities(Iterable)
      */
-    public boolean spawnEntities(Iterable<? extends Entity> entities) {
+    public Collection<Entity> spawnEntities(Iterable<? extends Entity> entities) {
```
Zidane: Now this one, I can see the reasoning behind. Else you can't know who spawned without a way to check each Entity (which makes things somewhat annoying).

parlough: This one is most likely less commonly used as well, so it hurts less to break. It is also what the original issue discussed.

src/main/java/org/spongepowered/api/world/Location.java
parlough: Fair enough. I will let it sizzle for a bit, but this one I am cool with reverting.

simon816: Would the returned entity ever _not_ be the entity given in the parameter? If not, I don't see a reason for this

parlough: It shouldn't be any other entity :P. This was kind of just for convenience, but yeah not really worth a breaking change. Can revert it later if the consensus continues to be so.

`world.spawnEntity(location.createEntity(EntityTypes.BOAT)).ifPresent(e -> e.addPassenger(player))` or something.

ryantheleach: It was more for consistency with the rest of the API / spawnEntities being a collection.

liach:
Should we return some sort of action results (true/false, tristate, transactionresult) to indicate for each entity better? 

> From https://github.com/SpongePowered/SpongeAPI/issues/1701#issuecomment-347379496

src/main/java/org/spongepowered/api/world/Location.java
parlough: This change would break a lot more plugins and overall it doesn't do much for the developer, so I will revert it.

parlough:
@liach I am not sure that is worth it as it simply complicates the result beyond a majority of use cases.

kashike (approved):

gabizou:
I'll get to merging this after #1554 is merged.

PR #1706, opened 2017-11-28
[API8] Add datapack support

mattmess1221:
I'm assuming the client's resource manager was reused for the server. Likely more edits will come when I have access to deobfuscated code.


liach:
should we merge resource path with catalog key?

mattmess1221:
If we do, we should settle on a better suited name. It doesn't make sense to get a `Resource` from a `CatalogKey`, and neither does a `CatalogType` with `ResourcePath`

Though they are slightly different. While both have a namespace and value, the resource path can contain a forward slash `/` and dot `.`. If anything, it should have a common parent-interface indicating it can have a namespace.

liach:
Something like `BinomialName` for a parent interface?

mattmess1221:
I was thinking it would include the word namespace. Something like `NamespacedIdentifier`

liach:
`SpacedName`

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -35,6 +35,7 @@
 import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.EventManager;
 import org.spongepowered.api.network.ChannelRegistrar;
+import org.spongepowered.api.resources.ResourceManager;
```
kashike: Singular package names - `resource`.

kashike (changes requested):

ryantheleach:
I don't like that this outright deprecates the Asset API.

The goals of the Asset API were to simplify the process for new programmers to extract resources from their jar files, to their plugin directories on disk, usually used for extracting default or example configurations.

If this is to happen, I'd like at least a little attention paid to giving those people a clear migration path, and maybe some docs to go along with it to explain exactly what this is and how it works.
  

I also think the name DataPack is pretty close to the Mojang defined term.

If this isn't intended to be directly compatible with that, or interact with that in some way then it's poorly named.

But maybe I'm just not following why the name change is needed?

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    ResourcePath getResourcePath();
+
+    /**
+     * Gets the {@link Pack} which owns this resource. The pack is set
+     * automatically when the resource is added to a pack.
+     *
+     * @return The parent pack.
+     */
+    Pack getPack();
```
ryantheleach: Do Resources always have Packs?

liach: I think there may be a default pack for minecraft

src/main/java/org/spongepowered/api/Sponge.java
```diff
@@ -152,10 +154,15 @@ public static EventManager getEventManager() {
      *
      * @return The asset manager instance
      */
+    @Deprecated
     public static AssetManager getAssetManager() {
         return check(assetManager);
     }
 
+    public static ResourceManager getResourceManager() {
```
liach: jdoc

src/main/java/org/spongepowered/api/resource/InputStreamSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@FunctionalInterface
+public interface InputStreamSupplier {
+
+    /**
+     * Returns a new {@link InputStream}. Each call of this method should
+     * create a new input stream.
+     *
+     * @return A new input stream
+     * @throws IOException if an error occurs
```
liach: Is there a case regulation for `@throws` tags?

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Reads a list of string lines from this resource using the given charset.
+     *
+     * @param charset The charset
+     * @return The list of strings
+     * @throws IOException if an error occurs
+     */
+    default List<String> readLines(Charset charset) throws IOException {
```
liach: Can we get a java stream here? We can use collector to collect it and do more complex operations

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,132 @@
...
+    interface Builder extends ResettableBuilder<ResourcePath, Builder> {
+
+        /**
+         * Sets the namespace of the {@link ResourcePath}.
+         *
+         * @param namespace The namespace
+         * @return This builder
+         * @throws IllegalArgumentException if the namespace contains illegal characters
+         */
+        Builder namespace(String namespace) throws IllegalArgumentException;
```
liach, line 208: Can we use a plugin object to supply the namespace, too?

src/main/java/org/spongepowered/api/resource/Pack.java
```diff
@@ -0,0 +1,136 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the name of this pack.
+     *
+     * @return The name
+     */
+    Text getName();
```
liach: Do we only have text names? String names are easier to identify compared to text ones (remember that we get a list of active packs from the resource provider and plugins may want to disable some of them)

src/main/java/org/spongepowered/api/resource/Resource.java
mattmess1221: If used properly, there are no orphans in the resource manager. The only way for this to return null is for the developer to use `Resource.builder()` and not add it to a pack.

I'm actually working on something that makes this a bit more foolproof. Standby.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
mattmess1221: Yes, I just added a method for that.

src/main/java/org/spongepowered/api/resource/Pack.java
```diff
@@ -0,0 +1,132 @@
...
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the name of this pack.
+     *
+     * @return The name
+     */
+    String getName();
```
Cybermaxke: This is a `String` but the `Builder` uses a `Text` object?
Should this name also be used in the `registerPack` method? If so, the name argument in the method is unnecessary.

Cybermaxke:
Please keep the `ResourcePath` separate from the `CatalogId`. They represent two different things and shouldn't be merged as one. They also have different contraints.

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Registers a pack to be usable.
+     *
+     * TODO move to {@link org.spongepowered.api.GameRegistry}?
+     *
+     * @param name
+     * @param pack
+     */
+    void registerPack(String name, Pack pack);
```
Cybermaxke: I think that it's fine to keep this method in here.

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Unregisters a pack by its name.
+     *
+     * @param name The name of the pack
+     * @return The pack which was unregistered or empty if it wasn't registered
+     */
+    Optional<Pack> unregisterPack(String name);
+
```
Cybermaxke, line 75: Maybe also `prioritizeAfter`, `prioritizeBefore`, `prioritizeFirst` and `prioritizeLast` methods?

liach: Can't we just arrange that list?

mattmess1221: The way I was thinking is the pack order would be determined by calling `reload(List<Pack>)`.  Though I also have an idea about using Comparable to sort the packs.

Cybermaxke: What happens if someone passes in a `Pack` that isn't registered?

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -88,6 +88,9 @@ static ResourcePath parse(String path) throws IllegalArgumentException {
      */
     String getPath();
 
+    @Override
+    int hashCode();
```
Cybermaxke, line 194: Why is this in the API?

liach: For arbitrary order?

src/main/java/org/spongepowered/api/resource/ResourceManager.java
Cybermaxke: What happens if you try to unregister a plugin pack?

src/main/java/org/spongepowered/api/resource/ResourcePath.java
mattmess1221: I put it there originally because I wanted it to be clear that this can be used in a Map. Now I'm just saying so in the docs.

src/main/java/org/spongepowered/api/resource/ResourceProvider.java
```diff
@@ -38,4 +39,18 @@
...
+    Collection<Resource> getResources();
+
+    /**
+     * Walks a resource tree starting at the given path.
+     *
+     * @param path The path to start at
+     * @return A stream of resource paths
+     */
+    Stream<ResourcePath> walk(ResourcePath path);
+
```
Cybermaxke: I wouldn't go with the `walk` name, it can be confusing when using in combination with `ResourceManager`. You can also return the `Resource` directly when walking or just add variants for `Resource`s and `ResourcePath`s. A depth could also be useful to limit the directory structure complexity:
`Stream<ResourcePath> listResourcePaths(ResourcePath resourcePath);`
`Stream<ResourcePath> walkResourcePaths(ResourcePath resourcePath, int maxDepth);`
`Stream<ResourcePath> walkResourcePaths(ResourcePath resourcePath);`
`Stream<Resource> walkResources(ResourcePath resourcePath, int maxDepth);`
`Stream<Resource> walkResources(ResourcePath resourcePath);`

src/main/java/org/spongepowered/api/resource/PackProvider.java
```diff
@@ -1,16 +1,75 @@
...
+ */
 package org.spongepowered.api.resource;
 
+import org.spongepowered.api.plugin.PluginContainer;
+
 import java.util.Map;
+import java.util.Optional;
 
-@FunctionalInterface
 public interface PackProvider {
```
Cybermaxke: Any reason why these methods are moved outside of the `ResourceManager`?

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -92,6 +93,54 @@ static ResourcePath parse(String path) throws IllegalArgumentException {
...
+     * List<ResourcePath> parents = getParents();
+     * if (parents.isEmpty()) {
+     *     return this;
+     * }
+     * return parents.get(parents.size() - 1);
+     * </pre>
+     *
+     * @return The root path
+     */
+    ResourcePath getRoot();
```
Cybermaxke: Root would return `namespace:`, path is empty, if the docs are true this would return the parent path. Add a separate `getParent` method for this.

src/main/java/org/spongepowered/api/resource/ResourceProvider.java
mattmess1221: What would be the difference between `listResourcePaths` and `walkResourcePaths`?

src/main/java/org/spongepowered/api/resource/ResourceId.java
```diff
@@ -24,13 +24,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
 /**
  * Annotation for injecting resources into a plugin instance.
  *
- * The type this should be used on is {@code Optional<Resource>}.
+ * The type this should be used on is {@code Optional<Resource>}. When applied
+ * to a method, the {@link Resource} will be
```
liach: will be what?

src/main/java/org/spongepowered/api/event/resource/ResourceReloadEvent.java
```diff
@@ -0,0 +1,45 @@
...
+public interface ResourceReloadEvent extends ResourceEvent {
+
+    /**
+     * Called before the {@link ResourceManager} is reloaded. At this point,
+     * the {@link ResourceManager#getActivePacks() active packs} can be added
+     * or removed from.
+     */
+    interface Pre extends ResourceEvent {
+
+    }
```
Cybermaxke: No `Post` event?

mattmess1221: I don't think having `onReload(ResourceReloadEvent.Post)` is good api. If someone were to do `ResourceReloadEvent` instead, it would receive both events, which is probably not what they meant.

Cybermaxke: The `Pre` event is called before the resources are reloaded, how do you it's finished reloading?

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Gets a reader for this resource using the given {@link Charset}.
+     *
+     * @param charset The charset
+     * @return A new reader
+     * @throws IOException if an error occurs
+     */
+    default Reader getReader(Charset charset) throws IOException {
+        return new InputStreamReader(openStream(), charset);
```
stephan-gh: I'd return a `BufferedReader` here. You'll likely almost always want a buffered reader, and if you don't you'll know to create it from `openStream()` manually.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Gets a list of parents for this {@link ResourcePath}. Order of parents
+     * will be closest first, with the root being the last element. If the
+     * returned list is empty, this path is the root path.
+     *
+     * @return The list of parents
+     */
+    List<ResourcePath> getParents();
```
Cybermaxke: Also a regular `getParent`?

src/main/java/org/spongepowered/api/resource/ResourceTypes.java
```diff
@@ -0,0 +1,15 @@
+package org.spongepowered.api.resource;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+// TODO I don't like this
+public class ResourceTypes {
```
Cybermaxke: I wouldn't add this in the API, this seems more like a implementation detail to me. The only reason to still use the `assets` package is because it's used by the Asset API, the assets package in vanilla will only be used for textures, etc. Maybe just use the `assets` package for the Asset API and the `data` package for the resource api?

src/main/java/org/spongepowered/api/resource/ResourceType.java
```diff
@@ -0,0 +1,8 @@
+package org.spongepowered.api.resource;
```
Cybermaxke: Missing header

src/main/java/org/spongepowered/api/resource/ResourceTypes.java
```diff
@@ -0,0 +1,15 @@
+package org.spongepowered.api.resource;
```
Cybermaxke: Missing header

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -94,36 +95,38 @@ static ResourcePath parse(String path) throws IllegalArgumentException {
...
+     * Gets a list of paths for this {@link ResourcePath}. Order of parents
+     * will be closest first, with the first element being this element and the
+     * root being the last element. If the returned list is empty, this path is
+     * the root path.
      *
-     * @return The list of names in the path.
+     * @return The list of paths
      */
-    List<String> getNames();
+    List<ResourcePath> getParts();
```
Cybermaxke: IMO the `getNames` and `getParents` methods were good, but `getParts` doesn't seem right. A resource path cannot be split into resource path parts, because it also contains the namespace. If it would split into stings that would make sense, but then it would just be the `getNames` method.

Javadocs still says it's the parents?

Cybermaxke: I already noticed that you fixed the javadoc.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,254 @@
...
+    boolean isParent(ResourcePath parent);
+
+    /**
+     * Gets the file name portion of this {@link ResourcePath}.
+     *
+     * @return The file name of the path
+     * @see #getBaseName()
+     * @see #getExtension()
+     */
+    String getFileName();
```
Cybermaxke, line 122: I wouldn't call this `getFileName` since there is no guarantee that it's actually a file, it can also be a directory. This also breaks the `getBaseName` method.

mattmess1221: I disagree. Yes, there is no guarantee that it is a file, but that's not up to the path to decide. The path has no say in whether something exists, or is a file (or directory).

And `getBaseName` was named after `FilenameUtils.getBaseName`, which I added because commons-io is not included in the API.

src/main/java/org/spongepowered/api/event/resource/ResourceEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.resource;
+
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.resource.ResourceManager;
+
+public interface ResourceEvent extends Event {
```
Lignium, line 34: Javadoc?

liach:
Can we move datapacks to each world, given that is how they are handled on both client and servers? Current structure won't work for clients.

mattmess1221:
@liach Nope. Resources are not per-world. There are only 2 instances of `SimpleReloadableResourceManager` in 1.13. They are located in `Minecraft.resourceManager` and `MinecraftServer.resourceManager`

Cybermaxke:
Resources are basically per save file, on the server there is only one. On the client its the one when you load a world (including all dimensions).

liach:
@killjoy1221 However, data packs are not loaded as resources; they are never included in the resource manager. They are simply associated with worlds.

mattmess1221:
I haven't looked too closely on how datapacks work yet, but I'm pretty sure the world will be added as a data pack, which can also be removed. I'm going to have to study the code to figure how how everything works with `PackFinder` and `ResourcePackList`

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -45,7 +49,7 @@
      * @param path The path to the resource
      * @return The resource
      */
-    Optional<Resource> getResource(ResourcePath path);
+    Resource getResource(ResourcePath path) throws IOException;
```
Cybermaxke: Maybe rename this to `openResource` and add a method `hasResource` to check whether a resource exists with a given path.

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,79 @@
...
+     * <pre>
+     * manager.getResources("loot_tables", name -> name.endsWith(".json"));
+     * </pre>
+     *
+     * @param path The path to search in
+     * @param filter The file name filter
+     * @return A collection of resource paths
+     */
+    Collection<ResourcePath> getResources(String path, Predicate<String> filter);
+
```
Cybermaxke, line 93: Maybe it's still worth adding some of the methods from my lantern datapacks pr?
https://github.com/LanternPowered/LanternServer/pull/53/files#diff-d8831f2ec678210c3618a5b72bbd5a47R47

src/main/java/org/spongepowered/api/event/resource/ResourceReloadEvent.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.api.resource.Resource;
+
+/**
+ * Base interface for resource reloading events.
+ */
+public interface ResourceReloadEvent extends ResourceEvent, Cancellable {
+
+    /**
+     * Called before the {@link ReloadableResourceManager} is reloaded. At this
+     * point, the {@link PackRepository#getEnabledPacks() enables packs} can be
```
liach: `enabled packs`

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -85,7 +85,7 @@ static ItemStack of(ItemType itemType, int quantity) {
      * @return The empty ItemStack
      */
     static ItemStack empty() {
-        return builder().itemType(ItemTypes.NONE).build();
+        return builder().itemType(ItemTypes.AIR).build();
```
liach: Should put this in a separate pr to prevent merge conflicts

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
```diff
@@ -131,15 +133,26 @@ default String getName() {
...
     }
 
+    /**
+     * Retrieves the {@link PackInfo} owned by this plugin from the
+     * {@link PackRepository}.
+     *
+     * @return The plugin's resource pack.
+     */
+    default PackInfo getPack() {
+        return Sponge.getServer().getPackRepository().getPack(this);
```
liach: As far as I know, resource packs and data packs have different information, even as they can share the same pack mcmeta file. This prevents a plugin from having both packs, which is the case for forge mods, etc.

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
Cybermaxke: The PR should be rebased on top of the latest changes.

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
mattmess1221: This will probably have to wait for a full impl until the client api gets fleshed out a bit more. As it is now, `ResourceTypes.ASSETS` is restricted to the client, and `DATA` to the server.

src/main/java/org/spongepowered/api/resource/ResourceTypes.java
Cybermaxke: Since we are also going to support the client, `ResourceType` can stay

liach:
@killjoy1221 no, data packs can be in integrated servers

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
liach: I suggest removing this method. Plugin Containers may exist without the server. This method itself is default, meaning that the removal will not reduce functionality.

liach: Instead of removing this method, we can leave it abstract so that implementations are correct.

src/main/java/org/spongepowered/api/resource/PackRepository.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return The disabled packs
+     */
+    Collection<PackInfo> getDisabledPacks();
+
+    /**
+     * Returns a collection of enabled packs.
+     *
+     * @return The enabled packs
+     */
+    Collection<PackInfo> getEnabledPacks();
```
liach: Mention that the iteration order or the collection is the packs' priority,

src/main/java/org/spongepowered/api/resource/PackRepository.java
```diff
@@ -0,0 +1,85 @@
...
+public interface PackRepository {
+
+    /**
+     * Sets the active packs to be used when reloading resources. Calls to this
+     * should be followed by {@link ReloadableResourceManager#reload()}.
+     *
+     * @param packs The active packs
+     * @see ReloadableResourceManager#reload()
+     */
+    void setActivePacks(Collection<PackInfo> packs);
```
liach: Mention that the iteration order of the `packs` is the priority of active packs.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+ * filesystem, a network location, or even generated at runtime. Use
+ * {@link #getInputStream()} to load the data held by a resource.
+ *
+ * <p>Resource extends Closeable, so don't forget to close your streams either
+ * with a try-with-resources or a close() inside a try-finally block. e.g.</p>
+ * <pre>
+ *     try (Resource res = resourceManager.getResource(path)) {
+ *         logger.info(res.readString(DefaultCharsets.UTF_8));
+ *     } catch (IOException e) {
+ *         e.printStackTrace();
```
liach: We should recommend users to log the error or throw it instead of `printStackTrace`

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * <p>A resource path should contain a namespace. If one is not provided,
+ * {@code minecraft} will be used instead. The namespace and path must not
+ * contain any special characters or uppercase letters.
+ *
+ * <p>In the pack, the path will point to a resource. The resource should
+ * be located roughly at {@code data/namespace/path}</p>
+ *
+ * <p>A resource path should be usable in a {@link Map}, so implementations
+ * should override {@link #hashCode()} and {@link #equals(Object)}.</p>
```
liach, line 45: Put this implnote in the method javadocs as well.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,245 @@
...
+ */
+public interface ResourcePath extends Comparable<ResourcePath>, Iterable<ResourcePath> {
```
liach: The iterable, in my opinion, is very rarely used and can cause mod conflicts like SpongePowered/SpongeForge#2540

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * <p>A resource path should contain a namespace. If one is not provided,
+ * {@code minecraft} will be used instead. The namespace and path must not
+ * contain any special characters or uppercase letters.
+ *
+ * <p>In the pack, the path will point to a resource. The resource should
+ * be located roughly at {@code data/namespace/path}</p>
```
liach: Do you need to note a transformation of resource path between user and actual usage? Users supply advancement id `minecraft:story/root` that will be translated to `minecraft:advancements/story/root.json`, which locates `data/minecraft/advancements/story/root.json`.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Creates a new {@link ResourcePath} using the given namespace and path.
+     *
+     * @param namespace The namespace to use
+     * @param path The path to use
+     * @return A new ResourcePath
+     */
+    static ResourcePath of(String namespace, String path) throws IllegalArgumentException {
```
liach, line 65: How about other `minecraft(String)` or `plugin(PluginContainer/Object, String)` methods?

src/main/java/org/spongepowered/api/resource/ResourcePath.java
liach: This method does not appear meaningful to me. Why?

liach:
@killjoy1221 Can you target 1.14 instead?

src/main/java/org/spongepowered/api/plugin/PluginContainer.java
liach: We should consider moving this to the client and server interfaces instead.

src/main/java/org/spongepowered/api/resource/ResourceTypes.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public class ResourceTypes {
+
+    // SORTFIELDS:ON
+
+    public static final ResourceType ASSETS = DummyObjectProvider.createFor(ResourceType.class, "ASSETS");
+
+    public static final ResourceType DATA = DummyObjectProvider.createFor(ResourceType.class, "DATA");
```
liach: It's not supposed to be like this... this is pack type. resource types are like `advancements` `loot_tables` `functions` `tags/functions` `tags/fluids` `models` `textures` `blockstates` etc...

src/main/java/org/spongepowered/api/resource/ResourcePath.java
mattmess1221: I don't feel those methods are necessary. You can just as easily stick in `"minecraft`" or `"foobar"`. I prefer it to be explicit for this.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
mattmess1221: This is specific to resources. A better class for that one would be `NamespacedId`. There's a conversation about this.

src/main/java/org/spongepowered/api/resource/AsyncReloader.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+
+public interface AsyncReloader {
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    ResourceManager getResourceManager();
```
liach: I don't think async reloaders are manager aware, are they?

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -48,17 +52,26 @@
...
+     * Returns all of the resources which exist in each {@link Pack}. Remember
+     * to close all of the resources.
      *
      * @return All of the resources
+     * @see #getResource(ResourcePath)
+     * @throws IOException If there was an error reading any of the resources
+     * @throws FileNotFoundException If no resources exist
      */
-    List<Resource> getResources(ResourcePath path);
+    List<Resource> getResources(ResourcePath path) throws IOException, FileNotFoundException;
```
liach, line 74: Remember to document the order of the list is based on priority. Higher priority entries are placed earlier in this list (I think this is the case; I may be wrong).

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
```diff
@@ -86,4 +80,22 @@
      */
     boolean isRemote();
 
+    boolean isLocked();
+
+    Priority getPriority();
+
+    Compatibility getCompatability();
+
+    enum Priority {
```
liach: `InsertionPosition`

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
```diff
@@ -86,4 +80,22 @@
...
+        FIRST,
+        LAST
+    }
+
+    enum Compatibility {
+        OLD,
+        NEW,
+        OK
```
liach: Consider using some other thing like a string version, like what we do for `GameVersion`

src/main/java/org/spongepowered/api/resource/AsyncReloader.java
mattmess1221: They are.

src/main/java/org/spongepowered/api/Engine.java
```diff
@@ -25,12 +25,16 @@
...
+import org.spongepowered.api.resource.pack.PackRepository;
 import org.spongepowered.api.scheduler.Scheduler;
 
+import java.util.concurrent.Executor;
+
 /**
  * Shared functionality between {@link Client} and {@link Server} engines.
  */
-public interface Engine {
+public interface Engine extends Executor {
```
Cybermaxke: You can't make engine a executor, all tasks need to be scheduled through the scheduler to know which plugin submitted them. You can however get a exectutor from the scheduler for a specific plugin.

mattmess1221: Minecraft and MinecraftServer are already executors. Adding a separate method would essentially be returning this. But I guess that's an implementation detail.

Cybermaxke: We need to track which plugin posts a task, vanilla mc does not. So we can't just expose this in the API.

Cybermaxke:
@killjoy1221 Can you rebase on top of `api-8`?

src/main/java/org/spongepowered/api/resource/ResourceManager.java
mattmess1221: Changing this into a `Stream` may be useful.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,246 @@
...
+        Builder namespace(String namespace) throws IllegalArgumentException;
+
+        /**
+         * Sets the namespace of the {@link ResourcePath} using the given
+         * plugin's id.
+         *
+         * @param plugin The owning plugin
+         * @return This builder
+         */
+        Builder namespace(PluginContainer plugin);
```
liach, line 217: Accept object as well for plugin instance?

mattmess1221: I was keeping consistent with `CatalogKey.Builder`

src/main/java/org/spongepowered/api/resource/pack/PackRepository.java
```diff
@@ -0,0 +1,85 @@
...
+    Collection<PackInfo> getAllPacks();
+
+    Collection<PackInfo> getAvailablePacks();
+
+    /**
+     * Returns a collection of enabled packs.
+     *
+     * @return The enabled packs
+     */
+    Collection<PackInfo> getEnabledPacks();
```
liach: Again, make it a list to show that enabled packs are ordered

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
```diff
@@ -0,0 +1,83 @@
...
+package org.spongepowered.api.resource.pack;
+
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.TextRepresentable;
+
+/**
+ * Basic info about a {@link Pack}
+ */
+public interface PackInfo extends DataSerializable, TextRepresentable {
```
liach: Why `DataSerializable`?

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    String getName();
+
+    /**
+     * Gets the display name of the {@link Pack} which is displayed to the
+     * user. This is loaded from the pack.mcmeta
+     *
+     * @return The display name
+     */
+    Text getDisplayName();
```
liach: More exactly, this is just a name pack repo gives and with the desc in the hover

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
liach: Also, return the pack type as well.

src/main/java/org/spongepowered/api/resource/pack/PackRepository.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Sets the active packs to be used when reloading resources. Calls to this
+     * should be followed by {@link Engine#reloadResources()} in order to take
+     * effect.
+     *
+     * @param packs The active packs
+     * @see Engine#reloadResources()
+     */
+    void setEnabledPacks(Collection<PackInfo> packs);
```
liach: `Iterable` or `List` is better; the enabled packs are ordered, which plain iterable and list can convey.

liach (changes requested):

src/main/java/org/spongepowered/api/resource/pack/Pack.java
```diff
@@ -72,14 +72,21 @@
...
+     *
+     * @return The set of namespaces
+     */
+    Set<String> getNamespaces();
+
     /**
      * Gets the metadata of this pack. The {@link DataView} represented is of
      * the pack.json file in the pack root. If the pack does not contain a
-     * pack.json, {@link Optional#empty()} is returned.
+     * pack.json, it is an error.
```
liach: wtf, it is `pack.mcmeta`, and if that one is absent, a default one will be returned

src/main/java/org/spongepowered/api/event/resource/ResourceEvent.java
```diff
@@ -22,14 +22,34 @@
...
+     * Gets the {@link ResourceManager}.
+     * @return The resource manager
+     */
+    ResourceManager getResourceManager();
 
+    /**
+     * Fired after a {@link ResourceManager} has been reloaded. This can be
+     * used to reload some systems that are resource based.
+     */
+    interface Reload extends ResourceEvent {
```
liach, line 52: Please allow plugins to work async! This will increase load on the server and cannot be profiled by vanilla loading bar, etc.

src/main/java/org/spongepowered/api/resource/pack/Pack.java
mattmess1221: And what should a default mcmeta contain?

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
mattmess1221: Should I use `String getId()` and `Text getName()`?

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
mattmess1221: It has no concept of `PackType`. I really don't remember why I made it data serializable.

src/main/java/org/spongepowered/api/resource/pack/Pack.java
liach: It just contains an assumed pack version and a broken message in its description

src/main/java/org/spongepowered/api/resource/pack/PackInfo.java
liach: I suggest you add a `getName` method additionally that returns the name used to refer to this pack info for commands.

src/main/java/org/spongepowered/api/resource/pack/PackTypes.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.spongepowered.api.resource.pack;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public class PackTypes {
+
+    // SORTFIELDS:ON
+
+    public static final PackType ASSETS = DummyObjectProvider.createFor(PackType.class, "ASSETS");
```
liach: Nah, this is actually resource.
image

mattmess1221: Yeah, but I don't really like that name.

liach: `RESOURCE` is already an established name. (I don't ask you to name this one with the client prefix) This name is not mapped, but rather inferred from the enum constructor parameters passed.

Well, minecraft wiki documents resource and data packs, not assets and data packs. Beware of that.

src/main/java/org/spongepowered/api/resource/ResourcePath.java
Zidane: We don't do Object plugin parameters anymore.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    String getPackName();
+
+    /**
+     * Gets a reader for this resource using the given {@link Charset}.
+     *
+     * @param charset The charset
+     * @return A new reader
+     */
+    default BufferedReader getReader(Charset charset) {
```
liach, line 105: `createReader`/`makeReader`/`newReader` as `get` implies this is an intrinsic property.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Reads the bytes from this resource and returns them in an array.
+     *
+     * @return The bytes
+     * @throws IOException if an error occurs
+     */
+    default byte[] readBytes() throws IOException {
+        return ByteStreams.toByteArray(getInputStream());
```
liach, line 137: Resource leak here.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+public interface Resource extends Closeable {
+
+    /**
+     * Returns a new {@link InputStream} of this resource. A new input stream
+     * should be created each time this method is called.
+     *
+     * @return A new input stream
+     */
+    InputStream getInputStream();
```
liach, line 68: `createInputStream`/`newInputStream`/`makeInputStream`

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+     * @see <a href=http://minecraft.gamepedia.com/Resource_pack#Contents> Minecraft Wiki/Resource Packs
+     */
+    Optional<DataView> getMetadata();
+
+    /**
+     * Gets the path of this resource.
+     *
+     * @return The path
+     */
+    ResourcePath getPath();
```
liach, line 90: In my opinion, `getName` sounds better. See how dinnerbone refers to `<namespace>:<name>` in https://minecraft.net/en-us/article/minecraft-snapshot-17w43a

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Reads this resource into a string using the given charset.
+     *
+     * @param charset The charset
+     * @return The string contents
+     * @throws IOException if an error occurs
+     */
+    default String readString(Charset charset) throws IOException {
+        return CharStreams.toString(getReader(charset));
```
liach, line 117: Resource leak

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Reads a list of string lines from this resource using the given charset.
```
liach, line 121: Note that the stream needs to be closed.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Reads this resource into a {@link DataView}.
+     *
+     * @param format The data format to use
+     * @return The dataview
+     * @throws IOException if an error occurs
+     */
+    default DataView readDataView(DataFormat format) throws IOException {
+        try {
+            return format.readFrom(getInputStream());
```
liach, line 149: Possible resource leak; unfortunately sponge api did not document if the stream will be closed after usage.

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Copies this resource to a given {@link Path}.
+     *
+     * @param path The file to write to
+     * @param options The options to use
+     * @throws IOException if an error occurs
+     */
+    default void copyTo(Path path, OpenOption... options) throws IOException {
+        MoreFiles.asByteSink(path, options).writeFrom(getInputStream());
```
liach, line 163: Possible leak?

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Copies this resource to a given {@link OutputStream}.
+     *
+     * @param out The output stream to write
+     * @throws IOException if an error occurs
+     */
+    default void copyTo(@WillNotClose OutputStream out) throws IOException {
+        ByteStreams.copy(getInputStream(), out);
```
liach, line 173: Need to close the input stream and note that the output stream is not flushed.

src/main/java/org/spongepowered/api/resource/pack/Pack.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.InputStream;
+import java.util.Collection;
+import java.util.Set;
+import java.util.function.Predicate;
+
+/**
+ * A pack can contain several {@link Resource Resources}. Each pack is
+ * independently loaded with a configured priority.
+ */
+public interface Pack extends Closeable {
```
liach, line 43: Mention that a pack can be used both as a data and resource pack, like that pack from the vanilla client.

liach (changes requested):
Must fix all those resource leaks before this merge.

liach (approved):

mattmess1221:
InputStream leakage is covered by `Resource#close()` and should be handled by the caller.

Zidane:
@killjoy1221 Need this finished up soon else it won't make it for API 8.0.0

mattmess1221:
#2135

PR #1708, opened 2017-11-30
Add fastutil as an API dependency - Closes #1590

### LINKED ISSUE #1590: Include fastutil as API dependency?
Since Minecraft is including the full, original fastutil(version 7.1.0) library now, it would be nice to have it in the API as a dependency like Guava so we can have a guarantee as a plugin developer that all Sponge implementations are including it, so we can use it safely ourselves.

parlough:
Adds fastutil as a dependency, since it is a library that Minecraft itself includes the full version of, we use it in the implementation, and many plugin developers will probably find use in it as well. Sure they could use it expecting Minecraft to have it, but that causes issues as there is then no guarantee a Sponge implementation includes a complete, compatible version of the library, or even one at all.

Closes https://github.com/SpongePowered/SpongeAPI/issues/1590

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     compile 'com.google.errorprone:error_prone_annotations:2.0.15'
     compile 'com.google.code.gson:gson:2.8.0'
     compile 'org.apache.commons:commons-lang3:3.5'
+    compile 'it.unimi.dsi:fastutil:7.1.0'
     // Only included in server
     compile 'com.google.code.findbugs:jsr305:3.0.1'
 


liach:
is there anything in the api that can be optimized by this for now? boxing type collections, etc...

stephan-gh (approved):

gabizou:
I'm not keen on having it as a requirement since it would require other implementations to now include a 17 Mb library that otherwise would not have been required to have, thereby increasing the size of downloads often times by at least double (in comparison with sponge forge's jars).

Zidane:
I am siding with @gabizou on this one. Including libraries isn't so cut and dry; the stuff we include currently are mostly "common sense" ones in the API (Apache Commons/Guava could be debated but they remain extremely useful). Fastutil does not fall under this "Its just useful for about anyone" as it is too specific in it being for faster collections. I very well can't see ALL implementations of Sponge outright using it just-because.

PR #1709, opened 2017-12-01
Add chorus fruit teleport type - Fixes #1556

### LINKED ISSUE #1556: [SpongeAPI 7.0] Add TeleportTypes for Chorus Fruit
Hi,
when player eat chorus fruit, the TeleportCause return TeleportTypes.UNKNOW
Can you add a new TeleportTypes for chorus fruit ?

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1629) 

Essentially adds a chorus fruit teleport type, to see if a teleport was caused by a player eating a chorus fruit.

Tested with: 

```Java
@Listener
public void onTeleport(MoveEntityEvent.Teleport event, @Root Player player) {
    event.getContext().get(EventContextKeys.TELEPORT_TYPE).ifPresent(t -> player.sendMessage(Text.of(TextColors.GOLD, "TeleportType: ", TextColors.GRAY, t.getId())));
}
```

### DIFF
--- a/src/main/java/org/spongepowered/api/event/cause/entity/teleport/TeleportTypes.java
+++ b/src/main/java/org/spongepowered/api/event/cause/entity/teleport/TeleportTypes.java
@@ -32,6 +32,8 @@ public final class TeleportTypes {
 
     public static final TeleportType COMMAND = DummyObjectProvider.createFor(TeleportType.class, "COMMAND");
 
+    public static final TeleportType CHORUS_FRUIT = DummyObjectProvider.createFor(TeleportType.class, "CHORUS_FRUIT");
+
     public static final TeleportType ENTITY_TELEPORT = DummyObjectProvider.createFor(TeleportType.class, "ENTITY_TELEPORT");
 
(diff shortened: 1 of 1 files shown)


mattmess1221:
Would probably be better to name it `ItemTeleport`? That would cover mods that add items that teleport you.

ryantheleach:
I believe changing it to ItemTeleport would make users expect enderpearls to be covered by it, even if it's not necessarily true (ender pearls are projectiles, and have different semantics). 

If it turns into a generic itemTeleport, would it be possible to include the snapshot of the enderpearl or something in the cause, noting that this would be a royal pain in the butt, as you would need to attach it as data to the enderpearl projectile.

But then what about modded items...

parlough:
@ryantheleach I have some work locally that supports both but it's a bit iffy and I've found some other issues with teleport cause/context tracking I'm trying to work through.

parlough:
Will replace this with a more generic item teleport type when I get a chance.

Issue #1665, opened 2017-10-12
Leftover methods with causes.


Cybermaxke:
`Container#open(Player, Cause)`: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/item/inventory/Container.java#L62
`Container#close(Player, Cause)`: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/item/inventory/Container.java#L71

Cybermaxke:
Got fixed in https://github.com/SpongePowered/SpongeAPI/pull/1710

PR #1710, opened 2017-12-01
CraftingInventory is not a Grid

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1710) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1585) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1833)

A crafting inventory is not a grid but contains one.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Container.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Container.java
@@ -56,18 +56,16 @@ public interface Container extends Interactable {
      * Shows this Inventory to the given viewer.
      *
      * @param viewer The viewer to show this inventory to
-     * @param cause The {@link Cause} to use when opening the inventory
      * @throws IllegalArgumentException if a {@link PluginContainer} is not the root of the cause
      */
-    void open(Player viewer, Cause cause) throws IllegalArgumentException;
+    void open(Player viewer) throws IllegalArgumentException;
 
     /**
      * Stops showing this Inventory to the given viewer.
      *
      * @param viewer The viewer to stop showing this inventory to
-     * @param cause The {@link Cause} to provide when closing the inventory
      * @throws IllegalArgumentException if a {@link PluginContainer} is not the root of the cause
      */
-    void close(Player viewer, Cause cause) throws IllegalArgumentException;
+    void close(Player viewer) throws IllegalArgumentException;
 
 }
--- a/src/main/java/org/spongepowered/api/item/inventory/crafting/CraftingInventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/crafting/CraftingInventory.java
@@ -24,7 +24,7 @@
  */
 package org.spongepowered.api.item.inventory.crafting;
 
-import org.spongepowered.api.item.inventory.type.GridInventory;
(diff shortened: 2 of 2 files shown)


parlough (approved):

JBYoshi (approved):

Cybermaxke:
@Faithcaio Could you also fix the following issue in this PR, it's related to `Container` methods still using `Cause`s: https://github.com/SpongePowered/SpongeAPI/issues/1665

kashike (approved):

Faithcaio:
@Cybermaxke it seems they dont even have an impl yet.
Same for `getViewers hasViewers canInteractWith`

DosMike:
The biggest problem with Inventories in my opinion is that there is no `Slot Inventory.getNthSlot(int n)`.
The InventoryBuilder returns a SpongeCommon CustomIventory, that has getter and setter for Slots (`getStackInSlot` and `setInventorySlotContents`) , but those are not exposed to the API, so I'm using reflection and a signature scanner to use `setInventorySlotContents` for my menus. @codeHusky seems to be using [the iterator and a counter variable when setting items at a specific slot](https://github.com/codeHusky/HuskyCrates-Sponge/blob/d1538d184af61890dae46efe511d1981145da32a/src/main/java/com/codehusky/huskycrates/crate/views/SpinnerCrateView.java#L94)
For me it is kinds important to be able to directly access slots in some way, and I think if makes writing menus with toggles (click at slot x, item at slot x e.g. changes color) way easier.

Thinking about it: would it be possible/viable to add something like `Inventory.getSlot(int... position)`? I'm thinking about non-grid-inventories being accessed as kind of grid inventories like this
```Java
public Inventory Inventory.getInventory(int... position) {
  Inventory target = this;
  int n=0;
  for (int index : position) {
    if (index < 0 || index >= target.capacity())
      throw IllegalArgumentException("Index "+n+" ("+index+)" was out of inventory bounds (max: "+target.capacity()+")");
    if (target.capacity()<=1) //Slot or Empty
      throw IllegalArgumentException("Position argument has more depth than this inventory");
    // go to 'index' sub-inventory:
    Iterator<Inventory> it = iterator(); for (int i=0; i<index; i++) target = it.next();
    n++;
  }
  return target;
}
```
For example in the MainPlayerInventory this would allow access to a slot in the GridInventory with `gridInventory.getInventory(row, slot)` as convenience method.

Faithcaio:
@Mumfrey ^?

Thats what query is for.
But `#query(SlotIndex.of(index))` (SlotPos for x/y coordinate) returns ALL slots that are at that SlotIndex including those from rows, columns etc.

You can also query for `OrderedInventory.class` or `GridInventory.class`
and then after an instanceof check use the methods on there. `#getSlot(index)` and `#getSlot(x, y)`

Anyways this would have to be another PR.

DosMike:
Oh, didn't know query accepts a SlotIndex instance as there's now overlaod with SlotIndex and the javadocs for query(Object...) don't mention it either as far as I can see 🤔

Faithcaio:
SlotIndex etc. are InventoryProperties
`#query(InventoryProperty<?, ?>... props)`

PR #1711, opened 2017-12-02
Fix criteria spelling from critieria to criteria - Fixes #1669

### LINKED ISSUE #1669: Misspelling in org.spongepowered.api.scoreboard.critieria packages
The correct spelling is criteria, not critieria.

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1630)

Fixes "criteria" spelling - #1669 

This is a minor spelling cleanup of a package name so it is breaking. So we have to decide if we want to make this change and whether we should make it now or in API 8. Since API 7 is not stable yet, may be best to make the change now to get it over with.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -188,7 +188,7 @@ sortClassFields {
     add 'main', 'org.spongepowered.api.item.inventory.equipment.EquipmentTypes'
     add 'main', 'org.spongepowered.api.scoreboard.CollisionRules'
     add 'main', 'org.spongepowered.api.scoreboard.Visibilities'
-    add 'main', 'org.spongepowered.api.scoreboard.critieria.Criteria'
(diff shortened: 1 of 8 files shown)


Cybermaxke:
Maybe this should be change to `criterion` since we are going with singular package names?

liach:
Not really, we have a package called `data` which is a plural (the singular form is `datum`)

Cybermaxke:
@liach I don't think we have a `data` package within the where the singular form would be `datum`, `org.spongepowered.api.data` has nothing to do with `datum`.

liach:
@Cybermaxke Now, I think this is for sponge leaders to decide.

gabizou:
@Zidane I think this should merge ni API 8, as commands breaking will fall in line with this change.

parlough:
Will handle this with a direct commit later on.

PR #1712, opened 2017-12-03
Refactor BlockChangeFlags

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1635)
This used to be an enum, now it isn't. This also eliminates the implementation detail of the various bitwise flags that Minecraft uses.

Will elaborate further in the morning or something tomorrow.


liach:
Can you add an `inverse` for flags?

PR #1716, opened 2017-12-11
Makes necessary changes for damageabledata implementation

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1643)

This makes changes to the API for implementing `DamageableData`. 

I need a lot of feedback on this, especially on turning `Living` to `EntitySnapshot`. I believe it does have to be changed but this may not be the best choice, possibly simply `UUID` is better.

More review is needed in the implementation though, as I wasn't exactly sure the best way to do everything. For example, I followed the old implemented fashion of `Living#lastDamage` where it would be empty if the `lastAttacker` was empty as well. Also I was not 100% on the various `toContainer` and `fill` methods.

To-do:

- [x] Clean up changes

- [x] Create a test plugin

- [x] Test changes thoroughly

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -1306,6 +1306,9 @@ public final class Keys {
      *
      * <p>This will usually be an entity snapshot of a {@link Living}.</p>
      *
+     * <p>This data will usually only be present within 100 ticks of the attack
+     * occurring.</p>
+     *
      * @see DamageableData#lastAttacker()
      */
     public static final Key<OptionalValue<EntitySnapshot>> LAST_ATTACKER = KeyFactory.fake("LAST_ATTACKER");
@@ -1321,6 +1324,9 @@ public final class Keys {
     /**
      * Represents the {@link Key} for the last amount of damage received by an
      * {@link Entity}.
+     *
+     * <p>This data will usually only be present within 100 ticks of the attack
+     * occurring.</p>
      *
      * @see DamageableData#lastDamage()
      */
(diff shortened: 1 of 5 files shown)


parlough:
So Minecraft resets the `revengeTarget` or in our case `LAST_ATTACKER` to null every 100 ticks, so this data is really only present for 5 seconds after it occurs. I have documented this in the API, but am looking for feedback if it should be kept longer ourselves. Since we use `EntitySnapshot` or could use `UUID` instead we could store it ourselves separately so we can keep it longer than 5 seconds(until the next attacker really) if that is the desired functionality. 

If we went this route, I think adding the `Instant` they were last attacked would be helpful as well.

ryantheleach:
How does it track the every 100 ticks? just tracks the entity life span and mod 100?

I think we should only expose the instant if there is an instant recorded, until someone requests it I'm wary of feature creep where someone could (in a theoretical Utopian Sponge that has perfect perf characteristics) listen for the event that's fired when this data changes, and cancel it or modify it how they like.

parlough:
@ryantheleach 

```java
if (this.revengeTarget != null)
{
    if (!this.revengeTarget.isEntityAlive())
    {
        this.setRevengeTarget((EntityLivingBase)null);
    }
    else if (this.ticksExisted - this.revengeTimer > 100)
    {
    this.setRevengeTarget((EntityLivingBase)null);
    }
}
```

ryantheleach:
Yeah, I'd expose the revenge timer, as an 'instant'(long) of the world time, as opposed to an Instant that represents real world time.

parlough:
@ryantheleach Hmm maybe I'm concerned that's very implemention specific though. I'd like to hear some other's opinions as well. 

How do you feel about idea of storing the entity snapshot so we can expose this information for longer than 5 seconds? That's my main concern.

ryantheleach:
I'd be more interested in a UUID or entity reference, I think we can quickly fall into the trap of having too many EntitySnapshots in Sponge, they arn't the most efficient things.

parlough:
@ryantheleach If we keep it in its current state of returning within only the 5 seconds Minecraft allows living would work fine, since we can just return empty if the entity is dead or removed.

PR #1718, opened 2017-12-17
Modify child command logic to allow for meaningful usage on subcommand failure

### LINKED ISSUE #1165: Command usage displayed incorrectly

### LINKED ISSUE #1272: arguments for fallback executors vs parents is ambiguous.
Given:

```
CommandExecutor pageExecutor = (src, args) -> {
                args.<ActivePagination>getOne("pagination-id").get().specificPage(args.<Integer>getOne("page").get());
            };
            SpongeImpl.getGame().getCommandManager().register(SpongeImpl.getPlugin(), CommandSpec.builder()
                .description(t("Helper command for paginations occurring"))
                .arguments(new ActivePaginationCommandElement(t("pagination-id")))
                //fallback
                .executor(pageExecutor)
                .child(CommandSpec.builder()
                           .description(t("Go to the next page"))
                           .executor((src, args) -> {
                               args.<ActivePagination>getOne("pagination-id").get().nextPage();
                           }).build(), "next", "n")
                .child(CommandSpec.builder()
                           .description(t("Go to the previous page"))
                           .executor((src, args) -> {
                               args.<ActivePagination>getOne("pagination-id").get().previousPage();
                           }).build(), "previous", "prev", "p")
                .child(CommandSpec.builder()
                           .description(t("Go to a specific page"))
                           .arguments(integer(t("page")))
                           .executor(pageExecutor).build(), "page")
                .build(), "pagination", "page");
```

Results in

image

Because the arguments of the parent command are assumed to be belonging to the fallback and not the parent command.

What would have been expected is 

Usage: /page [pagination-id] next|previous|page

The issue is that there is no concept of arguments belonging to a fallback executor separately to that of the parent.

Some ideas I have floating around is introducing an explicit .fallback() to the builder, and deprecating the use of parent executors for fallbacks. that way the fallback could additionally require additional commands?

Maybe I'm wrong and the fallback I'm attempting to create could be covered in another way.

https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/command/spec/CommandSpec.java#L249

Changing 
`arguments(firstParsing(childDispatcher, this.args));`
to  `arguments(this.args, firstParsing(childDispatcher, fallbackArgs));`
would fix this somewhat, but probably break someone's existing plugin somewhere.

After looking at this some more, it's currently not possible to set a child without an alias.

So maybe the functionality I'm thinking of would fit that better? 

Or maybe the child command element arguments could be specified, and if found replaced rather then pre/appended?

### LINKED ISSUE #1542: Child command argument parse error handling
If using a command spec with child commands and a fallback handler, the behaviour in case of an argument parse error is confusing. If I specify the child command name correctly, but have an error in the child command's arguments, the argument parser will fall back to parse the arguments for the fall back handler.

### LINKED ISSUE #1647: Confusing error with player argument
There is a confusing error when the `player` argument is used inside a subcommand and it's not the last argument of the command:
image
(The usage is also wrong, but it has already been said in another issue)

When it's not used inside a subcommand or when it's the last argument, it works as expected:
image

Example code to reproduce the issue:
```java
CommandSpec addEcuSpec = CommandSpec.builder()
        .executor((src, args) -> {
            Player player = args.<Player>getOne("player").get();
            double amount = args.<Double>getOne("amount").get();

            src.sendMessage(Text.of("Target: " + player.getName() + ", amount: " + amount));
            return CommandResult.affectedEntities(1);
        })
        .arguments(onlyOne(player(Text.of("player"))), doubleNum(Text.of("amount")))
        .build();

CommandSpec ecuSpec = CommandSpec.builder()
        .executor((src, args) -> CommandResult.empty())
        .child(addEcuSpec, "add")
        .build();

Sponge.getCommandManager().register(this, ecuSpec, "ecu");```

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1656)

One of the long standing issues with the Command API is that subcommands don't give useful feedback if a command fails. The current implementation of the command system is difficult to solve this effectively with - this is a best effort. It isn't breaking, but in order to activate behaviour where a subcommand failing gives meaningful usage requires `childArgumentParseExceptionFallback` on the CommandSpec builder to be set to `false`. It defaults to `true`, the current behaviour.

I felt it was better to propose this fix for API 7, even though #1587 will completely eclipse this in API 8. It gives devlopers the choice on how to handle their subcommands now. It honestly feels a little hacky, but it works. In theory, there is nothing stopping this going into API 5 and 6 either. Note that any plugin that doesn't use this method in the builder will get the old behaviour, so the change isn't breaking API wise or behaviour wise, plugins will have to update to use this new behaviour - there are some plugins out there, no doubt, that use the fallback as it is now.

Comments are welcome on naming (it's not my strong suit). Test plugin included in Common, some unit tests in API.

Fixes #1165, fixes #1272, fixes #1542, fixes #1647

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/ArgumentParseException.java
+++ b/src/main/java/org/spongepowered/api/command/args/ArgumentParseException.java
@@ -27,6 +27,7 @@ package org.spongepowered.api.command.args;
 import com.google.common.base.Strings;
 import org.spongepowered.api.command.CommandException;
 import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.annotation.NonnullByDefault;
 
 /**
  * Exception thrown when an error occurs while parsing arguments.
@@ -76,6 +77,10 @@ public class ArgumentParseException extends CommandException {
         }
     }
 
+    private Text getSuperText() {
+        return super.getText();
+    }
+
     /**
      * Return a string pointing to the position of the arguments when this
      * exception occurs.
@@ -120,4 +125,29 @@ public class ArgumentParseException extends CommandException {
     public String getSourceString() {
         return this.source;
     }
+
+    /**
+     * An {@link ArgumentParseException} where the usage is already specified.
+     */
+    public static class WithUsage extends ArgumentParseException {
+        private static final long serialVersionUID = -786214501012293475L;
+
+        private final Text usage;
+
+        WithUsage(ArgumentParseException wrapped, Text usage) {
+            super(wrapped.getSuperText(), wrapped.getCause(), wrapped.getSourceString(), wrapped.getPosition());
+            this.usage = usage;
+        }
+
+        /**
+         * Gets the usage associated with this exception
+         *
+         * @return The usage
+         */
+        public Text getUsage() {
+            return this.usage;
+        }
+
+    }
+
 }
--- a/src/main/java/org/spongepowered/api/command/args/ChildCommandElementExecutor.java
+++ b/src/main/java/org/spongepowered/api/command/args/ChildCommandElementExecutor.java
@@ -28,10 +28,12 @@ import static org.spongepowered.api.command.CommandMessageFormatting.error;
 import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Multimaps;
 import org.spongepowered.api.command.CommandCallable;
 import org.spongepowered.api.command.CommandException;
 import org.spongepowered.api.command.CommandMapping;
+import org.spongepowered.api.command.CommandMessageFormatting;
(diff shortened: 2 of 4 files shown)


kashike (approved):

parlough (approved):

dualspiral:
I'll merge this tomorrow if there are no late objections or naming suggestions. Would do it tonight, but other things are currently on my plate.

PR #1720, opened 2017-12-22
Move away from KeyFactoring to a Key builder

gabizou:
Eliminates some implementation detail in `Key`s and how they are used. Allows for some in depth registration to make an indexer for finding keys by plugin as they are registered.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Key.java
+++ b/src/main/java/org/spongepowered/api/data/key/Key.java
@@ -26,12 +26,19 @@ package org.spongepowered.api.data.key;
 
 import com.google.common.reflect.TypeToken;
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataQuery;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.value.BaseValue;
 import org.spongepowered.api.data.value.ValueContainer;
-import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.util.ResettableBuilder;
+import org.spongepowered.api.util.TypeTokens;
+
+import java.lang.reflect.Type;
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/data/key/KeyFactory.java
```diff
@@ -71,6 +71,7 @@ private KeyFactory() {}
      * @param name The name for the new key
      * @return The generated key
      */
+    @Deprecated
```
Deamon5550: Need some deprecation javadocs here with the alternative

Deamon5550 (approved):
lgtm pending javadocs

JBYoshi:
Will the deprecated `KeyFactory` methods be rewritten to delegate to `Key.Builder`?

liach:
I believe the whole class is going to be removed.

src/main/java/org/spongepowered/api/data/key/Key.java
```diff
@@ -66,4 +74,23 @@
      */
     DataQuery getQuery();
 
+    interface Builder<E, V extends BaseValue<E>> extends ResettableBuilder<Key<V>, Builder<E, V>> {
+
+        <T, B extends BaseValue<T>> Builder<T, B> valueElement(TypeToken<T> element, TypeToken<B> token);
```
Cybermaxke: Just the value typetoken is sufficient, you can extract the element typetoken from the value typetoken.

Cybermaxke: I would also name the method differently, maybe `valueType`?

src/main/java/org/spongepowered/api/data/key/Key.java
```diff
@@ -66,4 +74,23 @@
...
+
+        Builder<E, V> keyId(String id);
```
Cybermaxke: Just `id`?

Cybermaxke: We should be more consistent with builder methods for catalog types:
https://github.com/SpongePowered/SpongeAPI/issues/1721

PR #1723, opened 2017-12-23
Allow plugins to use a separate implementation class for custom data.

JBYoshi:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1666)

Sponge uses an interface and a separate implementing class for its `DataManipulator`s, and a few plugin developers have been [trying to use that model](https://forums.spongepowered.org/t/custom-data-not-working/21832?u=jbyoshi). However, `DataRegistration` doesn't allow multiple classes to be registered to it. This PR adds that option. (However, it remains optional for any plugins that don't currently use it, or don't want to use it.)

### DIFF
--- a/src/main/java/org/spongepowered/api/data/DataRegistration.java
+++ b/src/main/java/org/spongepowered/api/data/DataRegistration.java
@@ -54,11 +54,27 @@ public interface DataRegistration<T extends DataManipulator<T, I>, I extends Imm
     Class<T> getManipulatorClass();
 
     /**
+     * Gets the implementing class of the {@link DataManipulator} for this
+     * registration.
+     *
+     * @return The manipulator class registered
+     */
+    Class<? extends T> getImplementationClass();
+
+    /**
      * Gets the {@link ImmutableDataManipulator} class for this registration.
      *
      * @return The immutable class registered
      */
     Class<I> getImmutableManipulatorClass();
+
+    /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -53,13 +53,29 @@
      */
     Class<T> getManipulatorClass();
 
+    /**
+     * Gets the implementing class of the {@link DataManipulator} for this
+     * registration.
+     *
+     * @return The manipulator class registered
+     */
+    Class<? extends T> getImplementationClass();
```
gabizou, line 62: Is the implementing class optional?

gabizou (commented):
You said the implementing class is optional, but you don't have the return value as `Optional`. Or are you suggesting that if the manipulator class is the implemented class, that you have the same class being returned for both methods?

src/main/java/org/spongepowered/api/data/DataRegistration.java
JBYoshi: No. It returns the API class if there isn't a separate implementing class.

gabizou: Then you should clarify that in the documentation, personally speaking it makes better sense.

felixoi:
Is it possible to complete this up to API 7 release?

PR #1724, opened 2017-12-23
Deprecate skylands generator

Deamon5550:
Doesn't have any place being implemented in the api, could find a new home in a cookbook plugin.


gabizou (approved):
Since the Skylands never got fully developed, it's something that has little place in the API these days.

Maybe what it can be moved to is the implementation test plugin where you can optionally set up a world with the generator.

parlough (approved):

stephan-gh (approved):

kashike (approved):

parlough:
Most likely will remove in API 8 due to the high chance no one is using this.

PR #1728, opened 2017-12-28
Allow the Registration of EventListeners for Keys

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1670)
This is to supersede #1695 in some aspects to make it more functional to allow plugins to isolate specific key event listeners. The implementation of the event is also directing the flow of this being used.


Cybermaxke:
Could the `ChangeDataHolderEvent` be renamed to just `DataHolderEvent` and get a generic type `H` which represents the holder type of the event. The event can extend `GenericEvent` later when it is implemented, but the generic holder type is still useful because you don't need the generic event when registering the listener in the `Key`.

PR #1729, opened 2017-12-31
Add cooldown events

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1674) 

Adds need cooldown events which I should have added before and modifies the return types to reflect that.


parlough:
@gabizou I would like to get this in before the API 7 close tomorrow due to some of the method signatures changing.

PR #1730, opened 2018-01-01
add CraftItemEvent

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1730) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1690)

Adds CraftItemEvent Preview and Craft.

`CraftItemEvent.Preview` is fired every time the recipe is changed in the CraftingGrid.
You can set another item as the preview which will then later be crafted instead. `event.getPreview().setCustom(item)`
Canceling the Preview Event sets the outputslot to air as if the recipe did not exist.

`CraftItemEvent.Craft` is fired for every stack taken out of the output slot.
`event.getCrafted()` provides you with the actual crafted item.
Canceling the Craft Event prevents taking items out of the output slot.

Like with other inventory events you can still modify all slot transactions to your liking.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/CraftItemEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/CraftItemEvent.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/CraftItemEvent.java
```diff
@@ -56,7 +56,7 @@
      * This event is fired before the item is taken out of the
      * output slot but after completing the recipe in the grid.
```
ryantheleach: "This event is fired when a recipe has been completed and is rendering in the crafting output slot, but before the recipe is actually crafted."

ryantheleach:
https://github.com/SpongePowered/SpongeAPI/pull/1730/files/f7926cdbc4a389b8eb8c470106c31d9c6a0ffa70..0e4545da60ea9c411bd91b42da5841d369e2b375 Sorry for wasting your time faith, it would have been cool.

gabizou:
The real question is why you have to mock the slot transaction if it's supposed to be final?

ryantheleach:
He wasn't, the automated tests for the events was attempting to mock it or something.

Issue #1618, opened 2017-07-20
Checking if a player (or entity?) is currently elytra-flying (DataManipulator)


TBlueF:
There has to be a way to detect if a player/entity is currently flying with an elytra.
Something like:
```java
Optional<Boolean> isFlying = player.get(Keys.IS_ELYTRA_FLYING);
```

ryantheleach:
player.get(Keys.SUPER_STEVE)? @Meronat 

ELYTRA_FLYING mode should probably be in the API in some manner, @connorhartley will need it for guardian.

@Meronat Do you know if you can set this independently of the player wearing elytra in Minecraft? I know it's possible to tell that to the client at least thanks to Lantern.

I'm wondering if it would be a failable data, or if it's more like a property in Vanilla MC.

parlough:
Super Steve to the rescue! @ryantheleach 

I agree blue. This is a key we have in lantern, and it would be useful in Sponge. Shouldn't be too hard to add either, I'll do it :)

What data manipulator should I put it in though? `FlyingData` or a new `ElytraFlyingData`? 

@ryantheleach Hmm I will have to look into that. There is this method, not sure what it actually does yet though - 
![image](https://user-images.githubusercontent.com/18372958/28446894-4de8a436-6d93-11e7-927c-16cb8bfc91a8.png)

Almost have it implemented, now on to testing.

...

So it's all done, just having some issues with offering the value. I am missing something somewhere. Will try to finish this tomorrow. 

...

So after going on a trip, I hope to come back to this soon. I finished the core idea actually, but ended up adding more data to allow players to fly elytra like without an elytra on. There is some cleanup I have to do and I might also add a key that can control whether they stop elytra flying once the player hits the ground.

TBlueF:
Any news on this? :)

parlough:
@TBlueF Uh sorry, I haven't had a chance to be on much, but I will come back to this branch soon. If I can't find time in the next few days I'll just add the IS_ELYTRA_FLYING data for now since I think I have that done.

I'll probably create an issue on SpongeAPI with a larger proposal to see what people think of some of the others changes/additions I'd like to handle in these data manipulators.

Zidane:
Closing in favor of Meronat's proposal found here: https://github.com/SpongePowered/SpongeAPI/issues/1663

Issue #1663, opened 2017-10-11
ElytraFlyingData and ElytraCapability proposal


parlough:
_So I am looking for some input on this as some of it may not be fit to be in Sponge outside of the check to see if someone is flying with an elytra._ 

Essentially I would like to add `ElytraFlyingData` to the SpongeAPI, which allows you to not only check if a player is flying with an elytra, but also force them to start flying with an elytra and allow them to fly like they are using an elytra without one actually on. Some of this is not exactly the smoothest implementation in SpongeCommon(SV/SF), with some slightly weird changes to Minecraft. 

The main API this PR would add would be a way to check if a player is currently elytra flying, but it would also be possible to force them to start elytra flying or stop them from elytra flying by offering the relevant value. Due to this ability to force players to elytra fly, a sort of elytra capability is needed to control the check if a player can actually elytra fly.

A large part of this would be a player's "elytra capability" which would essentially be an enum that is accessible from a player's `ElytraFlyingData` which represents the various ways a player could elytra fly, whether that's disabled(even with elytra), enabled(even without elytra), or normal requiring an item with a true `ElytraFlyingProperty`(which in Vanilla is only the Elytra). One thing that must be recognized, is that a player cannot activate elytra flying themselves if they are not wearing one or some other item with the property not registered locally through Forge as an elytra item(would need to finish [this PR](https://github.com/MinecraftForge/MinecraftForge/pull/3851) in some fashion though first). As a whole the API would look something like this: 

<details>
 <summary>ElytraCapability.java</summary>

```java
/**
 * Represents the possible capabilities a player can have relating to
 * elytra flying.
 */
public enum ElytraCapability {

    /**
     * Requires an item with a true {@link ElytraFlyingProperty} to be
     * worn in the chest slot.
     */
    EQUIPMENT,

    /**
     * Allows a player to fly like they are wearing an elytra item 
     * without actually wearing one.
     */
    ENABLED,

    /**
     * Does not allow a player to fly like they are wearing an elytra
     * item, even if they are. 
     */
    DISABLED

}
```
</details>

<details>
 <summary>ElytraFlyingProperty.java</summary>

```java
/**
 * Represents the property which shows whether or not an item works
 * as an elytra for flying.
 */
public class ElytraFlyingProperty extends BooleanProperty {

    /**
     * Creates a new {@link ElytraFlyingProperty} with the provided boolean value.
     *
     * @param value The value
     */
[...]

PR #1737, opened 2018-01-10
Adds basic elytra data

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1702)

Adds a basic elytra data which can be expanded later on as this part of it is a common request.

With this API, you can get whether a player is flying elytra style or not. You can also stop them from flying elytra style with `player.offer(Keys.IS_ELYTRA_FLYING, false)` and have them start elytra flying if they are currently falling and have an elytra on with `player.offer(Keys.IS_ELYTRA_FLYING, true)`.

More capability will be introduced later down the road, but this is the most common request and was more easily done.

Fixes https://github.com/SpongePowered/SpongeAPI/issues/1618 and various Discord requests

Works towards https://github.com/SpongePowered/SpongeAPI/issues/1663


kashike (approved):

Zidane:
@Meronat 

Squash this into one commit for me please.

PR #1740, opened 2018-01-14
Fix SpongeCommon#1282

ImMorpheus:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1710)

Necessary for https://github.com/SpongePowered/SpongeCommon/issues/1282

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/Player.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/Player.java
@@ -255,7 +255,7 @@ public interface Player extends Humanoid, User, RemoteSource, Viewer, ChatTypeMe
      * @return True if played before, false otherwise
      */
     default boolean hasPlayedBefore() {
-        return !firstPlayed().equals(lastPlayed());
+        return !firstPlayed().get().equals(lastPlayed().get());
     }
 
     /**


src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -255,7 +255,7 @@ default JoinData getJoinData() {
      * @return True if played before, false otherwise
      */
     default boolean hasPlayedBefore() {
-        return !firstPlayed().equals(lastPlayed());
+        return !firstPlayed().get().equals(lastPlayed().get());
```
ryantheleach, line 258: You add the .gets, do Values not use deep equals?

ryantheleach:
Is removing the default on this interface a break that prevents it to be on API 7.1? 

Because my recommendation would be to remove the default, and remove the implementation from the API. But it might need to be delayed to API 8.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
ImMorpheus: They do, and I have no idea why this was not working on my dev environment.

gabizou: It's more than likely because the default values are differing. Troubleshoot through with AbstractBaseValue.

ryantheleach: That would make sense.

gabizou:
Going to close this as the API change is rendered unnecessary as the implementation can override the method itself. I'd rather break the API for bleeding to include a difference of `LastPlayed` versus `LastJoined` and being able to differentiate whether a player has played before if the `LastJoined` is the same as `FirstJoined`.

PR #1742, opened 2018-01-17
Minecraft 1.13 - Refactor statistics

Cybermaxke:
I separated this from https://github.com/SpongePowered/SpongeAPI/pull/1704 because it's getting quite big and getting a separate conversation going for redesigning the statistics API.

Goals:
- Update for 1.13 statistic names and categories (previously types)
- Remove statistic methods from the game registry, reworked into `CatalogStatisticCategory`
- Remove `ItemStatistic`, `BlockStatistic` and `EntityStatistic`, which got merged into `CatalogStatistic`

TODO:
- [ ] I don't like the names `CatalogStatistic` and `CatalogStatisticCategory`, suggestions?
- [ ] Update `Statistics` catalog class with proper names.


parlough:
@windy1 Was the last one to touch the statistic API and implementation to any large extent. If he has the time/chance would be nice to hear his input.

ImMorpheus:
This is carrying on the old broken division of StatisticTypes.
Vanilla has only 4 "StatisticTypes": BLOCKS, GENERAL, ITEMS, MOBS.

Note: in vanilla some Statistics are used for multiple StatisticTypes: DROPPED, CRAFTED, ecc.

Since this is an API pr, there's not much point in talking about the implementation but here's what I came up so far before Meronat pointed me to this issue.
https://github.com/ImMorpheus/SpongeAPI/commit/25634a66471e298d0800453cdd420e50ae3a7bb9
https://github.com/ImMorpheus/SpongeCommon/commit/c00518c0591930395e0b01ac8e2b48b42b67bcde

Cybermaxke:
@ImMorpheus I based the PR on the protocol changes:
http://wiki.vg/Pre-release_protocol#Statistics
And a category generates statistics based on all the catalog type of specific type, for example within the `mined` category are mined statistics for every block type in the game.
The custom category is different because it's not attached to catalog types.

ImMorpheus:
Fair enough. I based my comment on mojang code and the entries in the en_us.lang (broken getTranslation() was the reason I started working on statistics).

I'll wait the implementation and see how this will turn out.

ST-DDT:
Maybe we can simplify this a little bit by dropping the `CatalogStatisticType` class and just and add an `Optional<T> getReferencedCatalog()` (or `Data`) to Statistic.  That way we won't have issues with conflicting names either.

Then you could extract the targeted type(class) via the `StaticisticCategory` and the actually targeted entry via the statistic.

The only drawback of this is that supporting more specific statistics like "Killed Dragons with a Stick" won't be able to present all its data.

In addition to that I'm a little reluctant to put the catalog type in a generic category. Especially since there is no real type for `CUSTOM` and `Statistic` is just a placeholder for "nothing". At least allow everything including none Catalogs. That way we could just write `Void` (or `Tuple`) in there.

Cybermaxke:
Updated and merged into 1.13

PR #1744, opened 2018-01-22
Move text implementation to the implementation

kashike:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1724) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1971)


parlough:
Some related PRs:

https://github.com/SpongePowered/SpongeAPI/pull/1477
https://github.com/SpongePowered/SpongeAPI/pull/1634

Maybe for consideration:

https://github.com/SpongePowered/SpongeAPI/pull/1424

src/main/java/org/spongepowered/api/GameRegistry.java
```diff
@@ -374,13 +373,13 @@
...
+     * Gets the internal {@link TextFactory}.
      *
-     * @return The text serializer factory
-     * @deprecated Use {@link TextSerializers} instead.
+     * @return The text factory
+     * @deprecated Internal use only.
      */
     @Deprecated
-    TextSerializerFactory getTextSerializerFactory();
+    TextFactory getTextFactory();
```
gabizou, line 416: What is this used for?

gabizou: That needs to exist in api?

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -739,59 +460,27 @@ public Builder trim() {
...
+    }
 
+    /**
+     * Returns an empty, unformatted {@link Text} instance.
+     *
+     * @return An empty text
+     */
+    @SuppressWarnings("deprecation")
+    static Text of() {
+        return Sponge.getRegistry().getTextFactory().emptyText();
```
gabizou: Why do we use the factory instead of a builder?

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -802,14 +491,9 @@ public static Text of() {
...
-        if (checkNotNull(content, "content").isEmpty()) {
-            return LiteralText.EMPTY;
-        } else if (content.equals(NEW_LINE_STRING)) {
-            return NEW_LINE;
-        } else {
-            return new LiteralText(content);
-        }
+    @SuppressWarnings("deprecation")
+    static LiteralText of(String content) {
+        return Sponge.getRegistry().getTextFactory().literal(content);
```
gabizou: This honestly feels like we can static builder the individualized texts like all other builders instead of using a factory

gabizou (changes requested):
We shouldn’t have any more reason to use the factory. A combination of explicit builder methods like ‘LiteralText.builder()’ or ‘Text.literal()’

src/main/java/org/spongepowered/api/text/LiteralText.java
```diff
@@ -178,152 +76,69 @@ public final String getContent() {
...
-        public Builder remove(Text... children) {
-            return (Builder) super.remove(children);
-        }
+        Builder removeAll();
 
         @Override
-        public Builder remove(Collection<? extends Text> children) {
-            return (Builder) super.remove(children);
-        }
+        Builder from(Text value);
```
liach: Can a literal text's builder copy from any text? How does that work?

PR #1745, opened 2018-01-23
Change up game rules to be more flexible and follow Sponge practices

parlough:
Follows the style the rest of Sponge uses better while also providing a more flexible system.

Looking for some input on the set-up I have so far, then I will flesh out the rest of the API and write the implementation.

@Cybermaxke Any input here?


parlough:
Locally the file `GameRule.java` has the right name but Git seems to think the name is `Gamerule.java` :/

src/main/java/org/spongepowered/api/world/gamerule/GameRuleHolder.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+package org.spongepowered.api.world.gamerule;
+
+import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.storage.WorldProperties;
+
+import java.util.Optional;
+
+/**
+ * Someone who holds various {@link GameRule}s and the values they point to.
```
ryantheleach, line 33: Someone? Were you going to delegate per user gamerules to the world properties that currently controls them?

ryantheleach:
If you are going to refactor GameRules I'm expecting that you are going to come up with some sort of sharing context? 

e.g. nether and the_end and modded dimenson-worlds by default on sponge delegate to the overworld gamerules?

Where as Sponge worlds can share a 'multiverse'/world that contains the gamerules for a group of dimensions?

Note: My use of multiverse is just to convey concept, I am utterly aware how much of a Pain in the ass it would be to globally have support for true multiverses separating out inventory, stats, recipes. plugins etc on a Mojang based server.

I just can't help but think that the concept would be valuable, even if it only see's limited use on SpongeForge / SpongeVanilla. There isn't anything really forcing Lantern to having such limitations.

src/main/java/org/spongepowered/api/world/gamerule/GameRuleHolder.java
parlough: I meant "Something", just in-case an implementation wants to allow gamerules to extend beyond just a `WorldProperties`.

src/main/java/org/spongepowered/api/world/gamerule/GameRuleType.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.function.Function;
+
+/**
+ * A representation of the type a {@link GameRule}'s value has.
+ *
+ * @param <T> The value type of the game rule
+ */
+@CatalogedBy(GameRuleTypes.class)
+public interface GameRuleType<T> extends CatalogType {
```
parlough, line 39: I think I'll introduce a builder for this to make it easier to create a new one.

src/main/java/org/spongepowered/api/world/gamerule/GameRule.java
```diff
@@ -0,0 +1,54 @@
...
+import org.spongepowered.api.world.storage.WorldProperties;
+
+/**
+ * A representation of Minecraft's game rules which apply to worlds allowing
+ * access to their values from a world's {@link WorldProperties}.
+ *
+ * @param <T> The value type of the game rule
+ */
+@CatalogedBy(GameRules.class)
+public interface GameRule<T> extends CatalogType {
```
parlough, line 38: Will create a builder for this as well.

src/main/java/org/spongepowered/api/world/gamerule/GameRuleType.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return The type token of the value's type
+     */
+    TypeToken<T> getType();
+
+    /**
+     * Gets the deserializer for this game rule type.
+     *
+     * @return This type's deserializer
+     */
+    Function<String, T> getDeserializer();
```
gabizou, line 53: It’s not really a serializer, it’s more of a printer/parser combination, isn’t it?

src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -334,28 +332,6 @@ default Difficulty getDifficulty() {
         return getProperties().getDifficulty();
     }
```
Cybermaxke, line 334: Can you make the world also a `GameRuleHolder`? And then delegate all the methods to the properties.

src/main/java/org/spongepowered/api/world/gamerule/GameRuleHolder.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Sets the value of the specified game rule. If this game rule holder
+     * lacks this game rule, then it will be created automatically.
+     *
+     * @param gameRule The game rule to set the value for
+     * @param value The value to set the game rule to
+     * @param <T> The value type of the game rule
+     */
+    <T> void setGameRule(GameRule<T> gameRule, T value);
```
Cybermaxke, line 69: This method should also call the `ChangeWorldGameRuleEvent`, so return a boolean whether it was successful? That said, the `ChangeWorldGameRuleEvent` should also be refactored for the new API.

parlough: Ah yes I forgot to do this, thanks for reminding me. Will make these changes.

src/main/java/org/spongepowered/api/world/World.java
parlough: Yeah, I will just specify in the Javadoc that most implementations delegate the methods to the properties.

dualspiral:
While I know we're breaking commands and text for API 8 without deprecations, here, I'd argue deprecations serve a purpose. Deprecations will emit a warning on compilers that use the code to point them in the right direction - while we'll really hammer the change in commands when API 8 is out, game rules and smaller, easier to deprecate systems might end up falling by the wayside. We should make sure we are in the mindset of deprecating when we can, not just using major API versions as a reason to break with no consequence.

That being said, looking at this, I suspect it's possible for both systems to coexist peacefully with no breaking changes. Do you think you could do that in time for 7.1? If so, maybe that's a route forward - a deprecation in 7.1 with the new methods, removal in 8.

parlough:
@dualspiral I believe this is doable to an extent, the only problem would be for the events. I could introduce an event with a different name or package that supports the new changes while deprecating the old one.

How would you feel about that solution?

gabizou:
@dualspiral I disagree with trying to maintain deprecation policy for the bleeding branch. As we're already doing massive refactors left and right (Text has already changed, Value API is about to change, DataRegistration is partially changing due to values changing, Extent API is 100% rewritten at this point, Scheduling tasks will change as soon as we expose the client in the process, CauseStackManager is getting a rename, some major events are being redone, Inventory API changed, commands are going to be changing), I don't feel it necessary to keep something as simple as game rules deprecated for a version and then break it in a little way afterwards.

Not to mention that documentation alone for the migration from API 7 to API 8 will be somewhat extensive at that point given the PR's that were used to make all these changes, it'll be easy to show the migration via test plugin changes needed between the various pr's.

dualspiral:
@gabizou With the extent of changes since, where before, it was impossible to know with clarity how much we would need to change, sure. Deprecations _do_ serve a purpose and we should still pay mind to their use, but given how much we've changed since this was opened, I see the point.

Do bear in mind I made that comment 7 months ago, just after 7 was released. Things have changed!

gabizou:
> Do bear in mind I made that comment 7 months ago, just after 7 was released. Things have changed!

Of course, just resuscitating PR's (having went through I think all the open PR's currently) and making sure their either alive, or relegated for future development/management.

Cybermaxke:
The game rule is already refactored in the api-8 branch, comment on the status PR https://github.com/SpongePowered/SpongeAPI/pull/1995 if there's anything missing.

PR #1747, opened 2018-01-23
Fix checkstyle warnings, warnings during compilation, and other cleanup for stable-7

parlough:
A 7.1 version of my checkstyle PR without the breaking changes and targeting stable-7 instead of bleeding.


parlough:
Will clean this up and get it compiling once I get a chance, hopefully later tonight, then I'll get both the PRs in.

PR #1755, opened 2018-01-30
Add text replacement methods

pie-flavor:
These are very much needed, given the relative underpoweredness of `TextTemplate`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -29,6 +29,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataSerializable;
@@ -56,6 +57,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


ryantheleach:
@pie-flavor Is this still Work In Progress? if so reply with `~wip` to mark your issue as Work In Progress or reply with `~qa` to mark it as ready for Quality Assurance, if you wish it to be reviewed / tested by contributors / the community.

pie-flavor:
-qa

pie-flavor:
~qa

Defman:
Is Text not meant to be imutable: Represents an immutable tree-structure of formatted (text) components. Each instance consists of content and a list of children texts appended after the content of this text. The content of the text is available through one of the subclasses.

ryantheleach:
Sure, but it's a utility function for something that is inherently difficult to do yourself 100% correctly the first time. 

Strings are immutable too, doesn't stop them having replace, but it's not something you would want to (ab)use all the time.

My 2c.

dualspiral:
I don't see a problem with it as long as the javadocs make it clear (in both the body and the returns statement) that the method returns a copy of the `Text` object with the specified replacements.

pie-flavor:
@The-Defman I'm confused - where's the issue? The original text never gets modified.

@dualspiral I'll add that, though originally I never thought it could be an issue - anyone using Text.replace has inevitably used String.replace which behaves the same way.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -342,6 +346,116 @@ public final Text toText() {
...
+     * <p>This method produces a copy with the specified replacements. The
+     * original is not modified.</p>
+     *
+     * @param oldValue The pattern to replace
+     * @param newValue The value to replace with
+     * @param lossy Whether lossy mode is enabled
+     * @return The resulting (copied) text
+     */
+    public Text replace(Pattern oldValue, Text newValue, boolean lossy) {
+        Text t = this.children.isEmpty() ? this : this.toBuilder().removeAll().append(this.children.stream()
```
kashike: Please give this a longer name - single character variables should be reserved for `for`.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -342,6 +346,116 @@ public final Text toText() {
...
+     * 
+     * <p>This method produces a copy with the specified replacements. The
+     * original is not modified.</p>
+     *
+     * @param oldValue The pattern to replace
+     * @param newValue The value to replace with
+     * @param lossy Whether lossy mode is enabled
+     * @return The resulting (copied) text
+     */
+    public Text replace(Pattern oldValue, Text newValue, boolean lossy) {
```
kashike, line 307: Could you add a few comments to this method explaining what's going on? Will be nice for the next person who works on this.

kashike (changes requested):

randombyte-developer:
Can this be added to the next 7.x release or is that against semver?

pie-flavor:
No API methods get broken, so 7.x should be a valid target.

gabizou:
Needs to be updated since the text was refactored.

Cybermaxke:
@gabizou This targets API 7

ImMorpheus:
@Cybermaxke you need to change the target branch to stable-7 then

pie-flavor:
Oh good, the git history got clonked. Is this worth pursuing in its current state? Will there be another API 7 release?

ImMorpheus:
@pie-flavor Yes, please update and rebase

pie-flavor:
Looks all good from here.

XakepSDK:
Maybe `replace(Text, Text)`?

pie-flavor:
Properly searching for a Text would require remembering the effective formatting of the text in addition to the Text's own formatting, and additionally parsing children, greatly complicating the algorithm. And there wouldn't really be a point in a Text argument that's only ever used via toPlain.

ImMorpheus (approved):
Working as intended

Lignium:
You need to add these methods to API 8.

PR #1758, opened 2018-02-02
Added navigator interface for basic path operations

liach:
**SpongeAPI** | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1752)

This also adds a following range attribute modifier to the API to help plugins using the navigator.

To make things simple, the pull just exposed the most basic operation that is the most likely frequently used. I hope @Cybermaxke may comment on this so that different path finding systems may find an agreement.

Edit: added link to spongecommon

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/ai/navigation/Navigator.java
+++ b/src/main/java/org/spongepowered/api/entity/ai/navigation/Navigator.java
@@ -1,0 +1,104 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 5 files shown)


liach:
Just a self note, that sponge is currently in a cleanup phase; this feature may be left for review in the future time.

Cybermaxke:
Can't the `followRange` be task specific? And isn't it related to attributes?

liach:
@Cybermaxke Nope. It is based on attributes so that it cannot be task specific. Since it's based on attributes, I have both a data key and access (in the agent data) to the calcualted result (in navigator) available.

src/main/java/org/spongepowered/api/entity/ai/navigation/Navigator.java
```diff
@@ -0,0 +1,104 @@
...
+     * the entity's bounding box.</p>
+     *
+     * <p>Once set, the navigator will continuously try to
+     * guide the agent to the location of the target when this
+     * method is called: the location does not automatically update.</p>
+     *
+     * @param target The target entity
+     * @see #navigate(Location)
+     */
+    void navigate(Entity target);
```
ST-DDT, line 89: Please also add a method to clear the current path and check if a path is set.

JBYoshi:

liach:
Should start a new one after api 8 is in shape.

Issue #1586, opened 2017-06-18
CustomData and toContainer()


Faithcaio:
When implementing custom data people very often forget to implement toContainer().
probably because `AbstractData#toContainer` implements the Interface only adding the version.

Can we do something about this?

@gabizou

gabizou:
We can definitely provide an abstract method in bleeding and finalize the toContainer method. It was previously part of the class generation goal to eliminate this sort of issue.

PR #1762, opened 2018-02-03
add fillContainer for plugins to implement

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1761)

Plugindevs often forget to override `toContainer` this change forces them to implement `fillContainer`.
This has also the added benefit of always adding the content-version which could have been missed when not calling `super.toContainer()` in the override

https://github.com/SpongePowered/SpongeAPI/issues/1586

### DIFF
--- a/src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractData.java
+++ b/src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractData.java
@@ -207,9 +207,17 @@ public abstract class AbstractData<M extends DataManipulator<M, I>, I extends Im
     }
 
     @Override
-    public DataContainer toContainer() {
-        return DataContainer.createNew()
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractData.java
```diff
@@ -207,9 +207,17 @@ public boolean equals(Object obj) {
...
     }
 
+    /**
+     * Implement this method to add the data to be persisted.
+     *
+     * @param dataContainer The DataContainer
+     *
+     * @return The filled DataContainer
+     */
+    public abstract DataContainer fillContainer(DataContainer dataContainer);
```
mattmess1221: This should be made `protected` since it shouldn't need to be used outside this class structure.

gabizou (approved):

PR #1765, opened 2018-02-04
Adds some usefull ItemStack Comparators

### LINKED ISSUE #1257: ItemStack equality
In commit https://github.com/SpongePowered/SpongeAPI/commit/0c1d6c33d83d2b6fc909d611775b21185c76a6ce an additional check for item quantity was exposed. However, there is still a lack of a check to see if two ItemStacks are of the same variant, but with different damage/quantity. 

I currently use the following Forge code to check to see if my ItemStacks have the same variant, with a disregard to the damage value if the item is undamageable. It would be nice to see a similar easy to use check for Sponge.

``` java

import static com.skelril.nitro.transformer.ForgeTransformer.tf;
import static net.minecraft.item.ItemStack.areItemStackTagsEqual;
import static net.minecraft.item.ItemStack.areItemsEqual;

public class ItemComparisonUtil {
    public static boolean isSimilar(ItemStack a, ItemStack b) {
        net.minecraft.item.ItemStack stackA = tf(a);
        net.minecraft.item.ItemStack stackB = tf(b);

        if (stackA.isItemStackDamageable()) {
            return a.getItem() == b.getItem() && areItemStackTagsEqual(stackA, stackB);
        }

        return areItemsEqual(stackA, stackB) && areItemStackTagsEqual(stackA, stackB);
    }
}
```

Faithcaio:
fixes #1257 

For API 8 it would be great to move the impl of these to Common.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
@@ -29,9 +29,11 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.spongepowered.api.data.manipulator.mutable.entity.DamageableData;
 import org.spongepowered.api.item.ItemType;
 
 import java.util.Comparator;
+import java.util.Iterator;
 import java.util.List;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
```diff
@@ -160,11 +186,21 @@ public int compare(@Nullable final ItemStack o1, @Nullable final ItemStack o2) {
...
-                } else {
+                } else if (!isIgnored(manipulators, manipulator)) {
                     return -1;
                 }
             }
             return manipulators.size();
         }
+
+        private boolean isIgnored(List<DataManipulator<?, ?>> list, DataManipulator<?, ?> toCheck) {
+            for (Class<? extends DataManipulator<?, ?>> ignore : ignored) {
```
kashike: `this.`

kashike (changes requested):

gabizou (approved):

PR #1766, opened 2018-02-04
[API 8] Refactoring of LaunchProjectileEvent - Removal of TargetProjectileEvent

RedNesto:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1769)

This is a breaking change.

This PR removes the TargetProjectileEvent and changes the LaunchProjectileEvent.

Now, the event is an AffectEntityEvent: the point is to make the event handle several projectiles at the same time, which is possible in mods (like the Thermal Expansion Multishot enchantment).


src/main/java/org/spongepowered/api/event/entity/projectile/LaunchProjectileEvent.java
```diff
@@ -24,12 +24,31 @@
...
+import org.spongepowered.api.event.entity.AffectEntityEvent;
+import org.spongepowered.api.event.entity.ConstructEntityEvent;
+import org.spongepowered.api.event.entity.SpawnEntityEvent;
+import org.spongepowered.api.util.annotation.eventgen.PropertySettings;
+
+import java.util.List;
 
 /**
- * Called when a {@link Projectile} is launched.
+ * Called when a {@link Projectile} is launched. This event is always followed by a {@link SpawnEntityEvent}
```
gabizou, line 37: Why would this be followed by a spawn entity event and not just have this extend spawn entity event?

gabizou (commented):
Some clarification I think is needed for the change and the accuracy of how the implementation is handled.

src/main/java/org/spongepowered/api/event/entity/projectile/LaunchProjectileEvent.java
RedNesto: I did this because in the Minecraft code, the entity is first created, then is launched, and is finally spawned in the world, but I don't see anything which would prevent us to make LaunchProjectileEvent extend SpawnEntityEvent.

JBYoshi:

PR #1768, opened 2018-02-05
GenericArguments and CommandFlags Fixes

SimonFlash:
This PR fixes a couple issues within `GenericArguments` and `CommandFlags`, particularly with tab completion.

### UserCommandElement

Previously, the `UserCommandElement` would delegate it's `returnSource` value to the `PlayerCommandElement` which checked for players. As a result, if the name of an offline user was entered it tried to return the `CommandSource` instead, which resulted in the `UserCommandElement` never attempting to parse a `User` from the given name.

### SequenceCommandElement

The `SequenceCommandElement` could not handle tab completion well for optional-like elements (`OptionalCommandElement`, `UserCommandElement` returning the source, `FlagsElement`, etc.) because any `ArgumentParseException` thrown was caught and handled by that element. Now, if the state of the args is the same before and after parsing an element, the completions for those will be cached in a `Set` to be returned later. If an element is parsed successfully, this cache is cleared.

### CommandFlags

Much of `CommandFlags` has been cleaned up and simplified, placing a large priority on fixing tab completion. I would also like to look into a `tabCompleteUnusedFlags` option that will tab complete for flags that haven't been used yet, although this will take a bit of work doing.

### Minor Changes

* `MarkTrueCommandElement` now uses a `Text` instead of a `String` key, and has been made `public`
* `OptionalCommandElement#parseValue` had a reversed `args.hasNext()` check
* The `PlayerCommandElement` key is now `@Nullable`
* Updated `CommandFlags.Builder` functions to remove `@Nullable` and use lambdas

### Review & Testing

I have not begun testing these with Sponge yet, but these changes come from those that I have tested within TeslaLibs. I will begin setting up Sponge tests tomorrow, and any help testing would be much appreciated!

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
@@ -96,16 +96,17 @@ public final class CommandFlags extends CommandElement {
 
     private boolean parseLongFlag(CommandSource source, String longFlag, CommandArgs args, CommandContext context) throws ArgumentParseException {
         String[] flagSplit = longFlag.split("=", 2);
-        CommandElement element = this.longFlags.get(flagSplit[0].toLowerCase());
+        String flag = flagSplit[0].toLowerCase();
+        CommandElement element = this.longFlags.get(flag);
         if (element == null) {
             switch (this.unknownLongFlagBehavior) {
                 case ERROR:
                     throw args.createError(t("Unknown long flag %s specified", flagSplit[0]));
                 case ACCEPT_NONVALUE:
-                    context.putArg(flagSplit[0], flagSplit.length == 2 ? flagSplit[1] : true);
+                    context.putArg(flag, flagSplit.length == 2 ? flagSplit[1] : true);
                     return true;
                 case ACCEPT_VALUE:
-                    context.putArg(flagSplit[0], flagSplit.length == 2 ? flagSplit[1] : args.next());
+                    context.putArg(flag, flagSplit.length == 2 ? flagSplit[1] : args.next());
                     return true;
                 case IGNORE:
                     return false;
@@ -221,14 +222,14 @@ public final class CommandFlags extends CommandElement {
         Object state = args.getState();
         try {
             element.parse(src, args, context);
+            return null;
         } catch (ArgumentParseException ex) {
             args.setState(state);
-            String prefix = flagSplit.length == 2 ? flagSplit[0] + "=" : args.hasNext() ? "" : flagSplit[0] + " ";
+            String prefix = flagSplit.length == 2 ? flagSplit[0] + "=" : "";
             return element.complete(src, args, context).stream()
                     .map(s -> prefix + s)
                     .collect(Collectors.toList());
         }
-        return null;
     }
 
     @Nullable
--- a/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
@@ -35,7 +35,6 @@ import org.spongepowered.api.text.Text;
 import org.spongepowered.api.util.StartsWithPredicate;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -71,100 +70,75 @@ public final class CommandFlags extends CommandElement {
 
     @Override
     public void parse(CommandSource source, CommandArgs args, CommandContext context) throws ArgumentParseException {
-        Object startIdx = args.getState();
-        String arg;
+        Object state = args.getState();
         while (args.hasNext()) {
-            arg = args.next();
-            boolean ignore;
+            String arg = args.next();
+            boolean remove;
             if (arg.startsWith("-")) {
-                Object flagStartIdx = args.getState();
+                Object start = args.getState();
                 if (arg.startsWith("--")) { // Long flag
                     String longFlag = arg.substring(2);
-                    ignore = !parseLongFlag(source, longFlag, args, context);
+                    remove = parseLongFlag(source, longFlag, args, context);
                 } else {
                     arg = arg.substring(1);
-                    ignore = !parseShortFlags(source, arg, args, context);
+                    remove = parseShortFlags(source, arg, args, context);
                 }
-                if (!ignore) {
-                    args.removeArgs(flagStartIdx, args.getState());
+                if (remove) {
+                    args.removeArgs(start, args.getState());
                 }
             } else if (this.anchorFlags) {
                 break;
             }
         }
-
-        args.setState(startIdx);
+        args.setState(state);
         if (this.childElement != null) {
             this.childElement.parse(source, args, context);
         }
-
     }
 
     private boolean parseLongFlag(CommandSource source, String longFlag, CommandArgs args, CommandContext context) throws ArgumentParseException {
-        if (longFlag.contains("=")) {
-            final String[] flagSplit = longFlag.split("=", 2);
-            longFlag = flagSplit[0];
-            String value = flagSplit[1];
-            CommandElement element = this.longFlags.get(longFlag.toLowerCase());
-            if (element == null) {
-                switch (this.unknownLongFlagBehavior) {
-                    case ERROR:
-                        throw args.createError(t("Unknown long flag %s specified", args));
-                    case ACCEPT_NONVALUE:
-                    case ACCEPT_VALUE:
-                        context.putArg(longFlag, value);
-                        break;
-                    case IGNORE:
-                        return false;
-                    default:
-                        throw new Error("New UnknownFlagBehavior added without corresponding case clauses");
-                }
-            } else {
-                args.insertArg(value);
-                element.parse(source, args, context);
+        final String[] flagSplit = longFlag.split("=", 2);
+        CommandElement element = this.longFlags.get(flagSplit[0].toLowerCase());
+        if (element == null) {
+            switch (this.unknownLongFlagBehavior) {
+                case ERROR:
+                    throw args.createError(t("Unknown long flag %s specified", flagSplit[0]));
+                case ACCEPT_NONVALUE:
+                    context.putArg(longFlag, flagSplit.length == 2 ? flagSplit[1] : true);
+                    return true;
+                case ACCEPT_VALUE:
+                    context.putArg(longFlag, flagSplit.length == 2 ? flagSplit[1] : args.next());
+                    return true;
+                case IGNORE:
+                    return false;
+                default:
+                    throw new Error("New UnknownFlagBehavior added without corresponding case clauses");
             }
-        } else {
-            CommandElement element = this.longFlags.get(longFlag.toLowerCase());
-            if (element == null) {
-                switch (this.unknownLongFlagBehavior) {
-                    case ERROR:
-                        throw args.createError(t("Unknown long flag %s specified", args));
-                    case ACCEPT_NONVALUE:
-                        context.putArg(longFlag, true);
-                        break;
-                    case ACCEPT_VALUE:
-                        string(Text.of(longFlag)).parse(source, args, context);
-                        break;
-                    case IGNORE:
-                        return false;
-                    default:
-                        throw new Error("New UnknownFlagBehavior added without corresponding case clauses");
-                }
-            } else {
-                element.parse(source, args, context);
-            }
+        } else if (flagSplit.length == 2) {
+            args.insertArg(flagSplit[1]);
         }
+        element.parse(source, args, context);
         return true;
     }
 
     private boolean parseShortFlags(CommandSource source, String shortFlags, CommandArgs args, CommandContext context) throws ArgumentParseException {
-        for (int i = 0; i < shortFlags.length(); ++i) {
-            final String flagChar = shortFlags.substring(i, i + 1);
-            CommandElement element = this.shortFlags.get(flagChar);
+        for (int i = 0; i < shortFlags.length(); i++) {
+            String shortFlag = shortFlags.substring(i, i + 1);
+            CommandElement element = this.shortFlags.get(shortFlag);
             if (element == null) {
                 switch (this.unknownShortFlagBehavior) {
                     case IGNORE:
                         if (i == 0) {
                             return false;
                         }
-                        throw args.createError(t("Unknown short flag %s specified", flagChar));
+                        throw args.createError(t("Unknown short flag %s specified", shortFlag));
                     case ERROR:
-                        throw args.createError(t("Unknown short flag %s specified", flagChar));
+                        throw args.createError(t("Unknown short flag %s specified", shortFlag));
                     case ACCEPT_NONVALUE:
-                        context.putArg(flagChar, true);
+                        context.putArg(shortFlag, true);
                         break;
                     case ACCEPT_VALUE:
-                        string(Text.of(flagChar)).parse(source, args, context);
+                        string(Text.of(shortFlag)).parse(source, args, context);
                         break;
                     default:
                         throw new Error("New UnknownFlagBehavior added without corresponding case clauses");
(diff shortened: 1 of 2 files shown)


dualspiral:
I’ve not had much time to look over this, but with all of your minor changes, why? They either indicate a change in behaviour or just do nothing, as far as I can see.

I’ll investigate other points when I get a little more time.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -631,31 +634,26 @@ public static CommandElement optionalWeak(CommandElement element, Object value)
...
+            if (this.element.getKey() != null && this.value != null) {
+                context.putArg(this.element.getUntranslatedKey(), this.value);
+            }
         }
 
+        @Nullable
         @Override
         protected Object parseValue(CommandSource source, CommandArgs args) throws ArgumentParseException {
-            return args.hasNext() ? null : this.element.parseValue(source, args);
+            return args.hasNext() ? this.element.parseValue(source, args) : this.value;
```
dualspiral: What does this achieve? I don't think this is even called.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -71,100 +70,75 @@ protected CommandFlags(@Nullable CommandElement childElement, Map<List<String>,
 
     @Override
     public void parse(CommandSource source, CommandArgs args, CommandContext context) throws ArgumentParseException {
-        Object startIdx = args.getState();
```
dualspiral, line 74: There was _zero_ need to change this method, all you've done is change variable names and flipped a boolean.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -71,100 +70,75 @@ protected CommandFlags(@Nullable CommandElement childElement, Map<List<String>,
...
+        CommandElement element = this.longFlags.get(flagSplit[0].toLowerCase());
+        if (element == null) {
+            switch (this.unknownLongFlagBehavior) {
+                case ERROR:
+                    throw args.createError(t("Unknown long flag %s specified", flagSplit[0]));
+                case ACCEPT_NONVALUE:
+                    context.putArg(flagSplit[0], flagSplit.length == 2 ? flagSplit[1] : true);
+                    return true;
+                case ACCEPT_VALUE:
+                    context.putArg(flagSplit[0], flagSplit.length == 2 ? flagSplit[1] : args.next());
```
dualspiral: For reference, `string(Text.of(longFlag)).parse(source, args, context);` became `args.next();`

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -211,120 +185,83 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
...
+                    .collect(Collectors.toList());
+        } else if (flagSplit.length == 2) {
+            args.insertArg(flagSplit[1]);
+        }
+        Object state = args.getState();
+        try {
+            element.parse(src, args, context);
+        } catch (ArgumentParseException ex) {
+            args.setState(state);
+            String start = flagSplit.length == 2 ? flagSplit[0] + "=" : args.hasNext() ? "" : flagSplit[0] + " ";
```
dualspiral: If you're nesting a ternary in a ternary, then you should be splitting this out into if statements, it took me a moment to realise what this was.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -120,8 +122,14 @@ public static CommandElement none() {
...
 
-    static CommandElement markTrue(String flag) {
-        return new MarkTrueCommandElement(flag);
+    /**
+     * Expects no arguments. Adds 'true' to the context when parsed.
+     *
+     * @param key the key to store 'true' under
+     * @return the argument
+     */
+    public static CommandElement markTrue(Text key) {
```
dualspiral, line 131: Why does this need to be public?

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -312,40 +320,35 @@ public void parse(CommandSource source, CommandArgs args, CommandContext context
...
+                    } else if (!completions.isEmpty()) {
+                        completions.clear();
+                    }
                     if (!args.hasNext()) {
-                        args.setState(startState);
-                        List<String> inputs = element.complete(src, args, context);
-                        args.previous();
-                        if (!inputs.contains(args.next())) { // Tabcomplete returns results to complete the last word in an argument.
-                            // If the last word is one of the completions, the command is most likely complete
-                            return inputs;
```
dualspiral, line 334: Don't remove this behaviour.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -631,31 +634,26 @@ public static CommandElement optionalWeak(CommandElement element, Object value)
...
+                } catch (ArgumentParseException ex) {
+                    if (!this.considerInvalidFormatEmpty && !args.hasNext()) {
+                        throw ex;
                     }
-                } else {
-                    throw ex;
+                    args.setState(start);
                 }
             }
+            if (this.element.getKey() != null && this.value != null) {
```
dualspiral: Because there is no return after the parsing, you're now _always_ adding the fallback value to the context, regardless of whether you've parsed something successfully or not.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1125,7 +1123,7 @@ public Text getUsage(CommandSource src) {
 
         protected UserCommandElement(@Nullable Text key, boolean returnSource) {
             super(key);
-            this.possiblePlayer = new PlayerCommandElement(key, returnSource);
+            this.possiblePlayer = new PlayerCommandElement(key, false);
```
dualspiral: This change is good.

dualspiral (changes requested):
You've rearranged a lot of methods when they didn't need to be rearranged, and in the process you've actually introduced a bug.

When creating a PR, a minimal diff is a lot more helpful because we can then see what you've done and trace why you wanted to make the change. These rearrangments are not solutions to the original problem and worked fine. 

If this was sticking around then a cleanup might have been a good thing, but considering they are scheduled for removal in API 8 they will get blown away anyway.

I don't see how this helps tab completion for flags - it looks like the method is just rearranged to me. If you could point out your changes for that, it would help.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
SimonFlash: All of these changes are coming from another project, so the changes here were primarily to make it clear to me what's happening. It took me a minute to understand why the boolean `ignore` was being double negated here, and to confirm that I had to read through both of the parse methods. I can revert this if need be, but to me this shows the intent at a glance much more clearly.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
SimonFlash: Yep, which is why it's fine to do this simplification. There isn't really a need to create a new `CommandElement` for this as it only hurts performance.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
SimonFlash: Makes sense. This is one of the major fixes that prevents tab completion from overwriting the inputted flag.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
SimonFlash: I wanted it personally because I was re-implementing flags and had to copy this class because it was under default access. I would much rather refactor this into an element that inputs a given constant value when parsed as I think there's more use in that (and it's pretty simple to do).

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
SimonFlash: This is similar to the behavior fixed in #1660. Let's presume we are tab completing a `User` and that users by the names of `dualspiral` and `dual` exists. With this system, entering `dual` and completing will end up completing the *next* element, not to `dualspiral` as I might expect. I removed this because it makes an assumption that the argument is complete rather than letting the user decide that for themselves.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
SimonFlash: Good catch!

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
SimonFlash: This is a gross simplification over what it should be, as really it should be catching any thrown exception to have the same process as in `#parse`. I would personally prefer having things like this throw an `UnsupportedOperationException`, but it's too far along to change that now. This just brings it close to the expected behavior as it shouldn't be returning `null` if it has a next argument.

SimonFlash:
These changes are being migrated from another projects, but I did try to minimize any external changes that weren't focused on the aspect I was working on. In the case of flags, the primary fix is preventing tab completion from overwriting the inputted flag. I would also like to have some system in place that filters the available flags based on what has been inputted already. This would likely require a new option within the builder because the reuse of a flag is allowed by the specifications.

If you want me to revert the unnecessary changes, please confirm that for me.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
dualspiral: This then breaks the behaviour of all custom elements that don't use pattern matching and potentially use this behaviour. We should not break expected behaviour.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
dualspiral: Yes, sorry - it was a note to me to say where the diff really is, I should have made that clear.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
dualspiral: No, my point is there was no point in changing this method because the method `parseValue`, a protected method in a private class so cannot be overriden on the outside, is never called. This probably _should_ throw an exception of some kind, but far better just to leave it and ignore it for the purposes of this PR because it's harmless.

src/main/java/org/spongepowered/api/command/args/GenericArguments.java
SimonFlash: I'm not sure if this really is expected behavior though, and this change shouldn't break anything. If a user tries to tab complete an argument, I don't think it should try to complete the next one instead.

For example, with a command `/test <user> <item>`, this is what currently happens when tab completing:
* `/test dual` -> All items
* `/test dual (space)` -> All items

This looks nice, but is an issue if the sender was trying to tab complete `dual` to `dualspiral` in the first one; not complete all the items. Instead, I think that this is the expected behavior:
* `/test dual` -> `dual, dualspiral`
* `/test dual (space)` -> All items

dualspiral: OK, fair enough, I see what you're getting at now. The space should be a necessary delimiter.

#1660 will supersede part of this. I need to go back to it, test it and merge it, it was built when debugging something for blood then I got busy with other things. If I get to that before this gets sorted, just bear in mind that fix will need to be removed.

dualspiral:
It's done now but I am wary that by doing so that some behaviour changes may have been introduced that should not have been. Particularly for this, this is in the API project, so **changes** to stable should be **as minimal as possible**, which is ultimately where you'd want this to go.

Can you please explain what the flag tab completion problems are, why they are caused and how you fixed them? It's hard to make it out because of these diffs.

~wip

SimonFlash:
Previously, if you tab complete a flag that has an existing alias (similar to user thing discussed above) it would tab complete the element of the flag *without* attaching the flag to the result, so `--user` would tab complete `Simon_Flash` instead of `--user Simon_Flash` (or `--user=Simon_Flash`).

This was fixed in the first PR in [this line](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-bddcca313ee050f07039ec6b24e4653eR226), but I realized this is also incorrect as we're assuming the sender wants to tab complete the value of the flag and not the flag itself (same thing as with the user example above). The proper fix for that is in [this line](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-bddcca313ee050f07039ec6b24e4653eR213), which now also returns the filtered keys if we have a single flag without a value that does not have a next argument.

However, there are three other changes that I would like to make for this:
* Tab completing `""` (empty) should include all flags, not none.
* Tab completing `-` should include all flags, not only short flags.
* When tab completing a flag, only flags that are unused are included. Because the specifications for `CommandFlags` allow a flag to be used multiple times, this would be implemented as an option set within the builder with something like `setFilterUsedFlags`, defaulting to `false`.

The gif below demonstrates how these might work (particularly the last one); just note that in it long flags only have a single hyphen instead of two.

Filter Used Flags

ryantheleach:
I'll need time to review this in depth, but when I quickly skimmed it before dual reviewed, I had the same thoughts about not being able to see the direct changes due to not being a minimal diff, but didn't write it as I wanted to look in more depth and give some positive constructive feedback as well.

That said, this looks like a better contribution then the color Orange rabbit hole I led you down, So I'm glad to see you still contributing <3

"just note that in it long flags only have a single hyphen instead of two."

What do you mean by this? Is the gif representative of this PR (which changes the long flags to single hyphens)? Or is it using your alternate command system(and is just used for demonstration purposes, because you believe this PR will act similarly)?

SimonFlash:
Yes, some of the changes weren't really needed. I've reverted most of those, but I've kept the changes I made to the parsing of `CommandFlags` because it does make the code more readable (in my opinion) and was necessary for me to verify that it had the correct functionality that I was looking for. Without getting too off-topic, I have no complaints regarding the orange PR except I _still_ don't know what people want me to do with it.

I have a difficult time explaining thing (especially in text), so I included the gif to try to show what my final goal is. I wrote a flag parsing system into TeslaLibs [here](https://github.com/SimonFlash/TeslaPowered/blob/feature/arguments/TeslaLibs/src/main/java/com.mcsimonflash.sponge.teslalibs/command/arguments/FlagsElement.java) a couple days ago to fix some of the issues I had personally with `CommandFlags`, and this PR adds whatever fixes I can from there into Sponge. That's the system shown in the gif, and was only included to show what I meant by the tab completion pieces. For compatibility reasons, there are no changes made to how long flags work in this PR.

SimonFlash:
@dualspiral I just discovered an issue with tab completing for `PatternMatchingCommandElement`s, but I'm not sure how to fix it properly. 

When we're parsing an element such as a `UserCommandElement`, it attempts to get a value from the `PlayerCommandElement` first. However, unless you've typed the name of a player incorrectly this parsing will always succeed and ends up returning a list of all players on the server that match the argument. As a result, the parsing never fails and thus we don't know to complete the element. Here's a couple examples of this:

* `""` -> Parses all *players* into the context
* `"S"` -> Parses all *players* starting with `S` into the context
* `"Sin` -> Fails parsing. Tab completion would return any users starting with `Sin`, often resulting in an empty list.

This should be fixed as a result of the `!args.hasNext()` check [here](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-325508c3a8ee152cf22a18a16566a37R340) in `SequenceCommandElement`, and I should be doing the same [here](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-bddcca313ee050f07039ec6b24e4653eR223) in `CommandFlags`. However, I'm not sure if this is a proper fix or just a workaround as the parsed values for `UserCommandElement` are incorrect as well.

dualspiral:
I think that I solved this in #1660 by adding a completion context and moving the check for this element. Let me look at getting that in tonight (UK time), I should have some time this evening.

ryantheleach:
@SimonFlash I can't remember the exact reason for that behavior (players first) But I remember it being wanted, since 1. it was more efficient, and 2, people expected to be able to tab complete the online players first, because they were online. If it's failing to parse beyond that, then it's possible I made a mistake, or it's not the Command Element that I remembering editing. The git blame on that should hold a PR, with an issue with additional details if that's the case.

SimonFlash:
Status update on this:

* Waiting for #1660 so I can update accordingly. I did make an attempt to merge those changes with mine and do some testing with it, but was unsuccessful.
* Thoughts on converting `GenericArguments#markTrue` to something more like `GenericArguments.constant` that accepts a key and returns a constant value? Has the same functionality when `true` is passed in, but has a bit more uses and justifies making it public.
* Filtering out already defined flags from suggestions (see [this comment](https://github.com/SpongePowered/SpongeAPI/pull/1768#issuecomment-363245170)). This would be a bit of extra work and would require an additional setting to be defined in the builder, but shouldn't be a breaking change.

SimonFlash:
@dualspiral This has passed all of my tests and is ready for review/merge.

dualspiral:
> This has passed all of my tests

Do you have reduced test cases that we could use to see what exactly you've tested? It would be good for me to be able to look at a before/after. If you have any unit tests, that would be useful too.

I'm struggling to see what you've functionally changed and what formatting changes you've made so being able to understand what's going on by investigating the test cases would be much easier for me.

phit:
https://github.com/SpongePowered/SpongeAPI/pull/1768/files?w=1
ignoring whitespace helps a bit

SimonFlash:
Consider the following `CommandElement`s, which simply represent three arguments that are optional and their suggestions, and the `SequentialElement` created using the old logic compared to the new logic in this PR.

```java
CommandElement first = optionalWeak(choices(Text.of("first"), ImmutableMap.of("arg", "arg", "arg1", "arg1", "test1", "test1")));
CommandElement second = optionalWeak(choices(Text.of("second"), ImmutableMap.of("arg", "arg", "arg2", "arg2", "test2", "test2")));
CommandElement third = optionalWeak(choices(Text.of("third"), ImmutableMap.of("arg3", "arg3", "arg4", "arg4")));

CommandElement oldSeq = oldSeq(first, second, third);
CommandElement newSeq = newSeq(first, second, third);
```
Here's what happens when you complete these with the following inputs:

* `""` (empty)
  * OldSeq: `[]`
  * NewSeq: `[arg, arg1, arg2, arg3, arg4, test1, test2]`
* `"a"`
  * OldSeq: `[]`
  * NewSeq: `[arg, arg1, arg2, arg3, arg4]`
* `"arg"`
  * OldSeq: `[arg3, arg4]`
  * NewSeq: `[arg, arg1]`

So, what's happening here? The largest issue that this addresses is that the old `SequenceElement` skips over optional arguments because they are technically parsed successfully. To fix this, I've included a check to see whether the actual state of the arguments changes - if it doesn't, then we can presume that we're dealing with some type of optional element and can add its completions to the return list.

The second issue is that the old element makes the assumption that if a completion matches the input, we should jump to the next element. This is **wrong** - as I've discussed above with flags, we should only be completing the active element and require the user to input a space to indicate they're moving to the next one.

With the first two examples, none of the arguments are matched and so the `ChoicesElement` throws an exception, which is caught by `OptionalWeak`. As the parse was 'successful', the old sequence element moves on and eventually swallows the remaining elements without actually completing any of them. My changes make this system a bit smarter, and the state check realizes that we're probably dealing with an optional element and completes them.

In the last case, this is caused by the weird jumping and resetting. I'm honestly not 100% sure what the old sequence element is doing here, but it seems to parse `arg` twice for the first and second elements before eventually completing the third. The new sequence element, however, sticks with the first element as the parse was successful and returns it's completions. While the most correct response is to also complete the remaining elements, this is a limitation of the current command system because it is impossible to identify an element as being optional if an argument is consumed. As such, there is no way we can return the expected `[arg, arg1, arg2, arg3, arg4]` completions without making changes to the `CommandElement` class to indicate when arguments are optional.

Though the test case is quite odd, it is actually something that is seen commonly on a smaller scale. With these changes, any arguments such as `[player-or-source] <something>` will both be tab completed properly.

With flags, the changes to tab completion is with regards to this second issue. This ensures that tab completing `--fl` will show `[--fl, --flag]` instead of completing the element attached to the `--fl` element (poor example, but that's the gist of things). The focus here is requiring the using the specify when they're moving to the next argument rather than forcing it on them if there's a partial match.

dualspiral:
OK, so I've pulled this into my workspace and written a unit test or two to verify things. It doesn't pass all of my tests (and one of yours). This is what I see:

* **UserCommandElement**: not there - was fixed in #1660 and later commits so isn't needed as part of this.
* **SequenceCommandElement**: [I wrote this unit test](https://gist.github.com/dualspiral/331f1d1778aa3f9e883a7f16a657b254) to verify that this is the case, includes your tests and some other cases I thought of. Some are failing.
* **CommandFlags**: you're not accounting for optional elements or complete short flags by the looks of it - short flags probably don't need completion on their own to be fair but it's included in the test for sake of completion, happy to ignore that one. Others should be considered for consistency.

Minor changes:

* `MarkTrueCommandElement` now uses a `Text` instead of a `String` key, and has been made `public` - OK
* `OptionalCommandElement#parseValue` had a reversed `args.hasNext()` check - not there now.
* The `PlayerCommandElement` key is now `@Nullable` - not there now.
* Updated `CommandFlags.Builder` functions to remove `@Nullable` and use lambdas - OK.

SimonFlash:
#### UserCommandElement

I'm not sure if you're saying there's something I should change here or not, so let me know what you're looking for here. If there's something that needs to be redone or reversed, I'm going to need help as I don't understand Git well enough to do it myself.

#### SequenceCommandElement

I added the tests into my workspace and here were the initial results. Notably, the only test cases that fail with my system also fail with the old one anyways.

- My sequential element failed tests 1, 15, 16, and 17.
- The old sequential element failed tests 1, 2, 3, 6, 9, 10, 12, 13, 15, 16, and 17.

In the process of fixing test case 1, I found that tests 5, 7, and 8 are incorrect. Each of these makes the assumption that if the element is parsed successfully using an argument then it should complete the next element, which is not true. The completions of this element should be included through the next non-optional argument - this is the change that resolves test 1.

I've forked your gist of the tests above and added my edits [here](https://gist.github.com/SimonFlash/3f917303a73fd0712fc44aeea9d42ff6).

#### CommandFlags

Applying optional elements to flags is outside the scope of this PR, and as such I didn't worry about tests 15, 16, and 17. I wouldn't expect people to try using optional elements with flags as they're optional by nature anyways (though wrapping it with the new sequential argument could work). There are a number of fringe cases that could be addressed, but I'm not willing to put time into that because the system really needs to be rebuilt anyways. The largest problem with this is that there is no method to detect whether an element is optional or not; we can only make guesses based on whether the state of the arguments changes and that can't account for an optional element that's consumed anyways. I'm more than happy to put time into resolving these problems, but ultimately this is a project for API 8 that requires a redesign to include the necessary information to make this possible.

dualspiral:
Tests 5 6 and 7 mirrored what you said in your previous comment, so if you've fixed that, brilliant!

What I meant about the UserCommandElement was that I did it earlier, so it's not needed to be present here. My bad.

>  I wouldn't expect people to try using optional elements with flags as they're optional by nature anyways

I get what you're saying, but it is possible because there might be a sensible default associated with a flag where the flag changes the meaning of the command anyway. But I also understand your point about the work it is. However, as API 7 could be around for another year yet (as command changes will be to work with Brigadier for 1.13), it might be worth looking at it. It can be a future PR though.

To be honest, let's get this checked and in and worry about that later. I need to test this before making any more comments - I'll do the next time I get a chance to.

dualspiral:
I ran the code on my server and found some odd things happening with flag completion. So, I added some more tests and properly tested what was coming back from a completion (not just a count), and there were some errors that occurred:

[I added the tests here](https://gist.github.com/dualspiral/331f1d1778aa3f9e883a7f16a657b254#file-completiontest-java-L106-L108) along with the one on line 103 (`-f arg4`), which all failed. That latter one is an edge case, but the first three were broken when you rearranged some of the code.

Specifically, [you missed the two dashes here](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-bddcca313ee050f07039ec6b24e4653eL264) which [should have also been here](https://github.com/SpongePowered/SpongeAPI/pull/1768/files#diff-bddcca313ee050f07039ec6b24e4653eR226).

This is why a minimal diff for fixes and changes are preferred, because everyone missed it right up until the point I was doing final tests!

As it's a minor thing and I want to get this all in tonight, I'll make the changes. I've also updated the short flag completion code so that it properly handles elements that are complete and doesn't make them disappear on tab - my diff is here: https://gist.github.com/dualspiral/25fa09995a791972f2deca7f786457c1.

dualspiral:
https://github.com/SpongePowered/SpongeAPI/commit/675baab9b722abde24170903d043377d4585462e

This is currently only on bleeding, will cherry-pick in a bit.

PR #1769, opened 2018-02-06
[API 8 ?] Added/clarified documentation on arrows - Fix SpectralArrow

RedNesto:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1776)

This PR adds/changes the javadocs of the Arrows projectiles. It also makes the SpectralArrow extend Arrow.
Due to this last point, this may be considered as a breaking change, @Zidane and I would like to know if this can be merged in stable-7 or should be in bleeding.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/projectile/arrow/Arrow.java
+++ b/src/main/java/org/spongepowered/api/entity/projectile/arrow/Arrow.java
@@ -30,7 +30,10 @@ import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
 import org.spongepowered.api.entity.projectile.DamagingProjectile;
 
 /**
- * Represents an Arrow.
+ * Base interface representing an Arrow. You should not launch an Arrow, but instead on of its sub-interface (like {@link TippedArrow}).
+ *
+ * @see TippedArrow
+ * @see SpectralArrow
  */
 public interface Arrow extends DamagingProjectile {
 
--- a/src/main/java/org/spongepowered/api/entity/projectile/arrow/SpectralArrow.java
+++ b/src/main/java/org/spongepowered/api/entity/projectile/arrow/SpectralArrow.java
@@ -24,6 +24,11 @@
  */
 package org.spongepowered.api.entity.projectile.arrow;
 
-public interface SpectralArrow {
+import org.spongepowered.api.effect.potion.PotionEffectTypes;
+
+/**
+ * An {@link Arrow} applying the {@link PotionEffectTypes#GLOWING} effect on hit.
+ */
+public interface SpectralArrow extends Arrow {
 
 }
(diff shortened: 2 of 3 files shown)


RedNesto:
~qa

Zidane:
@gabizou @SpongePowered/developers 

This brings to mind the issue (as evidenced by the common PR) where we allow any Projectile to be launched or any Entity to be spawned and I also feel that we need a marker interface to solve the issue where you clearly can't launch an Arrow because in the MCP implementation Arrow is an abstract class.

I'm aware this leaks implementation details but its making for muddy usage as well because the act of allowing a spawnEntity(Entity e) or launch(Projectile p) gives the impression that it is possible to do so for everything.

Thoughts?

Cybermaxke:
If we would use `EntityType`s to spawn entities, this wouldn't be a issue. But then I would also make the `EntityType` class support generics. `EntityType<SpectralArrow>`, then you can use the generics to check if it's a projectile and use it to return the entity as a `SpectralArrow`.

gabizou:
Alternatively you can make the method to spawn an entity duck typed as such:
```java
<E> E createEntity(EntityType type, Vector3i pos);

final Arrow arrow = createEntity(EntityTypes.SPECTRAL_ARROW, pos);
```

While I agree with @Zidane that the overall exposure of the structure of entity classes, this was already done when we had to move away from skeletons and use `WhiteSkeleton`s etc. Likewise with horses. At this point, it's safe to say the exposure is required, but also that we can duck type the method itself to do as above mentioned, and the plugin developer would have to know that they are getting their hands dirty themselves. (The usage of duck typing here eliminates the need for generics on EntityType, and falls in line with the usages in Inventory API).

Cybermaxke:
@gabizou While that may work for creating entities, using a `EntityType` to launch projectiles allows you to pass in any entity, it doesn't even have to be a projectile. That's why the generic parameter for EntityType could be useful:
`<T extends Projectile> T launchProjectile(EntityType<T> projectileType);`

gabizou:
In some cases though, an entity is considered a projectile but isn’t mapped as a projectile in the API. I believe an example could be mod entities from bosses, like wither skulls or the equivalent.

amaranth:
Why does `SpectralArrow` extend `Arrow` instead of `TippedArrow`? Also, why does `SpectralArrow` even exist if you have `TippedArrow`? And this is definitely a breaking change although I doubt it'll actually break anything since `SpectralArrow` seems like it was unusable before this.

RedNesto:
`SpectralArrow` is just an `Arrow` since it can only give one special effect (glowing), unlike `TippedArrow`s which can have (or not) potion effects.

amaranth:
Huh, so you can have a `TippedArrow` with the glowing potion effect and the client will call it an "Arrow of Glowing" rather than a "Spectral Arrow"?

gabizou:
I believe this PR is good to go in stable-7 since it's not a breaking change, more of a fix. While it wouldn't exist in the major version release, it will be released in the minor version release of the API.

As far as spawn methods go, I believe the discussion should continue in a separate issue.

gabizou (approved):

parlough:
Closed by 7c8f70832ccb956cdb279b826bddf903847760a3 and dc5a4e736a2b2d8110be9f1684a6d8ecce1bc964

PR #1770, opened 2018-02-06
Move recipe registration and all recipes will be catalog types.

Cybermaxke:
* Moves the `SmeltingRecipe` and `CraftingRecipe` registration to the `GameRegistryEvent.Register` event.
* `CraftingRecipe`s must be reloadable, because they are loaded from json files in resource (data) packs.
* Turn `SmeltingRecipe` (`Recipe`) into a `CatalogType`, it's only a matter of time before smelting recipes are also loaded from json files, better prepare for this. This also makes it consistent with the crafting recipes which are already `CatalogType`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/recipe/Recipe.java
+++ b/src/main/java/org/spongepowered/api/item/recipe/Recipe.java
@@ -24,12 +24,20 @@
  */
 package org.spongepowered.api.item.recipe;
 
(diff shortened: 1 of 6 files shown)


Cybermaxke:
~wip

kashike:
~wip

Cybermaxke:
Will be covered in https://github.com/SpongePowered/SpongeAPI/pull/1952

PR #1777, opened 2018-02-10
Fix CatalogTypes

ImMorpheus:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1792)

### DIFF
--- a/src/main/java/org/spongepowered/api/item/ItemType.java
+++ b/src/main/java/org/spongepowered/api/item/ItemType.java
@@ -29,6 +29,7 @@ import org.spongepowered.api.block.BlockType;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.text.translation.Translation;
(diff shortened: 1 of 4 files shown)


ImMorpheus:
~WIP

ImMorpheus:
~qa

ST-DDT (approved):

src/main/java/org/spongepowered/api/item/ItemType.java
```diff
@@ -105,4 +106,6 @@ default ItemStackSnapshot getTemplate() {
         return ItemStack.of(this, 1).createSnapshot();
     }
 
+    Translation getTranslation(ItemStack stack);
```
ST-DDT, line 85: Isn't there already `ItemStack#getTranslation`?

ImMorpheus: Yes. This is used by ItemStack#getTranslation to take advantage of polymorphism since you need the ItemType to translate the ItemStack.

See Item#getTranslationKey(ItemStack stack) in minecraft code for a similar approach.

JBYoshi:

PR #1785, opened 2018-02-23
Enhance keepSpawnLoaded in WorldArchetype

liach:
**SpongeAPI** | [**SpongeCommon**](https://github.com/SpongePowered/SpongeCommon/pull/1810)

Now, it uses a `Tristate` instance to indicate its keep-spawn-loaded policy instead of an arbitrary boolean or a `@Nullable Boolean`.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/WorldArchetype.java
+++ b/src/main/java/org/spongepowered/api/world/WorldArchetype.java
@@ -34,6 +34,7 @@ import org.spongepowered.api.entity.living.player.gamemode.GameMode;
 import org.spongepowered.api.entity.living.player.gamemode.GameModes;
 import org.spongepowered.api.registry.CatalogTypeAlreadyRegisteredException;
 import org.spongepowered.api.util.ResettableBuilder;
+import org.spongepowered.api.util.Tristate;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 import org.spongepowered.api.world.difficulty.Difficulty;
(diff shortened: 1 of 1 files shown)


ryantheleach:
Hey Liach, can you add some context as to why it's now a Tristate? Thanks ~qa

dualspiral:
Deprecating `doesKeepSpawnLoaded` isn't a good idea.

The point of the method is that it hides an implementation detail, what the default is. Sure, in MCP, the default is to keep the overworld loaded but not other dimensions. What about Lantern? They should not be constricted by what you put into the documentation or how Mojang does it or how plugins interpret this (I dunno, plugins might use this for something!).

Having this true/false method indicates what the server is **going** to do, without referring to the configuration. This gives the impl the freedom to implement sane defaults for them.

If you want to expose what the user set as a tristate, that's fine, but it shouldn't replace the current method because what the user set and what the server does are two different things.

liach:
@dualspiral Nope. It depends if the server already has a world with a spawn or not. If it does not, then this archetype will be spawn loaded. (e.g. replacing the current overworld)

liach:
@ryantheleach it is from a bugfix in common's stable 7 branch.

dualspiral:
Oh, archetype, right. Yeah, this is fine then.

Zidane:
Closing as WorldArcheytpe is gone in 1.13.

Thanks anyways @liach .

liach:
thanks for the info! will follow the bleeding development more closely.

PR #1786, opened 2018-02-23
ItemStackComparators to Implementation

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1811)

see https://github.com/SpongePowered/SpongeAPI/issues/1257

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
@@ -24,23 +24,16 @@
  */
 package org.spongepowered.api.item.inventory;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Ordering;
 import org.spongepowered.api.data.Property;
 import org.spongepowered.api.data.manipulator.DataManipulator;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/item/inventory/ItemStackComparators.java
```diff
@@ -50,7 +44,7 @@
      * ItemStack.equals(ItemStack) for ItemStacks with extra attached data,
      * different damage values, or different sizes.
      */
-    public static final Comparator<ItemStack> TYPE = new Type();
+    public static final Comparator<ItemStack> TYPE = DummyObjectProvider.createFor(Comparator.class, "TYPE");
```
parlough: `DummyObjectProvider.createExtendedFor` for all of these should make it so you don't need your unchecked suppression.

liach (approved):

parlough (approved):

kashike (approved):

gabizou (approved):

gabizou:
This can be merged in `stable-7` since it's not a binary breakage. Since the declared types are only `Comparator` and the inner classes are package private, it isn't exposed to developers for use.

liach:
finally :tada:

PR #1787, opened 2018-02-24
Adds offline player location API

### LINKED ISSUE #1475: No way to set the location of an offline player
Currently there is only one way to alter the position of an offline player, which is by writing to the NBT's. As this is really unhandy and (IMO) unspongy, I would request another option to do this.
I've looked into the User class, which is the class that resembles an offline player. In this class there are ways to set/get the inventory,... but no way to alter the location of the player. If this is a deliberate choice, would someone explain to me why you chose not to?
Other contributions are welcome too! 😃

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1812)

fixes https://github.com/SpongePowered/SpongeAPI/issues/1475

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/User.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/User.java
@@ -24,14 +24,18 @@
  */
 package org.spongepowered.api.entity.living.player;
 
+import com.flowpowered.math.vector.Vector3d;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.manipulator.mutable.entity.StatisticData;
 import org.spongepowered.api.entity.ArmorEquipable;
+import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.Tamer;
 import org.spongepowered.api.profile.GameProfile;
 import org.spongepowered.api.service.permission.Subject;
+import org.spongepowered.api.world.Location;
 
 import java.util.Optional;
+import java.util.UUID;
 
 /**
  * A User is the data usually associated with a Player that is persisted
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/entity/living/player/User.java
```diff
@@ -70,6 +72,33 @@
...
+    Vector3d getPosition();
+
+    /**
+     * Gets the World UUID of this User.
+     *
+     * <p>May return empty when the world the player is in does not exist anymore</p>
+     *
+     * @return The World UUID of this User if found
+     */
+    Optional<UUID> getWorldUUID();
```
parlough: `getWorldUniqueId()` to match elsewhere in the API(`Identifiable`) and to fix checkstyle warning.

src/main/java/org/spongepowered/api/entity/living/player/User.java
```diff
@@ -70,6 +72,33 @@
...
+    /**
+     * Sets the position and world of this User.
+     * <p>The UUID must belong to an existing world.</p>
+     *
+     * @param position The position to set
+     * @param world The world UUID to set
+     *
+     * @throws IllegalArgumentException When the UUID does not belong to an existing world.
+     */
+    void setLocation(Vector3d position, UUID world);
```
parlough: Maybe we should name this `setPosition` to match the `getPosition`? Since it is really the position and not our version of "location".

Faithcaio: I chose the name because it also sets the world.
For online players this will literally call `setLocation(Location)`

parlough: Ah yeah I didn't consider that, makes sense to me. Keep doing your thing.

liach: I wonder if there is a way to remove the position of a user, i.e. to "unstuck" it

ST-DDT: What do you consider as unstuck?
Teleport him back to spawn?

Maybe we should state in the setters that they will delegate to Player#setX if the player is online

src/main/java/org/spongepowered/api/entity/living/player/User.java
```diff
@@ -70,6 +72,57 @@
...
+     * @param rotation The rotation to set the entity to
+     */
+    void setRotation(Vector3d rotation);
+
+    /**
+     * Gets the rotation.
+     *
+     * <p>The format of the rotation is represented by:</p>
+     *
+     * <ul><code>x -> pitch</code>, <code>y -> yaw</code>, <code>z -> roll
```
ST-DDT: Imo this needs `<li>`s.

src/main/java/org/spongepowered/api/entity/living/player/User.java
```diff
@@ -70,6 +72,57 @@
...
+     *
+     * <p>May return empty when the world the player is in does not exist anymore</p>
+     *
+     * @return The World UUID of this User if found
+     */
+    Optional<UUID> getWorldUniqueId();
+
+    /**
+     * Sets the position and world of this User.
+     * <p>The UUID must belong to an existing world.</p>
```
ST-DDT: Add an empty line in between

src/main/java/org/spongepowered/api/entity/living/player/User.java
Faithcaio: dont know why it even is a list...

liach:
how about three double versions for setter methods

Faithcaio:
I think 3 doubles would clutter the API

src/main/java/org/spongepowered/api/entity/living/player/User.java
Faithcaio: You can access player bedspawn and world spawn. Just set it accordingly.

PR #1800, opened 2018-03-30
Inventory Transformations

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1852)

predefined Transformations to order and filter inventories.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -667,6 +667,16 @@ public interface Inventory extends Iterable<Inventory>, Nameable {
     boolean containsInventory(Inventory inventory);
 
     /**
+     * Transforms this inventory using the given transformation.
+     *
+     * @param transformation The transformation
+     * @return The transformed Inventory
+     */
+    default Inventory transform(InventoryTransformation transformation) {
+        return transformation.transform(this);
+    }
+
+    /**
      * A Builder for Inventories based on {@link InventoryArchetype}s.
      */
     interface Builder extends ResettableBuilder<Inventory, Builder> {
--- a/src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
@@ -1,0 +1,87 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.query.QueryOperation;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A transformation can filter and order an inventory.
+ *
+ * <p>See {@link InventoryTransformations} for some predefined transformations</p>
+ *
+ * <p>Transformations built from the {@link Builder} use a combination of {@link Inventory#query(QueryOperation[])} and
+ * {@link Inventory#union(Inventory)}. Thus the transformed inventory will consist of only slots and will never contain duplicate slots.</p>
+ */
+public interface InventoryTransformation {
+
+    /**
+     * Transforms an Inventory
+     *
+     * @param inventory The inventory to transform
+     *
+     * @return The transformed inventory
+     */
+    Inventory transform(Inventory inventory);
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Creates a new transformation based on QueryOperations.
+     *
+     * @param operations The QueryOperations
+     *
+     * @return The new transformation
+     */
+    static InventoryTransformation of(QueryOperation... operations) {
+        if (operations.length == 0) {
+            return InventoryTransformations.NO_OP;
+        }
+        return builder().append(operations).build();
+    }
+
+    interface Builder extends ResettableBuilder<InventoryTransformation, Builder> {
+
+        /**
+         * Appends one or more {@link QueryOperation}s
+         *
+         * @param operation The operations to add
+         *
+         * @return Fluent pattern
+         */
+        Builder append(QueryOperation... operation);
+
+        /**
+         * Builds the transformation
+         *
+         * @return The built transformation
+         */
+        InventoryTransformation build();
+
+    }
+}
--- a/src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.item.inventory.entity.Hotbar;
+import org.spongepowered.api.item.inventory.entity.MainPlayerInventory;
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * Commonly used {@link InventoryTransformation}s
+ */
+public class InventoryTransformations {
+
+    /**
+     * A transformation that puts the Hotbar of a player inventory before the main inventory.
+     *
+     * <p>This transformation is equivalent to calling {@link Inventory#union(Inventory)} on the query
+     * for {@link Hotbar} with the query for {@link MainPlayerInventory} as parameter.</p>
+     */
+    public static final InventoryTransformation PLAYER_MAIN_HOTBAR_FIRST = DummyObjectProvider.createFor(InventoryTransformation.class, "PLAYER_MAIN_HOTBAR_FIRST");
+
+    /**
+     * A transformations that reverses the slot order.
+     */
+    public static final InventoryTransformation REVERSE = DummyObjectProvider.createFor(InventoryTransformation.class, "REVERSE");
+
+    /**
+     * A transformation that returns the incoming inventory without changing anything.
+     */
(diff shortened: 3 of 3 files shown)


src/main/java/org/spongepowered/api/item/inventory/Transformation.java
```diff
@@ -0,0 +1,91 @@
...
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.query.QueryOperation;
+import org.spongepowered.api.util.ResettableBuilder;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(Transformations.class)
+public interface Transformation extends CatalogType {
```
ST-DDT: I'm not sure whether this is a proper `CatalogType` based on its specs.
(You can create instances at runtime + it limits the usage to predefined instances)

src/main/java/org/spongepowered/api/item/inventory/Transformation.java
```diff
@@ -0,0 +1,91 @@
...
+         * Registers the Transformation with given id
+         *
+         * @param id The registration id
+         *
+         * @return The registered Transformation
+         */
+        Transformation build(String id);
+
+        /**
+         * Builds the Transformation with a random id
```
ST-DDT: Random ids are worthless at best and harmful at worst.
All `CatalogType`s should have meaningful ids.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/inventory/Transformation.java
Faithcaio: True. Also we dont really need the id. 
Do we have smth like this already?
Interface + Builder + some predefined instances

ST-DDT: I can't remember any occurrence of that, there is certainly no interface for that.

Just drop the interface. AFAICT you can still use the [RegistryHelper](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/registry/RegistryHelper.java)

kashike: I'd call this `InventoryTransformation` to avoid ambiguity in the future.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/inventory/Transformation.java
Faithcaio: I'll just set the fields myself in the RegistryModule

Ewoutvans (approved):

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+    public static final InventoryTransformation NONE = DummyObjectProvider.createFor(InventoryTransformation.class, "NONE");
+
+    /**
+     * A Transformation that puts the Hotbar of a player inventory before the main inventory.
+     * <p>
+     *     This transformation is equivalent to {@link Inventory#union(Inventory)} together a query
+     *     for {@link Hotbar} and {@link MainPlayerInventory} in this order
+     * </p>
+     */
+    public static final InventoryTransformation HOTBAR_FIRST = DummyObjectProvider.createFor(InventoryTransformation.class, "HOTBAR_FIRST");
```
Cybermaxke: The example is wrong in this case, since the `MainPlayerInventory` contains the main grid (3 rows) and the hotbar.

And transforming a `PlayerInventory` and `MainPlayerInventory` should give a different result. Transforming `MainPlayerInventory` would just include the main grid and hotbar where transforming the `PlayerInventory` would still include the equipment.

Cybermaxke: In lantern I am using a `View` which gives all the different kind of player inventory views used in vanilla minecraft:
https://github.com/LanternPowered/LanternServer/blob/master/src/main/java/org/lanternpowered/server/inventory/vanilla/AbstractUserInventory.java#L164
The default also prioritizes polling/offering ItemStacks to the hotbar, because it feels like the default behavior in minecraft. Not sure if it's intentional to have it the other way (priority main) in sponge, but if it is I will change it in lantern to be consistent.

Faithcaio: How is it wrong?
This is basicly your MAIN_AND_PRIORITY_HOTBAR View.
It filters out the EquipmentSlots as well as the Crafting slots in the Player Container and orders the hotbar before the grid.

Adding more predefined Transformations is possible.
I just thought this one would be the most used one.

e.g. Hotbar + Grid + Equipment would be:
`InventoryTransformation.of(INVENTORY_TYPE.of(Hotbar.class), INVENTORY_TYPE.of(MainPlayerInventory.class),
INVENTORY_TYPE.of(EquipmentInventory.class))`

Faithcaio: I changed the order in Sponge to be more consistent.
The MainPlayerInventory is a 9x4 Grid and its slots go from topleft to bottomright.
Like all other grid inventories.

Keep in mind that we need to support mods changing and adding inventories.

Cybermaxke: You're right, the doc is good the way it is, I was confused because the `MainPlayerInventory` already includes the `Hotbar`.

But I still think it would make more sense to allow `HOTBAR_FIRST` to be target specific, for example:
```
PlayerInventory inventory = ...;
// The following gives:
// InventoryTransformation.of(INVENTORY_TYPE.of(Hotbar.class), INVENTORY_TYPE.of(MainPlayerInventory.class), INVENTORY_TYPE.of(EquipmentInventory.class))
Inventory prioritizedHotbar = inventory.transform(InventoryTransformations.HOTBAR_FIRST);
// The following gives:
// InventoryTransformation.of(INVENTORY_TYPE.of(Hotbar.class), INVENTORY_TYPE.of(MainPlayerInventory.class))
inventory = inventory.getMain().transform(InventoryTransformations.HOTBAR_FIRST);
// 
```
Or if you prefer not to, change the transformation field name to say that it's only the main inventory. (to avoid confusion and conflicts with possible other transformations)
e.g. `PLAYER_MAIN_HOTBAR_FIRST`

Faithcaio: `Inventory#union` operates on the Set of Slots so no duplicates.

What do you mean by target specific?
Applying `InventoryTransformation.of(INVENTORY_TYPE.of(Hotbar.class), INVENTORY_TYPE.of(MainPlayerInventory.class), INVENTORY_TYPE.of(EquipmentInventory.class))`
to `inventory.getMain()` would result in the same result (as without EquipmentInventory) because the `#getMain` does not contain an `EquipmentInventory`.

The rename would be fine.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -666,6 +666,17 @@
      */
     boolean containsInventory(Inventory inventory);
 
+    /**
+     * Transforms this inventory using the transformation
+     *
+     * @param transformation The Transformation
+     *
```
ST-DDT: No empty line between `@param` and `@return`.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -666,6 +666,17 @@
      */
     boolean containsInventory(Inventory inventory);
 
+    /**
+     * Transforms this inventory using the transformation
```
ST-DDT: I would state that its transformed using the **given** transformation. Also add a dot at the end of the line.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -666,6 +666,17 @@
...
+     *
+     * @param transformation The Transformation
```
ST-DDT: Either use a lowercase `transformation` (recommended) or use a an `@link`. (same for the other references as well)

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.item.inventory.query.QueryOperation;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A Inventory Transformation can filter and order an inventory
+ * <p>
```
ST-DDT: Empty line before the `<p>` tag and no indentation after it.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
```diff
@@ -0,0 +1,85 @@
...
+ *     See {@link InventoryTransformations} for some predefined Transformations
+ * </p>
+ */
+public interface InventoryTransformation {
+
+    /**
+     * Transforms an Inventory
```
ST-DDT, line 42: Please add a little more description on how this inventory is supposed to be mutated. How does it behave in relation to duplicate slots and stuff alike. (and if there are duplicate slots, how the user can omit them).

Either here or on class level.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.item.inventory.entity.Hotbar;
+import org.spongepowered.api.item.inventory.entity.MainPlayerInventory;
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public class InventoryTransformations {
+
+    public static final InventoryTransformation NONE = DummyObjectProvider.createFor(InventoryTransformation.class, "NONE");
```
ST-DDT: What does this transformation do?
Does it anything at all?
If not maybe call it `NO_OP`?

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item.inventory;
+
+import org.spongepowered.api.item.inventory.entity.Hotbar;
+import org.spongepowered.api.item.inventory.entity.MainPlayerInventory;
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public class InventoryTransformations {
```
ST-DDT: Maybe at a transformation for `EMPTY` (so you have the basic equivalent of `true` (all) and `false` (nothing)).

ST-DDT: Also missing class javadoc.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final InventoryTransformation NONE = DummyObjectProvider.createFor(InventoryTransformation.class, "NONE");
+
+    /**
+     * A Transformation that puts the Hotbar of a player inventory before the main inventory.
+     * <p>
```
ST-DDT: Empty line before the `<p>` tag and no indentation after it.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
```diff
@@ -0,0 +1,85 @@
...
+import org.spongepowered.api.item.inventory.query.QueryOperation;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A Inventory Transformation can filter and order an inventory
+ * <p>
+ *     See {@link InventoryTransformations} for some predefined Transformations
+ * </p>
+ */
+public interface InventoryTransformation {
```
ST-DDT: Also it would be nice if you could add methods like [`andThen(InventoryTransformation)`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#andThen-java.util.function.Function-) and [`compose(InventoryTransformation)`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html#compose-java.util.function.Function-).

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+     *     This transformation is equivalent to {@link Inventory#union(Inventory)} together a query
+     *     for {@link Hotbar} and {@link MainPlayerInventory} in this order
+     * </p>
+     */
+    public static final InventoryTransformation PLAYER_MAIN_HOTBAR_FIRST = DummyObjectProvider.createFor(InventoryTransformation.class, "PLAYER_MAIN_HOTBAR_FIRST");
+
+    /**
+     * A Transformations that reverses the slot order.
```
ST-DDT: >  A **transformation** that reverses the slot order.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public static final InventoryTransformation REVERSE = DummyObjectProvider.createFor(InventoryTransformation.class, "REVERSE");
```
ST-DDT: Empty line after the last class member.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
Faithcaio: doing exactly nothing :)

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformation.java
Faithcaio: It really depends on the implementation of the specific transformation.
But I explained the ones you can create yourself with the builder.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A transformations that reverses the slot order.
+     */
+    public static final InventoryTransformation REVERSE = DummyObjectProvider.createFor(InventoryTransformation.class, "REVERSE");
+
+    /**
+     * A transformation that returns the incoming inventory without changing anything.
+     */
+    public static final InventoryTransformation NO_OP = DummyObjectProvider.createFor(InventoryTransformation.class, "NO_OP");
```
Zidane, line 52: How is this useful?

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A transformation that returns an empty inventory.
+     */
+    public static final InventoryTransformation EMPTY = DummyObjectProvider.createFor(InventoryTransformation.class, "EMPTY");
```
Zidane, line 57: Same as above ^

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
kashike: It's useful when you have a method that accepts an `InventoryTransformation` and want to have the option of doing nothing without creating a method that doesn't accept an `InventoryTransformation` (which would result in ugly code - `if(has tranformation) { method a } else { method b }`.

src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java
Faithcaio: Same as kasike above

PR #1804, opened 2018-04-12
client

kashike:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1868) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/2096)

### DIFF
--- a/src/main/java/org/spongepowered/api/Engine.java
+++ b/src/main/java/org/spongepowered/api/Engine.java
@@ -1,0 +1,49 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
+
+import org.spongepowered.api.client.Client;
+import org.spongepowered.api.scheduler.Scheduler;
+
+/**
+ * Shared functionality between {@link Client} and {@link Server} engines.
+ */
+public interface Engine {
+
+    /**
+     * Gets the scheduler used to schedule tasks.
+     *
+     * @return The scheduler
+     */
+    Scheduler getScheduler();
+
+    /**
+     * Checks if the {@link Thread#currentThread() current thread} is the main thread of the engine.
+     *
+     * @return {@code true} if main thread, {@code false} if not
+     */
+    boolean onMainThread();
+
+}
--- a/src/main/java/org/spongepowered/api/Game.java
+++ b/src/main/java/org/spongepowered/api/Game.java
@@ -25,6 +25,7 @@
 package org.spongepowered.api;
 
 import org.spongepowered.api.asset.AssetManager;
+import org.spongepowered.api.client.Client;
 import org.spongepowered.api.command.CommandManager;
 import org.spongepowered.api.config.ConfigManager;
 import org.spongepowered.api.data.DataManager;
@@ -86,7 +87,6 @@ public interface Game {
      */
     Path getSavesDirectory();
 
-
     /**
      * Returns if the {@link Server} is available for use. The result of this method is entirely
      * dependent on the implementation.
@@ -103,6 +103,50 @@ public interface Game {
      */
     Server getServer();
 
+    /**
+     * Gets the {@link Server}, if it's available.
+     *
+     * @return The server
+     */
+    default Optional<Server> optionalServer() {
+        if (this.isServerAvailable()) {
+            return Optional.of(this.getServer());
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Returns if the {@link Client} is available for use. The result of this method is entirely
+     * dependent on the implementation.
+     *
+     * @return True if the Client is available, false if not
+     */
+    default boolean isClientAvailable() {
+        return false;
+    }
+
+    /**
+     * Gets the {@link Client}.
+     *
+     * @return The client
+     * @throws UnsupportedEngineException If the client engine is not supported
+     * @throws IllegalStateException If the Client isn't currently available
+     */
+    default Client getClient() {
+        return Client.unsupportedEngine();
+    }
+
+    /**
+     * Gets the {@link Client}, if it's available.
+     *
+     * @return The client
+     */
+    default Optional<Client> optionalClient() {
+        if (this.isClientAvailable()) {
+            return Optional.of(this.getClient());
+        }
+        return Optional.empty();
+    }
 
     /**
      * Retrieves the GameDictionary (item dictionary) for this {@link Game}.
@@ -112,7 +156,6 @@ public interface Game {
     default Optional<GameDictionary> getGameDictionary() {
         return Optional.empty();
     }
-
 
     /**
      * Returns the current platform, or implementation, this {@link Game} is running on.
@@ -217,7 +260,9 @@ public interface Game {
      * Gets the scheduler used to schedule tasks.
      *
      * @return The scheduler
+     * @deprecated Use {@link Server#getScheduler()} or {@link Client#getScheduler()}
      */
+    @Deprecated
     default Scheduler getScheduler() {
         return Sponge.getScheduler();
     }
--- a/src/main/java/org/spongepowered/api/client/Client.java
+++ b/src/main/java/org/spongepowered/api/client/Client.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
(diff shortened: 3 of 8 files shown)


src/main/java/org/spongepowered/api/Engine.java
```diff
@@ -0,0 +1,42 @@
...
+import org.spongepowered.api.client.Client;
+import org.spongepowered.api.scheduler.Scheduler;
+
+/**
+ * Shared functionality between {@link Client} and {@link Server} engines.
+ */
+public interface Engine {
+
+    /**
+     * Gets the scheduler used to schedule tasks.
```
gabizou, line 36: What's special about this scheduler? Is it per "engine"? If there's a single player world, technically there's two engines present, no?

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -217,7 +260,9 @@ default ServiceManager getServiceManager() {
      * Gets the scheduler used to schedule tasks.
      *
      * @return The scheduler
+     * @deprecated Use {@link Server#getScheduler()} or {@link Client#getScheduler()}
```
gabizou, line 263: Why not make it based on current thread?

src/main/java/org/spongepowered/api/client/Client.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public interface Client extends Engine {
+
+    /**
+     * Throws an exception stating that the client engine is not supported.
+     *
+     * @return Nothing
+     * @throws UnsupportedEngineException The client engine is not supported
+     */
+    static <T> T unsupportedEngine() throws UnsupportedEngineException {
```
gabizou, line 41: what's the point of this? Why not throw it in Sponge?

src/main/java/org/spongepowered/api/scheduler/Task.java
```diff
@@ -219,6 +219,13 @@ default Builder execute(Runnable runnable) {
          */
         Builder name(String name);
 
+        /**
+         * Builds the task.
+         *
+         * @return The task
+         */
+        Task build();
```
gabizou, line 227: What? What about submitting it?

gabizou (changes requested):
Some clarifying commentary is needed. I know this is WIP, but simply providing feedback where I see fit in the development of this API.

src/main/java/org/spongepowered/api/Engine.java
kashike: Nothing special. This is a shared interface between `Client` and `Server`. Yes, when a client is running a singleplayer world it has two schedulers - one on `Client`, one on `Server.

src/main/java/org/spongepowered/api/Game.java
kashike: Because it shouldn't be a guessing game - someone should explicitly request which scheduler they want.

src/main/java/org/spongepowered/api/client/Client.java
kashike: This was used in more than one place before, and I haven't removed it yet as I'm not sure if it will be used again right now.

src/main/java/org/spongepowered/api/scheduler/Task.java
kashike: See the deprecation right below this method.

src/main/java/org/spongepowered/api/Engine.java
gabizou: Can you make that clearer in the docs here?

src/main/java/org/spongepowered/api/Game.java
gabizou: Right, it shouldn't be a guessing game, but it should also be able to detect which active thread is currently used based on the caller. This should be as easy as possible to use personally, and yes exposing the client and server, a majority of plugin developers will not know to use the client, until it's seemingly necessary in a corner case their plugin is installed on the client (like say VoxelSniper). All of the code should still work on the client regardless whether it was written in mind to be server side, because the simple call to `Sponge.getScheduler()` at the *call site* is on the server thread already.

src/main/java/org/spongepowered/api/client/Client.java
gabizou: Why not just make it a dummy object provider provided method? Or the actual uses for it can be DummyObjectProvider and remapped later.

src/main/java/org/spongepowered/api/scheduler/Task.java
gabizou: Why is that necessary? If I already started making a sync'ed task from the server scheduler, shouldn't it be possible to just call submit? Why clog up code to force me to submit it after the fact?

Or, maybe make t build method as such: `submit(Object plugin, Scheduler scheduler)`.

Zidane: @gabizou 

No the submit stuff made the Scheduler API foreign to other builder-like objects where you build your object and then decide later when you actually want to submit it to something. Its the same problem we have with stuff like WorldGeneratorModifiers where building it actually adds the CatalogType and makes the API muddy in that part.

Due to the deprecation (not getting a task builder via a Scheduler), I see no reason to keep submit and would much rather have a clean and consistent "Get a task builder, build my task, submit to a Scheduler of my choice".

src/main/java/org/spongepowered/api/Engine.java
Zidane: I don't see the need for further documentation here.

Even stating the fact that there may be two engines in SinglePlayer is implementation leak to the MCP implementation. Client itself is borderline that too but for the sake of the majority of our userbase, I deemed it okay to proceed.

src/main/java/org/spongepowered/api/Game.java
Zidane: I think I like gabizou's idea to get a Scheduler (An Optional one) that fits the current thread you are on.

I don't think this is the appropriate place for it and the Scheduler here does need to be removed.

In the Sponge god object would be okay imo.

Zidane: Actually I want to take some time to weigh the arguments on both sides and ponder this.

dualspiral: It's fine for schedulers on the server/client thread, but what about if I wanted to get a sync scheduler from an async thread to perform a block change after a database query? Unless you track which scheduler created which thread, if you remove the ability to select the correct scheduler, then it becomes a guessing game once again.

I think we need to be careful about making assumptions about things without providing a credible way to override them. Assumptions are not going to work 100% of the time, and if someone find some way to use SF for a client/internal-server hybrid (if that's even possible, I don't know!), then there might be a use case?

We should always be giving developers the chance to be explicit about what they want when it makes sense to. I would say that this is one of those times - if a plugin developer is developing for the client, they know that's the case! However, we also need through docs about where, for an integrated server, the server scheduler and the client scheduler should be used, because providing both is possibly going to muddy the waters for developers, I fear. I get that making assumptions would remove that need somewhat, but removing the chance to be explicit will bite us at some point.

With that in mind, what about `Optional<Scheduler> Game#getScheduler(SchedulerType);` where `SchedulerType` is a catalog type and is either `SERVER` or `CLIENT`?

gabizou: I’d wager the Scheduler can have a platform since we expose the platform type. So you can determine if there’s a client or server scheduled after querying for it from the god Object. This way you can determine which type you have to properly schedule a synchronized task. 

An example of this could be where you perform the verification of the scheduler passed to you, or we eliminate the guess work and define the method with passing in the platform type with an optional, due to the possibility of having no scheduler available for a client or server.

src/main/java/org/spongepowered/api/Engine.java
gabizou: I would just like to have explanaton that there may be multiple engines, or temporary engines etc. because in the event that you have a plug-in on the client, and scheduling a task for the server, the task would need to be recreated during the server start up again later. This would make it clear that there’s the potential for multiple thread schedulers.

src/main/java/org/spongepowered/api/scheduler/Task.java
gabizou: I can agree to that. On the other hand, the main reason why we moved away from registering catalog types in build methods like this was because we have a registry event for registration of those catalog types. As tasks are not registered in the same way, I had the feeling that this would make sense to have a differing way for building and submitting them. 

Would you then say that an asynchronous task would be regardless of the platform for submitting the task? Or would you still define the platform for submitting the task?

src/main/java/org/spongepowered/api/Game.java
Zidane: I've thought about it and we're sticking with the deprecation.

Its not hard to grab either Client or Server objects and call getScheduler....

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -103,6 +103,50 @@
...
+    default Client getClient() {
+        return Client.unsupportedEngine();
+    }
+
+    /**
+     * Gets the {@link Client}, if it's available.
+     *
+     * @return The client
+     */
+    default Optional<Client> optionalClient() {
```
XakepSDK, line 144: `getClientSafely`?

src/main/java/org/spongepowered/api/Game.java
```diff
@@ -103,6 +103,50 @@
      */
     Server getServer();
 
+    /**
+     * Gets the {@link Server}, if it's available.
+     *
+     * @return The server
+     */
+    default Optional<Server> optionalServer() {
```
XakepSDK, line 111: `getServerSafely`? or `getServerChecked`

Cybermaxke:
If there will be an API for the client, how can we separate catalog types that are available on the client and on the server? Recipes and advancements are good examples of this, they are synced with the client. Right now there is only one GameRegistry for both, with this PR there should be one for client and server?

gabizou:
This will be targeting API 8 as we're going to fully expose various client related changes come bleeding changes for 1.13 along with structural changes.

src/main/java/org/spongepowered/api/Game.java
liach: Hmm, what if a plugin wants the scheduler in loading stages? In that case, the scheduler would point to one that exists (on client the one points to client one, on dedicated server the one points to the server one)

src/main/java/org/spongepowered/api/scheduler/Scheduler.java
```diff
@@ -38,8 +38,12 @@
      * Gets a new instance of a {@link Task.Builder}.
      *
      * @return A new task builder
+     * @deprecated Use {@link Task#builder()}
      */
-    Task.Builder createTaskBuilder();
+    @Deprecated
+    default Task.Builder createTaskBuilder() {
```
liach, line 44: How about creating one defaulted to submit to this scheduler?

src/main/java/org/spongepowered/api/Game.java
ItsDoot: `getClientIfPresent` seems like a better choice.

src/main/java/org/spongepowered/api/Game.java
ItsDoot: or possibly `getServerIfPresent`

Zidane:
Manually merged into 1.13

PR #1808, opened 2018-04-20
Allow Entity deaths to be Cancellable

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1882) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2110)
This is mostly to bring in compatibility of entity deaths to match Forge's requirements, while this is somewhat of an implementation leak that it's "possible", it's just not recommended as the health status of the entity could have adverse effects.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/DestructEntityEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/DestructEntityEvent.java
@@ -26,6 +26,7 @@ package org.spongepowered.api.event.entity;
 
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.living.Living;
+import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.entity.living.TargetLivingEvent;
 import org.spongepowered.api.event.message.MessageChannelEvent;
 import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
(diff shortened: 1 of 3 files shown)


Zidane:
mmmm...The Rabbit Hole.

gabizou:
@Zidane precisely. Because Forge requires the event to be cancellable, and I remember this discussion where we elected not to cancel the entity death event for health and next tick update reasons.

I think we would have to perform a check on the entity that the plugin developer that cancelled the event would have to also increase the health in some way, otherwise reset the health to `1`.

ryantheleach:
Are we going to add the behavior on cancel to the JavaDocs? Or is it expected to be implementation dependent.

gabizou:
Spoke to @Zidane and @kashike and some forge members about the change:

It's currently unimplemented in Forge to handle the cancel of the entity death with the following logic: Modders who decide to cancel the entity death should "undeadify" the entity dying. It would otherwise not be the fault of the implementation if the entity does indeed die. What this will mean is that if the event is canceled, drops will still not occur, **but** every tick thereafter will still perform the death animation and process. What can happen however, is that the event is then *rethrown* during death ticks, at which point, if those are cancelled, death doesn't happen.

What @kashike and @Zidane did agree on however is that if the event is thrown **3** times for the same entity, and they're cancelled but the entity is not "undeaded", the entity will ultimately be allowed to die.

I'm comfortable with handling the implementation in Common to support these changes, but I'd like to get some more @SpongePowered/developers input on the idea since the API would react a little differently in this case compared to other events.

PR #1819, opened 2018-05-01
Update SpongeAPI for ChangeEntityEquipmentEvents

ustc-zzzz:
**SpongeAPI** | [SpongeCommon](//github.com/SpongePowered/SpongeCommon/pull/1895)

* `getOriginalItemStack` and `getItemStack` are marked as deprecated.
* a new method named `getTransaction` is added.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
+++ b/src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
@@ -48,4 +48,9 @@ public abstract class AbstractChangeEntityEquipmentEvent extends AbstractEvent i
     public Optional<Transaction<ItemStackSnapshot>> getItemStack() {
         return Optional.of(this.getTransaction());
     }
+
+    @Override
+    public Transaction<ItemStackSnapshot> getTransaction() {
+        return this.transaction;
+    }
 }
--- a/src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
+++ b/src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
@@ -1,0 +1,51 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    @Deprecated
+    public Optional<ItemStackSnapshot> getOriginalItemStack() {
+        return Optional.ofNullable(this.getTransaction().getOriginal());
+    }
+
+    @Override
+    @Deprecated
+    public Optional<Transaction<ItemStackSnapshot>> getItemStack() {
+        return Optional.of(this.getTransaction());
```
dualspiral, line 46: You've created an infinite loop here that will cause a stack overflow. [`getTransaction()` calls this method](https://github.com/SpongePowered/SpongeAPI/pull/1819/files#diff-679a51feabed6f5c847c734dfe49b02fR90), and this method calls `getTransaction()`.

Even if one or the other method gets overriden in a child class and this is hidden (though I don't see this in Common), better to avoid the potential for this in the first place.

dualspiral (dismissed):

src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
ustc-zzzz: @dualspiral No it doesn't, the implementation (`ChangeEntityEquipmentEvent.XXX.Impl`, which extends `AbstractChangeEntityEquipmentEvent`) will automatically generate a getter to override the `getTransaction` method, and the default implementation will not be used:
```java
@Override
public Transaction getTransaction() {
    return this.transaction;
}
```
However, It might be a better choice to declare the method overriding explicitly in `AbstractChangeEntityEquipmentEvent` rather than override the method implicitly in the automatically generated implementation. Thank you for reminding me.

dualspiral: Yes, that's better, thank you. While I figured that there wouldn't be one in our implementation, my concern was that the API on its own would cause an SOE, which the API should not suggest. As the API stood, if a plugin dev or other implementation just implemented the API and didn't worry about overriding, there would be trouble.

This resolves the concern, thanks.

src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.spongepowered.api.data.Transaction;
+import org.spongepowered.api.event.entity.ChangeEntityEquipmentEvent;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.util.annotation.eventgen.UseField;
+
+import java.util.Optional;
+
+/**
+ * @author ustc_zzzz
```
ImMorpheus: No need for an author tag. Keep it consistent with the other classes

ImMorpheus:
@ryantheleach Any reason why you put the bleeding label ? 

This is a bit old so I'm not sure if it was decided not to be included in stable

src/main/java/org/spongepowered/api/event/impl/AbstractChangeEntityEquipmentEvent.java
ustc-zzzz: Oh sorry. My development tool added the author tag automatically and I forgot to remove it.

ustc-zzzz: @ImMorpheus Fixed in 9aaa36fe9e44d587ac4ac8fdd683034467417008

alimtunc:
Is that PR will be added ?

Faithcaio:
https://github.com/SpongePowered/SpongeAPI/commit/23fe45713674be9065257192bdf1d4e6ada7ed2e

PR #1807, opened 2018-04-19
Added `OptionalValue<E>.asImmutable(): ImmutableOptionalValue<E>`.

Dykam:
This complements `ImmutableOptionalValue<E>.asMutable(): OptionalValue<E>`.

PR #1826, opened 2018-05-12
Add Value.copy().

JBYoshi:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1907)

### DIFF
--- a/src/main/java/org/spongepowered/api/data/value/mutable/CollectionValue.java
+++ b/src/main/java/org/spongepowered/api/data/value/mutable/CollectionValue.java
@@ -152,4 +152,7 @@ public interface CollectionValue<E, C extends Collection<E>, V extends Collectio
 
     @Override
     I asImmutable();
+
+    @Override
+    V copy();
 }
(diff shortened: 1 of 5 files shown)


gabizou:
Originally I was for this, but I decided against it for `Value`s because of the nature of them not knowing the types of elements they contain, so they cannot make a true copy by itself, whereas a `DataManipulator` can because it already knows what element types it contains.

JBYoshi:
From what I can see through a look through the `Keys` class, pretty much every value is either an immutable object or a collection of immutable elements handled by a custom `Value` sub-interface (e.g. `ListValue`, `SetValue`, `MapValue`).

Also, I don't think that this would cause any more problems than just using `value.asImmutable().asMutable()`. This would just be a slightly faster method of doing that.

JBYoshi:
Just realized the `OptionalValue` fix is already handled by another PR (#1807). I'll remove that from mine.

gabizou:
I'll agree, as we already can't verify the validity of copies in `asMutable()`/`asImmutable()`, `copy()` already makes sense.

gabizou (approved):

Issue #1287, opened 2016-06-30
Add Magma as a BlockDamageSource for entities walking over.


gabizou:
Pretty simple, as much as it is for block damage sources for lava and fire.

ImMorpheus:
PR merged (https://github.com/SpongePowered/SpongeCommon/pull/1914)

PR #1832, opened 2018-05-19
Add missing damage sources

### LINKED ISSUE #1496: Expose 1.10+ added DamageSources
In 1.10-1.10.2, a few new internal `DamageSource`s have been added. I'll put up a table for both versions as they need to be added to each API branch respectively.

1.10.2 | Description
------ | -------
`FLY_INTO_WALL` | Flying into a wall horizontally
`DRAGON_BREATH` | EnderDragon specific damage (currently not used however)

1.11.2 | Description
------ | -------
`FIREWORKS` | Apparently, fireworks now deal damage if exploding while still being ridden

All of these may need to be evaluated as to how they're being used internally and through the API, much like how there is the representation of `BlockDamageSource` versus a flat `DamageSource` without a location.

Relatively easy to implement, just marking this as an issue before it is forgotten.

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1914)

This along with the implementation PR should fix https://github.com/SpongePowered/SpongeAPI/issues/1496 and https://github.com/SpongePowered/SpongeAPI/issues/1287

Fireworks is handled with an explosion type and as an `EntityDamageSource` while magma is handled as a `BlockDamageSource`.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
+++ b/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
@@ -68,12 +68,6 @@ public final class DamageSources {
 
     public static final DamageSource VOID = DummyObjectProvider.createFor(DamageSource.class, "VOID");
 
-    /**
-     * Generally used to describe the damage taken when colliding into a wall
-     * while flying with an elytra.
-     */
-    public static final DamageSource WALL_COLLISION = DummyObjectProvider.createFor(DamageSource.class, "WALL_COLLISION");
-
     public static final DamageSource WITHER = DummyObjectProvider.createFor(DamageSource.class, "WITHER");
 
     // SORTFIELDS:OFF
--- a/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
+++ b/src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
@@ -45,6 +45,8 @@ public final class DamageSources {
 
     // SORTFIELDS:ON
 
+    public static final DamageSource DRAGON_BREATH = DummyObjectProvider.createFor(DamageSource.class, "DRAGON_BREATH");
+
     public static final DamageSource DROWNING = DummyObjectProvider.createFor(DamageSource.class, "DROWNING");
 
     public static final DamageSource FALLING = DummyObjectProvider.createFor(DamageSource.class, "FALLING");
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
```diff
@@ -45,6 +45,8 @@
 
     // SORTFIELDS:ON
 
+    public static final DamageSource DRAGON_BREATH = DummyObjectProvider.createFor(DamageSource.class, "DRAGON_BREATH");
```
gabizou, line 51: Javadocs?

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
```diff
@@ -63,6 +65,12 @@
 
     public static final DamageSource VOID = DummyObjectProvider.createFor(DamageSource.class, "VOID");
 
+    /**
+     * Generally used to describe the damage taken when colliding into a wall
+     * while flying with an elytra.
+     */
+    public static final DamageSource WALL_COLLISION = DummyObjectProvider.createFor(DamageSource.class, "WALL_COLLISION");
```
gabizou: I'm not fond of calling wall collision to be a raw DamageSource, maybe the `DamageType` supported for wall collisions, but I don't think we should be using just a vanilla `DamageSource` when we can provide some more information about the `DamageSource` (like, do we want to use `BlockDamageSource` or `EntityDamageSource`?)

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
parlough: Sure.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSources.java
parlough: And use the block that they collide into? That's definitely possible.

gabizou: Right, and if that's the case, this would only be restricted to a `DamageType`, and not a `DamageSource`.

gabizou (changes requested):
Fix the java docs and make wall collisions a DamageType (or however collisions are currently handled? I remember I do some translations and damage source creation potentially for blocks)

ImMorpheus:
Missing `DamageSource`s:

    DamageSource IN_FIRE
    DamageSource LIGHTNING_BOLT 
    DamageSource LAVA
    DamageSource HOT_FLOOR
    DamageSource IN_WALL 
    DamageSource CRAMMING 
    DamageSource CACTUS 
    DamageSource FLY_INTO_WALL 
    DamageSource ANVIL 
    DamageSource FALLING_BLOCK
    DamageSource FIREWORKS 

Based on what we are doing here https://github.com/SpongePowered/SpongeCommon/blob/fad041fc062c96fe847acbd54abd28f284a60160/src/main/java/org/spongepowered/common/registry/provider/DamageSourceToTypeProvider.java#L48 we are left with

    DamageSource HOT_FLOOR
    DamageSource CRAMMING 
    DamageSource FIREWORKS 

They should be handled in the common PR.

PR #1834, opened 2018-05-20
Event for recipebook interaction

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1919)

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/ClickInventoryEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/ClickInventoryEvent.java
@@ -61,6 +61,29 @@ public interface ClickInventoryEvent extends ChangeInventoryEvent, InteractInven
         interface Middle extends Drag, ClickInventoryEvent.Middle {}
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/ClickInventoryEvent.java
```diff
@@ -61,6 +63,29 @@
...
+     * Fires when the client requests a recipe to be crafted.
+     */
+    interface Recipe extends ClickInventoryEvent {
+
+        /**
+         * Returns the requested recipe-
+         *
+         * @return The requested recipe.
+         */
+        CraftingRecipe getRecipe();
```
Cybermaxke: This should be changed to `Recipe` instead, in 1.13, `FurnaceRecipe`s are also used in a recipe book.

ryantheleach (approved):

Ewoutvans (approved):

14mRh4X0r:
Sorry, off-by-one error in the issue number :slightly_frowning_face:

PR #1841, opened 2018-06-09
Inventory properties to implementation

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1935)

 - [x] SlotPos
 - [x] SlotSide
 - [x] StringProperty
 - [x] EquipmentSlotType
 - [x] GuiIdProperty
 - [x] Identifiable
 - [x] InventoryCapacity
 - [x] SlotIndex
 - [x] InventoryDimension
 - [x] InventoryTitle

Removed because unused / covered by other property
 - [x] ArmorSlotType
 - [x] MappedIntProperty
 - [x] AcceptsItems


Cybermaxke:
Are you also going to move the `Property`s that aren't related to inventories? The builder can then be moved to the basic `Property` class.
https://github.com/SpongePowered/SpongeAPI/tree/bleeding/src/main/java/org/spongepowered/api/data/property

Faithcaio:
I'd rather not in this PR.
I actually never used the other Properties.
I'm only breaking the stuff I know.

Cybermaxke:
Hmm, ok. Can you move the `InventoryProperty.Builder` to `Property.Builder` instead, so we don't have to break the API twice when moving over all the other properties.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperty.java
```diff
@@ -25,13 +25,42 @@
...
 public interface InventoryProperty<K, V> extends Property<K, V> {
 
+    /**
+     * Represents a builder class to create Inventory Properties
+     *
+     * @param <T>
+     * @param <B>
+     */
+    interface Builder<V, T extends InventoryProperty<?, V>, B extends Builder<V, T, B>> extends ResettableBuilder<T, B> {
+
```
Cybermaxke: A builder method for the key?

Faithcaio: I'm considering removing that.
It is always the default key atm.

Cybermaxke: Yeah, maybe remove the key from the API, haven't seen it used anywhere that it isn't default. If we remove this then it should be done for all the `Property`s.
@gabizou thoughts?

src/main/java/org/spongepowered/api/entity/vehicle/minecart/ContainerMinecart.java
```diff
@@ -31,7 +31,7 @@
  * Represents a minecart with a container inside it. Common extensions of this
  * are {@link ChestMinecart} and {@link HopperMinecart}.
  */
-public interface ContainerMinecart<M extends ContainerMinecart<M>> extends Minecart, Carrier, CarriedInventory<M> {
+public interface ContainerMinecart<M extends ContainerMinecart<M>> extends Minecart, Carrier {
```
gabizou: `M` doesn't look to be used.

gabizou (approved):

src/main/java/org/spongepowered/api/data/Property.java
```diff
@@ -222,4 +225,45 @@ public static Operator defaultOperator() {
...
     boolean matches(@Nullable Property<?, ?> other);
 
+    /**
+     * Represents a builder class to create Inventory Properties
+     *
+     * @param <V> The value type
+     * @param <T> The inventory property type
+     * @param <B> The builder type
+     */
+    interface Builder<V, T extends InventoryProperty<?, V>, B extends Builder<V, T, B>> extends ResettableBuilder<T, B> {
```
Cybermaxke, line 235: Generics need to be changed to target `Property` instead of `InventoryProperty`.

Issue #1824, opened 2018-05-10
Type inconsistencies in implementations of ValueContainer#getValue?


kory33:
Hello, I have a question on Data API.

I wanted to implement my own version of abstract data manipulator. I've already verified that it works for simple data constructs. While writing type proofs, I've found that it is not easy to ensure that the return value of `getValue` is `Optional<V>`. The question is; is it OK to return `Optional<Value<E>>` or `Optional<ImmutableValue<E>>` instead of `Optional<V>` in `ValueContainer#getValue`?

Looking at SpongeAPI code, it does the same thing as I do; it actually returns `Optional<Value<E>>` instead of `Optional<V>`: [AbstractData#getValue](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/data/manipulator/mutable/common/AbstractData.java#L164), and `Optional<ImmutableValue<E>>` instead of `Optional<V>`: [AbstractImmutableData#getValue](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/data/manipulator/immutable/common/AbstractImmutableData.java#L113)

In the former code for example, shouldn't there be a case where `V` is `ImmutableValue<E>` and the code leads to a potential heap pollution? If so, may I propose the signature of `ValueContainer#getValue` to be

```Java
<E> Optional<? extends BaseValue<E>> getValue(Key<? extends BaseValue<E>> key)
```

instead?

gabizou:
It's never OK to return a weaker type. The abstract implementation provided in the API is a lose contracted and *semi functioning* implementation to provide developers ways to implement their own Data without having to write all of the boilerplate code themselves. It's been intended to strip away the abstract implementation in #1198, but it's not completed as other priorities take precedence. 

To answer the question about whether the implementation does cause heap pollution: Short answer is **no**, but the long answer is dependent on the developer using the abstract implementation. If the plugin developer is returning a `Value` of a weaker type than the `Key` supports, that is their fault as the generic contract in their own created `Key` instances. Ultimately, Sponge's implementation of all Sponge provided Data does assure that heap pollution does not take place, because the return types of the `Supplier`s are matching the `Value` types identified by the `Key`.

It would be a good type check to use the `Key`'s `TypeToken` to validate that the supplier is returning the correct type, and point it out to the plugin developer, but in short, the implementation is not responsible for what is ultimately considered "user error".

[...]

PR #1843, opened 2018-06-09
Refactor Values for Data API 

gabizou:
In consideration of https://github.com/SpongePowered/SpongeAPI/issues/1824, I've elected to take the chance to ~royally break~ refactor the hierarchy of `Value` and the subclasses of such. 

The design for the new classes so to speak is to have separated the accessors from mutator methods. Accessor methods are available on the base classes (so, you have some list methods available in `ListValue`), but the mutator methods and modifier methods are available depending on whether you have a `Mutable` or `Immutable` subclass of the overall target.

The old usage  of values was something like so:

```java
final BlockState state = BlockTypes.STONE.getDefaultState();
final Optional<StoneType> stoneType = state.get(Keys.STONE_TYPE);
    // Up until this point, we're fine. This next line will cause a class cast exception
final Value<StoneType> value = state.getValue(Keys.STONE_TYPE).orElseThrow(() -> new IllegalStateException("Stone should have stone type")); 
// This won't work as the value is already an immutable value instance, throws an exception in the line above.
final ImmutableValue<StoneType> stoneTypeImmutableValue = value.asImmutable();
```

The biggest problem here is that the returned class type of a BlockState would not be a `Value`, but an `ImmutableValue`. Neither classes are interchangeable, let alone usable in this case because the returned value would still end up being invalid and cause a `ClassCastException`.

Now, what this refactoring intends to replace that with is:
```java
final BlockState state = BlockTypes.STONE.getDefaultState();
final Optional<StoneType> stoneType = state.get(Keys.STONE_TYPE);
final Value<StoneType> stoneTypeValue = state.getValue(Keys.STONE_TYPE).orElseThrow(() -> new IllegalStateException("Stone should have stone type!"));
final Value.Immutable<StoneType> stoneTypeImmutable = stoneTypeValue.asImmutable();
final Value.Mutable<StoneType> stoneTypeMutable = stoneTypeValue.asMutable();
```

The beauty of the new way is that you don't have to worry about whether it's mutable or not, since you can declare it as such and it will **work** without issue.

I'm expanding the PR a little to improve some more parts about the Values and how they are used, but this is it for now.


ryantheleach:
I'm assuming `Value.Immutable<StoneType>.asMutable()` isn't possible?

gabizou:
It is possible. That’s why there’s the asMutable and asImmutable methods available at the base class, just that the mutable and immutable variants will return themselves respectively.

src/main/java/org/spongepowered/api/data/value/Value.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Retrieves a mutable form of this value. Due to the vague nature of the
+     * value itself, some cases can already provide a {@link Mutable} instance
+     * where this would simply return itself. In other cases, where the retrieved
+     * value is an {@link Immutable} instance, a new mutable value is created
+     * with the same key, values, and defaults.
+     *
+     * @return A mutable value
+     */
+    Mutable<E> asMutable();
```
ST-DDT, line 106: Please add a method like `asMutableCopy()` that explicitly creates a new mutable copy (or move the `copy()` method here).
(This method could be default implemented in the mutable variant to call `copy()` and in the immutable variant by calling to `asMutable()`.) 

The goal of that method is to explicitly create a copy that can only be modified in the following code and not from the callee without unnecessary object creation.
Also it is more readable then `source.asImmutble().asMutable()` or `source.asMutable().copy()`.

gabizou: You already get copies of objects when retrieved from either `DataHolder`s or `ValueContainer`s (`DataManipulator` extends `ValueContainer`), so calling `asMutable()` is already going to either be a copy from the original container, or a new copy from an immutable value. Adding `asMutableCopy()` only really adds unnecessary steps for what's already intended as a safeguard for developers to use otherwise.

ryantheleach: Ah ok, on the surface I was thinking Value was going to act as a lens / accessor to immutable / mutable values, But I guess that's more Key's purpose. That's why I was worried about asMutable being exposed on the immutable Value, as if I pass an immutable value to a method, I don't expect it to be able to modify the underlying object.

Cybermaxke:
https://github.com/SpongePowered/SpongeAPI/commit/69b2f9bfc959f13b114eba09bfcca7e8a738a232

Issue #1294, opened 2016-07-04
GenericArguments.entity can not parse playernames, parsing errors confusing


ryantheleach:
GenericArguments.entity is supposed to be able to handle, uuid's, playernames, and selectors, but at the moment trying to parse UUID's fails with. 

No values matching pattern '39669d04-e02e-47e3-bf53-084d6fd7a6fc' present for uuid!

and attempting to parse player names results in 

Error occurred while executing command 'delball Player636' for source EntityPlayerMP['Player636'/361, l='spawntest', x=-78.28, y=68.37, z=244.06]: Invalid UUID string: Player636
java.lang.IllegalArgumentException: Invalid UUID string: Player636

ryantheleach:
Not a bug, just confusing error handling.

kashike:
Is this still an issue?

kashike:
@ryantheleach

ryantheleach:
Sorry been out of it for a while.

Tested just now, on Bleeding 726992c2595aab5d2fa9d80a5efb4bf52033035b.

![image](https://cloud.githubusercontent.com/assets/141554/25006604/b6486014-209c-11e7-967a-6cd5ae31e46a.png)

dualspiral:
8976964c1038066d3555ac6fb613ed3839037090

Issue #1192, opened 2016-04-20
GenericArguments.choices is case sensitive.


ryantheleach:
GenericArguments.choices is currently case sensitive. any chance of getting either a case insensitive version, or modifying choices to be case insensitive by default?

Maybe it should select the choice if there is no other casing, but bail if there are 2 choices with the same case.

modwizcode:
Couldn't you just register two choices, one uppercase and one lowercase?

kashike:
@modwizcode So you want someone to register all of these? :laughing: 
`"SPONGE", "SPONGe", "SPONgE", "SPONge", "SPOnGE", "SPOnGe", "SPOngE", "SPOnge", "SPoNGE", "SPoNGe", "SPoNgE", "SPoNge", "SPonGE", "SPonGe", "SPongE", "SPonge", "SpONGE", "SpONGe", "SpONgE", "SpONge", "SpOnGE", "SpOnGe", "SpOngE", "SpOnge", "SpoNGE", "SpoNGe", "SpoNgE", "SpoNge", "SponGE", "SponGe", "SpongE", "Sponge", "sPONGE", "sPONGe", "sPONgE", "sPONge", "sPOnGE", "sPOnGe", "sPOngE", "sPOnge", "sPoNGE", "sPoNGe", "sPoNgE", "sPoNge", "sPonGE", "sPonGe", "sPongE", "sPonge", "spONGE", "spONGe", "spONgE", "spONge", "spOnGE", "spOnGe", "spOngE", "spOnge", "spoNGE", "spoNGe", "spoNgE", "spoNge", "sponGE", "sponGe", "spongE", "sponge"` etc

modwizcode:
Good point. Maybe a `uncasedChoices` would make more sense.

windy1:
It should just be case insensitive by default IMO if we want to match the behavior of bash which seems to be the trend within the command API.

modwizcode:
Choices is used in a few places besides flags IIRC, I think a few of those uses would be best to remain case sensitive. Additionally command line programs often use multiple cases in flags, which I think is a sane default.

zml2008:
Yeah, I think both case-sensitive and -insensitive variants are needed for choices -- there are valid use cases for case-sensitive but case-insensitive is a reasonable default.

dualspiral:
8976964c1038066d3555ac6fb613ed3839037090

PR #1844, opened 2018-06-10
Quality of Life updates for the current command API

dualspiral:
This is actually more for stable-7, but I'll target bleeding and cherry-pick as and when so that I know I keep them in sync!

This is simply to make a couple of bugfixes and quality of life improvements before I focus on the MC 1.13 commands. Wouldn't mind an eye over them first as they do touch the API project, hence the PR. These changes are:

* Add `CommandContext#getOneUnchecked` for when you know you have that element in the `CommandContext` (avoids creating the `Optional`, similar to `ServiceManager#provideUnchecked`)
* Fix the `SpaceSplitInputTokenizer` (and add tests to verify)
* Add `CommandContext.Snapshot` and get/apply methods
* Fix `GenericArguments#firstParsing` sometimes leaving arguments in the context if an argument sequence fails part way though (and tests to verify)
* Add a case insensitive choices element option, #1192 
* Fix entity element not working with player names and player UUIDs, #1294

Some of this has been inspired by the previous PR by Cybermaxke, see #1284, and my previous PR, see #1323.


src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -170,4 +226,45 @@ public boolean hasAny(String key) {
...
+    public void applySnapshot(Snapshot snapshot) {
+        this.parsedArgs.clear();
+        this.parsedArgs.putAll(snapshot.args);
+    }
+
+    /**
+     * A snapshot of a {@link CommandContext}. This object does not contain any
+     * public API methods, a snapshot should be considered a black box.
+     */
+    public class Snapshot {
```
ST-DDT: Maybe make it final?

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -116,6 +125,53 @@ public CommandContext() {
...
+     * @throws java.util.NoSuchElementException if there is no element with the
+     *      associated key
+     * @throws IllegalArgumentException if there are more than one element
+     *      associated with the key (thus, the argument is illegal in this
+     *      context)
+     * @throws ClassCastException if the element type is not what is expected
+     *      by the caller
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getOneUnchecked(String key)
```
kashike: Should be `requireOne` to match the `require` name used in the Data API: https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/data/value/mutable/CompositeValueStore.java#L106

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -116,6 +125,53 @@ public CommandContext() {
...
+     * @param <T> the expected type of the argument
+     * @return the argument
+     * @throws java.util.NoSuchElementException if there is no element with the
+     *      associated key
+     * @throws IllegalArgumentException if there are more than one element
+     *      associated with the key (thus, the argument is illegal in this
+     *      context)
+     * @throws ClassCastException if the element type is not what is expected
+     */
+    public <T> T getOneUnchecked(Text key)
```
kashike: `requireOne`

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -170,4 +226,45 @@ public boolean hasAny(String key) {
...
+     *
+     * <p>This is only guaranteed to create a <em>shallow copy</em> of the
+     * backing store. If any value is mutable, any changes to that value
+     * will be reflected in this snapshot. It is therefore not recommended
+     * that you keep this snapshot around for longer than is necessary.</p>
+     *
+     * @return The {@link Snapshot} containing the current state of the
+     *      {@link CommandContext}
+     */
+    public Snapshot getSnapshot() {
```
kashike: `createSnapshot`

kashike (changes requested):

src/main/java/org/spongepowered/api/command/args/CommandContext.java
dualspiral: I have no issue with that. For reference, note that I got this name from the ServiceManager methods, `provide` and `provideUnchecked`, in case that wants to be rationalised later to `get` and `require`.

src/main/java/org/spongepowered/api/command/args/CommandContext.java
gabizou: Agreed, should be final if you're not planning on allowing it to be extended.

gabizou (approved):

kashike (approved):

src/main/java/org/spongepowered/api/command/args/CommandContext.java
dualspiral: Yep, that will be done before I merge

PR #1848, opened 2018-06-17
Fix/inventories

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1946) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2232)

Overview:
- `Container#getViewed()` added
- `Container#setCursor(ItemStack)` added
- `Container#getCursor:Optional<ItemStack>` added
- `Inventory#slots():List<Slot>` changed to return Slots
- `Inventory#children():List<Inventory>` instead of implementing Iterable
- `OrderedInventory` removed and its methods moved to `Inventory`
- `Interactable` -> `ViewableInventory`
- `Inventory#query(QueryOperations<?>...):Inventory` no ducktyping
- `Inventory#query(Class<T extends Inventory>):Optional<T>` returns none when not exact match
- `Inventory#asViewable():Optional<ViewableInventory>` 
- `Slot#transform` -> `Slot#viewedSlot` 
- `MainPlayerInventory` is not a Grid anymore `MainPlayerInventory#asGrid():GridInventory`
- `ViewableInventory#getViewers():Set<Players>` moved from `Container`
- `ViewableInventory#hasViewers():boolean` moved from `Container`
- `InventoryTransactionResult` removed unused UNDEFINED and CANCELLED Types
- `InventoryTransactionResult#getSlotTransactions` replaces `replaced` item list.
- `InventoryTransactionResult#revert()` and `InventoryTransactionResult#revertOnFailure()`
- `InventoryTransactionResult` and Builder moved to Impl
- `Inventory#peek` `Inventory#poll` do not return Optional anymore `ItemStack#empty()` instead
    - Variants with SlotIndex still return `Optional#empty` when the Slot does not exist.
- Updated/Simplified javadocs on `peek` `poll` `offer` `set`
- `UserInventory#getEquipment` now also has both main and offhand
- `UserInventory#getArmor()`
- `PrimaryPlayerInventory` renamed from MainPlayerInventory
- `UserInventory#getPrimary:PrimaryPlayerInventory` (previously `getMain`)
- `UserInventory#getStorage:GridInventory` (previously `getMainGrid`)
- `PrimaryPlayerInventory#getStorage:GridInventory` (previously `getMain`)


src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -72,40 +75,18 @@
...
+     * Returns a list of all direct child inventories.
      *
-     * @param <T> expected inventory type, specified as generic to allow easy
-     *      pseudo-duck-typing
-     * @return the next sibling inventory, or an {@link EmptyInventory} if
-     *      there are no further siblings
+     * @return a list of all direct child inventories.
      */
-    <T extends Inventory> T next();
+    List<Inventory> subInventories();
```
Cybermaxke: `children`?

src/main/java/org/spongepowered/api/item/inventory/type/Interactable.java
```diff
@@ -26,12 +26,45 @@
...
+     */
+    boolean hasViewers();
+
+    /**
+     * Shows this Inventory to the given viewer.
+     *
+     * @param viewer The viewer to show this inventory to
+     * @throws IllegalArgumentException if a {@link PluginContainer} is not the root of the cause
+     */
+    void open(Player viewer) throws IllegalArgumentException;
```
Cybermaxke: I would get rid of these and force devs to use `Player#openInventory` and `Player#closeInventory` instead.

Faithcaio: it really just redirects to exactly that

Cybermaxke: Then there is no need to keep them?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -223,6 +224,63 @@
...
+
+    /**
+     * Sets the item in the specified slot.
+     *
+     * @see Inventory#set(ItemStack)
+     * @param index Slot index to set
+     * @param stack Stack to insert
+     * @return matching stacks, as per the semantics of {@link Inventory#set}
+     */
+    InventoryTransactionResult set(SlotIndex index, ItemStack stack);
```
NeumimTo, line 254: Would it be possible to also introduce method 

InventoryTransactionResult set(SlotPos slotPos, ItemStack stack);

as well?

src/main/java/org/spongepowered/api/item/inventory/type/Interactable.java
```diff
@@ -26,12 +26,45 @@
...
  * Interface for inventories which may be interacted with by Players.
  */
 public interface Interactable extends Inventory {
 
+    /**
+     * Gets the current viewers looking at this Inventory.
+     *
+     * @return The current viewers of this inventory
+     */
+    Set<Player> getViewers();
```
Cybermaxke, line 43: Maybe rename to `getInteractors` to match the class name, can also avoid confusion without the difference between "viewing" and being able to modify the inventory. The same for `hasViewers`.

src/main/java/org/spongepowered/api/item/inventory/type/Interactable.java
```diff
@@ -26,12 +26,45 @@
...
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.item.inventory.Inventory;
+import org.spongepowered.api.plugin.PluginContainer;
+
+import java.util.Set;
 
 /**
  * Interface for inventories which may be interacted with by Players.
  */
 public interface Interactable extends Inventory {
```
Cybermaxke: Rename to `InteractableInventory`? To avoid possible conflicts with naming.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Cybermaxke: `SlotPos` is only used in `Inventory2D` (grid, row or column).

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -498,6 +534,15 @@ default Inventory transform(InventoryTransformation transformation) {
...
 
+    /**
+     * Returns this inventory as an interactable inventory if possible.
+     *
+     * <p>Not all inventories are interactable (e.g. a custom inventory of size 5x5)</p>
+     *
+     * @return This inventory as an interactable inventory if possible.
+     */
+    Optional<Interactable> asInteractable();
+
```
Cybermaxke: Maybe in case of a large grid inventory you can also provide a method like `asInteractableView` which takes the min and max coordinates to create view for.
E.g. for a 50x50 inventory, you can get a 3x9 view starting at row 4 and column 9, and display that to the player
Just an idea though

src/main/java/org/spongepowered/api/item/inventory/Slot.java
```diff
@@ -50,20 +38,6 @@
...
-     *       from Slot, you may call this with {@link Type#INVENTORY}.</dd>
-     * </dl>
-     *
-     * @param type the type to transform into
-     * @return the transformed Slot or itself if already the correct type
-     */
-    Slot transform(Type type);
-
     /**
      * Transforms this Slot into the default Type.
```
Cybermaxke: Javadoc will need to be updated.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: SlotPos is only available on Grids. Not all Inventories are Grids. 
So not possible

src/main/java/org/spongepowered/api/item/inventory/type/Interactable.java
Faithcaio: This is basicly a List of players having a Container open for this Inventory.
Maybe instead rename this interface to ViewableInventory?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: I'd rather not add features with this. I'll have enough work to fix impl.
But I'll keep this in mind for later

src/main/java/org/spongepowered/api/item/inventory/type/Interactable.java
Cybermaxke: That sounds good.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -243,6 +224,63 @@
...
     InventoryTransactionResult set(ItemStack stack);
 
+    /**
+     * Gets and remove the stack at the supplied index in this Inventory.
+     *
+     * @see Inventory#poll()
+     * @param index slot index to query
+     * @return matching stacks, as per the semantics of {@link Inventory#poll()}
+     */
+    Optional<ItemStack> poll(SlotIndex index);
```
XakepSDK, line 206: Maybe make these methods default?
```java
default Optional<ItemStack> poll(SlotIndex index) {
    return query(index.asQuery()).peek();
}
```

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -243,6 +224,63 @@
...
+
+    /**
+     * Gets and remove the stack at the supplied index in this Inventory.
+     *
+     * @see Inventory#poll()
+     * @param index slot index to query
+     * @param limit item limit
```
XakepSDK: What means this limit?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: no its not the same

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: see `Inventory#poll`
```
Get and remove up to <code>limit</code> items of the type in the first
available stack in this Inventory from all stacks in this Inventory...
```

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -421,12 +459,21 @@
...
+    /**
+     * Query this inventory for a single inventory matching the supplied inventory type.
+     * This query will return {@link Optional#empty()} when the query does not return a
+     * single inventory matching the supplied inventory type.
+     *
+     * @param inventoryType The inventory type to query for
+     * @param <T> The Type of inventory
+     * @return the query result
+     */
+    <T extends Inventory> Optional<T> query(Class<T> inventoryType);
```
Cybermaxke, line 456: Maybe rename this to `querySingle` (or something else)? The normal `query` method returns a collection `Inventory` and this method returning a single `Inventory` could make the naming confusing.

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
```diff
@@ -74,50 +76,84 @@ public static InventoryTransactionResult failNoTransactions() {
...
     }
 
-    final Type type;
+    private final List<SlotTransaction> slotTransactions;
     private final List<ItemStackSnapshot> rejected;
-    private final List<ItemStackSnapshot> replaced;
+
+    final Type type;
 
     InventoryTransactionResult(Builder builder) {
```
Cybermaxke: Maybe move the `InventoryTransactionResult` impl to SpongeCommon?

parlough: Yes please ^

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -243,6 +224,63 @@
      */
     InventoryTransactionResult set(ItemStack stack);
 
+    /**
+     * Gets and remove the stack at the supplied index in this Inventory.
```
gabizou, line 198: These methods need to be clear on when an `Optional` is empty or not.

Because, we can simplify this to `ItemStack.EMPTY` if we're returning it for a valid slot index, and `Optional.empty()` when the `SlotIndex` is invalid. OR, simplify these methods to always return an `ItemStack` or throw an exception if the slot index is invalid. My two cents on the matter.
The optionals are somewhat annoying to work with, and frankly, it's easier to say "Hey, you're working on an empty itemstack," versus "Hey, you might be working with an empty itemstack anyways, but you have to check the optional first!"

Faithcaio: I changed it to `ItemStack` instead of `Optional` for the ones without SlotIndex.
With SlotIndex still returns Optional because the Slot might not even exist.
`inv.peek(SlotIndex.of(1))` is the same as `inv.getSlot(SlotIndex.of(1)).map(Slot::peek)`

Cybermaxke:
I was discussing with @parlough about the fact that a `EquipmentInventory` which is located in a `PlayerInventory` doesn't contain the held item slots, in Mumfreys original PR there was spoken about a armor inventory (which are the 4 slots) and the `EquipmentInventory` should contain all the slots (ofc only the supported one).
Knowing that we could expose a `ArmorInventory` in the `PlayerInventory`, and make the `EquipmentInventory` return a inventory view containing the held and armor slots.

```java
interface PlayerInventory {

    ArmorInventory getArmor();

    EquipmentInventory getEquipment();

    // ...
}
```

src/main/java/org/spongepowered/api/item/inventory/Container.java
```diff
@@ -24,60 +24,50 @@
...
-     * <p>Examples for viewed inventory slots:</p>
-     * <p>All the slots of the opened chest and not the player grid below.</p>
-     * <p>The slots of the crafting grid in the opened player inventory.</p>
-     *
-     * @param slot The slot to check.
-     * @return {@code true} when the slot is part of the viewed inventories.
+     * @return The players cursor item.
      */
-    boolean isViewedSlot(Slot slot);
+    Optional<ItemStack> getCursor();
```
Zidane: Should we consider empty stacks for this?

src/main/java/org/spongepowered/api/item/inventory/Container.java
```diff
@@ -24,60 +24,50 @@
...
     /**
-     * Stops showing this Inventory to the given viewer.
+     * Sets the viewing players cursor item.
      *
-     * @param viewer The viewer to stop showing this inventory to
-     * @throws IllegalArgumentException if a {@link PluginContainer} is not the root of the cause
+     * @param item The item to set.
      */
-    void close(Player viewer) throws IllegalArgumentException;
+    void setCursor(ItemStack item);
```
Zidane: Could we mark parameter as Nullable and just set the cursor to empty stack then?

src/main/java/org/spongepowered/api/item/inventory/Container.java
```diff
@@ -24,60 +24,50 @@
...
+     * Returns the list of viewed inventories.
+     * <p>This is usually at least the inventory a player opened and the players inventory.</p>
+     * <p>It is not necessary, that all slots of the viewed inventories are visible or interactable with.</p>
      *
-     * @param viewer The viewer to show this inventory to
-     * @throws IllegalArgumentException if a {@link PluginContainer} is not the root of the cause
+     * @return the list of viewed inventories.
      */
-    void open(Player viewer) throws IllegalArgumentException;
+    List<Inventory> getViewed();
```
Zidane: Inventories are composite by nature, why is this needed when you can just query what you want?

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
```diff
@@ -59,15 +59,13 @@ default Hotbar getHotbar() {
      * @return The main inventory grid
      */
     default GridInventory getMainGrid() {
-        return this.getMain().getGrid();
+        return this.getMain().getMain();
```
Zidane: Naming seems odd here.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -30,20 +30,22 @@
...
+import java.util.List;
 import java.util.Optional;
-import java.util.Queue;
 import java.util.function.Consumer;
 
 /**
  * Base interface for queryable inventories.
  */
-public interface Inventory extends Iterable<Inventory>, Nameable, PropertyHolder {
+public interface Inventory extends Nameable, PropertyHolder {
```
Zidane, line 48: I want to hear the arguments for this change. It was at the heart of the Inventory API from the get go to not impose any presumption of ordering and, instead, leave that to the OrderedInventory interface (as that is more a Container concept).

src/main/java/org/spongepowered/api/item/inventory/Container.java
Faithcaio: Its actually Optional  for the case you have a Container which has been closed already.
If no player is looking at it there is no cursor.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: The reason is simple:
All Inventories that we currently have implementations for are accessible by index and getting a Slot of an Inventory by index was a lot of work.
You either had to check&cast to OrderedInventory, which only works for some inventories.
Or query for it but...
When you query for SlotIndex.of(0) you wont always get a single Slot instead all Slots that have index 0 recursivly. That includes the first Slot of every Row/Column.
When you query for OrderedInventory you wont always get one OrderedInventory instead EmptyInventory/OrderedInventory/multiple OrderedInventory.

Iterable was moved to `children()` instead. Because a lot of ppl. found it confusing.

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
Faithcaio: I agree and am open for suggestions

Faithcaio: `getPrimary().getStorage()`

src/main/java/org/spongepowered/api/item/inventory/Container.java
Faithcaio: We could but should we? I'd rather not have null items anymore.

Optional is for when a Slot actually does not exists. Not for empty Slots. 
See changes to `Inventory#peek()` etc.

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
```diff
@@ -42,32 +42,32 @@
      *
      * @return The main inventory
      */
-    MainPlayerInventory getMain();
+    MainPlayerInventory getPrimary();
```
Cybermaxke: What is wrong with `getMain` here?
If you change this then should the transformation `PLAYER_MAIN_HOTBAR_FIRST` (https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/item/inventory/InventoryTransformations.java#L42) be renamed to `PLAYER_PRIMARY_HOTBAR_FIRST` to be consistent

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
```diff
@@ -175,63 +156,41 @@ public Builder type(final Type type) {
...
+        Builder reject(ItemStack... itemStacks);
+
+        /**
+         * Adds the provided {@link ItemStack itemstacks} as stacks that have been
+         * "rejected".
+         *
+         * @param itemStacks The itemstacks being rejected
+         * @return This builder, for chaining
+         */
+        Builder reject(List<ItemStackSnapshot> itemStacks);
```
Cybermaxke: Maybe change the `List` to `Iterable`?

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
```diff
@@ -175,63 +156,41 @@ public Builder type(final Type type) {
...
-                this.replaced = new ArrayList<>();
-            }
-            for (ItemStack itemStack1 : itemStacks) {
-                if (itemStack1 != null) {
-                    this.replaced.add(itemStack1.createSnapshot());
-                }
-            }
-            return this;
-        }
+        Builder transaction(List<SlotTransaction> slotTransactions);
```
Cybermaxke: `List` to `Iterable`?

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
kashike: `getPrimary` is a nicer name, in my opinion.

Cybermaxke: Maybe also renaming the class then? `PrimaryPlayerInventory`?

XakepSDK:
ViewableInventory confuses me.
Player can view container, buy how can he view a inventory?
How can i get container viewers?

Faithcaio:
Stuff you can look into like Chests are/have `ViewableInventory`
When you look into a Chest you open a `Container`.
If two people look into the same Chest they each have their own Container.

I'll add a `getViewer:Optional<Player>` to Container

src/main/java/org/spongepowered/api/item/inventory/entity/PrimaryPlayerInventory.java
```diff
@@ -40,10 +41,18 @@
...
+    GridInventory asGrid();
 
-}
+    /**
+     * Gets the (3x9) grid of the players main inventory.
+     *
+     * @return The grid of the players main inventory
+     */
+    GridInventory getStorage();
+
```
Cybermaxke: Unnecessary empty line

src/main/java/org/spongepowered/api/item/inventory/entity/PrimaryPlayerInventory.java
```diff
@@ -40,10 +41,18 @@
...
+
+}
```
Cybermaxke: Missing newline

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
```diff
@@ -32,7 +32,7 @@
 
 /**
  * Represents the inventory of a {@link User}.
- * It consists of a {@link MainPlayerInventory} (containing the main {@link GridInventory}
+ * It consists of a {@link PrimaryPlayerInventory} (containing the main {@link GridInventory}
  * and the {@link Hotbar}) and an {@link EquipmentInventory}
  */
 public interface UserInventory<T extends User> extends CarriedInventory<T> {
```
Cybermaxke, line 38: Thinking about the armor and equipment inventory, what would be returned when querying for a `EquipmentInventory` class?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Cybermaxke: @Faithcaio thoughts?

src/main/java/org/spongepowered/api/item/inventory/entity/UserInventory.java
Faithcaio: It should detect that all armor slots are contained in the other. 
So you should get one EquipmentInventory (the bigger one) back

XakepSDK:
Bump: no method `offer(SlotIndex, ItemStack)`

ryantheleach:
@XakepSDK commenting on merged PR's has a high chance of being missed. Create a new issue, contact faith to see what they think (since this was their PR), or comment on https://github.com/SpongePowered/SpongeAPI/issues/1000 for eventual cleanup / addition.

PR #1864, opened 2018-07-14
InputTokenzier fixes

SimonFlash:
This PR contains a couple fixes to the `InputTokenizers` and some cleanup of unused features. Tests have also been updated/added for both.

### QuotedStringTokenzier
 - Instances (lenient and non-lenient) are cached in the class rather than being recreated. This should prevent `CommandSpecs` and other areas with tokenizers from having multiple, identical instances.
 - The `handleQuotedStrings` and `trimTrailingSpace` settings have been removed as they were always initialized to the same value. The logic behind them remains the same.
   - Note: `trimTrainigSpace` could actually be detrimental to tab completion, which relies on the existence of an empty argument to know when to continue to the next element.
- Fixed an off-by-one error if the last argument is empty. This would previously return something like [10, 9] (which marks the end index after the start), now it returns [10, 10].

### SpaceSplitInputTokenizer
 - This no longer ignores the last argument. I'm not sure how this happened or how it's been in Sponge for so long like this. Empty strings, space-only strings, and trailing spaces have the same behavior as in `QuotedStringTokenzier`.


gabizou:
@dualspiral I'm guessing this would be looking ok? I'm not one to mess with the commands api, but seeing as you've had experience with it, mind some input?

dualspiral:
The space split one is something that I’ve fixed in another PR of mine, #1844, which I was hoping to merge soon, so I’ll need to reconcile these two. As that has actual API changes, I’m waiting for a go ahead from a lead, I’ll try to get that today.

I’ll look at this in a bit, a basic look yesterday seemed fine, but I’d like to pull this into my IDE and make sure it doesn’t break the parsing.

SimonFlash:
I ran the changes for the space split tokenizer in #1844 through the test added in this PR and two of the tests failed, the one for a string with spaces and for trailing spaces. The behavior for those is based on the results for the quoted strings tokenizer.

dualspiral:
> I ran the changes for the space split tokenizer in #1844 through the test added in this PR and two of the tests failed

It would have been better had you commented on there about the missed cases, rather than redo the work here, though I appreciate that you might not have done the tests without giving it a go. Catch 22, I guess.

As I say, I'll look at the changes in the two PRs and reconcile them when I'm a bit more awake.

SimonFlash:
I wasn't trying to redo work, I didn't even know about your changes until you brought them up here. I tested it out right after you posted that and figured it would be best to continue the conversation here where it started and has the unit test in question.

src/main/java/org/spongepowered/api/command/args/parsing/RawStringInputTokenizer.java
```diff
@@ -24,18 +24,16 @@
...
 package org.spongepowered.api.command.args.parsing;
 
-import org.spongepowered.api.command.args.ArgumentParseException;
-
 import java.util.Collections;
 import java.util.List;
 
 class RawStringInputTokenizer implements InputTokenizer {
-    static final RawStringInputTokenizer INSTANCE = new RawStringInputTokenizer();
+    public static final RawStringInputTokenizer INSTANCE = new RawStringInputTokenizer();
```
dualspiral, line 31: Making this public doesn't matter in a package level class, this should remain as is. Indeed, there is no need to touch this class - remove your changes here.

src/main/java/org/spongepowered/api/command/args/parsing/SpaceSplitInputTokenizer.java
```diff
@@ -35,14 +35,18 @@
...
 
     @Override
-    public List<SingleArg> tokenize(String arguments, boolean lenient) throws ArgumentParseException {
-        List<SingleArg> ret = new ArrayList<>();
-        int lastIndex = 0;
-        int spaceIndex;
-        while ((spaceIndex = arguments.indexOf(" ")) != -1) {
-            ret.add(new SingleArg((arguments = arguments.substring(0, spaceIndex)), lastIndex, lastIndex + spaceIndex));
-            lastIndex += spaceIndex + 1;
+    public List<SingleArg> tokenize(String arguments, boolean lenient) {
```
dualspiral: This has been dealt with in my previous PR (#1844) and so the changes SpaceSplitInputTokenizer and associated tests should be removed from this PR (with an exception, see your tests).

src/test/java/org/spongepowered/api/command/args/parsing/SpaceSplitInputTokenizerTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    public void testEmptyString() throws ArgumentParseException {
+        assertEquals(ImmutableList.of(), parseFrom(""));
+    }
+
+    @Test
+    public void testSpaceString() throws ArgumentParseException {
+        assertEquals(ImmutableList.of(""), parseFrom(" "));
```
dualspiral, line 49: I suspect I missed this test now I look at this, if that's the case, feel free to update _my_ changes to accomodate this test (and include this test in the space tokenizer tests). It should be as simple as removing the regex check, I thought a string of spaces was no arguments.

That said, I'm not convinced that we should return any arguments at all if we have just a load of spaces, but the other side of this is that trailing spaces do create an extra arg at the end.

Dunno if anyone else has an opinion on that.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
```diff
@@ -46,17 +46,15 @@
...
     private static final int CHAR_BACKSLASH = '\\';
     private static final int CHAR_SINGLE_QUOTE = '\'';
     private static final int CHAR_DOUBLE_QUOTE = '"';
-    private final boolean handleQuotedStrings;
     private final boolean forceLenient;
-    private final boolean trimTrailingSpace;
 
-    QuotedStringTokenizer(boolean handleQuotedStrings, boolean forceLenient, boolean trimTrailingSpace) {
-        this.handleQuotedStrings = handleQuotedStrings;
+    private QuotedStringTokenizer(boolean forceLenient) {
```
dualspiral: Leave this package private to avoid creation of synthetic methods please

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
```diff
@@ -46,17 +46,15 @@
...
+    public static final QuotedStringTokenizer INSTANCE = new QuotedStringTokenizer(false);
+    public static final QuotedStringTokenizer LENIENT_INSTANCE = new QuotedStringTokenizer(true);
     private static final int CHAR_BACKSLASH = '\\';
     private static final int CHAR_SINGLE_QUOTE = '\'';
     private static final int CHAR_DOUBLE_QUOTE = '"';
-    private final boolean handleQuotedStrings;
     private final boolean forceLenient;
-    private final boolean trimTrailingSpace;
 
-    QuotedStringTokenizer(boolean handleQuotedStrings, boolean forceLenient, boolean trimTrailingSpace) {
```
dualspiral, line 56: Is there any reason where `handleQuotedStrings` or `trimTrailingSpace` would be useful? I'm wondering if it's better to expose them, rather than remove them.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
```diff
@@ -67,19 +65,11 @@
...
-                skipWhiteSpace(state);
-            }
+            skipWhiteSpace(state);
             int startIdx = state.getIndex() + 1;
             String arg = nextArg(state);
-            returnedArgs.add(new SingleArg(arg, startIdx, state.getIndex()));
-            if (this.trimTrailingSpace) {
-                skipWhiteSpace(state);
-            }
+            returnedArgs.add(new SingleArg(arg, startIdx, arg.isEmpty() ? startIdx : state.getIndex()));
```
dualspiral, line 72: This is where a comment would be handy to explain why you need to include this edge case.

That said, you say this is at the end of the string? Shouldn't `nextArg(state)` deal with this and return the correct state, rather than hacking it together here if `state.hasMore()` is saying that it has more to return?

dualspiral (changes requested):
Mostly niggles

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
dualspiral: (well, actually, not sure if they would be created in this case, but let's leave it package private)

src/main/java/org/spongepowered/api/command/args/parsing/RawStringInputTokenizer.java
SimonFlash: This was made public because both `QuotedStringsTokenzier` and `SpaceSplitInputTokenizer` have their instance marked as public. Regardless of which it is, they should be consistent. I'll leave that one to y'all to decide upon.

src/main/java/org/spongepowered/api/command/args/parsing/SpaceSplitInputTokenizer.java
SimonFlash: I'll compare and rework as necessary to meet the test cases.

src/test/java/org/spongepowered/api/command/args/parsing/SpaceSplitInputTokenizerTest.java
SimonFlash: This is consistent with the behavior of `QuotedStringsTokenzier` at the moment. I *believe* the fringe case is where you have something like `"/command  "` (double space), where that results in a leading space at the start of the argument (split doesn't remove this as it uses a negative index to separate the command from the argument). I'd have to dig through to check if this theory is correct, but for the meantime it should be reasonable enough.

src/main/java/org/spongepowered/api/command/args/parsing/RawStringInputTokenizer.java
dualspiral: Leave it as it was. You’re touching the API layer and we should be ensuring minimal, necessary, changes. There is no benefit to making this change, regardless of consistency.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
SimonFlash: Same thing as with the other two tokenizers, this was the only one that didn't have a private constructor - presumably because it was being called in `InputTokenzier`. I think it's best to be consistent here.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
SimonFlash: `handleQuotedStrings` I can't speak on much. Given that the class is *called* `QuotedStringsTokenzier`, it seems like that kind of defeats the point a bit. I can't think of a use case for it and it wasn't being used, so there's no harm in removing it for now.

`trimTrailingSpaces` actually *needs* to be removed, as the tab completion system relies on the existince of an empty argument to separate a completed element from the start of a new one. Having this set to true would result in a variety of bugs there, hence removal.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
SimonFlash: To be perfectly honest I have no idea; I saw the indexes were off and traced through this enough to verify that this was indeed a solution. I'd have to go back through and figure out exactly what's happening here. I believe implementing more consistent behavior would require a bit of a rework of the parser behind it.

src/main/java/org/spongepowered/api/command/args/parsing/RawStringInputTokenizer.java
SimonFlash: I'll revert it with the next commit, but I disagree with that statement.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
dualspiral: Unfortunately, while we’re in the API layer, anything destined for a stable API should not have their visibility downgraded, it is part of the API, even if package private. Change it back please.

src/main/java/org/spongepowered/api/command/args/parsing/QuotedStringTokenizer.java
dualspiral: Ok.

gabizou:
I hate doing this, but since #1844 was merged, this needs updating/rebasing.

SimonFlash:
I've had some external issues that I've needed to address, so I'm just getting caught up on things now.

With #1844 in this has to be updated - no way getting around that, so don't worry about it. I did attempt this a while ago but I had difficulty setting the workspace up with Git properly - I'll give it another go here shortly, and if I'm unable to get it working I'll reach out.

Zidane:
Closing as the Command Overhaul PR will encapsulate this change for 1.13. Thanks for the work anyways @SimonFlash !

PR #1866, opened 2018-07-17
Add item groups (aka creative tabs)

ghost:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2000) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2327)

* Added `ItemGroup` interface, cataloged by `ItemGroups`
* Added `Optional<ItemGroup> getItemGroup()` to `BlockType` and `ItemType`

Granted, creative tabs are not super important on the server, but it could be useful, for example, for detecting mod-added tabs.

<p align="right"><sub>have mercy on me, this is my first sponge pull request</sub></p>

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -28,6 +28,7 @@ import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.block.trait.BlockTrait;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.property.PropertyHolder;
+import org.spongepowered.api.item.ItemGroup;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
@@ -77,6 +78,13 @@ public interface BlockType extends CatalogType, Translatable, PropertyHolder {
      * @return The item type or {@link Optional#empty()} otherwise
      */
     Optional<ItemType> getItem();
+
+    /**
+     * Gets the {@link ItemGroup} that contains this block.
+     *
+     * @return The item group or {@link Optional#empty()} otherwise
+     */
+    Optional<ItemGroup> getItemGroup();
 
     /**
      * Gets if this BlockType is set to receive random block ticks.
--- a/src/main/java/org/spongepowered/api/block/BlockType.java
+++ b/src/main/java/org/spongepowered/api/block/BlockType.java
@@ -28,7 +28,6 @@ import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.block.trait.BlockTrait;
 import org.spongepowered.api.data.DataHolder;
 import org.spongepowered.api.data.property.PropertyHolder;
-import org.spongepowered.api.item.ItemGroup;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.text.translation.Translatable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
@@ -78,13 +77,6 @@ public interface BlockType extends CatalogType, Translatable, PropertyHolder {
      * @return The item type or {@link Optional#empty()} otherwise
      */
     Optional<ItemType> getItem();
-
-    /**
-     * Gets the {@link ItemGroup} that contains this block.
-     *
-     * @return The item group or {@link Optional#empty()} otherwise
-     */
-    Optional<ItemGroup> getItemGroup();
 
     /**
      * Gets if this BlockType is set to receive random block ticks.
--- a/src/main/java/org/spongepowered/api/item/ItemGroup.java
+++ b/src/main/java/org/spongepowered/api/item/ItemGroup.java
@@ -30,6 +30,9 @@ import org.spongepowered.api.util.annotation.CatalogedBy;
 
 /**
  * Represents an item group, or creative tab.
+ *
+ * <p>Item groups are signified by where they show on clients and otherwise have
+ * little usable functionality.</p>
  */
 @CatalogedBy(ItemGroups.class)
 public interface ItemGroup extends CatalogType, Translatable {
--- a/src/main/java/org/spongepowered/api/item/ItemGroup.java
+++ b/src/main/java/org/spongepowered/api/item/ItemGroup.java
@@ -1,0 +1,35 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/api/block/BlockType.java
```diff
@@ -78,6 +79,13 @@
      */
     Optional<ItemType> getItem();
 
+    /**
+     * Gets the {@link ItemGroup} that contains this block.
+     *
+     * @return The item group or {@link Optional#empty()} otherwise
+     */
+    Optional<ItemGroup> getItemGroup();
```
kashike: Remove this method - it be can retrieved from `getItem()` instead.

kashike (changes requested):

src/main/java/org/spongepowered/api/item/ItemType.java
```diff
@@ -61,6 +61,13 @@
     @Override
     String getName();
 
+    /**
+     * Gets the {@link ItemGroup} that contains this item.
+     *
+     * @return The item group or {@link Optional#empty()} otherwise
+     */
+    Optional<ItemGroup> getItemGroup();
```
mattmess1221: An item can be in multiple groups.

src/main/java/org/spongepowered/api/item/ItemGroup.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents an item group, or creative tab.
+ */
+@CatalogedBy(ItemGroups.class)
+public interface ItemGroup extends CatalogType, Translatable {}
```
liach: Braces on different lines and leave a blank line between. Also, can the icon of the item group be exposed?

src/main/java/org/spongepowered/api/item/ItemGroups.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static final ItemGroup MISC = DummyObjectProvider.createFor(ItemGroup.class, "MISC");
+
+    public static final ItemGroup REDSTONE = DummyObjectProvider.createFor(ItemGroup.class, "REDSTONE");
+
+    public static final ItemGroup TOOLS = DummyObjectProvider.createFor(ItemGroup.class, "TOOLS");
+
+    public static final ItemGroup TRANSPORTATION = DummyObjectProvider.createFor(ItemGroup.class, "TRANSPORTATION");
+
+    // SORTFIELDS:OFF
```
liach, line 57: You missed the `SEARCH` item group.

src/main/java/org/spongepowered/api/item/ItemGroup.java
liach: You should add a method to see if an item is in this item group possibly.

liach (changes requested):

src/main/java/org/spongepowered/api/item/ItemGroup.java
ghost: Unfortunately, item group icons are client only.

src/main/java/org/spongepowered/api/item/ItemGroups.java
ghost: Yes, `SEARCH`, along with `HOTBAR` and `INVENTORY`, were skipped because they are not so much item groups as they are creative tabs.

src/main/java/org/spongepowered/api/item/ItemGroup.java
gabizou: At the very least, make the class ending brace on a new line and leave an empty line in the middle.

src/main/java/org/spongepowered/api/item/ItemGroup.java
```diff
@@ -0,0 +1,35 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.text.translation.Translatable;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents an item group, or creative tab.
```
gabizou, line 32: How would this compare to being used? They're just signified by where they show on clients? There's otherwise no other usable functionality to gain from this, right?

src/main/java/org/spongepowered/api/item/ItemGroups.java
```diff
@@ -0,0 +1,63 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.item;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * An enumeration of all possible {@link ItemGroup}s in vanilla minecraft.
```
gabizou: Maybe phrase it as `An enumeration of the known vanilla {@link ItemGroup}s in Minecraft. Mods may change and/or introduce new groups at runtime`.

src/main/java/org/spongepowered/api/item/ItemType.java
gabizou: Quite possibly needs to be a `Collection<ItemGroup>`, since `Collections.empty()` is easy to call.

gabizou (changes requested):
Quite possibly could be named `ItemCollection` instead of `ItemGroup`. I'm impartial to either name, but it is something we can expose at least in writing but without any functionality.

src/main/java/org/spongepowered/api/item/ItemGroup.java
gabizou: Should document by answering these questions in some sense. That will fill out the javadoc for this class.

ghost:
~wip

ghost:
~qa

src/main/java/org/spongepowered/api/item/ItemGroups.java
ryantheleach: Are you attempting to expose the tab itself, for say, GUI click events. 

Or are you attempting to expose the property of items to belong to a creative tab?

ryantheleach:
@gbui Can you offer some usecases?

At the moment, without any sample usecases, or user stories, I'm not sure whether this API is sufficient or not.

I like that you have the possibility for an item to be in multiple groups.

@gabizou I'm not sold on ItemCollection as a name, it adds confusion with Java's Collections to make it seem like it's related to the Collections API / interfaces / classes.

ryantheleach:
I'm not sold on ItemCollection as a name, it adds confusion with Java's Collections to make it seem like it's related to the Collections API / interfaces / classes.

XakepSDK:
Why just not name it `ItemGroup` -> `CreativeTab`? `ItemGroup` may be confusing with 1.13 item tags.

src/main/java/org/spongepowered/api/item/ItemGroups.java
ghost: Yeah, just exposing the property of items to belong to an item group.

ghost:
@ryantheleach A sample usecase could be creating an in-game shopping catalog, organized by item group. This would make it easier for players to navigate the interface (which in this case would be an inventory "GUI").

@XakepSDK I went with `ItemGroup` since I'm just exposing the property of items to belong to an item group. `CreativeTab` is specific to the client side, while `ItemGroup` is a more general term that makes more sense on the server side.

kashike:
@XakepSDK It's worth noting that we've renamed `CreativeTabs` to `ItemGroup` in MCP.

ryantheleach:
@gbui That makes sense to me. Thanks.

PR #1869, opened 2018-07-19
Add get/applySnapshot for CommandArgs

### LINKED ISSUE #1867: Ensure resetting the state of CommandArgs reverts and inserted/removed arguments
*related to [SpongeCommon#2006](https://github.com/SpongePowered/SpongeCommon/issues/2006)*

The state of `CommandArgs` does not include the current list of arguments, so when setting the state any inserted or removed arguments from parsing will remain. The most notable example of this in use is with [CommandFlags](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/command/args/CommandFlags.java#L83), where parsed flags are removed so that they can be declared anywhere in the command. See [SpongeCommon#2006](https://github.com/SpongePowered/SpongeCommon/issues/2006) for issues that can result from this.

A workable solution for the meantime is to add a package-private `setArgs` method onto CommandArgs (as there is already a `getArgs` method as well). This can then be added to things like `GenericArguments#firstParsing` where needed on a case-by-case basis. Longer term, I believe that the object returned by `getState` needs to be updated to include this information. Alternatively, a snapshot system could be put in place, and this should supersede the use of `getState`/`setState`.

dualspiral:
* Deprecate CommandArgs#get/setState

Fixes #1867

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandArgs.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandArgs.java
@@ -26,12 +26,14 @@ package org.spongepowered.api.command.args;
 
 import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 
+import com.google.common.collect.ImmutableList;
 import org.spongepowered.api.command.args.parsing.SingleArg;
 import org.spongepowered.api.text.Text;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/command/args/CommandArgs.java
```diff
@@ -168,15 +177,31 @@ public void insertArg(String value) {
...
      */
+    @Deprecated
     public void removeArgs(Object startState, Object endState) {
         if (!(startState instanceof Integer) || !(endState instanceof Integer)) {
             throw new IllegalArgumentException("One of the states provided was not of the correct type!");
         }
-        int startIdx = (Integer) startState;
-        int endIdx = (Integer) endState;
+
+        removeArgs((int) startState, (int) endState);
```
gabizou: Wait, what?

gabizou (approved):

PR #1870, opened 2018-07-21
Make catalog type builders consistent.

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/2075)

Adds a common builder interface for catalog types to ensure consistency.All original build methods are deprecated and will be removed in bleeding.

Already makes `FurnaceRecipe` a catalog type. For stable, if no id is set, an auto generated one will be used instead.

Please let me know if I missed any catalog type builders. And is it worth to also target stable?

Impl will follow later.

### DIFF
--- a/src/main/java/org/spongepowered/api/util/CatalogBuilder.java
+++ b/src/main/java/org/spongepowered/api/util/CatalogBuilder.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.util;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.text.translation.Translation;
+
+/**
+ * A base builder to construct {@link CatalogType}s.
+ */
+public interface CatalogBuilder<C extends CatalogType, B extends ResettableBuilder<C, B>> extends ResettableBuilder<C, B> {
+
+    /**
+     * Sets the id of the {@link CatalogType} (without the namespace/plugin id).
+     *
+     * @param id The id
+     * @return This builder for chaining
+     */
+    B id(String id);
+
+    /**
+     * Sets the name of the {@link CatalogType}. Defaults to {@link #id(String)}.
+     *
+     * @param name The name
+     * @return This builder for chaining
+     */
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    B name(Translation translation);
+
+    /**
+     * Sets the id of the {@link CatalogType} (without the namespace/plugin id).
+     *
+     * @param id The id
+     * @return This builder for chaining
+     */
+    B id(String id);
```
ST-DDT: The id is the most important move it to the top.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Sets the name of the {@link CatalogType}. Defaults to {@link #id(String)}.
+     *
+     * @param name The name
+     * @return This builder for chaining
+     */
+    B name(String name);
+
+    /**
+     * Sets the name of the {@link CatalogType} as a {@link Translation}. Defaults to {@link #id}.
```
ST-DDT, line 53: Do not default to the id. We have a PR in common that actually tries to fix this.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.text.translation.Translation;
+
+/**
+ * A base builder to construct {@link CatalogType}s.
+ */
+public interface CatalogBuilder<C extends CatalogType, B extends ResettableBuilder<C, B>> extends ResettableBuilder<C, B> {
+
+    /**
+     * Sets the name of the {@link CatalogType}. Defaults to {@link #id(String)}.
```
ST-DDT: Do not default to the id. Internally you can do it but i would only do it for backwards compatibility.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    B name(String name);
+
+    /**
+     * Sets the name of the {@link CatalogType} as a {@link Translation}. Defaults to {@link #id}.
+     *
+     * @param translation The name translation
+     * @return This builder for chaining
+     */
+    B name(Translation translation);
```
ST-DDT, line 58: I'm not sure on the name here. Because on the server the name is something static usually english. However the translation is something dynamic. Maybe rename it to translation and explicitly state that the english content will be used for the name as well.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    B id(String id);
+
+    /**
+     * Builds the {@link CatalogType} of type {@link C}.
+     * <p>The last {@link PluginContainer} in the cause stack will be used to
+     * determine which plugin was used to construct the {@link CatalogType}.
+     *
+     * @return The built catalog type
+     */
+    C build();
```
ST-DDT: Maybe add some throws declarations.
ISE: Mandatory fields not set or no plugin container on the cause stack or run during the wrong game phase.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+     * Sets the id of the {@link CatalogType} (without the namespace/plugin id).
+     *
+     * @param id The id
+     * @return This builder for chaining
+     */
+    B id(String id);
+
+    /**
+     * Builds the {@link CatalogType} of type {@link C}.
+     * <p>The last {@link PluginContainer} in the cause stack will be used to
```
ST-DDT: Missing empty line before the `<p>` and missing closing tag.

ST-DDT (changes requested):

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
Cybermaxke: Some `CatalogType`s are `Translatable`, this would also apply this translation, and the plain version would be used as `getName` value.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
Cybermaxke: The id passed in here is without the namespace/plugin id, so if you provide `my_id` with `id(String)`, you would end up with a catalog type having `my_plugin_id:my_id` as id and `my_id` as name. If it's really desired to have a readable name, then I will enforce the name.

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
ST-DDT: Thats why `translation` might be an alternative.

XakepSDK: `Translarion extends Text`, iirc, why just not `Text`?

Cybermaxke: `Translation` doesn't extend `Text`, but `TranslatableText` does.

gabizou (approved):
Looks good. I like the extension and this is backwards compatible! Good changes all around.

PR #1871, opened 2018-07-21
Refactor properties.

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/2106) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/2451)

This PR is a overhaul of the `Property` API. This simplifies the implementation of new properties, separate matching behavior from the property itself and move all impl to the impl.

## Changes

- All the different property classes are now gone, there is only `Property` which contains a generic parameter representing the return type. Each property object represents a property, this was previously each property subclass. The `getProperty` method now returns the actual property value, not a value wrapped in a property object. All properties are now available in `Properties` or `InventoryProperties`.
- Removed unused property keys. Just create a new property instead.
- Added a `PropertyMatcher` which replaces the matching behavior that was previously included in the property object. This should avoid confusion between matching behavior and the actual property value.
- New properties can easily be constructed through a property builder. Properties are now also catalog types so they need to be registered.

## Examples
```java
final ItemStack itemStack = ...;
final Optional<RecordType> optRecord = itemStack.getProperty(Properties.RECORD_TYPE);
// Optional is present if the property is supported
```
```java
// No need to wrap every value in its property wrapper, just pass in the property and the value
final Inventory inventory = Inventory.builder().of(InventoryArchetypes.MENU_GRID)    
         .property(InventoryProperties.DIMENSION, new Vector2i(1, 9))
         .property(InventoryProperties.TITLE, Text.of("Slime Content"))
         .property(InventoryProperties.UNIQUE_ID, UUID.randomUUID())
         .property(InventoryProperties.GUI_ID, GuiIds.DISPENSER)
         .build(plugin);
```
```java
// Easy to create new properties
private Property<Integer> mySecretProperty;

@Listener
public void onRegisterProperties(GameRegistryEvent.Register<Property<?>> event) {
    this.mySecretProperty = Property.builder()
            .valueType(Integer.class)
            .id("my_secret_value")
            .build();
    event.register(this.mySecretProperty);
}
```


Cybermaxke:
~wip

src/main/java/org/spongepowered/api/data/property/DirectionRelativePropertyHolder.java
```diff
@@ -24,22 +24,50 @@
...
 
-import org.spongepowered.api.data.Property;
 import org.spongepowered.api.util.Direction;
 
 import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
 
-public interface DirectionRelativePropertyHolder {
+public interface DirectionRelativePropertyHolder extends PropertyHolder {
```
gabizou: We can't make them PropertyHolders because they are direction based, it's usually cases of things that literally cannot provide the property by itself

src/main/java/org/spongepowered/api/data/property/DoubleProperty.java
```diff
@@ -24,53 +24,9 @@
...
-     */
-    public DoubleProperty(Object value, Operator operator) {
-        super(Coerce.toDouble(value), operator);
-    }
-
-    @Override
-    public int compareTo(Property<?, ?> other) {
-        return this.getValue().compareTo(other == null ? 1 : Coerce.toDouble(other.getValue()));
-    }
+public interface DoubleProperty extends Property<Double> {
```
gabizou: If you're exposing `OptionalInt`s, we should consider making these return their primitive type values as well.

src/main/java/org/spongepowered/api/data/property/Properties.java
```diff
@@ -0,0 +1,384 @@
...
+
+    public static final class Block {
+
+        // SORTFIELDS:ON
+
+        /**
+         * A property to represent the blast resistance of a block.
+         */
+        public static final DoubleProperty BLAST_RESISTANCE =
+                DummyObjectProvider.createFor(DoubleProperty.class, "BLAST_RESISTANCE");
```
gabizou: Extra line difference?

src/main/java/org/spongepowered/api/data/property/PropertyHolder.java
```diff
@@ -24,38 +24,54 @@
...
      * not change the information about it's own properties either.</p>
      *
-     * @param propertyClass The property class
-     * @param <T> The type of property
-     * @return The property, if available
+     * @param property The property to retrieve the value for
+     * @param <V> The property value type
+     * @return The property value, if available
+     */
+    <V> Optional<V> getProperty(Property<V> property);
```
gabizou, line 49: So the change with usage is to just query for the property value fo the holder. 👍 How does this interact with int properties that are returning the primitive equivalent?

src/main/java/org/spongepowered/api/data/property/PropertyQuery.java
```diff
@@ -0,0 +1,143 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data.property;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * Represents a query for {@link Property} based queries.
```
gabizou: How is this used?

gabizou (changes requested):
Good structure so far, not sure what PropertyQuery does or is used for, but I'm going to allow some expanded explanation for it.

src/main/java/org/spongepowered/api/data/property/Properties.java
Cybermaxke: The lines are sometimes too long, so I moved it to a new line for everything.

src/main/java/org/spongepowered/api/data/property/DoubleProperty.java
Cybermaxke: Something like:
`int getProperty(IntProperty property, int defaultValue)`?

Cybermaxke: Then I can maybe get rid of the `OptionalInt` and `OptionalDouble` methods.

src/main/java/org/spongepowered/api/data/property/PropertyHolder.java
JBYoshi: My guess is that primitives will be autoboxed into `Integer`s and then unboxed after `Optional.get()`.

Cybermaxke: I added the following methods to get rid of boxing/unboxing of primitive property values:
```java
int getProperty(IntProperty property, int defaultValue);
double getProperty(DoubleProperty property, double defaultValue);
boolean getProperty(BooleanProperty property, boolean defaultValue);
```
I think that `int`, `double` and `boolean` would cover all the primitive properties types that are being used/will be used.

XakepSDK: String, byte, long, float arrays?

Cybermaxke: `String` isn't a real primitive, so it doesn't really need boxing/unboxing, except for the optional. `byte`, `long`, `float[]`, etc. are never used by properties.

JBYoshi: You can also use OptionalInt and OptionalDouble. OptionalBoolean doesn’t exist (in all systems at least) but a Tristate may suffice.

gabizou:
There wouldn't happen to be any implementation for this, would there?

Cybermaxke:
I started to work on it, but it's nowhere being done, doesn't compile yet. Will try to work on it this weekend.

src/main/java/org/spongepowered/api/data/type/Matter.java
```diff
@@ -22,5 +22,14 @@
...
  * THE SOFTWARE.
  */
-@org.spongepowered.api.util.annotation.NonnullByDefault
-package org.spongepowered.api.extra.fluid.data.property;
+package org.spongepowered.api.data.type;
+
+/**
+ * Represents the matter of something.
+ */
+public enum Matter {
```
liach, line 32: Implementation only?

liach (approved):

src/main/java/org/spongepowered/api/data/type/Matter.java
Cybermaxke: Why would this be impl only?

liach: This class has no apparent usage besides that it exists in a generic argument.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -658,7 +658,7 @@
      *
      * <p><em>NOTE:</em> Does not apply to {@link Player}s as their
      * {@link HandPreference} can not be changed server-side.
-     * See {@link DominantHandProperty}.</p>
+     * See {@link Properties.Entity#DOMINANT_HAND}.</p>
```
gabizou: You'll have to fully qualify this import for javadocs to properly link.

src/main/java/org/spongepowered/api/data/property/LocationBasePropertyHolder.java
```diff
@@ -25,36 +25,38 @@
...
-     * specific location. If the property is not applicable,
+     * Attempts to retrieve a value for the specified {@link Property} from
+     * the block at the specified location. If the property is not applicable,
      * {@link Optional#empty()} is returned.
      *
      * <p>{@link Property}s can define various immutable information about a
      * {@link PropertyHolder} that is dependent on the instance of the holder.
      * As {@link Property}s cannot be changed, the {@link PropertyHolder} can
      * not change the information about it's own properties either.</p>
      *
```
gabizou, line 51: May help to add a link to 
```
{@see Properties}
```
 to aid developers to look at what possible properties exist

src/main/java/org/spongepowered/api/data/property/LocationBasePropertyHolder.java
```diff
@@ -25,36 +25,38 @@
...
+import java.util.OptionalInt;
 
 public interface LocationBasePropertyHolder {
 
     /**
-     * Attempts to retrieve a specific {@link Property} type of the block at the
-     * specific location. If the property is not applicable,
+     * Attempts to retrieve a value for the specified {@link Property} from
+     * the block at the specified location. If the property is not applicable,
      * {@link Optional#empty()} is returned.
```
gabizou, line 41: Might also need to mention the possibility of directional property relations.

src/main/java/org/spongepowered/api/data/property/LocationBasePropertyHolder.java
```diff
@@ -105,29 +179,29 @@
...
-     * Gets an immutable collection of all known {@link Property}s pertaining to
-     * this {@link PropertyHolder}.
+     * Gets an immutable map of all known {@link Property}s that
+     * are supported by this {@link PropertyHolder} mapped to their value.
      *
      * <p>{@link Property}s can not be changed such that the property is
      * attached to the instance of the residing {@link PropertyHolder}.</p>
      *
      * @param coords The coordinates
      * @return An immutable collection of all known {@link Property}s
```
gabizou: s/collection/map

src/main/java/org/spongepowered/api/data/property/Property.java
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Represents a property type that can be used to
+ * retrieve immutable data from a {@link PropertyHolder}.
+ */
+@SuppressWarnings("unchecked")
+@CatalogedBy(Properties.class)
+public interface Property<V> extends CatalogType {
+
+    /**
+     * Constructs a new {@link Builder}.
```
gabizou, line 46: Add a note about calling `#valueType()` after creating the builder to get a generified version of the builder, as it's also a required first method to call when creating new properties.

src/main/java/org/spongepowered/api/data/property/PropertyRegistry.java
```diff
@@ -24,32 +24,58 @@
...
 package org.spongepowered.api.data.property;
 
-import org.spongepowered.api.data.Property;
-
-import java.util.Optional;
+import org.spongepowered.api.data.property.store.DoublePropertyStore;
+import org.spongepowered.api.data.property.store.IntPropertyStore;
+import org.spongepowered.api.data.property.store.PropertyStore;
 
 public interface PropertyRegistry {
```
gabizou: Might as well make this extend `CatalogRegistryModule` as well.

src/main/java/org/spongepowered/api/data/property/store/PropertyStore.java
```diff
@@ -81,11 +74,13 @@
...
-     * {@link PropertyStore}s have a priority of <code>100</code>.</p>
+     * that present value is returned. The default priority of a
+     * {@link PropertyStore} is <code>100</code>.</p>
      *
      * @return The priority
      */
-    int getPriority();
+    default int getPriority() {
+        return 100;
+    }
```
gabizou, line 85: May want to just have a `getProperty` for easy getter verification that you're using the right store.

src/main/java/org/spongepowered/api/data/type/Matter.java
```diff
@@ -22,5 +22,14 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-@org.spongepowered.api.util.annotation.NonnullByDefault
-package org.spongepowered.api.extra.fluid.data.property;
+package org.spongepowered.api.data.type;
+
+/**
+ * Represents the matter of something.
```
gabizou: Represents the physical matter state, usually of a {@link BlockState}.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -258,11 +266,13 @@
      *      note that upon successful insertion the supplied ItemStack itself
      *      will be reduced by the number of items successfully consumed by the
      *      Inventory
-     * @return A SUCCESS transactionresult if the entire stack was consumed and
+     * @return A SUCCESS transaction result if the entire stack was consumed and
      *      FAILURE when the stack was not or partially consumed or no Slot exists
      *      at given index.
      */
-    InventoryTransactionResult offer(SlotIndex index, ItemStack stack);
+    default InventoryTransactionResult offer(SlotIndex index, ItemStack stack) {
```
gabizou: Code smell changes, shouldn't really belong in the PR imo

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -0,0 +1,108 @@
...
+import org.spongepowered.api.item.inventory.gui.GuiId;
+import org.spongepowered.api.item.inventory.slot.SlotIndex;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.Direction;
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+import java.util.UUID;
+
+@SuppressWarnings("unchecked")
+public final class InventoryProperties {
```
gabizou, line 44: Javadoc mentioning something like:
Properties specific to `Inventory` objects that are not supported on other types of {@link PropertyHolder}s.

src/main/java/org/spongepowered/api/item/inventory/type/Inventory2D.java
```diff
@@ -44,7 +44,9 @@
      * @param pos Slot position to query
      * @return matching stacks, as per the semantics of {@link Inventory#poll()}
      */
-    Optional<ItemStack> poll(SlotPos pos);
+    default Optional<ItemStack> poll(Vector2i pos) {
```
gabizou, line 47: Likewise here, changes that aren't quite relevant to the PR.

gabizou (changes requested):
Small changes, functionality I approve.

src/main/java/org/spongepowered/api/data/property/store/PropertyStore.java
Cybermaxke: How will we handle directional properties without the other method? Or do you just mean renaming `getFor` to `getProperty`?

src/main/java/org/spongepowered/api/item/inventory/type/Inventory2D.java
Cybermaxke: I changed `SlotPos` to `Vector2i`, but I can get rid of the default impl if you want

src/main/java/org/spongepowered/api/data/property/store/PropertyStore.java
gabizou: No, as in get property type method returning Property, not the value of the property.

Cybermaxke: You don't need to know the `Property` (Type) of a store? You register a store for a specific `Property`.
https://github.com/SpongePowered/SpongeAPI/pull/1871/files#diff-f470a88385428e41ad4e5cb915ee3982R44

Cybermaxke:
https://github.com/SpongePowered/SpongeAPI/commit/15ef96be0153c0a5457b1645e0717386526af521

Issue #1885, opened 2018-08-01
Add ways to check whether player is shielding or eating something


ghost:
I couldn't find anything related to checking if player is currently shielding or eating something. It might helpful in the process of some sort of anti cheats or any other plugins creation that needs to for example calculate something differently if player is currently slowed down by either shielding or eating.

Methods (isShielding/isEating), Keys (IS_SHIELDING, IS_EATING) or any other viable ways to check that would be appreciated.

ryantheleach:
You are muddying a few different issues.

Do you 

1. want a way to see if the current speed is different.

or 

2. want a way to see if the player is eating or shielding for some other purpose? (Plugins? Spells? Damage?)

ghost:
I believe shielding and eating do change or at least should change `Keys.WALKING_SPEED` value.

As an example, I want to detect whether player is currently shielding/eating to properly block suspicious/impossible movement (like sprinting with shielding and so on). Someone from the staff showed me a workaround - listen to `UseItemStackEvent` and then do something if player is actually shielding/eating. This workaround have 2 issues though. It's over complicated for the thing that should be relatively easy to check and there are a lot of sub events and cases to consider so one mistake from me and the needed functionality will be gone.

parlough:
The PR above that I created will handle this in a more generic fashion where you can check what the active item is, so that could be a shield or food or whatever. `player.get(Keys.ACTIVE_ITEM).orElse(ItemStackSnapshot.NONE).getType().equals(ItemTypes.SHIELD)` or something along those lines.

ImMorpheus:
PR merged

PR #1888, opened 2018-08-03
Add active item data

parlough:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2042)

Fixes https://github.com/SpongePowered/SpongeAPI/issues/1885

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -104,6 +104,18 @@ public final class Keys {
      */
     public static final Key<Value<Double>> ABSORPTION = DummyObjectProvider.createExtendedFor(Key.class, "ABSORPTION");
 
+
+    /**
+     * Represents the {@link Key} for the item a {@link Living} is using.
+     * For example a player eating a food or blocking with a shield.
+     *
+     * <p>If there is no item, the snapshot will be empty. You can check this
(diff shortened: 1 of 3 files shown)


gabizou:
How does this get handled if it's set server side instead of client?

parlough:
@gabizou I simply return false if they try to set the item stack to anything besides empty. Empty does cancel it, but if they're holding right click it will just become active again.

https://github.com/SpongePowered/SpongeCommon/pull/2042/files#diff-eac4f1c95fc6a64813b25ffe4dde183dR53

Cybermaxke:
The active item feels more like a `Property` to me instead of a `Key` and value.
And maybe extend this PR to provide a active hand: `Optional<HandType>`, right know you don't know which hand is being used. With that knowledge, you can get the active item as well.

parlough:
Hmm currently properties are more things that don't really change once something is created, I think it fits better as data, especially since you can cancel an item being active, such as restarting a player drinking their milk, and other implementions may allow more. 

Relating to the active hand, I'm not really sure how useful that is, because I don't know a situation where it wouldn't be the secondary hand which can be retrieved through the dominant hand property. Then during events it should be part of the context or the event itself.

ImMorpheus:
https://github.com/SpongePowered/SpongeAPI/commit/859d917605f71aeca934c8a6095c533e1a104d2a

PR #1895, opened 2018-08-15
Add StatsConfigManager API

### LINKED ISSUE #1862: API for plugins to consult if stats collection is enabled
Passes plugin to the method, so Sponge can include each plugin separately in the config (updating as it gets new plugins), for per-plugin as well as global settings.

Code possible suggestions:

```java
Sponge.getStatsManager().isStatsEnabled(plugin);
Sponge.getPluginManager().isStatsEnabled(plugin);
Sponge.getPluginManager().getStats().isEnabled(plugin);
```

Example usage
```java
if (Sponge.getPluginManager().isStatsEnabled(this)) {
    Metrics metrics = new Metrics();
    // Stuff here
}
```

Config examples:

```json
{
  "stats": {
    "global-status": "disable",
    "per-plugin-enable": {
      "craftirc": false,
      "specatastic": false
    }
  }
}
```

```json

{
  "stats": {
    "global-status": "enable",
    "per-plugin-enable": {
      "craftirc": false,
      "specatastic": false
    }
  }
}
```

```json
{
  "stats": {
    "global-status": "per-plugin",
    "per-plugin-enable": {
      "craftirc": true,
      "spectastic": false
    }
  }
}
```

### Other thoughts

* Plugin devs will want a way to allow easy enabling
  * Proposal: General Sponge command. `/sponge stats <enable/disable> <plugin>`
    * This way, plugins could say (once, preferably) "Please consider enabling stats, run /sponge stats enable craftirc"
* There has been discussion of making a full stats service API. This could be part one of that.

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2060) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2382) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/385) | [Original Issue](https://github.com/SpongePowered/SpongeAPI/issues/1862)

This adds an API to check to see if a plugin is allowed to collect stats as per the user's config. I opted for only allowing a plugin to check to see if stats collection is allowed for that plugin, it shouldn't need to know about the global entry.

**This PR is not for discussion of whether stats are bad or not, Sponge's policies, or any general gripe. This is simply the proposed API for discussion. Please leave drama out of this PR, thanks.**

Fixes #1862

### DIFF
--- a/src/main/java/org/spongepowered/api/service/metric/MetricsConfigManager.java
+++ b/src/main/java/org/spongepowered/api/service/metric/MetricsConfigManager.java
@@ -22,7 +22,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.service.stat;
+package org.spongepowered.api.service.metric;
 
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.plugin.PluginContainer;
@@ -31,25 +31,25 @@ import org.spongepowered.api.plugin.PluginContainer;
  * Provides information about whether a server has granted permission for
  * server data to be transmitted on a per plugin basis.
  */
-public interface StatsConfigManager {
+public interface MetricsConfigManager {
 
     /**
-     * Gets whether permission for third-party statistics gathering has been
+     * Gets whether permission for third-party metrics gathering has been
      * granted for a specific plugin (or the global permission if the plugin
      * has not got a specific permission or lack thereof).
      *
      * <p>The value returned from this <em>should not be stored</em>. As the
      * configuration/permission can be updated at any time, it is best to
-     * check this each time stats collection is due to occur.</p>
+     * check this each time metrics collection is due to occur.</p>
      *
      * @param container The {@link PluginContainer} to check for permission for
-     * @return true if statistics gathering plugins have permission to gather
+     * @return true if metrics gathering plugins have permission to gather
      *         data about this plugin
      */
-    boolean areStatsEnabled(PluginContainer container);
+    boolean areMetricsEnabled(final PluginContainer container);
 
     /**
-     * Gets whether permission for third-party statistics gathering has been
+     * Gets whether permission for third-party metrics gathering has been
      * granted for a specific plugin (or the global permission if the plugin
      * has not got a specific permission or lack thereof).
      *
@@ -61,11 +61,11 @@ public interface StatsConfigManager {
      *               {@link org.spongepowered.api.plugin.Plugin})
      * @throws IllegalArgumentException if the supplied object is not a plugin
      *                                  object
-     * @return true if statistics gathering plugins have permission to gather
+     * @return true if metrics gathering plugins have permission to gather
      *         data about this plugin
      */
-    default boolean areStatsEnabled(Object plugin) throws IllegalArgumentException {
-        return areStatsEnabled(Sponge.getPluginManager().fromInstance(plugin)
+    default boolean areMetricsEnabled(final Object plugin) throws IllegalArgumentException {
+        return this.areMetricsEnabled(Sponge.getPluginManager().fromInstance(plugin)
                 .orElseThrow(() -> new IllegalArgumentException("The supplied object is not a plugin object.")));
     }
 
--- a/src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
+++ b/src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.stat;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.plugin.PluginContainer;
+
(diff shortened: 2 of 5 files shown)


src/main/java/org/spongepowered/api/service/stat/package-info.java
```diff
@@ -0,0 +1,26 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+@org.spongepowered.api.util.annotation.NonnullByDefault
+package org.spongepowered.api.service.stat;
```
ST-DDT: Maybe add some package javadocs.

ST-DDT (approved):

src/main/java/org/spongepowered/api/service/stat/package-info.java
Cybermaxke: Maybe `org.spongepowered.api.service.stats` instead of `org.spongepowered.api.service.stat`?

src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
```diff
@@ -0,0 +1,72 @@
...
+package org.spongepowered.api.service.stat;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.plugin.PluginContainer;
+
+/**
+ * Provides information about whether a server has granted permission for
+ * server data to be transmitted on a per plugin basis.
+ */
+public interface StatsConfigManager {
```
Zidane: This should simply be StatsService

ST-DDT: I would like to reserve the StatsService name for the actual statistic collector service.

src/main/java/org/spongepowered/api/service/stat/package-info.java
dualspiral: I got told somewhere that we should use singular names for packages - which is why I went for `stat`. I'm happy either way, but I'd like to make sure that I'm not going to be playing ping-pong with the letter s.

src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
dualspiral: I can change it, but I named it a `Manager` because it's not intended to be replacable. It's not registered in the service manager either (though it could be). It also doesn't handle stats itself, just whether the config gives it permission, which is why I went for `StatsConfigManager`. Could go for `StatsPermissionService` or something if people prefer.

gabizou: Maybe StatPermissionAuthority? It implies it’s an authority on permissions for statistics collections. One Manager away from having SPAM in the api ;)

Zidane: If that is the case, call it what you like but move it out of service package. Service package should imply things that are replaceable.

ST-DDT: If we call it like Permission then why not remove it entirely and use the `PermissionService` with a `SubjectType#PLUGIN` directly?
The plugin id could be the subject id and its even extensible for whatever plugin permissions we are going to add in the future (like `download.file` or whatever).

Even if we dont merge it with the permission service. The more complex/longer the name becomes the more I am in favor of just calling it `PluginPermission...` and allow it to be at least extensible for future additional plugin permissions. This would also make it easier to decide on a package for this class as `plugin` would be a good choice.

dualspiral: That's basically going to add a layer onto the steps that server admins would have to follow, and would basically _require_ a permission plugin to opt-in/out. We just need to make it easy for users to opt-in and be done with it.

All this is, and needs to be, is a simple config check. `StatAuthorityManager` would work as a name (as much as I like having the acronym SPAM in the API...), just got to decide the package. Any ideas welcome, I do not have much time to spend on this tonight, so it'll get done over the next day or so.

I'd rather get this straightened out before I have many commits that simply change the name of things.

src/main/java/org/spongepowered/api/service/stat/package-info.java
Cybermaxke: In this case `stats` sounds better than `stat`, and it makes also more sense when you have a
`StatsConfigManager` instead of `StatConfigManager`

src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
dualspiral: What about putting this in the `util` package? I thought about the `config` package, but that's for managing configs, not exposing Sponge configs.

Bit of a loss on where to put this otherwise.

dualspiral: kashike suggested `o.s.a.util.snooper` - avoiding the `stat` vs `stats` argument.

simon816:
I'd call it "Metrics" rather than "Stats" to avoid confusion with the [Statistics API](https://github.com/SpongePowered/SpongeAPI/tree/stable-7/src/main/java/org/spongepowered/api/statistic)

src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
```diff
@@ -0,0 +1,72 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.stat;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.plugin.PluginContainer;
+
+/**
+ * Provides information about whether a server has granted permission for
+ * server data to be transmitted on a per plugin basis.
```
gabizou: Just going to add that this should be `@Inject`able, and should say so on the javadocs.

src/main/java/org/spongepowered/api/service/stat/StatsConfigManager.java
gabizou: As per yesterday's discussion, we're keeping `StatsConfigManager` I may bend if we want to say `StatCollectionConfigManager` just to make it clearer it's about statistics collection.

gabizou (approved):
Looks good.

gabizou:
Reviewed API and implementation. Awesome work! I'll get to merging this this afternoon if someone else doesn't merge it first.

kashike:
I'd like to bump @simon816's comment:
> I'd call it "Metrics" rather than "Stats" to avoid confusion with the [Statistics API](https://github.com/SpongePowered/SpongeAPI/tree/stable-7/src/main/java/org/spongepowered/api/statistic)

Metrics, Snooper, ...

dualspiral:
It was discussed at SPAM, I think we ended up sticking with "Stats"...

edit: wait, that was said by Gabi. Never mind...

dualspiral:
(Also, I'm not gonna get to doing this myself tonight, so gabi, be my guest to merge it)

src/main/java/org/spongepowered/api/service/metric/MetricsConfigManager.java
```diff
@@ -0,0 +1,72 @@
...
+    boolean areMetricsEnabled(final PluginContainer container);
+
+    /**
+     * Gets whether permission for third-party metrics gathering has been
+     * granted for a specific plugin (or the global permission if the plugin
+     * has not got a specific permission or lack thereof).
+     *
+     * <p>The value returned from this <em>should not be stored</em>. As the
+     * configuration/permission can be updated at any time, it is best to
+     * check this each time stats collection is due to occur.</p>
```
Cybermaxke, line 58: `stats` -> `metrics`

PR #1896, opened 2018-08-17
Extent API 2.0

gabizou:
Heavily work in progress. 100% pending starting 1.13 migration. This is to properly ~smash with a hammer~ refactor Extent with better support for non `World`... extents... Yeah.... 

Our Extent hierarchy is problematic, we associated all `World` related methods to `Extent` including all the hierarchies of what is definitively a `World`, and then made `Chunk` extend `Extent`. Likewise, we decided early on that we would have this abstract `Location` that would be able to accept any `Extent`, along with all of the `DataHolder` methods relegated to the parented `Extent`...

You see where this is starting to become a problem, especially when #734 was being designed, new volume types had to be created #1137,  to support #1330... 

I'm intending to royally break the hierarchy and allow for more abstract types that will support the ability to work on more concrete volumes that aren't ~bloated~ swelling with extraneous methods that are more `World` centered. In doing so, this also allows a simultaneous rewrite of the world generation API to use a new world generation pipeline being introduced by Minecraft itself. The awesome thing about this is that the world generation pipeline should allow for some rather interesting uses for sponge plugins.

I'll do some code examples and usages etc. as the interfaces are more "implemented"


src/main/java/org/spongepowered/api/world/extent/EntityHit.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.flowpowered.math.vector.Vector3d;
+import org.spongepowered.api.entity.Entity;
+
+/**
+ * The result of an intersection between a ray and an entity.
+ */
+public class EntityHit {
```
Zidane: @gabizou Why is this in the API (or at the least as a concrete class?).

gabizou: It has been since it was created. I just moved it out of `EntityUniverse`, until I can figure out a good way to adapt AABB and expose VoxelShapes for better collision detection and raytracing.

src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -582,7 +582,7 @@ public boolean setBlockType(BlockType type) {
      * @return True if the block change was successful
      */
     public boolean setBlockType(BlockType type, BlockChangeFlag flag) {
-        return getExtent().setBlockType(getBlockPosition(), type, flag);
+        return getExtent().setblock(getBlockPosition(), type.getDefaultState(), flag);
```
Cybermaxke: `setBlock`

build.gradle
```diff
@@ -154,10 +154,10 @@ sortClassFields {
...
     add 'main', 'org.spongepowered.api.block.BlockTypes'
     add 'main', 'org.spongepowered.api.block.tileentity.TileEntityTypes'
-    add 'main', 'org.spongepowered.api.block.trait.BooleanTraits'
-    add 'main', 'org.spongepowered.api.block.trait.EnumTraits'
-    add 'main', 'org.spongepowered.api.block.trait.IntegerTraits'
-    add 'main', 'org.spongepowered.api.block.trait.EnumTraits'
+    add 'main', 'org.spongepowered.api.state.BooleanProperties'
+    add 'main', 'org.spongepowered.api.state.EnumProperties'
+    add 'main', 'org.spongepowered.api.state.IntegerProperties'
+    add 'main', 'org.spongepowered.api.state.EnumProperties'
```
Cybermaxke, line 160: Duplicate entry.

src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -83,11 +83,11 @@
  * <p>Locations are immutable. Methods that change the properties of the
  * location create a new instance.</p>
  *
- * @param <E> The type of extent containing this location
+ * @param  The type of world containing this location
  */
-public final class Location<E extends Extent> implements DataHolder {
+public final class Location implements DataHolder {
```
kashike: Uhhhh, why???

gabizou: Location as it's written currently can't use anything except World and maybe Chunk. Likewise, I'm moving Chunk away from a LocataionBasedDataHolder etc. and only making `ProtoWorld` implement that (since `ProtoWorld` can be recycled as a `GenerationRegion` of sorts as world gen gets fleshed out/refactored, and we'll have things extending `ProtoChunk`).

ryantheleach:
Posting to draw attention to: https://github.com/SpongePowered/SpongeAPI/issues/1932

Offline location classes have popped up quite a few times in chat, people loading and unloading many worlds typically run into this problem.

In my mind, Location itself (if it is to remain the unit extent, I'm not sure it should any more after learning  more about what that means) shouldn't be directly deserializeable.

Locations being applicable to other extents would be handy, I've experimented in the past attempting to create them for arena's to have arena relative co-ordinates.

But unless registered in a registry of UUID, I don't see how they could be deserialized nicely.

Design opinions for a replacement for a Config friendly location.

1. The 'context' of a location should be up to the plugin responsible for deserialization, they can add their own identifiers easily enough, but we could make this easier.

e.g.

A (Generic world location):
  `spawn: {x: 1, y: 1, z: 1,  world: "1234-2345-2345-456"}`
B (Generic world location + player rotation:
  `exactSpawn: {x: 1, y: 1,z: 1, yaw:1, pitch:1, world: "1234-2345-2345-456"}`
C Generic world location, specified by directory name
  `TNTEventSpawn: {x: 1, y: 1, z: 1,  world: "world-Copy"}`
D Generic Arena location, specified by resource location.
  `ballSpawn: {x: 1, y: 1, z: 1,  "minigameFramework:arena": "myMiniGame:Soccer1"}`
E Nether Fallback Spawn
  `hellSpawn: {x: 1, y: 1, z: 1,  "dimension": -1}`
F Target teleport location
  `comeToMe: {"entity":1234-2345-2345-456, "world:"franksWorld"}` // defaults to the position of that entity if found.
G Teleport to spawn
  `teleportTarget: {world:"franksWorld"}` //defaults to the world spawn point.
H Anonymous Location
  `cachedLocation: {x: 1, y: 1,z: 1, yaw:1, pitch:1}`


and Looking at the minecraft teleport command for inspiration, it's possible for there to be even yet more examples that people use on a regular basis.

What's common, is that depending on the usecase and plugin consuming the location, the context can vary wildly, and location resolving is **never** guaranteed to resolve. and that resolving a location may end up with a dynamic position that changes over time.

This would show that locations should be resolved as late as possible, and if failing, all data should be made available to the plugin.

What remains is how far should we take this extensibility, before it gets silly.

How can we design pre-built locations for most of the (common) use cases, whilst still allowing others?

How can we make this compatible between plugins, without too much effort on each plugin relying upon unique properties for each plugin?

How do we gracefully fail, and allow plugins to provide fallback locations based on the data provided by the user/config.

**Basic Usage:**
```
NewLocation loc = config.getLocation();
LocationResult target = Sponge.resolveLocation(targetPlayer, loc); //specify player for anonymous locations, or things that require a world context.
teleportee.setLocation(loc); //Will attempt to resolve, but will throw if failed.
teleportee.setLocation(target); // Will throw if LocationResult failed.
Location loc = target.get(); //Will throw if LocationResult was a failure
```

**Safe Usage:**
```
target.teleport(teleportee, fallbackLocation); // Will teleport to the target, or the fallback position.

if(!target.success){
  if(checkUnloaded(target){ // Checks to see if target failed due to an unloaded world, that can be loaded.
    LoadWorld(target.world.orElse(target.entity.world)); // world is optional, as it may not have resolved.
    //or alternately?
    LoadWorld(target.resolve(target.world)
  }
    target.teleport(teleportee, fallbackAction);
}
```

**Possible Advanced usages:**
```
LocationResult target = Sponge.resolveLocation(targetPlayer, loc, fallbackLocation); //result will still be a failure, but .get() will not throw.

target.teleport(teleportee, Function<LocationResult, LocationContext, Location> failResolver); // Will teleport to the target, or the provided fallback position, unless failAction returns null.
```

gabizou:
Manualy merging in 1.13 to keep it inline with API development since this introduces Fluid handling as well as other important base object changes (like Location).

randombyte-developer:
Since this is closed and being manually merged: I just want to make sure https://github.com/SpongePowered/SpongeAPI/issues/1932 isn't getting lost. This is a breaking change and would fit in here well.

PR #1898, opened 2018-08-18
UpdateAnvilEvent

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2084)

Adds an event to handle anvil results and costs.

The event fires before vanilla is doing its computation but after forge mods had a chance to set the values.

As such there is no initial level/material cost and the result is empty by default.
Only when the result is set by the plugin costs will be applied.

No result set => Vanilla Behaviour
Result set => Override Costs and Result
Cancel => Empty Result

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
@@ -1,0 +1,74 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.item.inventory;
+
+import org.spongepowered.api.data.Transaction;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.item.inventory.AnvilCost;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
+
+/**
+ * Fires whenever the left and right slots of an anvil are filled and a new result is computed.
+ */
+@GenerateFactoryMethod
+public interface UpdateAnvilEvent extends TargetInventoryEvent, Cancellable {
+
+    /**
+     * Returns the new item name.
+     *
+     * @return The new item name
+     */
+    String getItemName();
+
+    /**
+     * Returns the left input item.
+     *
+     * @return The left input item
+     */
+    ItemStackSnapshot getLeft();
+
+    /**
+     * Returns the right input item.
+     *
+     * @return The right input item
+     */
+    ItemStackSnapshot getRight();
+
+    /**
+     * Returns the transaction for the resulting item.
+     *
+     * @return The transaction for the resulting item.
+     */
+    Transaction<ItemStackSnapshot> getResult();
+
+    /**
+     * Returns the transaction for the repair costs.
+     *
+     * @return The transaction for the repair costs.
+     */
+    Transaction<AnvilCost> getCosts();
+
+}
--- a/src/main/java/org/spongepowered/api/item/inventory/AnvilCost.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/AnvilCost.java
@@ -1,0 +1,64 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/AnvilUpdateEvent.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return The right input item
+     */
+    ItemStackSnapshot getRight();
+
+    /**
+     * Returns the custom result.
+     *
+     * @return The custom result or {@link Optional#empty()} when not set
+     */
+    Optional<ItemStack> getResult();
```
gabizou: Couldn’t this be a transaction?

src/main/java/org/spongepowered/api/event/item/inventory/AnvilUpdateEvent.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param item The custom result
+     */
+    void setResult(@Nullable ItemStack item);
+
+    /**
+     * Returns the custom level cost.
+     *
+     * @return The custom level cost
+     */
+    int getLevelCost();
```
gabizou: Likewise this too.

src/main/java/org/spongepowered/api/event/item/inventory/AnvilUpdateEvent.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Fires whenever the left and right slots of an anvil are filled
+ * and a new result is computed.
+ *
+ * <p>Canceling this event prevents the vanilla behaviour and the
+ * output is set to empty</p>
+ *
+ * <p>Changed costs only apply when the result is also set</p>
+ */
+public interface AnvilUpdateEvent extends TargetInventoryEvent {
```
gabizou: Flip the name. It’s ‘ActionTargetEvent’ in this case UpdateAnvilEvent.

gabizou (changes requested):
Minor changes.

src/main/java/org/spongepowered/api/event/item/inventory/AnvilUpdateEvent.java
Faithcaio: If we fire the event where forge does it too then there is no default result/costs for the transaction

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return The right input item
+     */
+    ItemStackSnapshot getRight();
+
+    /**
+     * Returns the custom result.
+     *
+     * @return The custom result or {@link Optional#empty()} when not set
+     */
+    Optional<ItemStackSnapshot> getResult();
```
gabizou: I still say this should be a transaction, because the result is always available, even if the `ItemStack` is empty, in fact, that's what the implementation will do... it doesn't deal with a null item stack in the slot.

What I'd say the transaction would look like is the following:
- `getOriginal()` would almost always be empty, or whatever is currently in the slot
- `getDefault()` would have to simulate the vanilla behavior (I know, totally separate from what Forge does, more on that later)
- `getCustom()` would be the plugin provided one, if available
- `getFinal()` would be as `Transaction` already defines, if the custom is set, returns custom, otherwise return default. 

My reasoning for this:

Forge implements this event as more of a "pre" event where the desired output can be set, and likewise the costs. My guess for the original intention was due to how anvil repairs/enchantments are not jsonified, and the event predated the start of jsonifying/configuring/registering of things like normal-esque recipes. Part of the problem here is that you're exposing directly the forge event as the forge event was written, which doesn't really provide good extensibility from a plugin's perspective (imagine if a plugin wanted to modify the costs, output, and the ingredients provided, like a reactive item that would be "locked" or modified prior to the transaction. I know, it's a little awkward since the event is intended to be thrown whenever the anvil *container* is being told to refresh, which can be a little spammy.

I personally think we can and should be able to represent the playback of the information, whether it's done through an overwrite, or whether we do a couple of injectors/redirects.

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param item The custom result
+     */
+    void setResult(@Nullable ItemStackSnapshot item);
+
+    /**
+     * Returns the custom level cost.
+     *
+     * @return The custom level cost
+     */
+    int getLevelCost();
```
gabizou: As per above comment, if the event is to become a post calculation, this should be either a transaction, or have `getOriginalLevelCost()`.

gabizou (changes requested):

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,106 @@
...
+     * @return The transaction for the resulting item.
+     */
+    Transaction<ItemStackSnapshot> getItem();
+
+    /**
+     * Returns the level cost.
+     *
+     * @return The level cost
+     */
+    int getLevelCost();
```
gabizou: This too can be a transaction of integers (sure, it's not as clean as just primitive ints, but meh)

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param levelCost The new level cost
+     */
+    void setLevelCost(int levelCost);
+
+    /**
+     * Returns the material cost.
+     *
+     * @return The material cost
+     */
+    int getMaterialCost();
```
gabizou: Likewise this.

gabizou (changes requested):
Pretty much done, just move the other two groups of methods into a transaction and you're pretty much set.

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
Faithcaio: https://github.com/SpongePowered/SpongeAPI/blob/a1418ae3fccbfa1fbc56e05d2f3375dcd1344e1c/src/main/java/org/spongepowered/api/data/Transaction.java#L34
needs DataSerializable

Faithcaio: so like this now?

gabizou (approved):

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,108 @@
...
+     * Returns the transaction for the repair costs.
+     *
+     * @return The transaction for the repair costs.
+     */
+    Transaction<UpdateAnvilEvent.Cost> getCosts();
+
+    /**
+     * The level and material costs for an anvil usage.
+     */
+    interface Cost extends DataSerializable {
```
Cybermaxke: I would prefer this class to be immutable with a builder. That way is the transaction also completely immutable, so you cannot modify the original costs.

Faithcaio: How about immutable without a builder like it is now?

Cybermaxke: Oh, I didn't notice that it returns a new object. Javadoc is missing `@return`

Ant it's a common practice to use builders in sponge, not sure if it's worth adding one for this.

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,112 @@
...
+         *
+         * @return a modified copy of the Costs
+         */
+        Cost withLevelCost(int levelCost);
+
+        /**
+         * Returns a modified copy of the Costs
+         *
+         * @param materialCost The new material cost
+         *
```
Cybermaxke: No need for this line, and `A` instead of `a`

src/main/java/org/spongepowered/api/event/item/inventory/UpdateAnvilEvent.java
```diff
@@ -0,0 +1,110 @@
...
+     * @return The right input item
+     */
+    ItemStackSnapshot getRight();
+
+    /**
+     * Returns the transaction for the resulting item.
+     *
+     * @return The transaction for the resulting item.
+     */
+    Transaction<ItemStackSnapshot> getItem();
```
Cybermaxke: `getResult` would be more fitting

PR #1900, opened 2018-08-27
Update Javadocs for commands

### LINKED ISSUE #1892: Better javadocs for CommandElements, especially those which can return multiple instances
The documentation for many of the `CommandElement`s should be updated to inform the programmer that they are capable of returning multiple instances. In some cases, the vocabulary used suggests directly that only a single instance can be returned.

This is especially important as the method signature gives no information about the type(s) it stores (as it just returns a `CommandElement`).

For example:
playerdocsfial

It is inconvenient to keep referring back to the Sponge Docs to clarify. GenetricArguments.player may match with multiple players and an `onlyOne` wrapper may need to be used.

dualspiral:
Should make what is returned by a command a little clearer, just wanted to try to get this in for 7.1 so the updated descriptions fit in the docs.

No code changes, but as I tend to be a little flowery with my language at times, a proofread is what I'm looking for.

Fixes #1892

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
@@ -298,9 +298,14 @@ public final class CommandFlags extends CommandElement {
         return null;
     }
 
+    /**
+     * Indicates to the flag parser how it should treat an argument that looks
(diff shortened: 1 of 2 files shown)


gabizou (approved):
Nothing to complain about.

PR #1902, opened 2018-08-29
Expose static methods for plugins to detect whether they're running on Forge or Vanilla

ImMorpheus:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2082) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2417)

See common PR for related issue.

### DIFF
--- a/src/main/java/org/spongepowered/api/Sponge.java
+++ b/src/main/java/org/spongepowered/api/Sponge.java
@@ -280,4 +280,8 @@ public final class Sponge {
         return check(metricsConfigManager);
     }
 
+    public static boolean isVanilla() {
+        return true;
+    }
+
 }


kashike:
Why?

ImMorpheus:
> Why?

Quoting from the related issue on the SpongeCommon PR:

> it would be useful for plugins to have some static methods to detect the environment they're running in before the game has actually started.
> 
> They can be used to apply a certain mixin only to either SpongeForge or SpongeVanilla but not both (similar to what we do currently in SpongeVanilla/SpongeForge).

This was already implemented. I've moved it from `SpongeImplHooks` (see [SpongeImplHooks](https://github.com/SpongePowered/SpongeCommon/pull/2082/files)) to expose it to plugins.

XakepSDK:
What should return Lanternpowered server?

mosemister:
I was thinking the same.

And other implementations.

ImMorpheus:
> What should return Lanternpowered server?

True.
Lantern is a minecraft server.

isVanilla being false just means this is a Forge Server. 
The issue is [...] **detect whether they're running on Forge or Vanilla** [...], guess I should have linked it here.

mosemister:
I personally think that it would be useful for the mixin developers (I am not one myself) but I don't think that this is the way.
Personally I think there needs to be a single function that returns the implementation name even on PreInitialisation.

RedNesto:
@mosemister You should take a look at `Sponge#getPlatform`, you will be able to get the implementation PluginContainer.

Also, I think the method needs some documentation to explain when it should return what.

mosemister:
@RedNesto
I already knew that. Due to the fact im not a mixin guy I thought that wasnt available in the PreInitialisation state which is where I believe mixins are required to be initialized. If it is then I dont see the point for exposing this at all

Cybermaxke:
`Sponge.getPlatform().getContainer(Component.IMPLEMENTATION).getId() == "spongeforge"` can be easily added to a plugin if really needed.

The thing is that vanilla/modded is a blurred line, if you add a plugin to a sponge server, is it still vanilla? And even with lantern isn't even "vanilla", because you will never be able to replicate the features exactly. Another example, is forge without mods vanilla? Or is a forge server without client mods vanilla? What about server side mods?

Basically "vanilla" is minecraft without amything else, the moment sponge/forge comes into play it's no longer "vanilla". Unless you only use optimizations (and they aren't game changing).

XakepSDK:
@ImMorpheus but lantern isn't vanilla server. Someone may use `isVanilla` for some NMS hacks.

Cybermaxke:
The SpongeAPI isn't a place for `NMS hacks`, if you want that, depend on `SpongeCommon` instead. You can add the method there if you really need it. This breaks compatibility with any non NMS based impl.

XakepSDK:
And how you force someone to *not* to do these hacks?

Cybermaxke:
You don't. But we shouldn't encourage them either by exposing them in the API.

mosemister:
@XakepSDK
I think NMS is a little off topic. But one of the underlying principles of Sponge was to make a API that lacked the need for NMS to be done by exposing the NMS features in API.

ImMorpheus:
I can see the ambiguity with the "vanilla" term. 

Input wanted, whether this should be closed (with the related issue) or renamed to something else ("supportMods" or smth).

dualspiral:
Ultimately, why should a plugin that purely relies on the Sponge API alone care about this? This is exposing an implementation detail - and checking whether the implementation is a Vanilla one shouldn’t matter when the Sponge system is agnostic to this.

Just because it exists in Common doesn’t mean it should exist in the API.

So, to go back to @kashike’s question - why is this useful? Why should a non-implementation dependent plugin care about this? @Cybermaxke has already talked about the platform, surely that gives you everything you need to know already?

_edited to make a point clearer, posting from my phone!_

ImMorpheus:
That's what I wanted to hear. 

I'm closing this (and the related issue) as "invalid".


Just a sligthly unrelated nitpicking:

> Just because it exists in Common doesn’t mean it should exist in the API.

I've opened this PR because there was a feature request in common.

dualspiral:
> I've opened this PR because there was a feature request in common.

I think something needs to be clarified here as to what that issue was asking for.

The problem is that the [request was in _SpongeCommon_](https://github.com/SpongePowered/SpongeCommon/issues/804) and should be solved in SC because it's an implementation detail. It was open to explore exposing whether the impl was Vanilla to those that use the MCP implementation (that is, ours). Plugin-provided mixins intentionally depend on the implementation, so this doesn't need to be in the API layer. As such, it was never intended to hit the API, I'm sure Minecrell would have opened it in this repo if it was.

This PR probably would not have solved that issue because all you're doing is moving a method - if it was that easy, the original issue would not have been created. It might be that the original issue is no longer valid, it probably needs checking in the context it is presented in.

ryantheleach:
This is needed as per the original ticket.

It's needed to be done in it's own class outside the rest of SpongeAPI so that mixins / transformers DO NOT load the rest of SpongeAPI / SpongeCommon too early.

It NEEDS to be in the API, otherwise it's impossible for other implementations to implement, and have plugins be potentially compatible with other platform internals / extended API.

isVanilla() isn't enough in my eyes, and it also needs to be somewhere where it won't chain class load everything else.

The other option is just to rely on dependency resolution, but that may be awkward for some situations.

dualspiral:
Again, I never said the original issue wasn’t needed. I said this is the wrong place for the solution.

> mixins / transformers

These two words explain exactly why it does NOT go in the API. When you use mixins or transformers, you’re intentionally interacting with the implementation. In particular, this is wanted before the transformation takes place, particularly with Mixins - this makes a determination whether a mixin should be applied **before any mixins are applied**, meaning for the use case in the original issue, this would **always** say the impl was vanilla at the point it would most likely be checked.

(Aside: this PR would not have worked because of this fact, it relies on a transformation of an API method but the info would be needed before this happened)

Plugin provided mixins are designed as an MCP impl feature. There is no support for them in the Sponge API, the API and its dependency set makes **no guarantee** that you will be able to use them. I expect that Lantern doesn’t support mixins, why would it want to use them if it has control of the full stack?

The API is a runtime object that can already provide this information when needed at runtime, though a plugin generally does not need this information. The problem that needs to be solved is an implementation level, specifically _our_ implementation. Any mixins or transformers need, as you rightly say, to know WAY more information than just if it’s vanilla, and the API will not be ready at the time this info is needed.

It doesn’t need to be in the API layer, we should not pollute the layer with implementation specific calls. By making a call rely on a Mixin, we are creating a dependency (even if the dependency doesn’t move to the API) on Mixin. 

The solution to this needs to be in Common, for our impl. There is nothing stopping there being class or system outside the rest of SC but within the SC jar that does this, no need for the API to contain it.

ryantheleach:
I'll move my reply to the API issue, since it's been reopened, but originally I replied to this, since it was relating to it being implemented on the API side.

PR #1903, opened 2018-08-31
Add features specific to Shulkers, ShulkerBullets and ShulkerBoxes.

### LINKED ISSUE #1901: ShulkerBullet not implemented in Projectile.
When writing a plugin, I ran into a problem. I need to get the entity that fired the shulker bullet.

RedNesto:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2087)

This PR adds everything I found for Shulkers and its derived products.

I only have a doubt about `ShulkerBullet#getDirection`, as it may be a DirectionalData.

Reviews are welcome.

[Test Plugin](https://gist.github.com/RedNesto/6fdd0d1e8ce66a52374432e1544c43ba)

Fixes #1901

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -2103,7 +2103,7 @@ public final class Keys {
      *
      * @see TargetedEntityData#value
      */
-    public static final Key<Value<Entity>> TARGETED_ENTITY = DummyObjectProvider.createExtendedFor(Key.class,"TARGETED_ENTITY");
+    public static final Key<Value<EntitySnapshot>> TARGETED_ENTITY = DummyObjectProvider.createExtendedFor(Key.class,"TARGETED_ENTITY");
 
     /**
      * Represents the {@link Key} for the location targeted by an
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -609,7 +609,7 @@ public final class Keys {
 
     /**
      * Represents the {@link Key} for representing the {@link Direction}
-     * of a {@link BlockState}.
+     * of a {@link BlockState} or an {@link Entity}.
      *
      * @see DirectionalData#direction()
      */
@@ -2096,6 +2096,13 @@ public final class Keys {
      * @see TameableData#owner()
      */
     public static final Key<OptionalValue<UUID>> TAMED_OWNER = DummyObjectProvider.createExtendedFor(Key.class,"TAMED_OWNER");
+
+    /**
+     * Represents the {@link Key} for a targeted entity, like by a {@link ShulkerBullet}
+     *
+     * @see TargetedEntityData#value
+     */
+    public static final Key<OptionalValue<Entity>> TARGETED_ENTITY = DummyObjectProvider.createExtendedFor(Key.class,"TARGETED_ENTITY");
 
     /**
      * Represents the {@link Key} for the location targeted by an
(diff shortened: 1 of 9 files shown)


RedNesto:
~qa

XakepSDK:
Links are broken here and in common

src/main/java/org/spongepowered/api/entity/ShulkerBullet.java
```diff
@@ -24,6 +24,20 @@
...
+ * A {@link ShulkerBullet} is the projectile {@link Shulker}s shoot.
+ */
+public interface ShulkerBullet extends EntityTargetingProjectile, DamagingProjectile {
+
+    /**
+     * The direction this {@link ShulkerBullet} is going to.
+     *
+     * @return the direction of this bullet.
+     */
+    Direction getDirection();
```
Cybermaxke: Can be default implemented.

RedNesto: What should I put in the default implementation? There is no data for it in Entity.
And I'm wondering if it would be better to expose the Direction through the Data API instead of a getter.

Cybermaxke: Oh, I thought you already exposed DirectionalData for the bullet as well

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -2097,6 +2097,13 @@
      */
     public static final Key<OptionalValue<UUID>> TAMED_OWNER = DummyObjectProvider.createExtendedFor(Key.class,"TAMED_OWNER");
 
+    /**
+     * Represents the {@link Key} for a targeted entity, like by a {@link ShulkerBullet}
+     *
+     * @see TargetedEntityData#value
+     */
+    public static final Key<OptionalValue<Entity>> TARGETED_ENTITY = DummyObjectProvider.createExtendedFor(Key.class,"TARGETED_ENTITY");
```
gabizou: Why is this an optional value? If there's no targeted entity, then the value shouldn't exist. Inversely, the value should only be returned as a valid value if there's a targeted entity (so you can support removing the targeted entity).

src/main/java/org/spongepowered/api/entity/ShulkerBullet.java
gabizou: Can very well expose directional data on the bullet.

gabizou (changes requested):

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -2097,6 +2097,14 @@
      */
     public static final Key<OptionalValue<UUID>> TAMED_OWNER = DummyObjectProvider.createExtendedFor(Key.class,"TAMED_OWNER");
 
+    /**
+     * Represents the {@link Key} for a targeted entity,
+     * like by a {@link ShulkerBullet}.
+     *
+     * @see TargetedEntityData#value
+     */
+    public static final Key<Value<Entity>> TARGETED_ENTITY = DummyObjectProvider.createExtendedFor(Key.class,"TARGETED_ENTITY");
```
gabizou: I'm not a fan of reintroducing data targeting live `Entity` instances, Data API doesn't do well with it since it's not an immutable variation. When we handled targeting for entity AI, we elected to not represent that via Data API. I'm not against it, but I'm not a fan.

gabizou (approved):
Leaving a decision for @Cybermaxke to make about targeted entity pointing to `Entity` references. Otherwise, the PR looks fine.

PR #1911, opened 2018-09-18
Add BEETROOT_AGE integer trait

Grinch:


gabizou:
Just a note that this will not be available in implementation versions until implementations have a recommended build.

PR #1915, opened 2018-09-23
Inventory Transfer Events

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2107) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2452)

`.Post` was impossible to implement for mods.
https://github.com/SpongePowered/SpongeForge/issues/2411

Moves `ChangeInventoryEvent.Transfer` to `TransferInventoryEvent`
`TransferInventoryEvent.Post` cannot be cancelled anymore. It now includes the Source/Target Slot and the ItemStack transferred.


XakepSDK:
Maybe `Pre` should have source slot?

Faithcaio:
There is no Slot in Pre because its before any calculations happen.

Issue #1814, opened 2018-04-22
Replace integer ticks durations with Duration across the API.


Cybermaxke:
Originally @Meronat's idea

> What do you think about introducing methods which accept a Duration instead of the long/double/int representing ticks in the API. Works better for other implementations than vanilla(e.g. lantern), easier to understand for plugin developers, and can just deprecate the old methods.
I've noticed issues where plugin developers misinterpret the parameter and it will be better for our implementation anyway.

All the usages of ticks would be replaced by Duration, or other time API related objects (which fit the usecase). Ticks is widely used across the API, this change will cause a lot of breaking changes.

Thoughts?

parlough:
For where we accept integers, we can deprecate those more easily, but return types are a bit more challenging and will cause breaking changes.

Switching to `Duration` provides less confusion for developers as they often misinterpret the parameter or return type, even with proper javadocs. It is also a better choice as not all implementations will necessarily limit the game to 20 tps.

dualspiral:
I think we need to do this, as an addition, as I got caught out by it. [I mentioned at the time that a Duration version may be a good idea](https://github.com/NucleusPowered/Nucleus/issues/1122#issuecomment-379586153), though it slipped my mind as I was at work at the time so forgot to bring it up later here. I'm glad someone else remembered to do so.

It is important to note throught documentation that **supplying a `Duration` doesn't guarantee that the effect or whatever _will be_ that duration**, which might be due to the tick system or some other effect. We need to be clear in the Javadocs that the duration supplied will be converted to the internal representation and may differ slightly due to server performance. Whether that is by a lower ticks per second or that the duration specifies a period does not divide evenly into ticks (though in the vast majority of cases, it will) doesn't matter really.

In defence of the current function - having a tick based function indicates that you want something to run for x ticks. Ultimately, specifying 20 ticks is 20 ticks, which is an undefined temporal duration but is guaranteed to be 20 ticks in the absence of any overriding calls. How long a tick is is at the mercy of the impl, though [Sponge indicates that the server aims for 20 TPS, really, only the MCP implementations aim for that](https://jd.spongepowered.org/7.0.0/org/spongepowered/api/Server.html#getTicksPerSecond--). 

[...]

PR #1920, opened 2018-10-09
Replace integer ticks durations with Duration across the API.

Cybermaxke:
See https://github.com/SpongePowered/SpongeAPI/issues/1814

Right now I only added a catalog of `TemporalUnit`s, including minecraft related ones. But we have to keep the conversation going about this. 

Will very likely be held off after the 1.13 update to avoid conflics with changes. The catalog could be merged early, maybe even for API 7.x

### DIFF
--- a/src/main/java/org/spongepowered/api/util/TemporalUnits.java
+++ b/src/main/java/org/spongepowered/api/util/TemporalUnits.java
@@ -1,0 +1,146 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/api/util/TemporalUnits.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * @see ChronoUnit#HOURS
+     */
+    public static final ChronoUnit HOURS = ChronoUnit.HOURS;
+
+    /**
+     * @see ChronoUnit#HALF_DAYS
+     */
+    public static final ChronoUnit HALF_DAYS = ChronoUnit.HALF_DAYS;
```
ST-DDT, line 66: For simplicity reasons I would keep these out of here.

src/main/java/org/spongepowered/api/util/TemporalUnits.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * A {@link MinecraftTemporalUnit} which represents a half minecraft day.
+     */
+    public static final MinecraftTemporalUnit HALF_MINECRAFT_DAYS =
+            new MinecraftTemporalUnit("HalfMinecraftDays", MINECRAFT_TICKS.getDuration().multipliedBy(12000));
+
+    /**
+     * A {@link MinecraftTemporalUnit} which represents a minecraft day.
+     */
+    public static final MinecraftTemporalUnit MINECRAFT_WEEKS =
```
ST-DDT: Is there a concept of weeks in minecraft?

Cybermaxke: Not that I know of. A moon cycle is 8 minecraft days, would that be worth adding?

ST-DDT: That sounds like an implementation detail and I'm not sure whether someone uses that large duration units with a reference to minecraft moon cycles. (The vanilla time is not stable enough enough for a moonthly message in the chat)

Cybermaxke:
~wip

Cybermaxke:
Manually merged in 1.13 for further development.

PR #1921, opened 2018-10-21
Plugin client login phase.

Cybermaxke:
Support for the new login plugin message request/response in 1.13.

**This PR is based on top of my 1.13 changes, so only check the last diffs.**

How we could handle it is as following (inspired by the forge solution):
During a login event (`ClientConnectionEvent.Login`?), allow plugins to send login message to the client. The plugins get a `CompletableFuture` in which they can handle the responses and send new login messages (if needed).
The login event will not be processed until there are no more login messages send which are waiting for a response.

Thoughts?
Should I move this to a separate event instead of `ClientConnectionEvent.Login`? Suggestions for a name?


Cybermaxke:
~wip

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
```diff
@@ -35,6 +36,85 @@
...
+     * Creates a new login channel binding for the given channel name. The channel can
+     * be used to send and receive messages.
+     *
+     * @param plugin The plugin registering the channel
+     * @param channel The channel to register
+     * @return A new {@link LoginChannelBinding} instance bound to the channel name
+     * @throws ChannelRegistrationException The channel name is too long
+     * @throws ChannelRegistrationException The channel name is reserved
+     */
+    LoginChannelBinding.IndexedMessageChannel createLoginChannel(Object plugin, String channel) throws ChannelRegistrationException;
```
simon816, line 49: This seems like a lot of duplication with the existing API. Perhaps channel bindings could be per phase (login/play)?

Cybermaxke: Maybe I can expose the Request/Response system to normal `ChannelBinding`s, that would reduce the duplication. There is however one problem, the content of the Raw Login Channel can be null if the message wasn't handled by the client, and we can't use a empty buffer because that is also a valid payload.

Zidane:
Remove all the extra stuff here. Why is that in this PR?

Cybermaxke:
This PR is outdated, I moved my WIP changes to a Sponge branch.

PR #1939, opened 2018-12-10
Use null check from internal java library

lsiem:
The google "checkNotNull" function can be safety replaced with the java "requireNonNull" function.


kashike:
Yes, we are well aware.

At this time we are not going to accept this pull request. If you want to do this, you can submit a pull request across all Sponge repositories changing this in every file.

liach:
It can be easily done with the migration ability of intellij idea.

PR #1942, opened 2019-01-04
Add method to get the temperature at a specific position

JBYoshi:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2170)

Biomes can have different temperatures at different positions - for example, snow only occurs when the ground is higher up. This PR adds a method to get the temperature.


Cybermaxke:
The temperature can already be retrieved through properties. A separate method shouldn't be necessary.

PR #1944, opened 2019-01-07
Fix NegativeArraySizeException in BlockState.MatcherBuilder.build()

### LINKED ISSUE #1835: BlockState.MatcherBuilder#build() throws a NegativeArraySizeException when no traits are specified
This can be fixed by removing the ` - 1` part in `new BlockTrait<?>[this.traits.size() - 1]`.

(Creating an issue instead of a PR because it is a trivial change)

14mRh4X0r:
When no traits were specified, MatcherBuilder.build() would throw a
NegativeArraySizeException because of flawed logic (size always one too
small).

A size of 0 is specified here, because it's faster than getting the size
and passing it.

Fixes #1835 .

PS: Created this PR because asking to fix it in IRC didn't seem to help.


14mRh4X0r:
Force-push is because the commit message referenced the wrong issue.

phit:
~qa

PR #1950, opened 2019-01-19
1.14 - Village and Pillage

Cybermaxke:
Updates the API from 1.13 to 1.14


liach:
Any entity type?

Cybermaxke:
@liach Not yet, this is just auto generated. I haven't looked into those

liach:
Ah. I hope we can interface out village and raid data stored in world save.

gabizou:
Just, uh... Closing since this can be alive in `1.14` branch which will get merges between 1.13 etc.

PR #1951, opened 2019-01-20
Update channel buf for netty buffer changes.

Cybermaxke:
This updates the channel buf to reflect the latest netty changes, ignoring the netty version that minecraft is using (this can be easily avoided for these changes). I haven't added all the little endian methods yet because I would first like a bit of input.

Changes:
- Deprecate `ChannelBuf#order` in favor of little endian methods. 
- Deprecate marked reader/writer indexes, you can easily get them and reset them afterwards https://github.com/netty/netty/issues/8535


Cybermaxke:
https://github.com/SpongePowered/SpongeAPI/commit/6d3fd94efd8b9d3002eb989a75404fd6c8dea16b

PR #1952, opened 2019-01-20
Consistent catalog builders + move recipe registration

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/2174) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/2565)

Updates PR https://github.com/SpongePowered/SpongeAPI/pull/1870 for bleeding.


src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -82,6 +83,19 @@ static CatalogKey sponge(final String value) {
...
     }
 
+    /**
+     * Creates a catalog key with the specified value and
+     * the namespace of the active plugin in the {@link CauseStackManager}.
+     *
+     * @param value The value
+     * @return A new catalog key
+     */
+    static CatalogKey forActivePlugin(final String value) {
```
Cybermaxke: Is this name good?

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Sets the {@link CatalogKey} of the {@link CatalogType}.
+     *
+     * <p>Formatting for the namespace and value are lowercased as
+     * usual.</p>
+     *
+     * @param key The key
+     * @return This builder for chaining
+     */
+    B key(CatalogKey key);
```
Cybermaxke, line 47: Since the last PR, I added this method so you can pass in any plugin id and id. And `id` still uses the active plugin as namespace, which I added a method for in `CatalogKey`.

src/main/java/org/spongepowered/api/CatalogKey.java
XakepSDK: `forPlugin(Object plugin)`?

Cybermaxke: That signature doesn't reflect how this method works:

Key:
  namespace: id of plugin in cause stack
  value: value

I know that the name can be confusing, that's why I would like different options.

XakepSDK: Ah so
`namespaced(String)`
`of(String)`
`ofSelf(String)`
`forSelf(String)`

gabizou: Need to note that this **cannot** be called asynchronously

src/main/java/org/spongepowered/api/util/CatalogBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+    B name(Translation translation);
+
+    /**
+     * Builds the {@link CatalogType} of type {@link C}.
+     *
+     * <p>The last {@link PluginContainer} in the cause stack will be used to
+     * determine which plugin was used to construct the {@link CatalogType}.</p>
+     *
+     * @return The built catalog type
+     * @throws IllegalStateException If not all required options were specified
```
gabizou: Throws another exception if this is being called asynchronously as the CauseStackManager cannot be accessed off main thread(s) (when client is introduced, client thread will have a cause stack manager as well as the server thread). 

May want to just add documentation for the overriding methods in various builders that outline the "requirements" for this build to succeed and not fail.

gabizou (approved):
Seeing as we're royally breaking API and avoiding our deprecation policy for API 8, this change is fine. I wouldn't be opposed to having the legacy interface heavily deprecated and legacy methods in some places deprecated for easier transition, but the changes themselves are good.

gabizou (approved):

Issue #1303, opened 2016-07-07
Enchanting Events


ryantheleach:
There is a lack of Enchanting Events available in SpongeAPI

Needed:
EnchantItemEvent: an event that fires when an item's enchantments are being changed either through an anvil or enchanting table.

PreEnchantEvent an event that fires when the choices are given to a player when they insert an item into the enchanting table or anvil, giving them feedback on what happens next.

Usecase:
1. On my old Bukkit server I would add lore onto items that had been enchanted, with the players names. This made stealing items more interesting, because you could potentially see who the last person who upgraded or cared for an item was. I don't think this is currently possible without enchanting events.
2. Limiting higher level enchants to a magical crafters RPG class.

nihiluis:
agree

nihiluis:
bump

PR #1954, opened 2019-01-26
EnchantItemEvent

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2183)

3 new Events:

- `EnchantItemEvent.CalculateLevelRequirement`  first event ; calculates the level requirement for each enchant
- `EnchantItemEvent.CalculateEnchantment` second event ; calculates the list of enchantments to apply - fires twice - once for the preview then again when actually enchanting - plugins should return the same list for the same input
- `EnchantItemEvent.Post` the final event ; contains the slot transactions of the enchanted item getting changed and the lapis used.

For reference:
https://github.com/SpongePowered/SpongeAPI/issues/1303

### DIFF
--- a/src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
+++ b/src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
@@ -1,0 +1,157 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.item.inventory;
+
+import org.spongepowered.api.item.enchantment.Enchantment;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.item.inventory.Slot;
+
+import java.util.List;
+
+public interface EnchantItemEvent extends TargetContainerEvent {
+
+    /**
+     * The seed for pseudo random enchantment generation.
+     *
+     * @return the seed
+     */
+    int getSeed();
+
+    /**
+     * The enchantment option. 1, 2 or 3
+     *
+     * @return The enchantment option
+     */
+    int getOption();
+
+    /**
+     * Triggers when the enchantment costs for an item are calculated.
+     */
+    interface CalculateLevelRequirement extends EnchantItemEvent {
+
+        /**
+         * The enchantment power based on bookshelves around the enchantment table.
+         *
+         * @return the enchantment power
+         */
+        int getPower();
+
+        /**
+         * The itemstack to enchant.
+         *
+         * @return the itemstack to enchant.
+         */
+        ItemStackSnapshot getItem();
+
+        /**
+         * Returns the original level requirement
+         *
+         * @return the original level requirement
+         */
+        int getOriginalLevelRequirement();
+
+        /**
+         * Returns the level requirement
+         *
+         * @return the level requirement
+         */
+        int getLevelRequirement();
+
+        /**
+         * Sets the new level requirement
+         * <p>In vanilla the maximum value is 30.</p>
+         * <p>Returning a much higher level may result in no enchantments
+         * because the existing enchantments are too weak.</p>
+         *
+         * @param value the new level requrement
+         */
+        void setLevelRequirement(int value);
+    }
+
+    /**
+     * Triggers when the enchantments for an item are calculated.
+     * This happens after a new item to enchant is put in the table
+     * and again when the item is enchanted.
+     */
+    interface CalculateEnchantment extends EnchantItemEvent {
+
+        /**
+         * The itemstack to enchant.
+         *
+         * @return the itemstack to enchant.
+         */
+        ItemStackSnapshot getItem();
+
+        /**
+         * The final level requirement from {@link CalculateLevelRequirement}.
+         * <p>In vanilla the maximum value is 30.</p>
+         *
+         * @return the level requirement for the option
+         */
+        int getLevelRequirement();
+
+        /**
+         * Returns the original list of enchantments to apply to the item.
+         * <p>The first item in the list is used for display when previewing the enchantments.</p>
+         *
+         * @return the list of enchantments
+         */
+        List<Enchantment> getOriginalEnchantments();
+
+        /**
+         * Returns the list of enchantments to apply to the item.
+         * <p>The first item in the list is used for display when previewing the enchantments.</p>
+         *
+         * @return the list of enchantments
+         */
+        List<Enchantment> getEnchantments();
+
+        /**
+         * Sets the list of enchantments to apply to the item.
+         *
+         * <p>The first item in the list is used for display when previewing the enchantments.</p>
+         * <p>Note that when modifying the enchantment list you should
+         * return the same enchantments for the same seed and option.</p>
+         * <p>See {@link Enchantment#randomListBuilder()} to generate enchantment lists.</p>
+         *
+         * @param enchantments the new list of enchantments
+         */
+        void setEnchantments(List<Enchantment> enchantments);
+    }
+
+    /**
+     * This event is triggered when an item is enchanted at an enchanting table.
+     */
+    interface Post extends ClickInventoryEvent, EnchantItemEvent {
+        /**
+         * Returns the slot of the enchanted item.
+         *
+         * @return the slot if the enchanted item.
+         */
+        Slot getEnchantingSlot();
+
+    }
+
+}
--- a/src/main/java/org/spongepowered/api/item/enchantment/Enchantment.java
+++ b/src/main/java/org/spongepowered/api/item/enchantment/Enchantment.java
@@ -28,6 +28,9 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.DataSerializable;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.List;
 
 /**
  * Represents an {@link EnchantmentType} on an {@link ItemStack} that is paired
@@ -47,6 +50,10 @@ public interface Enchantment extends DataSerializable {
      */
     static Builder builder() {
         return Sponge.getRegistry().createBuilder(Enchantment.Builder.class);
+    }
+
+    static RandomListBuilder randomListBuilder() {
+        return Sponge.getRegistry().createBuilder(RandomListBuilder.class);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+        Transaction<EnchantmentList> getEnchantments();
+
+        interface EnchantmentList extends DataSerializable {
+
+            // Utility functions to generate different enchantment lists
+
+            List<Enchantment> enchantments();
+            // fixed enchantments defined by plugin
+            EnchantmentList withEnchantments(Enchantment... enchantments);
+            EnchantmentList withEnchantments(List<Enchantment> enchantments);
```
ImMorpheus: Why a List ?

Faithcaio: You add multiple enchantments in one go?

ImMorpheus: uh, I commented on the wrong line.

Anyway, why not a Collection ?

and about `List<Enchantment> enchantments();`, Can you have duplicate enchantments ?

Faithcaio: The first enchantment of the list is taken for the preview.
Duplicate enchantements probably makes no sense.
(and its a list in the impl)

Cybermaxke: Maybe put in the javadoc that the first one is taken for the preview?

ImMorpheus: IMO, that's an implementation detail.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+        int getSeed();
+        // 1,2,3
+        int getSlot();
+        // vanilla max 30
+        int getCost();
+
+        // First Enchantment in list is used for clue in client view
+        Transaction<EnchantmentList> getEnchantments();
+
+        interface EnchantmentList extends DataSerializable {
```
gabizou: I'd rather this get moved into it's own class/package where Enchantments resides.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
gabizou: I feel a `List` is still better since it imperatively retains order, whereas a `Collection` is a hit and miss of ordering. The aspect of "first enchantment is used for preview" is kinda meh, but it would make sense if we keep a `List` and just state something about it being used as a preview because plugin developers otherwise may not know that fact beforehand.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+            // Utility functions to generate different enchantment lists
+
+            List<Enchantment> enchantments();
+            // fixed enchantments defined by plugin
+            EnchantmentList withEnchantments(Enchantment... enchantments);
+            EnchantmentList withEnchantments(List<Enchantment> enchantments);
+
+            // redo random vanilla calculation - with different input values
+            // EnchantmentHelper.buildEnchantmentList
+            EnchantmentList with(int seed, int slot, int level, boolean treasure);
```
gabizou: Do we need to use slot indexes?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
gabizou: And what exactly is an enchantment list supposed to represent? Is it something closer to Data API's `Keys.ENCHANTMENTS`?

gabizou (changes requested):

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: Its not an inventory `Slot` but the enchantment button.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: It will get moved out. Just in the same file for visibility atm.

Using the `EnchantmentData` would probably be fine.
I wanted something where I could add utility methods to:
 - reroll random enchantments with different parameters
 - reroll random enchantments with fixed enchantment pool
 - set enchantment result 

move those to smth like `EnchantmentUtil` ?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
liach: We can just use an `Iterable` and document our stand on the iteration order.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+     * and again when the item is enchanted.
+     */
+    // Mixin ContaienrEnchantment.getEnchantmentList and/or
+    //       EnchantmentHelper.buildEnchantmentList-at-getEnchantmentDatas
+    interface CalculateEnchantment extends TargetContainerEvent {
+
+        // Seed + slot are seed for enchantment-list generation
+        int getSeed();
+        // 1,2,3
+        int getSlot();
```
liach: Mind interfacing these 2 methods out for `CalculateCost` and `CalculateEnchantment` to share?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Triggers when the enchantment costs for an item are calculated.
+     */
+    // Mixin onCraftMatrixChanged in loop after EnchantmentHelper.calcItemStackEnchantability
+    // Forge: ForgeEventFactory.onEnchantmentLevelSet
+    interface CalculateCost extends TargetContainerEvent {
+
+        int getSeed(); // pseudo random seed
+        int getSlot(); // 1,2,3
```
liach: The word "slot" is misleading. `Option` or `Line` is definitely better.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return the slot if the enchanted item.
+     */
+    Slot getSlot();
+
+    /**
+     * Triggers when the enchantment costs for an item are calculated.
+     */
+    // Mixin onCraftMatrixChanged in loop after EnchantmentHelper.calcItemStackEnchantability
+    // Forge: ForgeEventFactory.onEnchantmentLevelSet
+    interface CalculateCost extends TargetContainerEvent {
```
liach: Is this really a "cost"? This appears like a level requirement to me instead.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+            // fixed enchantments defined by plugin
+            EnchantmentList withEnchantments(Enchantment... enchantments);
+            EnchantmentList withEnchantments(List<Enchantment> enchantments);
+
+            // redo random vanilla calculation - with different input values
+            // EnchantmentHelper.buildEnchantmentList
+            EnchantmentList with(int seed, int slot, int level, boolean treasure);
+            // redo random vanilla calculation - with treasure enchantments
+            // EnchantmentHelper.buildEnchantmentList
+            EnchantmentList withTreasure(boolean treasure);
```
liach: Can curse be optional as well?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,114 @@
...
+
+            // redo modified random vanilla calculation - with custom enchantment list
+            // impl note: fixed getEnchantmentDatas then same as vanilla
+            EnchantmentList with(int seed, int slot, int level, List<Enchantment> list);
```
liach: why no treasure arg here?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
liach: We should not need the "slot" at all. We have the slot when the event is fired.

liach (changes requested):
Beyond all these issues, you need to remove the implnotes from the api.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Cybermaxke: My guess is that the enchantment list will be generated for the given slot, it doesn't have to be the one that you clicked on.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: Look like curses are treasure enchantments.
Not allowing curses would require us to write our own version for getting random enchantments.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: This uses the fixed list of enchantments as the pool. If you don't want treasure enchantments dont add them to the list.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: It is.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: Yes. In the impl. the slot/option is just added to the random seed.

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
kashike: Treasure and Curse are separate - there are only two curses (`binding_curse`, `vanishing_curse`), while there are 4 treasures (`binding_curse`, `frost_walker`, `mending`, `vanishing_curse`).

Faithcaio: `EnchantmentHelper#buildEnchantmentList(Random randomIn, ItemStack itemStackIn, int level, boolean allowTreasure)` doesn't allow filtering curses. 
We can still add it to the API if its doable.

liach (approved):

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,126 @@
...
+ * This event is triggered when an item is enchanted at an enchanting table.
+ */
+public interface EnchantItemEvent extends InteractContainerEvent {
+
+    /**
+     * Returns the slot of the enchanting item.
+     *
+     * @return the slot of the enchanting item.
+     */
+    Slot getEnchantingSlot();
```
Zidane: getOutputSlot instead? How do we want to word these? Other Tiles that have logic could have similar events (like the Anvil one we already do)

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
```diff
@@ -0,0 +1,126 @@
...
+         * @return the list of enchantments
+         */
+        Transaction<EnchantmentList> getEnchantments();
+
+    }
+
+    /**
+     * Triggers when the item was enchanted.
+     */
+    interface EnchantItem extends EnchantItemEvent, ChangeInventoryEvent {
```
Zidane: Maybe call this Post? The docs imply the action occurred, the name of the event implies it is about to occur.

src/main/java/org/spongepowered/api/event/item/inventory/enchantment/EnchantmentList.java
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param enchantments the enchantments
+     *
+     * @return the new enchantment list
+     */
+    EnchantmentList withEnchantments(List<Enchantment> enchantments);
+
+    // Utility functions to generate enchantment lists
+
+    /**
```
Zidane: The below two methods feel like they belong elsewhere. What if I want to build an EnchantmentList but in my code I don't have an EnchantmentList already on hand? How would I do this?

src/main/java/org/spongepowered/api/event/item/inventory/EnchantItemEvent.java
Faithcaio: Its the also the input slot.

src/main/java/org/spongepowered/api/event/item/inventory/enchantment/EnchantmentList.java
Faithcaio: Atm. you cant do it. 
I guess if you wanted to emulate an enchanting table somewhere else you might need this.
We would have to expose it in some kind of builder class then?

parlough (approved):

PR #1958, opened 2019-02-02
New high-level command system for 1.13

dualspiral:
Original PRs - [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1587) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1400) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1667)

---

Replaces the previous PR. No impl is available yet.

This is for 1.13, and the intention is that it will replace our current system. I'll write up more once we have a firmer idea on how this will follow implementation, **this API will change yet to accommodate this**. As such, this is still WIP, and I ask that any comments are on substance and not nitpicks.

This will not contain any annotation based command systems, these will occur after this is done and dusted. 

For original intentions and goals, see #1587


src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -260,18 +206,27 @@ public Builder affectedItems(@Nullable Integer affectedItems) {
...
+         * Sets or unsets the error message to return to the user without
+         * throwing an exception.
+         *
+         * <p>If this is set, then the command parser will send this message to
+         * the command invoker.</p>
+         *
+         * @param errorMessage The message to send to the user.
+         * @return This builder, for chaining
+         */
+        Builder error(@Nullable Text errorMessage);
```
ST-DDT, line 124: Does the caller have to apply the error formatting or is the command manager supposed to do that.

src/main/java/org/spongepowered/api/command/IncorrectCommandSourceException.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class IncorrectCommandSourceException extends CommandException {
+    private static final long serialVersionUID = -2927330349931825821L;
+
+    /**
+     * Create an exception with the default message.
+     */
+    public IncorrectCommandSourceException() {
+        this(t("You cannot execute this command!"));
+    }
```
ST-DDT: I'm torn by this method. On the one hand I would like to send a localized message, on the other hand I would also like to specify who is allowed to execute it.

Cybermaxke: `SpongeApiTranslationHelper` shouldn't be used anymore, translations go in the impl.

ST-DDT: Yes, which is why I thought about removing this hard coded text from the API, and replacing it with just a list of types that are allowed to execute the command. The impl is then responsible for checking whether a message is set or constructing it (in a localized way) using the list of allowed types.

src/main/java/org/spongepowered/api/command/parameter/managed/standard/CatalogedValueParameters.java
```diff
@@ -0,0 +1,392 @@
...
+    public static final CatalogedValueParameter<Long> LONG =
+            DummyObjectProvider.createExtendedFor(CatalogedValueParameter.class, "LONG");
+
+    /**
+     * Does not parse any arguments, returning nothing.
+     *
+     * <p>Returns nothing - no entry will be placed into any provided key.</p>
+     */
+    public static final CatalogedValueParameter<Void> NONE =
+            DummyObjectProvider.createExtendedFor(CatalogedValueParameter.class, "NONE");
```
ST-DDT, line 247: What would be the use case for this? Is it a requirement for commands to parse all parameters and throw an exception if there are any trailing ones and if they don't want to throw parse the remaining with `NONE`?

src/main/java/org/spongepowered/api/command/CommandResult.java
dualspiral: The command manager will apply a default formatting, at least as it stands - in the default impl it'll likely be light red. If a plugin applies a colour, that'll be the colour.

src/main/java/org/spongepowered/api/command/IncorrectCommandSourceException.java
dualspiral: Do bear in mind that this is a port from the previous PR and things have been missed. This will go in the impl.

I expect this message will be in two bits - one basic message and a list of allowed sources. So, something like (but not exactly):

```
You cannot execute this command using this source!
Allowable sources: Player, Console, Command Block
```

src/main/java/org/spongepowered/api/command/parameter/managed/standard/CatalogedValueParameters.java
dualspiral: Its use _was_ for flags, where the `buildWith` method required a `CommandElement`. It allowed you to have flags only - see `GenericArguments.none()`.

Not sure if there is a use right now, it's simply ported over at this moment in time.

ST-DDT:
This looks very promising. Please add an example command definition/implementation once your main refactoring is done. This will be helpful later, when we need to update the docs.

dualspiral:
Yup, once we're closer if I haven't opened a Docs branch, please prod me.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -0,0 +1,535 @@
...
+ *
+ * <p>Commands in Sponge are provided with a {@link Cause}, as well as an
+ * associated {@link CommandSource}. Note that the {@link CommandSource} is
+ * not necessarily the object that actually invoked the command, however, it
+ * is the source that is <strong>responsible</strong> for the command. Therefore,
+ * in general, command eligibility checks should be performed against the
+ * provided {@link CommandSource}. However, users and implementors may act upon
+ * the {@link Cause} alone as they see fit, should it be preferred.</p>
+ */
+public interface Command {
```
kashike, line 65: I disagree with this name, as I said before - this name should be reserved for annotation-based commands, when they come.

src/main/java/org/spongepowered/api/command/CommandResult.java
```diff
@@ -87,7 +83,7 @@ public static CommandResult successCount(int count) {
      * @param count The amount of blocks affected
      * @return The result
      */
-    public static CommandResult affectedBlocks(int count) {
+    static CommandResult withAffectedBlocks(int count) {
```
kashike: Vanilla removed all this - commands just return an `int` in vanilla, which is "result".

kashike (changes requested):

src/main/java/org/spongepowered/api/command/Command.java
dualspiral: **Absolutely not.** I don't see a compelling reason as to why this should be renamed. Why should this, which is a command and nothing more, be displaced by an annotation which builds on top of this? It's completely backwards.

A `@Command` annotation can still exist in a different package, done right, this interface won't need to be implemented in the same class as the annotation being used.

kashike: Sure, but how about cases where you'd have both types in a single class? Fully-qualified imports suck.

Feel free to suggest an alternative name for the annotation.

dualspiral: Sure, they do. I'll cross that bridge when we come to it.

ST-DDT: A good name for the annotation might be `@CommandDefinition`

Katrix:
Any plans for how custom command implementations can provide 1.13 style suggestions?

dualspiral:
As it stands here, my plan would be to provide a way to supply a `Parameter` chain to the system, with custom implementations implementing  and providing a set of `Parameter<>`s. However, I know this is not ideal as it is heavily skewed to the system we have here.

Once things solidify a bit more, it would be good to come and have a look at what you need and perhaps see if we can jointly build a solution.

Note that I am looking at how this all fits together with Brig (and the implementation within Minecraft) in the first place. While my plan has been to consider _not_ using Brig and instead just sending the required packets because there were things that we did that Brig didn't/doesn't that I'd like to keep, discussions away from here made me take a fresh look at this. Our biggest pain point will be flags, but with my musings, we might be able to have some "poor mans" solution to this. So I think I might have to rework parameters a little anyway.

dualspiral:
I've been spending some time playing with Brigadier (Mojang's system in 1.13) and there are a few features that are missing compared to our current library that I think we would need to address.

These are my thoughts and explain my thinking on the way forward to support the needs of the Sponge community while being as unobtrusive as possible in the Minecraft code. _Constructive_ comments are welcome.

# What I see as shortcomings (things we really want in Sponge but aren't in Brig)

## Command Flags

Brigadier does not support command flags at all. However, we could sidestep this by creating elements that mimic the flags (for example, a literal "-f" or "-s <int>" could become a literal "-s" followed by "<int>"), but positioning of these elements would be limited.

## Optional Elements (ish)

It turns out that this can be very cumbersome to do. This would effectively end up with duplicated nodes across the tree - while this is a necessary evil no doubt, trees _may_ become messy. Unfortunately, redirects won't help here so much. We could possibly wrap/extend `ArgumentType`s to do it for us in a way, or create our own `OptionalArgumentBuilder`.

## Overloading Parameters

Say you have a command `/poke player` that you want to accept a UUID or name. In Sponge, you might do the following:

`GenericArguments.firstParsing(GenericArguments.player("player"), new UUIDPlayerParser("player"))`

In this example, both the `Player` and `UUID` parameters use the key `player`. You can then write your command to check for a player in the "player" key of the context. In Brig, I've not found a simple way to do this. We _might_ be able to do something with redirect nodes and their modifiers, but it's far from simple as to what to do here. We could also create our own `RequiredArgumentBuilder` on top of this to support what we want to do.

## `ArgumentType` parsers are not context aware

`ArgumentTypes`, which perform the argument parsing, only have information on the string that has been passed to them (via a `StringReader`, which is similar to our `CommandArgs`). A lot of commands, Nucleus' `/home` being one, parse that string based on the invoker of the command, Brig cannot yet do this.

We could use our Cause system to determine who is invoking the command and call into argument types that way, but this isn't ideal. I'd much rather not rely on a state object - and it might cause problems when working with threads in the future, if commands go async (though an async cause tracker is also potentially on the cards at this point).

# What's good about Brig (in Minecraft) - and how we can use this

## Loads of `ArgumentTypes` we can hook into

We can mixin to the `ArgumentType` classes and catalog them, as well as add our own methods to modify the class to fit with how we envisage the system

## (Almost) everything is extendible

Both `StringReader` and `CommandContext` could be extended (or mixed into) by Sponge equivalents that include what we're used to, such as `CommandArgs`.

## Client Command Completion

The obvious reason for considering Brigadier above all else is this - client command completion. We can harness this and send down "examples" that tab complete can do right there on the client, no network packets.

Care would need to be made to ensure the tab complete is always valid.

## Subcommands can be created at any point in the parameter string

Something that a few people have asked for is in Brigadier. This is possible because of the way Brig determines what command to execute. In Sponge, we have a separation between command and argument such that you define a set of arguments for a command. In Brig, you define a command on an argument. This makes commands part of the tree, and is part of why optional arguments can be difficult to integrate. However, we should seek to support this too, which will most likely take the direction of having a special `subcommand()` argument that you attach a `Command` too.

# What tentative approach we should take

I don't believe that we should use Brigadier as our replacement API, but there is nothing wrong in using **parts** of it to enable convergence at a later time. Some thoughts on what will happen here:

* `ValueParameterModifier`s will go away in favour of specific `Parameters`, they are only generally used here to support defaults. Other things that it did can be built into the parameter generator.
* `ValueParameter`s are effectively `ArgumentType`s - the gap between then should shrink
* `ValueCompleter`s are effectively [`SuggestionProviders`](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/suggestion/SuggestionProvider.java)
* We can't seem to alter the examples in the same way
* For impl, consider extending Brig in our own ways to support what we want it to do.
* **PR things back to Mojang that we find useful, in the hope that others can use it too**.
* Consider how we manage commands. Brig has an interesting take on "merging command trees". I'm not convinced we want to do that with Sponge commands, and will likely want to namespace things as we do now.

gabizou:
> * **PR things back to Mojang that we find useful, in the hope that others can use it too**.

Yes, 1000x yes. 

> * `ValueParameterModifier`s will go away in favour of specific `Parameters`, they are only generally used here to support defaults. Other things that it did can be built into the parameter generator.

Simplification is always good.

> * For impl, consider extending Brig in our own ways to support what we want it to do.

Always an option. One thing to consider will be our maintenance overhead as new versions add features, assuming Brig is semver, and version bumps only apply to new versions of Minecraft. I'm not against mixing into Brig to achieve what we want, just a consideration for future maintenance as to *how much* we will be modifying. From your comments though, it doesn't seem like we'll have that much of a burden.



> * Consider how we manage commands. Brig has an interesting take on "merging command trees". I'm not convinced we want to do that with Sponge commands, and will likely want to namespace things as we do now.

If by merging command trees you mean that Brig will merge different commands of the same root (like `/tp`) to have the tree being extended based on multiple options of who accepts what argument types in order being entered, I'm partial to keeping namespacing in general as conflicts and confusion is best avoided. If two commands register the same arguments with the same root name, is that allowed in Brig? How would tree merging be affected by aliases?

dualspiral:
> If by merging command trees you mean that Brig will merge different commands of the same root (like /tp) to have the tree being extended based on multiple options of who accepts what argument types in order being entered

That's exactly it.

> If two commands register the same arguments with the same root name, is that allowed in Brig? 

As far as I can see, no. It's actually a problem for us, as you might want two arguments at the same level with the same key. I'm playing about with ways to resolve this in a (currently) personal project, https://github.com/dualspiral/General, such that we can wrap multiple parsers on the same node, which we can then put an API on top of.

(For anyone that goes looking, stuff in there is not final, please don't go nitpicking at the style because it's a personal testing repo)

> How would tree merging be affected by aliases?

Brigadier has the concept of a "redirect node", so `/tp` and `/teleport` can have exactly the same node chain by putting a redirect at the root of one of the command chains - the redirect can be thought of as a symbolic link. If you set up aliases using redirect nodes, a change to a node will affect them all.

It also has the concept of "forking" commands, but I don't plan to expose that.

>  I'm partial to keeping namespacing in general as conflicts and confusion is best avoided

I agree. My tentative approach is to replace the dispatcher with an extended dispatcher based on Brig to do this, but also try to be somewhat transparent to mods that are using Brig already - we'll see how that goes. Extending the dispatcher looks like it's going to be easy.

src/main/java/org/spongepowered/api/command/Command.java
aikar: the term I use as well as Mojang is RootCommand for the /root portion, then subcommands for nodes underneath that.

In ACF I named it @CommandAlias but @RootCommand would of been nice, but I already had that used as an interface.

pie-flavor:
My 2&cent;: 

> Our biggest pain point will be flags, but with my musings, we might be able to have some "poor mans" solution to this. 

Please do not. Flags as they currently are are *already* a hassle. The parent PR was originally 'redo commands, and do them right this time'. Patch-job solutions defeat the purpose.

aikar:
As the creator and maintainer of the most used annotation based framework for bukkit which also supports sponge, can I suggest please leaving annotation based management systems out of sponge-api, and let sponge-api simply provide 'capability' tools.

https://acf.emc.gs

ACF is years in the making with tons of features to come still. I haven't merged it into Paper because it would be a DRASTIC limitation on the ability to improve the framework. It locks the API in that it can't really be changed without the then risk of breaking plugins, where as providing a framework as a library built on top of the platforms foundation allows plugins to independently update and maintain their own version preference of the framework.

I will work with @bloodmc @kashike and others to ensure we can get ACF updated to this new API as soon as its landed in -stable, and would love if sponge encouraged people to use ACF for annotation based commands instead.

And the plan is to integrate ACF with brigadier, so that would be my desire w/ this command API Expansion is providing the capability for acf-sponge to register brigadier metadata through official API.

liach:
I definitely agree with @aikar. This not only increases maintenance cost for API but also brings trouble to implementations.

liach:
Can we just dump all causes in params first?

dualspiral:
> Can we just dump all causes in params first?

Can you be clearer on that? I'm reading that as "remove causes", but it could mean anything.

liach:
right now many methods require a cause object (the event one) for operations, which is no longer the case in the apo

dualspiral:
No.

Think about how a command is going to be used - effectively calling what is currently `CommandManager#process(String entry)`. That takes the cause from the cause stack manager - a cause should not be supplied there. The cause of the command that then gets parsed is fixed **at that point**, before any parsing occurs. That is what the `Cause` object is - and that's what we pass about. The cause of the parsing is the cause of the original invocation. Any commands that get executed at that point should be getting the **exact same cause** - _that_ is why they have `Cause` objects as parameters. If we just let everything get the current cause, the entire command could be executed under a myriad of multiple causes and thus be inconsistent with itself (as well as potential performance hits because we're continuously re-creating the `Cause`).

To put it another way - entry points from the game engine should not have a `Cause` parameter. That is the `CommandManager`. `Command`s are invoked by this manager which has already finalised the `Cause` of the command invocation and are not called directly by the game engine - thus an immutable `Cause` is passed through to them.

Finally, `Command`s are potentially user implementable, as are `ValueParameter`s. By providing such signatures, we're saying "we suggest you act upon the `Cause` that we supply", making it clear that we build our command system around causes. It also makes the commands more testable - testers can provide any cause they like without being constrained by the CSM.

dualspiral:
First pass is in `api-8` and more work will happen there. This is after reviews by Zidane, but it isn't that different from here. We remain open for discussion in our normal channels.

PR #1959, opened 2019-02-02
Invantory for API8 (part 1)  Inventory Builders

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2211) (outdated)

Redoing custom Inventory creation.
More changes based on this branch here: https://github.com/SpongePowered/SpongeAPI/pull/1998

The current API is split up into 3 
 - basic inventories - these inventories cannot be opened.
 - viewable inventories - inventories that can be opened.
 - Inventory menus - Wrapper around viewable inventory with various callbacks and utility methods.

also InventoryArchetypes are gone.

#### Basic Inventory ####
`Inventory.Builder`
Inventory is build using a sequence of either
 - indexed Inventory
 - grid Inventory
 - existing Inventory

after you are done building the inventory structure you can set 
 - UUID property 
 - Carrier

Sponge will not allow the resulting inventory to be opened.
For that you'll need a:

#### ViewableInventory ####
You start by choosing one vanilla `ContainerType`.
Then you can build your inventory from various existing Inventories or Dummy Inventories.
If you define a Slot multiple times only the last time will count. (see example below)
When completing the final structure all slots needed for the ContainerType need to be defined.
After you are done building the inventory structure you can set 
 - UUID property 
 - Carrier

If you want to add callbacks to specific Slots you'll need the:

#### InventoryMenu ####
 
All InventoryMenu are based on one ViewableInventory.
It keeps track of Containers opened by `InventoryMenu#open`.

By default all changes to the "top" inventory are cancelled.
You can change this by using `setReadOnly(false)`.

Simple callbacks can be registered for 
- `click`
- `change` (change won't happen when global readonly)
- `readonly` (for specific slots; no effect when already global readonly)
- `close`
and unregistered at any time.

You can change the current Inventory while it is opened by one or more players.
If the new inventory has a matching `ContainerType` Sponge will swap out the Slots silently.
Otherwise inventory will be reopened and all existing callbacks cleared.

Changing the Title will also result in the Container opening again.

Example Usages below


Faithcaio:
#### Example Usage: ####

```java
Inventory inv1 = Inventory.builder().grid(3, 3).completeStructure().build();
Inventory inv2 = Inventory.builder().grid(3, 3).completeStructure().build();
Inventory inv3 = Inventory.builder().grid(9, 3).completeStructure().build();

ViewableInventory inv = ViewableInventory.builder()
        .type(ContainerTypes.CHEST_3X9)
        .grid(inv1.slots(), new Vector2i(3,3), 0)
        .grid(inv2.slots(), new Vector2i(3,3), new Vector2i(3, 1))
        .grid(inv3.slots() /*TODO query for grid*/, new Vector2i(3, 3), new Vector2i(6, 3))
        .slots(Arrays.asList(inv3.getSlot(SlotIndex.of(0)).get()), 37)
        .dummySlots(1, 16)
        .fillDummy()
        .completeStructure()
        .identity(UUID.randomUUID())
        .build();
```


```java
Inventory inv = Inventory.builder().grid(3, 3).completeStructure().build();

ItemStackSnapshot disabled = ItemStack.of(ItemTypes.LIGHT_GRAY_STAINED_GLASS_PANE, 1).createSnapshot();
ItemStackSnapshot emerald = ItemStack.of(ItemTypes.EMERALD, 1).createSnapshot();

ViewableInventory display = ViewableInventory.builder().type(ContainerTypes.DISPENSER)
        .fillDummy().item(disabled)
        .slots(Arrays.asList(inv.query(GridInventory.class).get().getSlot(1,1).get()), new Vector2i(1, 1))
        .completeStructure().build();
display.query(GridInventory.class).get().set(1,1, ItemStack.of(ItemTypes.DIAMOND, 1));

ViewableInventory display2 = ViewableInventory.builder().type(ContainerTypes.DISPENSER)
        .fillDummy()
        .dummySlots(1, new Vector2i(1,1)).item(emerald)
        .completeStructure().build();
display.query(GridInventory.class).get().set(1,1, ItemStack.of(ItemTypes.DIAMOND, 1));


ViewableInventory basicChest = ViewableInventory.builder()
        .type(ContainerTypes.CHEST_3X9)
        .completeStructure()
        .build();


InventoryMenu menu = InventoryMenu.of(display);
menu.open(player);
menu.open(player2);

menu.setCurrentInventory(display2); // matching ContainerType so the inventory is silently swapped
menu.setTitle(Text.of("This reopens containers"));
menu.registerSlotClick((cause, container, slot, slotIndex, clickType) -> checkClick());

menu.setReadOnly(false);
MyHandler handler = new MyHandler();
menu.registerHandler(handler);
menu.registerChange((cause, container, slot, slotIndex, oldStack, newStack) -> checkAllChange());

menu.setReadOnly(true);
menu.setCurrentInventory(basicChest);
menu.unregisterAll(); // already done as changing the ContainerType clears all callbacks
```

Faithcaio:
Not sure about callback registration yet. 
Feels a bit clunky if you have smth. grid based (like give me a callback on the center 7x1 slots)
Also the Interfaces for registration. SlotClickHandler,SlotChangeHandler

NeumimTo:
The set method could be overloaded
```
query(GridInventory.class).get().set(1,1, ItemStack.of(ItemTypes.DIAMOND, 1));
query(GridInventory.class).get().set(1,1, ItemStack.of(ItemTypes.DIAMOND, 1), (container, slot, slotIndex, clickType) ->click_callback);
```
How would you create a single ViewableInventory from slots which are not in same type of inventory eg these slots (blue line)

inventory_thb

 .source(inv2).grid(1,1).from(1,1).at(1,1)

I have no idea what this will do.

Could we have a slots method which would take array of SlotPos
```
 .source(inv2).slots({SlotPos.of(1,1), SlotPos.of(1,2), SlotPos.of(1,3)})
```
or maybe an array of integers, representing inventory slot ids
```
 .source(inv2).slots(1,2,3)
```

Faithcaio:
Calling `source(inv)` sets the inventory where the next calls are taking slots from.
Lets say you've already got
`EquipmentInventory armor`
`GridInventory mainGrid`
`Slot offhand`
```java
.source(armor).slots(4)
.source(mainGrid).slots(5)
.source(offhand).slot()
```
Would then add all the slots you marked to the viewable Inventory.
Ofc. you need to define all slots of the Inventory but you can just fill up the rest with `fillDummy`

`.source(inv).grid(2, 2).from(1,1).at(2,2)`
This adds a 2x2 grid from the source inventory starting at x=1 y=1 to the new inventory starting at x=2 y=2
Source Inventory:
Source
New Inventory
NewInv

SlotPos Array and Int Array is possible.
But for both your examples i'd argue the current API proposal is better:
`grid(1,3).from(1,1)` 1x3 grid starting from x=1 y=1
`slots(3).from(1)`      3 slots starting from index=1

NeumimTo:
>But for both your examples i'd argue the current API proposal is better:
>`grid(1,3).from(1,1)` 1x3 grid starting from x=1 y=1
>`slots(3).from(1)` 3 slots starting from index=1

It depends on your use case i guess, Im letting my users to define inventory section, which is then handled by a different logic. 

Users have to configure the inventory class name and slot ids. With such a trivial mechanism my plugin can hook easily even into an inventory provided by a mod such as baubles, and that way i dont have to write extra code,
I see writing 6 slot ids much easier option than trying to figure out some kind of bounding box, and its deflection from origin.

Faithcaio:
If it is dynamic you can just call `slot().from(...)` multiple times.

src/main/java/org/spongepowered/api/item/inventory/menu/SlotClickHandler.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param container the container
+     * @param slot the primary clicked slot
+     * @param slotIndex the primary clicked slot index
+     * @param clickType the type of click
+     * @param slot2 the secondary slot in the interaction. not null for {@link ClickTypes#KEY_SWAP}
+     *
+     * @return false to prevent the click
+     */
+    boolean handle(Container container, Slot slot, SlotIndex slotIndex, ClickType clickType, @Nullable Slot slot2);
```
Cybermaxke: I don't like the second `slot` as a parameter, maybe we just need specialized click handlers in those cases. IMO there are too many arguments and adding new ones for every special case isn't good.

The SlotIndex is based on the `Container`? Then the parameter is unnecessary unless for convenience.

```java
interface SlotClickHandler {
    boolean handle(Container container, ClickType clickType, Slot slot);

    default boolean handleKeySwap(Container container, Slot hoveredSlot, int key) {
        return handle(container, ClickTypes.KEY_SWAP, hoveredSlot);
    }
}
```

Another thought, what happens when you return `true`/`false`? Does this prevent default behavior if you return `false`?

Maybe it's even an option to allow different `ClickType`s to use different handler interfaces.

Faithcaio: I guess I could make a second handler interface for the KEY_SWAP type
`SlotIndex` is indeed for convenience.
As the javadocs say. You return false to prevent the click from happening.

Faithcaio: Now there is
ClickHandler - no slot
SlotClickHandler - one slot
KeySwapHandler - two slots
also
CloseHandler
SlotChangeHandler

They can all be registered with one class implementing all of them. `#registerHandler`

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
```diff
@@ -0,0 +1,167 @@
...
+    void setTitle(Text title);
+
+    /**
+     * Registers click callbacks. Unregisters previously registered handler of the same type.
+     * <p>Return false in the handler to prevent changes.</p>
+     * <p>Possible handlers: {@link ClickHandler}, {@link SlotClickHandler}, {@link KeySwapHandler}, {@link SlotChangeHandler}, {@link CloseHandler}</p>
+     *
+     * @param handler the handler
+     */
+    void registerHandler(Object handler);
```
XakepSDK: Marker interface?

src/main/java/org/spongepowered/api/item/inventory/menu/KeySwapHandler.java
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param container the container
+     * @param slot the primary clicked slot
+     * @param slotIndex the primary clicked slot index
+     * @param clickType the type of click
+     * @param slot2 the secondary slot in the interaction.
+     *
+     * @return false to prevent the action
+     */
+    boolean handle(Container container, Slot slot, SlotIndex slotIndex, ClickType clickType, Slot slot2);
```
XakepSDK: Container class for these args

src/main/java/org/spongepowered/api/item/inventory/menu/SlotClickHandler.java
```diff
@@ -0,0 +1,45 @@
...
+     * Handler for inventory click callbacks except {@link ClickTypes#KEY_SWAP} see {@link KeySwapHandler}
+     *
+     * @param container the container
+     * @param slot the primary clicked slot
+     * @param slotIndex the primary clicked slot index
+     * @param clickType the type of click
+     *
+     * @return false to prevent the action
+     */
+    boolean handle(Container container, Slot slot, SlotIndex slotIndex, ClickType clickType);
```
XakepSDK: Container class for these

src/main/java/org/spongepowered/api/item/inventory/menu/SlotChangeHandler.java
```diff
@@ -0,0 +1,42 @@
...
+public interface SlotChangeHandler {
+
+    /**
+     * Handler for inventory change callbacks.
+     *
+     * @param container the container
+     * @param slot the changed slot
+     * @param slotIndex the changed slot index
+     */
+    boolean handle(Container container, Slot slot, SlotIndex slotIndex);
```
XakepSDK: Container class for args

src/main/java/org/spongepowered/api/item/inventory/menu/CloseHandler.java
```diff
@@ -0,0 +1,42 @@
...
+@FunctionalInterface
+public interface CloseHandler {
+
+    /**
+     * Handler for inventory closing.
+     *
+     * @param container the container
+     * @param player the viewing player
+     */
+    void handle(Container container, Player player);
```
XakepSDK: Container class for args

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
```diff
@@ -0,0 +1,167 @@
...
+     * @param handler the callback handler
+     */
+    void registerClose(CloseHandler handler);
+
+    /**
+     * Registers a callback handler that prevents any change
+     * <p>You can override the behaviour of {@link #setReadOnly(boolean)} with this.</p>
+     */
+    default void registerReadOnly() {
+        registerChange((container, slot, slotIndex) -> false);
```
XakepSDK: `setReadOnly(false)`?

XakepSDK (changes requested):
I suggest to make it bit harder, but more understandable.
Like you not writing indexes or grid positions, but you take `Slot` instances from `Inventory` by yourself and put them in `ViewableInventory`.

So we can just do this for example:
`insertSlotsAt(0, inv.query(EquipmentInventory.class).slots())`

For grids:
`insertGridAt(0, 0, inv.query(CraftingGridInventory.class).slots())`

instead of `.source(armor).slots(4)`, where `armor` - inventory with `WornEquipmentType` slots.

It's trillion thousand times more understandable.

Add way to get random grids from inventories, e.g.
`inv.query(grid(0,0,3,3))`, where `grid` returns corresponding `QueryOperationType`

Also, change handling feels bad, i write suggestion later.

src/main/java/org/spongepowered/api/item/inventory/menu/KeySwapHandler.java
Faithcaio: Why? Imo that would just be another additional interface for every handler added to the API without much benefit.

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
Faithcaio: Y u no union type java :/

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
Faithcaio: I had it with slotIndex filter before (also allowing multiple handlers). 
But it became too complex in impl.
So this will be gone completly.
`setReadonly(false)` does the same

src/main/java/org/spongepowered/api/item/inventory/menu/KeySwapHandler.java
XakepSDK: This change will avoid long lambda parameter list.
You even can remove these interfaces and use something like `Predicate<KeySwapHandlerData>` in `handleXXX` methods.
`Predicate`, `Function`, `Consumer` everything you want.

Faithcaio: You don't have to use lambdas. You can just pass an instance of smth. that implements all those interfaces or method reference that fits.
Still don't see the benefit of wrapping the data to just unwrap it in the handler again when you want to use it.

XakepSDK: This would give shorter lambda. Compare:
```java
KeySwapHandler = (container, slot, slotIndex, clickType, slot2) => {
    if (slot2 != null) {
        s.peek();
    }
};
```
and this:
```java
Consumer<KeySwapData> = data => {
  data.getSlot2().ifPresent(Slot::peek);
};
```

XakepSDK: Also, my approach is more deprecation/addition safer way.
You could deprecate getter methods and add new getter without breaking anything.
And i hope, it would work like sponge events, with Cause stack, event context and other features of sponge.

Faithcaio: The original idea of the callbacks was to reduce the massive overhead we currently have with inventory events. 
We have to detect and collect all changes and then based on cancel state rollback everything

If you just want to have some kind of immutable menu its easier to just have callbacks for clicks/changes instead.

While your approach would make deprecation/addition easier it also would make it impossible to just implement all those interfaces in one class.
Example here: https://github.com/SpongePowered/SpongeCommon/pull/2211/files#diff-a4229ef51d31c0aab92651c1562a8ab9R435
And let's be honest, we would find a way to break it anyways.

I'm still not convinced that a wrapper around those values is the best way to go.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -520,69 +519,81 @@ default Inventory transform(InventoryTransformation transformation) {
...
         /**
-         * Sets the InventoryArchetype and Properties for a default Inventory of
-         * given {@link Carrier}.
-         *
-         * @param carrier The Carrier class
-         * @return Fluent pattern
+         * The building step. The inventory structure can be completed at any time.
          */
-        Builder forCarrier(Class<? extends Carrier> carrier);
+        interface BuildingStep extends Builder {
```
Zidane: I see no point in separating this out. Should just be on the Builder imo.

src/main/java/org/spongepowered/api/item/inventory/custom/ContainerTypes.java
```diff
@@ -0,0 +1,92 @@
...
+ * An enumeration of all possible {@link ContainerType}s in vanilla minecraft.
+ */
+public final class ContainerTypes {
+
+    // SORTFIELDS:ON
+
+    /**
+     * Size 27 (3x9)
+     */
+    public static final ContainerType CHEST = DummyObjectProvider.createFor(ContainerType.class, "chest"); // "minecraft:chest"
```
Zidane: Honestly I dislike this and DOUBLE_CHEST simply because Mojang can redefine slot layouts/counts at any time (Looking at you 1.14...).

Instead, I'd do something like....

3X9
6X9

The other names are fine simply because they have specific Game logic attached. A Chest is pretty basic.

src/main/java/org/spongepowered/api/item/inventory/menu/ClickHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Handler for inventory click callbacks except without slot.
+     *
+     * @param container the container
+     * @param clickType the type of click
+     *
+     * @return false to prevent the action
+     */
+    boolean handle(Container container, ClickType clickType);
```
Zidane: Whereas Container nets you a lot of information, it looses a lot of the context of the click. Any argument for Cause/EventContext in this case? These aren't real events I'm aware but I feel like valuable information would be lost.

Same argument I made to @dualspiral for moving CommandExecutor over to having Cause/etc.

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
```diff
@@ -0,0 +1,167 @@
...
+    static InventoryMenu of(ViewableInventory inventory) {
+        return inventory.asMenu();
+    }
+
+    /**
+     * Returns the current inventory used in this menu.
+     *
+     * @return the current inventory
+     */
+    ViewableInventory getCurrentInventory();
```
Zidane: I feel that getCurrentInventory is not needed, getInventory should be fine.

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
Zidane: I agree with a marker interface here. Otherwise this encourages someone to pass whatever to it and now you have to perform validation.

It would be different if you made the handlers be annotated methods.

src/main/java/org/spongepowered/api/item/inventory/type/ViewableInventory.java
```diff
@@ -56,5 +70,322 @@
...
      */
     boolean canInteractWith(Player player);
-    
+
+    /**
+     * Returns the {@link ContainerType} of this viewable inventory.
+     *
+     * @return the ContainerType of this viewable inventory.
+     */
+    ContainerType getContainerType();
```
Zidane: getType

src/main/java/org/spongepowered/api/item/inventory/menu/ClickHandler.java
XakepSDK: Agree with `Cause` and `EventContext` thing, it may be very useful.

gabizou: I'll play both sides here:

We don't explicitly need to provide the cause, because the handler a) can be on either side of the thread management, b) still retrieved by `Sponge.getCauseStackManager().getCurrentCause()` if needed, c) Forces our implementation to generate the Cause even if it's not used (since this is not event based, we gain some free performance if the plugin is able to cope without cause usage at all).

On the other hand, Causes are cheap enough that we don't really lose anything by providing the extra information at thetime of the click/handle.

PR #1960, opened 2019-02-02
Do not append spaces after CommandElements whose usages are empty.

ItsDoot:
This PR fixes `GenericArguments.SequenceCommandElement` so that when any of its elements' usages are empty, it will not append extra spaces to the resulting usage.

#### Use Case:

I have a Kotlin annotation-based command system that represents functions' containing classes (needed to be able to call the function with `kotlin-reflect`) as hidden `CommandElement`s (i.e. they aren't part of the command's displayed usage). Currently, whenever the command functions are built with multiple containing classes, there are a bunch of extra spaces appended after the command's name. It would look better if those additional spaces were not there.


dualspiral (approved):
Looks fine to me.

kashike (approved):

PR #1967, opened 2019-02-16
Delegate getNearbyEntities to getEntities

### LINKED ISSUE #1607: Entity#getNearbyEntities will sometimes not include the originating entity
**I am currently running...**
- SpongeVanilla version: 1.12.2-7.0.0-BETA-352
- Java version: 8
- Operating System:

 - [x] Linux (Ubuntu, Debian, Fedora, etc)

- Plugins/Mods: 
 A simple test plugin

**What is the expected result?**
`getNearbyEntities` should include all entities from a given `Entity`'s location, including the `Entity` whose location is used per the javadoc (as it calls another method in `EntityUniverse` that takes a location).

**What is the current result?**
The `Entity` is missing from the resulting collection.

This can be seen by using `getNearbyEntities` with distances of 25 and 50m respectively.

ImMorpheus:
Should fix https://github.com/SpongePowered/SpongeCommon/issues/1607


ItsDoot:
Shouldn't you be using `distanceSquared` as it was before instead of `distance`, to bypass a square root calculation?

PR #1972, opened 2019-02-23
Added the tags API.

Cybermaxke:
**API** | [Common
](https://github.com/SpongePowered/SpongeCommon/pull/2197)
Adds a low level API which represents minecraft tags.

### DIFF
--- a/src/main/java/org/spongepowered/api/Tag.java
+++ b/src/main/java/org/spongepowered/api/Tag.java
@@ -25,7 +25,10 @@
 package org.spongepowered.api;
 
 import com.google.common.reflect.TypeToken;
-import org.spongepowered.api.block.BlockTags;
+import org.spongepowered.api.block.BlockTag;
+import org.spongepowered.api.entity.EntityTag;
+import org.spongepowered.api.fluid.FluidTag;
+import org.spongepowered.api.item.ItemTag;
 
 import java.util.Collection;
 
@@ -33,7 +36,10 @@ import java.util.Collection;
  * Represents a tag which can be applied to a specific {@link CatalogType}.
  *
  * @param <C> The catalog type
- * @see BlockTags
+ * @see BlockTag
+ * @see EntityTag
+ * @see FluidTag
+ * @see ItemTag
  */
 public interface Tag<C extends CatalogType> extends CatalogType {
 
@@ -51,6 +57,23 @@ public interface Tag<C extends CatalogType> extends CatalogType {
      * @return Whether this tag is applied to the catalog type
      */
     boolean contains(C catalog);
+
+    /**
+     * Gets whether this tag is applied to the given catalog types.
+     *
+     * @param iterable The catalog types
+     * @return Whether this tag is applied to the catalog types
+     */
+    boolean containsAll(Iterable<C> iterable);
+
+    /**
+     * Gets whether the specified {@link Tag} in
+     * included in this tag.
+     *
+     * @param tag The tag
+     * @return Whether the tag is included
+     */
+    boolean contains(Tag<C> tag);
 
     /**
      * Gets a unmodifiable {@link Collection} with all the catalog
--- a/src/main/java/org/spongepowered/api/Tag.java
+++ b/src/main/java/org/spongepowered/api/Tag.java
@@ -1,0 +1,62 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
+
+import com.google.common.reflect.TypeToken;
+import org.spongepowered.api.block.BlockTags;
+
+import java.util.Collection;
+
+/**
+ * Represents a tag which can be applied to a specific {@link CatalogType}.
+ *
+ * @param <C> The catalog type
+ * @see BlockTags
+ */
+public interface Tag<C extends CatalogType> extends CatalogType {
+
+    /**
+     * Gets the {@link TypeToken} of the target catalog type.
+     *
+     * @return The target catalog type
+     */
+    TypeToken<C> getTargetType();
+
+    /**
+     * Gets whether this tag is applied to the given catalog type.
+     *
+     * @param catalog The catalog type
+     * @return Whether this tag is applied to the catalog type
+     */
+    boolean contains(C catalog);
+
+    /**
+     * Gets a unmodifiable {@link Collection} with all the catalog
+     * types this tag is applied to.
+     *
+     * @return The collection with catalog types
+     */
+    Collection<C> getAll();
+}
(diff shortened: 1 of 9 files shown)


src/main/java/org/spongepowered/api/Tag.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.Collection;
+
+/**
+ * Represents a tag which can be applied to a specific {@link CatalogType}.
+ *
+ * @param <C> The catalog type
+ * @see BlockTags
+ */
+public interface Tag<C extends CatalogType> extends CatalogType {
```
gabizou, line 44: Can't enforce the `Tag` to only contain `CatalogType`s. Vanilla doesn't have a unified interface for `Tag` to use, so it's just `Tag<T>` sadly. Tags however can be catalog typed.

src/main/java/org/spongepowered/api/Tag.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Gets the {@link TypeToken} of the target catalog type.
+     *
+     * @return The target catalog type
+     */
+    TypeToken<C> getTargetType();
```
gabizou, line 51: No way to guarantee this either, Tags can be completely ignorant (or empty for that matter) to know what their generic type information would contain. Unless you plan on providing `TypeToken.of(Object.class)` as a worst case scenario.

src/main/java/org/spongepowered/api/Tag.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    boolean contains(C catalog);
+
+    /**
+     * Gets a unmodifiable {@link Collection} with all the catalog
+     * types this tag is applied to.
+     *
+     * @return The collection with catalog types
+     */
+    Collection<C> getAll();
```
gabizou, line 84: Possibly could add a method for getting a random element in valid tags by providing random.

src/main/java/org/spongepowered/api/block/BlockTag.java
```diff
@@ -0,0 +1,36 @@
...
+package org.spongepowered.api.block;
+
+import org.spongepowered.api.Tag;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents a {@link Tag} which targets {@link BlockType}s.
+ */
+@CatalogedBy(BlockTag.class)
+public interface BlockTag extends Tag<BlockType> {
```
gabizou, line 34: No sense in having the specific interface...

src/main/java/org/spongepowered/api/block/BlockTags.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Represents all possible {@link BlockTag}s that are known to exist in
+ * vanilla minecraft.
+ */
+public final class BlockTags {
+
+    // SORTFIELDS:ON
+
+    public static final BlockTag ACACIA_LOGS = DummyObjectProvider.createFor(BlockTag.class, "ACACIA_LOGS");
```
gabizou, line 37: Eliminate the interface and declare as `Tag<BlockType>`

gabizou (changes requested):

src/main/java/org/spongepowered/api/Tag.java
ST-DDT: If tags are not bound to a single type, as indicated by your next comment then you could simply skip the type parameter.

liach: Vanilla tag entries are all namespaced afaik, so they do actually contain catalog types.

src/main/java/org/spongepowered/api/Tag.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    TypeToken<C> getTargetType();
+
+    /**
+     * Gets whether this tag is applied to the given catalog type.
+     *
+     * @param catalog The catalog type
+     * @return Whether this tag is applied to the catalog type
+     */
+    boolean contains(C catalog);
```
liach, line 59: Also a contains method for `Tag<C>` argument

liach:
Moreover, do we want to expose the direct components of each tag (so a method to get direct catalog components and another to get directly referenced tags)

src/main/java/org/spongepowered/api/entity/EntityTag.java
```diff
@@ -27,6 +27,9 @@
 import org.spongepowered.api.Tag;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 
+/**
+ * Represents a {@link Tag} which targets {@link EntityType}s.
+ */
 @CatalogedBy(EntityTags.class)
 public interface EntityTag extends Tag<EntityType<?>> {
```
kashike, line 34: should be `EntityTypeTag`

kashike (changes requested):

src/main/java/org/spongepowered/api/entity/EntityTag.java
Cybermaxke: I didn't call the other ones `BlockTypeTag`, `ItemTypeTag`, `FluidTypeTag`. So why should this one be different?

kashike: `Entity` != `EntityType`. They're tags of entity types, not entities. "block type" and "block" are interchangeable

src/main/java/org/spongepowered/api/Tag.java
gabizou: > Vanilla tag entries are all namespaced afaik

But there's no unique interface to attach to them... I could literally make a `Tag<Entity>` and the game will be happy with that tag, even though it's not a valid use of the tag. Similarly, Sponge cannot enforce that the returned types are `CatalogType`s as mods can introduce new Tags of their own classes, which would immediately fail class casts to `CatalogType`.

Cybermaxke: Easy to work around that with the subtypes and wrappers. We don't have to do everything 1:1 with how things are done in vanilla mc.

gabizou: > Easy to work around that with the subtypes and wrappers.

Yes, it's easy to work around with, but it's extra maintenance and potentially bug inducing that I don't want to have to deal with. Since mods already do all sorts of things by having near unlimited access to internals, compared to plugins interacting with the API, I don't want there to be any room for overhead of maintenance costs for a wrapper system, let alone subtypes for an API that is supposed to represent/expose functionality implemented by vanilla. An easy example of how subtyping ends up adding extreme overhead of maintenance costs is the MaterialData/ItemMeta API in Bukkit, and I can tell you right now that that system was already an issue to represent mod provided datas.



> We don't have to do everything 1:1 with how things are done in vanilla mc.

You're right, we're not doing everything 1:1, but core functionality does dictate a lot in how our API can be designed, and the direction in which it is being designed. We've taken a lot of queues in our original API design to avoid wrapping for most things, and lately with 1.13's world gen changes, @Zidane and I already came to the agreement that we'd be making it a point to avoid wrapping at most all costs. These days, looking at the implementation, the biggest offenders of wrapping basically anything is to support API designs that are not representative of their implementation counterparts (Inventory and Data). This bit of API however, is **supposed** to represent the Tag system from vanilla, much like the system that was introduced for what was previously known as [`BlockTrait`s](https://github.com/SpongePowered/SpongeAPI/commit/1d4d3cff300003b9353a0d0415b8345e904da5b2#diff-4e4e8fb34984746e53f060d9a7a78eeb) because it needed to be representable for what is universally used in the implementation (and happily coexists with Data API's counterparts with keys and values).

gabizou: To add on, other examples of wrapper implemented APIs that have since disappeared: Text API implementation (used to be fully implemented in the API, now it's all interfaced onto the implementation using TextComponents as they should), Populators (quite literally re-implemented all vanilla populators with a mix and match of some implementation provided mixed in classes, and others that were full re-implementations using sponge classes), SpawnCauses (eliminated with the context system), etc.

Cybermaxke: Without subtypes, we can't make a Tag a catalogue type because we end up mixing different registries. We could add a prefix like `blocks/<id>` to resolve this, but then we'll still need to get it from somewhere, which can the cause the same problems with mods. This also mismatches ids with vanilla mc. IMO these are different catalog types and shouldn't be mixed, they have different registries in vanilla mc as well. There will always be a workaround to get this done properly if you want them to be catalog types, subtypes or not.

@kashike Maybe you can share your thoughts?

gabizou: > Without subtypes, we can't make a Tag a catalogue type because we end up mixing different registries.

Looking further, it's impossible then to provide all Tags regardless whether they're catalogued for Blocks, Items, or Fluids, since Mods will be able to create their own tag collections and using their own tags, being able to expose tags that are not registered in any of the three default "registries".



> We could add a prefix like `blocks/<id>` to resolve this, but then we'll still need to get it from somewhere, which can the cause the same problems with mods.

Right, because even a mod could make their own TagCollection of Tag<Item> and synchronize it on their own parts, which ends up with sponge not knowing how to represent that tag. I'd say it's easiest then to just have `Tag` exposing it's `CatalogKey` (which it always has to have), and that's about it. We can expose the `TagCollection` object and expose the three types of vanilla ones, but they are all temporarily unable to be `CatalogType`d to be searchable via `GameRegistry`. Given the dynamic data driven nature of Tags, I'd say this is the only way to go for now.

Cybermaxke: If we expose `TagCollection`, we end up with yet another different system. Which is inconsistent with how the rest of the API works. And they could be perfectly catalogued for the ones we know about. We shouldn't care about the custom tag collections that are created by mods.

JBYoshi:

Lignium:
@Cybermaxke Will work on this feature be resumed?

PR #1973, opened 2019-02-23
Improve the weather API to support local weathers based on biomes, etc.

Cybermaxke:


JBYoshi:

PR #1980, opened 2019-03-08
Migrate build scripts to Kotlin-DSL

gabizou:
[SpongeGradle](https://github.com/SpongePowered/SpongeGradle/pull/22) | **SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2374) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/419) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2970)

This is a **working** PR with the development lock step with **SpongeGradle**. While this technically is compatible with Gradle 4.10.3, I've elected to use the latest Gradle 5.6.2 (as of current writing) to take advantage of the improved Kotlin-DSL since those versions, and configuration avoidance throughout the build process. 

Some changes:
- Constants are just constants in files
- Things that should've been made properties are now properties and/or functions in SpongeGradle or in the `gradle.properties` of their respective projects
- Updated to use appropriate Gradle ideologies with configuration avoidance (lazy evaluation)
- Reduced the overall build file size and using as much handling with the gradle plugins to configure most things we've otherwise duplicated and/or scripted with other scripts.


buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -0,0 +1,25 @@
+object Libs {
+
+    const val slf4j = "${Deps.Groups.slf4j}:${Deps.Modules.slf4j}:${Versions.slf4j}"
+    const val guava = "${Deps.Groups.guava}:${Deps.Modules.guava}:${Versions.guava}"
```
stephan-gh: Shared constants are useful if they are used many times and change often. But having more or less every string in a constant actually makes the scripts less readable here in my opinion:

- Instead of having a plain simple `com.google.guava:guava:21.0`, you need to search through several files to find the exact dependencies that are being used by a module: 
  - Find referenced dependencies in `build.gradle.kts`
  - Look at the declarations in `Libs`
  - Get group ID / artifact ID from `Deps`
  - Get version from `Versions`

  Usually this won't be a big problem because Gradle and IDEs can display the resolved dependencies, but sometimes you still look at the source code directly (e.g. on GitHub).

- To declare a new dependency, you need to add lines in up to 5 different files now instead of adding a single line in one file.

Maven group/artifact IDs and Gradle plugin IDs barely ever change and in my opinion they are easier recognizable if referenced directly instead of through a intermediary variable.

As for the versions, I agree that this should be done for the Gradle plugin versions as they still need to be defined in each root Gradle build script, but even dependency versions should be usually only used once. Common dependencies are either added by a shared build script (e.g. test dependencies) or through transitive dependencies (e.g. SpongeAPI depends on plugin-meta, SpongeCommon depends on SpongeAPI).

build.gradle.kts
```diff
@@ -0,0 +1,229 @@
...
+    compileOnly(Libs.guice)
+
+    // High performing cache + guava
+    compileOnly(Libs.caffeine)
+    compileOnly(Libs.caffeine_guava) {
+        exclude(group = Deps.Groups.guava, module = Deps.Modules.guava)
+    }
+
+    // Plugin meta
+    compileOnly(Libs.plugin_meta)
```
stephan-gh: Why are all these `compileOnly` now? This means that they will not be exposed to any modules depending on SpongeAPI (plugins, implementations).

build.gradle.kts
```diff
@@ -0,0 +1,229 @@
...
+    }
+    val sourceJar by registering(Jar::class) {
+        archiveClassifier.set("sources")
+        from(sourceSets["main"].allSource)
+        from(sourceSets["ap"].allSource)
+    }
+    val javadocJar by registering(Jar::class) {
+        dependsOn(javadoc)
+        archiveClassifier.set("javadoc")
+        from(javadoc.get().destinationDir)
```
stephan-gh: `from(javadoc)`, remove `dependsOn(javadoc)`

build.gradle.kts
```diff
@@ -0,0 +1,229 @@
...
+    }
+    shadowJar {
+        archiveClassifier.set("shaded")
+    }
+    val sortClassFields by existing(TaskSortClassMembers::class) {
+        add(sourceSets.main.get(), "org.spongepowered.api.CatalogTypes")
```
stephan-gh: The duplication of `sourceSets.main.get()` is rather annoying here. With the amount of files we list here, one of the following options might be nicer:

- Search through all source files for `SORTFIELDS:ON`. Then we don't need to list any of them here.
- Add a `addAll()` method to `TaskSortClassMembers`
- Use something like
    ```kts
    listOf(
        "org.spongepowered.api.CatalogTypes",
        "org.spongepowered.api.advancement.AdvancementTypes", 
        "org.spongepowered.api.advancement.criteria.trigger.Triggers",
    ).forEach {
        add(sourceSets.main.get(), it)
    }
    ```

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+}
+repositories {
+    mavenCentral()
+    jcenter()
+    maven("https://files.minecraftforge.net/maven")
+
+}
+
+dependencies {
+    compileOnly("net.minecraftforge.gradle:ForgeGradle:2.3-SNAPSHOT")
```
stephan-gh: Is it possible to move all ForgeGradle related code to SpongeCommon? SpongeAPI does not rely on proprietary Minecraft code and we should try to keep it that way for other implementations (e.g. Lantern).

settings.gradle.kts
```diff
@@ -0,0 +1,9 @@
+pluginManagement {
+    repositories {
+        jcenter()
+        maven("https://repo.spongepowered.org/maven")
+        gradlePluginPortal()
```
stephan-gh, line 6: `gradlePluginPortal()` includes `jcenter()`, so

```kts
repositories {
    gradlePluginPortal()
    maven("https://repo.spongepowered.org/maven")
}
```
should be equivalent. However, I believe we have most (all?) of our Gradle plugins on the Gradle plugin portal anyway.

That reminds me, I did message them to transfer all plugins to the SpongePowered account, but never got any reply...

build.gradle.kts
```diff
@@ -0,0 +1,229 @@
...
+        include("org/spongepowered/api/event/*/**/*")
+        exclude("org/spongepowered/api/event/cause/")
+        exclude("org/spongepowered/api/event/filter/")
+        exclude("org/spongepowered/api/event/impl/")
+    }
+    findByName("setupDecompWorkspace")?.apply {
+        dependsOn(genEventImpl.get())
+    }
+    compileJava {
+        dependsOn(genEventImpl.get())
```
stephan-gh: Is this necessary? event-impl-gen should do this automatically.

build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+    // The sponge buildscript will apply various plugins that are
+    // defined in the buildscript classpath (including plugin versions)
+    // that are not listed in Plugins or Versions in SpongeAPI.kt.
+    // They are uniquely defined in buildSrc/build.gradle.kts
+    id(Plugins.sponge)
+    id(Plugins.`event-impl-gen`) version Versions.`event-impl-gen`
+}
+
+apply(plugin = Plugins.spongegradle)
+apply(plugin = Plugins.spongemeta)
```
Niko-sk2x: I'm really curious how this doesn't break automatic generated accessors in kotlin dsl. Last time I tried (gradle 4.9 and 4.8) automatic generated accessors wouldn't work if `apply plugin` was used.

gabizou: @Barteks2x because the main gradle plugin is still depended on in the class path, the "sub plugins" are still available to the class path, but not normally applied. Taking advantage of that allows us to apply it and use the `getting` or `existing` accessors with the appropriate classes.

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+    gradlePluginPortal()
+    jcenter()
+}
+
+dependencies {
+    implementation("net.minecrell.licenser:net.minecrell.licenser.gradle.plugin:0.4.1")
```
jamierocks, line 13: Any reason these can't be used in the `plugins` block above?

PR #1987, opened 2019-04-01
Sponge Schematic Format v2 Proposal

gabizou:
[Specification](https://github.com/SpongePowered/Schematic-Specification/pull/13) | **SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2231)

This adds API to `7.2` with default method implementations to avoid backwards incompatibilities with any plugins who implement the `Schematic` interface. Likewise, due to the nature of older schematics not having the same data as the new format would support, the older data will be considered `Optional`.

It is intended that the `Optional` information (such as `Collection<EntityArchetype>` or `Optional<MutableBiomeVolume>`) will remain optional and we will be offering those options when schematics are being created via builders.

Implementation is pending currently. Feedback is requested on the API, but Javadocs are not completed.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/EntityArchetype.java
+++ b/src/main/java/org/spongepowered/api/entity/EntityArchetype.java
@@ -28,11 +28,13 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.Archetype;
 import org.spongepowered.api.data.DataContainer;
 import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.Queries;
 import org.spongepowered.api.data.key.Key;
 import org.spongepowered.api.data.manipulator.DataManipulator;
 import org.spongepowered.api.data.persistence.DataBuilder;
 import org.spongepowered.api.data.persistence.InvalidDataException;
 import org.spongepowered.api.data.value.BaseValue;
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/api/entity/EntityArchetype.java
```diff
@@ -65,6 +67,11 @@ static EntityArchetype of(EntityType type) {
      * Gets the raw {@link Entity} data that would be applied to the generated
      * entity. Note that this is a copied container.
      *
+     * <p>Note: While normally, an archetype does not contain position information,
```
Deamon5550, line 70: Does it make sense to create a new type extending EntityArchetype which essentially archetype+position?

gabizou: I was debating about it. Something like SchematicEntity? But that'd be a little enterprisey. It's also possible to grab entities from the ArchetypeVolume when creating the archetype volume through `Extent`, so that they're not limited to just Schematics (I went through to update implementation and found that if it's easier to basically keep pre-built plugins working with the same API and including entities by default.

So far, from my implementation in progress, I've been able to avoid modifying the CopyPasta plugin entirely, using the same API without modification.

PR #1990, opened 2019-04-17
Add plugin adapters.

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/2238) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/2694) | Vanilla | [KotlinAdapter](https://github.com/Cybermaxke/KotlinAdapter)

Supports alternate plugin instantiation and injections. Can be used e.g. for kotlin/scala object support, kotlin delegate property injections, custom plugin injections, etc.

I am going to write a test project to demonstrate how it can be used. SpongeVanilla support will follow once it's successfully tested on SpongeForge.

Kotlin Adapter:
https://github.com/Cybermaxke/KotlinAdapter/blob/master/src/main/kotlin/org/lanternpowered/kt/adapter/KotlinAdapter.kt

Kotlin Test Plugin:
https://github.com/Cybermaxke/KotlinAdapter/blob/master/src/main/kotlin/org/lanternpowered/kt/plugin/TestPlugin.kt

### DIFF
--- a/src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
+++ b/src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.plugin;
 
+import com.google.inject.Guice;
 import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.Scopes;
@@ -49,10 +50,11 @@ public interface PluginAdapter {
      * all plugin specific {@link Injector}s. The returned injector will be used
      * to construct the plugin defined {@link Module}s.
      *
+     * @param defaultModule De default module
      * @return The global injector
      */
-    default Injector createGlobalInjector(Injector defaultInjector) {
-        return defaultInjector;
+    default Injector createGlobalInjector(Module defaultModule) {
+        return Guice.createInjector(defaultModule);
     }
 
     /**
--- a/src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
+++ b/src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
@@ -38,12 +38,25 @@ import java.util.Optional;
  *
  * <p>A plugin needs to specify this in the {@link Plugin} annotation.</p>
  *
- * <p>A plugin adapter requires a empty and public constructor.</p>
+ * <p>The plugin adapter will be constructed globally by a
+ * {@link Injector}. All injections that aren't plugin specific will
+ * be available for the {@link PluginAdapter}.</p>
  */
 public interface PluginAdapter {
 
     /**
-     * Gets the {@link Injector} that will be used to construct a plugin instance. Proper
+     * Creates the global {@link Injector} that will be used to construct
+     * all plugin specific {@link Injector}s. The returned injector will be used
+     * to construct the plugin defined {@link Module}s.
+     *
+     * @return The global modules
+     */
+    default Injector createGlobalInjector(Injector defaultInjector) {
+        return defaultInjector;
+    }
+
+    /**
+     * Creates the {@link Injector} that will be used to construct a plugin instance. Proper
      * bindings should be applied to the {@link Injector} that
      * {@code injector.getInstance(pluginClass)} would return the plugin instance.
      *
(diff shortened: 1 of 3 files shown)


XakepSDK:
Can you also add support for guice modules for plugins?
E.g. developer specifies module class in plugin annotation and then impl. installs that module for plugin's injector.

src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param defaultInjector The default injector
+     * @param pluginModules A immutable list of modules which are provided by the plugin
+     * @return The injector
+     */
+    <T> Injector getInjector(PluginContainer pluginContainer, Class<T> pluginClass, Injector defaultInjector, List<Module> pluginModules);
+
+    /**
+     * Represents the default {@link PluginAdapter}.
+     */
+    final class Default implements PluginAdapter {
```
kashike: Should not be implemented here - should be an empty interface, and implemented elsewhere.

kashike (changes requested):

src/main/java/org/spongepowered/api/plugin/PluginAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+     * <p>Calling {@link PluginContainer#getInstance()} during this
+     * method will result in a {@link IllegalStateException}.</p>
+     *
+     * @param pluginContainer The plugin container the injector is being created for
+     * @param pluginClass The plugin class
+     * @param defaultInjector The default injector
+     * @param pluginModules A immutable list of modules which are provided by the plugin
+     * @return The injector
+     */
+    <T> Injector getInjector(PluginContainer pluginContainer, Class<T> pluginClass, Injector defaultInjector, List<Module> pluginModules);
```
kashike: `createInjector`

kashike (changes requested):

mattmess1221:
👎 for cluttering up the `@Plugin` class.

If we go the route that FML went, the annotation would be specific to the default java loader. The Kotlin loader would have to provide its own loading scheme.

I would also like to see this done as a [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html).

Cybermaxke:
I don't think it's cluttering the annotation, it's not like it's required which could clutter the usage. Same goes for the modules. We could always move them to separate annotations if needed.

Going down the road of reimplementing the complete loader is going to complicate things, you'll need to reimplement the `PluginContainer`, handle class loading, annotation scanning. And in the end, we'll just end up with yet another `Plugin` annotation. I also don't see how this complete control is going to add any more value than what's already possible by injector transformations.

The goal of the plugin adapters isn't to reimplement of what's already there, replacing the complete loading system isn't worth it IMO.

If you also look at what forge uses it for in there scala adapter, it's just to provide language specific injections. Which can easily be handled by guice.

pie-flavor:
We'd need some way of instantiating PluginContainer in a safe way if we had separate annotations. This actually seems like *less* bloat to add another field to Plugin.

Katrix:
Will there be any way to avoid having to interface with Guice so that I can wire up by stuff statically? I don't really see any value in this from a Scala point of view at least if I can't.

Also, a bit unsure what you're talking about with the Scala adapter for Forge injecting language specific stuff. It doesn't inject anything. No mod or plugin annotation, or anything like that. You specify the name of the plugin object in the toml file, and go on your way.

Cybermaxke:
You don't have to use guice injections in your scala plugin, sponge uses it to construct the instance. You can register a provider and implement it however you like.

As for the scala adapter, I looked at a older version that didn't use any config, when it was still in Forge.

I also don't feel like exposing the construction of `PluginContainer`s is a good idea, maybe a way to register a alternate annotation and map those to PluginContainer values is better? And optionally registering adapters, etc.
E.g. 
```kotlin
annotation class KotlinPlugin(val id: String, val name: String)

// Find a way to register annotations?
registry.registerPluginAnnotation(KotlinPlugin::class.java, KotlinPlugin::id).mapName(KotlinPlugin::name).adapter(MyAdapter())
```

pie-flavor:
Actually it would be nice to not have to use the Plugin annotation, or any annotation at all. This API could then be extended to stuff not even Java-based, such as scripting plugins, so they could register each script set as a plugin.

dualspiral:
Are we going to revisit this for API 7, or are we just going to leave this for the [Launcher](https://github.com/SpongePowered/plugin-spi) project in API 8?

Zidane:
I am closing this.

As dual linked, plugin-spi is very lightweight and super simple to implement a language loading service for other JVM based languages (or even ones that aren't anchored in the JVM). You can easily add kotlin/scala features to your own service.

No plans for api-7 to gain this change.

PR #1991, opened 2019-04-17
Fix two typos

ghost:
This fixes two typos in the documentation of the `UseItemStackEvent` event.


ghost:
Well, apparently it's not ok to submit pull requests that are so small.

gabizou:
You could tackle cleaning up the OCD issues existing in the API, usually one of the core devs will perform such a task, but since the API hasn't been in heavy development as of late, I'd permit a contribution of that kind, that way it's a scatter gun PR that fixes all the OCD issues at once, instead of several smaller commits/prs.

Issue #1935, opened 2018-12-06
Create a BlockPositionCommandElement


ryantheleach:
https://github.com/SpongePowered/SpongeAPI/issues/1917 Shows that some confusion can arise using the existing Vector command elements.

I propose splitting the generic Vector command element to not accept #target and only accept raw vectors and instead introduce 2 alternatives.

BlockPositionCommandElement
and
ExactPositionCommandElement

BlockPositionCommandElement could potentially have additional parsing options.

e.g. if the player capable of flight? Should the nearest solid floor be used? Whether a block, or a space ontop, infront, of a block should be selected etc.

This would be a utility on top of the ray tracing to aid auto complete.

ExactPositionCommandElement would act like the current Vector3 command element.

Issue #1917, opened 2018-09-29
Incorrect work of GenericArguments#vector3d


Lignium:
Hi, SpongePowered!
When calling command with the "#target" special word, Vector3d argument points to a side block.

For example:

```
H> - the player head
.  - air
S  - the side block (air)
B  - the target block

H>.....SB
```

ST-DDT:
Have you verified this behavior for all directions (NESWUD)? IIRC there is/was something similar where the ray ended on the block surface and if you use the surface coordinate to get the block then you would end up with the wrong block

Lignium:
I tested it again. A command receives a position of point at a side of block, but not a position of block. Quite unexpectedly... If it was conceived, then an argument should be made to transfer to command coordinates of blocks, that is, Vector3i.

ST-DDT:
Could you post the code to help us to reproduce the error?

Lignium:
Simple command with Vector3d argument. Cloud particles spawn on sides of blocks.
https://gist.github.com/Lignium/93346438e007c132f04b5a713d3b9194

dualspiral:
This isn't a bug in the way you're thinking it is. The element is doing the right thing, returning a point on the boundary between two blocks. It's the assumption of which block the resulting co-ordinate will pick that's the issue.

> Have you verified this behavior for all directions (NESWUD)? IIRC there is/was something similar where the ray ended on the block surface and if you use the surface coordinate to get the block then you would end up with the wrong block

This is because [it's using a block ray to trace the line of sight to the absolute first point between air and block](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/command/args/GenericArguments.java#L1584-L1588). The intersection point happens to be at `<something>.0`, which is the boundary between two blocks - so which one does it choose? Ultimately, it truncates the decimal - so hits on the high end of the co-ordinate of the block you're interested in will actually select the air block - this will happen if you are predominately looking west, north or down.

> I tested it again. A command receives a position of point at a side of block, but not a position of block. Quite unexpectedly... 

Maybe it should be clearer that this is the case, but it's using a block ray under the hood and it is choosing the **exact point the player is looking at**. The javadocs are lacking in this though, I would agree, hence the confusion.

> then an argument should be made to transfer to command coordinates of blocks, that is, Vector3i.

[...]

PR #1995, opened 2019-05-01
[WIP] API 8.0.0

Zidane:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2519)

This is purely for a marker to see what is changing. We're not intending on merging the branches into each other. `stable-7` will be merged in to keep `api-8` up to date with whatever small code effects take place.


Cybermaxke:
Comments from https://github.com/SpongePowered/SpongeAPI/commit/3a583919c9a1033fbedca37c4b40a7aa89275ad7 still need to be addressed.

src/main/java/org/spongepowered/api/data/DataManager.java
```diff
@@ -192,14 +158,14 @@
     <T> Optional<DataTranslator<T>> getTranslator(Class<T> objectClass);
 
     /**
-     * Gets all {@link Class}es of all {@link DataManipulator}s registered for
+     * Gets all {@link Class}es of all {@link DataRegistration}s registered for
```
ItsDoot, line 163: This should be `Gets all {@link DataRegistration}s for`

src/main/java/org/spongepowered/api/CatalogTypes.java
```diff
@@ -122,22 +114,16 @@
 
     public static final Class<ArmorType> ARMOR_TYPE = ArmorType.class;
 
-    public static final Class<Art> ART = Art.class;
+    public static final Class<ArtType> ART = ArtType.class;
```
Lignium: ART -> ART_TYPE

src/main/java/org/spongepowered/api/CatalogTypes.java
```diff
@@ -324,11 +262,7 @@
 
     public static final Class<Statistic> STATISTIC = Statistic.class;
 
-    public static final Class<StatisticType> STATISTIC_TYPE = StatisticType.class;
-
-    public static final Class<StoneType> STONE_TYPE = StoneType.class;
-
-    public static final Class<StructureMode> STRUCTURE_MODE = StructureMode.class;
+    public static final Class<StatisticCategory> STATISTIC_TYPE = StatisticCategory.class;
```
Lignium: STATISTIC_TYPE -> STATISTIC_CATEGORY

src/main/java/org/spongepowered/api/CatalogTypes.java
```diff
@@ -346,31 +280,25 @@
 
     public static final Class<TransactionType> TRANSACTION_TYPE = TransactionType.class;
 
-    public static final Class<TreeType> TREE_TYPE = TreeType.class;
+    public static final Class<WoodType> TREE_TYPE = WoodType.class;
```
Lignium: TREE_TYPE -> WOOD_TYPE

README.md
```diff
@@ -35,16 +35,18 @@ Are you a talented programmer looking to contribute some code? We'd love the hel
...
-[Issues]: https://issues.spongepowered.org/
-[Wiki]: https://github.com/SpongePowered/SpongeAPI/wiki/
+[Issues]: https://github.com/SpongePowered/SpongeAPI/issues
+[Documentation]: https://docs.spongepowered.org/
 [Java]: http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html
 [Source]: https://github.com/SpongePowered/SpongeAPI/
 [MIT License]: https://www.tldrlegal.com/license/mit-license
-[Community Chat]: https://webchat.esper.net/?channels=sponge
-[Development Chat]: https://webchat.esper.net/?channels=spongedev
+[Community Discord]: https://discord.gg/PtaGRAs
```
ItsDoot, line 43: This should be using our vanity url: https://discord.gg/sponge

src/main/java/org/spongepowered/api/entity/living/animal/Ocelot.java
```diff
@@ -25,32 +25,21 @@
...
-     * Gets the {@link Value} for the {@link OcelotType} of this
+     * Gets the {@link Value.Mutable} for the {@link CatType} of this
      * {@link Ocelot}.
      *
-     * @return The ocelot variant value
+     * @return The ocelot type
      */
-    default Value<OcelotType> variant() {
-        return getValue(Keys.OCELOT_TYPE).get();
+    default Value.Mutable<CatType> type() {
```
Lignium: Ocelots and cats are different mobs. Ocelots do not have type, but cats have.

Zidane: That...shouldn't be here. I shall remove it.

Lignium:
Is it possible to rename `ChannelBuf` to `ChannelBuffer`? "Buf" is not even a generally accepted notation, such as "arg", for example.

ItsDoot:
> 
> 
> Is it possible to rename `ChannelBuf` to `ChannelBuffer`? "Buf" is not even a generally accepted notation, such as "arg", for example.

There is at least some precedent in having it, given Netty has ByteBuf, which ChannelBuf is a wrapper around. Not that I'm against changing it, just noting why it might be named what it is currently.

liach:
Can we remove jsr305 for API 8?

XakepSDK:
1. We can move to this: https://checkerframework.org/
2. Or use `--patch-module` `javac` option to fix split modules

src/main/java/org/spongepowered/api/world/schematic/PaletteTypes.java
```diff
@@ -24,22 +24,14 @@
...
-    /**
-     * A local palette containing only a subset of the global palette.
-     */
-    public static final BlockPaletteType LOCAL = DummyObjectProvider.createFor(BlockPaletteType.class, "LOCAL");
-
-    // Suppress default constructor to ensure non-instantiability.
-    private BlockPaletteTypes() {
-        throw new AssertionError("You should not be attempting to instantiate this class.");
-    }
+public class PaletteTypes {
```
Lignium: Missing default constructor with assertion error.

liach: Should be final class.

src/main/java/org/spongepowered/api/command/CommandResults.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public final static CommandResult SUCCESS = CommandResult.builder().setResult(1).build();
+
+    /**
+     * Indicates that the command executed but was unable to carry out its task.
+     */
+    public final static CommandResult EMPTY = CommandResult.builder().build();
+
+    private CommandResults() {
+        throw new AssertionError("You shouldn't be trying to instantiate this");
```
Lignium, line 43: At the end there should be an exclamation mark.

src/main/java/org/spongepowered/api/command/ParseResults.java
```diff
@@ -0,0 +1,36 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.command;
+
+import org.spongepowered.api.command.parameter.ArgumentReader;
+import org.spongepowered.api.command.parameter.CommandContext;
+
+public interface ParseResults {
```
Lignium: Missing javadoc

src/main/java/org/spongepowered/api/command/manager/CommandManager.java
```diff
@@ -0,0 +1,170 @@
...
+     * Suggests possible completions based on the input argument string.
+     *
+     * @param arguments The arguments
+     * @return The completions
+     */
+    List<String> suggest(String arguments);
+
+    /**
+     * Suggests possible completions based on the input argument string,
+     *  with a provided object that is both a {@link Subject} for permission
```
Lignium: Excess spaces at the beginning of subsequent lines.

src/main/java/org/spongepowered/api/command/manager/CommandManager.java
```diff
@@ -0,0 +1,170 @@
...
+     *
+     * @param subjectReceiver The {@link Subject} & {@link MessageReceiver}
+     * @param arguments The arguments
+     * @return The completions
+     */
+    <T extends Subject & MessageReceiver> List<String> suggest(T subjectReceiver, String arguments);
+
+    /**
+     * Suggests possible completions based on the input argument string,
+     *  with a provided a {@link Subject} for permission checks and a
```
Lignium: Here too.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -0,0 +1,471 @@
...
+        /**
+         * Provides the logic of the command.
+         *
+         * <p>This is only optional if child commands are specified. This will
+         * replace any previous executor that has been set.</p>
+         *
+         * @param executor The {@link CommandExecutor} that will run the command
+         * @return This builder, for chaining
+         */
+        Builder setExecutor(CommandExecutor executor);
```
Lignium, line 443: For builders, the "set" prefix should not be written. Missing "getExecutor" in Command.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -0,0 +1,471 @@
...
+         * {@link #setExecutionRequirements(Predicate)}. However, note that
+         * setting a permission here will not override anything set in
+         * {@link #setExecutionRequirements(Predicate)}, both will be checked
+         * during execution.</p>
+         *
+         * @param permission The permission that is required, or {@code null}
+         *                   for no permission
+         * @return This builder, for chaining
+         */
+        Builder setPermission(@Nullable String permission);
```
Lignium, line 525: Missing "getPermission" in Command.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -0,0 +1,471 @@
...
+         * Sets a function that determines what is required of the provided
+         * {@link Cause} and {@link CommandSource} before this command executes.
+         *
+         * <p>Any requirements here are in addition to the permission check
+         * from {@link #setPermission(String)}</p>
+         *
+         * @param executionRequirements A function that sets the
+         * @return This builder, for chaining
+         */
+        Builder setExecutionRequirements(@Nullable Predicate<Cause> executionRequirements);
```
Lignium: Missing "getExecutionRequirements" in Command.

src/main/java/org/spongepowered/api/command/manager/CommandMapping.java
```diff
@@ -0,0 +1,88 @@
...
+     */
+    Optional<Command> getCommand();
+
+    /**
+     * Gets a {@link Predicate} that
+     *
+     * @return A {@link Predicate} that determines whether the command can be
+     *         run by the given {@link Cause}.
+     */
+    Predicate<Cause> getRequirements();
```
Lignium: It seems this method does not fit here. This method is contained in Command, as well as many others that are not specified here. May be this method is needed here, but you need to add a lot of others from Command class?

\+ Javadoc comment is not fully written.

src/main/java/org/spongepowered/api/command/parameter/CommandContext.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param <T> the expected type of the argument
+     * @return the argument
+     * @throws NoSuchElementException if no value for the key was found
+     * @throws IllegalArgumentException if more than one value for the key was
+     *                                  found
+     */
+    <T> T requireOne(Parameter.Key<T> key) throws NoSuchElementException, IllegalArgumentException;
+
+    /**
+     * Gets all values for the given argument. May return an empty list if no#
```
Lignium: "#" symbol at the end :)

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -0,0 +1,1236 @@
...
+     * with the parameters supplied to {@link Command}. It is
+     * implementation dependent as to what happens if a subcommand
+     * fails to parse.</p>
+     *
+     * @param subcommand The {@link Command} to execute
+     * @param alias The first alias of the subcommand
+     * @param aliases Subsequent aliases, if any
+     * @return The {@link Subcommand} for use in a {@link Parameter} chain
+     */
+    static Subcommand subcommand(Command subcommand, String alias, String... aliases) {
```
Lignium: alias -> primaryAlias
aliases -> secondaryAliases

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -0,0 +1,1236 @@
...
+
+    /**
+     * Returns a {@link Parameter.SequenceBuilder} that parses arguments using
+     * the supplied parameters in order.
+     *
+     * @param parameter The first {@link Parameter} in the sequence
+     * @return The {@link Parameter.SequenceBuilder}, to continue building the
+     *         chain
+     */
+    static Parameter.SequenceBuilder seq(Parameter parameter) {
```
Lignium: Why abbreviate the method name?

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -0,0 +1,1236 @@
...
+     * Parses the next element(s) in the {@link CommandContext}
+     *
+     * @param args The {@link ArgumentReader.Mutable} containing the strings
+     *             that need to be parsed
+     * @param context The {@link CommandContext.Builder} that contains the
+     *                current state of the execution
+     * @throws ArgumentParseException thrown if the parameter could not be
+     *      parsed
+     */
+    void parse(ArgumentReader.Mutable args, CommandContext.Builder context) throws ArgumentParseException;
```
Lignium: args -> arguments

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -0,0 +1,1236 @@
...
+     *
+     * @param args The {@link ArgumentReader} containing the strings that need
+     *             to be parsed
+     * @param context The {@link CommandContext} that contains the
+     *                current state of the execution.
+     * @return The potential completions.
+     * @throws ArgumentParseException thrown if the parameter could not be
+     *      parsed
+     */
+    List<String> complete(ArgumentReader args, CommandContext context) throws ArgumentParseException;
```
Lignium: args -> arguments

src/main/java/org/spongepowered/api/command/parameter/managed/ValueCompleter.java
```diff
@@ -0,0 +1,47 @@
...
+public interface ValueCompleter {
+
+    /**
+     * Gets valid completions for this command.
+     *
+     * @param reader The {@link ArgumentReader.Immutable} containing the arguments
+     *               that needs to be completed.
+     * @param context The {@link CommandContext} that contains the parsed arguments
+     */
+    List<String> complete(ArgumentReader.Immutable reader, CommandContext context);
```
Lignium: reader -> arguments

src/main/java/org/spongepowered/api/command/source/RconSource.java
```diff
@@ -0,0 +1,45 @@
...
+ * Represents an Rcon client.
+ */
+public interface RconSource extends RemoteSource {
+
+    /**
+     * Gets whether this client is logged in, or authenticated.
+     *
+     * @return Whether this client is logged in
+     */
+    boolean getLoggedIn();
```
Lignium: May be "isLoggedIn"?

Lignium (changes requested):

src/main/java/org/spongepowered/api/command/Command.java
dualspiral: Wrong in this case. The `set` prefix indicates that in the builder, the executor will be replaced, not added to, unlike other methods in the builder.

This was discussed as part of the original PR. `set` stays.

src/main/java/org/spongepowered/api/command/source/RconSource.java
XakepSDK: isAuthorized?

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
dualspiral: No guarantee that `alias` will become the primary alias.

src/main/java/org/spongepowered/api/command/manager/CommandMapping.java
dualspiral: https://github.com/SpongePowered/SpongeAPI/pull/1995/files/5c80ba52af57cbaf90091cb56e4851ceb8da95e0..b40291577550b9dff9e0accf1755f365db77eeb8#diff-232ba333b2f62e82e882688724fff322R69

`Command` might be be available, hence it being here.

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
Lignium: Hmm... For sure! This is a subcommand.

src/main/java/org/spongepowered/api/command/source/RconSource.java
dualspiral: This will be moved once a better location for RCON has been decided on - note that this was from API 7 and has just been copied over.

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
dualspiral: Okay, I misread this - but I'm really not sold on that naming - primary/secondary alias is not appropriate here. It's just a set of aliases, this is a way to guarantee one alias and thus flag up to IDEs if the alias is forgotten. That's all.

alias/aliases is fine here. Javadocs should describe this enough.

src/main/java/org/spongepowered/api/command/Command.java
dualspiral: I'll update the javadocs, but it becomes part of `canExecute(Cause)`

src/main/java/org/spongepowered/api/command/Command.java
dualspiral: I'll update the javadocs, but it becomes part of `canExecute(Cause)`

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
dualspiral: Hangover from API 7, but `sequence` doesn't sit _completely_ well with me. Names are welcome for what this is.

Lignium:
@dualspiral You have no methods in interfaces for constructing a builder. Have you forgotten about them?

Zidane:
> 
> 
> Should be a bounded value then

@gabizou 
Vanilla doesn't actually enforce it to go to 24000 which is why I didn't bound it.

src/main/java/org/spongepowered/api/resourcepack/ResourcePack.java
```diff
@@ -68,4 +95,10 @@
      */
     Optional<String> getHash();
 
+    interface Factory {
```
liach, line 98: Consider merging #1706 for this pack change.

liach:
@Cybermaxke Please consider #1706 for the resource pack refactors.

XakepSDK:
Boxing. How boxing will be handled in data and property api?

kashike:
> Boxing. How boxing will be handled in data and property api?

https://github.com/SpongePowered/SpongeAPI/pull/1996#issuecomment-489378650

src/main/java/org/spongepowered/api/text/channel/ChatTypeMessageReceiver.java
```diff
@@ -30,14 +30,29 @@
...
+     * Sends a message as a {@link ChatTypes#CHAT} message.
+     *
+     * <p>If text formatting is not supported in the implementation
+     * it will be displayed as plain text.</p>
+     *
+     * @param message The message to send
+     */
+    @Override
+    default void sendMessage(Text message) {
+        sendMessage(ChatTypes.CHAT, message);
```
ItsDoot: Shouldn't the default `ChatType` be `ChatTypes.SYSTEM`, since most plugins' messages sent to players aren't really chat?

dualspiral: [Just following the conversation here](https://github.com/SpongePowered/SpongeAPI/pull/2004#discussion_r285146287) but I see your point.

@Zidane - thoughts?

kashike: Yes, it should be `SYSTEM`.

Zidane: Well I could be wrong for once :p.

dualspiral: Yeah, I just blindly did it to get out of the way. It'll get changed.

src/main/java/org/spongepowered/api/command/Command.java
Lignium: @dualspiral Where was it discussed? The fact is that in API 7 the style without the "set" prefix is ​​used everywhere, as is the case everywhere. In general, the "set" prefix assumes that a method sets a value and returns void or boolean.

dualspiral: Mostly internal, the outcome was https://github.com/SpongePowered/SpongeAPI/pull/1587#discussion_r129314548 - further discussion was also in the PR itself.

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
```diff
@@ -0,0 +1,118 @@
...
+    // SORTFIELDS:ON
+
+    /**
+     * Size 27 (3x9 grid)
+     */
+    public static final ContainerType CHEST_3X9 = DummyObjectProvider.createFor(ContainerType.class, "chest_3x9"); // "minecraft:chest"
+    /**
+     * Size 54 (6x9 grid)
+     */
+    public static final ContainerType CHEST_6x9 = DummyObjectProvider.createFor(ContainerType.class, "chest_6x9"); // "minecraft:chest"
```
Lignium: `CHEST_6x9` -> `CHEST_6X9`

src/main/java/org/spongepowered/api/item/inventory/menu/InventoryMenu.java
```diff
@@ -0,0 +1,162 @@
...
+import org.spongepowered.api.item.inventory.menu.handler.SlotClickHandler;
+import org.spongepowered.api.item.inventory.type.ViewableInventory;
+import org.spongepowered.api.text.Text;
+
+import java.util.Optional;
+
+/**
+ * Helper for Menus based on Inventories.
+ * <p>This helper provides simple callbacks that can be used instead of listening to inventory events.</p>
+ * <p>InventoryMenus are by default readonly and automatically prevent any changes made by players in the menu.</p>
```
Lignium, line 43: readonly -> read-only

src/main/java/org/spongepowered/api/command/parameter/managed/ValueParser.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param parameterKey The {@link Parameter.Key} of the parameter being parsed
+     * @param args The {@link ArgumentReader} that contains the unparsed arguments
+     * @param context The {@link CommandContext} containing the state about this command
+     * @return Returns the value(s)
+     * @throws ArgumentParseException if a parameter could not be parsed
+     */
+    Optional<? extends T> getValue(
+            Parameter.Key<T> parameterKey,
+            ArgumentReader.Mutable args,
```
Lignium: Here I did not notice. Now exactly the last one. :)

args -> reader

src/main/java/org/spongepowered/api/command/parameter/managed/ValueUsage.java
```diff
@@ -0,0 +1,45 @@
...
+public interface ValueUsage {
+
+    /**
+     * Gets the usage string for the argument.
+     *
+     * @param cause The {@link Cause} requesting the usage
+     * @param key The {@link Text} that defines the parameter key
+     * @return The usage
+     */
+    Text getUsage(Cause cause, Text key);
```
Lignium: Maybe `Parameter.Key parameterKey` instead of `Text key`?

dualspiral: No, as it's passing in how the command processor will render it, which would be more useful to the consumer as the API doesn't define what the usage would look like, the implementation does. This method can then inspect what the current state is and make a decision based on that, rather than just the key.

src/main/java/org/spongepowered/api/data/DataManager.java
```diff
@@ -203,30 +161,30 @@
     <T> Optional<DataTranslator<T>> getTranslator(Class<T> objectClass);
```
Lignium: @gabizou @Cybermaxke Is it possible to create a similar method, but with TypeToken? I have problems getting a general translator for a whole class of objects.

```
// public abstract class SomeClass<T extends SomeClass<T>>
// public final class SomeClassTranslator implements DataTranslator<SomeClass<?>>

// Type mismatch: cannot convert from Optional<DataTranslator<SomeClass>> to Optional<DataTranslator<SomeClass<?>>>
Optional<DataTranslator<SomeClass<?>>> translatorBox = Sponge.getDataManager().getTranslator(SomeClass.class);
```

src/main/java/org/spongepowered/api/command/parameter/managed/ValueCompleter.java
Lignium: 1. No description of the return value is specified.
2. Maybe replace `List` to `Set`? The proposed values ​​must be unique.

Cybermaxke: Must be a `List`, because the order matters.

Lignium: There is of course `SortedSet`, but it will probably be inconvenient. :smile:

ejm:
Hi there! Please excuse my ignorance but is there a way of translating a `BlockState` to a `String` version of that state currently in the API? So if, say, a player is looking at a lever on the floor, and I have the `BlockState` of that lever, I could return `minecraft:lever[face=floor]`? If not, is this something in scope that could be added to API 8? How about the other way around, going from `String` to `BlockState` or `Item`? Thanks!

liach:
I believe that should be part of the command argument system, as that's how it is in vanilla.

ejm:
Sure, a `Parameter` could be made that deserializes a `BlockType` or `Item`
from a string argument, but, what about the other way around? That seems to
not really be territory for the command system...

src/main/java/org/spongepowered/api/command/parameter/managed/standard/CatalogedValueParameters.java
```diff
@@ -0,0 +1,432 @@
...
+     * Does not parse any arguments, but instead returns a
+     * {@link org.spongepowered.api.block.BlockSnapshot} if the
+     * current {@link CommandSource} (which thus
+     * must be a {@link org.spongepowered.api.world.Locatable}).
+     *
+     * <p>This will always fail for non-locatable sources</p>
+     *
+     * <p>Returns a {@link org.spongepowered.api.block.BlockSnapshot}.</p>
+     */
+    public static final CatalogedValueParameter<BlockSnapshot> TARGET_BLOCK =
```
Lignium: There is no parameter to specify a block simply via coordinates (`CatalogedValueParameters.BLOCK`). There are no corresponding static methods in Parameter class for `BLOCK` and `TARGET_BLOCK`.

dualspiral: > There is no parameter to specify a block simply via coordinates (CatalogedValueParameters.BLOCK).

Yes there is, `LOCATION`. A `Location` will be specified by co-ordinates [and can get that information for you](https://github.com/SpongePowered/SpongeAPI/blob/d6a82f27cd42ab2a0fd7b8f06cb78475187cbc5a/src/main/java/org/spongepowered/api/world/Location.java#L225). 

> There are no corresponding static methods in Parameter class for BLOCK and TARGET_BLOCK.

`TARGET_BLOCK` doesn't need to be the `Parameter` class - not every parameter type needs to be represented there - it's intended to be easy access and we could remove some of the helper methods yet, it does feel a bit unwieldly.

Lignium: Okay... The fact is that I need an argument that would support ray tracing in order to detect the block that the player is looking at. (https://github.com/SpongePowered/SpongeAPI/issues/1917, https://github.com/SpongePowered/SpongeAPI/issues/1935)

src/main/java/org/spongepowered/api/client/ClientWorld.java
```diff
@@ -0,0 +1,42 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.client;
+
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.world.World;
+
+import java.util.Collection;
+
+public interface ClientWorld extends World {
```
liach: 1. Shouldn't be a real world; merely a view this is
2. Need a method to get the client player entity (which is always present)

src/main/java/org/spongepowered/api/client/LocalClientServer.java
```diff
@@ -0,0 +1,35 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.client;
+
+import org.spongepowered.api.server.Server;
+
+public interface LocalClientServer extends Server {
+
+    @Override
+    default boolean isDedicatedServer() {
```
liach: Should really handle this with inheritance

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -69,6 +74,17 @@
...
+    /**
+     * Gets whether this {@link Player player} is local to the game or not. Because there
+     * is the possibility of having a game with running on the game client, there
+     * exists the possibility of having {@link ClientPlayer}s as instances, with a
+     * {@link ClientWorld} is potentially populated by {@link RemoteClientPlayer remote players}
+     * and {@link SinglePlayer the single player entity}.
+     *
+     * @return True if this player is local to the game instance
+     */
+    boolean isLocal();
```
liach: does this return true if this is the server player instance for who runs the integrated server

src/main/java/org/spongepowered/api/server/Server.java
```diff
@@ -85,7 +88,7 @@
      * @param name The name to get the player from
      * @return The {@link Player} or empty if not found
      */
-    Optional<Player> getPlayer(String name);
+    Optional<? extends Player> getPlayer(String name);
```
liach: These always return the server player. raise the lower bound.

src/main/java/org/spongepowered/api/server/Server.java
```diff
@@ -242,4 +245,15 @@
...
+    /**
+     * Gets whether this server is dedicated to being a global server, or
+     * whether this server is local to a game client where a {@link Client}
+     * instance may be available. The primary difference will be the types
+     * of {@link Player players} there may exist in the server, whether
+     * they are {@link ServerPlayer server players} or {@link RemoteClientPlayer remote players}.
+     *
+     * @return True if this is a dedicated server without a game client
+     */
+    boolean isDedicatedServer();
```
liach: The main difference is functionalities like rcon, etc. which would be better represented by a difference in inheritance.

liach (changes requested):

src/main/java/org/spongepowered/api/client/LocalClientServer.java
gabizou: Easier to do a method check on `Server` than to instanceof.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
gabizou: Since the local player isn't local to the integrated server (`LocalServer`), no. It's still a separate player in the game. This only returns true for `SinglePlayer` instances.

src/main/java/org/spongepowered/api/client/LocalClientServer.java
liach: Why do you need this check when you don't do instanceof? Give me some example usage.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
liach: Then we can just check `instanceof SinglePlayer` or `Client.getPlayer().map(p -> p == this).orElse(false)`. This doesn't have a point as you don't get the main client player from a player list or such.

src/main/java/org/spongepowered/api/server/Server.java
gabizou: Still better to have both inheritance and a simple method to find out if the server is local or not. Since everything API wise wraps around the idea of "plugins are running on the server", we still have to expose the fact that the server could be remotely controlled, i.e. in the client connecting to a game server, there's still an integrated server.

liach: sure...

liach:
Just a side note:
We should prevent plugins from holding server instances in their plugin instances as servers are not always present on clients (and each time a new instance is created rather than an old server reused)

If we add a method to check for dedicated server, we should add one to check for integrated server as well in case some party write a custom server.

gabizou:
> Just a side note:
> We should prevent plugins from holding server instances in their plugin instances as servers are not always present on clients (and each time a new instance is created rather than an old server reused)

Just like with worlds, plugins are always told not to ever store references of servers, worlds, players, entities, etc. Nothing we can do to stop them from making that mistake.

> If we add a method to check for dedicated server, we should add one to check for integrated server as well in case some party write a custom server.

Literally just means that the `isDedicatedServer()` will return `false` always for the lifetime of the `Game` instance.

src/main/java/org/spongepowered/api/client/SinglePlayer.java
```diff
@@ -0,0 +1,34 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.client;
+
+public interface SinglePlayer extends ClientPlayer {
```
Zidane: LocalPlayer, SinglePlayer is the type of game and will confuse the crap out of people.

src/main/java/org/spongepowered/api/entity/EntityPredicates.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.spongepowered.api.data.Keys;
+import org.spongepowered.api.entity.living.player.gamemode.GameModes;
+import org.spongepowered.math.vector.Vector3d;
+
+import java.util.function.Predicate;
+
+public final class EntityPredicates {
+
+    // TODO - decide whether we want this in the implementation or not.
```
Zidane, line 36: As this stands? Implementation.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
Zidane: @liach has a point. The LocalPlayer of the client is kept to the Client and not stored in any spot that would be mixed with other Players/etc. Instanceof will suffice.

src/main/java/org/spongepowered/api/server/Server.java
Zidane: @liach is correct. RemotePlayer (which is what I'd call MCP's EntityOtherPlayerMP) is purely contained to the ClientWorld. Volume-like methods on Server return Server objects.

src/main/java/org/spongepowered/api/server/Server.java
Zidane: isDedicatedServer, unlike isLocal on Player, makes sense per Gabe's rebuttal.

src/main/java/org/spongepowered/api/client/ClientWorld.java
Zidane: @liach

Removed, I forgot they do make it always available to it.

I still think it is better off in Client.

src/main/java/org/spongepowered/api/client/ClientPlayer.java
```diff
@@ -0,0 +1,35 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.client;
+
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.world.World;
+
+public interface ClientPlayer extends Player {
```
Zidane: ClientPlayer should be the interface LocalPlayer and RemotePlayer extend.

src/main/java/org/spongepowered/api/client/Client.java
```diff
@@ -26,9 +26,29 @@
...
 
+    /**
+     * Gets the {@link ClientPlayer single player} responsible
+     * for controlling this client, expectedly in singleplayer.
+     * A client may not always have a local player if browsing
+     * menus prior to joining a world or server.
+     *
+     * @return The client player
+     */
+    Optional<? extends ClientPlayer> getLocalPlayer();
```
Zidane: No reason for the wildcard. LocalPlayer is LocalPlayer. Add the interfaces.

src/main/java/org/spongepowered/api/client/Client.java
```diff
@@ -26,9 +26,29 @@
...
+
+    /**
+     *
+     * @return
+     */
+    Optional<? extends LocalClientServer> getLocalServer();
```
Zidane: See above, remove the bound. Will be a LocalServer.

src/main/java/org/spongepowered/api/client/Client.java
```diff
@@ -26,9 +26,29 @@
...
+
+    Optional<? extends ClientWorld> getViewingWorld();
```
Zidane: See above, remove the bound. Will be a ClientWorld.

src/main/java/org/spongepowered/api/client/LocalClientServer.java
```diff
@@ -0,0 +1,35 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.client;
+
+import org.spongepowered.api.server.Server;
+
+public interface LocalClientServer extends Server {
```
Zidane: LocalClientServer -> LocalServer. No reason to throw Client in it imo.

liach:
Also if possible consider adding a camera view entity to the client api

Zidane:
> 
> 
> Also if possible consider adding a camera view entity to the client api

Not at this time. The amount @gabizou has added is honestly the extent we want to go to for the time being.

src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java
```diff
@@ -0,0 +1,470 @@
...
+    /**
+     * Creates a builder that builds a {@link ValueParameter} that tries to
+     * construct a {@link Text} from an argument.
+     */
+    public static TextBuilder textBuilder() {
+        return Sponge.getRegistry().createBuilder(TextBuilder.class);
+    }
+
+    /**
+     * Creates a builder that builds a {@link ValueParameter} that tries to
```
Lignium: Javadoc outdated, this method no longer returns a builder

dualspiral: Thanks, corrected.

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Size 3.
+     */
+    public static final ContainerType FURNACE = DummyObjectProvider.createFor(ContainerType.class, "furnace");
+
+    /**
+     * Size 9 (3x3 grid). (Dispenser)
+     */
+    public static final ContainerType GENERIC_3x3 = DummyObjectProvider.createFor(ContainerType.class, "generic_3x3");
```
Lignium: The character "x" will remain lowercase here and below? This is a violation of the generally accepted style for constants. Even GitHub does not highlight these constants.

src/main/java/org/spongepowered/api/world/chunk/ChunkStates.java
```diff
@@ -0,0 +1,121 @@
...
+     * in some regards for world generation.
+     */
+    public static final ChunkState EMPTY = DummyObjectProvider.createFor(ChunkState.class, "EMPTY");
+    /**
+     * A {@link ProtoChunk} that is at this state means that it is being generated
+     * with a "base" layer of terrain, usually by a {@link Surface}(s).
+     * The chunk should not have any {@link Entity} instances or {@link BlockEntity}
+     * instances and may have a valid {@link ProtoWorld} used for world generation.
+     */
+    public static final ChunkState BASE = DummyObjectProvider.createFor(ChunkState.class, "EMPTY");
```
Lignium: Everywhere is written `"EMPTY"`, so it should be?

src/main/java/org/spongepowered/api/world/gamerule/GameRuleHolder.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param <V> The value type
+     */
+    <V> void setGameRule(GameRule<V> gameRule, V value);
+
+    /**
+     * Gets a map with all the {@link GameRule}s and their values.
+     *
+     * @return The game rules
+     */
+    Map<GameRule<?>, ?> getGameRules();
```
Lignium, line 80: I think it should be: `<V> Map<GameRule<V>, V> getGameRules()`

src/main/java/org/spongepowered/api/world/gen/carver/CarvingLayers.java
```diff
@@ -0,0 +1,41 @@
...
+package org.spongepowered.api.world.gen.carver;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public final class CarvingLayers {
+
+    // SORTFIELDS:ON
+
+    public static final CarvingLayer AIR = DummyObjectProvider.createFor(CarvingLayer.class, "AIR");
+    public static final CarvingLayer LIQUID = DummyObjectProvider.createFor(CarvingLayer.class, "AIR");
```
Lignium: `"AIR"` -> `"LIQUID"`

Lignium:
There is no `package-info.java` files in the subpackages of `org.spongepowered.api.world.gen`.

src/main/java/org/spongepowered/api/world/gamerule/GameRuleHolder.java
liach: no, that means all game rules in the map have same value type V

Lignium: Oh... Yeah sure

src/main/java/org/spongepowered/api/data/type/PistonTypes.java
```diff
@@ -0,0 +1,46 @@
...
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ * An enumeration of vanilla known {@link PistonType}s.
+ */
+public final class PistonTypes {
+
+    // SORTFIELDS:ON
+
+    public static final PistonType NORMAL = DummyObjectProvider.createFor(PistonType.class, "NORMAL");
```
dualspiral: Should be `Supplier`s!

Zidane: Oh god damnit...

PR #1996, opened 2019-05-04
Simplification of the Unified Data API (Data API v4)

gabizou:
This PR is a WIP since the description and API itself needs to be javadoc'ed after all the manipulators were already removed. The description is being worked on as well.

#### Correcting All the Missteps

Since the ~dawn of time~ creation of what we know as "Data API", there have been several iterations of this system:

- *Predecessor*: [The initial Item Data API PR](https://github.com/SpongePowered/SpongeAPI/pull/495)
- *The First Version*:  [Data API v1, unifying `DataHolder` and introducing general `DataManipulator`s](https://github.com/SpongePowered/SpongeAPI/pull/542)
- *Adding Values*: [introduction of Value API to Data API, (Data API v2)](https://github.com/SpongePowered/SpongeAPI/pull/685)
- *Refactoring Registrations*: [Data API v2.1](https://github.com/SpongePowered/SpongeAPI/pull/1531) Adding `DataRegistration` and simplifying a majority of how `DataManipulator`s are persisted and registered.
- *Overhauling Values*: [Data API v3](https://github.com/SpongePowered/SpongeAPI/pull/1843) which sought to redo various pitfals with working with `Value`s and `ImmutableValue`s, along with their creation. 

But, there were several attempts at improving the system and usage:
- [Custom DataManipulator class generation](https://github.com/SpongePowered/SpongeAPI/pull/1198)
- [Allowing usages of `Key`s and `Value`s with various builders](https://github.com/SpongePowered/SpongeAPI/pull/1103)
- [Adding **more** abstract classes for easier overriding](https://github.com/SpongePowered/SpongeAPI/pull/1393)
- [Attempt No. 2 at class generation for easier custom data](https://github.com/SpongePowered/SpongeAPI/pull/1620)

It's been a clear given that we're also experiencing several requests at improving the quality of life with *just making simple data to attach a `String` to a `Player`*. So, here's our chance, and our proposal for a newly revamped **Data API v4: The Keep It Simple Stupid Edition**

### What's the proposal?

Ever heard of `DataManipulator`? If that name sends shivers down your spine, gives you a headache at thinking about implementing one, or simply makes you want to just say "screw that, i'll just use a `Map`", you can finally rest as it's going the way of the dodo. The first thing that we've done is *eliminated all accessors of "data" through an already simple system: `Key`s. 

### Whoa whoa whoa, wait, how do you do that? What about my custom data?!

Well, hold on, let me finish explaining the proposal! Geez.... Continuing, the accessors we've already established from Data API 2 was with `ValueContainer`, if you need a refresher, you can view it [here](https://github.com/SpongePowered/SpongeAPI/blob/ee2587f45a9948c19bc7ffa1258989024f285895/src/main/java/org/spongepowered/api/data/value/ValueContainer.java#L44). In short, it gives us the following methods:
```java
public interface ValueContainer<C extends ValueContainer<C>> {
  <V extends Value<E>, E> Optional<V> getValue(Key<V> key);
  <E> Optional<E> get(Key<? extends Value<E>> key);
}
``` 
And to mutate, we've had `DataManipulator`, and `CompositeValueStore` previously available to [`offer(Key, E)` values](https://github.com/SpongePowered/SpongeAPI/blob/ee2587f45a9948c19bc7ffa1258989024f285895/src/main/java/org/spongepowered/api/data/value/mutable/CompositeValueStore.java#L191).

The nice thing about these accessors is that they've previously been backed by Sponge's implementation to where certain `Key`s would always have a value for a particular `DataHolder`, and mutating/supporting those keys was backed by an implementation I ended up calling `ValueProcessor`s and `DataProcessor`s. As you can imagine, `ValueProcessor` was designed for a single `Key` to `Value` retrieval/offering, whereas `DataProcessor` handled retrieval and offering of `DataManipulator`s. Since we've had intentions of registering `Key`s to `PluginContainer`s or `DataRegistration`s before, it's been an easy step to make to saying "well, if `DataRegistration` has it's registered `Key`s, what purpose does the `DataManipulator` serve?" And at that point, there's a few key differences:
- `DataManipulator` served as the *container* of data that was `DataSerializable` and was contractually guaranteed to be serialized/deserialized from `DataHolder`s that would support it
- `Value`s didn't know what or how their held values came from, or how to apply them to a `DataHolder`, they just knew the `Key` to value pairing.
- `Key`s had a flakey system of exposing the `DataQuery` used to access the underlying data value from any `DataView`, but because Sponge added more `Key`s than vanilla supported, it was never implemented properly to get data out of `DataView`s from things like schematics.
- `DataManipulator`s required an equal `ImmutableDataManipulator` with simlar `DataManipulatorBuilder` and `ImmutableDataManipulatorBuilder` implementations registered through `DataRegistration`, but wouldn't be able to determine if any of it's values were `Optional` except at deserialization
- Plugin authors could not take advantage of the `Key` system by proxy to their own storage solutions without doing some hacks for serialization/deserialization.

Looking over that bulletpoint list, I admitted that it was a convoluded system that had a high bar of entry for plugin developers to do something that should be made simple. So, I nuked what we know as `DataManipulator`s and their registrations, and their builders, and transformed `DataRegistration` into something *far far far cooler*: **A Registration object that plugins can register their `Key`s with appropriate serialization/providers for *persisting* data, while dynamic virtual data that is only retained as a cache requires no registration otherwise**.

### Huh? Virtual dynamic data? Providers? Caches? Wat?

Ok, yeah, that was a huge mental leap from ~nuking~ removing `DataManipulator`s to dynamic virtual data. I'll explain it this way:

If I want to just store some ~data~ values to an `Entity`, and use the handy `getValue()` and `offer()` accessors, I just register my `Key` normally, and that's it. If I `offer(Value<MyCustomObject>)` afterwards, I know my ~data~ value will remain on that `Entity` except after a restart or reload. In other words, it is non-persisting values!

HOWEVER, if I am a plugin that wants to *associate ~data~ values* to a `Player`, and store that data to the `Player`'s data file, I'd make a `Key`, like normal, and then use a `DataRegistration` to register the key, along with a `DataStore`. A `DataStore` is basically a glorified `DataBuilder` that can serialize and deserialize my custom values to and from a `DataView`. Simple enough, right? Right.

### But what about this dynamic data and providers?

Ah yes, providers and dynamic data. Let's step back a little and [look at the implementation for how we got `Value`s out of the SpongeAPI provided `Key`s](https://github.com/SpongePowered/SpongeCommon/blob/5f19dd24c9969574b90d36caa1552015a383d94a/src/main/java/org/spongepowered/common/data/processor/value/entity/FallingBlockStateValueProcessor.java#L41), seems simple enough, right? It was designed to support getting values out of any `DataHolder` from the implementation that would support it. It can be read about [here](https://github.com/SpongePowered/SpongeCommon/blob/5f19dd24c9969574b90d36caa1552015a383d94a/src/main/java/org/spongepowered/common/data/ValueProcessor.java#L75). The neat thing about this was that we knew these values would be persisted in some way, but because the plugins were not having to witness that it was being serialized, **it provided a dynamic way to say "here's a value for this key, but you don't know how it was created"**. 

Thankfully, after discussing with @dualspiral and @Cybermaxke and @Faithcaio a bit, we essentially designed the `DataProvider` to be exposed for plugin authors to be able to always provide their values, regardless whether the values are stored to the `DataHolder` or through a database. So, we had to make it a possibility for `DataProvider` to be exposed in the API.


src/main/java/org/spongepowered/api/data/DataStore.java
```diff
@@ -0,0 +1,38 @@
...
+package org.spongepowered.api.data;
+
+import com.google.common.reflect.TypeToken;
+import org.spongepowered.api.data.value.Value;
+
+public interface DataStore {
+
+    TypeToken<?> getSupportedToken();
+
+    DataView serialize(Iterable<Value<?>> values, DataView view);
```
liach: Why does this serialize/deserialize iterables instead of simple values?

liach:
Will the pr be ready for review when data manipulator is removed?

gabizou:
> Will the pr be ready for review when data manipulator is removed?

It'll be ready in probably a day or two once documentation is finished. `DataManipulator` as a whole is already nuked to orbit, just the replacement system for what's going to power getting/offering via `Key`s is needing to be drafted and reviewed.

src/main/java/org/spongepowered/api/data/DataStore.java
gabizou: That's what it's getting, an `Iterable` of `Value`s...

liach: I mean `DataView serialize(Value<?> value, DataView view);` instead of all values saved at once

gabizou: What's the point of getting only one `Value` when there's possibly multiple values being stored?

liach: Ah. Will see when you finish the basic jd (in which you will explain the saving model, etc.)

So the `Value` here actually contains both the actual data and the key, right? If that's the case, all goes very well.

JBYoshi:
How will migration of existing `DataManipulator`s to the new system work? Or does that still need to be written?

gabizou:
> How will migration of existing `DataManipulator`s to the new system work?

Since the `DataView`s are already being passed to `DataManipulatorBuilder`s, the simpler explanation is this: Pass the `DataView` of that custom data to the new `DataStore` that is registered with the `DataRegistration` object (by requirement, otherwise, failed data takes over), and because they're deserialized very much the same way, the system still gets the `Value`s back. Alternatively, as @dualspiral was mentioning yesterday, Nucleus would be doing some sort of proxy handiwork where the old data will be removed and he'll be taking advantage of the `DataProvider` that ends up providing those values externally.

ItsDoot:
In an effort to make things even more straightforward I propose also adding the following to `CompositeValueStore` (method naming I am unsure about):

`<E> DataTransactionResult add(Key<? extends CollectionValue<E>> key, E value);`
`<E> DataTransactionResult remove(Key<? extends CollectionValue<E>> key, E value);`
`<K, V> DataTransactionResult put(Key<MapValue<K, V>> key, K valueKey, E value);`
`<K> DataTransactionResult removeKey(Key<MapValue<K, ?>> key, K valueKey);`

#### Example Usecase

Currently, in API 7, it is a bit convoluted to add a potion effect to a player:
```
player.transform(Keys.POTION_EFFECTS, (effects) -> {
    ArrayList<PotionEffect> effects = new ArrayList<>(effects);
    effects.add(<potion effect>);
    return effects;
});
```
or
```
ArrayList<PotionEffect> effects = player.get(Keys.POTION_EFFECTS).orElse(new ArrayList<>());
effects.add(<potion effect>);
player.offer(Keys.POTION_EFFECTS, effects);
```

With these new methods added it would become as simple as:
```
player.add(Keys.POTION_EFFECTS, <potion effect>);
```

gabizou:
@ItsDoot there’s already transform

ItsDoot:
I understand but it shouldn't take multiple lines to just add a potion effect.
It's also undefined whether the list or map thats passed into `transform` is mutable or immutable.

kashike:
Primitive data types to avoid dealing with boxed types would be nice - `BooleanData crouching()` for example

src/main/java/org/spongepowered/api/data/DataStore.java
gabizou: What's been added is also serializing by `DataManipulator` so it's easier to "get" only the usable keys that are wanted.

ryantheleach:
> If I offer(Value<MyCustomObject>) afterwards, I know my data value will remain on that Entity except after a restart or reload. In other words, it is non-persisting values!

Only advice I'd offer here is to have explicit subtypes for the keys for Non Persisting Values.

Main usecase would be when plugins are registering them, to make sure they realize they are registering a Non Persisting value, as opposed to taking a shortcut for registering a persisting value.

Small UX thing sure, but there may be other ways of accomplishing the same goal. Just want the 2 registration methods **Crystal Clear** that not having a DataStore doesn't somehow auto magically create one for you using "Sponge Magic". e.g. default implementing a data store based off a key name and plugin name for instance, for simple value types.

**Edit:** I may have mangled some of the terms used. but I hope my point comes across.

src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -48,9 +49,33 @@ static Builder builder() {
...
+     * {@code key} is not actually registered with this {@link DataRegistration},
+     * an {@link UnregisteredKeyException} is thrown. If there is no
+     * {@code DataProvider} registered for the particular {@code Key},
+     * {@link Optional#empty()} is returned.
+     *
+     * @param key The requested key
+     * @return The provider, if there is one for the key
+     * @throws UnregisteredKeyException If the key is not registered in this registration
+     */
+    Optional<DataProvider<?, ?>> getProviderFor(Key<?> key) throws UnregisteredKeyException;
```
Cybermaxke: `<V extends Value<E>, E>Optional<DataProvider<V, E>> getProviderFor(Key<V> key) throws UnregisteredKeyException;`

src/main/java/org/spongepowered/api/data/DataStore.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+package org.spongepowered.api.data;
+
+import com.google.common.reflect.TypeToken;
+import org.spongepowered.api.data.value.MutableValueStore;
+import org.spongepowered.api.data.value.Value;
+
+public interface DataStore {
+
+    TypeToken<?> getSupportedToken();
```
Cybermaxke: `TypeToken<? extends ValueContainer>` to match `DataRegistration`

src/main/java/org/spongepowered/api/block/entity/BlockEntityArchetype.java
```diff
@@ -115,13 +130,62 @@ static Builder builder() {
...
-        <E, V extends Value<E>> Builder set(V value);
+        /**
+         * Sets the {@link DataManipulator}'s {@link Value}s onto this builder,
+         * to apply to the resulting {@link BlockEntity} when the built
+         * {@link BlockEntityArchetype} is used.
+         *
+         * @param manipulator The manipulator containing values
+         * @return This builder, for chaining
+         */
+        default Builder set(DataManipulator manipulator) {
```
Cybermaxke, line 181: `set`/`add` should be used consistently across builders, e.g. `DataBuilder.Immutable` uses `add`

src/main/java/org/spongepowered/api/data/value/MergeFunction.java
```diff
@@ -102,20 +89,20 @@ default MergeFunction andThen(final MergeFunction that) {
      * Represents a {@link MergeFunction} that ignores all merges and uses the
      * replacement, or the original if the replacement is {@code null}.
      */
-    MergeFunction IGNORE_ALL = new MergeFunction() {
+    MergeFunction RIGHT_PREFERRED = new MergeFunction() {
```
Cybermaxke: `ORIGINAL_PREFERRED` and `REPLACEMENT_PREFERRED`?

src/main/java/org/spongepowered/api/data/Key.java
```diff
@@ -0,0 +1,241 @@
...
+         * Builds the {@link Key}.
+         *
+         * @return The built key
+         * @throws IllegalStateException If not all required options were specified;
+         *                               {@link #key(CatalogKey)} and {@link #type(TypeToken)}.
+         */
+        @Override
+        Key<V> build();
+
+        interface BoundedBuilder<E, V extends Value<E>> extends Builder<E, V> {
```
Cybermaxke: Should use `BoundedValue`, I forgot to do that.

src/main/java/org/spongepowered/api/data/DataManipulator.java
```diff
@@ -0,0 +1,448 @@
...
+
+        /**
+         * Gets a {@link Immutable} with the given values.
+         *
+         * @param values The values
+         * @return The immutable data manipulator
+         */
+        static Immutable of(Iterable<? extends Value<?>> values) {
+            return Sponge.getRegistry().requireFactory(Factory.class).of(values);
+        }
```
Cybermaxke: The factory methods should be moved to `DataManipulator` and be renamed. E.g. `DataManipulator.immutableOf` instead of `DataManipulator.Immutable.of`. IMO it feels cleaner. Also applies to the mutable one.

liach: `ofImmutable` for naming.

kashike: :+1: for `immutableOf`.

gabizou:
> Only advice I'd offer here is to have explicit subtypes for the keys for Non Persisting Values.
> 
> Main usecase would be when plugins are registering them, to make sure they realize they are registering a Non Persisting value, as opposed to taking a shortcut for registering a persisting value.
> 
> Small UX thing sure, but there may be other ways of accomplishing the same goal. Just want the 2 registration methods **Crystal Clear** that not having a DataStore doesn't somehow auto magically create one for you using "Sponge Magic". e.g. default implementing a data store based off a key name and plugin name for instance, for simple value types.
> 
> **Edit:** I may have mangled some of the terms used. but I hope my point comes across.

All of this will absolutely need to be documented on SpongeDocs (I'm sure @Inscrutable will appreciate the practical deletion of half of the custom data doc pages as it's pretty much simplified to a few liners). And it will be documented on the Javadocs. Nothing else about changing types will be needed.

src/main/java/org/spongepowered/api/data/value/ValueContainer.java
```diff
@@ -65,6 +68,12 @@
      */
     <E> Optional<E> get(Key<? extends Value<E>> key);
 
+    OptionalInt getInteger(Key<? extends Value<Integer>> key);
```
Cybermaxke, line 71: `getInt` (to be consistent with `getIntProperty`)

src/main/java/org/spongepowered/api/data/DataHolder.java
```diff
@@ -24,46 +24,310 @@
  */
 package org.spongepowered.api.data;
 
-import org.spongepowered.api.data.manipulator.DataManipulator;
+import org.checkerframework.checker.nullness.qual.NonNull;
```
liach, line 27: Just curious, where did you get this checker framework annotation from? I fear my change in #2001 might clash with yours.

kashike: Transient dependency from configurate, most likely.

PR #1998, opened 2019-05-05
Inventory for API8 (part 2)

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2394) | [SpongeCommon (old)](https://github.com/SpongePowered/SpongeCommon/pull/2211) 

based on changes in: https://github.com/SpongePowered/SpongeAPI/pull/1959
more changes in: https://github.com/SpongePowered/SpongeAPI/pull/2050

#### SlotIndex -> int (already merged to API-8) ####
* Replaced `SlotIndex` with Integers
* renamed `poll`/`peek` with index -> `pollFrom`/`peekAt`

#### Changes to Inventory operations ####
 - `InventoryTransactionResult` is used for `set`/`offer`/`poll` including uses with index/pos  
 - added `InventoryTransactionResult#getPolledItems` for poll operations
 - operations on absent slot index/pos return `Type.NO_SLOT` result

#### Other API Changes ####
* removed `peek` with limit. (instead use query and totalQuantity)
* set(stack) moved from Inventory to Slot (set on multiple slots was not clearly defined)
* `size` replaced by `freeCapacity` (inverted meaning: size was non-empty slots)
* `totalItems` renamed to `totalQuantity` (analog to `getQuantity` on `ItemStack`)
* `getMaxStackSize` removed and added back as a `InventoryProperty`
* `setMaxStackSize` removed (not implementable for most Inventory)
* `getPlugin` removed and added back as a `InventoryProperty`
* `ViewableInventory` gets `getViewers`, hasViewers`
* `Container` gets `setCursor` and `getCursor`

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
@@ -35,37 +35,61 @@ public final class ContainerTypes {
 
     // Containers backed by an inventory.
 
+    // TODO check container creation in 1.14 code especially merchant/horse
+
+
+    // TODO add new inventories
+
     // SORTFIELDS:ON
+
+    public static final ContainerType BLAST_FURNACE = DummyObjectProvider.createFor(ContainerType.class, "blast_furnace");
+
+    /**
+     * Size 5.
+     */
+    public static final ContainerType BREWING_STAND = DummyObjectProvider.createFor(ContainerType.class, "brewing_stand");
+
+    /**
+     * Size 3.
+     */
+    public static final ContainerType FURNACE = DummyObjectProvider.createFor(ContainerType.class, "furnace");
+
+    /**
+     * Size 9 (3x3 grid). (Dispenser)
+     */
+    public static final ContainerType GENERIC_3x3 = DummyObjectProvider.createFor(ContainerType.class, "generic_3x3");
+
+    public static final ContainerType GENERIC_9x1 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x1");
+
+    public static final ContainerType GENERIC_9x2 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x2");
 
     /**
      * Size 27 (3x9 grid)
      */
-    public static final ContainerType CHEST_3X9 = DummyObjectProvider.createFor(ContainerType.class, "chest_3x9"); // "minecraft:chest"
+    public static final ContainerType GENERIC_9x3 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x3");
+
+    public static final ContainerType GENERIC_9x4 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x4");
+
+    public static final ContainerType GENERIC_9x5 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x5");
+
     /**
      * Size 54 (6x9 grid)
      */
-    public static final ContainerType CHEST_6x9 = DummyObjectProvider.createFor(ContainerType.class, "chest_6x9"); // "minecraft:chest"
-    /**
-     * Size 3.
-     */
-    public static final ContainerType FURNACE = DummyObjectProvider.createFor(ContainerType.class, "furnace"); // "minecraft:furnace"
-    /**
-     * Size 9 (3x3 grid).
-     */
-    public static final ContainerType DISPENSER = DummyObjectProvider.createFor(ContainerType.class, "dispenser"); // "minecraft:dispenser"
-    /**
-     * Size 5.
-     */
-    public static final ContainerType BREWING_STAND = DummyObjectProvider.createFor(ContainerType.class, "brewing_stand"); // "minecraft:brewing_stand"
+    public static final ContainerType GENERIC_9x6 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x6");
+
     /**
      * Size 5 (1x5 grid).
      */
-    public static final ContainerType HOPPER = DummyObjectProvider.createFor(ContainerType.class, "hopper"); // "minecraft:hopper"
+    public static final ContainerType HOPPER = DummyObjectProvider.createFor(ContainerType.class, "hopper");
+
+    public static final ContainerType LECTERN = DummyObjectProvider.createFor(ContainerType.class, "lectern");
 
     /**
      * Size 27 (3x9 grid). Shulker boxes are not allowed in shulker boxes.
      */
-    public static final ContainerType SHULKER_BOX = DummyObjectProvider.createFor(ContainerType.class, "shulker_box"); // "minecraft:shulker_box"
+    public static final ContainerType SHULKER_BOX = DummyObjectProvider.createFor(ContainerType.class, "shulker_box");
+
+    public static final ContainerType SMOKER = DummyObjectProvider.createFor(ContainerType.class, "smoker");
 
     // SORTFIELDS:OFF
 
@@ -74,22 +98,32 @@ public final class ContainerTypes {
     // SORTFIELDS:ON
 
     /**
-     * Size 0. All slots present in the container only: 10 (3x3+1).
-     */
-    public static final ContainerType CRAFTING_TABLE = DummyObjectProvider.createFor(ContainerType.class, "crafting_table"); // "minecraft:crafting_table"
-    /**
-     * Size 0. All slots present in the container only 2.
-     */
-    public static final ContainerType ENCHANTING_TABLE = DummyObjectProvider.createFor(ContainerType.class, "enchanting_table"); // "minecraft:enchanting_table"
-    /**
      * Size 0. All slots present in the container only: 3.
      */
-    public static final ContainerType ANVIL = DummyObjectProvider.createFor(ContainerType.class, "anvil"); // "minecraft:anvil"
+    public static final ContainerType ANVIL = DummyObjectProvider.createFor(ContainerType.class, "anvil");
 
     /**
      * Size 0. All slots present in the container only: 1.
      */
-    public static final ContainerType BEACON = DummyObjectProvider.createFor(ContainerType.class, "beacon"); // "minecraft:beacon"
+    public static final ContainerType BEACON = DummyObjectProvider.createFor(ContainerType.class, "beacon");
+
+    public static final ContainerType CARTOGRAPHY = DummyObjectProvider.createFor(ContainerType.class, "cartography");
+
+    /**
+     * Size 0. All slots present in the container only: 10 (3x3+1).
+     */
+    public static final ContainerType CRAFTING = DummyObjectProvider.createFor(ContainerType.class, "crafting");
+
+    /**
+     * Size 0. All slots present in the container only 2.
+     */
+    public static final ContainerType ENCHANTMENT = DummyObjectProvider.createFor(ContainerType.class, "enchantment");
+
+    public static final ContainerType GRINDSTONE = DummyObjectProvider.createFor(ContainerType.class, "grindstone");
+
+    public static final ContainerType LOOM = DummyObjectProvider.createFor(ContainerType.class, "loom");
+
+    public static final ContainerType STONECUTTER = DummyObjectProvider.createFor(ContainerType.class, "stonecutter");
 
     // SORTFIELDS:OFF
 
@@ -100,12 +134,12 @@ public final class ContainerTypes {
     /**
      * Create a subtype of a {@link Horse} Entity instead of using this ContainerType.
      */
-    public static final ContainerType HORSE = DummyObjectProvider.createFor(ContainerType.class, "horse"); // "minecraft:horse" internally "EntityHorse"
+    public static final ContainerType HORSE = DummyObjectProvider.createFor(ContainerType.class, "horse");
 
     /**
      * Create a {@link Villager} Entity instead of using this ContainerType.
      */
-    public static final ContainerType VILLAGER = DummyObjectProvider.createFor(ContainerType.class, "villager"); // "minecraft:villager"
+    public static final ContainerType MERCHANT = DummyObjectProvider.createFor(ContainerType.class, "merchant");
 
     // SORTFIELDS:OFF
 
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -350,13 +350,6 @@ public interface Inventory extends Nameable.Translatable, PropertyHolder {
     <T extends Inventory> Optional<T> query(Class<T> inventoryType);
 
     /**
-     * Returns the {@link PluginContainer} who built this inventory.
-     *
-     * @return The container
-     */
-    PluginContainer getPlugin();
-
-    /**
      * Intersects the slots of both inventories.
      * The resulting inventory will only contain slots
      * that are present in both inventories.
@@ -414,6 +407,15 @@ public interface Inventory extends Nameable.Translatable, PropertyHolder {
     default Inventory transform(InventoryTransformation transformation) {
         return transformation.transform(this);
     }
+
+    /**
+     * Returns true if the given inventory is a direct child of this one.
+     *
+     * @param child the child inventory to check for.
+     *
+     * @return whether the given inventory is a direct child of this one.
+     */
+    boolean containsChild(Inventory child);
 
     /**
      * Returns this inventory as a viewable inventory if possible.
--- a/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/Inventory.java
@@ -28,6 +28,7 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.data.property.Property;
 import org.spongepowered.api.data.property.PropertyHolder;
 import org.spongepowered.api.data.property.PropertyMatcher;
+import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.item.inventory.query.QueryOperation;
 import org.spongepowered.api.item.inventory.query.QueryOperationTypes;
@@ -49,7 +50,7 @@ public interface Inventory extends Nameable.Translatable, PropertyHolder {
     /**
      * Creates a new {@link Inventory.Builder} to build a basic {@link Inventory}.
      * <p>Inventories created by this builder cannot be opened.</p>
-     * <p>If you want to show the inventory to a Client use {@link ViewableInventory#builder()}</p>
+     * <p>If you want to show the inventory to a {@link Player} use {@link ViewableInventory#builder()}</p>
      *
      * @return The builder
      */
@@ -60,8 +61,7 @@ public interface Inventory extends Nameable.Translatable, PropertyHolder {
     /**
      * Gets the parent {@link Inventory} of this {@link Inventory}.
      *
-     * @return the parent inventory, returns this inventory if there is no
-     *      parent (this is a top-level inventory)
+     * @return the parent inventory, returns this inventory if there is no parent (this is a top-level inventory)
      */
     Inventory parent();
 
@@ -69,8 +69,7 @@ public interface Inventory extends Nameable.Translatable, PropertyHolder {
      * Gets the root {@link Inventory} of this {@link Inventory}.
      * This is equivalent to calling {@link #parent()} until it returns itself.
      *
-     * @return the root inventory, returns this inventory if there is no
-     *       parent (this is a top-level inventory)
+     * @return the root inventory, returns this inventory if there is no parent (this is a top-level inventory)
      */
     Inventory root();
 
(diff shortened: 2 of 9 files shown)


src/main/java/org/spongepowered/api/item/inventory/custom/ContainerTypes.java
```diff
@@ -37,26 +37,31 @@
 
     // SORTFIELDS:ON
```
Cybermaxke, line 44: `ContainerType`(s) shouldn't be in `custom` IMO. Since they are also used for vanilla.
This catalog should also be updated to 1.14: https://gist.github.com/Cybermaxke/c91032ba8571bddfaf3048250cbe6d10

Cybermaxke:
Why remove `containsInventory`? How can you otherwise know if the inventory contains a child? E.g. a container contains the player main grid. Possibly rename it to `containsChild`?

Faithcaio:
I guess i'll leave it in. But the way its implemented now its not always correct.

src/main/java/org/spongepowered/api/item/inventory/custom/ContainerTypes.java
Zidane: I agree with @Cybermaxke on this.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -364,70 +298,83 @@
...
-     * least an amount use {@link #contains(ItemStack)}.
+     * least a given quantity use {@link #contains(ItemStack)}.
      *
      * @param stack The stack to check for
+     *
      * @return True if the stack is present in this inventory
      */
     boolean containsAny(ItemStack stack);
 
+    // TODO remove from API? do we need to get a property relative to another parent in API?
```
Zidane, line 293: Agreed. if they want the parent? Call getParent().getProperty().

Should be only one of those.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -364,70 +298,83 @@
...
+     *
+     * @return The queried inventory
      */
-    Inventory query(QueryOperation<?>... operations);
+    Inventory query(Query query);
 
     /**
-     * Query this inventory for inventories matching
-     * the supplied {@link PropertyMatcher}.
+     * Query this inventory with given {@link QueryType} and {@code param}.
```
Zidane: Unsure about the OneParam, TwoParam system. jOOQ Does something similar in how they handle DatabaseRows and their types. This is inherently inflexible because you have to keep littering this interface with additional param methods and interfaces.

Just my thoughts.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -45,6 +45,10 @@
 
     // SORTFIELDS:ON
 
+    // TODO new properties
+    public static final Property<PluginContainer> CREATOR = null;
```
Zidane: Should be called PLUGIN

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -472,10 +413,12 @@ default Inventory query(PropertyMatcher<?> propertyMatcher) {
      * in the second one is removed.</p>
      *
      * @param inventory the other inventory
+     *
      * @return an inventory wrapping all slots of both inventories.
      */
     Inventory union(Inventory inventory);
```
Zidane, line 380: As cyber mentioned, we still need this. containsChild would be a better name.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -53,20 +57,23 @@
 
     /**
      * Represents the capacity of a inventory.
+     * TODO not present in impl | do we need this for query? {@link Inventory#capacity()} exists
```
Zidane: No, you have the Inventory object you can figure this out straight away with getting a property.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -53,20 +57,23 @@
...
+     *
      */
     public static final Property<Integer> CAPACITY =
             DummyObjectProvider.createFor(Property.class, "CAPACITY");
 
     /**
      * Represents the dimension of a inventory.
+     * TODO not present in impl | do we need this for query? {@link GridInventory#getRows()} and {@link GridInventory#getColumns()} exists
```
Zidane: If all Inventories effectively have a dimension, even if 1-dimensional, then I would say we still need this as someone may not care about if it is a grid and only cares about the particular dimensions of the inventory.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -78,22 +85,22 @@
...
     public static final Property<UUID> UNIQUE_ID =
             DummyObjectProvider.createFor(Property.class, "UNIQUE_ID");
 
     /**
-     * Represents the title of a inventory, is
-     * viewable by players looking at the inventory.
+     * Represents the title of a inventory, is viewable by players looking at the inventory.
      */
+    // TODO maybe not present in impl? most inventories dont have a title. also title only shows in Containers
     public static final Property<Text> TITLE =
```
Zidane: Getting a properties' value returns an Optional, yes? If so, this is fine to keep. Maybe document that all of these properties are not guaranteed to exist on all Inventories.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -78,22 +85,22 @@
...
-     * for use in querying "sided inventories".
+     * A property which represents a "side" for a particular slot, for use in querying "sided inventories".
      */
     public static final Property<Direction> SLOT_SIDE =
             DummyObjectProvider.createFor(Property.class, "SLOT_SIDE");
 
     /**
      * Represents the {@link UUID} of a inventory.
      */
+    // TODO not present in impl for query atm.
```
Zidane: I want to see this one happen if possible.

src/main/java/org/spongepowered/api/item/inventory/Slot.java
```diff
@@ -35,4 +37,18 @@
      * @return the viewed inventory slot or itself if not a container slot
      */
     Slot viewedSlot();
+
+    /**
+     * Adds the ItemStack to this slot overwriting the existing item.
```
Zidane, line 42: Per your docs, this is really a consume method and not just a set if your ItemStack object is not being used by the Slot.

src/main/java/org/spongepowered/api/item/inventory/custom/ContainerTypes.java
```diff
@@ -37,26 +37,31 @@
...
 
+    /**
+     * Size 5.
+     */
+    public static final ContainerType BREWING_STAND = DummyObjectProvider.createFor(ContainerType.class, "brewing_stand"); // "minecraft:brewing_stand"
```
Zidane: Remove the id comments before we merge this please.

src/main/java/org/spongepowered/api/item/inventory/query/Query.java
```diff
@@ -24,26 +24,37 @@
...
+     *
+     * @param queries The queries to append
      *
-     * @param arg The object to test the inventory against
-     * @return A query that may be passed to {@link
-     * Inventory#query(QueryOperation...)}
+     * @return The new query
      */
-    QueryOperation<T> of(T arg);
+    Query append(Query... queries);
```
Zidane: This should be and(..)

src/main/java/org/spongepowered/api/item/inventory/query/Query.java
```diff
@@ -24,26 +24,37 @@
...
 /**
- * Represents a possible type of operation for an {@linkplain
- * Inventory#query(QueryOperation...) inventory query}.
- *
- * @param <T> The argument type for the query
+ * An inventory query. See {@link QueryTypes} for possible types of queries.
  */
-@CatalogedBy(QueryOperationTypes.class)
-public interface QueryOperationType<T> extends CatalogType {
+public interface Query {
```
Zidane: This design to me is odd only because it has composite methods.

Generally you build a composite from a set of queries and have an adapter execute them. Here the query builds a composite in on itself (?) and then executes itself (??)

I think Inventory should get the execute method as you are basically asking for a view of it based on a provided query. I would then create utility methods for common composites (and, not, or) and pass Query objects to them.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: Thats not the same. Properties of a child are relative to its parent.
The way queries work there can be multiple parents returning the "same" (in terms of the same lens not instance) child.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: Alternative Ideas?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: the current impl. (when it works) checks not only for childs but recursivly

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
Faithcaio: Dimension is for grids. Currently used to create grids for custom inventories, which is done differently with the new builders

src/main/java/org/spongepowered/api/item/inventory/query/Query.java
Faithcaio: The logic for what to do is in the `Query` impl. It cannot be in the Inventory/Adapter.
`Inventory` has `query(QueryType, param(s))` and `query(Query)`  which (creates) + executes the query with itself as input.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Zidane: Right but again there is no reason to call a method to get one from the parent. You could just get the Parent inventory and call it instead no?

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
Zidane: * without

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Zidane: It scans recursively? Is that how get works? If not, I don't see a problem with containsChild and not analyzing the entire trunk.

src/main/java/org/spongepowered/api/item/inventory/query/Query.java
Zidane: Which is fine, in the implementation you could keep the logic in Query. I just would rather not see that at the API level.

From the API's perspective, Query should be kept dumb. There should be methods to build composite queries. Inventory should have the execute method that takes the Query. In the impl, if you want to have Query do the logic and basically forward the Inventory call to it then knock yourself out. At least the API would remain clean in that regard.

Faithcaio:
.

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
```diff
@@ -138,6 +139,13 @@ static InventoryTransactionResult failNoTransactions() {
      */
     List<ItemStackSnapshot> getRejectedItems();
 
+    /**
+     * If items were polled by the operation this will return the polled item.
+     *
+     * @return the item that was polled from the inventory.
+     */
+    Optional<ItemStackSnapshot> getPolledItem();
```
Cybermaxke: I prefer subclasses of `InventoryTransactionResult` instead of putting everything here.

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
```diff
@@ -35,28 +35,35 @@
...
 
+    /**
+     * Size 5.
+     */
+    public static final ContainerType BREWING_STAND = DummyObjectProvider.createFor(ContainerType.class, "brewing_stand"); // "minecraft:brewing_stand"
+
     /**
      * Size 27 (3x9 grid)
      */
     public static final ContainerType CHEST_3X9 = DummyObjectProvider.createFor(ContainerType.class, "chest_3x9"); // "minecraft:chest"
```
Cybermaxke: Becomes `GENERIC_3X9`, `GENERIC_6X9` for the one below

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -145,9 +145,7 @@
      * Adds an ItemStack to this inventory.
      * Returns a {@link InventoryTransactionResult.Type#SUCCESS} only if the entire {@link ItemStack} fits the inventory.
      *
-     * <p>The size of the supplied stack is reduced by the number of items successfully consumed by the inventory.</p>
```
Cybermaxke: Why remove this? IMO it was very convenient to use it like this. E.g. I can just offer a itemstack to a list of inventories and it will be added until it's consumed.

For the `set` method it makes sense to remove it.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
```diff
@@ -45,50 +45,47 @@
...
+    public static final Property<UUID> UNIQUE_ID = DummyObjectProvider.createFor(Property.class, "UNIQUE_ID");
 
     /**
-     * Represents the title of a inventory, is viewable by players looking at the inventory.
+     * Represents the title of a inventory, is viewable by players looking at the inventory. Not all inventories have a title.
      */
-    // TODO maybe not present in impl? most inventories dont have a title. also title only shows in Containers
-    public static final Property<Text> TITLE =
-            DummyObjectProvider.createFor(Property.class, "TITLE");
+    public static final Property<Text> TITLE = DummyObjectProvider.createFor(Property.class, "TITLE");
```
Cybermaxke, line 88: Originally for the capacity and dimension property, my thought that it would be useful to find subinventories in built inventories. But maybe it would be better to "tag" inventories instead and use that tag to query for them. Or this could be the unique id, if it's used uniquely per inventory (and children) and not globally.

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: You could also do this by offering to the union of your list of inventories.

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
Faithcaio: Where do you get the names from?

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
Faithcaio: And to answer why because @Zidane 
fight!

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
Cybermaxke: It's the "menu" registry: https://gist.github.com/Cybermaxke/c91032ba8571bddfaf3048250cbe6d10

Faithcaio: in code `net.minecraft.inventory.container.ContainerType`

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
Faithcaio: That would break the ability to combine multiple InventoryTransactionResults (`InventoryTransactionResult#and`)

On that note it has also to be `List<ItemStackSnapshot> getPolledItems()`

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
```diff
@@ -35,37 +35,61 @@
...
+     * Size 5.
      */
-    public static final ContainerType CHEST_3X9 = DummyObjectProvider.createFor(ContainerType.class, "chest_3x9"); // "minecraft:chest"
+    public static final ContainerType BREWING_STAND = DummyObjectProvider.createFor(ContainerType.class, "brewing_stand");
+
     /**
      * Size 54 (6x9 grid)
      */
-    public static final ContainerType CHEST_6x9 = DummyObjectProvider.createFor(ContainerType.class, "chest_6x9"); // "minecraft:chest"
+    public static final ContainerType CHEST_9x6 = DummyObjectProvider.createFor(ContainerType.class, "generic_9x6");
```
Cybermaxke: `GENERIC_9x6`

src/main/java/org/spongepowered/api/item/inventory/ContainerTypes.java
```diff
@@ -35,37 +35,61 @@
...
+
     /**
-     * Size 3.
+     * Size 9 (3x3 grid).
      */
-    public static final ContainerType FURNACE = DummyObjectProvider.createFor(ContainerType.class, "furnace"); // "minecraft:furnace"
+    public static final ContainerType DISPENSER = DummyObjectProvider.createFor(ContainerType.class, "dispenser");
```
Cybermaxke: `GENERIC_3x3`

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
Cybermaxke: Ok, you can still use subclasses in this case, because the moment you start merging things `getPolledItem` is meaningless. Which of the polled items is it, the first one, or the last one. So the subclass provides the `getPolledItem`, and the merged transactions (or other operations) use the standard one.

Another thing, I can see it becoming quite annoying that you have to call `getPolledItem` after calling `poll`. Most of the time users don't even care about the transactions that happened, just about the returned item.

Cybermaxke:
`containsInventory` can be removed, since it was replaced by `containsChild`?

Faithcaio:
no `containsInventory` still checks recursivly `containsChild` only one layer down

src/main/java/org/spongepowered/api/item/inventory/type/Inventory2D.java
```diff
@@ -72,23 +73,10 @@
...
-    /**
-     * Gets without removing the stack at the supplied position in this
-     * Inventory.
-     *
-     * @see Inventory#peek()
-     * @param pos Slot position to query
-     * @param limit item limit
-     * @return matching stacks, as per the semantics of {@link Inventory#peek()}
-     */
-    default Optional<ItemStack> peek(Vector2i pos, int limit) {
```
Cybermaxke, line 84: Why remove this?

Faithcaio: see above: 
> removed peek with limit. (instead use query and totalQuantity)

src/main/java/org/spongepowered/api/item/inventory/transaction/InventoryTransactionResult.java
Faithcaio: can't have everything. but the subclass is here

src/main/java/org/spongepowered/api/item/inventory/type/Inventory2D.java
Cybermaxke: Then why keep `poll` with limit, if it can also be done with query and total quantity. How do you use "query and totalQuantity"?

Faithcaio: It works for replacing "peek" because its readonly.
```
Inventory queryResult = inventory.query(ITEM_STACK_IGNORE_QUANTITY.of(stack));
int quantityOfStack = queryResult.totalQuantity();
```
But it won't work for polling a specific quantity as it also modifies the inventory.
Including only removing partial stacks.
```
Inventory queryResult = inventory.query(ITEM_STACK_IGNORE_QUANTITY.of(stack));
queryResult.poll(32);
```

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
JBYoshi: See https://github.com/SpongePowered/SpongeCommon/issues/2378#issuecomment-536341811. Each slot in a chest has three `SlotIndex` properties: one for the index within the grid, one for the index within the row, and one for the index within the column. Each is accesed by specifying a different parent inventory: the grid, the row, or the column.

JBYoshi: Although personally, I think `SlotIndex` (and other properties that depend on a parent, such as `SlotPos`) ought to be methods instead of properties - that way you have to specify a parent. For example, `SlotIndex` could be replaced by `Inventory.getSlot(int index)` and `Inventory.indexOf(Inventory inv)`.

src/main/java/org/spongepowered/api/item/inventory/InventoryProperties.java
Cybermaxke: Keep in mind that properties also serve a purpose for queries, so even if this may be duplicate for getting the capacity, it's still useful to use for querying for inventories with a specific capacity.

The same applies to the other properties.

PR #1999, opened 2019-05-05
Data improvements

Cybermaxke:
@gabizou 

I replaced the `Simple` value containers with the recycled `DataManipulator`s. I split it up into 3 interfaces `DataManipulator`, `MutableDataManipulator` and `ImmutableDataManipulator`. Making the mutable and immutable ones inner classes is still an option to be consistent with `Value`s.

I also cleaned up the `Value` constructed and added `mutableOf` and `immutableOf` factory methods. Due the nature of generics, the appropriate `asMutable` and `asImmutable` method need to be called if you want to cast.

Bounded value keys now provide default min/max values and comparator at construction. So they can use the same factory methods.

And some other misc stuff.


gabizou:
Manually merged in these changes into a single commit. Thanks for the improvements!

PR #2001, opened 2019-05-09
Migrate from jsr305 to checker framework

liach:

Fixes https://github.com/SpongePowered/SpongeAPI/pull/1995#issuecomment-490771299

Some immutable and threadsafe annotations are dumped; they should be replaced by proper "nolock" "locking" annotations instead.


liach:
@Xakep_SDK Requesting a review from you as well.

XakepSDK:
Looks good for me.
I request removing jsr305 from here too: https://github.com/SpongePowered/plugin-meta
p.s. i never used checkerframework

kashike (approved):

kashike:
Big :+1: from me.

ST-DDT:
Mind sharing the reason for this change?
What are the benefits from using the checkerframework vs jsr305? Is it because jsr305 is dormant?

XakepSDK:
1. JSR305 never was standard
2. Doesn't support java 8 features, e.g. you can't do `List<@Nullable String> s`
3. Has compatibility issues with java modules
JSR305 works with modules, but you need to know how to deal with split packages, while checkerframework just works

liach:
Reasons for @ST-DDT:
1. jsr305 breaks with Java 9+ module system;
1. checker framework has type use annotations, e.g. `Supplier<@NonNull T>` cannot supply `null` while `Predicate<@Nullable T>` can test `null`.
1. checker framework is more complete: it has annotations for calls to methods in constructors, locking conditions, optional presence, taint/no taint, etc., compared to jetbrains annotations or jsr305.
1. if we want to verify the validity, we can add checker framework's non-qual part to our test dependencies so that we can use checker itself to test the validity of our annotations.

dualspiral:
I don't see a specific issue with this in general.

I'm not familiar with checkerframework so I'll ask, is there a way to use meta-annotations on `NonnullByDefault` to alias `@org.checkerframework.framework.qual.DefaultQualifier(org.checkerframework.checker.nullness.qual.NonNull.class)` like was done with `@NonNull` such the checkerframework can still work with it? I know we're breaking things left right and centre for 8 **but** keeping this annotation, if possible, doesn't really cost anything, is arguably more readable and it does make the diff a lot smaller making it easier to port any changes we _might_ have to make on top. It would also help if we have to make a wholesale change across all packages in the future for whatever reason and we can just change the meta-annotation on our annotation.

If not, sure, no problem, but I figured I'd ask.

liach:
Nah. I don't think checker framework allows annotations to be an alias for other annotations, hence even if we keep `@NonnullByDefault`, it would no longer be effective.

liach:
On a side note, we might consider adding a qualifier annotation set for server thread access, similar to that for gui access check. Opinions? https://checkerframework.org/manual/#guieffect-checker

stephan-gh (approved):

Zidane:
Manually merged into api-8. Thanks @liach !

liach:
thanks zidane!

PR #2002, opened 2019-05-11
Play with builders for goal and such

liach:
Finished the first stage. Need some comments for the best design possible.

### DIFF
--- a/src/main/java/org/spongepowered/api/ai/goal/GoalSelector.java
+++ b/src/main/java/org/spongepowered/api/ai/goal/GoalSelector.java
@@ -30,6 +30,7 @@ import org.spongepowered.api.entity.living.monster.Monster;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * Represents a set of tasks that will be updated together by an {@link Agent}.
@@ -58,24 +59,24 @@ public interface GoalSelector<O extends Agent> {
     O getOwner();
 
     /**
-     * Adds a new {@link org.spongepowered.api.ai.goal.Goal} to this {@link GoalSelector}.
+     * Adds a new {@link Goal} to this {@link GoalSelector}.
      *
      * @param priority The priority this goal should run at
      * @param task The goal to run
      * @return This goal, for chaining
      */
-    GoalSelector addTask(int priority, org.spongepowered.api.ai.goal.Goal<O> task);
+    GoalSelector<O> addTask(int priority, Goal<O> task);
 
     /**
-     * Removes a specific {@link org.spongepowered.api.ai.goal.Goal} from this {@link GoalSelector}.
+     * Removes a specific {@link Goal} from this {@link GoalSelector}.
      *
      * @param task The goal to remove
      * @return An optional containing the goal entry removed or {@link Optional#empty()} if none found
      */
-    Optional<GoalEntry<O>> removeTask(org.spongepowered.api.ai.goal.Goal<O> task);
+    Optional<GoalEntry<O>> removeTask(Goal<O> task);
 
     /**
-     * Removes all {@link org.spongepowered.api.ai.goal.Goal}s whose {@link GoalType} matches
+     * Removes all {@link Goal}s whose {@link GoalType} matches
      * the provided type in this {@link GoalSelector}.
      *
      * @param type The type to remove
@@ -84,7 +85,7 @@ public interface GoalSelector<O extends Agent> {
     Collection<GoalEntry<O>> removeTasks(GoalType type);
 
     /**
-     * Gets all {@link org.spongepowered.api.ai.goal.Goal}s whose {@link GoalType} matches
+     * Gets all {@link Goal}s whose {@link GoalType} matches
      * the provided type in this {@link GoalSelector}.
      *
      * @param type The type to look for
@@ -93,20 +94,24 @@ public interface GoalSelector<O extends Agent> {
     List<GoalEntry<O>> getTasksByType(GoalType type);
 
     /**
-     * Gets all {@link org.spongepowered.api.ai.goal.Goal}s in this {@link GoalSelector}.
+     * Gets all {@link Goal}s in this {@link GoalSelector}.
      *
      * @return The tasks
      */
     List<GoalEntry<O>> getTasks();
 
     /**
-     * Clears all {@link org.spongepowered.api.ai.goal.Goal}s from this goal.
+     * Clears all {@link Goal}s from this goal.
      */
     void clear();
 
-    interface GoalEntry <O extends Agent> extends Comparable<GoalEntry<O>> {
+    default Set<GoalCategoryFlag> getAvailableFlags() {
+        return getOwner().getAvailableGoalCategoryFlags();
+    }
 
-        GoalSelector getSelector();
+    interface GoalEntry<O extends Agent> extends Comparable<GoalEntry<O>> {
+
+        GoalSelector<O> getSelector();
 
         Goal<O> getGoal();
 
--- a/src/main/java/org/spongepowered/api/ai/goal/agent/LookIdleGoal.java
+++ b/src/main/java/org/spongepowered/api/ai/goal/agent/LookIdleGoal.java
@@ -29,18 +29,18 @@ import org.spongepowered.api.ai.goal.Goal;
 import org.spongepowered.api.ai.goal.GoalBuilder;
 import org.spongepowered.api.entity.living.Agent;
 
-public interface LookIdleGoal extends Goal<Agent> {
+public interface LookIdleGoal<O extends Agent> extends Goal<O> {
 
     /**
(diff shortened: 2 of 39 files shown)


ST-DDT:
IMO adding the agent type to the goal might be helpful, but adding the other parameters to the goal builder severally  reduces the code readability/usability when using the builders.

````java
interface Builder<O extends Agent, A extends LookIdleGoal<O>, B extends Builder<O, A, B>> extends GoalBuilder<O, A, B>
````
I might be wrong but the user would then have to use the following builder definition:

````java
LookIdleGoal.Builder<Zombie, LookIdleGoal<Zombie>, ???>
````
Is it even possible without generic type parameters or raw types?
Try storing it in a variable.

It should be possible to simplify it like this:

````java
interface Builder<O extends Agent> extends GoalBuilder<O, LookIdleGoal<O>, Builder<O>>
````
without reducing the extensibility much.
It is very unlikely that these interfaces are extended anyways, so a usage simplification would definitely be worth it.

The same also applies to some none goal builders.

**Comment by liach** [comment]
Nah. @ST-DDT Look at this for the sample usage of these new generic builders. https://github.com/SpongePowered/SpongeAPI/pull/2002/files#diff-e7a8c43871ed676925c231f14465bf8a

You declare a generic variable like this in a private method:
```java
    <B extends WatchClosestGoal.Builder<Fox, WatchClosestGoal<Fox>, B>> void handle(B builder) {
        // do stuff with your builder!
    }
```

When we update to java 10, we can also use `var` to have this type inferred correctly as well.

ST-DDT:
Is java 10 support guaranteed to be available with the new release?
IIRC var is not usable for fields though.
And if you create templates, you definitely need to store it somewhere.
Lets assume you want to spawn three/n monsters each with the same goals atbdifferent. You'll want to configure each goal builder only once and then just invoke build on them three/n times.

Please make the API simple instead of fancy.

src/main/java/org/spongepowered/api/ai/goal/GoalSelector.java
```diff
@@ -58,24 +59,24 @@
...
     /**
-     * Adds a new {@link org.spongepowered.api.ai.goal.Goal} to this {@link GoalSelector}.
+     * Adds a new {@link Goal} to this {@link GoalSelector}.
      *
      * @param priority The priority this goal should run at
      * @param task The goal to run
      * @return This goal, for chaining
      */
-    GoalSelector addTask(int priority, org.spongepowered.api.ai.goal.Goal<O> task);
+    GoalSelector<O> addTask(int priority, Goal<O> task);
```
ST-DDT, line 68: Maybe use `Goal<? super O>`. That way you could reuse a goal builder for different monster types e.g. zombies and skeletons.

liach: Nah. Now the `O` is the exact type, no longer fuzzy like before.

liach:
Having a concrete builder and an abstract builder is definitely not fancy and harms interfacing for modded goals as well.

liach:
If things do go bad, just use raw types. We always have that backing us up. Moreover, builders aren't meant to be kept for long; they are intended to be consumed and disposed quickly.

ST-DDT:
> Having a concrete builder and an abstract builder is definitely not fancy and harms interfacing for modded goals as well.

I don't want abstract builders. i don't want generic builders. I just want plain old builders. The current builders use recursive generics (A and B) to save a tiny amount of work when subclassing them. Also these classes are the leaves of the tree and are unlikely to be extended at all. It doesn't even help with the implementation.

The API should be easy to use. Having a fancy API design that make extending the API easy while it complicates the usage the API heavily, is not worth it.

Maybe we should discuss this in its own issue?

> If things do go bad, just use raw types. 

That does not sound like good API design.

> Moreover, builders aren't meant to be kept for long; they are intended to be consumed and disposed quickly.

So you force the users to use the classes in an a single specific way or create their own wrappers for them? Why did we make them resettable, copyable and reusable then?

liach:
@ST-DDT If the builders aren't generic, I am fine with overriding each method that return the object itself manually. Persuade zidane or gabizou to remove the self types. It is a shame that java does not support it natively indeed.

> unlikely to be extended at all

I promise that you are very wrong here. Many mods have extended vanilla goals (ai tasks), which we cannot represent correctly according to our api 7 model (especially if the base owner type get more specific).

> That does not sound like good API design.

Look into `CatalogTypes`. All those catalog types with possible generics are raw types there! You need to blame Java, not me.

> I just want plain old builders

If you really want simple and stupid, I suggest turning all goal builders into data builders backed by json data, or just make them return void on all methods and add a utility like 
```java
public static <T> void setup(T instance, Consumer<T> consumer) {
    consumer.accept(instance);
}
```
And do all setup of the builder in that consumer.

Again, the usage is not complicated. You just need to understand generics.

Zidane:
@gabizou Make the call on this please.

src/main/java/org/spongepowered/api/ai/goal/agent/LookIdleGoal.java
```diff
@@ -29,18 +29,18 @@
...
+public interface LookIdleGoal<O extends Agent> extends Goal<O> {
 
     /**
      * Creates a new {@link Builder} to build a new {@link LookIdleGoal}.
      *
      * @return A new builder
      */
-    static Builder builder() {
+    static <O extends Agent, B extends Builder<O, LookIdleGoal<O>, B>> B builder(O type) {
         return Sponge.getRegistry().createBuilder(Builder.class);
```
gabizou, line 40: if you're using the raw type, is there a raw type cast or unchecked cast going on?

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
```diff
@@ -49,7 +49,7 @@
      *
      * @return A new builder
      */
-    static Builder builder() {
+    static <B extends Builder<DamageSource, B>> B builder() {
```
gabizou, line 52: What's the point of the generic if it's just making a `Builder` that would otherwise build a `DamageSource`? Is it magically going to return a builder that would build an `EntityDamageSource` or say a plugin's customized damage source builder?

liach: The point of this generic is that the `Builder` is not a leaf in interface inheritance; other `Builder` can extend this builder with a different bound specified for `A` (the returned built type). Thus, we just need to have one builder declared for each damage source declared in inheritance instead of two,

src/main/java/org/spongepowered/api/ai/goal/agent/LookIdleGoal.java
liach: The use of the `Builder` must be a raw type without generics because the generic self type can never be specified. When users need to use, they need another method in which this generic type can be specified again for usage.

src/main/java/org/spongepowered/api/event/cause/entity/damage/source/DamageSource.java
ST-DDT: Another reason to dump the generic self types.

liach: So @ST-DDT, if we dump generic self types, two ways to go:
1. Make all builder methods return void and use things like 
```java
Sponge.configure(DamageSource.builder(), $ -> {
    $.setMagic(true);
    $.setByPassesArmor(true);
    // More
});
```
instead.

2. Manually override the return type in each subclass (which is almost impossible).

Gabi and ST, thoughts?

ST-DDT: 1) No.
2) Yes. With the help of an annotation processor that should be possible. It's a one time efford as the root builders dont change. It also has the added benefit that you can see all options of a builder by looking at its source.

liach: Sure, we could do that, just like what we did to sponge event factory. now, i think it is up to gabizou to decide.

gabizou: Annotation processor sounds fine, making it part of the build process to ensure we don't distribute jars with invalid builder types.

liach: So do add another gradle task that use spoon or one that uses asm to patch the built api jar? I don't know if spoon can modify existing generated source.

gabizou: I'd rather we use spoon over patching the built api jar.

Zidane:
@liach doesn't appear he will be doing anything more with stuff like this, closing.

Issue #1975, opened 2019-02-26
Enhance Visibility of EventContextKey Usage


gabizou:
#### First, some background (there is a TLDR at the bottom, but really you should read this issue)
As the API and official implementations have evolved over the years, @Zidane @bloodmc and I are coming to a head about exposing bits of what we can expect for being able to better use Sponge's implementation for control flow of various events. As most plugin developers are aware, Sponge's tracking system does an impressive job at capturing and emitting objects for [`Cause`](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/event/tracking/phase/tick/EntityTickPhaseState.java#L67), and [`EventContext`](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/event/tracking/phase/packet/player/AttackEntityPacketState.java#L67). However, we've had a recent discussion to add more `EventContextKey`s to provide some insight to plugins about using [`ChangeBlockEvent.Pre`](https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/event/SpongeCommonEventFactory.java#L515) as it serves as a "pre-process a bunch of things" event that can safely be used to cancel things like Pistons (because pistons are already [requiring a rewiring/re-engineering of some bits about how captures are done](https://github.com/SpongePowered/SpongeCommon/pull/2163)...

#### The point
So, having given some background about how `EventContextKey`s are being populated, I'd like to get to the root of this discussion: **We want to provide visibility via documentation and possibly default provided methods for common expected `EventContextKey`s that would be more than likely available for various events**.

So far, we're agreeing and will be committing to adding Javadoc annotations (like `@see`) to various `Event` classes to suggest possible `EventContextKey`s that would be available to those events, and documenting when those keys would be optionally provided (better to sanely explain the contracts defined in the API for implementations to abide by). What is an option, and the root of this discussion is whether we, the API, will also define *default implemented method accessors* for these optional `EventContext`s that we're almost sure will exist, but still having an `Optional` return. This would be the same treatment that [`Entity` interfaces went through to facilitate "guaranteed" accessors with Data API](https://github.com/SpongePowered/SpongeAPI/commit/26791b9aeb927f0f8959b6a5950e63b153b12d85), except, because the events are still contracted to have no guarantee on what will exist in `EventContext`, the methods would still be `Optional`.

#### Examples

[...]

PR #2004, opened 2019-05-14
Remove the CommandSource in favour of Causes.

dualspiral:
This is marked as a Draft PR so that the extent of the changes I'm proposing can be seen. It is intended as a way to foster discussion on a specific point that previous PRs could not.

**We are investigating how best to utilise the Cause system in our Command framework - the code in this PR is the purest option** - but hear me out as I explain what we're really looking at. This is long, but please read all of it before passing comment because I'm more interested in reaching convergence of ideas rather than a "I don't like this, put it back how it was."

### A recap of `Cause`s (and why we all should love them!)

The SpongeAPI and its implementation is a system driven by `Cause`s. They have primarily been used in events, though they have crept into data registration and other uses where we determine which plugin is doing what. It is a powerful tool in the plugin developer's toolkit, enabling not only the determination of redstone being caused by redstone, but being able to track the player who spawned the sheep who aimlessly walked onto a pressure plate turning on redstone wire and causing an explosion, or opening a door, or firing a command block...

...except actually, we don't fully expose the cause to commands (we could get the cause from the CSM, but that's a live snapshot and _not_ the cause at the specific time a command was sent). So, we are setting out to remedy this. If a player fires a command block because they stood on a pressure plate, we want to be able to directly tell a command consumer "hey, this player was the one who set your command off on this command block. Maybe you want to target them?" 

This could be really handy for game servers who use command blocks or signs to trigger events, but want the context of who triggered it, rather than who is nearby. So, we have been talking a lot internally (and in previous command PRs), but I wanted to really get opnions on this point, rather than as part of a big API PR because this is rather important.

### Causing those Commands

The basic idea is that you could check the `Cause` for what's going on. The root might be a sign, but you can the scan the cause and find a `Player` who clicked on the sign. That sign could be a `Subject`, so permission checks are performed on that sign - but the player is the first `MessageReceiver` - so you send the messages to them! Similarly, you might have a command block executed by a sheep on a pressure plate - but you want to execute the command on a player only. So, no problem, let's just find the first `Entity` on the stack and just not execute if it's a player. By opening causes to the command system, commands can take advantage of this rich, contextual system we already have and make Sponge an even better system to develop commands against.

Even better, we could expose our Event filters to Commands, so we can reduce the amount of boilerplate you write - want to have a command that only executes if the root of the command is a `Player`? Perhaps two different logic paths if you have a root `Player` or a root `Console`? While this is a little later down the line - `Cause`s will really enable all this, allowing command executors to get the same treatments as events (though specialised).

### Removing those Sources?

Of course, that shouldn't all be too contriversial... but you may have been alarmed by the title of this PR. Yes, we are considering the removal of `CommandSource`s in the API. The simple reason is, the cause already tells you the source of the command - particularly the root of the cause. The `CommandSource` is in API 7 a collection of the following:

* `Subject` - an object that permissions can be attached to;
* `MessageReceiver` - an object that can receive `Text` messages;
* A `getLocale()` method (moved to `MessageReceiver` in API 8 as that is a more logical place for it
* A `getName()` method for identifying this source as a `String` - this has been moved to `NamedIdentifiable` (but that name sucks IMO and could do with a better one if this goes ahead)

So, in most cases, these interfaces will simply replace the `CommandSource` and the various sub sources that existed prior to this. **However**, there are some sub sources that have already disappeared:

* **`ProxySource`**: this is a source that wraps around a source to intercept permission checks etc. A common use for it _should_ be to front a player when executing a command from a different source, such as a Discord bot. In future, the Discord bot plugin should already be in the cause, but can push a `Player` to the cause prior to execution (or the player an then some other proxy `Subject` if they want to control permission checks, for example).
* **`RemoteSource`** didn't belong - the command system in itself shouldn't care if a command executor is remote.
* **`RconSource`** is now an `RconConnection` in the appropriate package
* **`SignSource`** is now just the `Sign` that would be pushed to the `Cause`, **and would be a `Subject` only**.
* **`CommandBlockSource`** similarly is just the `CommandBlock` and `MinecartCommandBlock`

### Cause and Effect (on Plugin Developers)

You may see a problem here - plugin developers will still likely want to get a traditional command source out of the cause - such that they get an easy object to obtain the console or the player out of. So, we might want to consider a `CommandTarget` interface that can be used to group the `Console`, `Rcon` and `Player`s together - or more precisely, interfaces that indicate objects that would operate with commands more _interactively_ and would normally be the recipient of messages. Of course, command blocks would probably have to be in that category too as the output of command are important to them.

We also have to think about how Sponge managed components (parameters etc.) handle the cause for permissions and messages, but I was likely to go with first `Subject` and first `MessageReceiver` (there are currently overrides in the Context, but I fear they would be ignored and we are better just focusing on the direct `Cause`).

There is the potential to be more confusing - but with good documentation and established standards we can absolutely make the `Cause` the new `CommandSource`. We just need to make sure we do this right so it works for everyone.

### Sourcing those Opinions!

So, the question is, do we:

* Just remove all `CommandSource`s like this and document that permission checks should be performed on the first `Subject`, similarly, messages should go to the first `MessageReceiver`?
* Add a `CommandTarget` marker interface that performs a "quasi-`CommandSource`" like job, but this rename indicating that command sources are no longer actual sources (that's the root of the `Cause` now!)
* Always make `CommandSource`s the ultimate root of an object (this is closest to what Mojang does but we don't necessarily need to follow that for Sponge plugins)

Open for discussion, but as always, please keep it civil and please give reasoning. Note that we will likely go somewhere between the first two points as it stands.

---

Just on the note of the `Console` object appearing, it's there simply because I'd completely forgotten that I'd started working on it until most of the way through this PR. Note that the `Server` and `Console` objects are both `Subject`s to account for the possibility that (as said elsewhere) one console could support multiple servers and so someone, somewhere, might have different permissions per server... but that could also be too complex for the API. That, however, will likely go into api-8 sooner, once @Zidane, @Cybermaxke and I have settled on this (though I doubt we're far off that).


src/main/java/org/spongepowered/api/Game.java
```diff
@@ -86,6 +86,13 @@
      */
     Server getServer();
 
+    /**
+     * Gets the {@link Console}
+     *
+     * @return The console
+     */
+    Console getConsole();
```
liach: Should bound it to a server as the console ain't really doing anything until the server has started. Also consider preserving integrated server compatibility.

dualspiral: https://github.com/SpongePowered/SpongeAPI/commit/a84bf1def1e44936cea81b5a278714fa6f0f9857#r33517014

liach: How can you ensure a game only has one console then? At least document its availability (in vanilla, it is only available when a **dedicated** server is present)

Cybermaxke: Does a integrated server have a console directly? Or is it tied to the client (for logging)?

dualspiral: > (in vanilla, it is only available when a dedicated server is present)

Yes, in vanilla. We very well could have a console object that is available at game start, rather than server start if we wanted. We might mod the game, but that doesn't mean we need to follow absolutely _everything_ that Mojang does.

I really don't care where this goes and when the console is available, but the thought behind this is that we have a separation of console and server and this is what made sense to me as a result of the conversation there. We might always want a console to exist represent a super user, regardless of whether a physical console exists.

I'm just glad that putting this method here has sparked conversation about it, I just wish I'd put it in a separate PR now and wasn't being so lazy...

liach: Ah, didn't notice that console was not for command but is for a more general purpose, mainly logging. I agree with this change then.

ItsDoot:
I vote for the first option of selecting the first `Subject` and first `MessageReceiver`.

1. It's a relatively simple solution to describe to developers new to Sponge.
2. Works across any command framework one may use, be it the one Sponge provides or a third-party library.
3. It unifies the command system with the event system.
4. It just *feels* right.

For a while I was really adamant against replacing `CommandSource` with `Cause`, but after some thought it actually seems like the best way going forward.

src/main/java/org/spongepowered/api/entity/living/player/Player.java
```diff
@@ -68,7 +69,7 @@
  * <p>Any methods called on Player that are not on User do not store any data
  * that persists across server restarts.</p>
  */
-public interface Player extends Humanoid, User, Viewer, ChatTypeMessageReceiver, CommandSource {
```
Zidane, line 73: ChatTypeMessageReceiver is not a MessageReceiver?

dualspiral: [Nope](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/text/channel/ChatTypeMessageReceiver.java#L40)

Zidane: I feel like we could extend MessageReciever and default the methods to call the chat type methods with ChatType.CHAT?

dualspiral: I'll put that into another PR (or directly onto `api-8`)

Zidane: Do it directly in api-8.

dualspiral: Okay.

src/main/java/org/spongepowered/api/util/NamedIdentifiable.java
```diff
@@ -22,5 +22,18 @@
...
  * THE SOFTWARE.
  */
-@org.spongepowered.api.util.annotation.NonnullByDefault
-package org.spongepowered.api.command.source;
+package org.spongepowered.api.util;
+
+/**
+ * An identifiable object that uses a {@link String} as an identifier.
+ */
+public interface NamedIdentifiable {
```
ItsDoot: `UniquelyNameable` or perhaps `UniquelyNamed`?

kashike: Should just be `Named` or `Nameable` - existing `Nameable` interface should be renamed to something else.

JBYoshi:
I might be misunderstanding how this will work, but I see a few problems with simply selecting the first `Subject` or `MessageReceiver`.

Say that a player steps on a pressure plate that triggers a command block. I'm assuming the player would be earlier in the cause (because the pressure plate processing occurs before the command block processing).
- This would mean that a plugin grabbing the first `MessageReceiver` would get the player, and output messages would then go to the player. Normally, these messages would be [stored in the command block](https://github.com/SpongePowered/SpongeAPI/blob/3fdbf2993d487e0170ce8f77823805e86fee254f/src/main/java/org/spongepowered/api/data/key/Keys.java#L1364) instead of being sent to the player.
- If a plugin selects the first `Subject`, it would check the permissions of the player instead of the command block. The command execution should usually have the permissions of the command block, but it gets the player instead - who may have fewer permissions. (This case would also occur if a player clicks a sign and ends up before the sign in the cause.)

On the other hand, selecting the last one would also break in some cases. Take `/execute` as an example. Say that the console types `/execute as Steve run somecommand` (or `/execute Steve ~ ~ ~ somecommand` in 1.12 notation). The command should check the permissions of the console. However, if the console is before the player, selecting the last `Subject` would return the player, which would break the permission test.

Is there something I'm missing here?

dualspiral:
Remember causes are last in, first out.

> This would mean that a plugin grabbing the first MessageReceiver would get the player, and output messages would then go to the player
> 
> If a plugin selects the first Subject, it would check the permissions of the player instead of the command block. The command execution should usually have the permissions of the command block, but it gets the player instead - who may have fewer permissions. (This case would also occur if a player clicks a sign and ends up before the sign in the cause.)

No, the first `MessageReceiver` and `Subject` would be the `CommandBlock` as the cause is effectively a last in, first out - `@First` really means latest on the stack. So, when the command block is triggered, that command block should be the root and the "first" subject/receiver.

> The command should check the permissions of the console. However, if the console is before the player, selecting the last Subject would return the player, which would break the permission test.

This is the tricky part. The original solution was to use the `EventContext` as an override (so, if there is an object in `EventContextKeys.SUBJECT`, that would be checked as opposed to the direct cause). However, I explain above:

> We also have to think about how Sponge managed components (parameters etc.) handle the cause for permissions and messages, but I was likely to go with first Subject and first MessageReceiver **(there are currently overrides in the Context, but I fear they would be ignored and we are better just focusing on the direct Cause)**.

In this case, there would be a specialist `Subject` that contained the permissions of the super user on the top of the stack, but the player second in the stack. This would allow the first `Subject` to be super users permissions, and the first `MessageReceiver` to be the player.

Now, this is why I want to use event contexts as part of our command structure! The player isn't a direct cause, and so this is part of the event context - but if we do that, plugin developers aren't going to want to check the context before the cause each time. As I write this, I wonder if we'd be better served by creating a `CommandCause` object that has `getCause()`, `getSubject()` and `getMessageReceiver()` methods on it so that most plugins can just use the subject and message receiver easily, but have access to the `Cause` (which is all controlling this anyway).

Of worthy note, if we were to retain `CommandSource`, then we'd still have to use a special proxy source to mask the player permissions when using `/execute`, so this is not a problem that has been created by making this switch - we would still have to make a subject proxy.

src/main/java/org/spongepowered/api/util/NamedIdentifiable.java
liach: The existing one should become `Translatable`.

ejm:
> As I write this, I wonder if we'd be better served by creating a `CommandCause` object that has `getCause()`, `getSubject()` and `getMessageReceiver()` methods on it so that most plugins can just use the subject and message receiver easily, but have access to the Cause (which is all controlling this anyway).

So, if I'm understanding this correctly, a `CommandCause` would be effectively the drop-in replacement for a `CommandSource`? I don't see why this is any better than just having `Location`, `Rotation`, `Subject`, etc. each stored as an `EventContextKey`? Yes, plugin developers would now need to check those, but, if you have mappings to them in the existing `CommandContext` (which already has methods for a lot of these things already), you wouldn't need to introduce a new dummy `Cause`. I don't think it is too much to ask from developers to utilize the `CommandContext`.

I understand the challenges in that, though. It is tempting to just grab the first `Player` off the `Cause` stack and run with it, but, commands aren't that simple, especially with `/execute` being as powerful as it is. Plugin developers will need to keep it in mind no matter what or risk their commands having unexpected behaviors, so, using `CommandContext` to check for permissions, location, etc. doesn't seem like a huge ask.

> Of worthy note, if we were to retain `CommandSource`, then we'd still have to use a special proxy source to mask the player permissions when using `/execute`, so this is not a problem that has been created by making this switch - we would still have to make a subject proxy.

One of the developers on a project I work on currently does this using delegation, and, it *almost* works, but, Vanilla commands deal with `CommandSource`s a little funny so it causes weird side effects with Vanilla permissions. No matter what ends up happening, could this possibly be rectified?

dualspiral:
> I don't think it is too much to ask from developers to utilize the CommandContext.
> 
> Plugin developers will need to keep it in mind no matter what or risk their commands having unexpected behaviors, so, using CommandContext to check for permissions, location, etc. doesn't seem like a huge ask.

I think you're assuming that command consumers will always have a `CommandContext`. They won't - if you implement a command without using our framework, you simply get the `Cause`, not this context. That's why I'm proposing a `CommandCause`, it will effectively be a wrapper around the `Cause` with those methods on the `Context`. Everything will still be controlled by the `Cause` itself - the `CommandCause` would not be a dummy, but informed by (and exposes) the cause itself.

Effectively, I'm proposing that this line:

https://github.com/SpongePowered/SpongeAPI/blob/0520f48b05bc026c083f490991cdf3a39fd58cbf/src/main/java/org/spongepowered/api/command/Command.java#L122

becomes 

```java
CommandResult process(CommandCause cause, String arguments) throws CommandException; 
```

where `CommandCause` will have a `getCause()` method, plus the utility methods on `CommandContext`. As `CommandContext` already contains the methods, that wouldn't change on the context. It's simply a way to ensure that these methods are always available.

As for vanilla commands, yes, I'll keep this in mind.

ejm:
I was just writing a reply after I remembered that `Command` doesn't necessarily need to use `CommandContext`. `CommandCause` makes sense in that case! Disregard my above comment aside from the vanilla-related concern. Thanks!

dualspiral:
Note to self - `CommandSendEvent` was forgotten... need to readd.

src/main/java/org/spongepowered/api/util/NamedIdentifiable.java
dualspiral: Just had a look at what the existing `Nameable` is used for, it's used for `Inventory` only (and in the JDs for `QueryOperationTypes`). One thing we can do is:

* Rename `NamedIdentifiable` to `Nameable`
* Rename `Nameable` to `Translatable`
* Rename `Translatable#getName` to `Translatable#getTranslation`

We could also look to make player names `Translations`, but that's probably overkill.

Thoughts?

dualspiral: I've been thinking about this more: the name `Translatable` for the current `Nameable` interface doesn't quite fit, nor does the proposed method rename. An inventory is not translatable, its name is, and that's what the interface indicates. So, I don't think `Translatable` is right here, nor is `getTranslation`.

This is the remaining thing left to decide before I merge this and I really don't have a good idea for what we should do - a name that indicates that it has a translatable name eludes me.

Cybermaxke:
`ClickAction.ExecuteCallback` still uses `Cause`, which should be changed to `CommandCause`. Its factory method in `TextActions` also needs to be readded.

src/main/java/org/spongepowered/api/command/CommandCause.java
```diff
@@ -0,0 +1,187 @@
...
+     * may be thought of as a traditional entity executing the command.
+     * For the object that invoked the command, check the
+     * {@link Cause#root()} of the {@link #getCause()}.</p>
+     *
+     * @return The {@link MessageReceiver} to send any messages to.
+     */
+    default MessageReceiver getMessageReceiver() {
+        return getCause().getContext()
+                .get(EventContextKeys.MESSAGE_TARGET)
+                .orElseGet(() -> getCause().first(MessageReceiver.class).orElseGet(Sponge::getSystemSubject));
```
Cybermaxke, line 167: Wouldn't it be better to return a `Optional` here. Right now falling back to the "system subject" will result in logging all messages to the console, e.g. for command blocks, signs. Which isn't desired in most cases IMO. You could make your own method, but that's less convenient. Adding a separate method which falls back to the system subject is also an option.

src/main/java/org/spongepowered/api/command/CommandCause.java
```diff
@@ -0,0 +1,187 @@
...
+ */
+public interface CommandCause {
+
+    /**
+     * Creates a {@link CommandCause} from the provided {@link Cause}
+     *
+     * @param cause The {@link Cause}
+     * @return The {@link CommandCause}
+     */
+    static CommandCause create(Cause cause) {
```
Cybermaxke: `of` to be consistent with all the other factory methods

src/main/java/org/spongepowered/api/command/CommandCause.java
dualspiral: > Which isn't desired in most cases IMO. 

And won't normally happen in this system either. Command blocks are message receivers anyway so messages will go to them. Signs are explicity _not_ `MessageReceivers` (as it stands) and I would imagine that a sign would be triggered by a player, which would also be in the cause and so the message would end up with the player anyway. In the cases you've described the message generally _wouldn't_ go to the system subject - and I expect that to be the exception, rather than the norm.

This is intended to catch the scenario that a plugin executes a command that wasn't caused by a player, e.g. during a scheduled task, where a command would end up going to the "console" (as that is the command source you picked).

Finally, getting a `MessageReceiver` is a fundamental thing that command executors will do - I suspect wrapping it in an `Optional` is going to frustrate more than help, and that the vast majority of players will just add a `orElseGet(() -> Sponge.getSystemSubject())`. This is intended to point them at the appropriate receiver - if all else fails, that's the `SystemSubject`. If I add another optional method, you can bet no-one is going to use that in favour of this.

pie-flavor:
+1 for the CommandCause idea. One of the primary failings of a map-and-key system is that what there is supposed to be is not immediately clear; which ones are optional in a particular context vs which ones are just never present in that context, for example. EventContextKeys should still be accepted *in* that cause, but a wrapper API guaranteeing things like location of execution, dimension of execution, alignment, rotation, executor, original executor, should have dedicated methods in some kind of wrapper class, that *don't* return Optional, as guarantees that these things always exist. Especially because these things are only used in commands and nowhere else, and context keys used lots of other places won't be used here.

liach:
Soon:tm: sponge will be the plugin api that can detect the malicious plugin that called "player.setOp" :laughing:

Note for implementation: While most parts of minecraft can be simulated by the vm stack, the command functions are stored in an external stack and their cause, etc. needs to be handled delicately.

src/main/java/org/spongepowered/api/util/NamedIdentifiable.java
dualspiral: [Okay, I've done this - folding the existing `Nameable` into the "new" `Nameable` as `Nameable.Translation`](https://github.com/SpongePowered/SpongeAPI/blob/1584092d1eb8fc8a578370ced2e6ce2d76254f91/src/main/java/org/spongepowered/api/util/Nameable.java). Provides the link, resolves the concerns.

dualspiral:
Any more comments? I'm going to merge this soon otherwise (and fix a couple of javadocs that have TODOs on them).

pie-flavor:
We have the location but not the rotation of the command. Maybe change the Location to a Transform?

dualspiral:
How does that work if a block is the focus of the command?

pie-flavor:
It would face the direction the block is facing (since command blocks are directional). Or maybe it'd be a rotation of zero if it's a nondirectional block, I don't know. But commands can have relative rotations to the execution targets, so a. it's gotta be in there and b. vanilla knows what to do.

liach:
Making sure that all quantities in vanilla `ServerCommandSource` is present in the cause stack, and we should be fine.

dualspiral:
> Or maybe it'd be a rotation of zero if it's a nondirectional block, I don't know.

And this is why it shouldn’t be a `Transform`, there is a difference between 0,0,0 and no rotational data. Adding a rotation separately makes more sense in this case.

I do have to point out that this is all driven by the `Cause` and the `CommandCause` is solely driven by that. There is no separate cause stack for a command in normal usage. I’m not keen on adding every helper method under the sun, in the example of a command block, it’s going to be the message receiver so you can always get more information from that. Similarly, we do have the target block method which would also have this info.

I am not saying I’m against adding the rotation, but do remember that the `CommandCause` is to simplify common actions and to allow plugins to harmonise with the impl (what Subject is being checked, etc.) and is not intended as a draconian toolkit that removes the point of the Cause in the first place.

Plugins are allowed to think for themselves too!

pie-flavor:
Right. The point of a CommandCause is to provide stuff that's guaranteed to be relevant.
The reason, however, that you *can't* just get the rotational data from the command block, is because commands have a relative rotation just like they have a relative position that may or may not be the execution target's actual position or rotation.

dualspiral:
> The point of a CommandCause is to provide stuff that's guaranteed to be relevant.

Where did I say that? Also what determines what is relevant? The `Cause` contains all information. The `CommandCause` contains that `Cause` and helper methods. **That's it**.

Very much like how events don't have any particular guarantees, we've made that the case with commands too by making it use the `Cause` system. The point is that command can see _exactly_ where the command came from, who executed, why that execution occured, etc. etc. It all goes on the `Cause` and the `CommandCause` is a simple wrapper for the `Cause` that explains how the implementation will select a `Subject`, for example, so that plugins don't need to check the `SUBJECT` context on the cause before looking for the first `Subject` in the `Cause`. Same with the `MessageReceiver`.

Quite simply, the point of the `CommandCause` is to harmonise how the `Cause` is used across Sponge and plugins, particularly across obtaining the `Subject` and `MessageReceiver`s.

The `Cause` is the ultimate control of this process. Everything will be present in the `Cause`, the `CommandCause` has no other information beyond this. This wrapper can simply be thought of as a specialised "event". That's it.

pie-flavor:
Exactly. And the event has a rotation associated with it.

dualspiral:
Okay. I don't really think you're getting this point of this entire command system here and you are fixated on one point that I've not even rejected. 

Commands in Sponge can be summed up like this:

**Commands using the Sponge system are different to those using the Vanilla system. By using the Cause system we, in effect, state that our commands have a different set of criteria upon execution - which is entirely our right to do.**

Originally, after much discussion with @Zidane, I was just going to provide the `Cause` of the command and tell developers to go nuts. We  make absolutely no guarantees about _anything_ when a Sponge command is invoked. The is no requirement for a `CommandSource` (indeed, they won't exist in our system), no requirement for a `Subject` to be present, or a `MessageReceiver`, it could just be a block and a sheep for all we care. The `CommandCause` is simply a set of methods that allow plugins to use the same objects that Sponge selects for the subject of a command, or the message receiver. It's not a guarantee - in fact, it fills two gaps - what we do if there is no `Subject` or `MessageReceiver`.

We make no guarantees about anything, because **Sponge commands have no guarantees about how  they are executed**. Now, if someone uses the `/execute` command to build up a source that has a rotation, then we'll add that to our cause. Yes, I will likely add an `Optional<Vector3d>` (even though the rotation is a `Vec2f` in the mojang code) that adds a rotation - again, not something I ever said I wouldn't do. In fact, using `/execute` will likely end up filling the `EventContext` with keys _such as_ `ROTATION`.

Should `Rotation` be on the `CommandCause`? Yes, considering `Location` is on there it makes sense and it is something I missed. **I never said that I wouldn't do it**, I just haven't had enough time with all my other personal commitments to actually do it. But it seems to me that you've misunderstood what this system really is getting at - if you want something, go get it from the `Cause` because that really has all the information you need.

Of course, as a quasi-event, I'd be following what https://github.com/SpongePowered/SpongeAPI/issues/1975 does.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -63,10 +63,10 @@
  * as required. This <em>may</em> enable the use of client side command
  * completions, if the implementation is equipped to do so.</p>
  *
- * <p>Commands in Sponge are provided with a {@link Cause}, which explains
- * who <strong>directly</strong> invoked the command. In line with causes used
- * in events, you may assume that the {@link Cause#root()} is the
- * <strong>intended</strong> direct invoker.</p>
+ * <p>Commands in Sponge are provided with a {@link CommandCause}, which
```
Zidane: May need re-wording.

dualspiral: Opted to move this part of the description to the CommandCause interface, makes sense to lump that all together.

PR #2007, opened 2019-05-17
Eliminate ticks

Cybermaxke:
Eliminates the remaining usage of ticks. Timings still need to be done.

### DIFF
--- a/src/main/java/org/spongepowered/api/EngineStatistics.java
+++ b/src/main/java/org/spongepowered/api/EngineStatistics.java
@@ -1,0 +1,104 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api;
+
+import org.spongepowered.api.util.temporal.TemporalUnits;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.TemporalUnit;
+import java.util.List;
+
+/**
+ * Represents the statistics of a {@link Engine}.
+ */
+public interface EngineStatistics {
+
+    /**
+     * Gets the amount of game loop updates that are being
+     * performed per {@link TemporalUnit}.
+     *
+     * @param unit The temporal unit
+     * @return The amount of updates
+     */
+    double getUpdatesPerUnit(TemporalUnit unit);
+
+    /**
+     * Gets the amount of game loop updates that are being
+     * performed per second.
+     *
+     * <p>In vanilla minecraft this value represents the
+     * amount of ticks per second.</p>
+     *
+     * @return The amount of updates
+     */
+    default double getUpdatesPerSecond() {
+        return getUpdatesPerUnit(TemporalUnits.SECONDS);
+    }
+
+    /**
+     * Gets the average amount of game loop updates that are
+     * being performed per {@link TemporalUnit}.
+     *
+     * @param unit The temporal unit
+     * @return The amount of updates
+     */
+    double getAverageUpdatesPerUnit(TemporalUnit unit);
+
+    /**
+     * Gets the average amount of game loop updates that are
+     * being performed per second.
+     *
+     * <p>In vanilla minecraft this value represents the
+     * average amount of ticks per second.</p>
+     *
+     * @return The amount of updates
+     */
+    default double getAverageUpdatesPerSecond() {
+        return getAverageUpdatesPerUnit(TemporalUnits.SECONDS);
+    }
+
+    /**
+     * Gets a list with {@link Instant}s at which updates
+     * have occurred recently.
+     *
+     * <p>In vanilla minecraft is the size of the list limited
+     * to a 100 values.</p>
+     *
+     * @return The recent update times
+     */
+    List<Instant> getRecentUpdateTimes();
+
+    /**
+     * Gets the duration that the engine has been running for, since the
+     * engine started.
+     *
+     * <p>This value is not persisted across engine restarts, it is set to zero
+     * each time the engine starts.</p>
+     *
+     * @return The duration the engine has been running for
+     */
+    Duration getRunningTime();
+}
--- a/src/main/java/org/spongepowered/api/EngineStatistics.java
+++ b/src/main/java/org/spongepowered/api/EngineStatistics.java
@@ -59,28 +59,6 @@ public interface EngineStatistics {
     }
 
     /**
-     * Gets the average amount of game loop updates that are
-     * being performed per {@link TemporalUnit}.
-     *
-     * @param unit The temporal unit
-     * @return The amount of updates
-     */
-    double getAverageUpdatesPerUnit(TemporalUnit unit);
-
-    /**
-     * Gets the average amount of game loop updates that are
-     * being performed per second.
-     *
-     * <p>In vanilla minecraft this value represents the
-     * average amount of ticks per second.</p>
-     *
-     * @return The amount of updates
-     */
-    default double getAverageUpdatesPerSecond() {
-        return getAverageUpdatesPerUnit(TemporalUnits.SECONDS);
-    }
-
-    /**
      * Gets a list with {@link Instant}s at which updates
      * have occurred recently.
      *
--- a/src/main/java/org/spongepowered/api/util/TickBasedByDefault.java
+++ b/src/main/java/org/spongepowered/api/util/TickBasedByDefault.java
@@ -22,43 +22,24 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.entity.living;
+package org.spongepowered.api.util;
 
-import org.spongepowered.api.data.key.Keys;
-import org.spongepowered.api.data.value.BoundedValue;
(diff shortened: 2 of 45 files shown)


src/main/java/org/spongepowered/api/EngineStatistics.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    default double getUpdatesPerSecond() {
+        return getUpdatesPerUnit(TemporalUnits.SECONDS);
+    }
+
+    /**
+     * Gets a list with {@link Instant}s at which updates
+     * have occurred recently.
+     *
+     * <p>In vanilla minecraft is the size of the list limited
```
Zidane, line 65: Grammar, sentence doesn't  make sense. Remove the is :p.

simon816:
Ticks are a fundamental idea in just about any game engine. Is there a reason to remove the notion of a tick other than "ticks are not a real unit"?

Cybermaxke:
@simon816 Completely depends on the implementation, and even if you use ticks, who says you need to use 20 ticks per second, maybe 50 per second is an option. This also eliminates all the different units being used across the API, like this setter uses ticks, another one seconds or milliseconds.

Cybermaxke:
Note that you can still use "minecraft ticks" as a `TemporalUnit`, see `TemporalUnits.MINECRAFT_TICKS`

Zidane:
If the Temporal classes are remaining concrete classes in the API, they need some tests.

src/main/java/org/spongepowered/api/util/TickBasedByDefault.java
```diff
@@ -0,0 +1,45 @@
...
+ * that e.g. a duration of 5 seconds will be ideally 100 minecraft ticks.
+ * But in reality can 100 ticks be longer then 5 seconds depending on the
+ * actual tick length.
+ *
+ * <p>You can't assume that the duration of a given length will be exactly
+ * the same, this depends on the implementation.</p>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE_USE)
+public @interface TickBasedByDefault {
```
dualspiral, line 43: If the whole point of this PR is to remove any notion of a "tick", this name is screaming of "our reference implementations use ticks". It does, but that's an implmentation detail and doesn't really explain why that's important. I also know we're not going to get away from talking about ticks completely - but I think references to ticks should be consigned to the javadocs.

What you're _really_ saying here is that the input is the minimum time - if the server is overloaded the target may take longer than indicated. `@PerformanceDependent` might be a better name, or just some name that indicates that a lagging server may extend the time, rather than stating that we're just estimating based on tick time. That way we're stating that a lagging server may extend the duration, without suggesting that we're backing the field with ticks, and we're not expressing an implementation detail.

This PR is also a break from working with ticks for a lot of developers. If we're removing ticks, then this needs to be crystal clear as to what these durations mean. I think the javadoc needs rewording, something like (but perhaps not exact):

> Marks a parameter or return type that represents a duration as dependent on server performance, and should be considred a minimum time, rather than an absolute time.
>
> If the server operates under perfect conditions (for example, 20 ticks per second (TPS) in vanilla minecraft), then the duration that has been set or returned will be accurate. However, if the server performance is degraded at any moment (in vanilla, a drop in the TPS), then the given duration might be considered an underestimate if the underlying implementation uses ticks rather than wall clock time to time actions. 
>
> The effect of server performance drops on annotated types is dependent on the implementation that is being used and the settings that are applied. Therefore, no assumptions should be made about how durations are calculated, it is implementation dependent. 
>
> Plugins wishing to execute tasks based on any returned duration annotated with this annotation should check that the desired action has occured rather than solely relying on what is returned. Plugins setting times annotated with this annotation should be aware that the underlying implementation may convert the time into a different format and, as such, the stored time may not be exactly the same as requested.
>
> Any duration annotated with this annotation may be used with any of the units provided in `{@link TemporalUnits}`.

dualspiral:
I'm not completely against this, but there have been some discussions over this and I wanted to get my thoughts down, as well as try to help get discussion going where it's going to be recorded.

The fundamental problem with this PR is the following:

## Ticks Are Not Time Units and Should Not Be Treated As One

While I can understand the reason why we might want to remove the idea of a tick from the API, as noted above, a tick is not a time unit, it is a unit of work that the server does and the beginning and end of a tick represent synchronisation points, snapshots where everything in the game is consistent with each other. A common misconception that a tick is 50ms - it is infact not at all. The vanilla game strives to **complete** each tick within 50ms, such that it can perform 20 units of work a second. A tick itself could last a much smaller amount of time and is different from tick to tick.

In a perfect world, the difference in time between the **beginning** of each tick is 50ms. Note that I didn't say **end**. So if we could guarantee that the tick **spacing** is 50ms, then it is a valid unit of time. **However, we cannot, and we can never guarantee that a tick _itself_ takes 50ms.** Now, that latter part is me being pedantic over phaseology but I feel it needs cleaning up. This is especially important if we run a task at the end of a tick - the task may happen 54ms later if the first tick took 2ms and the second 6ms.

We shouldn't be inferring a concrete connection between ticks and time, as is being proposed by including the `TemporalUnits#MINECRAFT_TICKS` unit for use with `Duration`s, because that really is comparing apples with oranges, about all you can say in that case is that they are both fruit, or in this case, indicate an action somewhere in the future. We say in the javadocs that one tick is 50ms... we'll we've now coupled back up to the implementation. Another implementation might not use a 20 TPS model, they might have 40, or 10, or 13, or 3.

I don't, however, see the issue of having `expectedTickRate()` and `actualTickRate()` methods somewhere that exposes this information. Could also have a `float currentServerPerformance()` that returns a float between 0 and 1 that indicates what percentage of updates are being fired per second.

## Ticks over Durations?

Shoving the notion of units of work that only has some approximate time meaning into a time based `Duration`, in light of this, is just plain wrong. If a server takes longer than 50ms to tick and you therefore don't get 20 ticks per second, then the duration is wrong. I know about this annotation that you've proposed (and I've commented on at length), but I almost feel like we're doing a disservice to API consumers by saying "well, this storm will last 40 seconds, but your TPS is 15 so actually, it'll be longer, but I won't tell you that directly, you should just know that, have fun". If they then set a scheduler to run after 40 seconds, the storm still has 10 seconds or so to go - if they were able to schedule on ticks, they would have been right on the mark. Durations in this way strip the consumer of syncing up with the game.

The counterpoint is that we should have events that cover these possibilities - which is probably true, but you may want to sync up with a mod that works in ticks.

I suppose an alternative is to create an event that fires on tick end, but I can see that being abused by plugin developers very quickly.

## Durations over Ticks?

There are situations where 20 ticks doesn't actually mean 1 second. 

* By enabling the realtime mixin, you are fundamentally changing the assumption of the game engine, "do something in 20 ticks" may suddenly become "do something in 10 ticks because our ticks have taken twice as long as the game is expecting" - in fact, this is the engine deciding what it should run per tick based on the duration, not the tick count. In that sense, the Duration is more accurate, not ticks.
* A tick time is arbitary, and 20 ticks per second might not be used in all implementations. Exposing ticks means that plugin developers are going to be lazy and assume 20 TPS across all implementations (which is, I suspect, a fundamental driving force in this PR)
* There is also the potential for an implementation to _not_ use ticks.

Also, a duration is a natural unit that we're all used to. Specifying ticks is not. However, with the way the game engine works, **all** javadocs (and even method names) for methods that get/set durations with a tick based storage with the words "expected" or "approximate".

## Ticks or Durations?

The question is do we want plugin authors to be concerned about wall-clock time or synchronised tick times? In most cases, I imagine a plugin will want something to happen for a specific period of time, rather than ticks which is already understood to be highly variable - in which case this PR is a good thing. Specifying in ticks, on the other hand, allows for easy synchronisation across tasks when bad performance hits. You might want to process every tick or two, or after x ticks when you want a guarantee that something has developed, or on every entity update.

We need to choose what we really want our API to work with, on wallclock times or ticks. Or both. I don't see the removal of ticks from the API as a bad thing, but we must make sure that what was possible is possible after - and we should not pretend that converting between ticks and time makes sense, it does not.

simon816:
I agree with @dualspiral.
Ticks and durations are fundamentally different concepts that should not be mixed.

As an example, you've changed `Keys.MAX_BURN_TIME` to be of type `java.time.Duration`. There is no way to properly implement this without stripping out massive parts of the `TileEntityFurnace` code to support wall clock time.
e.g. If I set the value to "10 seconds" then I expect that the fuel lasts 10 seconds. If normal TPS is 20 but the server is overloaded and runs at 10 TPS, I still want it to finish after 10 seconds.
From a behavioural point of view, the "fuel efficiency" therefore changes depending on server load - I only get 50% efficiency at 10 TPS. Now, if the entire game engine somehow used wall-clock time then it would still be consistent (i.e. `10 seconds = x fuel usage units`, regardless of TPS speed). But this fundamentally changes how the engine works, good luck with that.

What I described there is how I would expect a duration-based API to behave, this is how `SyncScheduler` works when given a duration (`delay(long delay, TimeUnit unit)`). `SyncScheduler` will run the task on the next tick where the wall-clock time equals (or exceeds) the given duration.
The scheduler API provides `delayTicks` and `intervalTicks` because ticks are not the same concept as duration.

The only feasible implementation of `MAX_BURN_TIME` would be to convert the duration into ticks using some sort of multiplier, but "10 seconds is actually 20 because the server is overloaded" is not how real-world wall clock time works (thus ticks are a different concept).

ryantheleach:
When I was developing game mechanic plugins, I was nearly always interested in counting ticks, not wall time.

Even when teleporting players, and lying to the players for the sake of communication, "You will be teleported in 5 seconds" The goal of that cooldown, is to make sure they are not in danger or being attacked by another creature or player.

When the tick rate suffers, gameplay turns into slow motion, which means I want my plugins delays delayed as well so that it's repeatable, no advantage is gained, and so that custom circuits/contraptions can be relied upon to have reliable timing in redstone contraptions.

Whilst the Wall-Clock mixin can have a positive player experience for newer players, hiding some aspects of performance issues, it absolutely wrecks redstone contraptions that rely upon those timings.

There are **some** circumstances when wall time is appropriate, but 9/10 for plugins I've made in the past I've wanted them to slow when the server does. Especially if they themselves are the cause of the performance issues, and violates player expectations of durations.

pie-flavor:
+1 against this change. `Duration`s should be for time independent of ticks, and tick counts should be for ticks independent of time, and there should absolutely not be an attempt to unify them because they are fundamentally different constructs. `@TickBasedByDefault` is a poor way of putting it; it's not just that it's tick-*based*, it's that you're suddenly treating a `Duration` like another way of counting tick increments. Annotating a type is for what kind of values it accepts or produces; if you want a type used in a totally different way, use a different type. If the methods need to be unified so there's only one way of specifying durations, then intent should be specified by the *user*, not by the API; e.g. a `ServerDuration` parameter type, where you can pass `FlexibleDuration` which is tick-based or `ConcreteDuration` which is time-based.

Cybermaxke:
I was thinking about providing a custom "duration" class to represent ticks/time, `Duration` is also not as flexible as I originally thought. Properly naming this new duration is currently a problem, maybe `ServerDuration` is an option (`EngineDuration` to target client and server), but I am not sure about it. The new duration would hold the `TemporalUnit` which can be tick based or time based, and the duration value based on the unit.

Providing a time based duration to a function that still uses ticks could convert that time based value to an estimated tick value based on the server performance.

ItsDoot:
+1 for `EngineDuration`

pie-flavor:
`TemporalUnit` being tick-based or time-based is again the wrong way of thinking about it. The time API is meant to be used in terms of increments of time, not in terms of increments of arbitrarily-timed events. The reason I mentioned subclassing is for that very reason, so that you can express something both in ticks and in time periods. A 50ms `TemporalUnit` *would* be useful, I've used them myself, but not as a tick increment but as a tick-sized time increment.

Pulverizer:
I tried reading all this. As far as I can tell you are trying to remove the usage of Ticks and that it may mean plugin developers can't use a variable time frame to time actions in their plugins.

I am on rewriting my port of the Movecraft plugin. The reason why the plugin uses Ticks, a variable time frame, over a set time frame, like Seconds, is due to lag.

Movecraft is very performance intensive. By using Ticks, I can still synchronise plugin actions without putting additional strain on the server.

I don't know if I explained that very well, but I hope it helps.

Cybermaxke:
Will be revisited in https://github.com/SpongePowered/SpongeAPI/pull/2107

PR #2008, opened 2019-05-18
Channel API Improvements + Client login phase

Cybermaxke:
A few major additions are in this PR, plugin channel support for the login phase and request/response based message registration. The request/response system was originally added for the login phase but will be expanded to the normal ones as well. 

- `RawDataChannel` - Only during play phase.
- `RawLoginDataChannel` - Only during login phase.
- `MessageChannel` - Both login and play phases.`sendToServer` is limited to the play phase.

Handlers are now attached to the `MessageBinding` which is returned by the register method. The reduces the amount of method parameters that were used previously.


src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
```diff
@@ -84,11 +87,39 @@
...
+     * can be terminated by calling {@link ClientConnection#disconnect(Text)}.</p>
+     */
+    interface Handshake extends ClientConnectionEvent {
+
+        /**
+         * Gets the {@link ClientConnection}.
+         *
+         * @return The client connection
+         */
+        ClientConnection getConnection();
```
Zidane: If we're a ClientConnectionEvent, should this not be in the root event?

src/main/java/org/spongepowered/api/network/message/MessageChannel.java
```diff
@@ -0,0 +1,185 @@
...
+import org.spongepowered.api.network.ClientConnection;
+import org.spongepowered.api.network.NoResponseMessageException;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Represents a channel binding that sends and receives messages.
+ */
+public interface MessageChannel extends ChannelBinding {
```
Zidane: This will confuse the crap out of people thinking of Text MessageChannels.

src/main/java/org/spongepowered/api/network/message/RequestMessageHandler.java
```diff
@@ -0,0 +1,56 @@
...
+ * @param <R> The response message type
+ */
+public interface RequestMessageHandler<M extends RequestMessage<R>, R extends ResponseMessage> {
+
+    /**
+     * Handles the {@link RequestMessage} which was send by a specific
+     * {@link RemoteConnection}. A proper {@link ResponseMessage} should
+     * be answered with.
+     *
+     * <p>If this handlers fails with an {@link Exception}, then
```
Zidane: then the other Endpoint...

src/main/java/org/spongepowered/api/network/message/TransactionalMessageBinding.java
```diff
@@ -0,0 +1,141 @@
...
+import org.spongepowered.api.network.ClientConnection;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A message binding that represents a request/response pair. One endpoint of a connection
+ * can send a {@link RequestMessage} and the other endpoint will normally respond with a
+ * {@link ResponseMessage}.
+ *
+ * <p>Request/response pairs share a opcode, a additional byte will specify whether it's a
```
Zidane: share an opcode

src/main/java/org/spongepowered/api/network/message/TransactionalMessageBinding.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * <p>Binding a response {@link MessageHandler} is completely optional. It is also
+     * possible to append a handler to the {@link CompletableFuture} returned by
+     * {@link MessageChannel#sendTo(ClientConnection, RequestMessage)}.</p>
+     *
+     * @param responseHandleSide The side the response message should be handled on
+     * @param responseHandler The handler of the response message
+     * @return This message binding, for chaining
+     */
+    TransactionalMessageBinding<M, R> addReponseHandler(Platform.Type responseHandleSide, MessageHandler<R> responseHandler);
```
Zidane: Response

Fix the below ones.

src/main/java/org/spongepowered/api/network/raw/RawDataListener.java
```diff
@@ -22,9 +22,11 @@
...
  */
-package org.spongepowered.api.network;
+package org.spongepowered.api.network.raw;
 
 import org.spongepowered.api.Platform;
+import org.spongepowered.api.network.ChannelBuf;
+import org.spongepowered.api.network.ClientConnection;
 
 /**
  * Represents a listener for data being sent to a raw channel.
```
Zidane: Rename this class to RawDataHandler?

src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
Cybermaxke: I can move it, but it was originally added to the subclasses.

src/main/java/org/spongepowered/api/network/message/MessageChannel.java
Cybermaxke: Ah yes, I will change it to `NetworkMessageChannel`, and maybe rename `Message`, etc. as well?

src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
```diff
@@ -160,7 +179,23 @@
...
+         */
+        void setLocation(Location location);
+
+        /**
+         * Sets the rotation the {@link User} will have
+         * at login.
+         *
+         * @param rotation The rotation
+         */
+        void setRotation(Vector3d rotation);
```
Zidane: For the other events like this, I didn't separate out rotation. It is easy enough to construct a transform to set it. I don't want to start adding all the methods to set the stuff on transform.

src/main/java/org/spongepowered/api/network/ClientConnection.java
```diff
@@ -0,0 +1,55 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.network;
+
+import org.spongepowered.api.profile.GameProfile;
+import org.spongepowered.api.text.Text;
+
+/**
+ * Represents a connection between a minecraft client and the server.
```
Zidane: Can remove the, reads better.

src/main/java/org/spongepowered/api/network/PlayerConnection.java
```diff
@@ -27,9 +27,10 @@
 import org.spongepowered.api.entity.living.player.Player;
 
 /**
- * Represents a connection of a client to the server.
+ * Represents an connection of a client to the server where
```
Zidane: a connection

src/main/java/org/spongepowered/api/network/packet/PacketIOException.java
```diff
@@ -0,0 +1,50 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.network.packet;
+
+import org.spongepowered.api.network.ChannelException;
+
+/**
+ * Represents a {@link Exception} which is thrown when a
```
Zidane: an

src/main/java/org/spongepowered/api/event/network/ClientConnectionEvent.java
```diff
@@ -160,7 +179,23 @@
          * @param world The world
          * @param transform The transform
          */
-        void setLoginPosition(World world, Transform transform);
+        void setLocation(World world, Transform transform);
```
Zidane: Per my comment below, call this setLoginPosition. Or come up with a better combined method name. If you do, correct the other events.

src/main/java/org/spongepowered/api/network/packet/Packet.java
```diff
@@ -22,28 +22,29 @@
...
+
+import org.spongepowered.api.network.ChannelBuf;
 
 /**
- * A message transmitted over the connection of a client and a server.
+ * A packet transmitted over the connection of a client and a server.
  *
- * <p>Note to plugin implementations: This must have a publicly accessible
- * no-args constructor.</p>
+ * <p>Note to plugin implementations: This must have a no-args constructor.</p>
```
simon816: Why remove `publicly accessible` here?

src/main/java/org/spongepowered/api/network/packet/PacketChannel.java
```diff
@@ -0,0 +1,193 @@
...
+     */
+    void sendTo(ClientConnection connection, Packet packet);
+
+    /**
+     * Sends the {@link Packet} to the player using this channel. This method returns a
+     * {@link CompletableFuture} that will be accepted once the response is received.
+     *
+     * <p>The {@link CompletableFuture} may fail exceptionally by a
+     * {@link NoResponseException} if there wasn't a valid response
+     * received for the given request.</p>
```
simon816: How is this determined? Is there some kind of timeout?

src/main/java/org/spongepowered/api/network/ChannelRegistrar.java
```diff
@@ -38,91 +38,37 @@
...
+     * Gets or creates a {@link PacketChannel} by the given channel key. If the
+     * channel exists and it matches the given channel type, it is returned. If
+     * the channel doesn't match a {@link IllegalStateException} is thrown.
+     * Otherwise, a new channel is created.
      *
-     * @param plugin The plugin to register the channel if it doesn't exist
-     * @param channelKey The channel id
+     * @param channelKey The channel key
      * @return A new or existing indexed message channel binding
      * @throws IllegalStateException if the existing channel is not an
```
simon816: javadocs need updating

src/main/java/org/spongepowered/api/network/raw/RawDataChannel.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param handler The handler
+     */
+    void addHandler(Platform.Type side, RawDataHandler handler);
+
+    /**
+     * Removes the handler from handling data.
+     *
+     * @param handler The handler
+     */
+    void removeListener(RawDataHandler handler);
```
simon816: `removeHandler`

src/main/java/org/spongepowered/api/network/packet/Packet.java
XakepSDK: Why empty constructor? Maybe inject it via guice?

Cybermaxke: There must be a empty constructor, so the fields can be read from a `ChannelBuf.` And I removed the requirement to make it public, because you can instantiate classes with private constructors without any (or minimal) performance penalty (e.g. with https://github.com/LanternPowered/Lmbda/).

src/main/java/org/spongepowered/api/network/packet/PacketChannel.java
Cybermaxke: Timeout or a "no response" packet.

src/main/java/org/spongepowered/api/network/packet/Packet.java
XakepSDK: Then maybe separate packet serializer and packet data?

Cybermaxke: Possibly, I could provide a `Codec` like in lantern, but you will end up with a lot more classes. I just kept it like this for simplicity, and because it was like this before.
@Zidane Thoughts?

Zidane: I want to keep it together, I don't see us needing to do Codec.

If we wrote the server, I'd reconsider.

Zidane:
@Cybermaxke Need this updated.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterChannelEvent.java
```diff
@@ -0,0 +1,52 @@
...
+ * Lifecycle event to indicate when network channels should be created and registered.
+ */
+public interface RegisterChannelEvent extends LifecycleEvent {
+
+    /**
+     * Gets the {@link ChannelRegistry}.
+     *
+     * @return The channel registry
+     */
+    ChannelRegistry getRegistry();
```
Zidane: For what purpose would you expose the registry at this time?

The only goal of this event is registering a channel. If they want to query for another channel, best do that in LoadedGameEvent.

src/main/java/org/spongepowered/api/network/ClientPlayerConnection.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface ClientPlayerConnection extends PlayerConnection, ClientSideConnection {
+
+    /**
+     * Gets the associated {@link ClientPlayer player} for this connection.
+     *
+     * @return The associated player
+     */
+    @Override
+    ClientPlayer getPlayer();
```
i509VCB: Shouldn't this return a `LocalPlayer` since you can't get the connection of a remote player?

Zidane: @Cybermaxke That is a good point, LocalPlayer is "me" in Client engine.

ClientPlayer -> LocalPlayer/RemotePlayer

Issue #1966, opened 2019-02-15
Rework of the Metrics API (Missing undecided config value)


BrainStone:
As per the discussion on the Discord, I'm writing this down as an issue/feature request.

Per older and newer discussion it was initially agreed upon that Sponge provided an API for plugins to check if they may send metrics. And that they are supposed to ask for consent. The API was added in a very basic sense. The issue being the plugins only knowing if metrics are enabled or not. Meaning that even if an admin purposely disabled them, they would get asked every time, if a plugin ever decided to do that notification.

That's the core issue. Whie discussing it several suggestions came up to streamline the process and provide a nice and consitent UI to the user/admin.

I propose the following:

- Let's have the following global stats options:
  - Undecided (admins get asked to decide) (`undecided`)
  - Enabled by default (`true` or `enabled`)
  - Disabled by default (`false` or `disabled`)
  - Please ask individually (`ask`)
- Every individual plugin gets these options:
  - Undecided. If global is set to enabled or disabled, this defaults to that value. If the global setting is on ask, the admins get individually asked for each plugin checking if it's allowed to send. (`undecided`)
  - Enabled  (`true` or `enabled`)
  - Disabled (`false` or `disabled`)

This is to allow people to give general consent or dissent, for the ones that really don't want to be bothered by individual prompts and also to give those who want it, full controll out of the box.  
This natually needs commands to allow changing all these settings directly, so they can be used in clickable links/texts.

Now this shouldn't be too hard to implement.

In addition I had a few other suggestions to improve usability:

- Move all the metrics stuff into it's own config, so it can be easily found. I find that the metrics settings are super hidden inside the `global.conf` and spent quite a while looking for them. Additionally it would make it easier to migrate any existing settings, since they will definately need to change if they are supposed to be kept, as the old system defaults to `false`, which would mean that when the new system gets in place, everything is still disabled, which would be a considerable loss for plugin devs.
- Actually implement the prompts in Sponge itself. This has quite a few advantages:
  - You make it a lot easier for plugin and stats devs to work with the restrictions. Like that basically the same stuff doesn't need to be implemented a billion times over
  - It would make sure the messages and prompts are standardized and consitent. Having every plugin have their own style and timing can get super annoying
  - If there ever will be a rivaling metrics system to bStats (or some dev makes their own custom) the user doesn't get two messages for the global setting
  - It might even be possible to prevent plugins executing the commands on behalf of users this way, as it would be all Sponge internal.

[...]

PR #2018, opened 2019-06-20
Introduce Metrics Collection States

Grinch:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2300) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/411) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/2824)

  - Deprecate `MetricsConfigManager#areMetricsEnabled(PluginContainer)`
  - Deprecate `MetricsConfigManager#areMetricsEnabled(Object)`
  - Add `MetricsConfigManager#getGlobalCollectionState`
  - Add `MetricsConfigManager#getCollectionState(PluginContainer)`

### DIFF
--- a/src/main/java/org/spongepowered/api/util/metric/MetricsConfigManager.java
+++ b/src/main/java/org/spongepowered/api/util/metric/MetricsConfigManager.java
@@ -26,6 +26,7 @@ package org.spongepowered.api.util.metric;
 
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.plugin.PluginContainer;
+import org.spongepowered.api.util.Tristate;
 
 /**
  * Provides information about whether a server has granted permission for
@@ -48,6 +49,7 @@ public interface MetricsConfigManager {
      * @return true if metrics gathering plugins have permission to gather
      *         data about this plugin
      */
+    @Deprecated
     boolean areMetricsEnabled(final PluginContainer container);
 
     /**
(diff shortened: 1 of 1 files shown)


ST-DDT (commented):
Not sure about the tristates.
This complicates the logic for plugin authors as they dont know what to do if it returns undefined.

Grinch:
> Not sure about the tristates.
> This complicates the logic for plugin authors as they dont know what to do if it returns undefined.

I can see why it would be confusing, how would you recommend going about it? Undefined is effectively false but without expressed intent on it being false.

ST-DDT:
What is the purpose of changing the API?
Why should plugin devs care whether it is implicitly false or explicitly false?

Grinch:
> What is the purpose of changing the API?
> Why should plugin devs care whether it is implicitly false or explicitly false?

For what purpose would you hide that information from them? One defines intent not to allow stat collection while the other is just "They haven't specifically decided but it's still disabled by default". You still haven't told me how you would recommend going about it.

We could stick with only exposing a yay or nay to the plugin developers but I don't see the harm in them knowing a bit more.

ST-DDT:
> > What is the purpose of changing the API?
> > Why should plugin devs care whether it is implicitly false or explicitly false?
>
> For what purpose would you hide that information from them? One defines intent not to allow stat collection while the other is just "They haven't specifically decided but it's still disabled by default". You still haven't told me how you would recommend going about it.
> 
> We could stick with only exposing a yay or nay to the plugin developers but I don't see the harm in them knowing a bit more.

My recommendation is: don't change the API.
For the following (related) reasons:
* Using a Tristate does not implicate that false is the default.
* It requires additional documentation that clearly explains what the states refer to
  * Otherwise you might be mislead to "if the plugin is not specified, you have to ask the server, if the server is not specified you can decide on your own"
* Checking for it is more complicated than a boolean
* I don't see a benefit in exposing it
  * Adding functions without real benefit is harmful in the long run.

The sole purpose of exposing this I can think of would be adding a plea to enable metrics if it is not defined. However if a plugin author actually pleas to enable metrics for their plugin, then I assume that they would do so even if metrics are explicitly disabled for them (and we shouldn't encourage plugin devs to do that either way).

ryantheleach:
> However if a plugin author actually pleas to enable metrics for their plugin, then I assume that they would do so even if metrics are explicitly disabled for them (and we shouldn't encourage plugin devs to do that either way).

On the contrary, enabling plugin authors to mute their plugins willingly, will better their reputation to users. Once that happens, plugins that continue to beg after disabling will be looked down upon the community.

Without revealing that information, if plugin developers want to ask users to enable, they are forced to do so even when the user has explicitly opted out, or be forced to implement their own muting mechanism for each plugin.

+1

src/main/java/org/spongepowered/api/util/metric/MetricsConfigManager.java
```diff
@@ -66,9 +68,23 @@
...
      *         data about this plugin
      */
+    @Deprecated
     default boolean areMetricsEnabled(final Object plugin) throws IllegalArgumentException {
         return this.areMetricsEnabled(Sponge.getPluginManager().fromInstance(plugin)
                 .orElseThrow(() -> new IllegalArgumentException("The supplied object is not a plugin object.")));
     }
 
+    /**
+     * Gets the current <em>global</em> state of collection.
```
Zidane: There should be documentation what each state means.

Grinch: Requested changes have been made.

BrainStone:
@ST-DDT also have a look at #1966  
@Grinch in case you haven't check that issue, I think would be great if you could check it too.

PR #2019, opened 2019-06-22
Add Acceleration Data

### LINKED ISSUE #1981: Unable to adjust the Entity "power" tag for Fireballs.
Description of the tag by the Official Minecraft Wiki:

> power: List of 3 doubles that adds to direction every tick. Act as the acceleration.
   ~ [Official Minecraft Wiki - Ghast](https://minecraft.gamepedia.com/Ghast#Ghast_fireball)

Just from reading this description you can see why, when altering the course of a Fireball entity, you would need to be able to adjust this tag.

Grinch:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2303)

Resolves #1981 

### DIFF
--- a/src/main/java/org/spongepowered/api/data/key/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/key/Keys.java
@@ -108,6 +108,13 @@ public final class Keys {
     public static final Key<Value<Double>> ABSORPTION = DummyObjectProvider.createExtendedFor(Key.class, "ABSORPTION");
 
     /**
+     * Represents the {@link Key} for the acceleration of a {@link Fireball}.
+     *
+     * @see AccelerationData#acceleration()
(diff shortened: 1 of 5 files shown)


liach:
:tada: Finally we don't need to register an event listener to constantly change the velocity of the fireball like in the cookbook!

Pulverizer:
This variable is also needed for WitherSkull.

Grinch:
@Pulverizer Updated based on your feedback.

NeumimTo:
may this affect any entity?

Pulverizer:
There may be some entities we've missed. But in vanilla, it only affects certain entities.

Maybe it should be discussed if we want to modify Minecraft to have it for all entities.

PR #2022, opened 2019-07-11
Fix PlayerCommandElement autocompletion and enhance CommandArgs

ImMorpheus:
The issue can be replicated using the skull testplugin.

user "sponge" join the game

before: `/skullme sponge bukk<tab>` -> `/skullme sponge sponge`
with this PR: `/skullme sponge bukk<tab>` -> `/skullme sponge bukk`

Also, add `size` and `get` to `CommandArgs` to help plugin provided command elements

Note: the two added methods can be replicated using `getAll().size()` and `getAll().get(index)`, however `getAll()` has quite a bit of overhead.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -1433,6 +1433,19 @@ public final class GenericArguments {
         }
 
         @Override
+        public List<String> complete(CommandSource src, CommandArgs args, CommandContext context) {
+            if (args.size() == 0) {
+                return Collections.emptyList();
+            }
+            final String last = args.get(args.size() - 1);
+
+            return Sponge.getServer().getOnlinePlayers().stream()
+                    .filter(p -> p.getName().startsWith(last))
+                    .map(User::getName)
+                    .collect(Collectors.toList());
+        }
+
+        @Override
         protected Iterable<String> getChoices(CommandSource source) {
             return Sponge.getGame().getServer().getOnlinePlayers().stream()
                 .map(input -> input == null ? null : input.getName())
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -1433,16 +1433,6 @@ public final class GenericArguments {
         }
 
         @Override
-        public List<String> complete(CommandSource src, CommandArgs args, CommandContext context) {
-            final String last = (args.size() == 0) ? "" : args.get(args.size() - 1);
-
-            return Sponge.getServer().getOnlinePlayers().stream()
-                    .filter(p -> p.getName().startsWith(last))
-                    .map(User::getName)
-                    .collect(Collectors.toList());
-        }
-
-        @Override
         protected Iterable<String> getChoices(CommandSource source) {
             return Sponge.getGame().getServer().getOnlinePlayers().stream()
                 .map(input -> input == null ? null : input.getName())
--- a/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
+++ b/src/main/java/org/spongepowered/api/command/args/GenericArguments.java
@@ -1434,10 +1434,7 @@ public final class GenericArguments {
 
         @Override
         public List<String> complete(CommandSource src, CommandArgs args, CommandContext context) {
-            if (args.size() == 0) {
-                return Collections.emptyList();
-            }
-            final String last = args.get(args.size() - 1);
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/command/args/GenericArguments.java
```diff
@@ -1432,6 +1432,16 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
             }
         }
 
+        @Override
+        public List<String> complete(CommandSource src, CommandArgs args, CommandContext context) {
```
dualspiral: No. You're ~~completely~~ changing the completion of this element - ~~this removes the ability to use regex to complete, _and_~~ you are removing the ability to complete a selector.

See 

https://github.com/SpongePowered/SpongeAPI/blob/a3e2b07a0de0877037eb16091425ee9757649c57/src/main/java/org/spongepowered/api/command/args/SelectorCommandElement.java#L61-L71

and

https://github.com/SpongePowered/SpongeAPI/blob/a3e2b07a0de0877037eb16091425ee9757649c57/src/main/java/org/spongepowered/api/command/args/PatternMatchingCommandElement.java#L76-L83

for what this should be doing.

Honestly, this is masking a problem I think. If it's a problem with the player element then I'm pretty sure it'll be a problem elsewhere and there might be a bug with the completion code somewhere. Overriding the complete method in the player element is _not_ the way to do it.

**EDIT:** This is a completion, so the regex bit doesn't apply. Shouldn't have looked at this as I was going to bed. The selectors bit does, however.

dualspiral (changes requested):

dualspiral:
[This was the problem](https://github.com/SpongePowered/SpongeAPI/commit/240a43e70e2445f5169636a6f84de1cf52c4a0da). The `seq` element handles the completion and parsing elements before it, not the `CommandSpec`, but in the case of only one parameter being set on a `CommandSpec`, the spec didn't bother wrapping the element and so the system just ran the complete method with no regard to the current number of elements. This would have affected everything, not just the player element.

I see no issue with adding the other methods you propose, so I didn't want to close the PR. However, the override on the complete method for the `PlayerCommandElement` should be removed.

ImMorpheus:
Thank you for the explanation.

I missed the fact you could autocomplete selectors (explained in `GenericArguments`, my bad).
Yeah, overriding the complete method in the player element is wrong, this is also happening with UserCommandElement.. It's a broader bug with autocompletion.


> This was the problem. The seq element handles the completion and parsing elements before it, not the CommandSpec, but in the case of only one parameter being set on a CommandSpec, the spec didn't bother wrapping the element and so the system just ran the complete method with no regard to the current number of elements. 

I have to disagree. That commmit has just broken autocompletion for command elements returning multiple values.
https://github.com/SpongePowered/SpongeAPI/blob/240a43e70e2445f5169636a6f84de1cf52c4a0da/src/main/java/org/spongepowered/api/command/args/GenericArguments.java#L176

If a command element may return multiple values, I expect completition to work if I provide multiple arguments.

This is after you commit:

user "sponge" join the game
`/skullme sponge bukk<tab>` -> `/skullme sponge bukk` (which is good)
`/skullme sponge spon<tab>` -> `/skullme sponge spon` (This should not happen)
`/skullme sponge spon<tab>` -> `/skullme sponge sponge` (Expected behaviour)

dualspiral:
> If a command element may return multiple values, I expect completition to work if I provide multiple arguments.

You're misunderstanding how these elements work. The singular element consumes one argument, but may return multiple results. For example, if you have players `sponge1` and `sponge2`, `spon` will return both of these because it's pattern matching. If you have a selector (`@a`, for example), that might return multiple players. That's what that javadoc comment means - for one input you can get multiple outputs.

For the behaviour you want, you need to use `GenericArguments.allOf(...)`, or `GenericArguments.repeated(...)` - these indicate to the command parser that there may be more than one element - and that will work as intended.

> I have to disagree. That commmit has just broken autocompletion for command elements returning multiple values.

Ignoring the above, given that it wasn't working correctly in the first place, my commit hasn't broken it. Nor has it broken existing behaviour when multiple elements are specified. So to tell me that I broke it is downright wrong. If there was a player called `sponge` and `bukkit`, you still would have ended up with `/skullme sponge sponge` because the completion relied upon the first element _only_.

ImMorpheus:
> You're misunderstanding how these elements work. The singular element consumes one argument, but may return multiple results. For example, if you have players sponge1 and sponge2, spon will return both of these because it's pattern matching. If you have a selector (@a, for example), that might return multiple players. That's what that javadoc comment means - for one input you can get multiple outputs.

This cleared up everything.

PR #2028, opened 2019-08-04
Add SoundEvents

clienthax:
Implementation: https://github.com/SpongePowered/SpongeCommon/pull/2335

Allows the canceling of Broadcast sounds, eg Wither/Dragon death, without causing event spam.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/sound/PlaySoundEvent.java
+++ b/src/main/java/org/spongepowered/api/event/sound/PlaySoundEvent.java
@@ -1,0 +1,113 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.sound;
+
+import org.spongepowered.api.block.tileentity.Jukebox;
+import org.spongepowered.api.effect.sound.SoundCategory;
+import org.spongepowered.api.effect.sound.SoundType;
+import org.spongepowered.api.effect.sound.record.RecordType;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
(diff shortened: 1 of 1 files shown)


ImMorpheus:
Not sure about this, this event looks too specific.

I'd prefer something like 

```
SoundEvent
SoundEvent.Broadcast
SoundEvent.Record
```

mosemister (changes requested):
I would recommend having a GetSound() function that returns the SoundType as well as a getCategory() function that returns the SoundCategory of the sound. 

This saves plugin developers from looking up the soundType and/or category type of the sound. Makes it like the other events such as all entity deaths allow you to gain the Entity involved. Not just the location at which the entity was. 

I also agree with @imMorpheus (p.s cool name)

clienthax:
@mosemister Broadcast Sounds do not have SoundTypes linked to them, they seem to use some internal MC ids that have been stripped, see previous comments, the Entity involved is already in the causetracker from upstream.

@ImMorpheus I can get behind that idea, the only problem is we don't want to be spaming events for every sound.

ImMorpheus:
> I can get behind that idea, the only problem is we don't want to be spaming events for every sound.

Put a ShouldFire check before firing the event. That's no different that ServerTickEvent, CollideEntityEvent and other expensive events.

mosemister:
I think you missread the entity bit. It was just an example. I was expecting if an entity caused the sound to have that entity in the cause. 

If the ids are internal MC ids then maybe expose them as an catalogue type (BroadcastType as an example) just to stop confusion as I as a plugin developer who just jumps to assumptions when in the mood would attempt to get a sound type from the id using the sponge register. If its a different type then i wouldn't make that wrong assumption.

clienthax:
@mosemister There are only three broadcast sound types, personally as a plugin developer I would just  grab the Entity causing the sound from the cause and compare against that.

RenderGlobal.broadcastSound for an example.
I might just remove the ID as its practically useless as a data point.

clienthax:
@mosemister Just noticed there is a reference to these type of ids in util.Constants, I may use that for these.

clienthax:
@ImMorpheus Updated

src/main/java/org/spongepowered/api/event/sound/PlaySoundEvent.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Gets the location of the sound being played.
+     *
+     * @return the location of the sound being played
+     */
+    Location<World> getLocation();
+
+    /**
+     * Called when a call to World.playBroadcastSound is made, for global sounds
```
dualspiral: You shouldn't specify how you can call an event as you do, because it may not be the only way to do it, or other implementations of the server software might have other systems that fire this event. Instead, you could suggest it as examples, so that it should be something like:

```java
/**
 * Fired when a sound is broadcasted to all {@link Player}s on the server. 
 * <p>Examples of when this might be fired include:</p>
 *
 * <ul>
 * <li>When a Wither or Dragon dies;</li>
 * <li>When requested by a plugin via {@link World#playBroadcastSound()}</li>
 * </ul>
 */
```

However, I'm not convinced that you even need the examples.

You don't need to specify what happens when an event is cancelled - cancelling an event implicity means that you are preventing the action from happening.

src/main/java/org/spongepowered/api/event/sound/PlaySoundEvent.java
```diff
@@ -0,0 +1,76 @@
...
+
+    }
+
+    /**
+     * Called when a record is inserted into a Jukebox
+     * <p>
+     * If this event gets cancelled, it will not send the sound to the clients.
+     * </p>
+     */
+    interface Record extends PlaySoundEvent {
```
dualspiral, line 88: I don't think this should be here. A sound (in this context) and a record are two different things - a sound is usually a reaction to an event and is quick, while a record is background music as opposed to a quick "sound".

I think this should be a `PlayRecordEvent` or something similar that can be subinterfaced if necessary (start/stop, for example).

Then, the `SoundType` (or whatever) that the `PlaySoundEvent` represents can be moved to the `PlaySoundEvent` interface.

clienthax: I think it can still be here, it is a sound as far as the server and client are concerned, it's not like broadcast sounds are 'quick' either.

dualspiral: Right, yes, reviewing the server code that makes sense. I would _much_ rather have it split out _conceptually_ but looking into the MCP implementation, and crucially, Forge, that's not going to happen.

My other thoughts are general and I will put them in the main body..

src/main/java/org/spongepowered/api/event/sound/PlaySoundEvent.java
clienthax: Noted, i think the examples are better than not having examples, but it was mostly for my self reference when implementing the events.

clienthax:
Updated

clienthax:
Ready for review.

PR #2029, opened 2019-08-05
Add possibility to remove a child at an index

### LINKED ISSUE #1727: Allow deleting of a specific child of a Text.Builder
Currently, there is no way to remove a specific child of a `Text.Builder`. This is particularly bothersome when creating delimited lists, where one would normally remove the trailing delimiter after the list is built. This could be achieved by introducing a `Builder.remove(int index)` method, or by making `Builder.getChildren()` return a mutable list.
Something like
```Text.Builder textBuilder;
int i;
for (Something s : listOfSomethings) {
    textBuilder.append(Text.of(s));
    if(i < listOfSomethings.size() - 1){
         textBuilder.append(Text.of(", "));
    }
}
```
could be simplified to
```Text.Builder textBuilder;
for (Something s : listOfSomethings) {
    textBuilder.append(Text.of(s))
            .append(Text.of(", "));
}
textBuilder.remove(textBuilder.getChildren().size() - 1);
```

Stupremee:
A user can now remove a child at a specific index.

This pull request fixes #1727

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -766,6 +766,30 @@ public abstract class Text implements TextRepresentable, DataSerializable, Compa
         }
 
         /**
+         * Removes the last child in this builder.
+         *
+         * @return This text builder
+         * @see #remove(int)
+         */
+        public Builder removeLastChild() {
+            int lastIndex = this.children.size() - 1;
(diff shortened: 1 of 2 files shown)


gabizou:
Can you squash the commits? Once done, can merge.

Stupremee:
You can squash the commits when you merge.

PR #2030, opened 2019-08-08
Give permissions integer values to allow weighting

zml2008:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/2340) | no forge changes

Based on discussions with @lucko and @bloodmc, this gives more flexibility with how different levels of permission take priority over each other.

PEX has had weighted permissions for quite a while now, so hopefully it makes sense to bring them into Sponge as well.

I've kept the old methods around to avoid breaking changes -- the deprecated methods and classes can go away for API 8.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
+++ b/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.service.permission;
 
+import com.google.common.collect.Maps;
 import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.text.Text;
 
@@ -160,7 +161,9 @@ public interface PermissionDescription {
      * @return A reference to any subject known to have this permission
      *         set, and the value this permission is set to
      * @see SubjectCollection#getAllWithPermission(String)
+     * @deprecated See {@link #findAssignedSubjectValues(String)}
      */
+    @Deprecated
     CompletableFuture<Map<SubjectReference, Boolean>> findAssignedSubjects(String collectionIdentifier);
 
     /**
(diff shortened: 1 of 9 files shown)


bloodmc:
@kashike Does this PR look good to you?

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -233,8 +279,34 @@
...
+         * {@link #register()}.</p>
+         *
+         * @param role The role-template to assign the permission to. See
+         *             constants in {@link PermissionDescription} for common
+         *             roles (not exhaustive).
+         * @param value The value to to assign
+         * @return This builder for chaining
+         */
+        default Builder assign(String role, int value) {
+            return assign(role, value > 0);
```
kashike, line 307: `assign(String, boolean)` is deprecated - why are you calling it?

zml2008: it's a fallback so that calling added api methods with a permissions plugin that hasn't been updated to specifically implement them won't cause breakage. These default implementations (and deprecated methods) would then be removed for API 8.

src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java
```diff
@@ -233,8 +279,34 @@
          * @param value The value to to assign
          * @return This builder for chaining
          */
+        @Deprecated
         Builder assign(String role, boolean value);
```
ST-DDT, line 283: I'm not sure whether the description actually needs integer values as it is only intended as a description for humans whether a given role should have this permission or not.

ST-DDT: https://github.com/zml2008/SpongeAPI/blob/fa812713b34c8d8b446308b93cc22e8b73718777/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java#L43

ST-DDT:
Dont forget to create a PR for our docs repo explaining the new API properly.

ST-DDT:
The javadocs on the new methods dont explain what use/impact the integer values have, do they? How is 5 any different from 3?

zml2008:
Closing in favor of #2065

Issue #1978, opened 2019-03-06
Incorrect behavior of flags with optional value


ya-evgeniy:
I registered a command with a flag that takes an optional value.

`CommandElement flagValue = GenericArguments.bool(Text.of("flag_value"));`
`CommandElement optionalFlagValue = GenericArguments.optional(flagValue);`
`CommandFlags.Builder flags = GenericArguments.flags();`
`flags.valueFlag(optionalFlagValue, "-flag");`
`CommandElement commandElement = flags.buildWith(GenericArguments.none());`

As well as a handler that displays data on the flag. ps: Just in case, I tried to use "-".

`Text.of("has flag_value ", args.hasAny("flag_value") )`
`Text.of("has -flag_value ", args.hasAny("-flag_value"))`
`Text.of("has flag ", args.hasAny("flag"))`
`Text.of("has -flag ", args.hasAny("-flag"))`
`Text.of("get flag_value ", args.getOne("flag_value"))`
`Text.of("get -flag_value ", args.getOne("-flag_value"))`
`Text.of("get flag ", args.getOne("flag"))`
`Text.of("get -flag ", args.getOne("-flag"))`

For command "/test_flags --flag", the result can be seen in the screenshot below.
![image](https://user-images.githubusercontent.com/17035475/53866085-c1147880-4001-11e9-86b5-c4d049604ea1.png)

Expected behavior:
By my logic, `args.hasAny("flag")` should have returned true, and `args.getOne("flag_value")` should return `Optional.empty()`. If I am mistaken in some way, then correct me.

Server: `spongevanilla-1.12.2-7.1.5`

p.s. The text is translated by a translator.

WillBAnders:
The way flags work are a bit different than you expect. Flags are not themselves `CommandElements`, but each flag uses an element to handle it's parsing. A value flag is one where you provide this element so it can parse additional values after the flag. The `"flag"` parameter is only used to identify the flag and is not part of this element, thus it is not included in the final context.

Additionally, flags should not be combined with optional elements because it adds multiple layers of optional. Flags are inheritably optional, so you should stick to using `--flag value` for present values and `` (empty) when the flag is not defined. As such, of the calls given only `hasKey("flag_value")` and `getOne("flag_value")` will work.

The idea of including a way to determine what flags were specified themselves is definitely meritable and worth discussion for API 8.

dualspiral:
I wondered how I'd missed this in the first place, but then realised I was on holiday at the time.

[...]

PR #2033, opened 2019-08-11
Add defined flags to CommandContext

WillBAnders:
### Overview
Based on discussion in #1978, being able to differentiate between an absent flag and a flag with an absent optional value could be useful for handling default values. Additionally, this is extremely helpful when working with complex flag systems where the same value could be set through multiple flags (used in validation). The PR is intended to provide a base implementation and serve as discussion for whether this is worth adding to the API.

Additionally, given the new `hasFlag` method it would make sense to have this be the recommend way for checking non-value flags.

### Implementation

This adds methods to `CommandContext` for defining and checking flags, implemented with a set. Methods that accept a `Text` are also provided to match existing `putArg` methods, but this may not be necessary.

To handle parsing, a new `FlagElement` has been added that ensures flags are added to the context. The key used is the first flag found, same as for the element used previously. Notably, this replaces the use case for the `MarkTrue` element used previously, so this could potentially be deprecated as well.

Additionally, I made the decision to also define the flag for the `ACCEPT_VALUE` and `ACCEPT_NONVALUE` behaviors for unknown flags. This is probably worth discussing as well.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/args/CommandContext.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandContext.java
@@ -282,7 +282,7 @@ public final class CommandContext {
      *      {@link CommandContext}
      */
     public Snapshot createSnapshot() {
-        return new Snapshot(this.parsedArgs);
+        return new Snapshot(this.parsedArgs, this.definedFlags);
     }
 
     /**
@@ -294,6 +294,8 @@ public final class CommandContext {
     public void applySnapshot(Snapshot snapshot) {
         this.parsedArgs.clear();
         this.parsedArgs.putAll(snapshot.args);
+        this.definedFlags.clear();
+        this.definedFlags.addAll(snapshot.flags);
     }
 
     /**
@@ -303,9 +305,11 @@ public final class CommandContext {
     public final class Snapshot {
 
         final Multimap<String, Object> args;
+        final Set<String> flags;
 
-        Snapshot(Multimap<String, Object> args) {
+        Snapshot(Multimap<String, Object> args, Set<String> flags) {
             this.args = ArrayListMultimap.create(args);
+            this.flags = Sets.newHashSet(flags);
         }
 
     }
--- a/src/main/java/org/spongepowered/api/command/args/CommandContext.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandContext.java
@@ -29,6 +29,7 @@ import static org.spongepowered.api.util.SpongeApiTranslationHelper.t;
 
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
 import org.spongepowered.api.command.CommandException;
 import org.spongepowered.api.command.CommandSource;
 import org.spongepowered.api.text.Text;
@@ -38,6 +39,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * Context that a command is executed in.
@@ -57,12 +59,14 @@ public final class CommandContext {
     public static final String TAB_COMPLETION = "tab-complete-50456"; // Random junk afterwards so we don't accidentally conflict with other args
 
     private final Multimap<String, Object> parsedArgs;
+    private final Set<String> definedFlags;
 
     /**
      * Create a new empty CommandContext.
      */
     public CommandContext() {
         this.parsedArgs = ArrayListMultimap.create();
+        this.definedFlags = Sets.newHashSet();
     }
 
     /**
@@ -194,6 +198,24 @@ public final class CommandContext {
     }
 
     /**
+     * Defines the flag as being present when parsing this context.
+     *
+     * @param key the key for the flag defined
+     */
+    public void addFlag(String key) {
+        definedFlags.add(key);
+    }
+
+    /**
+     * Defines the flag as being present when parsing this context.
+     *
+     * @param key the key for the flag defined
+     */
+    public void addFlag(Text key) {
+        addFlag(ArgUtils.textToArgKey(key));
+    }
+
+    /**
      * Perform a permissions check, throwing an exception if the required
      * permissions are not present.
      *
@@ -225,6 +247,26 @@ public final class CommandContext {
      */
     public boolean hasAny(Text key) {
         return hasAny(ArgUtils.textToArgKey(key));
+    }
+
+    /**
+     * Returns whether the given flag has been defined in this context.
+     *
+     * @param key The key to look up
+     * @return whether the flag is defined
+     */
+    public boolean hasFlag(String key) {
+        return definedFlags.contains(key);
+    }
+
+    /**
+     * Returns whether the given flag has been defined in this context.
+     *
+     * @param key The key to look up
+     * @return whether the flag is defined
+     */
+    public boolean hasFlag(Text key) {
+        return hasFlag(ArgUtils.textToArgKey(key));
     }
 
     /**
--- a/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
@@ -340,8 +340,6 @@ public final class CommandFlags extends CommandElement {
 
         Builder() {}
 
-        private static final Function<String, CommandElement> MARK_TRUE_FUNC = input -> new FlagElement(Text.of(input), null);
-
         private Builder flag(Function<String, CommandElement> func, String... specs) {
             final List<String> availableFlags = new ArrayList<>(specs.length);
             CommandElement el = null;
@@ -388,7 +386,7 @@ public final class CommandFlags extends CommandElement {
          * @return this
          */
         public Builder flag(String... specs) {
-            return flag(MARK_TRUE_FUNC, specs);
+            return flag(input -> new FlagElement(Text.of(input), null), specs);
         }
 
         /**
--- a/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
+++ b/src/main/java/org/spongepowered/api/command/args/CommandFlags.java
@@ -504,14 +504,12 @@ public final class CommandFlags extends CommandElement {
         @Override
         public void parse(CommandSource source, CommandArgs args, CommandContext context) throws ArgumentParseException {
             String key = getUntranslatedKey();
-            if (key != null) {
-                context.addFlag(key);
-                if (valueElement != null) {
-                    valueElement.parse(source, args, context);
-                } else {
-                    context.putArg(key, true);
-                }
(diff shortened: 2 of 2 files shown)


src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -227,6 +249,26 @@ public boolean hasAny(Text key) {
         return hasAny(ArgUtils.textToArgKey(key));
     }
 
+    /**
+     * Returns whether the given flag has been defined in this context.
+     *
+     * @param key The key to look up
+     * @return whether the flag is defined
+     */
+    public boolean hasFlag(String key) {
```
ST-DDT, line 258: Do you actually need this new method?
AFIACT the hasAny method does the same (not limited to flags though).

WillBAnders: Yes, defined flags are stored in a set separate from the multimap used by arguments. This is because flags do not always have a value (or only one) and to avoid conflictions between flags and arguments with the same key. The split here was recommended by dual in the issue introducing this and I support the distinction as well.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -486,4 +490,41 @@ public CommandElement buildWith(CommandElement wrapped) {
                     this.unknownLongFlagBehavior, this.anchorFlags);
         }
     }
+
+    private static class FlagElement extends CommandElement {
```
dualspiral, line 492: I don't like this. The whole point of having the `hasFlag` method is to keep the fact a flag found and correctly parsed separate from _what_ was parsed. This kind of blends the two together again.

Remember we're dealing with the API layer too. I'd rather make the system change as little as possible.

I'm pretty sure this can be done in the `CommandFlags` class alone - when everything has parsed (and no exception thrown), then add the flag to the context.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -486,4 +490,41 @@ public CommandElement buildWith(CommandElement wrapped) {
...
+        private FlagElement(Text key, @Nullable CommandElement valueElement) {
+            super(key);
+            this.valueElement = valueElement;
+        }
+
+        @Override
+        public void parse(CommandSource source, CommandArgs args, CommandContext context) throws ArgumentParseException {
+            String key = getUntranslatedKey();
+            if (key != null) {
+                context.addFlag(key);
```
dualspiral: If the flag value fails to parse and is then taken as, say, a negative number, the flag will still be in the context when it shouldn't be, right?

src/main/java/org/spongepowered/api/command/args/CommandContext.java
```diff
@@ -193,6 +197,24 @@ public void putArg(Text key, Object value) {
         putArg(ArgUtils.textToArgKey(key), value);
     }
 
+    /**
+     * Defines the flag as being present when parsing this context.
+     *
+     * @param key the key for the flag defined
+     */
+    public void addFlag(String key) {
+        definedFlags.add(key);
```
dualspiral, line 206: Make sure to handle this when creating/applying snapshots of the context.

src/main/java/org/spongepowered/api/command/args/CommandContext.java
dualspiral: @ST-DDT See https://github.com/SpongePowered/SpongeAPI/issues/1978 for where the concern lies. There is a difference to "if a flag was specified" and "does it have a value".

dualspiral (changes requested):
I think this needs to change as suggested in the comments.

I'm still not sure about whether we need this. A well built command should make it simple for the consumer to know which value comes from where. If the changes are minimal and purely additive, then there isn't _really_ a reason to reject it outright either, sans that it might be seen as bloat.

Please note that as the API 8 command system stands, we don't have flags. That's not for lack of trying, I'm trying to figure out the best way to re-add them, but the client completion stuff is actually making it really difficult. They will likely become "anchor flags", at least at first. 

Just be aware of that.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
WillBAnders: Unfortunately it can't as is - I thought that too when I started, but after trying it I realized that the flag being parsed is not necessarily the key used for that flag either. For example, in the flags `xzvf` the key used is `x`. The flag `-v` should then be recognized as the same as flag `x`, which is what's added to the context in the same way as elements.

At the `CommandFlags` level, we don't really have this information anymore - just a map of flags to elements. The element should contain the correct key for non-value flags, but for value flags it could be anything, hence this PR and why it's can't be done this way.

One alternative here is to also send `CommandFlags` a map of flags to keys, which would allow us to look up the key. This seemed a bit too costly so I went with this approach. As far as the API goes, this wouldn't be visible as well (and it replaces the mark true element used previously).

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
WillBAnders: That's true, I mistook any failures there as throwing an exception that would simply fail but it can still be caught later on. The flag should be added afterwards. The key can't be null either, so that can be taken out as well.

src/main/java/org/spongepowered/api/command/args/CommandContext.java
WillBAnders: Good catch, I will do that.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
dualspiral: You could do it at this point and remove the addition of the key from everywhere else. `remove` is `true` when parsing succeeded.

https://github.com/SpongePowered/SpongeAPI/blob/9067212f054eee4187c623cca520f1d8c063f3f2/src/main/java/org/spongepowered/api/command/args/CommandFlags.java#L82-L84

However, if a command has specified x and v as the same, then why do they need to know that x and v have both been specified? Surely thy only need to know the intended outcome?

The more I think about this, the more I think this really isn't needed. It's not as if a command doesn't have full control over the parameters it requires. With each flag, a value is placed into the context, which will be placed against a key that the developer should be aware of, and if they duplicate the key in other elements, thats on them.

I just don't see the use case. If you need to differentiate between -x and -v, then make them two separate flags, you should not be relying on the parser to tell you that separately.

WillBAnders: So at the point of `remove` we don't have the actual flags anyways - additional work would have to be done to re-extract them. Regardless, the main problem is that we can't use the flag the user entered for this when working with flag aliases. If you added the user input to the context, then `-x` and `-v` would be separate when they should be aliases.

The additional work done with the `FlagsElement` is so aliases of a flag still add the same key to the context. It's essentially what `markTrue` did with an additional step. This is what ensures the context *doesn't* differentiate between `-x` and `-v` because they're aliases for the same flag.

dualspiral: Ok, so you want the first alias to be added to the flag set. I would still prefer this information being entirely within the command flags class.

Before you do though, the question still remains, is there a use case? I do not see one that should be solved with careful command planning. What does having a “hasFlag” method solve that you can’t solve with flags and their associated values in the command context?

WillBAnders: If it's done in `CommandFlags`, we need to also add a map for looking up the key as well, performing a lookup twice (or alternatively store a tuple instead).

This is useful when you have two flags which share the same value but are used differently. An example of this is defining a range for a random number command.

```
/rand {--lowerInclusive=<lower>| --lowerExclusive=<lower> | --upperInclusive=<upper> | --upperExclusive=<lower>}

/rand --li 1 --ui 9
/rand --li 0 --ue 10
```
The handling for this would be:
```
int lower = args.requireOne("lower");
int upper = args.requireOne("upper");
BoundType lBound = args.hasFlag("lowerInclusive") ? BoundType.CLOSED : BoundType.OPEN;
BoundType uBound = args.hasFlag("upperInclusive") ? BoundType.CLOSED : BoundType.OPEN;
Range range = Range.range(lower, lBound, upper, uBound);
```
It's a bit contrived of course, but I think for this particular example it's the cleanest way to express this command.

Also worth noting is having `args.hasFlag("key")` rather than `args.hasAny("key")`, which might be cleaner from an API perspective.

Zidane:
@dualspiral Now or never on this for API-7. If not close it.

dualspiral:
I'll test, it's my fault it's been held and I think it would be unfair to close it. Will decide based on that.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
```diff
@@ -336,7 +340,7 @@ protected Object parseValue(CommandSource source, CommandArgs args) throws Argum
 
         Builder() {}
 
-        private static final Function<String, CommandElement> MARK_TRUE_FUNC = input -> markTrue(Text.of(input));
+        private static final Function<String, CommandElement> MARK_TRUE_FUNC = input -> new FlagElement(Text.of(input), null);
```
dualspiral: The name of this should change - it no longer marks as true!

dualspiral (approved):
I think I owe you an apology @WillBAnders. After working on the new API 8 system, parsed flags are indeed separate from the context. I know I argued against it but... it made sense in the new system and so it deserves to be here. My apologies for running you around the houses.

This looks okay as is, one minor comment. If you don't want to do it, let me know and I'll get around to doing so.

src/main/java/org/spongepowered/api/command/args/CommandFlags.java
WillBAnders: `FlagElement` will still add `true` to the context so it's backwards compatible. This is only used once now so it might be best to just inline it anyways.

dualspiral: It's used multiple times, just in the same function. I think we're just better off keeping it in a field, as I say, it just needs renaming because [it doesn't necessarily mark as true](https://github.com/SpongePowered/SpongeAPI/pull/2033/files#diff-bddcca313ee050f07039ec6b24e4653eR508).

WillBAnders: `valueElement` at that point is always `null`. I don't see where other uses of `MARK_TRUE_FUNC` are, I can only find one in [`#flag(String...)`](https://github.com/SpongePowered/SpongeAPI/pull/2033/files#diff-bddcca313ee050f07039ec6b24e4653eR391).

dualspiral: Oh, yeah... I'm tired. You're right. Bin it then.

PR #2035, opened 2019-08-14
Create RegainHealthEvent for Entities

gabizou:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2350)
Deprecates HealEntityEvent that is/was impossible to implement with modifiers and sources. 

Adds RegainHealthEvent with appropriate usages and `HealingType`s.


gabizou:
Closing as we're going to be looking at an alternative option for adding this event. It'll be a sub-event of ChangeValueEvent via some generation of annotated Keys.

PR #2037, opened 2019-08-16
Fix DataView equality for arrays (#1983)

### LINKED ISSUE #1983: MemoryDataView with array doesn't equal equivalent MemoryDataView
When you have a `MemoryDataView` that contains an array and call `#equals` on with another `MemoryDataView` that contains an equivalent or even the same array, `#equals` will return false regardless.

Example:

```java
DataQuery path = DataQuery.of("path");
int[] array1 = new int[] {1, 2, 3};
int[] array2 = new int[] {1, 2, 3};

MemoryDataView dataView1 = new MemoryDataView(); // Or DataContainer.createNew();
dataView1.set(path, array1);

MemoryDataView dataView2 = new MemoryDataView();
dataView2.set(path, array1);

MemoryDataView dataView3 = new MemoryDataView();
dataView3.set(path, array2);

dataView1.equals(dataView2); // -> false
dataView1.equals(dataView3); // -> false
```

### LINKED ISSUE #3086: MemoryDataView with array doesn't equal equivalent MemoryDataView
When you have a `MemoryDataView` that contains an array and call `#equals` on with another `MemoryDataView` that contains an equivalent or even the same array, `#equals` will return false regardless.

Example:

```java
DataQuery path = DataQuery.of("path");
int[] array1 = new int[] {1, 2, 3};
int[] array2 = new int[] {1, 2, 3};

MemoryDataView dataView1 = new MemoryDataView(); // Or DataContainer.createNew();
dataView1.set(path, array1);

MemoryDataView dataView2 = new MemoryDataView();
dataView2.set(path, array1);

MemoryDataView dataView3 = new MemoryDataView();
dataView3.set(path, array2);

dataView1.equals(dataView2); // -> false
dataView1.equals(dataView3); // -> false
```

WillBAnders:
Fixes #1983.

The current implementation of `MemoryDataView#equals` uses normal equality comparisons for the internal map, which does not support arrays. This PR replaces this segment of the equals method with one that checks values for arrays and performs the proper comparisons from there. I have also added in a test using a `byte[][]` array that shows this is working.

Whether this is correct solution, however, is a bit more debatable. The use of `DataTranslator`s was discussed in the issue, but given the explicit checks that `MemoryDataView` has for arrays in other locations this seems to be what's required.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/MemoryDataView.java
+++ b/src/main/java/org/spongepowered/api/data/MemoryDataView.java
@@ -844,9 +844,39 @@ public class MemoryDataView implements DataView {
             return false;
         }
         final MemoryDataView other = (MemoryDataView) obj;
+        if (!Objects.equal(this.path, other.path) || this.map.size() != other.map.size()) {
+            return false;
+        }
+        for (Map.Entry<String, Object> entry : this.map.entrySet()) {
+            if (!equalsCheckArrays(entry.getValue(), other.map.get(entry.getKey()))) {
+                return false;
+            }
+        }
+        return true;
+    }
 
-        return Objects.equal(this.map.entrySet(), other.map.entrySet())
-               && Objects.equal(this.path, other.path);
+    private static boolean equalsCheckArrays(Object o1, Object o2) {
+        if (o1.getClass().isArray() && o2.getClass().isArray()) {
+            if (o1 instanceof byte[] && o2 instanceof byte[]) {
+                return Arrays.equals((byte[]) o1, (byte[]) o2);
+            } else if (o1 instanceof short[] && o2 instanceof short[]) {
+                return Arrays.equals((short[]) o1, (short[]) o2);
+            } else if (o1 instanceof int[] && o2 instanceof int[]) {
+                return Arrays.equals((int[]) o1, (int[]) o2);
+            } else if (o1 instanceof long[] && o2 instanceof long[]) {
+                return Arrays.equals((long[]) o1, (long[]) o2);
+            } else if (o1 instanceof float[] && o2 instanceof float[]) {
+                return Arrays.equals((float[]) o1, (float[]) o2);
+            } else if (o1 instanceof double[] && o2 instanceof double[]) {
+                return Arrays.equals((double[]) o1, (double[]) o2);
+            } else if (o1 instanceof boolean[] && o2 instanceof boolean[]) {
+                return Arrays.equals((boolean[]) o1, (boolean[]) o2);
+            } else if (o1 instanceof Object[] && o2 instanceof Object[]) {
+                return Arrays.deepEquals((Object[]) o1, (Object[]) o2);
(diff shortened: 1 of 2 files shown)


dualspiral:
I realise this has been here some time now, apologies.

Any changes here will need to be mirrored in Common: https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/data/MemoryDataView.java. Best to make sure that this is OK first though, no point copying something that a change is then requested on.

It looks okay to me, but @gabizou might have a better insight.

src/main/java/org/spongepowered/api/data/MemoryDataView.java
```diff
@@ -844,9 +844,39 @@ public boolean equals(@Nullable Object obj) {
             return false;
         }
         final MemoryDataView other = (MemoryDataView) obj;
+        if (!Objects.equal(this.path, other.path) || this.map.size() != other.map.size()) {
+            return false;
+        }
+        for (Map.Entry<String, Object> entry : this.map.entrySet()) {
+            if (!equalsCheckArrays(entry.getValue(), other.map.get(entry.getKey()))) {
```
ImMorpheus, line 851: This doesn't account for `other.map.get(entry.getKey())` returning null and is duplicating what's inside `Arrays#deepEquals`.
This check can be reduced to `if (!Objects.deepEquals(entry.getValue(), other.map.get(entry.getKey())))`

ImMorpheus (changes requested):

src/test/java/org/spongepowered/api/data/MemoryDataTest.java
```diff
@@ -406,4 +407,12 @@ public void testDeepSerialization() {
         main.getMap(of()).get();
     }
 
+    @Test
+    public void testArrayEquality() {
+        DataQuery path = DataQuery.of("path");
+        DataContainer container1 = DataContainer.createNew().set(path, new byte[][] {{1, 2}, {3}, {}});
+        DataContainer container2 = DataContainer.createNew().set(path, new byte[][] {{1, 2}, {3}, {}});
+        assertEquals(container1, container2);
+    }
```
gabizou, line 416: Should test for nested array equality along with nested array inequality.

src/main/java/org/spongepowered/api/data/MemoryDataView.java
gabizou: Agreed, a lot of the deep equals can be replacing some of the equality methods written here. Plus, this class being deprecated, will need an update to the SpongeCommon variation.

gabizou (changes requested):

Zidane:
@dualspiral Another one, now or never.

dualspiral:
This actually needs to be done in Common, if it's done there, I'll backport it here as an "impl in API" thing.

Otherwise, closing this.

PR #2041, opened 2019-08-21
Implement New Explosion Settings

Pulverizer:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2358)

Code required for the implementation of three new variables in ExplosionMixin in SpongeCommon.

All variables to be used in the newer versions of my plugin.

JavaDocs still required in Explosion.java for the new methods. I don't know how to explain it in a clear manner.


ImMorpheus:
This is exposing a few implementation details. (`isRandomnessReduced` is really bad)

boundingBoxSize shouldn't be touched IMO, you can just use size.
EDIT: no, you can't, but I still don't think we should expose this.

> All variables to be used in the newer versions of my plugin, Movecraft for Sponge.

This doesn't really add anything to the PR since you didn't provide any use case (which would have been useful) and looks like self-promotion to me.

I will leave the final say to an api developer.

Pulverizer:
@ImMorpheus Right, sorry, I'm not very good at explaining it.

Reduced randomness - Reduces how random each ray is. This gives a more even explosion.

boundingBoxSize - Controls the number of rays cast by the explosion. This is **NOT** the same as `size`. Increasing this number **WILL** make explosions generate more lag. However, if a higher value is used on explosions larger than `size` 16 it will provide a nicer crater and generally removes the tunnel bores that usually occur. It can also be used to improve performance slightly by reducing the number of rays cast by explosions where `size` is less than 16.

I'm sorry that seems like self-promotion I was just trying to explain that these values are required for my plugin.

I've got a family matter to deal with atm as my Grandma has just passed away earlier today, but I will try to get some screenshots to you in the next few days demonstrating the results of using the variables.

Pulverizer:
TBF, these settings should come with a warning label and are not required for most explosions.

Pulverizer:
Imgur Album, each picture is described. https://imgur.com/a/A9Cqpap

dualspiral:
So, just looking at the API and impl together:

* I don't like `boundingBoxSize` at all. For a start, that name is not really describing what it is doing. You're increasing the number of rays from the source point of the explosion and it basically generates more explosions. I suspect a power of 2 was chosen for a reason, though I haven't combed through it enough to see what is going on. I don't think we should be touching that - what would you use it for?
* `isReducedRandomness` - I actually don't have much of an issue with this in general, though what I would prefer to see is this be a double between 0 and 1 that indicates how random the generated explosions are - and a better name!
* `entityKnockbackMultiplier` could be fun to play with...

I think we need to see what your use cases are for all these - you've said what they do, but not what they are.

Pulverizer:
I would recommend getting familiar with the Effect section of the [Minecraft Wiki Explosion page](https://minecraft.gamepedia.com/Explosion) as I am rubbish explaining things.

> So, just looking at the API and impl together:
> 
> * I don't like `boundingBoxSize` at all. For a start, that name is not really describing what it is doing. You're increasing the number of rays from the source point of the explosion and it basically generates more explosions. I suspect a power of 2 was chosen for a reason, though I haven't combed through it enough to see what is going on. I don't think we should be touching that - what would you use it for?

You're right, the name irks me too, but as with all the names, they are what I could think of.
I welcome any edits.

Minecraft Wiki:
`A cube around the explosion is divided into a 16×16×16 grid, and rays are created from the center to each outer point of this grid.`

With Vanilla you can see where the rays start to separate at the edges, the larger the explosion the more noticeable it is and tunnels start to form through the surrounding blocks.

Increasing this variable will cause addition lag and therefore should be done with caution. However, doing so allows you to give a nicer explosion crater with less to no tunnelling depending on the situation.

> * `isReducedRandomness` - I actually don't have much of an issue with this in general, though what I would prefer to see is this be a double between 0 and 1 that indicates how random the generated explosions are - and a better name!

I'm not sure what you mean?

Minecraft Wiki:
`Each ray is given an intensity, calculated as (0.7 + [a random value from 0 to 0.6]) × [power].`

`power` is what we'd call radius.

This variable simply sets if the rays should all use the same random value or not. With it on explosion craters are more symmetrical, as seen in the picture album.

> * `entityKnockbackMultiplier` could be fun to play with...

It's very fun.

> I think we need to see what your use cases are for all these - you've said what they do, but not what they are.

I haven't actually written it into my code yet but this is what my code would look like:
[Gist](https://gist.github.com/Pulverizer/6cc8b8d9b1ef7ad4645f59e9ede2183e)

dualspiral:
> `boundingBoxSize`

So, I actually went down the route of looking at the MCP code for explosions to see what it did, rather than reading the wiki. It's a set of unit vector directions, so rather than increasing the size of the rays, you're increasing the number of rays (as you've said). Going from 16 to 32 means that you basically increase the number of explosions by 8 (well, ish). You can see this might get bad very quickly.

It's nothing to do with size at all, it's not a bouding box. It's more to do with the "resolution" of the explosion. It _does_ increase the power of the explosion - and so I'm somewhat wary of this addition for the number of side effects it really will have.

Again - why do you want this? It's not code we need to see, nor what it does - it's an explanation as to why it's important to you. Having an API that no-one uses is not really a good API!

> isReducedRandomness - I actually don't have much of an issue with this in general, though what I would prefer to see is this be a double between 0 and 1 that indicates how random the generated explosions are - and a better name!

`(0.7 + [a random value from 0 to 0.6]) × [power]` -> `(1.0 + ([a random value from 0 to 0.6] × [randomness] - 0.3)) × [power])` - this isReducedRandomness would become randomness, centred around 1.0 x power (zero randomness would be "power", one would be current behaviour). The higher it is, the more random the pattern that comes out of explosions would appear.

Pulverizer:
@dualspiral So I've taken some time to think about how I can do a good job of explaining why I need/want these variables.

`boundingBoxSize` - I think you're right, `resolution` would be a much better name.

This is what the rays look like on a really big explosion:

Big Explosion Rays

You can just barely see the effects of using this variable in the picture album.

`radius` 16, current:



`radius` 16 with `resolution` 32:



While this will generate more lag it's a sacrifice I'm willing to make for tidier craters.

`isReducedRandomness` - I like what you're trying to do but this will mean that explosions with reduced randomness are always the lowest possible strength for their `radius`.

dualspiral:
> I like what you're trying to do but this will mean that explosions with reduced randomness are always the lowest possible strength for their radius

I got the equation wrong, I meant

```
(1.0 + (([a random value from 0 to 0.6] - 0.3) × [randomness])) × [power]
```

At randomness = 0, power is of strength 1 (times power). At a half, 0.85-1.15. At full, we get 0.7 to 1.3. That's what I was intending.

Pulverizer:
Okay, I'll try to revise the PRs with these changes in the next few days.

Pulverizer:
@dualspiral Made those changes. Think it's good to go?

dualspiral:
I think I'd like to get a set of opinions from the devs. I agree with @ImMorpheus that _some_ of this is just implementation that we probably don't want to expose.

I'm not trying to give anyone the runaround here, but I wanted to get a better understanding of what's going on - including names.

@SpongePowered/developers 

I think we need to come to a consensus on this. My feeling on this PR is that the knockback and randomness parts are probably okay. I don't think I'm particularly happy with the "resolution" being exposed though, other implementations such as Lantern might do explosions very differently and so resolution would have no meaning.

What do you all think?

WillBAnders:
I've added a couple of thoughts on the Common PR with reference to how the implementation is handled, but figured I'd note the overarching idea here as well.

The changes made seem to be arbitrary in the sense that there's a lot of areas for customization, both in what settings are exposed and how they're implemented. I think it makes the most sense for this to be it's own project separate from Sponge, that way you can tailor the customized behavior to your needs.

Additionally, reading through things in the implementation the names feel really weird and I can't predict what they do. I think the easiest way to get this across would be specifying how the explosions are related to these ray calcuations - ultimately `resolution` acts like a `rayDensity`, and `randomness` is basically `rayVariance`. Knockback also feels more like a `knockbackScale` as well - it's a multiplier there (on it's own it feels like a boolean).

dualspiral:
> randomness is basically rayVariance

I just want to quickly say no to this particular name. I suggested randomness as a name for the simple reason that you're otherwise specifying an implementation detail. Lantern, for example, does not have to use ray tracing for explosions if they don't want to.

If any of this is getting pulled, the names need to be sufficiently descriptive to explain what the outcome is, without explaining how that is done.

Pulverizer:
I agree with dualspiral 

In regards to being a separate project, are you suggesting I create a competing API or something because, to my understanding, it isn't possible to override a mixin with the @Final tag? As you said, bukkit could do it.

ImMorpheus:
> other implementations such as Lantern might do explosions very differently and so resolution would have no meaning.

This is exactly the "problem" I have with `randomness` and `resolution`. We shouldn't be biased by the current minecraft implementation.

Don't have an issue with knockback, maybe switch to a Vector3d to support different knockback for different coordinates ? 

Additionally pinging @Cybermaxke since working on Lantern/different implementation could really help with giving feedback on this.

ryantheleach:
@Pulverizer Aside from wanting to 'change explosions' What does Movecraft need/want from the API here?

Are you trying to control the explosion size? Density? or do you just want to basically just reimplement the type of explosion to look completely different, in a way that aligns with what you think explosions should look like?

It's my understanding that Movecraft is a port, are you porting functionality that already hooks into NMS and you are trying to replicate this feature exactly? Or is this a more free-form thing, where you would like to add a feature to Movecraft to change the explosions themselves.

Ultimately I believe that plugins currently have full control over the explosions, by modifying the list of affected blocks. Although the docs do say that may have a strange effect, and I havn't tested it myself. https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java#L89-L91

It would be my preference to remove any strange effects from removing the list of affected locations, and allow you to change the explosion to exactly how you would like it handled, without exposing implementation un-necessarially. (Even if this means that you need to basically copy the NMS algo + your changes.)

Is it possible to change the knockback of entities using the entity snapshots? with the current event? or is it such a performance drain, that it would be worth adding additional API for explosion knockback (Not that I think adding knockback API is bad, it just feels like extra maintenance that needs reasoning through).

dualspiral:
> or is it such a performance drain, that it would be worth adding additional API for explosion knockback (Not that I think adding knockback API is bad, it just feels like extra maintenance that needs reasoning through).

That's the one thing I think most people here agree is an okay thing to add. [The implementation amounts to being a multiplier like so](https://github.com/SpongePowered/SpongeCommon/pull/2358/files#diff-8b417570b299753ff70d99886c28d1a6R254-R256), I don't really think the maintainence burden is high on that and it saves recalculations later, I think it would be a performance hit to do otherwise.

As for the other points, yup, totally see where you are coming from - we really should remove side effects from that method but that may end up being done for 1.14 if it's decided that this is a deficient API.

I do really want to re-iterate a point you made ryan because I think this is really important to understand the motivation here:

> What does Movecraft need/want from the API here?
>
> Are you trying to control the explosion size? Density? or do you just want to basically just reimplement the type of explosion to look completely different, in a way that aligns with what you think explosions should look like?
>
> It's my understanding that Movecraft is a port, are you porting functionality that already hooks into NMS and you are trying to replicate this feature exactly? Or is this a more free-form thing, where you would like to add a feature to Movecraft to change the explosions themselves.

Yes. This. Exactly. I asked previously:

> Again - why do you want this? It's not code we need to see, nor what it does - it's an explanation as to why it's important to you. Having an API that no-one uses is not really a good API!

@Pulverizer, we never got a use case, we got examples of what you are proposing does - except that you might want a more even explosion by removing the randomness. As you can see, I am in favour of the knockback and I'm relatively okay with the idea of tweaking the randomness of the explosion, though as others have commented, this can be seen as an implementation detail, Lantern might not have randomness (though you could also argue that the API specifies that there should be some randomness in explosions, but I digress).

The reason we are asking for use cases is so that we can consider whether the API supports what you want to do, or if there are better solutions to it. But we need those use cases to help understand what you need from us, and whether this really is the best way forward.

What are you doing in Movecraft that **needs** these changes?

Pulverizer:
I am on holiday atm with only a phone. I thought I had managed to give the information that you desired. I see now that this is not the case but am not sure what you want.

I will try again...

Movecaft currently has a bit of code that allows for different levels of armour. However the method used is very crude and therefore, imo, damages gameplay.

My plan was to alter TNT explosions so that instead of blocks randomly being told to not break I could instead give the BlockType a higher blast resistance.

After having created a listener to condense explosions into a single larger explosion (to break the now stronger blocks) I found some issues.

A. Large explosions suffer from lack of ray density. (resolution) 

B. The strength of the blast along a given ray could not be guaranteed. (randomness) 

C. Knockback on a size 1 blast is the same as a size 1,000,00... (you get the point) blast. (knockback) 

So rather than copy the algorithm and have to run it in my listener (potentially increasing overhead) I thought it would to better to add the required control to Sponge.

@dualspiral @ryantheleach

ryantheleach:
@Pulverizer Thanks, that's exactly what I was looking for.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -91,6 +91,32 @@ static Builder builder() {
      */
     boolean shouldDamageEntities();
 
+    /**
+     * Gets the number of rays cast per length of each side of the explosion bounding box.
+     *
+     * @return The number of rays cast per length of each side of the explosion bounding box.
+     */
+    int getResolution();
```
gabizou: Because not default implementing this leads to a binary breakage for any thing that hasn't implemented the interface in previous builds, this would need to be default implemented to return the default resolution. This will allow for classes that implemented the class prior to this change to still be usable in other environments.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -152,6 +178,39 @@ static Builder builder() {
...
+         * <pre>
+         * <b>WARNING: It is recommended that this variable be set no higher than the default of 16!</b>
+         * The number of rays cast per length of each side of the explosion bounding box.
+         * The total number of rays that will be cast can be calculated using {@code [resolution] * 12}
+         * </pre>
+         *
+         * @param resolution The desired resolution of the explosion
+         * @return The builder, for chaining
+         */
+        Builder resolution(int resolution);
```
gabizou: Likewise with the builder.

gabizou (changes requested):
Default implement the methods that are being added. You can see it's been done with other interfaces being added on to to avoid binary breakage of precompiled classes.

Pulverizer:
@gabizou 👍 I'll get on that as quickly as possible. Been AFK for 2 weeks and am currently on getting settled in at uni.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -161,4 +232,4 @@ static Builder builder() {
         Explosion build() throws IllegalArgumentException;
 
     }
-}
+}
```
gabizou: Add new line at end of file, and then squash the commits, many thanks!

gabizou (approved):

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
Pulverizer: How do I squash the commits?

Pulverizer: And I tried quickly adding a new line but I think I messed up. :(

JBYoshi: @gabizou If this is ready to merge otherwise, GitHub can squash the commits when the PR is merged.

Pulverizer:
Is there anything I still need to do?

JBYoshi:
I'll go ahead and merge this since @gabizou approved it a while back.

dualspiral:
...but common wasn’t necessarily ready - which I noticed you haven’t merged, and is likely why this wasn’t merged.

Either that should be moved, or this should be reverted (can be done any time up to 7.2s release) We can’t do one and not the other. Please check the implementation PR and take appropriate action @JBYoshi

JBYoshi:
Whoops - completely forgot about that. I don't see any complaints on the implementation at the moment, so I'll look at it and run some tests.

PR #2045, opened 2019-09-07
Add ability to change potion color

### LINKED ISSUE #1961: Ability to modify potion colors
As discussed in this Thread https://forums.spongepowered.org/t/modify-color-of-potion/27170 it seems like there currently is no option to modify the color of the potion items. Per default, every potion is pink. Since you can modify the potion colors in the give command, it should be possible to implement this in the API.

RedNesto:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2370)

This PR introduces a new Key: `POTION_COLOR`. 

- It is used to change the color of an ItemStack potion.
- Getting the value will return the custom color if set, or the default color for the attached effects.
- Thrown potions color can be changed by updating their represented item.
- It is possible to remove the custom color.

Resolves #1961


JBYoshi (approved):

gabizou (approved):
Looks good. Will get to merging when I'm home.

PR #2047, opened 2019-09-20
Add economy account deletion

fingoltin:
**API only**

Adds support for deleting economy accounts. Left as a default method so as to not break any existing economy plugins.

@ItsDoot ;)

### DIFF
--- a/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
+++ b/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
@@ -26,6 +26,8 @@ package org.spongepowered.api.service.economy;
 
 import org.spongepowered.api.service.context.ContextualService;
 import org.spongepowered.api.service.economy.account.Account;
+import org.spongepowered.api.service.economy.account.AccountDeletionResultType;
+import org.spongepowered.api.service.economy.account.AccountDeletionResultTypes;
 import org.spongepowered.api.service.economy.account.UniqueAccount;
 import org.spongepowered.api.service.economy.account.VirtualAccount;
 
@@ -115,4 +117,32 @@ public interface EconomyService extends ContextualService<Account> {
      * @return The {@link Account}, if available.
      */
     Optional<Account> getOrCreateAccount(String identifier);
+
+    /**
+     * Deletes the account for the user with the specified {@link UUID}.
+     *
+     * <p>Deletion might fail if the provided {@link UUID} does not correspond to
+     * an actual player, or for an implementation-defined reason.</p>
+     *
+     * @param uuid The {@link UUID} of the account to delete.
+     * @return The result of the deletion.
+     */
+    default AccountDeletionResultType deleteAccount(UUID uuid) {
+        return AccountDeletionResultTypes.UNSUPPORTED;
+    }
+
+    /**
+     * Deletes the account with the specified identifier.
+     *
+     * <p>If an account exists with the specified identifier,
+     * it will be deleted.</p>
+     *
+     * <p>Deletion may fail for an implementation-defined reason.</p>
+     *
+     * @param identifier The identifier of the account to delete.
+     * @return The result of the deletion.
+     */
+    default AccountDeletionResultType deleteAccount(String identifier) {
+        return AccountDeletionResultTypes.UNSUPPORTED;
+    }
 }
--- a/src/main/java/org/spongepowered/api/service/economy/account/AccountDeletionResultType.java
+++ b/src/main/java/org/spongepowered/api/service/economy/account/AccountDeletionResultType.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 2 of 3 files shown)


liach:
Can it return a boolean to indicate operation success or failure?

fingoltin:
> Can it return a boolean to indicate operation success or failure?

It could, and I had it there to begin with. It just wouldn't be as pretty. Should I add it?

liach:
If doot asked for the deletion functionality, let him decide it.

fingoltin:
> If doot asked for the deletion functionality, let him decide it.

He just mentioned it was bothering him too. @ItsDoot what'cha think?

ItsDoot:
Yea, definitely needs to return `boolean`. We gotta know if it was successful or not.

src/main/java/org/spongepowered/api/service/economy/account/Account.java
```diff
@@ -372,8 +372,10 @@ default TransferResult transfer(Account to, Currency currency, BigDecimal amount
 
     /**
      * Deletes the account, removing all traces of the account having existed.
+     *
+     * @return Whether the deletion was successful.
      */
-    default void delete() {
-
+    default boolean delete() {
```
ST-DDT: This should indicate possible reasons for the failure of the deletion in the javadocs.

fingoltin: "Implementation related issues". What else could even be a factor?

dualspiral: Possible reasons include that the account doesn't exist in the first place, or that account deletion is not supported (default case).

I'd consider whether this boolean could be a new `AccountDeletionResult` (name pending) `CatalogType` that has the common reasons as failiure options, an `OTHER_FAILURE` (name also pending) option, as well as a success option, at least a plugin using this might have a better idea as to what happened.

@Flibio and @Erigitic might have an interest in this too as economy plugin authors.

fingoltin: I suppose you mean on the method in the service primarily.

dualspiral: I would remove this method and only keep the service one. It makes no sense to have a `delete` method on the `Account` object as you'd still have a reference to the account and so would have an object that is practically worthless to you. 

Better to just have it on the service where you don't have the object created at all.

ST-DDT:
I considered the following results:

* ABSENT ( doesnt exist )
* SUCCESS
* RESET  (player that is online)
* UNDELETEABLE (virtual and game accounts)
* UNSUPPORTED (feature not available)
* FAILED_IN_USE ? (player online)
* FAILED_ERROR (internal error)
* ?

fingoltin:
Revisiting this PR; should the `Account` even have a `delete()` method?

Zidane:
No, put it on the EconomyService only.

ST-DDT (approved):

src/main/java/org/spongepowered/api/service/economy/EconomyService.java
```diff
@@ -25,9 +25,7 @@
 package org.spongepowered.api.service.economy;
 
 import org.spongepowered.api.service.context.ContextualService;
-import org.spongepowered.api.service.economy.account.Account;
-import org.spongepowered.api.service.economy.account.UniqueAccount;
-import org.spongepowered.api.service.economy.account.VirtualAccount;
+import org.spongepowered.api.service.economy.account.*;
```
dualspiral: No star imports please.

src/main/java/org/spongepowered/api/service/economy/account/AccountDeletionResultTypes.java
```diff
@@ -0,0 +1,64 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.economy.account;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+public class AccountDeletionResultTypes {
```
dualspiral, line 29: Unlike most of the rest of the econ API, these will need implementing in Common as these are intended to be constants - they wouldn't be implemented by other plugins.

src/main/java/org/spongepowered/api/service/economy/account/AccountDeletionResultType.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.service.economy.account;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(AccountDeletionResultTypes.class)
+public interface AccountDeletionResultType extends CatalogType {
```
dualspiral, line 31: Perhaps add a `boolean isSuccess()` method on to here for easy use in things like if statements? Sometimes you just want to know if it worked.

dualspiral (dismissed):

dualspiral:
There are some niggles (javadocs over 80 characters wide, etc. - [please note that we do have a style guide](https://docs.spongepowered.org/stable/en/contributing/implementation/codestyle.html) that should be adhered to), but otherwise, there needs to be a companion Common PR that adds a registry module and implementations to the `AccountDeletionResultTypes` before this is pulled.

dualspiral (approved):
Nitpicks aside, this is okay. I want to see this merged soon, but if someone merges this, _please_ sort out the line line lengths and whatnot.

Give it a couple of days in case anyone has any objections first. Econ plugin authors (current and emerging), any comments before we do?

Issue #2048, opened 2019-09-23
Fix no permission message


MoeBoy76:
after https://github.com/SpongePowered/SpongeAPI/commit/b38390e1751fff802d355155c45b2eb4913f5817 when a player hasn't got permission they get a message saying the command doesn't exist, this is very frustrating when attempting to set up a new server/plugin as it appears that the plugin isn't loaded when you just don't have permission for the command(s)

dualspiral:
I was discussing this with someone on the Nucleus Discord as to what the change was intended to do, message reproduced below.

I'll take this on over the next couple of days when I get a moment, I know how I want to fix this in a simple way, at least in the short term - mentioned below.

---

For background - that commit comes from this PR - https://github.com/SpongePowered/SpongeAPI/pull/2038. The idea was that it would prevent the commands appearing in /help and though tab completion - before this commit this was happening. This is something that a lot of people have been asking for and so makes sense. That method basically returns the command if the source should be able to use it - and is used during execution too. 

A solution would be to pass in `null` as the source during execution, it would then suppress tab completion and usage in /help, but allow for it to be executed if desired. Could use a config option too.

PR #2049, opened 2019-10-06
Slight refactor of SimpleDispatcher to support SpongePowered/SpongeCommon#2385

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2385) | [Original Issue](https://github.com/SpongePowered/SpongeAPI/issues/2048)

This PR is mostly to remove the `final` modifier from `SimpleDispatcher` and rearrange a little bit of logic to support the goals of SpongePowered/SpongeCommon#2385 and SpongePowered/SpongeAPI#2048.

Big thing here is removing the `final` modifier. I don't see it as a problem. However, if we are not okay with that, that's okay, but the whole Dispatcher will be copied into Common.

Please see the Common PR for a lot more information as to what this will achieve.

---

* Move some logic into protected methods to reduce deduplication
* As this was final, no previous plugins will have extended this - changes are safe.
* Logic has not changed except for a reversion (below) that will go into the impl instead.
* Revert "Filter CommandMapping result based on source permission.", this reverts commit b38390e1751fff802d355155c45b2eb4913f5817 (the spirit of this will be in the Common PR).

### DIFF
--- a/src/main/java/org/spongepowered/api/command/dispatcher/SimpleDispatcher.java
+++ b/src/main/java/org/spongepowered/api/command/dispatcher/SimpleDispatcher.java
@@ -68,7 +68,7 @@ import javax.annotation.Nullable;
 /**
  * A simple implementation of a {@link Dispatcher}.
  */
-public final class SimpleDispatcher implements Dispatcher {
+public class SimpleDispatcher implements Dispatcher {
 
     /**
      * This is a disambiguator function that returns the first matching command.
@@ -170,7 +170,6 @@ public final class SimpleDispatcher implements Dispatcher {
         checkNotNull(callback, "callback");
 
         // Invoke the callback with the commands that /can/ be registered
-        // noinspection ConstantConditions
         aliases = ImmutableList.copyOf(callback.apply(aliases));
         if (aliases.isEmpty()) {
             return Optional.empty();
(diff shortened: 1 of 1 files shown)


liach:
can't you just make a command dispatcher that use this one in a field and just delegate most calls?

liach:
and if you do decide to remove final, consider making some fields protected

dualspiral:
> can't you just make a command dispatcher that use this one in a field and just delegate most calls?

It becomes tricky to ensure that the common PR can do what it does without copying huge chunks of code - if you read how the common PR works compared to this you'll see that the `get` method requires a third parameter there - and not being able to override methods would have severely hampered what I was trying to do.

Quite frankly, I'm not wasting more time on something that is going to be gutted in API 8 anyway. The only reason I've done it this way is to minimise the code duplication, such that if we _do_ need to change some logic to fix a bug or whatever, we're not duplicating these fixes. There may be others using such a dispatcher.

It's either this or copying it into common. I've made no secret about that, it's something I muse upon in the Common PR. It's a draft because I'm not sure I want to do this at all, and I simply want opinions on final vs. non final from @gabizou and @Zidane.

> consider making some fields protected

No. I am simply going to do what is necessary to support the implementation. Honestly, this class should not be in the API layer at all and so, without breaking things, I'm going to consider it as implementation. If someone then wants something to be exposed for whatever reason, they can PR it, but again, I see little reason to.

dualspiral:
I've talked to @Zidane about this, and all I'm going to do is copy this into Common - no-one should be using this anyway...

PR #2050, opened 2019-10-09
Inventory for API8 (part 3) Queries

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2395)

based on changes in: #1998 

#### Merged QueryOperations and InventoryTransformations ####
  * `Query`, `QueryType.NoParam|OneParam|TwoParam`
  * `transform` replaced by `query(Query)`
  * new query type to query for a grid in another grid.


src/main/java/org/spongepowered/api/item/inventory/query/QueryTypes.java
```diff
@@ -73,32 +80,47 @@
...
+
+    /**
+     * Query for a reverse order of slots.
+     */
+    public static final NoParam REVERSE = DummyObjectProvider.createFor(NoParam.class, "REVERSE");
+
+    /**
+     * A grid query. Only works on grids. The first value is the offset the second value is the grid size.
+     */
+    public static final TwoParam<Vector2i, Vector2i> GRID = DummyObjectProvider.createExtendedFor(TwoParam.class, "REVERSE");
```
ST-DDT: Should be `GRID`

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -318,23 +319,46 @@
...
+    /**
+     * Query this inventory with given {@link QueryType} and {@code param}.
+     *
+     * @param type The query type
+     * @param param The query parameter
+     * @param <P> The query parameter type
+     *
+     * @return The queried inventory
+     */
+    <P> Inventory query(QueryType.OneParam<P> type, P param);
```
ST-DDT: IMO this method should be removed as it isn't any better readable than query(type.of(param))

src/main/java/org/spongepowered/api/item/inventory/Inventory.java
```diff
@@ -318,23 +319,46 @@
...
+     * @param type The query type
+     * @param param1 The first query parameter
+     * @param param2 The second query parameter
+     * @param <P1> The first query paramater type
+     * @param <P2> The second query parameter type
+     *
+     * @return The queried inventory
      */
-    Inventory query(QueryOperation<?>... operations);
+    <P1, P2> Inventory query(QueryType.TwoParam<P1, P2> type, P1 param1, P2 param2);
```
liach: Agree with ST-DDT. We can virtually have infinite number of params and the list of such methods can go on, but we don't want to explode the class with such methods

liach:
@Faithcaio can you rebase api8/redesign/inventory onto api-8 so that changes are better reflected?

src/main/java/org/spongepowered/api/item/inventory/query/QueryType.java
```diff
@@ -29,21 +29,52 @@
...
  */
-@CatalogedBy(QueryOperationTypes.class)
-public interface QueryOperationType<T> extends CatalogType {
+@CatalogedBy(QueryTypes.class)
+public interface QueryType extends CatalogType {
+
+    /**
+     * A type of query that requires no parameters. It can directly be used as a query.
+     */
+    interface NoParam extends QueryType, Query {
```
liach: `NoParam` shouldn't directly extend `Query`; should rather add a factory method and leave this as an impl detail

Faithcaio: Why not? They're basically just predefined Queries (previously InventoryTransformations)
I don't see a need for a factory method creating more instances.

`inv.query(QueryTypes.REVERSE)`
`playerInv.query(QueryTypes.PLAYER_PRIMARY_HOTBAR_FIRST)`

liach: I don't see a need for object types to implement `execute` method.

PR #2054, opened 2019-10-22
Add compact methods to Text

### LINKED ISSUE #1312: Stupidly Difficult Idea (TM) - Text Data Smart Trimming

Katrix:
Retargets #1424

Fixes #1312

### DIFF
--- a/src/main/java/org/spongepowered/api/text/Text.java
+++ b/src/main/java/org/spongepowered/api/text/Text.java
@@ -194,6 +194,25 @@ public interface Text extends Comparable<Text>, DataSerializable, TextRepresenta
     }
 
     /**
+     * Builds a compact {@link Text} from a given array of objects.
+     *
+     * <p>Like the normal {@link #of(Object...)} method, but will try to
+     * compact the passed in text objects if possible. No guarantees are given
+     * regarding the structure of the {@link Text}, only that it will
+     * display the same.</p>
+     *
+     * <p>The result of this might be, but does not need to be equivalent to
+     * {@code Text.of(arrayOfObjects).compact()}.</p>
+     *
+     * @see #of(Object...) For more information on the normal of factory.
+     * @param objects The object array
+     * @return The built compact text object
+     */
+    static Text ofCompact(Object... objects) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -729,6 +759,18 @@ default Text toText() {
          */
         List<Text> getChildren();
 
+		/**
+		 * Gets if this builder should try to compact when appending text.
+		 *
+		 * @return If this builder will compact when appended text
+		 */
+		boolean shouldCompact();
+
```
ST-DDT, line 768: Use spaces instead of tabs.

src/main/java/org/spongepowered/api/text/Text.java
```diff
@@ -193,6 +193,25 @@ static Text of(Object... objects) {
...
+     * display the same.</p>
+     *
+     * <p>The result of this might be, but does not need to be equivalent to
+     * {@code Text.of(arrayOfObjects).compact()}.</p>
+     *
+     * @see #of(Object...) For more information on the normal of factory.
+     * @param objects The object array
+     * @return The built compact text object
+     */
+    static Text ofCompact(Object... objects) {
```
ST-DDT, line 211: I would rename it to `compactOf` or `compactedOf` same for the factory method.

PR #2056, opened 2019-11-01
[API8] Allow regular gradle dsl to have spongeapi's kotlin dsl subproject

liach:
Without these 3 id declarations, the groovy-based regular gradle dsl will error out when encountering these lines, trying to resolve these as class names instead of gradle plugin ids.


gabizou:
This is already done. Thanks for the tip!

PR #2066, opened 2019-12-04
Updates to the Command API to support implementation

dualspiral:
The big thing here is the addition of a "CommandRegistrar". This is meant to be a lightweight interface that tries to give external command frameworks (Aikar's ACF, Katrix's Scammander, Faith's Butler, amongst others) a simple way to add their commands to Sponge. The idea of registrars was originally an implementation one - if you look at that code, you'll see that the vanilla brig system is separated from the Sponge one - it takes little work to add support for arbitary registrars, so why not add them?

The Javadocs should provide a fair amount of explanation. There will be more for client completions, I'm sure - though enabling that is the thing that is missing right now. That'll come later - it'll likely mimic Brig's suggestion provider with a tree like format to build the command elements up with.

This is simply for review of code style and Javadocs.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/manager/CommandManager.java
+++ b/src/main/java/org/spongepowered/api/command/manager/CommandManager.java
@@ -25,8 +25,11 @@
 package org.spongepowered.api.command.manager;
 
 import org.spongepowered.api.command.Command;
+import org.spongepowered.api.command.CommandCause;
 import org.spongepowered.api.command.CommandResult;
 import org.spongepowered.api.command.exception.CommandException;
+import org.spongepowered.api.command.registrar.CommandRegistrar;
+import org.spongepowered.api.command.registrar.tree.CommandTreeBuilder;
 import org.spongepowered.api.service.permission.Subject;
 import org.spongepowered.api.text.channel.MessageChannel;
 import org.spongepowered.api.text.channel.MessageReceiver;
@@ -34,7 +37,7 @@ import org.spongepowered.plugin.PluginContainer;
 
 import java.util.Collection;
 import java.util.List;
-import java.util.Optional;
+import java.util.function.Predicate;
 
 /**
  * Registers and dispatches commands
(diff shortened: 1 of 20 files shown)


ItsDoot:
So effectively, `CommandRegistrar` is similar to a `Dispatcher` + `CatalogType`?

src/main/java/org/spongepowered/api/command/manager/CommandManager.java
```diff
@@ -112,7 +113,42 @@
     List<String> suggest(Subject subject, MessageChannel receiver, String arguments);
 
     /**
-     * Registers a {@link Command} with the {@link CommandManager}.
+     * Registers a {@link Command} with the Sponge {@link CommandRegistrar}.
```
Zidane: I'd remove Sponge in this sentence.

src/main/java/org/spongepowered/api/command/manager/CommandManager.java
```diff
@@ -112,7 +113,42 @@
...
+     * @return The {@link CommandMapping} containing the command mapping
+     *         information.
+     * @throws FailedRegistrationException thrown if the command could not be
+     *                                     registered.
+     */
+    CommandMapping registerAlias(CommandRegistrar registrar,
+            PluginContainer container,
+            String primaryAlias,
+            String... secondaryAliases)
+            throws FailedRegistrationException;
```
Zidane: CommandFailedRegistrationException ?

dualspiral:
I guess, though it doesn't really need to be a dispatcher so much. A registrar could conceivebly be a command executor - I don't put a restriction on it.

Note that there are a couple of classes that don't have JDs. That's a mistake on my part, some API elements have made it through that were nowhere near ready. I won't be merging tonight as it is, so pick on anything and I'll review over the next couple of days.

src/main/java/org/spongepowered/api/command/parameter/CommandContext.java
```diff
@@ -176,24 +177,69 @@
...
+             */
+            void rollback();
+
+            /**
+             * A {@link Transaction} cannot be started from within a transaction.
+             *
+             * @return The current transaction.
+             */
+            @Override
+            @Deprecated
```
Zidane: Why is new API deprecated?

dualspiral: The `Transaction` was intended to be a replacement for `CommandContext.Builder` when sending it through elements. Building the builder is part of that interface, but transactions shouldn't be built. Same with starting a transaction from a transaction.

I think I'll ditch that notion and just have a Transaction as a stack frame like in causes. It'll minimise changes I have to make elsewhere, but just makes the context builder impl more complicated.

That's what implementation is for though!

PR #2067, opened 2019-12-07
Remove blocktype from location#tocontainer

ImMorpheus:
Location#toContainer() calls World#getBlockType, leading to an Illegal Async Chunk Load if you try to serialize a location asynchronously.
It's also lost when deserializing* a location object (the blocktype is saved but it's ignored by locationbuilder) so there's no need for a contentupdater.

*https://github.com/SpongePowered/SpongeCommon/blob/0b6e0819d266e9f92770a6024fd26458b7264a90/src/main/java/org/spongepowered/common/data/builder/world/LocationBuilder.java#L46


gabizou:
Should bump the version to omit the block type.

PR #2070, opened 2019-12-17
Make TradeOfferGenerator extend a BiFunction instead to handle 1.14 implementation.

i509VCB:
This is to compensate for 1.14 now requiring a Merchant to call `VillagerTrades$ITrade#getOffer(Entity, Random)` since some Villagers such as the Cartographer Villager require access to the world from the entity to find a Woodland Mansion or a Monument.

This here is part of the changes with https://github.com/SpongePowered/SpongeCommon/pull/2434


liach:
Is this used as a bifunction so often that it must implement `BiFunction`? Or is it used rarely enough that a method reference that convert it to a bifunction suffices?

Issue #2069, opened 2019-12-13
[1.14] Add a mechanism to get raids which are in the ServerWorld


i509VCB:
I noticed this while looking around the raid api code. There is currently no way to get a raid from the world.

Based on how the implementation would probably act, the raids are in the ServerWorld internally, so for consistency this would make sense in the api's ServerWorld. Since raids can happen in any dimension (yes it sounds ridiculous but you can have a raid in the end [game code doesn't allow raids in the nether]) it further justifies binding raid access to the world. 

The best option would probably be to add a method to get a list of raids in the world since multiple raids can happen concurrently.

i509VCB:
Closed #2073

Issue #2068, opened 2019-12-12
[1.14] New Villager and Mechant related Events


i509VCB:
Since 1.14 made villagers more complex, some new events would be convient. These are the concepts for some events that could be added. 

CareerChangeVillagerEvent:
Since villagers can now change their careers, this event would have a reason to exist. 
This would be fired in a context of a villager taking a profession (Where the POI block is the cause) or where a villager loses their profession (either when the POI block is broken or the Villager can no longer pathfind to the POI). Note that once a villager has more than 1 profession expirence, the Villager will not lose its profession. 

ProfessionLevelVillagerEvent
Villagers have multiple levels of Mastery of their trades. You can see this by the colored badge on their texture. This event would be fired when a villagers mastery level increases. 
The event's context would consist of the Player whose trade caused the level to increase (and possibly the trade which made the level increase). 

TradeMerchantEvent
This is fired when a player trades with a Merchant (Villager or Wandering trader) 
Part of the event's context could include the player's reputation level [if a villager]).

RaidEvent.Start
Event which is fired when a raid starts. Would provide information such a the level of the raid starting and would be Cancellable. 
Context:
Player who started the raid. 
Possibly a list of villagers which are subjects of the raid. 

RaidEvent.Wave
An event for when a raid wave is started. 
Context:
Raid - The raid which is occurring
Player - The player who started the raid
Raiders - The raiders spawned from this event 

RaidEvent.End
An event which signifies the end of a raid. This includes both a victory and defeat. 
Context:
Raid - The raid which just ended
Player - The player who started the event
Player(s) - The player(s) if the result was a victory.
Raider(s) - The winning raiders if the result was a defeat.

PanicVillagerEvent
Event which is fired due to a villager starting to panic. This should contain a flag somewhere to specify if an iron golem will be spawned. 
Context:
Raid - The raid which caused the paniccing if the event was caused by a raid. 
Entity(s) - The entities which are causing the Villager to panic.

~~GossipVillagerEvent~~ (This would require a gossip API which we do not have yet).
Should be self explanatory, but plenty of different contexts to handle the situations from a player killing a villager (gossip event which is instantly transmitted) up to curing. This definitely will need more discussion for each specific case it could be toggled from. 

Events for schedules 
This is a bit more advanced due to how the schedule system works. This could use more discussion. This may classify much more as an AI Category than a villager issue. 

Other cases:
Villagers gifting players with hero of the village. 
[...]

PR #2073, opened 2019-12-18
Add new 1.14 Merchant, Raid and Villager related features. #2068 #2069

i509VCB:
**This adds some new events (#2068):**

RingBellEvent (Needs to have a way to get a list of all villagers which will hear the bell ring).

VillagerEvent.ProfessionChange
VillagerEvent.ProfessionLevelUp
VillagerEvent.Panic

TradeMerchantEvent

RaidEvent.Start
RaidEvent.StartWave
RaidEvent.End

**Other additions:**

Adds a new Key to get a Villager's Profession level.

Add a way to reference all the Raids within a world, inside of ServerWorld. (#2069)

### DIFF
--- a/src/main/java/org/spongepowered/api/event/block/entity/RingBellEvent.java
+++ b/src/main/java/org/spongepowered/api/event/block/entity/RingBellEvent.java
@@ -22,42 +22,25 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.entity.living.trader;
+package org.spongepowered.api.event.block.entity;
 
-import org.spongepowered.api.data.Keys;
-import org.spongepowered.api.data.type.Profession;
-import org.spongepowered.api.data.type.VillagerType;
-import org.spongepowered.api.data.value.Value;
-import org.spongepowered.api.entity.living.Ageable;
+import org.spongepowered.api.block.entity.Bell;
+import org.spongepowered.api.effect.potion.PotionEffectTypes;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.entity.AffectEntityEvent;
 
 /**
- * Represents a Villager.
+ * An event when a {@link Bell} is rung.
+ *
+ * <p>Entities which are affected by this event will be given a {@link PotionEffectTypes#GLOWING} effect.</p>
  */
-public interface Villager extends Trader, Ageable {
+public interface RingBellEvent extends AffectEntityEvent, Cancellable {
 
-    /**
-     * {@link Keys#VILLAGER_TYPE}
-     * @return The villager type
-     * @see org.spongepowered.api.data.type.VillagerTypes
-     */
-    default Value.Mutable<VillagerType> type() {
-        return getValue(Keys.VILLAGER_TYPE.get()).get().asMutable();
-    }
+	/**
+	 * The {@link Bell} which was rung.
+	 *
+	 * @return The bell which was rung.
+	 */
+	Bell getBell();
 
-    /**
-     * {@link Keys#PROFESSION}
-     * @return The profession of this villager
-     * @see org.spongepowered.api.data.type.Professions
-     */
-    default Value.Mutable<Profession> profession() {
-        return getValue(Keys.PROFESSION.get()).get().asMutable();
-    }
-
-    /**
-     * {@link Keys#PROFESSION_LEVEL}
-     * @return The profession level of this villager
-     */
-    default Value.Mutable<Integer> professionLevel() {
-        return getValue(Keys.PROFESSION_LEVEL).get().asMutable();
-    }
 }
--- a/src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
@@ -1,0 +1,135 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.entity.living.trader;
+
+import org.spongepowered.api.data.type.Profession;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.living.Hostile;
+import org.spongepowered.api.entity.living.golem.IronGolem;
+import org.spongepowered.api.entity.living.monster.raider.Raider;
+import org.spongepowered.api.entity.living.monster.zombie.ZombieEntity;
+import org.spongepowered.api.entity.living.trader.Villager;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.raid.Raid;
+import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
+
+/**
+ * An event which involves a {@link Villager}.
+ */
+public interface VillagerEvent extends Event {
+
+	/**
+	 * Gets the {@link Villager} involved with this event.
+	 *
+	 * @return The villager.
+	 */
+	Villager getVillager();
+
+	/**
+	 * Fired when a {@link Villager}'s profession changes.
+	 *
+	 * <p>This can include both gaining or losing a {@link Profession}.</p>
+	 */
+	@GenerateFactoryMethod
+	interface ChangeProfession extends VillagerEvent, Cancellable {
+
+		/**
+		 * Gets the villager's original {@link Profession}.
+		 *
+		 * @return The {@link Profession} the {@link Villager}'s will take.
+		 */
+		Profession getOriginalProfession();
+
+		/**
+		 * Gets the {@link Profession} the villager will change to.
+		 *
+		 * @return The {@link Villager}'s next {@link Profession}.
+		 */
+		Profession getProfession();
+
+		/**
+		 * Sets the {@link Villager}'s next {@link Profession}.
+		 */
+		void setProfession(Profession profession);
+	}
+
+	/**
+	 * Fired when a {@link Villager} levels up it's {@link Profession}.
+	 */
+	@GenerateFactoryMethod
+	interface LevelUpProfession extends VillagerEvent, Cancellable {
+
+		/**
+		 * Gets the {@link Villager}'s original {@link Profession} level.
+		 *
+		 * @return The {@link Profession} level.
+		 */
+		int getOriginalProfessionLevel();
+
+		/**
+		 * Gets the {@link Profession} level this {@link Villager}'s will level up to.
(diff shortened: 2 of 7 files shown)


src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,104 @@
...
+		 * @return The {@link Profession} level.
+		 */
+		int getNextProfessionLevel();
+
+		/**
+		 * Sets the profession level of this {@link Villager}.
+		 *
+		 * @param level The level to set the {@link Villager}'s {@link Profession} to.
+		 *
+		 * <p>Note that any level above 5 will usually be ignored.</p>
```
ST-DDT: Maybe phrase it like this:

> Please note that any level above 5 usually does not define `TradeOfferGenerator`s and is therefore ignored.

i509VCB: Changed docs to your tweak.

src/main/java/org/spongepowered/api/event/item/merchant/TradeMerchantEvent.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.api.entity.living.Humanoid;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.item.merchant.Merchant;
+import org.spongepowered.api.item.merchant.TradeOffer;
+
+/**
+ * An event when a {@link Humanoid} trades with a {@link Merchant}.
+ */
+public interface TradeMerchantEvent extends Event, Cancellable {
```
Zidane: TradeWithMerchantEvent

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Fired when a {@link Villager}'s profession changes.
+	 *
+	 * <p>This can include both gaining an losing a {@link Profession}.</p>
+	 */
+	@GenerateFactoryMethod
+	interface ProfessionChange extends VillagerEvent, Cancellable {
+
+		/**
```
Zidane, line 66: Something we do that other APIs really don't is add getOriginalXXXX() to events

getOriginalHealth()
getOriginalExperience()
etc, you get the idea

We do this so a plugin can know what it was to begin with. I'd like to see that in these events.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,106 @@
...
+	@GenerateFactoryMethod
+	interface ProfessionLevelUp extends VillagerEvent, Cancellable {
+
+		/**
+		 * Gets the {@link Villager}'s current {@link Profession} level.
+		 *
+		 * @return The {@link Profession} level.
+		 */
+		default int getCurrentProfessionLevel() {
+			this.getVillager().professionLevel().get();
```
Zidane: This needs to be passed to the event factory in the impl, never default and point to the source like this.

src/main/java/org/spongepowered/api/event/raid/RaidEvent.java
```diff
@@ -0,0 +1,56 @@
...
+	@GenerateFactoryMethod
+	interface StartWave extends RaidEvent, Cancellable {
+
+		/**
+		 * The {@link Wave} which is starting.
+		 *
+		 * @return The current wave.
+		 */
+		default Wave getWave() {
+			return this.getRaid().getCurrentWave().get();
```
Zidane: Same here, no default.

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
```diff
@@ -147,4 +150,32 @@ default UUID getUniqueId() {
...
+     */
+    Optional<Raid> getRaidAt(Vector3i blockPosition);
+
+    /**
+     * Checks if a {@link Raid} is occurring at a position in the world.
+     *
+     * @param blockPosition The location of the Raid.
+     * @return True if a Raid is present at the location.
+     */
+    default boolean hasRaidAt(Vector3i blockPosition) {
```
Zidane: Our plugin devs understand Optional. Not necessary to sugar this.

src/main/java/org/spongepowered/api/event/block/entity/RingBellEvent.java
```diff
@@ -0,0 +1,30 @@
...
+	 * @return The bell which was rung.
+	 */
+	Bell getBell();
+
+	/**
+	 * The {@link Entity}s which will get a {@link PotionEffectTypes#GLOWING} effect because the {@link Bell} was rung.
+	 *
+	 * @return A list of Entities that will get the glowing effect.
+	 */
+	List<Entity> entitiesToGlow();
```
gabizou: This might be better to mimic AffectEntityEvent where there's the immutable list of entities originally, versus the now mutable list that'll be used.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
gabizou: Right, and especially since a plugin can change the profession level value between event listeners.

gabizou: Mind you that this would technically be possible to be part of `ChangeValueEvent` as a sub class `LevelUpProfession` that is based on the `Keys.PROFESSION_LEVEL`. I'll be making adjustments to `event-impl-gen` to allow annotating `Key`s fields to generate an event based on it.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,106 @@
...
+		 * Sets the {@link Villager}'s next {@link Profession}.
+		 */
+		void setNextProfession(Profession profession);
+	}
+
+	/**
+	 * Fired when a {@link Villager} levels up it's {@link Profession}.
+	 */
+	@GenerateFactoryMethod
+	interface ProfessionLevelUp extends VillagerEvent, Cancellable {
```
gabizou: `LevelUpProfession` since it's `ActionTargetEvent` style.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	Villager getVillager();
+
+	/**
+	 * Fired when a {@link Villager}'s profession changes.
+	 *
+	 * <p>This can include both gaining an losing a {@link Profession}.</p>
+	 */
+	@GenerateFactoryMethod
+	interface ProfessionChange extends VillagerEvent, Cancellable {
```
gabizou: `ChangeProfession` like `ChangeBlockEvent`

src/main/java/org/spongepowered/api/event/item/merchant/TradeMerchantEvent.java
```diff
@@ -0,0 +1,35 @@
...
+	 * @return The trading merchant.
+	 */
+	Merchant getMerchant();
+
+	/**
+	 * Gets the selected {@link TradeOffer} a {@link Humanoid} could buy.
+	 *
+	 * @return The trade offer.
+	 */
+	TradeOffer getSelectedOffer();
```
gabizou: You should be able to change the offer iirc, since we can throw the event prior to the offer being sent down the wire.

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
gabizou: I agree, defaulted simplistic "check if the optional is present" methods are bloat.

gabizou (changes requested):
Well done, only went through functional aspects of the PR, will do another review of documentation a bit later, but a cursory look was fine.

ST-DDT:
Do not force push your changes as it complicates the review process.

i509VCB:
> 
> 
> Do not force push your changes as it complicates the review process.

Will note that.

gabizou (changes requested):
Run a `licenseFormat` and it looks good to me.

src/main/java/org/spongepowered/api/raid/Wave.java
```diff
@@ -40,13 +42,21 @@
...
+     * <p>Bonus waves exist because {@link PotionEffectTypes#BAD_OMEN} had a level greater than 1.
+     *
      * @return True if bonus, false if not
      */
     boolean isBonus();
 
     /**
      * Determines if this wave was the final wave of the {@link Raid}.
      *
+     * <p>Weather a wave is final depends on the {@link Difficulty} of the world.
```
ST-DDT: Typo: whether

src/main/java/org/spongepowered/api/event/item/merchant/TradeWithMerchantEvent.java
```diff
@@ -0,0 +1,66 @@
...
+public interface TradeWithMerchantEvent extends Event, Cancellable {
+
+	/**
+	 * Gets the {@link Merchant} which is trading.
+	 *
+	 * @return The trading merchant.
+	 */
+	Merchant getMerchant();
+
+	/**
```
Zidane, line 45: It would be nice to have the original TradeOffer here.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,135 @@
...
+import org.spongepowered.api.raid.Raid;
+import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
+
+/**
+ * An event which involves a {@link Villager}.
+ */
+public interface VillagerEvent extends Event {
+
+	/**
+	 * Gets the {@link Villager} involved with this event.\
```
Zidane: \

src/main/java/org/spongepowered/api/event/item/merchant/TradeWithMerchantEvent.java
ST-DDT: I'm not sure whether the `TradeOffer` should be settable in this event.

Zidane: @ST-DDT Would you care to share with the class why?

ST-DDT: It's hard to explain. I dont see a reason for that. You are in the middle of the trade. Why would the trade change at this point. If this were something like `SelectTradeOfferEvent` it would make sense, but not while actually doing it.

Edit: Is the server expected to verify that the requirements for the old or the new trade offer is met? What happens if the player no longer meets the requirement will the event be cancelled afterwards? If not how will the missing resources be withdrawn?

i509VCB: Considering that case you mentioned, I am not too sure if you can make the client forcefully select a TradeOffer. Maybe it would be best to make the TradeOffer not settable during the event.

src/main/java/org/spongepowered/api/event/item/merchant/TradeWithMerchantEvent.java
```diff
@@ -0,0 +1,59 @@
...
+	 * Gets the selected {@link TradeOffer} a {@link Humanoid} could buy.
+	 *
+	 * @return The trade offer.
+	 */
+	TradeOffer getSelectedOffer();
```
ST-DDT: Maybe we should name this method `getTradeOffer` and change the description to `the trade offer being traded` to avoid confusion about whether the trade offer is selected or traded.

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
```diff
@@ -147,4 +150,23 @@ default UUID getUniqueId() {
 
     @Override
     Collection<ServerPlayer> getPlayers();
+
+    /**
+     * Gets all {@link Raid}s occuring in this {@link ServerWorld}
```
ST-DDT: Missing dot at EOL

ST-DDT:
Only these last minor issues otherwise it looks good to me.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,135 @@
...
+		 */
+		Profession getOriginalProfession();
+
+		/**
+		 * Gets the {@link Profession} the villager will change to.
+		 *
+		 * @return The {@link Villager}'s next {@link Profession}.
+		 */
+		Profession getNextProfession();
+
```
Zidane, line 72: Remove the next in the naming.

src/main/java/org/spongepowered/api/event/entity/living/trader/VillagerEvent.java
```diff
@@ -0,0 +1,135 @@
...
+		 * @return The {@link Profession} level.
+		 */
+		int getOriginalProfessionLevel();
+
+		/**
+		 * Gets the {@link Villager}'s next {@link Profession} level.
+		 *
+		 * @return The {@link Profession} level.
+		 */
+		int getNextProfessionLevel();
```
Zidane: Remove the next in the naming.

src/main/java/org/spongepowered/api/event/item/merchant/TradeWithMerchantEvent.java
Zidane: If we can't reasonably set the TradeOffer then just do getTradeOffer as he said above.

PR #2078, opened 2020-01-09
Add useful methods to RailDirection and LogAxis interfaces

### LINKED ISSUE #2026: Rotating of blocks (issue with RailDirection)
Hello! In my project, I needed to rotate some objects from blocks, that is, what the `//rotate` command of WorldEdit does, too. I have already found a way to transform the coordinates of blocks, however, some blocks have their own rotation (for example, a furnace or rails). Some blocks can be rotated right now, some can only be in API 8 (1.14), and some have problems.

Here is a small list of data that needs to be changed for the blocks that have them in order to rotate structures correctly (if I forgot something, please remind):

- **AxisData** - there are no problems.
- **ConnectedDirectionData** - here too. The `Direction` class is very well designed.
- **DirectionalData** - obviously everything is okay too.
- **LogAxisData** - logs can be rotated using `Axis` in API 8, although it would be nice to add a method `Optional<Axis> LogAxis#toAxis()` and `LogAxis#fromAxis(Axis axis)` to API 7.
- **RailDirectionData** - and here there are problems. The `RailDirection` class does not have any useful methods. Transforming of this data is not possible. There is an option to write own conversion code, but it will turn out to be very cumbersome and not beautiful.
- **WireAttachmentData** - I think there will be no problems.

As you can see the main problems with `RailDirection`... ~~Will it be possible to somehow improve this class in API 8 or maybe even in API 7?~~

Lignium:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2474)

Added several missing methods, without which it is impossible to transform rails and logs.

Resolves: https://github.com/SpongePowered/SpongeAPI/issues/2026

### DIFF
--- a/src/main/java/org/spongepowered/api/data/type/LogAxis.java
+++ b/src/main/java/org/spongepowered/api/data/type/LogAxis.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.api.data.type;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.util.Axis;
 import org.spongepowered.api.util.Cycleable;
@@ -35,8 +37,29 @@ import java.util.Optional;
 public interface LogAxis extends CatalogType, Cycleable<LogAxis> {
 
     /**
+     * Converts the given {@link Axis} to the {@link LogAxis} equivalent.
+     *
+     * @param axis The axis
+     * @return The {@link LogAxis}
+     */
+    static LogAxis fromAxis(Axis axis) {
+        checkNotNull(axis, "axis");
+
+        switch (axis) {
+            case X:
+                return LogAxes.X;
+            case Y:
+                return LogAxes.Y;
+            case Z:
+                return LogAxes.Z;
+            default:
+                throw new AssertionError();
+        }
+    }
+
+    /**
      * Returns the {@link Axis} from this log axis, if it is not {@link LogAxes#NONE}.
-     * 
+     *
      * @return The {@link Axis}, if possible
      */
     Optional<Axis> toAxis();
--- a/src/main/java/org/spongepowered/api/data/type/LogAxis.java
+++ b/src/main/java/org/spongepowered/api/data/type/LogAxis.java
@@ -25,10 +25,19 @@
 package org.spongepowered.api.data.type;
 
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.Axis;
 import org.spongepowered.api.util.Cycleable;
 import org.spongepowered.api.util.annotation.CatalogedBy;
+
+import java.util.Optional;
 
 @CatalogedBy(LogAxes.class)
 public interface LogAxis extends CatalogType, Cycleable<LogAxis> {
 
+    /**
+     * Returns the {@link Axis} from this log axis, if it is not {@link LogAxes#NONE}.
+     * 
+     * @return The {@link Axis}, if possible
+     */
+    Optional<Axis> toAxis();
 }
--- a/src/main/java/org/spongepowered/api/data/type/RailDirection.java
+++ b/src/main/java/org/spongepowered/api/data/type/RailDirection.java
@@ -24,7 +24,10 @@
  */
 package org.spongepowered.api.data.type;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.util.Cycleable;
 import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.util.annotation.CatalogedBy;
@@ -35,7 +38,61 @@ import java.util.Optional;
 public interface RailDirection extends CatalogType, Cycleable<RailDirection> {
 
     /**
-     * Returns the ascending direction, if this rail direction is ascending.
+     * Returns the {@link RailDirection} from the given facing directions.
+     *
+     * @param firstDirection The first direction
+     * @param secondDirection The second direction
+     * @return The {@link RailDirection}
+     * @throws IllegalArgumentException If the given directions don't differ
+     */
+    static RailDirection fromFaces(Direction firstDirection, Direction secondDirection) {
+        checkNotNull(firstDirection, "firstDirection");
+        checkNotNull(secondDirection, "secondDirection");
+
+        Direction cardinalFirstDirection = Direction.getClosest(firstDirection.asOffset(), Direction.Division.CARDINAL);
+        Direction cardinalSecondDirection = Direction.getClosest(secondDirection.asOffset(), Direction.Division.CARDINAL);
+
+        if (cardinalFirstDirection.equals(cardinalSecondDirection)) {
+            throw new IllegalArgumentException("The directions should be different!");
+        }
+
+        for (RailDirection currentRailDirection: Sponge.getRegistry().getAllForMinecraft(RailDirection.class)) {
+            if (currentRailDirection.isFacing(cardinalFirstDirection) && currentRailDirection.isFacing(cardinalSecondDirection)) {
+                return currentRailDirection;
+            }
+        }
+
+        throw new AssertionError();
+    }
+
+    /**
+     * Returns the {@link RailDirection} from the given ascending direction.
+     *
+     * @param ascendingDirection The ascending direction
+     * @return The {@link RailDirection}
+     */
+    static RailDirection fromAscendingDirection(Direction ascendingDirection) {
+        checkNotNull(ascendingDirection, "ascendingDirection");
+
+        Direction cardinalAscendingDirection = Direction.getClosest(ascendingDirection.asOffset(), Direction.Division.CARDINAL);
+
+        for (RailDirection currentRailDirection: Sponge.getRegistry().getAllForMinecraft(RailDirection.class)) {
+            Optional<Direction> currentAscendingDirectionBox = currentRailDirection.getAscendingDirection();
+
+            if (currentAscendingDirectionBox.isPresent()) {
+                Direction currentAscendingDirection = currentAscendingDirectionBox.get();
+
+                if (currentAscendingDirection.equals(cardinalAscendingDirection)) {
+                    return currentRailDirection;
+                }
+            }
+        }
+
+        throw new AssertionError();
+    }
+
+    /**
+     * Returns the direction it's ascending to, if this rail direction is ascending.
      *
      * @return The ascending direction, if possible
      */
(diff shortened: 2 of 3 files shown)


src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -26,9 +26,43 @@
...
+     * @return The ascending direction, if possible
+     */
+    Optional<Direction> getAscendingDirection();
+
+    /**
+     * Checks if this rail direction is ascending.
+     *
+     * @return <b>true</b> if is ascending, else <b>false</b>
+     */
+    default boolean isAscending() {
```
Zidane: No reason for the default, can simply do an isPresent().

Lignium: Should I remove default or remove this method?

i509VCB: Since ascending direction is optional, remove the default method `isAscending`

src/main/java/org/spongepowered/api/util/Axis.java
```diff
@@ -132,6 +134,23 @@ public Vector3d toVector3dWithDirection(final AxisDirection axisDirection) {
         return this.direction.mul(axisDirection.getSignum());
     }
 
+    /**
+     * Returns the {@link LogAxis} from this axis.
+     * 
+     * @return The {@link LogAxis}
+     */
+    public LogAxis toLogAxis() {
+        if (equals(Axis.X)) {
```
ST-DDT: Maybe do a switch case here.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -26,9 +26,34 @@
...
+
 @CatalogedBy(RailDirections.class)
 public interface RailDirection extends CatalogType, Cycleable<RailDirection> {
 
+    /**
+     * Returns the ascending direction, if this rail direction is ascending.
+     *
+     * @return The ascending direction, if possible
+     */
+    Optional<Direction> getAscendingDirection();
```
ST-DDT, line 98: Maybe instead of using an `Optional`, we should use just a `boolean` for `isAscending`/`isSlope`, because there is no `DOWN` here.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -26,9 +26,34 @@
...
+     */
+    Direction getFirstDirection();
+
+    /**
+     * Returns the second direction of this rail direction.
+     *
+     * @return The second direction
+     * @see RailDirection#getFirstDirection()
+     */
+    Direction getSecondDirection();
```
ST-DDT, line 114: Maybe add a method like `isFacing(Direction) : boolean` to avoid the trouble to check both direction components in the own code.

src/main/java/org/spongepowered/api/util/Axis.java
ST-DDT: Also not sure whether this should be here or inside the `LogAxes` class

Cybermaxke: I would put this in `LogAxes` or `LogAxis`.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
ST-DDT: Never mind, I misunderstood the method. Maybe clarify it a bit in the javadocs, that this is the direction it's ascending to.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
Lignium: Good idea! I also thought of adding something like this here, this will really reduce a condition code in a plugin.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
 import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.util.Cycleable;
+import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 
+import java.util.Optional;
+
 @CatalogedBy(RailDirections.class)
 public interface RailDirection extends CatalogType, Cycleable<RailDirection> {
```
dualspiral, line 38: This class needs a javadoc explaining what it represents. I'm also not sure about the name `RailDirection` because it doesn't really represent the direction of the rails. It represents which edge the two rail piece termini are at, not really the direction of the rail itself.

Would like others' input on this.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+            throw new IllegalArgumentException("The directions should be different!");
+        }
+
+        for (RailDirection currentRailDirection: Sponge.getRegistry().getAllForMinecraft(RailDirection.class)) {
+            if (currentRailDirection.isFacing(cardinalFirstDirection) && currentRailDirection.isFacing(cardinalSecondDirection)) {
+                return currentRailDirection;
+            }
+        }
+
+        throw new AssertionError();
```
dualspiral: No, never throw an `AssertionError` - this should be a `IllegalArgumentException`.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+     * @param firstDirection The first direction
+     * @param secondDirection The second direction
+     * @return The {@link RailDirection}
+     * @throws IllegalArgumentException If the given directions don't differ
+     */
+    static RailDirection fromFaces(final Direction firstDirection, final Direction secondDirection) {
+        checkNotNull(firstDirection, "firstDirection");
+        checkNotNull(secondDirection, "secondDirection");
+
+        Direction cardinalFirstDirection = Direction.getClosest(firstDirection.asOffset(), Direction.Division.CARDINAL);
```
dualspiral: If you post an invalid direction, it's an exception, don't try to correct for them.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+            if (currentAscendingDirectionBox.isPresent()) {
+                Direction currentAscendingDirection = currentAscendingDirectionBox.get();
+
+                if (currentAscendingDirection.equals(cardinalAscendingDirection)) {
+                    return currentRailDirection;
+                }
+            }
+        }
+
+        throw new AssertionError();
```
dualspiral: As above.

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+    }
+
+    /**
+     * Returns the {@link RailDirection} from the given ascending direction.
+     *
+     * @param ascendingDirection The ascending direction
+     * @return The {@link RailDirection}
+     */
+    static RailDirection fromAscendingDirection(final Direction ascendingDirection) {
```
dualspiral, line 71: This feels very confusing to me - and the javadocs aren't particularly informative as to why there is one direction, not two.

`fromAscendingDirection` doesn't really make any sense from a language point of view. I know what you're trying to say, that you are defining the direction where the rail terminus is at the top of the block, not the bottom, and that by Minecraft logic it must be the opposite direction which is the "direction" of the bottom terminus.

`withUpperDirection`/`withUpperAxis`?

Opinions?

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+
+        throw new AssertionError();
+    }
+
+    /**
+     * Returns the direction it's ascending to, if this rail direction is ascending.
+     *
+     * @return The ascending direction, if possible
+     */
+    Optional<Direction> getAscendingDirection();
```
dualspiral, line 98: Should we just update `Direction` to make it a true 3D directional enum (so, `UP_NORTH` for example) and do away with this method? Alternatively, should we have a separate `RailAxis` instead that have a combination that contains a combination of `Direction`s?

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
 
+    /**
+     * Returns the {@link RailDirection} from the given facing directions.
+     *
+     * @param firstDirection The first direction
+     * @param secondDirection The second direction
+     * @return The {@link RailDirection}
+     * @throws IllegalArgumentException If the given directions don't differ
+     */
+    static RailDirection fromFaces(final Direction firstDirection, final Direction secondDirection) {
```
dualspiral, line 48: I do not like these static methods having impl code. Can we create a builder and have these methods call this builder?

src/main/java/org/spongepowered/api/data/type/RailDirection.java
```diff
@@ -24,11 +24,103 @@
...
+    /**
+     * Returns the second direction of this rail direction.
+     *
+     * @return The second direction
+     * @see RailDirection#getFirstDirection()
+     */
+    Direction getSecondDirection();
+
+    /**
+     * Checks if the rail direction is facing to the given direction.
```
dualspiral: Get rid of `to`.

dualspiral (changes requested):
I think there needs to be a bit more discussion about how this all comes together. Javadocs need improving.

Zidane:
@dualspiral Yay or nay?

dualspiral:
No, too many issues as it stands and I'm not sure how much of an impact it'll make.

src/main/java/org/spongepowered/api/data/type/LogAxis.java
```diff
@@ -24,11 +24,43 @@
...
+
+        switch (axis) {
+            case X:
+                return LogAxes.X;
+            case Y:
+                return LogAxes.Y;
+            case Z:
+                return LogAxes.Z;
+            default:
+                throw new AssertionError();
```
dualspiral, line 56: No assertion errors...

dualspiral (commented):
The more I look at this, the more I think it needs redoing. There are many corner cases here, code in API that should be in impl, and badly named methods.

I will take some blame for some of the stagnation, I asked questions of others and got no answer - perhaps should have poked harder. However, this really cannot be accepted as is.

Part of the problem is that there are static methods everywhere that shouldn't be there. As I've said, they are badly named and there are too many of them. For defining a `RailDirection`, a builder that allows you to pull out the correct type would be better, for example:

```java
RailDirection.builder().from(Direction).to(Direction).ascending().build();
```

In this chain it's clear where the rail is going from and to, and now makes sense as to which direction it's ascending as we have defined a from and to. This allows us to ditch the various factory static methods you have and also put validating logic in `build` (which throws an `IllegalStateException` if such validation fails), in common with other builders, and crucially, it's in impl. The only thing that I kind of umm and ahh about is the fact it's not really a builder but a selector because we're not creating a type but grabbing the appropriate `CatalogType`. Maybe it should be:

```java
RailDirection.selector().from(Direction).to(Direction).ascending().select();
```

@Zidane, any thoughts on that last bit on naming, given it's a "builder" that doesn't build, but selects.

Then, your documentation for `RailDirection` can state that if `isAscending` is `true`, then the second direction is above the first, and you don't need to return an optional, a boolean suffices. If you can do that, then we'll revisit from there.

Also, please avoid static imports.

Lignium:
Given the fact that in API 8 it became possible to easily transform any blocks, including from modifications, this PR now does not make much sense, so I will close it (and the implementation PR).

PR #2079, opened 2020-01-10
Move Spellcaster getters to keys and add missing value accessor for enderman screaming

i509VCB:


Zidane:
Merged manually. Thanks as always i50.

PR #2081, opened 2020-01-14
Add DataValues to the BlockEntities where they belong. [API-8]

i509VCB:
This puts all the Value.Mutable blah blah blah in the BlockEntity classes. This is here to check incase I've added Values to things that don't apply to BlockEntities or some blatantly missing ones are not there.

~~So far there are two obvious values I have not done yet (DISPLAY_NAME for BlockEntities) and all of the `STRUCTURE` keys due to the optionality of the methods depending on the specific state of the Structure Block.~~

Renames `PlayerHead` -> `Skull` to match the vanilla impl since all heads/skulls share the same BlockEntityType. Also the `representedPlayer` data value is now `Optional`.

Renames `ComparatorType` -> `ComparatorMode` as per Zidane's request.

Changes `Keys#REPRESENTED_PLAYER` to a simple `Keys#GAME_PROFILE` for the Skull.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/entity/Banner.java
+++ b/src/main/java/org/spongepowered/api/block/entity/Banner.java
@@ -33,7 +33,7 @@ import org.spongepowered.api.data.value.Value;
 /**
  * Represents a Banner {@link BlockEntity}.
  */
-public interface Banner extends BlockEntity {
+public interface Banner extends NameableBlockEntity {
 
     /**
      * Gets the {@link org.spongepowered.api.data.value.Value.Mutable} for the base {@link DyeColor}.
--- a/src/main/java/org/spongepowered/api/block/entity/Comparator.java
+++ b/src/main/java/org/spongepowered/api/block/entity/Comparator.java
@@ -24,9 +24,20 @@
  */
 package org.spongepowered.api.block.entity;
 
+import org.spongepowered.api.data.Keys;
+import org.spongepowered.api.data.type.ComparatorMode;
+import org.spongepowered.api.data.value.Value;
+
 /**
  * Represents a Redstone Comparator.
  */
 public interface Comparator extends BlockEntity {
 
+    /**
+     * {@link Keys#COMPARATOR_MODE}
+     * @return The mode of the Comparator.
+     */
+    default Value.Mutable<ComparatorMode> mode() {
+        return this.requireValue(Keys.COMPARATOR_MODE).asMutable();
+    }
 }
--- a/src/main/java/org/spongepowered/api/block/entity/Piston.java
+++ b/src/main/java/org/spongepowered/api/block/entity/Piston.java
@@ -24,6 +24,10 @@
  */
 package org.spongepowered.api.block.entity;
 
+import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.data.Keys;
+import org.spongepowered.api.data.value.Value;
+
 /**
  * Represents a piston moving in the world.
  *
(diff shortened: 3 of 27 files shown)


src/main/java/org/spongepowered/api/block/entity/StructureBlock.java
```diff
@@ -34,4 +43,88 @@
...
+     */
+    default Value.Mutable<Double> integrity() {
+        return this.requireValue(Keys.STRUCTURE_INTEGRITY).asMutable();
+    }
+
+    /**
+     * {@link Keys#STRUCTURE_AUTHOR}
+     * @return The author of the structure loaded by the structure block.
+     */
+    default Optional<Value.Mutable<String>> structureAuthor() {
```
Zidane: Just author will do

src/main/java/org/spongepowered/api/block/entity/Banner.java
```diff
@@ -33,7 +33,7 @@
 /**
  * Represents a Banner {@link BlockEntity}.
  */
-public interface Banner extends BlockEntity {
+public interface Banner extends BlockEntity, NameableBlockEntity {
```
i509VCB: Discord comment:

Banner and enchantment table only need to extend NameableBlockEntity because of inheritance

src/main/java/org/spongepowered/api/block/entity/Comparator.java
```diff
@@ -24,9 +24,20 @@
...
 /**
  * Represents a Redstone Comparator.
  */
 public interface Comparator extends BlockEntity {
 
+    /**
+     * {@link Keys#COMPARATOR_TYPE}
+     * @return The type of the Comparator.
+     */
+    default Value.Mutable<ComparatorType> type() {
```
Zidane: mode

Rename ComparatorType to ComparatorMode

src/main/java/org/spongepowered/api/block/entity/Skull.java
```diff
@@ -0,0 +1,55 @@
...
+     * <p>Gets the current represented player {@link GameProfile} for this
+     * {@link Skull}.</p>
+     *
+     * <p>This value is only present when the Skull is bound to a
+     * {@link BlockTypes#PLAYER_HEAD} or
+     * {@link BlockTypes#PLAYER_WALL_HEAD}.</p>
+     *
+     * @return The current represented player value
+     */
+    default Optional<Value.Mutable<GameProfile>> representedPlayer() {
```
Zidane: Rename to gameProfile

Rename key to GAME_PROFILE

src/main/java/org/spongepowered/api/block/entity/Piston.java
```diff
@@ -33,4 +37,19 @@
...
+     * {@link Keys#PISTON_TYPE}
+     * @return The type of this piston.
+     */
+    default Value.Mutable<PistonType> type() {
+        return this.requireValue(Keys.PISTON_TYPE).asMutable();
+    }
+
+    /**
+     * {@link Keys#EXTENDED}
+     * @return Whether this piston is extended.
```
gabizou: Note that this is somewhat of a misnomer. The `TileEntity` is only going to exist while the piston is extending or retracting. It'd be better to explain that while a `Piston` can be sticky or normal, the `Piston` as a `BlockEntity` instance is serving only to produce "movement" of the blocks being pushed (I know this because I've had to work around these sort of changes in the world with the enhanced tracker).

gabizou (commented):
I just want to have some discussion about the `Piston` `BlockEntity` as it serves not as the piston itself, but the blocks being moved in the process of the piston (the 30 ticks or so that it takes to move i think).

Other than that, nothing else is wrong that I can see.

src/main/java/org/spongepowered/api/block/entity/Piston.java
i509VCB: So what I've thought from that statement is the piston type doesn't really apply in this case. A look at the wiki for the data structure does show an extended value (which we have)  and I don't really see a is sticky attribute. So remove that value from the block entity? (of course the piston block type would still accept that key)

i509VCB: Ive removed the PistonType from the block entity and added a data value for getting the state which the piston block entity represents.

PR #2085, opened 2020-02-02
Add Entity Attribute API [Redux of #1428] [API-8]

i509VCB:
Definitely not complete or structured entirely. A few things definitely need to be moved around into proper packages.

The system is based off this:
http://minecraft.gamepedia.com/Attribute

Replaces #1428

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/attribute/type/AttributeType.java
+++ b/src/main/java/org/spongepowered/api/entity/attribute/type/AttributeType.java
@@ -1,0 +1,60 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.entity.attribute.type;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.entity.attribute.AttributeModifier;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+import java.util.Optional;
+
+/**
+ * Represents an attribute type that can apply effects or modify traits related to an entity.
+ */
+@CatalogedBy(AttributeTypes.class)
+public interface AttributeType extends CatalogType {
+
+    /**
+     * Gets the parent attribute type of this attribute type.
+     *
+     * @return The parent type if present.
+     */
+    Optional<AttributeType> getParent();
+
+    /**
+     * Gets the default value of this attribute type before any {@link AttributeModifier}s are applied.
+     *
+     * @return The default value of this attribute type.
+     */
+    double getDefaultValue();
+
+    /**
+     * Clamps a value to be within the bounds of this attribute type.
+     *
+     * @param value The value to clamp
+     * @return A value within this attribute type's bounds.
+     */
+    double clampValue(double value);
+}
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/api/entity/attribute/type/AttributeOperation.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.entity.attribute.Attribute;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * Represents an operation which can be applied to transform the value of an
+ * {@link Attribute}.
+ */
+@CatalogedBy(AttributeOperations.class)
+public interface AttributeOperation extends CatalogType {
```
gabizou: Should we expose `AttributeOperation` as a `DoubleFunction` as well? Or a `DoubleBiFunction` of sorts?

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeTypes.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public static final Supplier<RangedAttributeType> KNOCKBACK_RESISTANCE = Sponge.getRegistry().getCatalogRegistry().provideSupplier(RangedAttributeType.class, "KNOCKBACK_RESISTANCE");
+
+    public static final Supplier<RangedAttributeType> LUCK = Sponge.getRegistry().getCatalogRegistry().provideSupplier(RangedAttributeType.class, "LUCK");
+
+    public static final Supplier<RangedAttributeType> MAX_HEALTH = Sponge.getRegistry().getCatalogRegistry().provideSupplier(RangedAttributeType.class, "MAX_HEALTH");
+
+    public static final Supplier<RangedAttributeType> MOVEMENT_SPEED = Sponge.getRegistry().getCatalogRegistry().provideSupplier(RangedAttributeType.class, "MOVEMENT_SPEED");
+
+    public static final class Horse {
```
gabizou: I'd rather the fields be prefixed with their intended targets rather than being nested.

I get that `AttributeTypes.Horse.JUMP_STRENGTH` is a nice way of writing it, but it's also a little cluttered looking at it without the prefix (otherwise, we'd be stylistically needing to do the same treatment to `Keys` etc.

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeType.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Optional;
+
+/**
+ * Represents an attribute type that can apply effects or modify traits related to an entity.
+ */
+@CatalogedBy(AttributeTypes.class)
+public interface AttributeType extends CatalogType {
+
+    Optional<AttributeType> getParent();
```
gabizou, line 44: Would it make sense to expose what the `AttributeType`'s related `Key` is?

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -164,6 +170,39 @@ static ItemStack empty() {
...
+    /**
+     * Gets all {@link AttributeModifier}s on this item stack.
+     *
+     * @param attributeType The {@link AttributeType} of the modifier.
+     * @param equipmentType The {@link EquipmentType} this modifier is applied
+     * to.
+     *
+     * @return A collection of {@link AttributeModifier}s.
+     */
+    default Collection<AttributeModifier> getAttributeModifiers(Supplier<? extends AttributeType> attributeType, Supplier<? extends EquipmentType> equipmentType) {
```
gabizou, line 203: Need to overload with both as optionally suppliers (two added methods).

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -164,6 +170,39 @@ static ItemStack empty() {
...
+     * @return A collection of {@link AttributeModifier}s.
+     */
+    Collection<AttributeModifier> getAttributeModifiers(AttributeType attributeType, EquipmentType equipmentType);
+
+    /**
+     * Adds an {@link AttributeModifier} to this item stack.
+     * @param attributeType The attribute ty
+     * @param modifier
+     */
+    void addAttributeModifier(AttributeType attributeType, AttributeModifier modifier);
```
gabizou: Needs an overload with Supplier for AttributeType.

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeOperation.java
i509VCB: Looking at the code again mojang seems to store all the modifiers by their operations within the EntityAttributeInstance [Attribute in API] (sorry just like using yarn names). 

It could be done but since the game seems to calculate from ADD -> MULTIPLY_BASE -> MULTIPLY_TOTAL the predictions might be different depending on which operation is being added into the modifier and where it sits within the list of Modifiers on the attribute.

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeType.java
i509VCB: Could you possibly elaborate on this idea?

src/main/java/org/spongepowered/api/entity/attribute/type/AttributeType.java
```diff
@@ -0,0 +1,68 @@
...
+     * @return The parent type if present.
+     */
+    Optional<AttributeType> getParent();
+
+    /**
+     * Gets the key this attribute type is related to.
+     *
+     * @return The key if any key is related to this attribute type.
+     */
+    Optional<Key<?>> getAttributeKey();
```
Zidane: Why? Seems like we're mixing concepts here.

i509VCB: Might have misunderstood an earlier comment gabi left, I'll look at it again

i509VCB: Okay this was related to an earlier question:
> Would it make sense to expose what the AttributeType's related Key is?

So I've thought it over and decided that exposing the related key isn't necessary since the two systems are completely separate and unrelated.

PR #2091, opened 2020-02-24
Add supplier overloads to all builder methods and trait -> property

i509VCB:
Adds Supplier overloads in many methods for CatalogTypes, this one mainly covering builders. Of course more of these commits will come along in time to handle all the Supplier overloads.

### DIFF
--- a/src/main/java/org/spongepowered/api/advancement/DisplayInfo.java
+++ b/src/main/java/org/spongepowered/api/advancement/DisplayInfo.java
@@ -31,6 +31,8 @@ import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 import org.spongepowered.api.text.Text;
 import org.spongepowered.api.util.CopyableBuilder;
 
+import java.util.function.Supplier;
+
 public interface DisplayInfo {
 
     /**
@@ -110,6 +112,17 @@ public interface DisplayInfo {
          * @param advancementType The advancement type
          * @return This builder, for chaining
          */
+        default Builder type(Supplier<? extends AdvancementType> advancementType) {
+            return this.type(advancementType.get());
+        }
+
+        /**
+         * Sets the {@link AdvancementType}. Defaults
+         * to {@link AdvancementTypes#TASK}.
+         *
+         * @param advancementType The advancement type
+         * @return This builder, for chaining
+         */
         Builder type(AdvancementType advancementType);
 
         /**
@@ -135,8 +148,19 @@ public interface DisplayInfo {
          * @param itemType The item type
          * @return This builder, for chaining
          */
+        default Builder icon(Supplier<? extends ItemType> itemType) {
+            return icon(itemType.get());
+        }
+
+        /**
+         * Sets the icon of the advancement with the
+         * specified {@link ItemType}.
+         *
+         * @param itemType The item type
+         * @return This builder, for chaining
+         */
         default Builder icon(ItemType itemType) {
-            return icon(ItemStack.of(itemType, 1));
+            return this.icon(ItemStack.of(itemType, 1));
         }
 
         /**
@@ -147,7 +171,7 @@ public interface DisplayInfo {
          * @return This builder, for chaining
          */
         default Builder icon(ItemStack itemStack) {
-            return icon(itemStack.createSnapshot());
+            return this.icon(itemStack.createSnapshot());
         }
 
         /**
--- a/src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
+++ b/src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
@@ -32,14 +32,15 @@ import org.spongepowered.api.util.CopyableBuilder;
 
 import java.util.List;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 /**
  * A {@link BlockState} matcher that will match various block states
  * according to a pre-built list of {@link StateProperty}s and their
  * values, such that not all {@link StateProperty}s contained in a
  * {@link BlockState} must be matched. (Such as if a block state
- * that contains 4 traits, and only 2 are wanting to be matched,
- * then the other two traits may be variable).
+ * that contains 4 properties, and only 2 are wanting to be matched,
+ * then the other two properties may be variable).
  */
 public interface BlockStateMatcher extends Predicate<BlockState>, DataSerializable {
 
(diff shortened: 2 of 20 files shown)


src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -128,6 +128,10 @@ static Builder builder() {
      */
     boolean doesGrantExperience();
 
+    float getPriceGrowthRate();
```
gabizou: Can this be a `double` instead? And can you javadoc it?

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -186,6 +190,24 @@ static Builder builder() {
...
+         */
+        Builder grantsExperience(int experience);
+
+        /**
+         * Sets the rate at which this trade offer's price will grow by when used.
+         *
+         * <p>By default this multiplier is 0.05</p>
+         *
+         * @param priceGrowthRate The offer's rate of price growth.
+         * @return This builder.
```
gabizou: Extra `.`

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -186,6 +190,24 @@ static Builder builder() {
...
+         * @return This builder
+         */
+        Builder grantsExperience(int experience);
+
+        /**
+         * Sets the rate at which this trade offer's price will grow by when used.
+         *
+         * <p>By default this multiplier is 0.05</p>
+         *
+         * @param priceGrowthRate The offer's rate of price growth.
```
gabizou: Extra `.`

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -186,6 +190,24 @@ static Builder builder() {
...
+         *
+         * @param experience The amount of experience.
+         * @return This builder
+         */
+        Builder grantsExperience(int experience);
+
+        /**
+         * Sets the rate at which this trade offer's price will grow by when used.
+         *
+         * <p>By default this multiplier is 0.05</p>
```
gabizou: If you're going to link modifiers like this, complete the sentence, and possibly link the MC wiki where they explain this.

src/main/java/org/spongepowered/api/fluid/FluidState.java
```diff
@@ -85,6 +87,21 @@ static Builder builder() {
      */
     interface Builder extends SerializableDataHolderBuilder.Immutable<FluidState, Builder> {
 
+        /**
+         * Sets the {@link FluidType} for the {@link FluidState} to build.
+         *
+         * <p>The {@link FluidType} is used for some pre-validation on addition of
+         * {@link DataManipulator}s through {@link #add(DataManipulator)}. It is
```
gabizou: There's no more manipulators, if there are, they are missed to be removed.

gabizou (changes requested):
Small nitpicks, but it looks like something leaked from another feature into this PR.

For the few classes that you're modifying, could you javadoc the methods that aren't javadoc'ed already?

As always, many thanks for your contributions

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
i509VCB: Oh yeah forgot to mention that these are to compensate for the new properties of trade offers. Also thought it would be best not do have a 4 line PR just for that

gabizou: Need to java doc/link to mc wiki for information imo if you're doing that then.

src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
```diff
@@ -92,6 +105,26 @@ default boolean test(BlockState blockState) {
...
+         *
+         * @param trait The desired block trait
+         * @param value the desired value
+         * @param <T> The type of comparable
+         * @return This builder
+         * @throws IllegalArgumentException If the block trait does not match
+         *     the block type, or if the value does not belong to the trait
+         *     with the desired block type
+         */
+        default <T extends Comparable<T>> Builder trait(Supplier<? extends StateProperty<T>> trait, T value) throws IllegalArgumentException {
```
Cybermaxke: `trait` should be renamed to `property` or `stateProperty`

src/main/java/org/spongepowered/api/advancement/DisplayInfo.java
```diff
@@ -128,6 +141,17 @@ static Builder builder() {
...
 
+        /**
+         * Sets the icon of the advancement with the
+         * specified {@link ItemType}.
+         *
+         * @param itemType The item type
+         * @return This builder, for chaining
+         */
+        default Builder icon(Supplier<? extends ItemType> itemType) {
+            return icon(itemType.get());
```
Cybermaxke, line 152: `this.icon`

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -122,12 +122,29 @@ static Builder builder() {
...
+     * @return True if using this trade offer will grant experience to the customer
      */
     boolean doesGrantExperience();
 
+    /**
+     * Gets the amount of experience granted to the merchant when this trade offer is used.
+     *
+     * @return The experience to be granted to the merchant
+     */
+    int getMerchantExperience();
```
i509VCB: @gabizou Do you have a better name for this? It's the experience granted to the merchant when a trade offer is used.

src/main/java/org/spongepowered/api/block/BlockStateMatcher.java
i509VCB: I'll look around the api and include those renames with this pr EDIT: (I went for property).

gabizou (approved):

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
gabizou: I'd say it's fine. It's the experience granted to the merchant on use. just maybe `getExperienceGrantedToMerchantOnUse` but that's verbose.

Issue #2083, opened 2020-01-31
RequiringPermission fails even though child CommandElement is optional.


alexstaeding:
Currently, supplying an argument like:
```java
GenericArguments.requiringPermission(  
    GenericArguments.optional(GenericArguments.onlyOne(GenericArguments.user(Text.of("user")))),
     "some.permission"
)
```
will always fail for a user without the permission "some.permission" because parsing fails before it checks whether a child argument is optional.

ItsDoot:
Do this instead:
```java
GenericArguments.optional(
    GenericArguments.requiringPermission(
        GenericArguments.onlyOne(GenericArguments.user(Text.of("user"))),
        "some.permission"
    )
)
```

dualspiral:
Okay, I finally had some time to consider this issue and... this issue isn't really representative of what was talked about. I recally reading the issue at the time it was made, and the conversation surrounding it. What @ItsDoot said here is correct, but it's not actually the reason I asked for an issue to be created.

The problem is to do with the usage, it'll potentially leave `[]` in the usage string. I just need to eliminate that ghost.

PR #2097, opened 2020-02-29
Add requiringPermissionWeak to GenericArguments

dualspiral:
`requirePermission` doesn't actually make an element optional, sometimes you might want that. Due to the way `optional` works, I ended up deciding that a new API method was the best way to suit the needs of people who want `requirePermission` to only parse an element if a player has a permission, else it is ignored by the parser.

Unit tests added to suit.

* Also don't put [] around usage text that is empty when wrapping element in "optional" or "optionalWeak"

See #2083


ImMorpheus (approved):
Working as intended

dualspiral:
b3236fa8281059ff30e74b59d5bcffe9bae60803

PR #2101, opened 2020-03-25
Add new features to trade offers [API8].

i509VCB:
As per Zidane's request with #2091

### DIFF
--- a/src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
+++ b/src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
@@ -122,11 +122,42 @@ public interface TradeOffer extends DataSerializable {
     boolean hasExpired();
 
     /**
-     * Gets whether this trade offer will grant experience upon usage or not.
+     * Gets whether this trade offer will grant experience to the customer upon
+     * usage or not.
      *
-     * @return True if using this trade offer will grant experience
+     * @return True if using this trade offer will grant experience to the customer
      */
     boolean doesGrantExperience();
+
+    /**
+     * Gets the amount of experience granted to the merchant when this trade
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -179,13 +196,32 @@ static Builder builder() {
...
+        /**
+         * Sets how much experience the trade offer will grant to the merchant upon use.
+         *
+         * @param experience The amount of experience.
+         * @return This builder
+         */
+        Builder merchantExperience(int experience);
+
+        /**
+         * Sets the rate at which this trade offer's price will grow when demand for an item increases.
```
ST-DDT: Maybe there should be a hint what this rate does. Does every trade increase it? Is this rate incremented by this value or multiplied by it? How does the user pay 1.05 emeralds?

ST-DDT (approved):

src/main/java/org/spongepowered/api/item/merchant/TradeOffer.java
```diff
@@ -179,13 +203,48 @@ static Builder builder() {
...
+         */
+        Builder priceGrowthMultiplier(double priceGrowthMultiplier);
+
+        /**
+         * Sets the demand bonus for this trade offer.
+         *
+         * @param bonus The offer's demand bonus
+         * @return This builder
+         */
+        Builder demandBonus(int bonus);
```
ST-DDT, line 253: There doesn't seem to be a getter for this value on the final TradeOffer.

i509VCB: Now there is.

PR #2104, opened 2020-04-04
Add SaveChunkEvent.

Zidane:
For discussion, @dualspiral .

### DIFF
--- a/src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
@@ -1,0 +1,57 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.event.world.chunk;
+
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.world.Chunk;
+
+/**
+ * Called when a {@link Chunk} is performing a save.
+ */
+public interface SaveChunkEvent extends Event {
+
+    /**
+     * Gets the position of the {@link Chunk}.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import org.spongepowered.api.world.Chunk;
+
+/**
+ * Called when a {@link Chunk} is performing a save.
+ */
+public interface SaveChunkEvent extends Event {
+
+    Vector3i getPosition();
+
+    interface Pre extends SaveChunkEvent, TargetChunkEvent, Cancellable {}
```
dualspiral, line 48: Why might you want to cancel a chunk save? Isn't that what the serialisation behaviours for?

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.world.Chunk;
+
+/**
+ * Called when a {@link Chunk} is performing a save.
+ */
+public interface SaveChunkEvent extends Event {
+
+    Vector3i getPosition();
```
dualspiral: I'm assuming that this is the chunk position - in which case I think we should name this `getChunkPosition()` to make it crystal clear that's what this is. I know we'll add javadocs to this BUT many a developer may not read the javadocs and assume.

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+    interface Pre extends SaveChunkEvent, TargetChunkEvent, Cancellable {}
+
+    interface Post extends SaveChunkEvent {}
```
dualspiral, line 56: Just read Discord, this basically means that the save is going ahead, but may not yet be on the disc. We will want to mention that in the JDs.

dualspiral (commented):
Discussion, as requested. Just trying to get a feel for what everything is about.

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
Zidane: Agreed

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
Zidane: SerializationBehavior is all or nothing. You could choose to not save certain chunks. Debate could be made either way

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
Zidane: It was but ultimately that isnt enough for Joni's usecase. He needs to be sure that file wrote to disk. Therefore this is as post as you get, the file is on the disk.

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
dualspiral: I don't have an issue with this personally - dunno if anyone else does.

Zidane: One thought I had in mind is an area of the world where any changes wouldn't be wrote. You enter this "zone" and when you leave and it unloads and loads again the zone resets.

Think "mining worlds" but no need to do more worlds, lol.

dualspiral:
For note, associated PR: https://github.com/SpongePowered/SpongeCommon/pull/2551

Is this good to go, or is there more to consider? Anyone else have any comments on this?

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
gabizou: I can see some issues with this though, if handled improperly, couldn't the chunk load having not been saved cause some entity weirdness?

src/main/java/org/spongepowered/api/event/world/chunk/SaveChunkEvent.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    interface Pre extends SaveChunkEvent, TargetChunkEvent, Cancellable {}
+
+    /**
+     * Called after the {@link Chunk} is saved. Guaranteed to exist in the chunk's
+     * storage container.
+     *
+     * Depending on the implementation, this event may be called off-thread.
+     */
+    interface Post extends SaveChunkEvent {}
```
ImMorpheus, line 56: Shouldn't this `extends TargetChunkEvent` ?

EDIT: no, it's too late

ImMorpheus (approved):

gabizou (approved):
LGTM, Can always refigure this in API 8 due to the nature of the chunk management being redone there.

PR #2107, opened 2020-04-11
Duration (ticks vs clock)

Cybermaxke:
Continuation of https://github.com/SpongePowered/SpongeAPI/pull/2007
everything in this PR will be revisited for the latest changes of the API-8 branch and the discussion there. Still WIP

The java `Duration` will be replaced by a custom one, which has two subclasses `Ticks` and `Time` (renaming is an option), explicit conversion is needed between these two durations if you want to use methods like `toSeconds` or `toTicks`. 

If you pass in a time based duration to something that in vanilla uses ticks, it will be converted to ticks based on the engine performance. If you want to ticks exactly, you'll have to pass in a tick based duration. There are methods for performance based and optimal conversions.

`Engine.getEstimatedTime`, `Engine.getEstimatedTicks` - performance based conversion
`Duration.toTicks`, `Duration.toTime` - optimal conversion, assumes every tick is 50ms


XakepSDK:
I heard that `java.time` interfaces will be sealed, but i'm not sure. So this may break compatibility with future java versions
https://openjdk.java.net/jeps/360

Pulverizer:
@Cybermaxke Would this mean all Task scheduling in Sponge will be Tick based? Or can we still use exact time?

Cybermaxke:
@Pulverizer You can still use both. E.g. you use `Duration.ofTicks` you'll get tick based scheduling, if you use `Duration.ofSeconds` you'll use time based scheduling.

Cybermaxke:
@XakepSDK Where did you hear that it will be sealed? I can't find any information about it.

XakepSDK:
@Cybermaxke amber or valhalla mailing list, but i'm not sure if it was about making these classes inline or inline and sealed

Pulverizer:
I feel that changing the wording of classes and documentation to use "cycle" instead of "tick" may confuse users that are used to calling a game cycle a "tick", eg. from using a different API.

XakepSDK:
I think it's not worth.
1. It adds maintenance burden
2. Increases API complexity
3. Nobody really will implement SpongeAPI outside of vanilla minecraft. Look at bukkit, there are several implementations like glowstone, but nobody really uses them and none of them are actually have full game features
4. If somebody will implement SpongeAPI, it's his work to interpret ticks, ticks are standardized in vanilla, so implementor can easily implement ticks mechanism on top of his engine which may have no real ticks

Cybermaxke:
I don't think it will add a maintance burden, sure you need to implement it, but after that's done you don't have a lot of maintance. 

Sure it adds a new data type, but it will make the API more clear what's expected, there are places where ticks are expected, milliseconds, seconds, etc. all represented by a long or int. This makes it clear that a duration is expected, and you can't be mistaken by the unit. It also makes the API easier to use for people who don't care about ticks and just want things to happen based on time, which I think are the most cases.

Ofc other projects like lantern will benefit from this change, but that's not the main reason to do this. Even the sponge has a realtime module that could benefit from this change.

XakepSDK:
I'm using `j.t.Duration` in my plugins. Can sponge duration class renamed to something else?

Also would be good if sponge time classes implement `java.time` classes, even if `java.time` may become sealed. I think it's too preemptive to not implement `java.time`, it adds more mental complexity for a developer.

Cybermaxke:
@XakepSDK I wouldn't mind renaming it, but I'm not able to come up with a good alternative name. And yeah, it makes sense to implement the TemporalAmount interface, but only for `Duration.Time`.

ItsDoot:
An alternative naming scheme could be `Duration` -> `TimeDuration`, then `Duration.Time` -> `TimeDuration.Real`. Then at least they don't conflict with java.time.Duration in the same class.

XakepSDK:
I can't understand how it will work.
1. Let's imagine i submit task to execute every tick using `Tick` time.
Task will be executed every tick?

2. Let's imagine i submit task to execute every second using `Time` time and current TPS is 20.
Task will be executed every 20 ticks?

3. Let's imagine i submit task to execute every second using `Time` time and current TPS is 10
Task will be executed every 10 ticks?

Cybermaxke:
@XakepSDK

1 -> Yes, since you used "ticks" with `Duration.Ticks`
2, 3 -> No, you used `Duration.Time`, so it will use the clock time, so it will be a realtime second

Ticks and real time tasks were always a thing in sponge, e.g. if you used `TaskBuilder.delayTicks` it would use tick based scheduling, if you used `TaskBuilder.delay` it was time based scheduling.

XakepSDK:
Then, i think, it would be better to:
1. Introduce `Tick` `TimeUnit`, expose as `Supplier<TickUnit>` in API, so engine can implement it
2. Implement `TickDuration` `TemporalAmount`, API like in `Duration`
3. In API, accept `TemporalAmount` everywhere, so implementation could:
3.1. Check if it's `TickDuration` and use ticks
3.2. Or calculate time manually based on `TemporalAmount`

So `Duration.of(...)` or `Period` or etc for real time clock and `TickDuration.of(...)` for engine clock

Cybermaxke:
"Ticks Are Not Time Units and Should Not Be Treated As One" is something said by @dualspiral in the discussion of the previous PR and I agree with that. That's why I want to enforce explicit conversion between normal and tick based durations, and not implementing `TemporalAmount` because that would break the explicit conversion.

Cybermaxke:
I'll rename `Duration` to `EngineDuration`, this was a suggestion I did in the previous PR.

Zidane:
Superceded by the ticks framework. Thanks for the work on this Cyber, really helped devise a compromise.

PR #2110, opened 2020-04-15
Add new functions to replace stopFilter and skipFilter

dualspiral:
These functions have been deprecated and replaced with `whilst` and `select`, along with clearer javadocs.

~~* Fix bug where a BlockRay containing a ContinueAfterFilter will give a different result upon ray reset.~~

@Zidane - I did see your comments, but I mostly used Github for the diff! This is a better thing to review.

Note that `selectWhen` is what you suggested `completeWhen` should be. In my first iteration, I got the function of `skipFilter` wrong... - `selectWhen` is meant to indicate there can be multiple selections, they just have a `BlockRay#next()` call between them.

### DIFF
--- a/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
+++ b/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
@@ -83,9 +83,11 @@ import java.util.function.Predicate;
  * at based on some metric, like transparency, block model, or even distance.
  * The standard Bukkit-like behavior for finding the target block can be
  * achieved with using {@link BlockRay#ONLY_AIR_FILTER} as the
- * {@code stopFilter}, combined with
+ * continue filter, combined with
  * {@link #continueAfterFilter(Predicate, int)} with a second argument of 1, to
- * obtain the block just after the last air.</p>
+ * enable the ray to iterate such that it can select the block just after the
+ * last air. A {@link #notAirFilter()} can then be used as the select filter
+ * to select the non-air block.</p>
  *
  * <p>To get a block ray for an entities' line of sight, use
  * <pre>{@code BlockRay.from(entity);}</pre></p>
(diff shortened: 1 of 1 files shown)


dualspiral:
Okay, so, something I've noticed in the javadocs:

https://github.com/SpongePowered/SpongeAPI/blob/f12f8e7743844b8761b890ce3e440911e97c6f38/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java#L68-L71

directly conflicts with the existance of a `reset()` method...

https://github.com/SpongePowered/SpongeAPI/blob/f12f8e7743844b8761b890ce3e440911e97c6f38/src/main/java/org/spongepowered/api/util/blockray/BlockRay.java#L189

meaning that resetting a ray and re-iterating could end up with a different result. 

~~I have fixed it for our `ContinueAfterFilter`, but if we want to do this for external filters, we might need to introduce the concept of a "resettable" predicate...~~ no, I haven't, and no, I won't, given how we're going to ask plugin authors to compose their own predicates instead...

dualspiral:
I've had a think, but I couldn't come up with a better word than "whilst" to avoid a Java keyword, unless we use "whileTrue".

@Zidane Any more thoughts on this yourself or does it look okay? Don't want to accidently let this fall into the abyss.

dualspiral:
Does anyone have any more comments on this? Want to merge this soon. Particularly the method names - `select` is fine, but currently using `whilst` as `while` is a reserved keyword. Any better ideas?

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
```diff
@@ -790,8 +927,32 @@ public Vector3d position() {
...
+     * represents <strong>any</strong> of the supplied {@link BlockType}s.
+     *
+     * @param types The type of block to allow
+     * @param <E> The extent to be applied in
+     * @return The filter instance
+     */
+    public static <E extends Extent> Predicate<BlockRayHit<E>> blockTypeFilter(final BlockType... types) {
+        return lastHit -> {
+            for (BlockType type : types) {
+                if (lastHit.getExtent().getBlockType(lastHit.getBlockX(), lastHit.getBlockY(), lastHit.getBlockZ()).equals(type)) {
```
ImMorpheus: Move `lastHit.getExtent().getBlockType(lastHit.getBlockX(), lastHit.getBlockY(), lastHit.getBlockZ())` outside the loop

ImMorpheus (changes requested):

src/main/java/org/spongepowered/api/util/blockray/BlockRay.java
dualspiral: Good spot.

ImMorpheus (approved):

PR #1162, opened 2016-03-29
[WIP] Maps

MoeHouse:
This is the API from the last PR (https://github.com/SpongePowered/SpongeAPI/pull/531), updated to match current API conventions.

For the WIP Implementation see: https://github.com/SpongePowered/SpongeCommon/pull/625

Remaining Tasks:
- [x] Document `SerializationPolicy`'s effect on saving changes to a `MapView`.

PR #2112, opened 2020-04-17
Map api

tyhdefu:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2573) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/429) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/3149)

### **Keys**
- MAP_CANVAS - Key for obtaining the MapCanvas object from a map - represents pixels. **This is the only key that causes the pixels on the map to change immediately**, if required, offer a MapCanvas.blank() if you want to blank the map etc.
- MAP_WORLD - Key for which world a map "maps".
- MAP_LOCATION - Key for where in a world a world maps.
- MAP_TRACKS_PLAYERS - Key for whether a map tracks players
- MAP_UNLIMITED_TRACKING - Key for whether a map can track players from anywhere in the world.
- MAP_LOCKED - Key for whether a map should update as a player explores it.
### **DataManipulator**
Standard stuff, has all the keys etc.
### **Event**
Added a MapInfo creation event. This is fired when a new map is added to the server. (basically when a map file is created). This may not be caused by a player, it could be caused by a plugin.
### **MapCanvas**
A Map canvas represents the pixels on a map. It is made using a builder, which has basic paint methods, or you can create a canvas from an `java.awt.Image.` The image must be 128*128, and must be directly convertable to one of the MapColorTypes, with a shade. It can also be converted to a `java.awt.Image`
### **MapColors**
Map Colors represent a fully created color, which has had shading added. A MapColorType, does not have shading added. Theres a Registry and MapColorTypes for this.


src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -1458,6 +1459,44 @@
...
+     * @see MapItemData#unlimitedTracking()
+     */
+    public static final Key<Value<Boolean>> MAP_UNLIMITED_TRACKING = DummyObjectProvider.createExtendedFor(Key.class, "MAP_UNLIMITED_TRACKING");
+
+    /**
+     * Represents the {@link Key} for the scale of a map
+     *
+     * @see MapItemData#scale()
+     */
+    public static final Key<Value<Byte>> MAP_SCALE = DummyObjectProvider.createExtendedFor(Key.class, "MAP_SCALE");
```
Cybermaxke: Use `Integer`

tyhdefu: minecraft's scale field in mapdata is only byte, so it could cause problems if we allowed more than a byte's range

Cybermaxke: Then you can make it a bounded value, which will restrict the min/max values: `Key<BoundedValue<Integer>>`
Using a byte leaks implementation details that shouldn't be in the API, this also prevent future changes and it's more conventient to work with integers.

simon816:
Is this intended to be a "full" map API with methods for creating canvases etc? If so it may be worth looking at the previous efforts for a map API here: #1162

src/main/java/org/spongepowered/api/world/map/MapStorage.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Optional;
+
+public interface MapStorage {
+    /**
+     * Gets the highest known map id if there are any
+     * May be inaccurate if maps overflow
+     * @return int Highest known map
+     */
+    Optional<Integer> getHighestMapId();
```
dualspiral: This isn't a useful method, is it? Having a highest ID of 42, for example, doesn't really mean much, especially if ID 40 doesn't exist, for example. Another implementation might also register maps differently, though there is a wider discussion to be had one that one (I don't like the idea of the int ids being exposed, but I can see why that's the case here).

src/main/java/org/spongepowered/api/world/map/MapStorage.java
```diff
@@ -0,0 +1,45 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.map;
+
+import org.spongepowered.api.data.manipulator.mutable.item.MapItemData;
+
+import java.util.Optional;
+
+public interface MapStorage {
```
dualspiral, line 38: Missing Javadocs

src/main/java/org/spongepowered/api/map/MapColor.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+package org.spongepowered.api.map;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.util.Color;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(MapColorTypes.class)
```
dualspiral: Shouldn't this be on `MapColorType` instead?

Javadocs?

src/main/java/org/spongepowered/api/map/MapColorType.java
```diff
@@ -0,0 +1,32 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.map;
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.Color;
+
+public interface MapColorType extends CatalogType {
```
dualspiral: Javadocs?

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -439,4 +440,10 @@
      * @return True if main thread, false if not
      */
     boolean isMainThread();
+
+    /**
+     * Gets the map storage for this server
+     * @return MapStorage map storage
+     */
+    Optional<MapStorage> getMapStorage();
```
dualspiral, line 453: We have this on `Server` and `World` - shouldn't it just be one or the other?

src/main/java/org/spongepowered/api/data/manipulator/immutable/item/ImmutableMapItemData.java
```diff
@@ -24,15 +24,63 @@
...
+import org.spongepowered.api.world.World;
 
 /**
  * Represents an {@link ImmutableDataManipulator} hosting the specific map
  * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
+ * Warning: This will affect all maps with this id (damage value).
  */
 public interface ImmutableMapItemData extends ImmutableDataManipulator<ImmutableMapItemData, MapItemData> {
+    /**
+     * Gets the centre of where the map refers to
```
dualspiral: As much as the inner Brit is crying for this to remain like this, all comments in US English please, so, "center".

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
```diff
@@ -0,0 +1,85 @@
...
+import org.spongepowered.api.data.manipulator.immutable.item.ImmutableMapItemData;
+import org.spongepowered.api.data.manipulator.mutable.item.MapItemData;
+import org.spongepowered.api.data.type.HandType;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.event.Cancellable;
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
+import org.spongepowered.api.world.World;
+
+public interface CreateMapEvent extends Event, Cancellable {
```
dualspiral, line 35: Javadocs

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,132 @@
...
+package org.spongepowered.api.map;
+
+import com.flowpowered.math.vector.Vector2i;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.DataView;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.data.persistence.InvalidDataException;
+
+import java.awt.*;
```
dualspiral: No star imports please

src/main/java/org/spongepowered/api/world/World.java
```diff
@@ -456,6 +457,12 @@ default void setSerializationBehavior(SerializationBehavior behavior) {
      */
     WorldStorage getWorldStorage();
 
+    /**
+     * Gets the map storage for this world.
+     * @return Optional<MapStorage>
+     */
+    Optional<MapStorage> getMapStorage();
```
dualspiral: This is on both `Server` and `World` - should it be?

src/main/java/org/spongepowered/api/world/map/MapStorage.java
```diff
@@ -0,0 +1,45 @@
...
+     * @return int Highest known map
+     */
+    Optional<Integer> getHighestMapId();
+
+    /**
+     * Gets {@link MapItemData}
+     * @param id int Map id
+     * @return MapItemData of that map id.
+     */
+    Optional<MapItemData> getMapData(int id);
```
dualspiral: This is another reason why I don't necessarily like this returning the manipulator and wonder if some intermediate object would be better. For example, how would you offer this back without a `Map` object with the correct ID to offer it to?

dualspiral (changes requested):
I have a few comments, but one bigger discussion point that I think @SpongePowered/developers might want to get involved with on a design front.

I saw the following line on a data manipulator:

```
Warning: This will affect all maps with this id (damage value).
```

I'm not sure I like this. If I apply the data to a specific map, it should change that specific map. In this case, that indicates to me that the implementation should do whatever it can to ensure that's the case. (There should also be no mention of damage values).

Either:

* We treat each map separately and if you offer data to a map, that map becomes distinct from all other maps (in impl, that'll be changing the map ID). I don't think this is right though as you might want to affect all maps at once, say, for some objective based game you run as a team with.
* We have the `Map` item and some sort of definition object, say `MapInfo`, separated, so that it's clear that there can be a many to one relationship between map and backing data. This map data then becomes this `MapInfo`, and changing the map info changes all the maps that are linked to it. You can then offer a different `MapInfo` to the map to change its ID internally.

I don't think we can have it the way currently presented to us.

There are other comments. I'm not going to look at the impl right this moment because I expect it'll have to change as this API gets hammered out.

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -1458,6 +1461,61 @@
...
+     * Represents the {@link Key} for whether a map can track
+     * a player from anywhere in the world
+     *
+     * @see MapItemData#unlimitedTracking()
+     */
+    public static final Key<Value<Boolean>> MAP_UNLIMITED_TRACKING = DummyObjectProvider.createExtendedFor(Key.class, "MAP_UNLIMITED_TRACKING");
+
+    /**
+     * Represents the {@link Key} for the scale of a map
+     *
```
gabizou, line 1528: Nitpick: Maybe can link to the Minecraft wiki what scale means or how it's interpreted?

src/main/java/org/spongepowered/api/data/key/Keys.java
```diff
@@ -1458,6 +1461,61 @@
...
+    /**
+     * Represents the {@link Key} for whether a map tracks player
+     *
+     * @see MapItemData#trackPosition()
+     */
+    public static final Key<Value<Boolean>> MAP_TRACKS_PLAYERS = DummyObjectProvider.createExtendedFor(Key.class, "MAP_TRACKS_PLAYERS");
+
+    /**
+     * Represents the {@link Key} for whether a map can track
+     * a player from anywhere in the world
```
gabizou: Does this mean there's the possible tracking of a player by `UUID` specifically? Might want to also link the mc wiki for clarification and avoiding java docs having to re-explain.

src/main/java/org/spongepowered/api/data/manipulator/immutable/item/ImmutableMapItemData.java
```diff
@@ -24,15 +24,63 @@
...
+import org.spongepowered.api.data.value.mutable.Value;
 import org.spongepowered.api.item.ItemTypes;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.world.World;
 
 /**
  * Represents an {@link ImmutableDataManipulator} hosting the specific map
  * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
+ * Warning: This will affect all maps with this id (damage value).
```
gabizou: There's no understanding of damage value or id in the API, do you mean the map data itself? There's no possible way to alter an immutable data once it's created, so it shouldn't be altered by any existing map.

src/main/java/org/spongepowered/api/data/manipulator/mutable/item/MapItemData.java
```diff
@@ -24,14 +24,63 @@
...
+import org.spongepowered.api.data.value.mutable.Value;
 import org.spongepowered.api.item.ItemTypes;
 import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.world.World;
 
 /**
  * Represents an {@link DataManipulator} hosting the specific map
  * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
+ * Warning: This will affect all maps with this id (damage value).
```
gabizou: What does this warning actually mean? If you mean that a `MapData` instance can be applied/cloned to another map `ItemStack` and the mutations can affect *that other* `ItemStack`, then that's against the Data API contract that we're always returning copies of data, never the live data (unless it's impossible or by structure).

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Get the player creating this map
+     *
+     * @return Player who right clicked on a blank map
+     */
+    Player getCreator();
```
gabizou: Is this always the creator? Could anything make a new map? If anything can technically make a new map, the player would be part of the Cause of the event.

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Gets the map data for this map
```
gabizou: Is it intended that this map data is mutable? Are you suggesting that the map data itself can be modified within the event and whatever is left will be applied to the final `ItemStack`? If true, it somewhat defies what the transaction method is trying to achieve below.

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    Transaction<ItemStackSnapshot> getItemTransaction();
+
+    /**
+     * Get the id assigned to the map about to be created
+     * (This id will not exist if the event is cancelled)
+     *
+     * @return int id assigned to the new map
+     */
+    int getNewId();
```
gabizou: Does the API have to use this for any case whatsoever?

src/main/java/org/spongepowered/api/world/map/MapStorage.java
gabizou: To make a point, I do understand that we should have a Map data representation, but I feel that we do somewhat need some registry of sorts, unfortunately, Maps use just an `int` id system, whereas the API doesn't really need to adhere to the id system to retain a map, I feel we could have a world-based id system in place for the world stored maps, versus the server-sided/virtual maps that plugins create and send to clients. Would this be doable?

gabizou (changes requested):
Small documentation changes. Some design decisions need to be ironed out, but I think this is close to doable. 

To put a finer point: We still want to avoid using `int`s as a damage value -> identification system. Its somewhat legacy, and it's questionable how the maps are actually to be used. Do we want maps to be replaceable? Modifiable? Stored virtually for clients to receive, but not stored in the server's files?

tyhdefu:
(Was also posted in discord)
I have been thinking for a while about your comments and I think I have a solution. I create a MapInfo class as dualspiral suggested. This replaces a lot of the uses of the datamanipulator, and also holds the id for the map, but this will not be accessable via the api, it will purely be in the implementation. You will be able to create new MapInfo's, which will assign new ids as required, (technically it should be possible not to add a new map id until it is actually applied). There could be methods for checking if two MapInfo's refer to the same underlying data, that is, their int ids are the same, (they have been copied via crafting or copied by a plugin). You can get the MapInfo from a key, and the .getMapData(int id), will either be replaced with .getMapInfo(ItemStack itemStack), or removed. You would then be able to apply the MapInfo to an itemStack, which would change its id to the MapInfo's, unlinking it from previous ones meaning it shouldn't change ItemStacks unless they are copied/also applied the MapInfo to.
Ultimately, we need some reference to int id in the implementation, because thats how clients understand what goes where, but atleast this should eliminate any need for it in the API.

tyhdefu:
MapInfo is a DataHolder, meaning it can be offered keys, all old keys that used to apply to ItemStacks, now work on MapInfo. We still have to use a backing of int ids because we need to send packets to the clients and identify them with int ids, but they are no longer exposed in the API.

Not quite finished yet but I wanted to see if I was on the right track

Still working on MapDecorations, currently they are mutable, as it seems like a huge annoyance to repeatedly reoffer them when you're only changing cordinates (which could easily be changed by a plugin every tick). Theres also quite a few complications to work out still.

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+import org.spongepowered.api.data.value.mutable.MutableBoundedValue;
+import org.spongepowered.api.data.value.mutable.SetValue;
+import org.spongepowered.api.data.value.mutable.Value;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.map.decoration.MapDecoration;
+import org.spongepowered.api.world.World;
+
+/**
+ * Represents an {@link DataManipulator} hosting the specific map
+ * information of an {@link org.spongepowered.api.map.MapInfo}
```
gabizou:
```suggestion
 * information of an {@link org.spongepowered.api.map.MapInfo}.
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public interface MapInfoData extends DataManipulator<MapInfoData, ImmutableMapInfoData> {
+    /**
+     * Gets the center of where the map refers to
```
gabizou:
```suggestion
     * Gets the center of where the map refers to.
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     * @return Vector2d centre
```
gabizou:
```suggestion
     * @return The Vector2i center point
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    Value<Vector2i> location();
+
+    /**
+     * Gets the World that this map refers to
```
gabizou:
```suggestion
     * Gets the {@link World} that the map refers to.
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     * @return World
```
gabizou:
```suggestion
     * @return The world
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    Value<World> world();
+
+    /**
+     * Gets whether this map tracks player position
+     * @return boolean If this map tracks players
```
gabizou:
```suggestion
     * Gets whether this map tracks {@link Player} positions. A player would appear to be
     * a cursor on the map when visible and when a {@link Player} is within range.
     * 
     * @return True if this map tracks players..
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    Value<Boolean> trackPosition();
+
+    /**
+     * Gets whether this map tracks players from unlimited distance away
+     * @return boolean If the map has unlimited tracking
```
gabizou, line 63:
```suggestion
     * Gets whether this map tracks {@link Player Players} from unlimited distance away.
     * 
     * @return True If the map has unlimited player distance tracking
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    Value<Boolean> unlimitedTracking();
+
+    /**
+     * Gets the scale of this map
```
gabizou, line 68:
```suggestion
     * Gets the scale of this map.
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     * @see <a href="https://minecraft.gamepedia.com/Map#Zoom_details">Minecraft Wiki - Zoom Details</a>
+     * @return Integer The scale of this map
```
gabizou, line 70:
```suggestion
     * @return The integer scale of this map
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    MutableBoundedValue<Integer> scale();
+
+    /**
+     * Gets the canvas from this map
+     * This contains the colors in the map
+     * @return MapCanvas canvas for this map
```
gabizou:
```suggestion
     * Gets the {@link MapCanvas canvas} from this map, containing colors,
     * 
     * @return The canvas for this map
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    Value<MapCanvas> canvas();
+
+    /**
+     * Gets the MapDecorations from this map
+     * This is a list of MapDecorations
+     * @return List of MapDecorations
```
gabizou:
```suggestion
     * Gets the {@link MapDecoration MapDecorations} for this map.
     *
     * @return The list of decorations
```

src/main/java/org/spongepowered/api/data/manipulator/mutable/MapInfoData.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    SetValue<MapDecoration> decorations();
+
+    /**
+     * Gets whether this map is locked
+     * Brought forward version of <a href="https://minecraft.gamepedia.com/Map#Locking">Minecraft Wiki - Map Locking</a>
+     * (Plugins can still modify the map)
+     * @return Boolean if it is locked
```
gabizou, line 92:
```suggestion
     * Gets whether this map is locked, or whether the map is modifiable by normal
     * gameplay mechanics.
     * @see <a href="https://minecraft.gamepedia.com/Map#Locking">Minecraft Wiki - Map Locking</a>
     * 
     * @return True if the map is locked, irrelevant of plugin interactions
```

src/main/java/org/spongepowered/api/data/manipulator/immutable/item/ImmutableMapInfoItemData.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.map.MapInfo;
+
+/**
+ * A {@link ImmutableDataManipulator} representing the {@link org.spongepowered.api.map.MapInfo}
+ * of a {@link ItemStack} of type {@link ItemTypes#FILLED_MAP}
+ */
+public interface ImmutableMapInfoItemData extends ImmutableDataManipulator<ImmutableMapInfoItemData, MapInfoItemData> {
+    /**
+     * Gets the MapInfo that relates to this map
```
gabizou, line 40:
```suggestion
     * Gets the {@link MapInfo} that relates to this map.
```

src/main/java/org/spongepowered/api/data/manipulator/immutable/item/ImmutableMapInfoItemData.java
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @return {@link MapInfo} that correlates to this map
```
gabizou, line 42:
```suggestion
     * @return The map info that correlates to this map
```

src/main/java/org/spongepowered/api/data/manipulator/immutable/item/ImmutableMapInfoItemData.java
```diff
@@ -0,0 +1,45 @@
...
+import org.spongepowered.api.data.manipulator.ImmutableDataManipulator;
+import org.spongepowered.api.data.manipulator.mutable.item.MapInfoItemData;
+import org.spongepowered.api.data.value.immutable.ImmutableValue;
+import org.spongepowered.api.item.ItemTypes;
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.map.MapInfo;
+
+/**
+ * A {@link ImmutableDataManipulator} representing the {@link org.spongepowered.api.map.MapInfo}
+ * of a {@link ItemStack} of type {@link ItemTypes#FILLED_MAP}
```
gabizou, line 36:
```suggestion
 * of an {@link ItemStack} of type {@link ItemTypes#FILLED_MAP}.
```

src/main/java/org/spongepowered/api/data/manipulator/immutable/ImmutableMapInfoData.java
```diff
@@ -0,0 +1,96 @@
...
+     * This is a list of MapDecorations
+     * @return List of MapDecorations
+     */
+    ImmutableSetValue<MapDecoration> decorations();
+
+    /**
+     * Gets whether this map is locked
+     * Brought forward version of <a href="https://minecraft.gamepedia.com/Map#Locking">Minecraft Wiki - Map Locking</a>
+     * (Plugins can still modify the map)
+     * @return Boolean if it is locked
```
gabizou, line 93: Clean up java docs like the mutable variant I commented on please.

src/main/java/org/spongepowered/api/data/manipulator/mutable/item/MapInfoItemData.java
```diff
@@ -25,13 +25,18 @@
...
 /**
- * Represents an {@link DataManipulator} hosting the specific map
- * information of an {@link ItemStack} of the type {@link ItemTypes#FILLED_MAP}.
+ * A {@link DataManipulator}
  */
-public interface MapItemData extends DataManipulator<MapItemData, ImmutableMapItemData> {
+public interface MapInfoItemData extends DataManipulator<MapInfoItemData, ImmutableMapInfoItemData> {
+    /**
+     * Gets the MapInfo that relates to this map
+     * @return {@link MapInfo} that correlates to this map
```
gabizou, line 42: Same here.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,134 @@
...
+import org.spongepowered.api.map.color.MapColor;
+import org.spongepowered.api.map.color.MapColorTypes;
+
+import java.awt.Image;
+import java.util.Optional;
+
+/**
+ * Represents the canvas of a {@link org.spongepowered.api.item.inventory.ItemStack}
+ * with {@link org.spongepowered.api.item.ItemType} {@link org.spongepowered.api.item.ItemTypes#FILLED_MAP}
+ * Changes to maps will not apply until the client relogs
```
gabizou: Javadocs should be complete sentences, not fragments of a phrase, as well as being filled with correct punctuation.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,134 @@
...
+
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the map color at a location on the canvas
+     * @param x int x location
+     * @param y int y location
+     * @return MapColor color at this location
```
gabizou: You don't need to repeat the argument name or the return type, it's already implied.. You just comment on what the parameter is doing or describing it itself, or the return value, whether it's a copy, mutable, live, editable etc.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    Image toImage();
+
+    /**
+     * Creates an image and replaces all {@link MapColorTypes#AIR}
+     * with the specified color.
+     * @param color to replace the air MapColorType
+     * @return Image from the canvas
+     */
+    Image toImage(java.awt.Color color);
```
gabizou, line 94: What are the implications of using a `java.awt.Image`? IS it mutable, going to be shared, actually used by implementation? Copied?

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Called when a new map is registered.
+ * Can be caused by players or a plugin.
+ */
+public interface CreateMapEvent extends Event, Cancellable {
+    /**
+     * Gets the {@link org.spongepowered.api.map.MapInfo} for this map
+     * @return MapInfoData Map data
+     */
+    MapInfo getMapInfo();
```
gabizou, line 40: Is it possible to change the map info instance? Or are we to expect it being mutable from the get go?

src/main/java/org/spongepowered/api/map/MapInfo.java
```diff
@@ -0,0 +1,56 @@
...
+import org.spongepowered.api.item.inventory.ItemStack;
+import org.spongepowered.api.map.decoration.MapDecoration;
+
+import java.util.Map;
+
+/**
+ * Represents data of any number of linked {@link ItemStack} of type {@link ItemTypes#FILLED_MAP}
+ * @see MapInfoData
+ */
+public interface MapInfo extends DataHolder {
```
gabizou: Could you explain why you made this a `DataHolder`?

tyhdefu: I think this makes sense as a `DataHolder` because it is mutable. If it wasn't a dataholder I don't think it would fit as well in the api since not making a dataholder would mean it would not have any keys, and the idea of sponge is to have keys for pretty much everything. I think it would mean repeating code and making it fit less in the sponge "model".

Zidane:
@dualspiral See if you can wrap this up with the op.

dualspiral:
@tyhdefu I'm sorry I've been quiet, I've been trying to get an initial implementation of commands set up on 1.14.4 so that's taken all of my time. That's nearly there though (!) so I figured I'd come revisit this.

I've not attempted to run this yet, is it functional in the state it's in now? Do you have a test plugin or two that we can use to verify this?

tyhdefu:
@dualspiral It was functional with the previous features, but I was working on being able to use custom MapDecorations. I do have a test plugin, but its mostly just commands that test different features since its not that easy to test as maps are a visual thing. I will try to finish that off, clean up anything. My hope is after this I will help get 1.14 running, then after that, hopefully implement a map api in 1.14 too.

tyhdefu:
Alright, good news. The map api is now finished as I see it.

*Why I think MapInfo should be a DataHolder*
- It is mutable, like an entity
- Changing its values has a direct consequence ingame
- I think it fits better in the current style of having keys to access any data.

I will resolve the review about it implementing DataHolder once we have agreed upon what if any action needs to be taken.

src/main/java/org/spongepowered/api/Server.java
Zidane, line 453: When would this not be present?

tyhdefu: If the there were no overworlds/worlds, or the they were not loaded. This is because minecraft normally uses one map storage per world, which sponge changes to a single storage for all worlds. I will add a clarification to the javadocs. This is the same circumstance as when the scoreboard would not be present so I will copy the relevant paragraph.

src/main/java/org/spongepowered/api/Server.java
```diff
@@ -443,7 +443,12 @@
 
     /**
      * Gets the map storage for this server
-     * @return MapStorage map storage
+     *
+     * <p>The server scoreboard may not be available if dimension 0
```
dualspiral: Copy/paste error - this isn't a scoreboard!

src/main/java/org/spongepowered/api/event/action/CreateMapEvent.java
tyhdefu: MapInfo is mutable, but it can refer to more than one map. I can see that maybe you would want to swap out the event's MapInfo with your own that already refers to a lot of maps. I think the thing that would make the most sense to cancel the creation of the new map, then get the used item (assuming there was one), and switch it out for a ItemStack with type filled_map, and then offer it your map info, because the event is for the creation of a new map, not the item use.

Probably the better idea is to use the Interact event and not create map event.

dualspiral:
Can we delete a `MapInfo` so we can reuse the map ID?

I say remove rather than clear because if someone creates a new map, we will want to reuse the ID. If it's cleared but not removed, we can't reuse it.

Should we have maps associated with a UUID too so we can get a specific map we know about? I guess there would be maintenance mapping UUID to map ID, but I can see that being useful too.

tyhdefu:
> Can we delete a `MapInfo` so we can reuse the map ID?
> 
> I say remove rather than clear because if someone creates a new map, we will want to reuse the ID. If it's cleared but not removed, we can't reuse it.
> 
> Should we have maps associated with a UUID too so we can get a specific map we know about? I guess there would be maintenance mapping UUID to map ID, but I can see that being useful too.

I like the idea of having a mapping to UUID. I think that this UUID probably would also need saving into the map files, (I think UUIDs need to be persistant across restarts. As for being able to delete maps, minecraft stores the highest known map id in a file called idcounts, which is how it knows what to set the next map id as. I think most likely the best way to implement a feature to delete maps would be to store a list of map ids that are free for use, (unless the highest map id is deleted). The problem with this, is that any map item with that map id, would not have any map attached to it, we would probably need some checks to say that its okay to have maps that don't exist. When a plugin added a map back, it would fill the id in, and all of sudden, unknowingly, a plugin has created a map info thats already linked to a map. At that point it seems pretty useless to delete maps, if the next map created will "undelete" it, with different data.

tyhdefu:
Potetially, if we think there are going to be a lot of plugins that want to use temporary maps, I could write a service called "MapRecycler" or similar that you gave a certain amount of maps as a pool, then you requested a map, handed it out to a player, then deleted it from the player's inventory (and made sure it was not stored in a chest or similar), then marked it as recycled where it would be re-added to the pool. This service would make note of the UUIDs of the map pool, and would have to store it in config somewhere so that no maps were "leaked".
Also take note that 1.13+ uses an int instead of a short for the amount of maps. This means we would be able to have a lot more maps. Potentially at that point it could make sense to allow plugins to delete maps, which would delete the map file, and just let the map number progress upwards without having to worry about using a map hidden in a player's chest somewhere.

tyhdefu:
As far as I know, this is complete and ready to merge.

gabizou:
Gonna give this a review sometime this week if not today/tomorrow. Great work getting this fleshed out and implemented.

src/main/java/org/spongepowered/api/map/MapInfo.java
```diff
@@ -0,0 +1,54 @@
...
+ * @see MapInfoData
+ */
+public interface MapInfo extends DataHolder, Identifiable {
+    /**
+     * Checks if a different MapInfo refers the the
+     * same map
+     * @param other MapInfo
+     * @return true if they refer to the same map
+     */
+    boolean isLinked(MapInfo other);
```
dualspiral: I'm confused by this. Looking at the implementation, you're suggesting that multiple `MapInfo` objects can have the same backing ID. When would this be the case? Surely this _cannot_ be the case?

src/main/java/org/spongepowered/api/map/color/MapColor.java
```diff
@@ -0,0 +1,104 @@
...
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.map.MapCanvas;
+import org.spongepowered.api.util.Color;
+
+/**
+ * A MapColor that represents a color that can be used to
+ * paint on a {@link MapCanvas}
+ */
+public interface MapColor extends DataSerializable {
```
dualspiral, line 38: There should be a method that allows for the shade to be returned. Perhaps it should become a `MapColorShade` catalog type?

src/main/java/org/spongepowered/api/map/color/MapColorType.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.util.Color;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+/**
+ * A type of color used to create a {@link MapColor}
+ */
+@CatalogedBy(MapColorTypes.class)
+public interface MapColorType extends CatalogType {
```
dualspiral, line 36: If a `MapColorShade` type is added, consider adding `Color getColor(MapColorShade)` here so that a colour can be retrieved for a type/shade combo.

src/main/java/org/spongepowered/api/map/color/MapColorTypes.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public static final MapColorType RED = DummyObjectProvider.createFor(MapColorType.class, "RED");
+
+    public static final MapColorType BLACK = DummyObjectProvider.createFor(MapColorType.class, "BLACK");
+
+    public static final MapColorType GOLD = DummyObjectProvider.createFor(MapColorType.class, "GOLD");
+
+    public static final MapColorType DIAMOND = DummyObjectProvider.createFor(MapColorType.class, "DIAMOND");
+
+    public static final MapColorType LAPIS = DummyObjectProvider.createFor(MapColorType.class, "LAPIS");
```
dualspiral: `LAPIS_LAZULI` (also change the underlying implementation ID et. al.)

src/main/java/org/spongepowered/api/map/color/MapColorTypes.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+package org.spongepowered.api.map.color;
+
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/*
+* Contains all the colors minecraft supports
+* All of these have no shading/darkness attached.
+ */
+public final class MapColorTypes {
```
dualspiral, line 37: These need to be sorted.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,163 @@
...
+import com.flowpowered.math.vector.Vector2i;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.DataSerializable;
+import org.spongepowered.api.data.persistence.DataBuilder;
+import org.spongepowered.api.util.Direction;
+
+/**
+ * A MapDecoration that represents a Decoration on a map,
+ * e.g player marker, monument marker etc.
+ * All coordinates are world coordinates, not relative to the map
```
dualspiral: > All coordinates are world coordinates, not relative to the map

but then you say on anything that _uses_ the co-ordinates:

> 0,0 is the centre of a map.
> Ranges from {@value java.lang.Byte#MIN_VALUE}-{@value java.lang.Byte#MAX_VALUE}. AKA any valid byte value

Which is it? For a custom map, it's not going to make sense for it to be world co-ordinates, right?

src/main/java/org/spongepowered/api/map/decoration/MapDecorationTypes.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.spongepowered.api.map.decoration;
+
+import org.spongepowered.api.map.color.MapColorType;
+import org.spongepowered.api.util.generator.dummy.DummyObjectProvider;
+
+/**
+ *
+ */
+public final class MapDecorationTypes {
```
dualspiral, line 35: Need to be sorted.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,163 @@
...
+        Builder type(MapDecorationType type);
+
+        /**
+         * Sets the X position of the MapDecoration
+         * (for when it is applied to a {@link org.spongepowered.api.map.MapInfo})
+         * @param x map coordinate where 0 is the centre of the map.
+         * Ranges from {@value java.lang.Byte#MIN_VALUE}-{@value java.lang.Byte#MAX_VALUE}. AKA any valid byte value
+         * @return this Builder, for chaining
+         */
+        Builder x(int x) throws IllegalStateException;
```
dualspiral: This should throw an `IllegalArgumentException` as the error is in what is supplied, not the state of the builder.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,163 @@
...
+
+        /**
+         * Sets the Y position of the MapDecoration
+         *
+         * (for when it is applied to a {@link org.spongepowered.api.map.MapInfo})
+         * @param y map coordinate where 0 is the centre of the map.
+         * Ranges from {@value java.lang.Byte#MIN_VALUE}-{@value java.lang.Byte#MAX_VALUE}. AKA any valid byte value
+         * @return this Builder, for chaining
+         */
+        Builder y(int y) throws IllegalStateException;
```
dualspiral: As above.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
tyhdefu: Yep you're correct, I read the source code wrong originally and forgot to correct it.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,164 @@
...
+     * @return Vector2i Co-ordinate position in world
+     */
+    Vector2i getPosition();
+
+    /**
+     * Gets the X position on a map, or where it will be when applied
+     * Ranges from {@value java.lang.Byte#MIN_VALUE}-{@value java.lang.Byte#MAX_VALUE}. AKA any valid byte value
+     * @return int Y world coordinate
+     */
+    int getX();
```
dualspiral: Why do we have `getPosition` _and_ `getX/getY` methods? Same with `setPosition/X/Y` - if you're setting the position it should probably just be done in one go (keep the `Vector2i` variants).

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    void setRotation(Direction rot);
+
+    /**
+     * Gets the direction the Map Decoration is pointing
+     * @return Direction, a cardinal/ordinal/secondary
+     */
+    Direction getRotation();
+
+    interface Builder extends DataBuilder<MapDecoration> {
```
dualspiral: This is really a sweeping comment here - please _do not_ have developer facing builders implement `DataBuilder`. They are two completely different concepts. I just looked at the implementation of this builder and there is _zero relation_ between the two functions. `build()` uses the methods in `Builder`, `DataBuilder` does not at all.

Keep them separate - that includes in the implementation. You don't even need to have an interface extend `DataBuilder` anywhere.

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,164 @@
...
+        Builder type(MapDecorationType type);
+
+        /**
+         * Sets the X position of the MapDecoration
+         * (for when it is applied to a {@link org.spongepowered.api.map.MapInfo})
+         * @param x map coordinate where 0 is the centre of the map.
+         * Ranges from {@value java.lang.Byte#MIN_VALUE}-{@value java.lang.Byte#MAX_VALUE}. AKA any valid byte value
+         * @return this Builder, for chaining
+         */
+        Builder x(int x) throws IllegalArgumentException;
```
dualspiral: `x` and `y` methods are redundant, remove them (in favour of `position`)

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,164 @@
...
+         */
+        Builder y(int y) throws IllegalArgumentException;
+
+        /**
+         * Sets the rotation in degrees for this builder
+         * (for when it is applied to a {@link org.spongepowered.api.map.MapInfo})
+         * @param rot Direction, any cardinal/ordinal/secondary ordinal direction. (Not up/down)
+         * @return this Builder, for chaining
+         */
+        Builder rotation(Direction rot) throws IllegalArgumentException;
```
dualspiral: Would using the `Rotation` class be better here? That has simple directions in a non-cardinal form and given it's said a map doesn't represent the world, N/S/E/W don't really make as much sense as `TOP`, `BOTTOM` etc.

(Also, JDs say "rotation in degrees" - that's not what's happening here.)

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Gets the {@link MapColor} at the requested location.
+     *
+     * @param x The x location
+     * @param y The y location
+     * @return MapColor color at this location
+     * @throws IllegalArgumentException if either of the requested
+     *      co-ordinates are out of bounds.
```
dualspiral, line 74: What are the bounds? Where is (0, 0)?

[Take a look at how I've described it here in `MapDecoration`](https://github.com/SpongePowered/SpongeAPI/pull/2112/files#diff-93c4ab7cf6dd5ffeb933fbd0791f389bR36) - this is the sort of information a user will need.

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,150 @@
...
+         */
+        Builder from(MapCanvas canvas);
+
+        /**
+         * Fills a canvas from an image
+         * This given image will be copied and thus changes to the
+         * image after this is called will not be reflected.
+         * @param image to take from
+         * @return This Builder, for chaining
+         * @throws IllegalArgumentException if width/height is wrong or not known yet
```
dualspiral, line 173: > if width/height is wrong or not known yet

What does this mean?

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,150 @@
...
+     * Creates an image and replaces all {@link MapColorTypes#AIR}
+     * with the specified color.
+     * It creates a copy of held data, therefore changing this image will have
+     * no effect on the MapCanvas.
+     * @param color to replace the air MapColorType
+     * @return Image from the canvas
+     */
+    Image toImage(java.awt.Color color);
+
+    Builder toBuilder(MapCanvas canvas);
```
dualspiral: Why does this accept a `MapCanvas` when this is on a `MapCanvas` object?

src/main/java/org/spongepowered/api/map/MapCanvas.java
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * @param canvas
+     * @return A {@link Builder}
+     */
+    Builder toBuilder(MapCanvas canvas);
+
+    /**
+     * A builder that creates a {@link MapCanvas}.
+     */
+    interface Builder extends DataBuilder<MapCanvas> {
```
dualspiral: Builder should not extend `DataBuilder`.

src/main/java/org/spongepowered/api/map/color/MapColor.java
```diff
@@ -0,0 +1,152 @@
...
+         */
+        Builder darkest();
+
+        /**
+         * Sets the {@link MapColorType} that will form the basis of the built
+         * {@link MapColor}.
+         *
+         * @return This builder, for chaining
+         */
+        Builder baseColor(MapColorType mapColor);
```
dualspiral, line 142: `colorType`?

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
```diff
@@ -0,0 +1,214 @@
...
+ * co-ordinates (-128, -128).</p>
+ */
+public interface MapDecoration extends DataSerializable {
+
+    /**
+     * Creates a {@link Builder} for generating {@link MapDecoration}s.
+     *
+     * @return A {@link Builder}
+     */
+    static Builder builder() {
```
dualspiral, line 62: It might be useful to have 

```
MapDecoration of(final MapDecorationType type) {
    return MapDecoration.builder().type(type).build();
}
```

as you can set the position and rotation outside of the builder.

dualspiral (changes requested):
I've re-done some more of the javadocs - this is helping me with review. However, by doing so, there's more that needs to be thought about...

src/main/java/org/spongepowered/api/map/decoration/MapDecoration.java
tyhdefu: I am all for this, but we need 16 directions, whereas currently Rotations currently has 8. Should I add fields for TOP_TOP_LEFT, in the same style as Direction, or just keep using Direction? Have you got better names for the fields?

dualspiral: Or just make a `MapDecorationOrientation` psuedo enum and use that.

dualspiral:
So, thinking on this, all our work has pretty much shifted to API-8 and the Map API was implemented there. I don't think anyone has a particular interest in making the required changes to API-7 to get this working. I'm therefore going to close this and associated PRs.

I will note (again) that this work was not wasted as it is in API-8, thanks @tyhdefu.

PR #2114, opened 2020-04-19
Improve the effects API. Added SoundEffect.

Cybermaxke:
This makes it easy and consistent to play effects, previously you had to use `playSound`, `spawnParticles`, `stopSounds`, or create a particle effect with firework effects.

This PR cleans that up and now there's a single method for `Viewer.play(PlayableEffect)`, supported `PlayableEffect` types are `SoundEffect`, `ParticleEffect`, `FireworkEffect`, `StopSoundEffects` (possibly a music disc related one, still TODO). `SoundEffect`, `StopSoundEffects` is also serializable, which makes it easy to use in config files.

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/Viewer.java
+++ b/src/main/java/org/spongepowered/api/effect/Viewer.java
@@ -27,9 +27,6 @@ package org.spongepowered.api.effect;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import org.spongepowered.api.block.BlockState;
-import org.spongepowered.api.effect.particle.ParticleEffect;
-import org.spongepowered.api.effect.sound.SoundCategories;
-import org.spongepowered.api.effect.sound.SoundCategory;
 import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.effect.sound.music.MusicDisc;
 import org.spongepowered.api.text.BookView;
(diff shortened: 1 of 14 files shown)


src/main/java/org/spongepowered/api/effect/sound/SoundEffect.java
```diff
@@ -0,0 +1,291 @@
...
+     *
+     * @param type The type
+     * @param category The category
+     * @param volume The volume
+     * @param pitch The pitch
+     * @return The sound effect
+     */
+    static SoundEffect of(SoundType type, SoundCategory category, double volume, double pitch) {
+        return builder().type(type).category(category).volume(volume).pitch(pitch).build();
+    }
```
ItsDoot, line 51: Not sure how I feel about `of`s with more than 3 parameters. I know this is just mirroring the old `playSounds` methods, but doesn't that defeat the purpose of using builders in the first place?

Cybermaxke: Yeah, I was also thinking about that. What about just the methods with the type, volume and pitch?

ItsDoot: All the other values' defaults seem to follow principle of least surprise, so yea seems good to me. 👍

ItsDoot:
Seem to be missing `minVolume` for `SoundEffect`.

Cybermaxke:
@ItsDoot I don't see why `minVolume` is in the api.  You can easily do `Math.max(minVolume, volume)`, or am I missing something?

ItsDoot:
Well, it's part of the `/playsounds` command, but doesn't seem to be part of the packet itself? So unsure what it actually does.

Zidane:
@Cybermaxke Need this updated.

ItsDoot:
Can `Location` (or `ServerLocation`) also be made a `Viewer`?

src/main/java/org/spongepowered/api/effect/Viewer.java
```diff
@@ -48,260 +45,21 @@
...
-     */
-    void stopSoundTypes(Supplier<? extends SoundType> sound);
-
-    /**
-     * Stops all the sounds that are played in the
-     * given {@link SoundCategory}.
-     *
-     * @param category The sound category
-     */
-    void stopSounds(SoundCategory category);
```
gabizou, line 278: What's the logical replacement for these?

src/main/java/org/spongepowered/api/effect/sound/SoundEffect.java
gabizou: Any `of` with more than a few is gonna be too much.

Zidane:
@Cybermaxke Can you update this for Adventure?

Zidane:
Superceded by Adventure or may return at a later date if deemed necessary.

Issue #1891, opened 2018-08-04
TextTemplate Redesign


ryantheleach:
# WIP - Work in Progress

I'm looking for feedback from plugin and Sponge devs to come up with the list of goals and usecases to effectively design this, and someone to champion this after / with the Text moving to implementation.

# Why Redesign?

Text Templates have numerous reported issues:

https://github.com/SpongePowered/SpongeAPI/issues/1856 Complex Serialization Structure, making them a pain for config / user editing.

https://github.com/SpongePowered/SpongeAPI/issues/1855 Having surprising behavior regarding colors.

https://github.com/SpongePowered/SpongeAPI/issues/1792 Their builders acting differently to TextBuilder & Text.of "magic" method.

https://github.com/SpongePowered/SpongeAPI/issues/1493 General serialization problems

https://github.com/SpongePowered/SpongeAPI/issues/1452 (OCD) Poor naming regarding arguments vs parameters.

https://github.com/SpongePowered/SpongeAPI/issues/1462 and Confusion surrounding 'meta' formatting options for Actions.

They are poorly understood, poorly used, and as such, people avoid them for their own solutions constantly, even implementing "search and replace" algorithms manually for Text's.

# Goals of a new TextTemplate:

1. Serialize very similarly to Texts, with a **standard** way of marking arguments, common to all plugins.
2. Leave evaluation of placeholders as LATE as possible, and allow plugins to determine the context. (Per player formatting? per receiver formatting?)
3. Have serializers that closely match the current Text Serializers.
4. Maybe incorporating / Promoting a 'global' arg registration system, akin to, https://ore.spongepowered.org/rojo8399/PlaceholderAPI
5. Use a similar parent interface as server side translations, so that a given TextTemplate can be 'resolved' for a given receivers Locale? 
6. Maintaining performance, by generating as much Text ahead of time as possible, TextTemplates should be reusable.

# Usecases.

1. Plugins specifying configurable text in configs, that also have tokens that need replacing by dynamic values at runtime.

2. Chat Formatting plugins would use TextTemplates to replace known variables, from other plugins.

# What Next?

If you disagree, or have further Usecases / Goals please add them.

If you have a strawman API to prototype with / play with, please make a pull request, EXPLAINING it's a prototype, and link it to this issue.

XakepSDK:
I like how formatting code serializer works. Really simple and short output. Maybe somehow fit that serializer for text templates?

ST-DDT:
[...]

PR #2116, opened 2020-05-01
Add support for centralised Placeholders

dualspiral:
**Sponge API (that's here!)** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2588) | [SpongeDocs](https://github.com/SpongePowered/SpongeDocs/pull/834)

The intent behind this is to harmonise the various templating systems to use a common bank of placeholders, and comes from initial ideas from when trying to rebuild my placeholder system for Nucleus. I thought that this would be valuable to have in Sponge some time ago (akin to Econ and Permissions), but [this forum post](https://forums.spongepowered.org/t/help-best-approach-for-a-chat-plugin/36135) kicked me into actually doing it. 

This simply provides `PlaceholderParsers` that plugins can supply to the Sponge registry, and `PlaceholderContext` and `PlaceholderText` objects that templating engines can use 

* `PlaceholderParser`s will be registered centrally using the Sponge registry system such that there is a standard way for all plugins to get a placeholder, along with a standard way to include them in text constructions.
* Consumers wanting a placeholder grab the appropriate `PlaceholderParser` from the Sponge Registry, and create `PlaceholderContext` objects to supply to the parser to get a concrete `Text`.
* `PlaceholderText`s are `TextRepresentable`s that bundle the `PlaceholderParser` along with a `PlaceholderContext` so can be used easily within `Text.of(...)` or `Text.Builder`, minimising the effort required on the plugin side. The `Text` is evaluated at the time the text is built.

Part of the ideas are from #1891, but notably, this PR does NOT include a templating system. This is intentional, to allow plugins to provide their own systems as they do now.

---

<details> 
  <summary>The original text from this PR is here, when there was also a service.</summary>

The text below is from the previous iteration of the PR, the service no longer exists, but the idea is about the same.

The intent behind such a service is to harmonise the various templating systems to use a common bank of placeholders, and comes from initial ideas from when trying to rebuild my placeholder system for Nucleus. I thought that this would be valuable to have in Sponge some time ago (akin to Econ and Permissions), but [this forum post](https://forums.spongepowered.org/t/help-best-approach-for-a-chat-plugin/36135) kicked me into actually doing it. 

This is for discussion, particularly javadocs - is there anything I could do there to make them clearer?

* `PlaceholderParser`s will be registered centrally using the Sponge registry system such that there is a standard way for all plugins to get a placeholder, along with a standard way to include them in text constructions.
* Consumers wanting a placeholder can use this service to either parse a token, or they can go for a more DIY approach and supplying a builder (which is supplied by the service, rather than the normal thing of `PlaceholderText.builder()`, though I could do that I guess and just reference back to the service)
* `PlaceholderText`s are `TextRepresentable`s so can be used easily within `Text.of(...)` or `Text.Builder`, minimising the effort required on the plugin side.

The basic idea is that you build a `PlaceholderText` with a `PlaceholderParser` and use that in `Text.of(...)` or ` Text.builder()` objects.

This service is replaceable, allowing for a plugin to support a superset of the required base - so for example, I could add my more advanced stuff in, or a dedicated placeholder plugin could make use of the service. Other plugins can also just make use of the parser registry if they wish - they don't need to parse via the service if they don't want to.

Part of the ideas are from #1891, but notably, this PR does NOT include a templating system. This is intentional, to allow plugins to provide their own systems as they do now.
</details>

### DIFF
--- a/src/main/java/org/spongepowered/api/text/placeholder/PlaceholderText.java
+++ b/src/main/java/org/spongepowered/api/text/placeholder/PlaceholderText.java
@@ -1,0 +1,130 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.placeholder;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.text.TextRepresentable;
+import org.spongepowered.api.util.ResettableBuilder;
+
+/**
+ * A {@link TextRepresentable} that can be used in {@link Text} building methods
+ * that represents a placeholder in text.
+ *
+ * <p>A {@link PlaceholderText} is the collection of a {@link PlaceholderParser}
+ * along with contextual data in the supplied {@link PlaceholderContext},
+ * enabling its use in a {@link Text} object.</p>
+ *
+ * <p>Such placeholders will generally be built from tokenized strings, however
+ * these objects make no assumption about the format of text templating. Such a
+ * system can therefore be used by other templating engines without conforming
+ * to a particular standard.</p>
+ *
+ * <p>The {@link PlaceholderContext} is fixed when this object is created, but
+ * {@link PlaceholderParser#parse(PlaceholderContext)} is not called until
+ * {@link #toText()} is called. Thus, any {@link Text} object that is created
+ * will reflect the time that the {@link Text} object was requested, and not when
+ * this object itself was created. It therefore follows that implementations must
+ * not cache the result of {@link #toText()} unless it is known that the supplied
+ * parser is <strong>not</strong> sensitive to the time of invocation.</p>
+ */
+public interface PlaceholderText extends TextRepresentable {
+
+    /**
+     * Gets a builder for creating {@link PlaceholderText}.
+     *
+     * @return A {@link Builder}
+     */
+    static PlaceholderText.Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * Gets the {@link PlaceholderContext} that is to be supplied to the
+     * {@link PlaceholderParser} to create the {@link Text}.
+     *
+     * @return The {@link PlaceholderContext}
+     */
+    PlaceholderContext getContext();
+
+    /**
+     * Gets the {@link PlaceholderParser} that handles this placeholder.
+     *
+     * @return The {@link PlaceholderParser}
+     */
+    PlaceholderParser getParser();
+
+    /**
+     * Creates a {@link Text} from the supplied {@link PlaceholderParser} in the
+     * context of the supplied {@link PlaceholderContext}.
+     *
+     * <p>This will always return a {@link Text} object, however, if the parser
+     * could not handle the provided context, this will be {@link Text#EMPTY}.
+     * </p>
+     *
+     * @return The parsed {@link Text}
+     */
+    @Override
+    default Text toText() {
+        return getParser().parse(getContext());
+    }
+
+    /**
+     * A builder for {@link PlaceholderText} objects.
+     */
+    interface Builder extends ResettableBuilder<PlaceholderText, Builder> {
+
+        /**
+         * Sets the token that represents a {@link PlaceholderParser} for use
+         * in this {@link PlaceholderText}.
+         *
+         * @param parser The {@link PlaceholderParser} to use
+         * @return This, for chaining
+         */
+        Builder setParser(PlaceholderParser parser);
+
+        /**
+         * Sets the {@link PlaceholderContext} that will be provided to the
+         * {@link PlaceholderParser} to create the {@link Text} when
+         * {@link TextRepresentable#toText()} is called.
+         *
+         * @param context The {@link PlaceholderContext} to use.
+         * @return This, for chaining
+         */
+        Builder setContext(PlaceholderContext context);
+
+        /**
+         * Builds and returns the placeholder.
+         *
+         * @return The appropriate {@link PlaceholderText}
+         * @throws IllegalStateException if the builder has not been completed,
+         *  or the associated {@link PlaceholderParser} could not validate the
+         *  built {@link PlaceholderText}, if applicable.
+         */
+        PlaceholderText build() throws IllegalStateException;
+
+    }
+
+}
--- a/build.gradle
+++ b/build.gradle
@@ -198,6 +198,7 @@ sortClassFields {
     add 'main', 'org.spongepowered.api.text.chat.ChatVisibilities'
     add 'main', 'org.spongepowered.api.text.format.TextColors'
     add 'main', 'org.spongepowered.api.text.format.TextStyles'
+    add 'main', 'org.spongepowered.api.text.placeholder.PlaceholderParsers'
     add 'main', 'org.spongepowered.api.text.selector.SelectorTypes'
     add 'main', 'org.spongepowered.api.util.TypeTokens'
     add 'main', 'org.spongepowered.api.util.ban.BanTypes'
--- a/src/main/java/org/spongepowered/api/CatalogTypes.java
+++ b/src/main/java/org/spongepowered/api/CatalogTypes.java
@@ -70,6 +70,7 @@ import org.spongepowered.api.scoreboard.displayslot.DisplaySlot;
 import org.spongepowered.api.scoreboard.objective.displaymode.ObjectiveDisplayMode;
 import org.spongepowered.api.service.economy.Currency;
 import org.spongepowered.api.service.economy.transaction.TransactionType;
+import org.spongepowered.api.text.placeholder.PlaceholderParser;
 import org.spongepowered.api.statistic.BlockStatistic;
 import org.spongepowered.api.statistic.EntityStatistic;
 import org.spongepowered.api.statistic.ItemStatistic;
@@ -268,6 +269,8 @@ public final class CatalogTypes {
     public static final Class<PickupRule> PICKUP_RULE = PickupRule.class;
 
     public static final Class<PistonType> PISTON_TYPE = PistonType.class;
+
+    public static final Class<PlaceholderParser> PLACEHOLDER_PARSER = PlaceholderParser.class;
 
     public static final Class<PlantType> PLANT_TYPE = PlantType.class;
 
--- a/src/main/java/org/spongepowered/api/text/placeholder/PlaceholderContext.java
+++ b/src/main/java/org/spongepowered/api/text/placeholder/PlaceholderContext.java
@@ -1,0 +1,172 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.text.placeholder;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.Optional;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+/**
+ * Contains the context that a {@link PlaceholderParser} can use to determine
+ * what to display.
+ */
+public interface PlaceholderContext {
+
+    /**
+     * Creates a {@link PlaceholderContext} for a {@link PlaceholderParser} to
+     * consume.
+     *
+     * @return The builder.
+     */
+    static Builder builder() {
+        return Sponge.getRegistry().createBuilder(Builder.class);
+    }
+
+    /**
+     * If provided, the {@link Object} which to pull information from
+     * when building the placeholder text.
+     *
+     * <p>Examples of how this might affect a placeholder are:</p>
+     *
+     * <ul>
+     *     <li>
+     *         For a "name" placeholder that prints out the source's name,
+     *         the name would be selected from this source.
+     *     </li>
+     *     <li>
+     *         For a "current world" placeholder that returns a player's current
+     *         world, this would pull the name of that current world from the
+     *         player.
+     *     </li>
+     * </ul>
+     *
+     * <p>It is important to note that the associated context does not
+     * necessarily have to be the sender/invoker of a message, nor does it
+     * have to be the recipient. The source is selected by the context of
+     * builder. It is up to plugins that use such placeholders to be aware
+     * of the context of which the placeholder is used.
+     * {@link PlaceholderParser}s should make no assumption about the origin of
+     * the context.</p>
+     *
+     * <p>If an invalid {@link Object} is provided for the context
+     * of the placeholder, then the associated {@link PlaceholderParser} must
+     * return a {@link Text#EMPTY}.</p>
+     *
+     * @return The associated {@link Object}, if any.
+     */
+    Optional<Object> getAssociatedObject();
+
+    /**
+     * The variable string passed to this token to provide contextual
+     * information.
+     *
+     * @return The argument, if any
+     */
+    Optional<String> getArgumentString();
+
+    /**
+     * A builder for {@link PlaceholderText} objects.
+     */
+    interface Builder extends ResettableBuilder<PlaceholderContext, Builder> {
+
+        /**
+         * Sets the {@link Object} to use as a source of information
+         * for this {@link PlaceholderText} to the supplied {@link Player}.
+         *
+         * @param player The player to associate this text with.
+         * @return This, for chaining
+         *
+         * @see #getAssociatedObject()
+         */
+        default Builder setAssociatedObject(Player player) {
+            final UUID uuid = player.getUniqueId();
+            return setAssociatedObject(() -> Sponge.getServer().getPlayer(uuid).orElse(null));
+        }
+
+        /**
+         * Sets the {@link Object} to use as a source of information
+         * for this {@link PlaceholderText}. If {@code null}, removes this
+         * source.
+         *
+         * <p>If you are intending to keep the associated
+         * {@link PlaceholderText} for any period of time and that you wish to
+         * associate a game object with the placeholder, use
+         * {@link #setAssociatedObject(Supplier)} instead, supplying a function
+         * that can recreate the object if necessary.</p>
+         *
+         * <p>If supplying a {@link Player}, use
+         * {@link #setAssociatedObject(Player)}, which will handle creating the
+         * correct supplier for you.</p>
+         *
+         * @param object The {@link Object} to associate this text with.
+         * @return This, for chaining
+         *
+         * @see #getAssociatedObject()
+         */
+        Builder setAssociatedObject(@Nullable Object object);
+
+        /**
+         * Sets the {@link Object} to use as a source of information
+         * for this {@link PlaceholderText}. If {@code null}, removes this source.
+         *
+         * @param supplier A {@link Supplier} that provides the {@link Object}
+         * @return This, for chaining
+         *
+         * @see #getAssociatedObject()
+         */
+        Builder setAssociatedObject(@Nullable Supplier<Object> supplier);
+
+        /**
+         * Sets a string that represents variables for the supplied token.
+         * The format of this argument string is dependent on the parser
+         * that consumes this string and thus is not prescribed here.
+         *
+         * @param string The argument string, may be null to reset to
+         *      the default argument string
+         * @return This, for chaining
+         *
+         * @see #getArgumentString()
+         */
+        Builder setArgumentString(@Nullable String string);
+
+        /**
+         * Builds and returns the placeholder.
+         *
+         * @return The appropriate {@link PlaceholderText}
+         * @throws IllegalStateException if the builder has not been completed,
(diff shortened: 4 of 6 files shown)


src/main/java/org/spongepowered/api/service/placeholder/PlaceholderService.java
```diff
@@ -0,0 +1,97 @@
...
+     * how to select a placeholder if the supplied {@code token} is not a
+     * known ID for a registered {@link PlaceholderParser}.</p>
+     *
+     * @param messageReceiver The source that tokens should use as a context.
+     *  May be {@code null}, but some tokens may not parse without this
+     *  supplied.
+     * @param token The token to obtain.
+     * @return The parsed {@link Text}.
+     */
+    default Optional<PlaceholderText> parse(@Nullable MessageReceiver messageReceiver, String token) {
```
Zidane: I dislike API methods where a parameter may come at them as null. It would be far cleaner to do separated methods that make the intent clear.

dualspiral: Normally, I'd agree, and don't mind the idea of making more overloads. However, this is inteded to be plugin implemented, and as a group of methods that do the same thing with different conditions I opted to have a "master" method that has two nullable parameters.

* If we want plugins to do as little as needed by using default methods, we're going to need to have a `@Nullable` source somewhere in one of the methods.
* Alternatively, we just do away with default methods as this is a totally new interface, which I admit, I am tempted to do.

That said, I do need to alter the order of the parameters anyway. The "token" should come first, as that's the most important part of the call.

Zidane: I'd do away with the default methods. That way we don't muddy the contract.

dualspiral: Okie doke, all sorted.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderService.java
```diff
@@ -0,0 +1,116 @@
...
+     * Gets a {@link PlaceholderText} based on the provided token,
+     * using the provided {@link MessageReceiver} and supplied argument string
+     * for context.
+     *
+     * <p>It is entirely up to the implementation of the service to determine
+     * how to select a placeholder if the supplied {@code token} is not a
+     * known ID for a registered {@link PlaceholderParser}.</p>
+     *
+     * @param token The token name.
+     * @param arguments The arguments for the token.
```
ST-DDT: `arguments`  sounds like there are many, but there is only a single string.

ST-DDT:
Could you add an example for both specifing and using those placeholders? Or will you write a docs PR as well?

dualspiral:
@ST-DDT I'll write a Docs PR, because at least then it's translatable.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    Optional<MessageReceiver> getAssociatedReceiver();
+
+    /**
+     * The variable string passed to this token to provide contextual
+     * information.
+     *
+     * @return The argument, if any
+     */
+    Optional<String> getArgument();
```
NickImpact:
```suggestion
    Optional<String> getArguments();
```
Since the argument in PlaceholderService was modified from argument to argumentString, it'd likely be best to make that modification here as well.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,176 @@
...
+         * supplied to {@link #setParser(PlaceholderParser)} and thus is
+         * not prescribed here.
+         *
+         * @param string The argument string, may be null to reset to
+         *      the default argument string
+         * @return This, for chaining
+         *
+         * @see PlaceholderText#getArgument()
+         */
+        Builder setArgument(@Nullable String string);
```
NickImpact:
```suggestion
        Builder setArguments(@Nullable String string);
```
Like the above thought, probably want to update this builder function to be plural rather than singular

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
dualspiral: Gone for `getArgumentString` instead.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
dualspiral: As above.

dualspiral:
I've updated the API to include a `getParser` method - explanation why is on the javadocs. I've also added a Docs PR with some more information on how it works - please read that in tandem with this to understand the thought process.

ST-DDT:
How about turning argumentString into a List<String> or something similar. That allows consistent parameter transmissions across plugins / avoids different separator/escape characters per plugin.

dualspiral:
> That allows consistent parameter transmissions across plugins / avoids different separator/escape characters per plugin.

So does a string. If a templating system is passing its own implementation specific escape strings then it's absolutely doing it wrong - only the templating system should care about its escape strings and should be removing them before passing them along. Futher, specifying that a plaveholder may accept a single argument string which it will manage itself is a simpler concept than asking systems to parse multiple arguments.

Note that these placeholders have nothing to do with templating systems (from an API and design perspective) and so should not have to compensate for them. They are two different things.

`PlaceholderParsers` are, in effect, mini commands and so they should parse an argument string in the way they see fit. They can use our command system if they want to.

NickImpact:
That’s the thought I had with that argument. It really can be whatever given the placeholder parser should be the one processing the arguments, whereas the request should only be sending them through. For my case, since I’ll need placeholders to be focused on a source that doesn’t implement the MessageReceiver interface, I very likely will consider using json or something along the lines for my parsers. 

So in a sense, processing it exactly like a command makes sense.

dualspiral:
Here's a good discussion point

> For my case, since I’ll need placeholders to be focused on a source that doesn’t implement the MessageReceiver interface

The choice of that return object as a `MessageReciever` wasn't particularly arbitary, because it's chat related. But, in many cases, the associated object isn't actually the _receiver_, but the _sender_. To take my use case, for example, a chat message sent by player X might be formatted like:

```
{{nucleus:prefix}} {{nucleus:displayname}} > here is my message
```

The "associated receiver" in the `PlaceholderText` for these tokens would actually be the sender! But in another case, Nucleus' `/broadcast`, for example, the "associated receiver" is actually the receiver, except in the case of my special token `{{sender}}` - at this point, it's my templating engine that decides what the context is. So the method is named badly, really only named after the choice of `MessageReceiver` as the asssociated context.

Before people suggest the notion of "receiver" and "sender", that's not going to happen. Most, if not all, placeholders are going to need to consider at most _one_ "associated" object. For Nucleus' `{{displayname}}`, which context would I pick if both were provided?

The plugin and/or templating engine knows what it wants the placeholder to be about and should be the one to provide the "associated context". Remember what placeholders are: `TextRepresentable`s that require a little more information to know what to output.

So, should I make the "associated receiver" an `Object` and let plugins just do the checks they need on it? Or should it be some other bound? Considering that `MessageChannel`s support `Object` senders, there is precedent for the fomer. Just remember, the control of this object isn't in the placeholder's hands, but the plugin providing the message - no guarantees what it will be, just like the Cause!

(I did consider the `Cause`, but the cause of a message is not the same as who the subject of a placeholder is. So that's out.)

XakepSDK:
What if placeholder wants `User`?
I would like to implement some placeholders for offline users.

NickImpact:
I've done some thinking on this, namely going as far as trying to consider a means of a placeholder parser being able to make some call to report a _potentially_ non-empty text only when it meets the source requirements of the parser. 

Unfortunately, I haven't been able to come up with a design for such a concept that'd be an acceptable way to do such. Thought of having a default method that will take in a class type and supplier for a Text, but given that the parse call will be the function calling this method more than likely, the whole block of code that should be in that parse function ends up as a supplier to the convenience function.

Anyways, I believe an Object here for the associated receiver would be ideal, as it tremendously expands the range of possible placeholder values. This is one of the great things with Nucleus Placeholder Variables, as those allow for plenty of control over the input string. 

The only case where I can see this being potentially disadvantageous would be for a message string that contains more than one placeholder expecting different sources. You could have something like `{{sponge:player}} is selling {{plugin:item_id}}`, where the item_id placeholder expects an ItemStack(Snapshot) or something similar. Both placeholders could be parsed individually, but the problem I see comes down to message services that only support the passing of a singular source. 

I could be overthinking the concept at this point, so correct me if I'm wrong here.

dualspiral:
_Sorry for the long post, but there is a lot of text here to explain why this entire PR has been written the way it has, mostly to explain why it might seem simplistic._

---

Let's take it back a step and talk through what this is, why I've gone for the minimal set I have, and why adding Nucleus style variables was not proposed here (though they could be).

Right now, we have Nucleus, UltimateChat, NanoChat, PlaceholderAPI<sup>1</sup> (the Sponge version), and FeatherChat (at least, we did), probably others, all offering chat customisation functionality. These systems come in two parts:

* A way to provide variables in a message that require context to render (things that change, such as current server time or who sent a message) - the **placeholders**
* A way to specify what the chat message will look like that contains these dynamic text placeholders - the formatting **template**

Templates are out of scope of this system because they operate entirely in the scope of the plugin generating a message - it makes little to no sense to require a central templating method. On the other hand, some plugins want to provide ways for server owners to add their information to player chat messages, amongst over things. Thus the ultimate goal of this PR is for plugins to provide placeholders these placeholders with little regard to templates.

However, the templating engines **are** important and do need to be considered, in this system:

* **Templates** supply the `PlaceholderText` and _adds_ the context
* **Plugins** supply the `PlaceholderParser` which _acts_ on the context

So, say your plugin provides a `plugin:placeholder` placeholder, and I set Nucleus' chat message prefix template to:
```
{{nucleus:displayname}} {{plugin:placeholder}} > 
```

and then I wrote a message, Nucleus would create a `PlaceholderText` that sets `getAssociatedReceiver()` (bad name, I know) to me, and there would be nothing more it could add. It would pass this to the `PlaceholderParser` which would then consider what to do with it.

> The only case where I can see this being potentially disadvantageous would be for a message string that contains more than one placeholder expecting different sources. 

Note that a templating engine could decide different placeholders have different sources. In fact, for Nucleus' broadcast command, most placeholders have an associated source of the chat sender, but if Nucleus comes across the `{{sender}}` token, it replaces it with a "displayname" token with an associated source of the receiver. It is entirely up to the **templating** engine to decide what placeholders to fetch and what context to build into them.

> This is one of the great things with Nucleus Placeholder Variables, as those allow for plenty of control over the input string.

[Nucleus Placeholder Variables](https://github.com/NucleusPowered/Nucleus/blob/v2/S7/nucleus-api/src/main/java/io/github/nucleuspowered/nucleus/api/placeholder/PlaceholderVariables.java#L13-L135) are a way for a templating engine to provide additional context to a `PlaceholderText`-like object. It's basically a way to provide type safe context _by the templating engine_, in addition to the associated source and argument string (which is generally user specified)<sup>2</sup>.

So, to tie it in to the previous point, this would only become useful if you controlled the templating engine and/or built the `PlaceholderText` yourself - because the variables would be provided by the templating engine. As a result, given the usage of this will be primarily for chat plugins wanting to grab placeholders they know nothing about, I have opted to keep this simple by just supporting an associated object (usually the chat sender, but is the context the placeholder should use) and an argument string (which are effectively user specified options). I felt that the variable system was too complicated for this and was rarely/never going to be used because it _was_ generally used as an implementation detail between different parts of Nucleus - it was actually a `Map<String, Object>` and so was _very_ implementation specific. Suffice to say, it was originally born from laziness on my part and was actually slated for removal!<sup>3</sup>

It must also be said that there is no _requirement_ that the `PlaceholderText` object presented to `PlaceholderParser`s is the one provided by the service. You can roll your own that includes extra context that your templating system provides, that's totally okay. I plan to do this with Nucleus with the variable system, which is why this is specified as a service. As long as the `PlaceholderParser` and `PlaceholderText` conform to the API, any extensions of that, as long as they are properly checked for via `instanceof` checks and the like, aren't going to interfere with the way this system is envisaged to run.

@NickImpact I hope this helps bring some clarity as to why variables are something I actually was keen to avoid.

> I've done some thinking on this, namely going as far as trying to consider a means of a placeholder parser being able to make some call to report a potentially non-empty text only when it meets the source requirements of the parser.

I'm not really sure what you mean by this, but if the `PlaceholderText` your parser is supplied doesn't contain the right information (or is not of the right type because it doesn't have the variables, see above) then you could just reject it with an empty text. If you have your own templating engine, or you use Nucleus' (which will still be available and will continue to have the variable system in a custom `PlaceholderText`), then output text, else, output nothing.

---
<sup>1</sup> PlaceholderAPI is a weird one in that it has a templating engine that _other_ plugins use, rather than managing chat itself, but I'll include it because it's relavent.
<sup>2</sup> As an example, a `econplugin:balance` placeholder than may want to accept additional user input that specifies the currency type to get the balance from. How that is presented to the **templating** engine, we don't care, (Nucleus might do `{{econplugin:balance|GBP}}`, PlaceholderAPI might do `%econplugin:balance_GBP%` - we'd expect `getArgumentString()` to return `GBP` in both cases.)
<sup>3</sup> But if there is enough support for it being central, I'll add it, I just don't see it being useful in the context above.

NickImpact:
Definitely makes sense, I think I was starting to think outside of scope and envision a little too far beyond based on the background of everything. If I made it sound like I was suggesting the addition of variables to this PR, that's my mistake. I wanted to reference them given you actually know the concept. 

Given that, I believe my thoughts started to pander into the template system you're mentioning, which led me on the path to what I wrote up. I was thinking in terms of how I do things now where an entire string of text is translated based on one CommandSource and the variables, which does indeed go into the template system.

As for the point of confusion in the last bit of your response, that was based on some ideas I was trying to brainstorm for the earlier discussion idea, but the concept would have been unnecessary. 

Anyways, back to the topic of the PR, definitely agree with the simple approach, and I don't believe we will need variables with this system. I do believe there's power in using just an Object for the associated receiver, but I don't want to be the deciding factor on that detail.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,176 @@
...
+         * Sets the {@link MessageReceiver} to use as a source of information
+         * for this {@link PlaceholderText} to the supplied {@link Player}.
+         *
+         * @param player The player to associate this text with.
+         * @return This, for chaining
+         *
+         * @see PlaceholderText#getAssociatedReceiver()
+         */
+        default Builder setAssociatedSource(Player player) {
+            UUID uuid = player.getUniqueId();
```
Zidane: I like final things.

dualspiral:
> If I made it sound like I was suggesting the addition of variables to this PR, that's my mistake.

My comment was meant to explain to anyone else reading the issue thread what they were, and then explain why I didn't bring that over.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * <ul>
+     *     <li>
+     *         For a "name" placeholder that prints out the source's name,
+     *         the name would be selected from this source.
+     *     </li>
+     *     <li>
+     *         For a "current world" placeholder that returns a player's
+     *         monetary current world, this would pull the balance from the
+     *         player.
```
ST-DDT: Another current world vs balance issue

ST-DDT (approved):

gabizou (approved):
LGTM

dualspiral:
I have added two methods to the service to make it clear a `Supplier<Object>` can be provided when creating `PlaceholderText`s. Similarly, added a `PlaceholderParser.Builder` for simple parsers.

I do find myself wondering if the service is necessary and if we should just move parsers and texts into the text package. All this really provides is a registry for placeholders and a way to parse them, the service mostly just provides a way to get a parser given a token. I know I wrote the service but it might have been me overthinking it!

I suspect we should keep it for now - see how it goes, revisit for API 8.

src/main/java/org/spongepowered/api/service/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,199 @@
...
+         * <p>If supplying a {@link Player}, use
+         * {@link #setAssociatedObject(Player)}, which will handle creating the
+         * correct supplier for you.</p>
+         *
+         * @param object The {@link Object} to associate this text with.
+         * @return This, for chaining
+         *
+         * @see PlaceholderText#getAssociatedObject()
+         */
+        Builder setAssociatedObject(Object object);
```
ST-DDT: This one does not state its nullability.

ST-DDT:
> I suspect we should keep it for now - see how it goes, revisit for API 8.

I would rather remove it, to avoid causing unnecessary steps when migrating to API 8 as an API user.

AFAICT the service currently has two functions:

1. token or alias to Parser mapping
2. utility methods to create `PlaceholderTexts`

Regarding the first point: We should encourage users to use the original ids instead of aliases, especially since the parameters are parser dependent. If we discover an actual need for it we can still add it later.
Regarding the second point: Static helper methods would probably do the same job.

What I don't like about the current approach is, that I need a `PlaceholderParser` to create a `PlaceholderText` but I there is a `parse(PlaceholderText)` method on the parser. What is supposed to happen if I create the PlaceholderText with parser A and pass it to parser B?

To avoid this problem you should change the PlaceholderParser to only have the following method or no such method at all:

`newTextBuilder() : PlaceholderTextBuilder`


-------------------------------------------
<details>
<summary>
How about a structure like this: (Click to expand)
</summary>

- `PlaceholderParser<T extends PlaceholderText.Builder> extends CatalogType`
  - `newTextBuilder() : T`
- `PlaceholderText extends TextRepresentable`
  - `toText() : Text`

The `PlaceholderParser` is a catalog type, that does not contain any logic except creating a new builder.
By default, the PlaceholderText/Builder "supports" some basic options (those that are currently specified in the `PlaceholderText.Builder`), but the PlaceholderParser might as well return custom subclasses.
Templating plugins can then inspect those custom implementations for attributes to set or they just use the default ones.
Plugin developers that invoke a specific parser (e.g. via catalog class) can use the setters to set specific values directly without guessing how to set it. E.g. set the format via `setFormat("#,###.00")` instead of `setArgumentString("format=#,###.00")`.

Example:

````java
interface PlaceholderParser<T extends PlaceholderText.Builder > extends CatalogType {

    Class<T> getTextBuilderClass(); // <-- Optional, helper for templating plugin introspection (It would return interface class instead of impl class)

    T newTextBuilder();

}
````

````java
interface PlaceholderText extends TextRepresentable {

    interface Builder<T extends Builder> extends ResetableBuilder {

        T setAssociatedObject(Object object);

        T setArgumentString(String argument);

        PlaceholderText build();

    }

}
````

````java
final class PlaceholderParsers {

    public static final PlaceholderParser<BalancePlaceholderTextBuilder> BALANCE = ...;

}
````

````java
interface BalancePlaceholderTextBuilder extends PlaceholderText.Builder<BalancePlaceholderTextBuilder> {

    BalancePlaceholderTextBuilder setCurrencyId(String currencyId);

    BalancePlaceholderTextBuilder setCurrency(Currency currency);

    BalancePlaceholderTextBuilder setAddSendWithdrawTextButton(boolean add);

    [...]

}
````
````java
class BalancePlaceholderTextBuilderImpl extends BalancePlaceholderTextBuilder {

    private Supplier<Currency> currencySupplier = () -> null;

    BalancePlaceholderTextBuilder setCurrencyId(String currencyId) {
        currencySupplier  = () -> toCurrency(currencyId);
        return this;
    }

    BalancePlaceholderTextBuilder setCurrency(Currency currency) {
        currencySupplier  = () -> currency;
        return this;
    }

    BalancePlaceholderTextBuilder setArgumentString(String argument) {
        argument.splitAsStream(";")
                .map(s -> s.split("="))
                .filter(s ->  s[0] == "currency")
                .forEach(s -> setCurrencyId(s[1]));
        return this;
    }
}
````

````java
class BalancePlaceholderText extends PlaceholderText {

    Text toText() {
        try{
             Currency currency = currencySupplier.get().orElse(economyService.getDefaultCurrency()):
             return Text.of(formatter.format(economyService.getBalance(getAssociatedObject(), currency)));
        } catch (Exception e) {
            return Text.EMPTY;
        }
    }

    String getArgumentString() {
        return joiner(";")
                .addIfSet("currency", currencySupplier)
                .addIfSet("format", formatter.getFormat()):
}
````

````java
// From code (known parser in dependency)
PlaceholderParsers.BALANCE.newTextBuilder()
        .setAccountOwner(player)
        .setFormat("#,###.00")
        .toText();
// From template/config (plugin X)
messages.show_current_player_balance=You have ${foobar:balance;format="#,###.00"}.
registry.getCatalog(PlaceholderParser.class, "foobar:balance")
        .setAssociatedObject(player)
        .setArgumentString("format="#,###.00") // Requires lookup from documentation
        .toText();
````
<details>

**Comment by dualspiral** [comment]
First thing I'm going to do is direct you right back to https://github.com/SpongePowered/SpongeAPI/pull/2116#issuecomment-624334776, that explains why things are separated.

> The PlaceholderParser is a catalog type, that does not contain any logic except creating a new builder.

This is not going to happen at all. What you're suggesting is violating _exactly_ what I said about separating templates from parsers. The templating engine **does not care about the parser and its function at all**. It is not, and will not be, starting to inspect the variable and go "hmm, so now I have to cast to placeholder X's builder, but wait, I have no awareness of the plugin, _what do I do_?"

If a plugin wants to provide a way to make a specific text builder with specific parameters, I'm not going to stop them. If you know about the plugin, then what do you need this for? I'm also not going to require it, and make every parser implement a builder. That's madness. Think about what developers want to do. They want to say "here's my context, this is the text I want to display based on that context". Are you saying that I should start adding builders for context on commands? Commands do effectively the same thing as we're doing here, someone provides a string, the command parses it, neither have knowledge of the other.

> Plugin developers that invoke a specific parser (e.g. via catalog class) can use the setters to set specific values directly without guessing how to set it. E.g. set the format via setFormat("#,###.00") instead of setArgumentString("format=#,###.00").

The `PlaceholderParser` can handle the string. **It is not for the templating engine to do.** For the same reason I was talking about removing the service, I am not about to add plugin dependent builders so that the templating engine has to be aware of other plugins. 

**The respononsibility for configuring the parser is the `PlaceholderParser` alone, with the parameters defined in `PlaceholderText`. The templating engine should simply pass on an argument string that the plugin has defined in their docs.**

All plugin developers who write templating systems should need to do is define something like `{{ testplugin:balance|current=GBP}}`, get the `testplugin:balance` parser, set the context, and pass the string `current=GBP` to it. The parser can then decide what to do with that. The templating engine has no role in this. It is a similar reason I didn't include Nucleus' "placeholder variables" system I talked about above.

The point of a `PlaceholderParser` is to simply say "given this context, produce this `Text`". The context is `PlaceholderText`. The `PlaceholderParser` holds no state. You might want to use the parser outside of a `Text.of()` context and just throw in different contexts. `parse(PlaceholderText)`. Plugin developers, by extension, have no reason to create `PlaceholderText`s. Keep it simple. They just want to add their own tokens (parsers) that recieve context (placeholder text).

Finally, note it's only called `PlaceholderText` because it's usable in a `Text` object.

> What I don't like about the current approach is, that I need a PlaceholderParser to create a PlaceholderText but I there is a parse(PlaceholderText) method on the parser.

It would be worth you reading the implementation to understand the logic, but it's basically so `PlaceholderText` can be used as a `TextElement` and know which parser to use.

> What is supposed to happen if I create the PlaceholderText with parser A and pass it to parser B?

Imagine that I create a `PlaceholderContext` object that doesn't include the `PlaceholderParser` so it can't be used in Text objects, but otherwise has the same arguments. The basic sentiment of this will be "it doesn't care", it'll just use it. Unless it specifically checks for the parser, it's only going to care about the associated object and string. So it's really not a problem.

If you _really_ want it, I can make a `PlaceholderContext` that doesn't include the parser, so you have a context to pass around, but I didn't really think that people were going to use it that way.

**Comment by dualspiral** [comment]
I actually have one more point to make that came to mind later, rather than edit the above, I'll add it as another comment:

> Plugin developers that invoke a specific parser (e.g. via catalog class) can use the setters to set specific values directly without guessing how to set it.

That would require a dep on the plugin in question. In that scenario, either the parser class can be referenced directly and if a builder is supplied, it can discover it _without_ needing to rely on Sponge. So parsers really should not enforced to build `PlaceholderText`s at all, such a method is pointless.

What I will probably do is separate out the text such that we have a `PlaceholderContext` that holds the associated object and string, and the `PlaceholderText` will be the following:

```java
interface PlaceholderText extends TextRepresentable {

    PlaceholderContext getContext();

    PlaceholderParser getParser();

    Text toText() { // impl will be something like
        getParser().parse(getContext());
     }
}
```

It changes very little, but will mean that the context has no dep on the parser whatsoever. I will not change anything else to complicate matters for templating systems or plugins that want to provide a simple placeholder - they can do it in one class or with our builder.

---

Do remember what this placeholder system is all about. It is about unifying all the current Placeholder based chat systems so that plugins can just provide placeholders and we can all use them. This is intended as a _simple_ way to provide a placeholder with a simple way for templating engines to interface with it without having to understand parsers. This puts the burden of correctly filling out the argument string on the user, as it should be, to make it simple for templating engines to do what they do best, provide a template.

If you want to get a text outside of this with a parameter builder, it might be worth asking the plugin developer to provide a builder outside of the parser altogether. That, however, is outside of the scope of what this will do.

ST-DDT:
Thank you. I understand your argument, however I dont fully agree with it. We both stated our arguments, so there is no need to repeat them again.

dualspiral:
I've ditched the service and there are now simply three components: `PlaceholderParser`, `PlaceholderContext` and `PlaceholderText` - the service didn't add value as explained above, except to increase awareness of the system. Have a look at the added test plugin in the Common PR to see how it all fits together in code, simply put:

* Plugins wanting to provide placeholders provide `PlaceholderParsers`
* Templating plugins wanting to use placeholders grab these from the registry, fill out the `PlaceholderContext` (based on the token they've been provided, users should use the documentation of the plugins providing the placeholder to configure correctly) and optionally use a `PlaceholderText` to add the placeholder to `Text` objects.

As this is a little bit different from before because of the movement of stuff from PlaceholderText to PlaceholderContext and the removal of the service, I've dismissed the reviews - so a fresh look would be nice. The general idea is the same, however. Feedback is welcome on the javadocs etc. I'll update the Docs PR once this is all done and dusted.

src/main/java/org/spongepowered/api/text/placeholder/PlaceholderText.java
```diff
@@ -0,0 +1,122 @@
...
+ * <p>A {@link PlaceholderText} is the collection of a {@link PlaceholderParser}
+ * along with contextual data in the supplied {@link PlaceholderContext},
+ * enabling its use in a {@link Text} object.</p>
+ *
+ * <p>While such placeholders will generally be built from tokenised strings,
+ * these objects make no assumption about the format of text templating. Such a
+ * system can therefore be used by other templating engines without conforming
+ * to a particular standard.</p>
+ */
+public interface PlaceholderText extends TextRepresentable {
```
ST-DDT, line 53: Maybe add a note, that the placeholder will be resolved when converting it to a `Text` and not during the creation of the `PlaceholderText` instance.

ST-DDT (approved):

PR #2120, opened 2020-05-19
Add totem effect to ParticleTypes

fingoltin:
Pretty self explanatory I think. Common PR is [here](https://github.com/SpongePowered/SpongeCommon/pull/2613).


Faithcaio:
Was that the only one missing?

PR #2124, opened 2020-05-28
Fix unsafe copy method for Location

danorris709:
In the super method for copy it says the following:
image
However, the current implementation doesn't clone the Vector object holding the position of the location.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/Location.java
+++ b/src/main/java/org/spongepowered/api/world/Location.java
@@ -884,7 +884,7 @@ public final class Location<E extends Extent> implements DataHolder {
 
     @Override
     public Location<E> copy() {
-        return new Location<>(getExtent(), getPosition());
+        return new Location<>(getExtent(), getPosition().clone());
     }
 
     @Override


dualspiral:
It's actually okay as it is. Surely if we clone the position, we'd have to clone the `Extent` too?

It boils down to three things:

1: `Vector3d` is immutable and so is safe to just reuse.
2: The position is not actually a `BaseValue` or, indeed, part of a value container.
3: The javadocs for the class says the following:

https://github.com/SpongePowered/SpongeAPI/blob/4a53e6d12a36ddc2698996169d512658faa1b38e/src/main/java/org/spongepowered/api/world/Location.java#L76-L78

The location itself is not a data holder, but is a proxy for a specfic block in the selected `Extent` - sure, perhaps the Javadocs should be clearer, but cloning the position is an unnecessary use of memory and processing time _especially_ as the position is immutable.

Thank you for looking at this and providing a PR, but I'm closing this for the reasons above.

danorris709:
> It's actually okay as it is. Surely if we clone the position, we'd have to clone the `Extent` too?
> 
> It boils down to three things:
> 
> 1: `Vector3d` is immutable and so is safe to just reuse.
> 2: The position is not actually a `BaseValue` or, indeed, part of a value container.
> 3: The javadocs for the class says the following:
> 
> https://github.com/SpongePowered/SpongeAPI/blob/4a53e6d12a36ddc2698996169d512658faa1b38e/src/main/java/org/spongepowered/api/world/Location.java#L76-L78
> 
> The location itself is not a data holder, but is a proxy for a specfic block in the selected `Extent` - sure, perhaps the Javadocs should be clearer, but cloning the position is an unnecessary use of memory and processing time _especially_ as the position is immutable.
> 
> Thank you for looking at this and providing a PR, but I'm closing this for the reasons above.

I see, I didn't realize it was immutable because this method's name is misleading. Given it's immutable should the "copy" method not be deprecated, and only return ``this``? Surely it's also a waste of memory to return a copy of an immutable object?

PR #2130, opened 2020-06-14
New Explosion Event for BlockExplosionResistance

### LINKED ISSUE #2093: Event to alter block blast resistance during explosion calculations
This is a feature that I intend to make use of and would like to open a PR in the future.
However, I am not sure what would be the best way to implement this feature.

Pulverizer:
[ **SpongeAPI** | [SpongeCommon ](https://github.com/SpongePowered/SpongeCommon/pull/2640)]

Resolves https://github.com/SpongePowered/SpongeAPI/issues/2093

The intention of this PR is to allow plugins to intercept the fetching of a block's explosion resistance and adjust the value dynamically based on the block's location and type.

Example Usage:
```
   public void changeBlockResistance(ExplosionEvent.BlockResistance event) {

        for (Map.Entry<Vector3i, Pair<BlockState, Float>> entry : event.getBlocks().entrySet()) {
            Vector3i blockPos = entry.getKey();
            Pair<BlockState, Float> pair = entry.getValue();


            float newResistance = pair.getValue();
            if (Settings.DurabilityOverride != null && Settings.DurabilityOverride.containsKey(pair.getKey().getType())) {
                newResistance = Settings.DurabilityOverride.get(pair.getKey().getType());
            }

            for (Craft craft : CraftManager.getInstance().getCraftsFromLocation(new Location<>(event.getTargetWorld(), blockPos))) {
                float testNum = pair.getValue() * (1F + ((float) craft.getSize() / 50000F));

                if (testNum > newResistance) {
                    newResistance = testNum;
                }
            }

            event.getBlocks().put(entry.getKey(), Pair.of(pair.getKey(), newResistance));
        }
    }
```

### DIFF
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -24,18 +24,14 @@
  */
 package org.spongepowered.api.event.world;
 
-import com.flowpowered.math.vector.Vector3i;
-import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.block.ChangeBlockEvent;
 import org.spongepowered.api.event.entity.AffectEntityEvent;
-import org.spongepowered.api.util.Tuple;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.explosion.Explosion;
 
-import java.util.HashMap;
 import java.util.List;
 
 /**
@@ -93,19 +89,4 @@ public interface ExplosionEvent extends Event {
      * updated and/or changed.
      */
     interface Post extends ExplosionEvent, ChangeBlockEvent.Post {}
-
-    /**
-     * An event that is fired when the Explosion fetches the explosion resistance of a block.
-     * Provides the world, block position, block state, and default and current explosion resistance.
-     */
-    interface BlockResistance extends ExplosionEvent, TargetWorldEvent {
-
-        /**
-         * Gets the block positions that will potentially be effected as Vector3i mapped to a Pair containing the BlockState and
-         * explosion resistance of the block located at that position.
-         *
-         * @return The blocks that the explosion may effect
-         */
-        HashMap<Vector3i, Tuple<BlockState, Float>> getBlocks();
-    }
 }
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -25,12 +25,12 @@
 package org.spongepowered.api.event.world;
 
 import com.flowpowered.math.vector.Vector3i;
-import org.apache.commons.lang3.tuple.Pair;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.block.ChangeBlockEvent;
 import org.spongepowered.api.event.entity.AffectEntityEvent;
+import org.spongepowered.api.util.Tuple;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.explosion.Explosion;
@@ -106,6 +106,6 @@ public interface ExplosionEvent extends Event {
          *
          * @return The blocks that the explosion may effect
          */
-        HashMap<Vector3i, Pair<BlockState, Float>> getBlocks();
+        HashMap<Vector3i, Tuple<BlockState, Float>> getBlocks();
     }
 }
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -25,7 +25,7 @@
 package org.spongepowered.api.event.world;
 
 import com.flowpowered.math.vector.Vector3i;
-import javafx.util.Pair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -88,5 +88,6 @@ public interface ExplosionEvent extends Event {
      * affected blocks) will be included. This is where the block changes can be
      * updated and/or changed.
      */
-    interface Post extends ExplosionEvent, ChangeBlockEvent.Post {}
+    interface Post extends ExplosionEvent, ChangeBlockEvent.Post {
+    }
 }
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.api.event.world;
 
+import com.flowpowered.math.vector.Vector3i;
+import javafx.util.Pair;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
@@ -33,6 +35,7 @@ import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.explosion.Explosion;
 
+import java.util.HashMap;
 import java.util.List;
 
 /**
@@ -93,36 +96,16 @@ public interface ExplosionEvent extends Event {
 
     /**
      * An event that is fired when the Explosion fetches the explosion resistance of a block.
-     * Provides the world, block position, block state, and current explosion resistance.
+     * Provides the world, block position, block state, and default and current explosion resistance.
      */
-    interface FetchBlockExplosionResistance extends ExplosionEvent, TargetWorldEvent {
+    interface BlockResistance extends ExplosionEvent, TargetWorldEvent {
 
         /**
-         * Gets the explosion resistance of the block which may have already been altered.
+         * Gets the block positions that will potentially be effected as Vector3i mapped to a Pair containing the BlockState and
+         * explosion resistance of the block located at that position.
          *
-         * @return The current explosion resistance of the block
+         * @return The blocks that the explosion may effect
          */
-        float getResistance();
-
-        /**
-         * Sets the explosion resistance of the block.
-         *
-         * @param resistance The new explosion resistance of the block
-         */
-        void setResistance(float resistance);
-
-        /**
-         * Gets the location of the block being affected.
-         *
-         * @return The location of the block
-         */
-        Location<World> getLocation();
-
-        /**
-         * Gets the BlockState of the block being affected.
-         *
-         * @return The BlockState of the block
-         */
-        BlockState getBlockState();
+        HashMap<Vector3i, Pair<BlockState, Float>> getBlocks();
     }
 }
--- a/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.api.event.world;
 
+import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.block.ChangeBlockEvent;
@@ -89,4 +90,39 @@ public interface ExplosionEvent extends Event {
      * updated and/or changed.
      */
     interface Post extends ExplosionEvent, ChangeBlockEvent.Post {}
+
+    /**
+     * An event that is fired when the Explosion fetches the explosion resistance of a block.
+     * Provides the world, block position, block state, and current explosion resistance.
+     */
+    interface FetchBlockExplosionResistance extends ExplosionEvent, TargetWorldEvent {
+
+        /**
+         * Gets the explosion resistance of the block which may have already been altered.
+         *
+         * @return The current explosion resistance of the block
+         */
+        float getResistance();
+
+        /**
+         * Sets the explosion resistance of the block.
+         *
+         * @param resistance The new explosion resistance of the block
+         */
+        void setResistance(float resistance);
+
+        /**
+         * Gets the location of the block being affected.
+         *
+         * @return The location of the block
+         */
+        Location<World> getLocation();
+
+        /**
+         * Gets the BlockState of the block being affected.
+         *
+         * @return The BlockState of the block
+         */
+        BlockState getBlockState();
+    }
 }
--- a/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
+++ b/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
@@ -25,7 +25,6 @@
 package org.spongepowered.api.world.explosion;
 
 import org.spongepowered.api.Sponge;
-import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.explosive.Explosive;
 import org.spongepowered.api.util.ResettableBuilder;
 import org.spongepowered.api.world.Locatable;
@@ -133,6 +132,15 @@ public interface Explosion extends Locatable {
     }
 
     /**
+     * Gets the {@link ResistanceCalculator} of the Explosion.
+     *
+     * @return The ResistanceCalculator or null if not supported by the implementation
+     */
+    default Optional<ResistanceCalculator> getResistanceCalculator() {
+        return Optional.empty();
+    }
+
+    /**
      * A builder for {@link Explosion}.
      */
     interface Builder extends ResettableBuilder<Explosion, Builder> {
@@ -232,6 +240,17 @@ public interface Explosion extends Locatable {
          * @return This builder, for chaining
          */
         default Builder knockback(double knockback) {
+            return this;
+        }
+
+        /**
+         * Sets the method by which the explosion resistance of an affected block can be obtained or calculated.
+         * Expects a lambda in the form of (BlockState, Vector3i, Explosion) that returns a float.
+         *
+         * @param resistanceCalculator The lambda to use
+         * @return This builder, for chaining
+         */
+        default Builder resistanceCalculator(ResistanceCalculator resistanceCalculator) {
             return this;
         }
 
--- a/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
+++ b/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
@@ -31,8 +31,9 @@ import org.spongepowered.api.world.Locatable;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 
-import javax.annotation.Nullable;
 import java.util.Optional;
+
+import javax.annotation.Nullable;
 
 /**
  * Represents an explosion in a {@link World}.
@@ -104,8 +105,8 @@ public interface Explosion extends Locatable {
      *
      * <p>This value indicates how close to being symmetrical the explosion is.
      * A value of zero indicates a completely symmetrical blast (in all three
-     * dimensions). A larger value indicates a lower likelihood of symmetry. A
-     * value of one indicates the platform default.</p>
+     * dimensions). A larger value indicates a lower likelihood of symmetry.
+     * A value of one indicates the platform default.</p>
      *
      * <p>Note, this is a hint to the implementation. Implementations may not
      * provide the means to produce semi-random form explosions.</p>
@@ -120,8 +121,8 @@ public interface Explosion extends Locatable {
      * Gets the relative strength of the knockback applied to nearby
      * objects that can be knocked back.
      *
-     * <p>Note that the default behavior and strength is not defined here. A
-     * return value of 1 simply indicates the default behavior which is
+     * <p>Note that the default behavior and strength is not defined here.
+     * A return value of 1 simply indicates the default behavior which is
      * implementation dependent.</p>
      *
      * @return The multiple by which the knockback of entities will be changed
@@ -216,11 +217,11 @@ public interface Explosion extends Locatable {
          *
          * <p>This value indicates how close to being symmetrical the explosion is.
          * A value of zero indicates a completely symmetrical blast (in all three
-         * dimensions). A larger value indicates a lower likelihood of symmetry. A
-         * value of one indicates the platform default.</p>
+         * dimensions). A larger value indicates a lower likelihood of symmetry.
+         * A value of one indicates the platform default.</p>
          *
-         * <p>Note, this is a hint to the implementation. Implementations may not
-         * provide the means to produce semi-random form explosions.</p>
+         * <p>Note, this is a hint to the implementation. Implementations might
+         * not provide the means to produce semi-random form explosions.</p>
          *
          * @return This builder, for chaining
          */
@@ -232,8 +233,8 @@ public interface Explosion extends Locatable {
          * Sets the relative strength of the knockback applied to nearby
          * objects that can be knocked back.
          *
-         * <p>Note that the default behavior and strength is not defined here. A
-         * return value of 1 simply indicates the default behavior which is
+         * <p>Note that the default behavior and strength is not defined here.
+         * A return value of 1 simply indicates the default behavior which is
          * implementation dependent.</p>
          *
          * @param knockback The knockback multiple
@@ -244,7 +245,8 @@ public interface Explosion extends Locatable {
         }
 
         /**
-         * Sets the method by which the explosion resistance of an affected block can be obtained or calculated.
+         * Sets the method by which the explosion resistance of an affected
+         * block can be obtained or calculated.
          *
          * @param resistanceCalculator The calculator
          * @return This builder, for chaining
--- a/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
+++ b/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
@@ -31,9 +31,8 @@ import org.spongepowered.api.world.Locatable;
 import org.spongepowered.api.world.Location;
 import org.spongepowered.api.world.World;
 
-import java.util.Optional;
-
 import javax.annotation.Nullable;
+import java.util.Optional;
 
 /**
  * Represents an explosion in a {@link World}.
--- a/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
+++ b/src/main/java/org/spongepowered/api/world/explosion/Explosion.java
@@ -132,8 +132,9 @@ public interface Explosion extends Locatable {
 
     /**
      * Gets the {@link ResistanceCalculator} of the Explosion.
+     * Optional#empty() indicates default behaviour.
      *
-     * @return The ResistanceCalculator or null if not supported by the implementation
+     * @return The ResistanceCalculator
      */
     default Optional<ResistanceCalculator> getResistanceCalculator() {
         return Optional.empty();
@@ -244,9 +245,8 @@ public interface Explosion extends Locatable {
 
         /**
          * Sets the method by which the explosion resistance of an affected block can be obtained or calculated.
-         * Expects a lambda in the form of (BlockState, Vector3i, Explosion) that returns a float.
          *
-         * @param resistanceCalculator The lambda to use
+         * @param resistanceCalculator The calculator
          * @return This builder, for chaining
          */
         default Builder resistanceCalculator(ResistanceCalculator resistanceCalculator) {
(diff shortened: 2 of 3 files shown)


src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
```diff
@@ -89,4 +90,39 @@
      * updated and/or changed.
      */
     interface Post extends ExplosionEvent, ChangeBlockEvent.Post {}
-}
+
+    /**
+     * An event that is fired when the Explosion fetches the explosion resistance of a block.
+     * Provides the world, block position, block state, and current explosion resistance.
+     */
+    interface FetchBlockExplosionResistance extends ExplosionEvent, TargetWorldEvent {
```
dualspiral: Why does this need to be a separate event? Why can't such things go in `Pre` and/or `Detonate`?

Pulverizer: `Pre` fires before the rays are cast and block resistances checked.
`Detonate` fires after the blocks have already been processed and would mean having to do all the ray casting again.

dualspiral: This is going to get expensive because you've done it on a per block basis. I also don't really agree with the idea of exposing resistances when we could have a proper filter on `Detonate`. 

Why do you want to change the resistance? If it is just to guarantee that a block will be removed (or vice versa) then it really should be in `Detonate`. Indeed, I'd just prefer capturing all blocks that have been processed by the explosion, whether or not they would explode, and provide them in `Detonate`, with whether or not they're going to be removed. A plugin can then work though that list and decide if that needs changing, even with a random number generator if required.

If this event is to stay (someone else can decide that), this needs to be a bulk event, not a per block event. Constructing and firing this event hundereds of times in a tick (potentially) is going to _really_ hurt the performance of the server.

Pulverizer: You'd have to capture all blocks in the radius of the explosion, not just the ones the explosion is currently effecting. Maybe this would be more optimised.

Secondly, using `Detonate` in my use case, would require running raytracing in the listener after the source code and rerunning the entity knockback too.

So the best new option I can see is gathering all the blocks in the radius and providing a map of BlockType and Location to explosion resistance that listeners can edit. The map returned at the end can then be used for the raytracing.

dualspiral: > Why do you want to change the resistance?

Please answer my question. All you've done is say "in my use case" and come up with some reason why this doesn't work for you without explaining what it is you're trying to do. The question has been asked for a reason - to get an idea of what you _really_ want to do and so figure out what, if any, a better solution is.

Pulverizer: Also, I think `Pre` might be getting triggered in the wrong place?

Because I would suggest putting this better way of doing it there but it is located with the WorldServerMixin but would be needed within the ExplosionMixin for this.

Pulverizer: > > Why do you want to change the resistance?
> 
> Please answer my question. All you've done is say "in my use case" and come up with some reason why this doesn't work for you without explaining what it is you're trying to do. The question has been asked for a reason - to get an idea of what you _really_ want to do and so figure out what, if any, a better solution is.

Sorry, I missed that part and this is where I struggle to explain. To best explain what I am trying to achieve by adding this feature to the API I should probably start by explaining my plugin.

My plugin is a port of an existing Spigot plugin that allows players to pilot and move a collection of blocks called a "craft". Some examples of this:




One of its features is to override the explosion resistance of blocks to create various armour types for these "crafts".

I want to implement this feature and expand it so that by using a config setting blocks can be made stronger or weaker, and blocks that are located within a "craft" receive a bonus to explosion resistance based on the size of the "craft".

If you want different details or a better explanation in a specific area of the above I will need help to understand what you require.

Pulverizer: While working to move it to a single event per explosion I have discovered that explosions continue a given raytrace outside of the explosion radius if it does not meet enough resistance.

This could be optimized so that it stops if it exits the radius, however, this may have unforeseen side-effects. What would you advise?

Yeregorix:
I'm really worrying about performance. `FetchBlockExplosionResistance` would be fired multiple times for each block. Given that an explosion affects many blocks and they can be many explosions at once on a server I expect a huge cost from this approach. I might be wrong, it needs to be tested.

I suggest a different approach, the one I used in my plugin SuperPiston. The problem was similar, I needed to replace the logic of each block affected by a piston. See https://github.com/Yeregorix/SuperPiston/blob/master/src/main/java/net/smoofyuniverse/superpiston/api/event/PistonStructureCalculationEvent.java for example.

Here is the trick, make two cancellable events `Pre` and `Post`. During `Pre`, plugins must choose a `Calculator`, then the explosion logic is fully replaced and calculated using the `Calculator`, and then `Post` is fired, with the result of the explosion. 

This approach has the advantage to give more freedom to plugins, they can fully replace the explosion logic by a custom one, totally different. It's a bulk operation, no need to fire multiple events for each block. It ensures only one logic is calculated to avoid performance overhead, for example if the vanilla logic is fully replaced it will never be calculated, only the custom one will. The only problem of this approach is that it's possible but difficult for two "conflicting" plugins to merge their `Calculator`s.

This approach would be better **in my opinion**, but I let people here debate about it.

NB: `Pre` and `Post` was the names I chose in my plugin, in case of modifying the current api, editing `Pre` and `Detonate` should be more appropriate.

Pulverizer:
> While working to move it to a single event per explosion I have discovered that explosions continue a given raytrace outside of the explosion radius if it does not meet enough resistance.
> 
> This could be optimized so that it stops if it exits the radius, however, this may have unforeseen side-effects. What would you advise?

@dualspiral @Yeregorix 

Also, if I was to do as you suggest Yeregorix where should the new classes be located?

src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
```diff
@@ -24,6 +24,9 @@
  */
 package org.spongepowered.api.event.world;
 
+import com.flowpowered.math.vector.Vector3i;
+import javafx.util.Pair;
```
clienthax: Swap for org.spongepowered.api.util.Tuple?, Ideally we don't want javafx imports

clienthax (changes requested):

src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java
Pulverizer: Ah, okay. I've not used this sort of thing before. I'll try to fix it tomorrow morning.

clienthax: We have a tuple class here which is library non dependent it would be better to use this https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/util/Tuple.java

clienthax (changes requested):

Zidane:
@dualspiral Yay or nay?

dualspiral:
It's active _but_ I really do not think this is the right way to go. I still think this should be on `Pre` and/or `Detonate` rather than its own event, but I admit I do admit I've not looked into what it would take for it to go onto one of these events and I've been working on other things (unsurprisingly).

My major concern is that by putting this onto 7, we're going to be expected to put this onto 8 in the same way. I'm not convinced this is right for 8. I also think that, having a quick look at the Common PR, there may be some work involved before this is ready for pulling (there are numerous style issues and I think some of the methods can indeed be minimised).

If the use case is to have a simple multiplier for explosion resistance that can be specified in `Pre`, _maybe_ on a per-block type/state level, that can then be used in the calculations, I'd be more receptive to that - it would be minimal on performance and we're not having a whole event just to tack on numbers to blocks and potentially hurting performance.

For those reasons, I'm leaning towards a no, this should not be a separate event.

Pulverizer:
> It's active _but_ I really do not think this is the right way to go. I still think this should be on `Pre` and/or `Detonate` rather than its own event, but I admit I do admit I've not looked into what it would take for it to go onto one of these events and I've been working on other things (unsurprisingly).

As I think I explained previously, `Pre` is not located within the ExplosionMixin (not sure why not tho?), and `Detonate` is too late unless we want to run the entire block damage, entity damage, and entity knockback code again.

> My major concern is that by putting this onto 7, we're going to be expected to put this onto 8 in the same way. I'm not convinced this is right for 8. I also think that, having a quick look at the Common PR, there may be some work involved before this is ready for pulling (there are numerous style issues and I think some of the methods can indeed be minimised).

Yeah, I'm trying my best but I don't know how to correctly do a lot of stuff.

> If the use case is to have a simple multiplier for explosion resistance that can be specified in `Pre`, _maybe_ on a per-block type/state level, that can then be used in the calculations, I'd be more receptive to that - it would be minimal on performance and we're not having a whole event just to tack on numbers to blocks and potentially hurting performance.
> 
> For those reasons, I'm leaning towards a no, this should not be a separate event.

Unfortunately, my use case requires block location as well and it does not provide a multiplier, it completely replaces the existing number. Perhaps we should review why the `Pre` event is not located within the Explosion Mixin. I suspect it is so that we don't waste time processing code unnecessarily?

@dualspiral

Pulverizer:
Perhaps, I could drastically alter the Mixin so that there are two versions of the effected method and use an if-else containing the ShouldFire for the Block Resistance Event to decide which version to use. This would mean that performance is only affected when there is a listener for the event?

dualspiral:
So, why not just have new methods on `Pre`, `get/setExplosionResistanceCalcuator` which is a `ResistanceCalculator` which takes a `Location`, the current resistance as a `float`, and returns the new resistance as a `float`. We can then inject that into the explosion code to make the calculation. That way, we avoid a new event and we tell the engine how to calculate resistance for a given block.

I think that's in line with what @Yeregorix suggested and it'll completely remove the requirement to have a second event. It could even go on the `Explosion` builder too, but let's not worry about that for now.

Pulverizer:
@dualspiral This seems like a good solution once we have resolved a few concerns that were raised.

> This approach has the advantage to give more freedom to plugins, they can fully replace the explosion logic by a custom one, totally different. It's a bulk operation, no need to fire multiple events for each block. It ensures only one logic is calculated to avoid performance overhead, for example if the vanilla logic is fully replaced it will never be calculated, only the custom one will. The only problem of this approach is that it's possible but difficult for two "conflicting" plugins to merge their `Calculator`s.

How will we handle multiple plugins trying to alter the results?

And where should the new Interface and Implementation classes for `ExplosionResistanceCalculator` be located?

dualspiral:
> How will we handle multiple plugins trying to alter the results?

I, personally, don't care too much about that, because if you can get the resistance calculator you can chain it in your own. If you do though (and others), then the event needs to become more complex, where you supply your own calculator which adds it to a list of calculators and order of registration will determine order of applicaton.

> And where should the new Interface and Implementation classes for ExplosionResistanceCalculator be located?

Interface should be wherever the explosion classes are. I'd imagine it's going to be a `FunctionalInterface`, so there wouldn't need to be an implementation class - the default could always be `Optional.empty()` so the impl wouldn't need one.

Yeregorix:
> How will we handle multiple plugins trying to alter the results?

If you do this way it would not be the role of the implementation but the role of the plugins.

eg.
Plugin A: `event.setCalculator(new MyCalculatorA())`

then

Plugin B have two choices:
erase A's calculator: `event.setCalculator(new MyCalculatorB())`
or
register a special calculator based on the previous one: `event.setCalculator(new MyCalculatorAdapterB(event.getCalculator()))`

Even without speaking of calculators, this is always the case for all events of the API. When you set a value in an event, you can ignore the previous value, or do calculation based on it.

Pulverizer:
How does one access the functional interface ResistanceCalculator in their plugin if it is located in SpongeAPI? Should I have put it somewhere else?

dualspiral:
The interface should be in the API - when I said explosion classes, I meant interfaces in the API.

Pulverizer:
> The interface should be in the API - when I said explosion classes, I meant interfaces in the API.

That's what I thought. I've put it at `org.spongepowered.api.world.explosion.ResistanceCalculator`. However, when I do `gradlew build install` and point my plugin to it the IDE errors when my plugin tries to use it "Cannot resolve symbol 'ResistanceCalculator'" and asks if I want to add `SpongeVanilla.SpongeCommon.SpongeAPI.main` to the dependencies.

dualspiral:
That's likely the IDE being weird. If you're using IntelliJ, try invalidating the caches and restarting. Don't know about other IDEs.

Pulverizer:
@dualspiral I think that this PR is complete where functionality is concerned. Any remaining issues I suspect are bad coding/formatting on my part.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -132,6 +130,15 @@ default double getKnockback() {
         return 1;
     }
 
+    /**
+     * Gets the {@link ResistanceCalculator} of the Explosion.
+     *
+     * @return The ResistanceCalculator or null if not supported by the implementation
+     */
+    default Optional<ResistanceCalculator> getResistanceCalculator() {
```
dualspiral, line 140: I wrote the common comments before I saw this and the javadoc comments.

> or null if not supported by the implementation

I'd... rather not put that. Simply, `Optional#empty()` will indicate default behaviour.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -235,6 +242,17 @@ default Builder knockback(double knockback) {
             return this;
         }
 
+        /**
+         * Sets the method by which the explosion resistance of an affected block can be obtained or calculated.
```
dualspiral: Javadocs must be at most 80 characters wide.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
```diff
@@ -235,6 +242,17 @@ default Builder knockback(double knockback) {
...
+         * Expects a lambda in the form of (BlockState, Vector3i, Explosion) that returns a float.
+         *
+         * @param resistanceCalculator The lambda to use
```
dualspiral: They don't have to add a lambda here. They can create a subclass in the traditional way if they prefer.

`The calculator` will suffice.

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
```diff
@@ -0,0 +1,44 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.explosion;
+
+import com.flowpowered.math.vector.Vector3i;
+import org.spongepowered.api.block.BlockState;
+
+/**
+ * A Functional Interface for the fetching of a Block's explosion resistance during an explosion
```
dualspiral: Well... we already know it's a functional interface.

Javadocs just need to describe what the class/method does.

```
Calculates a {@link Block}s resistance to a {@link Explosion}.
```

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@FunctionalInterface
+public interface ResistanceCalculator {
+    /**
+     * A blueprint for a lambda that takes three variables; BlockState, Vector3i, and Explosion and returns a float.
```
dualspiral: The javadoc comment is pointless because we already know what has been written here. An interface is a blueprint, and we can see return and parameter types.

Focus on what the method does and the context around that in a doc comment - it should tell you what the code cannot. For example, what values can I return? Is it 0 to 100? 0 to 1? Is a negative value valid?

```
Sets the explosion resistance for a block at a given location during an {@link Explosion}.

<p>The resistance must a value between zero and one. A resistance of zero will ensure that the given block will be affected by the explosion. A resistance of one ensures that the block will be unaffected.</p>
```

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param blockState    The state of the block
+     * @param blockPosition The position of the block in the World
+     * @param explosion     The explosion affecting the block
+     * @return The explosion resistance of the block
+     */
+    float calculateResistance(final BlockState blockState, final Vector3i blockPosition, final Explosion explosion);
```
dualspiral: It's not obvious how to get a `World` here. Should the block position be a `Location<World>` instead, and drop the block state (which you can get from the location - fairly standard thing we do?)

What about if a plugin wants to affect obsidian blocks only? How would they determine the resistance for other blocks?

Earlier, I said:

> a `ResistanceCalculator` which takes a Location, **the current resistance as a float**, and returns the new resistance as a float.

That's the reason why I suggested it.

dualspiral (changes requested):
The calculator isn't exactly what I had in mind because it doesn't take into account what you might do if you only wanted to alter _some_ of the block resistances, or apply a multiplier.

Also, the javadocs do need some work, also please conform to our style guide: https://docs.spongepowered.org/stable/en/contributing/implementation/codestyle.html

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
Pulverizer: Could you clarify what you mean? I don't quite understand.

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
Pulverizer: > It's not obvious how to get a `World` here. Should the block position be a `Location<World>` instead, and drop the block state (which you can get from the location - fairly standard thing we do?)

If that's the standard way of doing it then I'll rewrite that section.

> What about if a plugin wants to affect obsidian blocks only? How would they determine the resistance for other blocks?
> 
> Earlier, I said:
> 
> > a `ResistanceCalculator` which takes a Location, **the current resistance as a float**, and returns the new resistance as a float.
> 
> That's the reason why I suggested it.

Uh... Mistakes have been made 🤕 . I'll review this and get back to you.

src/main/java/org/spongepowered/api/world/explosion/Explosion.java
dualspiral: Clarify the javadocs, don't refer to implementations.

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
dualspiral: I mean, you can keep the block state in, to be honest, it's already been determined. It's more important that the default resistance is provided.

Pulverizer: I believed it would be more optimized to pass it than fetch it again.
The world can be obtained through `explosion.getWorld()`.

Pulverizer:
Hopefully, I haven't marked anything as resolved without actually resolving it.
I'll recheck all code style related stuff and push a final commit after everything else is confirmed to be correct.

Pulverizer:
I think the code style is now correct?

dualspiral:
I'm iffy on the block resistance because resistance seems to be a semi arbitary number and I'd imagine that such a number would be different across differing implementations. I don't like that we're so tied to the implementation on that.

I've asked a couple of people to look at this aspect but had no respose. I'm gonna ask again because I'm really not sure about the notion of telling people to go look at the Minecraft wiki to get the values on this.

@SpongePowered/developers. Are we okay with block resistance being arbitary and thus kind of trap all other impls into this notion or not? If we are, then minor issues aside, this looks okay to me. If we are not, then I think this is ultimately going to either get closed or simply based on a multiplier.

The use case seems to me to be to expose an internal resistance and potentially modify it. Is this one of those cases where we say "this is clearly impl dependent, you may just want to use the internals"?

src/main/java/org/spongepowered/api/world/explosion/ResistanceCalculator.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Calculates a Block's resistance to a {@link Explosion}.
+ */
+@FunctionalInterface
+public interface ResistanceCalculator {
+
+    /**
+     * Sets the explosion resistance for a block at a given location
+     * during an {@link Explosion}.
```
gabizou, line 38: What am I supposed to understand about setting the resistance for a block at a given location?

Is it that I decrement the returned value from the original resistance?

Can I chain on the vanilla calculator?

Is it possible to return some bounds?

What are the numbers meaning? Do I return `0` to say "don't break me" or do I use `1f` to say "I'm impenetrable? Or is it a resistance factor based on `0-10`?

gabizou (changes requested):
I'm hazy on what the Calculator is actually to achieve.. I get that it's to "replace" the vanilla one for a particular event, but I don't know off the java docs what to make of it. Do I return some values between some bounds?

Pulverizer:
> I'm iffy on the block resistance because resistance seems to be a semi arbitary number and I'd imagine that such a number would be different across differing implementations. I don't like that we're so tied to the implementation on that.
> 
> I've asked a couple of people to look at this aspect but had no respose. I'm gonna ask again because I'm really not sure about the notion of telling people to go look at the Minecraft wiki to get the values on this.
> 
> @SpongePowered/developers. Are we okay with block resistance being arbitary and thus kind of trap all other impls into this notion or not? If we are, then minor issues aside, this looks okay to me. If we are not, then I think this is ultimately going to either get closed or simply based on a multiplier.
> 
> The use case seems to me to be to expose an internal resistance and potentially modify it. Is this one of those cases where we say "this is clearly impl dependent, you may just want to use the internals"?

How would one use internals? If this becomes a multiplier it won't provide what I require for my plugin. I do agree that this might not be the best, but how else could it be done?

I would point out that the arbitrary numbers exist because blast resistance in vanilla Minecraft is just that.

dualspiral:
> I would point out that the arbitrary numbers exist because blast resistance in vanilla Minecraft is just that.

And I would point out that one of the points of our API is to, effectively, provide a layer between Minecraft and plugins that provides predictable, known behaviour. We also consider third party implementations who might not want to use Mojang's blast resistance mechanic.

Take @gabizou's comment above. Some of it is geared at the javadocs, but there is also this:

> Is it possible to return some bounds?
>
> What are the numbers meaning? Do I return 0 to say "don't break me" or do I use 1f to say "I'm impenetrable? Or is it a resistance factor based on 0-10?

which I think is the sticking point and why I'm reticent to go with this. [Indeed, I also reference this here](https://github.com/SpongePowered/SpongeAPI/pull/2130#discussion_r447060681) and while I simply say "what can you do", linking to the Minecraft wiki, which can change over time, isn't the way to go. Our docs should really be self contained.

If this numbers were 0 to 1, that's something that makes sense - but it's not. I understand how Minecraft does it (now I've looked at it) but it's just a magic number that makes no sense to plugin developers on their own. The range shouldn't just be 0 to Infinity where the scale is some non-linear function. I'd rather say "okay, the probability of this block going boom is 75%" or whatever, because, from what I can see, that's really what this resistance does, it just has a weird non-linear scale.

If you can make the block resistance a sane number and a sane scale, that is, 0 to 1 where it's effectively a percentage. I'd be happier. You can probably work what the percentages represent by looking at how Minecraft deals with resistance and work out probabilities from there. That way, it makes sense to any plugin developer that uses this (they just need to say "I want this chance") and any third party implementor, and gives a common implementation base, not just some weird numbers that base Minecraft uses.

Pulverizer:
> > I would point out that the arbitrary numbers exist because blast resistance in vanilla Minecraft is just that.
> 
> And I would point out that one of the points of our API is to, effectively, provide a layer between Minecraft and plugins that provides predictable, known behaviour. We also consider third party implementations who might not want to use Mojang's blast resistance mechanic.
> 
> Take @gabizou's comment above. Some of it is geared at the javadocs, but there is also this:
> 
> > Is it possible to return some bounds?
> > What are the numbers meaning? Do I return 0 to say "don't break me" or do I use 1f to say "I'm impenetrable? Or is it a resistance factor based on 0-10?
> 
> which I think is the sticking point and why I'm reticent to go with this. [Indeed, I also reference this here](https://github.com/SpongePowered/SpongeAPI/pull/2130#discussion_r447060681) and while I simply say "what can you do", linking to the Minecraft wiki, which can change over time, isn't the way to go. Our docs should really be self contained.

I agree but am not sure what to do about it.

> If this numbers were 0 to 1, that's something that makes sense - but it's not. I understand how Minecraft does it (now I've looked at it) but it's just a magic number that makes no sense to plugin developers on their own. The range shouldn't just be 0 to Infinity where the scale is some non-linear function. I'd rather say "okay, the probability of this block going boom is 75%" or whatever, because, from what I can see, that's really what this resistance does, it just has a weird non-linear scale.
>
> If you can make the block resistance a sane number and a sane scale, that is, 0 to 1 where it's effectively a percentage. I'd be happier. You can probably work what the percentages represent by looking at how Minecraft deals with resistance and work out probabilities from there. That way, it makes sense to any plugin developer that uses this (they just need to say "I want this chance") and any third party implementor, and gives a common implementation base, not just some weird numbers that base Minecraft uses.


There is no randomness to whether a block is broken or not where Blast Resistance is involved.
The explosion calculation gives each raytrace a `strength`, e.g. 20.
For each step it moves along the raytrace it deducts a hardcoded number.
When it finds a block it fetches the `resistance` of the block and deducts it from the `strength` that is remaining.
If there is still `strength` remaining then the block will be broken.
Once there is no `strength` remaining the raytrace will stop.

dualspiral:
Herein lies the issue - this reeks of implementation details. Using our API:

* How do you know what blast rays exist?
* How do you know what the strength of the blast ray is?
* How are you meant to know that this block resistance does this?
* Why is it even assumed that ray tracing is done for an explosion?
* Will the block stop the ray if given a specific resistance or not (you might want to know this?)

There is a **lot** more to just exposing the block resistance than a magic number that requires innate knowledge of the system to understand correctly. We're not going to expose the strength of a ray, or even talk about the strength of a ray. So the block resistance number we talk about here inherently will _never_ make sense to a consumer on its own, or to a third party implementor that is completely writing their own server software (see, Lantern).

Some of us have had a discussion about this and we decided that we're going to close the PR. As much as we'd love to support you fully with the API, we honestly don't see how to do this in the API without leaking all this implementation detail - you're asking for modification of the implementation (rather than the concept) at a really granular level. I realise that you've put a fair amount of effort into this (some of it, at our behest!) but now knowing exactly what is going on, this really strays into something you'd use Forge for.

It may be worth looking at converting your plugin into a hybrid Sponge/Forge mod. You can still use the Sponge API as much as you like by depending on it with in a mod workspace, but you'll also get access to the base classes and implementation detail that we're unwilling to provide access to.

Thank you for your contribution regardless - I hope that some of these discussions we've had refining your idea have helped you see how we might tackle things too!

Pulverizer:
> Herein lies the issue - this reeks of implementation details. Using our API:
> 
> * How do you know what blast rays exist?
> * How do you know what the strength of the blast ray is?
> * How are you meant to know that this block resistance does this?
> * Why is it even assumed that ray tracing is done for an explosion?
> * Will the block stop the ray if given a specific resistance or not (you might want to know this?)
> 
> There is a **lot** more to just exposing the block resistance than a magic number that requires innate knowledge of the system to understand correctly. We're not going to expose the strength of a ray, or even talk about the strength of a ray. So the block resistance number we talk about here inherently will _never_ make sense to a consumer on its own, or to a third party implementor that is completely writing their own server software (see, Lantern).
> 
> Some of us have had a discussion about this and we decided that we're going to close the PR. As much as we'd love to support you fully with the API, we honestly don't see how to do this in the API without leaking all this implementation detail - you're asking for modification of the implementation (rather than the concept) at a really granular level. I realise that you've put a fair amount of effort into this (some of it, at our behest!) but now knowing exactly what is going on, this really strays into something you'd use Forge for.
> 
> It may be worth looking at converting your plugin into a hybrid Sponge/Forge mod. You can still use the Sponge API as much as you like by depending on it with in a mod workspace, but you'll also get access to the base classes and implementation detail that we're unwilling to provide access to.
> 
> Thank you for your contribution regardless - I hope that some of these discussions we've had refining your idea have helped you see how we might tackle things too!

I understand your position but with every one of these repeat experiences, I feel more and more like Sponge is just overly hyped up.

I will offer one suggestion, and it's really about Mixins, make it possible to overwrite mixin methods that are annotated with `@Final`. Give it a big huge warning label for all I care, but I wouldn't have even bothered with this PR if I could have supplied an `Unstable` Mixin with my plugin.

Zidane:
@Pulverizer You clearly do not understand any of this.

You are clearly angry because we denied an API request that very much is tied to what Mojang does. SpongeAPI is an API for implementations that want to re-create Minecraft or Minecraft itself. Because it is, that means that the API CANNOT simply do "what Mojang does". You were told this above, that this is very much an implementation detail. I am sorry if you're offended.

Instead of being professional, you commence with a snide comment. If you actually understood the technology you're using, you would already know that you CAN do per-platform mixins (Forge/Vanilla), "`@Final` Mixins" as you put it makes not a bit of sense, and ultimately would be done with this already.

Don't blame us for issues with your code and your environment.

PR #2131, opened 2020-06-18
Clarify some stuff with Server players on a server

i509VCB:


src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
```diff
@@ -32,4 +33,6 @@
     @Override
     ServerWorld getWorld();
 
+    Server getServer();
```
Zidane: ServerWorld has a getServer I believe.

i509VCB: Yes that is true. This was mainly a thought from vanilla having a getServer method on it's own server player entity

PR #2133, opened 2020-06-27
Replace Service Manager with Provider

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2653)

@Zidane For discussion. Very proof of concept-y, but this was what I was thinking. Plugin recieves event. Plugin chooses whether to supply to event. Event only sent to a plugin if:

* It is named as the provider for the given service in the config (see Common)
* There is no named provider and a registration hasn't been successfully made yet

You'll see in Common that I send the event to one plugin at a time. I've not added the service provider to `Game` or `Sponge` yet, that's part of the discussion!

Vanilla/Sponge services have low priority and will always be created last (or as a fallback to a named provider that failed to load)


src/main/java/org/spongepowered/api/event/service/ProvideServiceEvent.java
```diff
@@ -0,0 +1,34 @@
...
+ * service. This may happen if the server is configured to select a particular
+ * service.</p>
+ *
+ * @param <T> The service to provide.
+ */
+public interface ProvideServiceEvent<T> extends GenericEvent<T> {
+
+    /**
+     * Provides a suggestion for the given service. <strong>This may only be
+     * called once for a given service.</strong>
```
ST-DDT: Maybe append `per plugin` otherwise the reader expects a reason or error here.

src/main/java/org/spongepowered/api/service/ServiceProvider.java
```diff
@@ -0,0 +1,86 @@
...
+public interface ServiceProvider {
+
+    /**
+     * Provides the service represented by the supplied {@link Class}.
+     *
+     * @param serviceClass The class
+     * @param <T> The type of service
+     * @return The service, if one exists
+     */
+    <T> Optional<T> provide(Class<T> serviceClass);
```
ST-DDT, line 49: Please add a `provideUnchecked`, like before.

src/main/java/org/spongepowered/api/service/ServiceProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Provides the {@link ServiceRegistration} for the supplied {@link Class}.
+     *
+     * @param serviceClass The class
+     * @param <T> The type of service
+     * @return The registration, if one exists
+     */
+    <T> Optional<ServiceRegistration<T>> registration(Class<T> serviceClass);
```
ST-DDT: I would call it `getRegistration`, not sure which prupose this method has though.

ST-DDT (commented):
Interesting concept.
But how do I get access to the ServiceProvider? AFAICT there is no longer a static way to get that instance.

src/main/java/org/spongepowered/api/service/ServiceProvider.java
dualspiral: No. 

As this is a PoC I've not fully explained what's going on here but this is not intended for third party service interfaces. The only reason I've added this is to enable third party implementations to add extra first-party service interfaces.

dualspiral:
> But how do I get access to the ServiceProvider? AFAICT there is no longer a static way to get that instance.

You need to remember that this is a proof of concept and is a draft - something that's come up after discussion with Zidane. Because init is still up in the air and it's not clear where the access will go, I've not added it, and that's concious. It _will_ be added.

This concept may not go futher - if it does, I'll expend the effort then.

src/main/java/org/spongepowered/api/service/ServiceProvider.java
dualspiral: Yeah, name is in the air. A example is that a plugin might want to know who supplies a service, so they then know they can use an expanded API from said plugin that they are aware of.

src/main/java/org/spongepowered/api/Sponge.java
```diff
@@ -254,4 +240,14 @@ public static Scheduler getAsyncScheduler() {
...
         return check(commandManager);
     }
+
+    /**
+     * Gets the {@link ServiceProvider} for providing services.
+     *
+     * @return The service provider.
+     */
+    public static ServiceProvider getServiceProvider() {
+        return getGame().getServiceProvider();
```
Zidane, line 247: this.

dualspiral: `getGame()` is static.

Zidane: Sigh, I am up too early then.

PR #2135, opened 2020-06-28
[API-8] Resource Management

mattmess1221:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/3108)

It's been nearly 3 years since I opened #1706, so here's starting fresh on a new branch with mostly new code.

The only thing I don't like is `PackDiscoverer`s. That part isn't critical to resources, so we can remove it and do it for another time if we're short on time.

I'm also not sure if we need a separate event for registering reload listeners. We may be able to get away with having them be registered during ServerAboutToStartEvent. Input is appreciated.


src/main/java/org/spongepowered/api/event/resource/PackEvent.java
```diff
@@ -0,0 +1,71 @@
...
+         * <p>Returns a {@link MutablePackList} to expose methods for adding listeners.</p>
+         * @return The pack list
+         */
+        @Override
+        MutablePackList getPackList();
+
+        /**
+         * Register event for client resources.
+         */
+        interface Client extends RegisterPackDiscoverer {
```
Zidane: Maybe just one event that is generic for engine?

See `StartingEngineEvent`

src/main/java/org/spongepowered/api/event/resource/ResourceEvent.java
```diff
@@ -0,0 +1,74 @@
...
+         *
+         * @return The reloadable resource manager
+         */
+        @Override
+        ReloadableResourceManager getResourceManager();
+
+        /**
+         * Register event for client resources.
+         */
+        interface Client extends RegisterReloadListener {
```
Zidane: Same as the above, a generic engine event.

If lifecycle, put it there along with the above.

src/main/java/org/spongepowered/api/event/resource/PackEvent.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    interface RegisterPackDiscoverer extends PackEvent {
+
+        /**
+         * {@inheritDoc}
+         * <p>Returns a {@link MutablePackList} to expose methods for adding listeners.</p>
+         * @return The pack list
+         */
+        @Override
+        MutablePackList getPackList();
```
Zidane: Do not expose the pack list, instead add a register method like the registry events.

src/main/java/org/spongepowered/api/event/resource/PackEvent.java
```diff
@@ -0,0 +1,71 @@
...
+ * Base {@link Pack} event.
+ */
+public interface PackEvent extends Event {
+
+    /**
+     * Gets the relevant {@link PackList} for this event.
+     *
+     * @return The pack list
+     */
+    PackList getPackList();
```
Zidane: See below.

src/main/java/org/spongepowered/api/event/resource/ResourceEvent.java
```diff
@@ -0,0 +1,74 @@
...
+ * Base event for {@link Resource}s.
+ */
+public interface ResourceEvent extends Event {
+
+    /**
+     * Gets the relevant {@link ResourceManager} for this event.
+     *
+     * @return The resource manager
+     */
+    ResourceManager getResourceManager();
```
Zidane: Do not expose the manager, expose register methods.

src/main/java/org/spongepowered/api/event/resource/ResourceEvent.java
mattmess1221: I wouldn't consider it a lifecycle event. It's just a specific event that does a specific task once at a specific time during init.

Zidane:
Closing in favor of https://github.com/SpongePowered/SpongeAPI/tree/feature/resources

Thank you for all your work on this Killjoy.

PR #2137, opened 2020-06-30
[API-8] Improve Tristate

ItsDoot:
Adds:
* `Tristate#not`
* `Tristate#fromNullableBoolean`
* `Tristate#asNullableBoolean`

### DIFF
--- a/src/main/java/org/spongepowered/api/util/Tristate.java
+++ b/src/main/java/org/spongepowered/api/util/Tristate.java
@@ -33,7 +33,7 @@ public enum Tristate {
     TRUE(true) {
         @Override
         public Tristate and(Tristate other) {
-            return other == TRUE || other == UNDEFINED ? TRUE : FALSE;
+            return other == FALSE ? FALSE : TRUE;
         }
 
         @Override
--- a/src/main/java/org/spongepowered/api/util/Tristate.java
+++ b/src/main/java/org/spongepowered/api/util/Tristate.java
@@ -147,8 +147,8 @@ public enum Tristate {
      * where {@link Tristate#UNDEFINED} equates to null.
      *
      * <p>Note: To ensure that there is a one-to-one mapping of nullable
-     * booleans to and from {@link Tristate}, this method may return null
-     * rather than using an {@link java.util.Optional Optional}.</p>
+     * booleans to and from {@link Tristate}, this method will return null
+     * rather than use an {@link java.util.Optional Optional}.</p>
      *
      * @return The nullable boolean tristate representation
      */
--- a/src/main/java/org/spongepowered/api/util/Tristate.java
+++ b/src/main/java/org/spongepowered/api/util/Tristate.java
@@ -102,7 +102,7 @@ public enum Tristate {
      * @param val The nullable boolean value
      * @return The appropriate tristate
      */
-    public static Tristate fromNullable(@Nullable Boolean val) {
+    public static Tristate fromNullableBoolean(@Nullable Boolean val) {
         if (val == null) {
             return UNDEFINED;
         } else {
@@ -140,5 +140,20 @@ public enum Tristate {
      */
     public boolean asBoolean() {
         return this.val;
+    }
+
+    /**
+     * Returns the nullable boolean representation of this tristate,
+     * where {@link Tristate#UNDEFINED} equates to null.
+     *
+     * @return The nullable boolean tristate representation
+     */
+    @Nullable
+    public Boolean asNullableBoolean() {
+        if (this == Tristate.UNDEFINED) {
+            return null;
+        } else {
+            return this.val;
+        }
     }
 }
--- a/src/main/java/org/spongepowered/api/util/Tristate.java
+++ b/src/main/java/org/spongepowered/api/util/Tristate.java
@@ -24,6 +24,8 @@
  */
 package org.spongepowered.api.util;
 
+import org.checkerframework.checker.nullness.qual.Nullable;
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/util/Tristate.java
```diff
@@ -94,6 +126,13 @@ public static Tristate fromBoolean(boolean val) {
      */
     public abstract Tristate or(Tristate other);
 
+    /**
+     * NOTs this tristate.
```
Zidane: Maybe call it negate?

src/main/java/org/spongepowered/api/util/Tristate.java
```diff
@@ -78,6 +95,21 @@ public static Tristate fromBoolean(boolean val) {
...
     }
 
+    /**
+     * Returns the appropriate tristate for a given nullable boolean value,
+     * where null equates to {@link Tristate#UNDEFINED}.
+     *
+     * @param val The nullable boolean value
+     * @return The appropriate tristate
+     */
+    public static Tristate fromNullable(@Nullable Boolean val) {
```
dualspiral: I feel like this should be named `fromNullableBoolean`.

If we have this, should we also have an `asNullableBoolean` analogue?

dualspiral (commented):
Looks okay, I think a method needs renaming and one should be added for consistency.

src/main/java/org/spongepowered/api/util/Tristate.java
ItsDoot: See, I was thinking about doing `asNullableBoolean`, but what should it return? Returning an actual nullable boolean would make sense, but Sponge prefers Optionals... which to me would feel a little weird.

src/main/java/org/spongepowered/api/util/Tristate.java
ItsDoot: Edited

ST-DDT:
While we are at it we should also change:
https://github.com/SpongePowered/SpongeAPI/blob/d0426575e3c6197363550a072f2a7c80b5d384d1/src/main/java/org/spongepowered/api/util/Tristate.java#L33-L37

to

````java
return other == FALSE ? FALSE : TRUE;
````

This is shorter/easier to understand and also matches the `FALSE#or(...)` style.

File: src/main/java/org/spongepowered/api/util/Tristate.java (hunk 2 above)
Comment (dualspiral) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: This is one of those rare cases I'd have a nullable return type as a one-to-one mapping of `Boolean`<->`Tristate` - if you can get a `Tristate` from a `Boolean`, it stands to reason that you should be able to go the other way just as simply.

I'll let @Zidane call this one, but I think it should just be a `@Nullable Boolean` return type. There is an extra clue in the name of the method here, if you're not expecting the possibility of it being `null` if you grab the `Boolean` then that's really your lookout.

Comment (ST-DDT) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: I would prefer, not to have a `asNullableBoolean`, because this class is the replacement for the nullable/unset boolean.

As for the `fromX` methods, I would rather name them both `from` (or `fromBoolean`) with just different parameters. Let the compiler decide which one it should use.

**Comment by ItsDoot** [comment]
Agreed @ST-DDT

File: src/main/java/org/spongepowered/api/util/Tristate.java (hunk 1 above)
Comment (ST-DDT) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: Maybe also change the method name accordingly?

File: src/main/java/org/spongepowered/api/util/Tristate.java (hunk 2 above)
Comment (dualspiral) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: > because this class is the replacement for the nullable/unset boolean.

That doesn't change my argument at all. I'm saying the mapping should work both ways in case you then interact with some system that _isn't_ `Tristate` aware - plugins can (and often do) work with third party libraries.

File: src/main/java/org/spongepowered/api/util/Tristate.java (hunk 1 above)
Comment (ItsDoot) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: IMO `not` lines up better with `and`/`or` than `negate` does.

**Review by ST-DDT — APPROVED** [review]

File: src/main/java/org/spongepowered/api/util/Tristate.java (hunk 2 above)
Comment (ItsDoot) [reply src/main/java/org/spongepowered/api/util/Tristate.java]: Changed `fromNullable` to `fromNullableBoolean`, and added `asNullableBoolean`.

> As for the fromX methods, I would rather name them both from (or fromBoolean) with just different parameters. Let the compiler decide which one it should use.

@ST-DDT I'd rather users explicitly opt-in to using the nullable boolean conversions. Then it's clear when someone else is reading your code, for example.

File: src/main/java/org/spongepowered/api/util/Tristate.java [hunk 3]
```diff
@@ -102,4 +141,23 @@ public static Tristate fromBoolean(boolean val) {
     public boolean asBoolean() {
         return this.val;
     }
+
+    /**
+     * Returns the nullable boolean representation of this tristate,
+     * where {@link Tristate#UNDEFINED} equates to null.
+     *
+     * <p>Note: To ensure that there is a one-to-one mapping of nullable
+     * booleans to and from {@link Tristate}, this method may return null
```
Zidane: may return? I see no Optional return so state it will.

ItsDoot: Fixed

PR #2141, opened 2020-07-01
[API-8] MessageChannel/MessageReceiver improvements

ItsDoot:
Two things this PR does:

> Relax the `MessageChannel#getMembers()` return type from `Collection<MessageReceiver>` to `Collection<? extends MessageReceiver>`

This makes it easier to create `MessageChannel`s that just delegate `getMembers()` to another `Collection<SomeTypeThatImplementsMessageReceiver>`. ~~It also functionally prevents consumers from calling `Collection#add` or other mutating operations.~~

> Pull `MessageReceiver#setMessageChannel(MessageChannel)` into a separete subinterface `MessageReceiver.Mutable`

Not all `MessageReceiver`s need or should be able to have their `MessageChannel` changed.

TODO:
- [X] Figure out what other API interfaces besides `Player` should be `MessageReceiver.Mutable`.
    - I think we're good for now, and can make other receivers mutable as needed.


dualspiral:
Completely agree with everything you propose. Just one point:

> It also functionally prevents consumers from calling Collection#add or other mutating operations.

We shouldn't rely on that because type erasure and raw types are a thing - the returned collection should be immutable or a copy to ensure that is the case (crucially, javadocced as such). As `MutableMessageChannel` has methods that already mutate the list, I don't really think this is a problem.

ItsDoot:
@dualspiral For sure. I tacked it on in case others wanted more than 1 reason. Will leave that part out of any javadocs of course. 👍

dualspiral:
Actually, I'd go the other way and make it explicit. I personally don't think that the collection returned should be able affect the channel in any situation - so I think it should either be returning a copy or an immutable view. I can see the situation where a dev gets the collection and alters it, not realising how it's affecting the channel.

Could also just note that the returned collection backs the message channel and that any changes are reflected - but that's at odds with the extends clause!

ItsDoot:
Alright, I've updated `#getMembers()` javadocs to reflect this. @dualspiral

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -135,4 +127,18 @@ default Locale getLocale() {
...
     }
 
+    /**
+     * Represents a {@link MessageReceiver} that can have their
+     * {@link MessageChannel} changed.
+     */
+    interface Mutable extends MessageReceiver {
+
+        /**
+         * Sets the message channel that messages sent by this source should be sent
```
gabizou, line 137: So what does this mean exactly? Is this effectively removing the need for events on a channel to have a player's message forwarded to other channels?

src/main/java/org/spongepowered/api/text/channel/MessageChannel.java
```diff
@@ -254,9 +254,12 @@ default void send(@Nullable Object sender, Text original, Supplier<? extends Cha
     /**
      * Gets a collection of all members in this channel.
      *
-     * @return A collection of all members of this channel
+     * <p>Implementors of this interface should always return a copy or
+     * immutable view of the underlying container.</p>
```
gabizou, line 258: Copy or immutable? The return tag says immutable, but a copy could be mutable.

src/main/java/org/spongepowered/api/text/channel/MessageReceiver.java
```diff
@@ -135,4 +127,18 @@ default Locale getLocale() {
         return Locales.DEFAULT;
     }
 
+    /**
+     * Represents a {@link MessageReceiver} that can have their
+     * {@link MessageChannel} changed.
```
gabizou, line 132: So could you clarify here the uniqueness of what this means? That a Mutable variation can still only naturally send to one single channel (which can be comprised of multiple channels), and some will only send to a specific channel?

Is it still possible that there's events for the message sent to "forward" messages to other channels?

ItsDoot:
No longer valid as of #2180.

The replacement now has the `? extends` for getMembers(),
though a Mutable `Audience` is now missing.

PR #2142, opened 2020-07-02
[API-8] Allow resolving of CatalogKeys with custom default namespaces.

ItsDoot:
Allowing API consumers to resolve `CatalogKey`s with custom default namespaces will make it easier to handle user input of custom `CatalogType`s.

For example, my "Avatar the Last Airbender in Minecraft" plugin defines abilities as `CatalogType`s (which allows other plugins to register their own custom abilities), and it would be handy to by default resolve non-namespaced keys to my plugin's namespace with a single `CatalogKey.resolve` call.

I've also taken the opportunity to add missing javadocs to the interface.

--- 

To end off, my final concern is whether we should be defaulting the `namespace` in `CatalogKey.Builder`. IMO, we should prefer for API consumers to specifically set the namespace, or use builtin factories like `CatalogKey#minecraft` or `CatalogKey#sponge`.

I wager that plugin devs will rarely want `minecraft`-namespaced keys, especially since most of the time they'll be calling `CatalogKey` builders/factories for their own custom `CatalogType`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/ResourceKey.java
+++ b/src/main/java/org/spongepowered/api/ResourceKey.java
@@ -95,38 +95,64 @@ public interface ResourceKey extends Key {
     }
 
     /**
-     * Creates a catalog key.
+     * Creates a catalog key from the provided namespace and value.
      *
      * @param namespace The namespace
      * @param value The value
      * @return A new catalog key
      */
     static ResourceKey of(final String namespace, final String value) {
-        return Sponge.getRegistry().getBuilderRegistry().provideBuilder(Builder.class).namespace(namespace).value(value).build();
+        return builder().namespace(namespace).value(value).build();
     }
 
     /**
-     * Creates a catalog key
+     * Creates a catalog key from the provided {@link PluginContainer}'s
+     * identifier and the provided value.
      *
      * @param container The container
      * @param value The value
      * @return A new catalog key
      */
     static ResourceKey of(final PluginContainer container, final String value) {
-        return Sponge.getRegistry().getBuilderRegistry().provideBuilder(Builder.class).namespace(container).value(value).build();
+        return builder().namespace(container).value(value).build();
     }
 
     /**
      * Resolves a catalog key from a string.
      *
-     * <p>If no namespace is found in {@code string} then
+     * <p>If no namespace is found in {@code value} then
      * {@link #MINECRAFT_NAMESPACE} will be the namespace.</p>
      *
      * @param value The value
      * @return A new catalog key
      */
     static ResourceKey resolve(final String value) {
-        return Sponge.getRegistry().getBuilderRegistry().provideBuilder(Builder.class).value(value).build();
+        return resolve(value, MINECRAFT_NAMESPACE);
+    }
+
+    /**
+     * Resolves a catalog key from a string, using the provided default
+     * namespace if no namespace was found in {@code value}.
+     *
+     * @param value The value
+     * @param defaultNamespace The default namespace
+     * @return A new catalog key
+     */
+    static ResourceKey resolve(final String value, final String defaultNamespace) {
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).resolve(value, defaultNamespace);
+    }
+
+    /**
+     * Resolves a catalog key from a string, using the provided
+     * {@link PluginContainer}'s identifier as the default namespace if no
+     * namespace was found in {@code value}.
+     *
+     * @param value The value
+     * @param defaultNamespace The default namespace
+     * @return A new catalog key
+     */
+    static ResourceKey resolve(final String value, final PluginContainer defaultNamespace) {
+        return resolve(value, defaultNamespace.getMetadata().getId());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -119,14 +120,39 @@ static CatalogKey of(final PluginContainer container, final String value) {
...
-     * <p>If no namespace is found in {@code string} then
+     * <p>If no namespace is found in {@code value} then
      * {@link #MINECRAFT_NAMESPACE} will be the namespace.</p>
      *
      * @param value The value
      * @return A new catalog key
      */
     static CatalogKey resolve(final String value) {
-        return Sponge.getRegistry().getBuilderRegistry().provideBuilder(Builder.class).value(value).build();
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).resolve(value, MINECRAFT_NAMESPACE);
```
ST-DDT:
```suggestion
        return resolve(value, MINECRAFT_NAMESPACE);
```
Reuse your existing logic for fallback handling or delegate to the factory (new method).

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -119,14 +120,39 @@ static CatalogKey of(final PluginContainer container, final String value) {
...
+     * Resolves a catalog key from a string, using the provided
+     * {@link PluginContainer}'s identifier as the default namespace if no
+     * namespace was found in {@code value}.
+     *
+     * @param value The value
+     * @param defaultNamespace The default namespace
+     * @return A new catalog key
+     */
+    static CatalogKey resolve(final String value, final PluginContainer defaultNamespace) {
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).resolve(value, defaultNamespace);
```
ST-DDT:
```suggestion
        return resolve(value, defaultNamespace.getMetadata().getId());
```
IMO we shouldn't hide the logic in the factory as it is clearly defined in the javadocs.

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -157,14 +183,69 @@ static CatalogKey resolve(final String value) {
...
+        /**
+         * Resolves a catalog key from a string, using the provided
+         * {@link PluginContainer}'s identifier as the default namespace if no
+         * namespace was found in {@code value}.
+         *
+         * @param value The value
+         * @param defaultNamespace The default namespace
+         * @return A new catalog key
+         */
+        CatalogKey resolve(final String value, final PluginContainer defaultNamespace);
```
ST-DDT: IMO this should be default implemented or removed, as the javadocs only allow a single way of implementing it.

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -157,14 +183,69 @@ static CatalogKey resolve(final String value) {
...
+
+        /**
+         * Resolves a catalog key from a string, using the provided default
+         * namespace if no namespace was found in {@code value}.
+         *
+         * @param value The value
+         * @param defaultNamespace The default namespace
+         * @return A new catalog key
+         */
+        CatalogKey resolve(final String value, final String defaultNamespace);
```
ST-DDT: Also add a variant without a `defaultNamespace`. Especially if you implement a custom "default namespace".

ST-DDT (commented):
I agree with your opinion, that defaulting to the minecraft namespace is only useful for the gamers, but the plugin developers usually "know better" which namespace it should default to.

Having a default namespace for a `CatalogType` seems like a good addition.

Having the `PluginContainer` parameter seems like a useful addition to the catalog command parameter.
https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java#L171
Maybe we should mention the default prefix/namespace in the parameter builder as well then.

src/main/java/org/spongepowered/api/CatalogKey.java
ItsDoot: I don't think this is necessary. What I mean by custom default namespace, is whatever the plugin developer defines when calling this method.

ItsDoot:
@ST-DDT Adjusted based on your feedback.

ST-DDT (approved):

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -119,14 +120,39 @@ static CatalogKey of(final PluginContainer container, final String value) {
...
+     * @param defaultNamespace The default namespace
+     * @return A new catalog key
+     */
+    static CatalogKey resolve(final String value, final String defaultNamespace) {
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).resolve(value, defaultNamespace);
+    }
+
+    /**
+     * Resolves a catalog key from a string, using the provided
+     * {@link PluginContainer}'s identifier as the default namespace if no
```
Zidane: `{@Link PluginContainer plugin}'s`

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -119,14 +120,39 @@ static CatalogKey of(final PluginContainer container, final String value) {
...
+     * namespace was found in {@code value}.
+     *
+     * @param value The value
+     * @param defaultNamespace The default namespace
+     * @return A new catalog key
+     */
+    static CatalogKey resolve(final String value, final PluginContainer defaultNamespace) {
```
Zidane: `final PluginContainer plugin`

src/main/java/org/spongepowered/api/CatalogKey.java
```diff
@@ -119,14 +120,39 @@ static CatalogKey of(final PluginContainer container, final String value) {
...
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).resolve(value, defaultNamespace);
+    }
+
+    /**
+     * Resolves a catalog key from a string, using the provided
+     * {@link PluginContainer}'s identifier as the default namespace if no
+     * namespace was found in {@code value}.
+     *
+     * @param value The value
+     * @param defaultNamespace The default namespace
```
Zidane: `@param plugin `

ItsDoot:
Closed in favor of https://github.com/SpongePowered/SpongeAPI/pull/2268 and https://github.com/SpongePowered/SpongeCommon/pull/3217

PR #2144, opened 2020-07-02
[API-8] Add missing Living value accessors

ItsDoot:
Adds value accessors for the following keys to `Living`:
- `ABSORPTION`
- `ACTIVE_ITEM`
- `EYE_HEIGHT`
- `EYE_POSITION`
- `FALL_DISTANCE`
- `MAX_AIR`
- `REMAINING_AIR`
- `STUCK_ARROWS`
- `WALKING_SPEED`

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/Living.java
+++ b/src/main/java/org/spongepowered/api/entity/living/Living.java
@@ -30,6 +30,7 @@ import org.spongepowered.api.effect.potion.PotionEffect;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.api.entity.attribute.Attribute;
 import org.spongepowered.api.entity.attribute.AttributeHolder;
+import org.spongepowered.api.item.inventory.ItemStackSnapshot;
 import org.spongepowered.api.projectile.source.EntityProjectileSource;
 import org.spongepowered.api.scoreboard.TeamMember;
 import org.spongepowered.math.imaginary.Quaterniond;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -78,6 +78,22 @@
         return this.requireValue(Keys.HEAD_ROTATION).asMutable();
     }
 
+    /**
+     * {@link Keys#EYE_HEIGHT}
+     * @return The height of the eyes
+     */
+    default Value.Immutable<Double> eyeHeight() {
```
Zidane: Only do Value.

I now remember the discussion ages ago. Getting a Value doesn't mean setting the `value` will actually set it back on the data holder. You have to offer it back. A DataProvider in common can reject it.

Nothing should stop your Value from not being written to how you like though.

@gabizou

ItsDoot: Fixed

src/main/java/org/spongepowered/api/entity/living/Living.java
```diff
@@ -47,6 +48,22 @@
...
+     */
+    default BoundedValue.Mutable<Double> absorption() {
+        return this.requireValue(Keys.ABSORPTION).asMutable();
+    }
+
+    /**
+     * {@link Keys#ACTIVE_ITEM}
+     * @return The active item, such as food being eaten
+     */
+    default Value.Mutable<ItemStackSnapshot> activeItem() {
```
Zidane, line 62: @gabizou 

Can you give some guidance on how the Values in these interfaces should be returned to consumers?

Like, should they always be Value.Mutable? Just Value? Does it depend on what it is returning and what you can do with it in the game?

gabizou: I am preferring they're always `Mutable` from entities. It does make sense that they are reflective of their `Key` variables, but at the end of the day, the developer is likely to want to mutate the value and maybe offer it back, or present it to another entity after mutation.

PR #2145, opened 2020-07-02
[API-8] Add Location#getBlockType

ItsDoot:


src/main/java/org/spongepowered/api/world/Location.java
```diff
@@ -290,6 +290,15 @@
      */
     BlockState getBlock();
 
+    /**
+     * Gets the {@link BlockType} for this position.
```
Zidane: ** for this location

ItsDoot: Fixed

PR #2146, opened 2020-07-03
[API-8] Add ScheduledTask#isCancelled()

ItsDoot:
My "Avatar the Last Airbender in Minecraft" plugin manages a Task per ability execution, so that abilities can be executed over an indefinite amount of time. I need to be able to check if tasks are cancelled, so that I can correctly cleanup the object that wraps them.


ST-DDT (approved):
How about also adding an optional callback on the task that will be called if the task gets cancelled?
That way you don't have to check the task periodically.

PR #2149, opened 2020-07-04
[API-8] Add AbsoluteSortPosition annotations

ItsDoot:


src/main/java/org/spongepowered/api/event/entity/ConstructEntityEvent.java
```diff
@@ -36,28 +37,31 @@
  */
 public interface ConstructEntityEvent extends Event {
 
+    /**
+     * Gets the {@link EntityType} of the target {@link Entity} that is going to be
+     * constructed.
+     *
+     * @return The target entity type
+     */
+    @AbsoluteSortPosition(1)
```
Zidane, line 46: Arrange these in the order of the sort as well please.

Zidane:
Gonna close this for now as AbsoluteSortPosition is quite borked in eventimplgen. Need to spend the time to fix it.

PR #2150, opened 2020-07-04
[API-8] Reroute BlockSnapshot#empty into a factory, and add misc missing javadocs

ItsDoot:
The concept of creating an empty value of a given object, belongs in a factory, and not a builder.

### DIFF
--- a/src/main/java/org/spongepowered/api/block/BlockSnapshot.java
+++ b/src/main/java/org/spongepowered/api/block/BlockSnapshot.java
@@ -51,11 +51,11 @@ public interface BlockSnapshot extends LocatableSnapshot<BlockSnapshot> {
     /**
      * Represents a {@link BlockSnapshot} with the default state of
      * {@link BlockTypes#AIR} and a {@link ServerLocation} that cannot be determined.
+     *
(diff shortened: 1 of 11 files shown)


src/main/java/org/spongepowered/api/block/BlockSnapshot.java
```diff
@@ -201,7 +204,19 @@ static Builder builder() {
...
+    }
 
+    /**
+     * Represents a factory to create {@link BlockSnapshot}s.
+     */
+    interface Factory {
+
+        /**
+         * Represents a {@link BlockSnapshot} with the default state of
+         * {@link BlockTypes#AIR} and a {@link ServerLocation} that cannot be determined.
```
Zidane: Not sure "cannot be determined" is the appropriate phrasing.

Maybe state that the location is up to the implementation as to where "default" snapshots occur at.

As an example, In the mcp implementation, it would be 0, 0, 0 but Lantern could put it at Integer.MIN_VALUE.

ItsDoot: Fixed

PR #2151, opened 2020-07-04
[API-8] Expose PaginationCalculator

### LINKED ISSUE #2138: Expose Pagination API/PaginationCalculator
It would be nice if the `PaginationCalculator` was exposed by the API, so we could use it to make centered texts, add padding to texts and etc. Since the class is private in SpongeCommon, we can't even add it to the classpath to be able to use the class, the only way being copying the whole class.

ItsDoot:
Closes #2138.

### DIFF
--- a/src/main/java/org/spongepowered/api/service/pagination/PaginationCalculator.java
+++ b/src/main/java/org/spongepowered/api/service/pagination/PaginationCalculator.java
@@ -1,0 +1,51 @@
+package org.spongepowered.api.service.pagination;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.text.Text;
+
+/**
+ * Handles calculation of text widths, centering text, adding padding, adding
+ * spacing, and more.
+ */
+public interface PaginationCalculator {
+
+    /**
+     * Gets the number of lines the specified text flows into.
+     *
+     * @param text The text to calculate the number of lines for
+     * @return The number of lines that this text flows into
+     */
+    int getLines(Text text);
+
+    /**
+     * Calculates the width of a given text as the number of character
+     * pixels/columns the line takes up.
+     *
+     * @param text The text to get the width of
+     * @return The amount of character pixels/columns the text takes up
+     */
+    int getWidth(Text text);
+
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/service/pagination/PaginationCalculator.java
```diff
@@ -0,0 +1,51 @@
+package org.spongepowered.api.service.pagination;
```
Zidane, line 1: Forgot license headers.

Zidane:
Also, this could target api-7 as well.

simon816:
I think it's worth noting in the documentation that this assumes the vanilla font. In ChatUILib I have [`FontData`](https://github.com/simon816/ChatUI/blob/ba4fc94457d6dc741e6b7bafe0605096f7a6e5af/ChatUILib/src/main/java/com/simon816/chatui/util/FontData.java) which reads a .png to determine character sizes, allowing non-vanilla ascii.png measurements.
For unicode characters there's `glyph_sizes.bin`, though I'm not sure if this is modifiable in a datapack.

simon816:
Also, to centre text means knowing the width of the chat box. SpongeCommon [assumes the default](https://github.com/SpongePowered/SpongeCommon/blob/7ea7c5d8e69f84fa9b1c1c3eddb90d4fff5d4dd7/src/main/java/org/spongepowered/common/service/pagination/PaginationCalculator.java#L56) (320), but it is customizable (see https://ore.spongepowered.org/simon816/Chat-UI-Library#player-settings)

zml2008:
I think it's entirely fine that the Sponge Common implementation only supports vanilla content. Perhaps it could be expanded to be able to load data from resource packs and automatically configure based on the server pack, but it *is* a service and can be replaced by other more customizable implementations.

Another theoretical implementation could integrate with a client mod that passed the necessary sizing and resource pack information to the server,but that would be imo out of scope for sponge itself. However, to make this more friendly to such an implementation, the `getCalculator` method should probably take a `Player` parameter?

dualspiral:
This has sat around for some time - what do we want to do with this?

ImMorpheus:

___

PR #2152, opened 2020-07-04
 Add a way to get all unique and virtual accounts from the economy service

### LINKED ISSUE #1398: Can't get list of accounts from EconomyService

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/1398

### DIFF
--- a/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
+++ b/src/main/java/org/spongepowered/api/service/economy/EconomyService.java
@@ -24,7 +24,6 @@
  */
 package org.spongepowered.api.service.economy;
 
-import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.service.context.ContextualService;
 import org.spongepowered.api.service.economy.account.Account;
 import org.spongepowered.api.service.economy.account.AccountDeletionResultType;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/service/economy/EconomyService.java
```diff
@@ -108,6 +109,20 @@
      */
     Optional<Account> getOrCreateAccount(String identifier);
 
+    /**
+     * Returns a {@link Stream} of all stored {@link UniqueAccount} objects.
+     *
+     * @return A {@link Stream} of all stored {@link UniqueAccount} objects.
+     */
+    Stream<UniqueAccount> uniqueAccounts();
```
Zidane: Add Stream and Collection getters.

I like Streams but we should not shoehorn people into doing them as a default. I made the same change for the CatalogRegistry.

ImMorpheus: Done

Zidane: Maybe also use the same phrasing in the CatalogRegistry that I've seen in other libraries?

https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/registry/CatalogRegistry.java#L96

Zidane: @ImMorpheus I meant the method names.

ImMorpheus: Oops, done

PR #2153, opened 2020-07-04
Improve JoinData

### LINKED ISSUE #1845: Improve PlayerJoinData to support extra data.
History:

https://github.com/SpongePowered/SpongeCommon/issues/1282
https://github.com/SpongePowered/SpongeCommon/pull/1710

PlayerJoinData is inconsistent and impossible to fix correctly without API breaks due to badly named & semantics in API7.

It is planned, that the following data is going to be added to PlayerJoinData.

hasplayedbefore, firstjoined, lastjoined, lastplayed

This will clear up the ambiguity in the current data, as well as giving plugin developers the information they need.

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/1845

Offline players:
```
A joined  -> first joined
A left
A joined
A left
A joined  -> last joined
A left    -> last played
```

Online players:
```
A joined  -> first joined
A left
A joined
A left    -> last played
A joined  -> last joined
```

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -788,9 +788,9 @@ public final class Keys {
     public static final Supplier<Key<BoundedValue<Integer>>> FIRE_TICKS = Sponge.getRegistry().getCatalogRegistry().provideSupplier(Key.class, "FIRE_TICKS");
 
     /**
-     * The time a {@link Player} first played on the Server.
+     * The time a {@link Player} first joined on the Server.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -788,9 +788,9 @@
     public static final Supplier<Key<BoundedValue<Integer>>> FIRE_TICKS = Sponge.getRegistry().getCatalogRegistry().provideSupplier(Key.class, "FIRE_TICKS");
 
     /**
-     * The time a {@link Player} first played on the Server.
+     * The time a {@link Player} first joined on the Server.
```
Zidane: {@link Player} -> {@link ServerPlayer player}

ImMorpheus: Actually, this should be a User

Issue #1853, opened 2018-06-26
Marker for classes that should not be referenced directly


ST-DDT:
Based on https://github.com/SpongePowered/SpongeCommon/issues/1956.

Currently the information about stale/leaked references is present only in the [Docs: Storing References](https://docs.spongepowered.org/stable/en/plugin/practices/bad.html#storing-references). It would be nice if we could add some kind of marker that the user will see during development time that references should not be stored.

### Suggestions

1. **Annotation**
  Has to be set only on the root/base class and can be inherited and will be visible in javadocs.
2. **JavaDocs**
  Add a clause to every affected class and all of its subclasses that references should not be stored.
3. **(Marker-) Interface**
  Add a marker interface that shows. Do not cache these references.  Unfortunately you won't see the marker interface unless you explicitly search for it. (Maybe with a method: `isValid()`)

What do you think? Which would work best with tooling/IDE-support?

XakepSDK:
Annotation + support for that annotation in javac and/or support in idea plugin.
https://github.com/minecraft-dev/MinecraftDev

ryantheleach:
Sounds good to me, maybe start this by creating a summary of what classes should be annotated, and people can bikeshed about the specifics later?

ImMorpheus:
PR merged

PR #2154, opened 2020-07-04
Add marker for classes that should not be stored

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/1853

### DIFF
--- a/src/main/java/org/spongepowered/api/util/annotation/DoNotStore.java
+++ b/src/main/java/org/spongepowered/api/util/annotation/DoNotStore.java
@@ -24,17 +24,27 @@
  */
 package org.spongepowered.api.util.annotation;
 
+import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Target;
 
 /**
- * The instance of the class to which this annotation is applied cannot be stored.
+ * <p>In particular, you should not:</p>
  *
- * The annotation is applied to classes to indicate that their instances should not
- * be stored in a field or top-level variable.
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/api/util/annotation/DoNotStore.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * The instance of the class to which this annotation is applied cannot be stored.
+ *
+ * The annotation is applied to classes to indicate that their instances should not
+ * be stored in a field or top-level variable.
+ * The annotation is inherited.
+ */
+@Inherited
+@Target(ElementType.TYPE)
+public @interface DoNotStore {
```
dualspiral, line 50: Add `@Documented` as this is a documentation marker so should appear in the javadocs and is considered part of the public contract.

I'd change the javadocs on this too to give developers a hint on what (not) to do, something like:

```
Objects of types that are annotated with this method <strong>must not</strong> be stored as they may be invalidated at any time.

<p>In particular, you should not:</p>

<ul>
    <li>Store annotated objects on a field</li>
    <li>Cache annotated objects in any collection</li>
</ul>

<p>Doing so may result in memory leaks and/or result in you operating on a stale object, meaning your operations will fail, often silently.</p>

<p>Where objects have identifiers (usually a {@link UUID} or name), you should store the identifiers instead, and use these to get the required object from the API when needed.</p>
```

dualspiral (commented):
I think things like the `Cause`, `CommandCause` and other storing objects should also be annotated with this annotation as they store game objects.

Issue #1682, opened 2017-11-06
Request to support per-player view distance


phyne:
Referencing [this sponge thread](https://forums.spongepowered.org/t/would-it-be-possible-for-sponge-to-support-per-player-view-distance/21379/3)

Would it be possible to support per-player view distance? A method to send a certain amount of chunks based on a player's individual setting (either specified by them or the server) would be unbelievable. I stumbled upon paperspigot, and apparently they support this feature.

In any case, thanks for the great work you guys do on Sponge!

ryantheleach:
Would this prevent the chunks from even loading? or is this just a move to save on serialization / sending to the player?

parlough:
The related Paper patch can be found here - https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0036-Add-player-view-distance-API.patch

parlough:
@dags- I saw you were working on a plugin like this. If you wanted to make PRs to Sponge or at least provide some feedback on how you'd like to see this done in the API that would be fantastic!

ryantheleach:
https://github.com/SpongePowered/SpongeAPI/pull/1659 has merged per world view distances FYI.

dags-:
World and Player both have methods 'getViewDistance' which refer to different things:
- World's being the radius of chunks around a player which get sent to the client etc.
- Player's being the [clientside view-distance setting](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/entity/living/player/Player.java#L115) (clarified in [Common](https://github.com/SpongePowered/SpongeCommon/blob/c8e575216424089b5c5e75add57c0223560969a5/src/main/java/org/spongepowered/common/mixin/core/entity/player/MixinEntityPlayerMP.java#L428)).

To me, 'get/setChunkRadius' (or similar) makes sense for the server-side notion of chunks around a player (the server setting doesn't change the client's view-distance setting), but even in the server.properties file it's called 'view-distance'.
Alternatively the method currently in Player could just be renamed to something like 'getClientViewDistance' or 'getRenderDistance'?

I'm a little out of touch with the latest Sponge developments in all honestly so not sure what the preferred plan of attack would be.

Other than getter/setters, a reset method as added to the per-World PR would be a good idea.

PR #2155, opened 2020-07-04
Add view-distance setter

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/1682

In stable-7 `Player#getViewDistance` was used on the client to get the clientside view-distance setting. In api-8 the method was moved to `ServerPlayer` but the setter was missing.

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
@@ -123,6 +123,14 @@ public interface ServerPlayer extends Player, Subject {
     int getViewDistance();
 
     /**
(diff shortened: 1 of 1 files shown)


ImMorpheus:
Actually, `getViewDistance` still refers to the client view-distance. We're hijacking `CPacketClientSettings` to know the client setting, since the client is telling us its view-distance.

This PR is still valid but the methods need to be renamed.

concept (ignore the names)
```
getClientSideViewDistance -> this is the setting on the client. The client tells us what it is and we can't change that.
getServerSideViewDistance -> server-side view-distance, enforced (the client-side view-distance is ignored if this is set) 
setServerSideViewDistance -> setter for the above
```

gabizou (approved):

PR #2159, opened 2020-07-06
Command API tweaks

dualspiral:
Tweaks for commands, from as I implement and realise I did things wrong (or can add super simple things).

### DIFF
--- a/src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java
+++ b/src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java
@@ -31,6 +31,7 @@ import org.spongepowered.api.text.Text;
 import org.spongepowered.api.text.serializer.TextSerializer;
 import org.spongepowered.api.util.ResettableBuilder;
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java
```diff
@@ -168,7 +224,7 @@ public static TextBuilder textBuilder() {
          *               without such a prefix
          * @return This builder, for chaining.
          */
-        CatalogedTypeBuilder<T> prefix(String prefix);
+        CatalogedTypeBuilder<T> defaultNamespace(String prefix);
```
ST-DDT, line 227: I would go for `tryNamespace`/`tryDefaultNamespace` or something similar because you can have multiple. `defaultNamespace` sounds like you can only define one.

ST-DDT (commented):
Why did you change the following?

````diff
- Supplier<Iterable<String>> literalSupplier
+ Supplier<? extends Collection<String>> literalSupplier
````

PR #2163, opened 2020-07-10
Refactor EconomyTransactionEvent

### LINKED ISSUE #1501: Allow EconomytransactionEvent to be cancelable

ImMorpheus:
Part of the Economy API refactor.

Fix https://github.com/SpongePowered/SpongeAPI/issues/1501

`EconomyTransactionEvent` is now split between

Pre: is fired **BEFORE** the transaction has been processed (it can still fails and is cancellable).
Post: is fired **AFTER** the transaction has been processed. This is read-only and the transaction result can be used to reliably check the result of the transaction.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/economy/EconomyTransactionEvent.java
+++ b/src/main/java/org/spongepowered/api/event/economy/EconomyTransactionEvent.java
@@ -24,20 +24,71 @@
  */
 package org.spongepowered.api.event.economy;
 
+import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
+import org.spongepowered.api.service.economy.Currency;
 import org.spongepowered.api.service.economy.EconomyService;
+import org.spongepowered.api.service.economy.account.Account;
 import org.spongepowered.api.service.economy.transaction.TransactionResult;
+import org.spongepowered.api.service.economy.transaction.TransactionType;
+
+import java.math.BigDecimal;
 
 /**
- * Fired when the {@link EconomyService} has processed a transaction.
+ * Fired when the {@link EconomyService} is processing a transaction.
  */
 public interface EconomyTransactionEvent extends Event {
 
     /**
-     * Gets the {@link TransactionResult} for the transaction that occurred.
-     *
-     * @return The {@link TransactionResult}
+     * Called before processing a transaction
      */
-    TransactionResult transactionResult();
+    interface Pre extends EconomyTransactionEvent, Cancellable {
+
+        /**
+         * Gets the {@link Account} involved in the transaction.
+         *
+         * @return The {@link Account}
+         */
+        Account getTargetAccount();
+
+        /**
+         * Gets the {@link Currency} involved in the transaction.
+         *
+         * @return The {@link Currency}
+         */
+        Currency getCurrency();
+
+        /**
+         * Gets the amount of the {@link Currency} involved in the transaction.
+         *
+         * @return The amount
+         */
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/economy/EconomyTransactionEvent.java
```diff
@@ -24,20 +24,71 @@
...
+         * Gets the amount of the {@link Currency} involved in the transaction.
+         *
+         * @return The amount
+         */
+        BigDecimal getAmount();
+
+        /**
+         * Sets the amount of the {@link Currency} involved in the transaction.
+         */
+        void setAmount(BigDecimal amount);
```
Zidane, line 78: Provide the original amount as well.

src/main/java/org/spongepowered/api/event/economy/EconomyTransactionEvent.java
```diff
@@ -24,20 +24,78 @@
...
      */
-    TransactionResult getTransactionResult();
+    interface Pre extends EconomyTransactionEvent, Cancellable {
+
+        /**
+         * Gets the {@link Account} involved in the transaction.
+         *
+         * @return The {@link Account}
+         */
+        Account getTargetAccount();
```
Zidane: Could simply be getAccount()

src/main/java/org/spongepowered/api/event/economy/EconomyTransactionEvent.java
```diff
@@ -24,20 +24,78 @@
...
+     * Called after processing a transaction
+     */
+    interface Post extends EconomyTransactionEvent {
+
+        /**
+         * Gets the {@link TransactionResult} for the transaction that occurred.
+         *
+         * @return The {@link TransactionResult}
+         */
+        TransactionResult getTransactionResult();
```
Zidane: Does the result contain the account, currency type, etc?

ImMorpheus: Yep https://github.com/SpongePowered/SpongeAPI/blob/94d7fb22fcc0b77e5af44b9f9e7dc68534ecc3da/src/main/java/org/spongepowered/api/service/economy/transaction/TransactionResult.java#L38

ImMorpheus:
Looking back at this, I missed something about `EconomyTransactionEvent.Pre` and `getAccount`. Specifically: while the method works fine for deposit/withdraw, it's unclear for transfer since that involves two accounts. Something to retrieve the other account is needed. 
Do we add a method returning an Optional<Account> ? A subclass for the event ? 
Input wanted.

dualspiral:
Would it be better to just have an event that can handle multiple transactions at once? Rather than having a `getAccount()` and `getCurrency()` etc., have a collection of `ProposedTransactions` (or whatever) that contains this info, each one being a simple deposit or withdraw That way, a transfer is one event and the same event, and a plugin listening to this can treat/use it the same way as other plugins.

This could also allow economy plugins to build a complex web of transactions that might happen - a currency exchange plugin might allow two players to exchange their currencies so you might have two deposits and two withdrawals that all depend on each other (two accounts, two currencies, deposit and withdrawal on each side).

dualspiral:
I've shifted the base branch here to API-9, but is this something we want to revisit, or are we better off closing and thinking about Economy in a wider sense?

PR #2164, opened 2020-07-10
Add AccountHolder

ImMorpheus:
This should make it easier for 3rd party AccountHolders (towns, shop, chest shop, etc.) as well as clearly define the Account of a Player.


src/main/java/org/spongepowered/api/service/economy/account/AccountHolder.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An account holder is a holder of an {@link Account}.
+ */
+public interface AccountHolder {
+
+    /**
+     * Gets the {@link Account} for this account holder,
+     *
+     * <p>If an account does not already exist for this account holder,
```
Zidane: I disagree with this notion. I am almost not wanting this interface as a whole because creating the account should be opt-in and the notion of checking if something has one shouldn't force the creation of one.

Unfortunately the name of interface gives the impression "Yes I will hold one if Economy is present".

ImMorpheus:
Expanding a bit on why and how of this PR.

### Why

#### a. No standard way to get an Account for API defined classes (for this PR: ServerPlayer)

Plugins using the Economy API are left guessing how to get a Player account. 
This is the interface we provide:

https://github.com/SpongePowered/SpongeAPI/blob/94d7fb22fcc0b77e5af44b9f9e7dc68534ecc3da/src/main/java/org/spongepowered/api/service/economy/EconomyService.java#L95-L124

UniqueAccounts (identified by `UUID`s) and VirtualAccount (identified by `string`s) is all we have.
Sure, account creation and all that is a job of the Economy plugin providing the EconomyService (eco plugin from now on) but the API should provide a clue on what a Player Account is.

There's only one reference to Player accounts here
https://github.com/SpongePowered/SpongeAPI/blob/94d7fb22fcc0b77e5af44b9f9e7dc68534ecc3da/src/main/java/org/spongepowered/api/service/economy/account/UniqueAccount.java#L36-L37
suggesting `UniqueAccount`s are Player accounts identified by their `UUID`s.
That's just a suggestion though and what is actually happening is plugin devs being creative with their code, resorting to things like:

ecoService.getOrCreate(uuid)   // Player#getUniqueID
ecoService.getOrCreate(name) // Player#getName
ecoService.getOrCreate(id)       // Player#getIdentifier

The first one is an `UUID`, the others are strings. 
The first one is returning a `UniqueAccount`, the others are returning a `VirtualAccount`.

#### b. Provide 3rd party plugins a way to mark an Object as an AccountHolder (it holds an Account)

_Note: Account creation is still delegated to the eco plugin._ 

Ex. 
A is a towny plugin
B wants to use the towny API and wants to get the town bank. 

He can:
1. call ecoservice#getOrCreate guessing what A is using as the town identifier for the Account. (An update from A is all it takes to screw this up)
2. pray really hard that A has a Town#getBank method.

This PR aims to make 2. easier by providing a common means to define `AccountHolder`s and get their `Account`.

### How

An `AccountHolder` interface is added to solve problem b. and a method in `ServerPlayer` is added to solve problem a.

___

Now, based on a quick discussion on discord I want to address a few points.

1. It's not up to the EconomyService to decide how an `AccountHolder` is identified. It's up to the owner of the class. 
For `ServerPlayer` it's us. For a town is the town plugin.
The eco plugin handles Account creation and manages transaction. That's it.
This is for a number of reasons, namely:
An eco plugin can't possibly know every `AccountHolder` implementation out there and have 200 dependencies just to handle specific logic for each `AccountHolder`.
3rd party plugins can only query an `Account` by identifier. They have no way to know how eco plugins decide to handle a specific `AccountHolder`s because that's an implementation detail. Knowing that would defy the point of having a common interface for the `EconomyService` in the API.

> Take Pieconomy for example. Why would you use that method when, actually, his account is the inventory? 

2. Even with Pieconomy a plugin would query the `EconomyService` (provided by Pieconomy) to get the player account. The problem we should try to solve (problem a.) is how to query the player account.

> What if an economy plugin just doesn't want to tie accounts to players. 

3. There's no way to do that and this PR doesn't change it. An eco plugin provide an EconomyService which receives either a string identifier or a UUID and return an Account. It doesn't know who owns that Account. It just knows by what it's identified.


> Honestly, I'd ditch the interface and just have a getAccount() on ServerPlayer

4. That's fine by me, it solves problem `a.` which is the main focus of the PR. (assuming we keep the current default implementation)

### Alternative solution

An alternative approach is to replace `UniqueAccount` with `PlayerAccount`.
It would remove the method from the ServerPlayer interface, leave the AccountHolder as a marker for plugins devs and the EconomyService interface with:

getOrCreate(string identifier)        //for 3rd party AccountHolders
getOrCreate(ServerPlayer player) //for player accounts

But what about offline players (User) ? Use a GameProfile instead ?

dualspiral:
I was thinking about this last night and I can see some more value in this - though I think the problem is not solved by simply saying "slap this interface on, you're good".

You've missed off some parts of the conversation that happened in Discord which I think are pretty key to this. @Zidane said

> That can't be implemented though
> Without saying "It must be a UUID"
> What if my accounts are based on names?
> Me = Econ plugin

of which I then said:

> Right
> It would be better to have something like  EconService#getAccounts(AccountHolder)
> because then the economy plugin has full control over what it does for a player, or a town, or whatever
> the control needs to be on the econ service as much as possible.

The point that was being made was that we do not know how an economy plugin will store this data and what parts of the identifier it will use. Zidane's point was that they might not choose to store by UUID, or you might store by UUID from a town, or whatever. The econ plugin might very well have a different identifying mechanism.

Thinking back, however, that way of thinking is actually kind of bedded into the way things are set up now. Yep, the object, in general, is identifying itself and the econ plugin should take note of that (but said plugin could then choose to use a different identifier to store the values, of course). So this interface _might_ hold value, but I think the whole economy service needs to be looked at in tandem.

This interested me:

> It's not up to the EconomyService to decide how an AccountHolder is identified. It's up to the owner of the class.

If that's the case, why is there no way to identify the `AccountHolder` with just that interface? If it's not a player, the economy plugin may not have any clue as to whether it's a town, or whatever. There needs to be a way to identify that. There may also be a case for identifying whether the account holder is a player, or a town, or a world bank (if someone chose to go mega-economy on their server).

The major problem with what you've done is said that the implementation controls the identifer, and the econ plugin doesn't have any information beyond that - and that limits the control of the econ plugin. It actually leads to what is really a deficiency in the econ service itself, which is really the larger problem here. You've correctly noted that the `getAccount(UUID)` method only provides a suggestion that the `UUID` is a player's `UUID` - and that's not good.

Your alternative solution is better because the control properly goes through to the econ plugin - they have _more_ information that they can work with. I'd just have the `Account` interface. I'd have, for example:

```
Account getOrCreate(String identifier); // create arbitary account - this is always going to be important.

Account getOrCreate(AccountHolder holder);  // create account for (registered?) account holder.
```

A `User` would be an `AccountHolder`.

If you were to then have a `getAccount()` on the player, I think it should be implemented as `EconomyService#getAccount(this)`. `AccountHolder` would need to provide identifying inforation, alternatively, you could create a catalog type that converts an `AccountHolder` to `UUID` or `String` function that provides a way to perform identification if you wanted to (e.g. User -> getUniqueId().toString()), or register class types with the econ plugin and say "hey, this is a type of `AccountHolder` we support, and this is how you identify it". 

You could make it a requirement to register `AccountHolder` types and define how accounts should operate - when withdrawing, should there be specific people in the cause? Do they need a permission? I seem to recall that there was the notion of an `AccountArchetype` in a previous PR which seemed to be superflous to requirements there, but may have a use here - register an account holder to have an archetype or some other control system to also give plugins some of that control.

These are just ideas and some might not be viable, but I just feel like slapping this interface on just papers over issues in the EconomyService that we have a golden oppotunity to fix, so that Econ plugins get a lot more control. Plugins that tightly integrate with the economy could benefit from a full discussion of how this might work, such that they can register town accounts with their own control over how the account runs without having to guess whether account "abc-def" is a town account or not.

ImMorpheus:
> That can't be implemented though
> Without saying "It must be a UUID"
> What if my accounts are based on names?
> Me = Econ plugin

> The point that was being made was that we do not know how an economy plugin will store this data and what parts of the identifier it will use. Zidane's point was that they might not choose to store by UUID, or you might store by UUID from a town, or whatever. The econ plugin might very well have a different identifying mechanism.
> 
> Thinking back, however, that way of thinking is actually kind of bedded into the way things are set up now. Yep, the object, in general, is identifying itself and the econ plugin should take note of that (but said plugin could then choose to use a different identifier to store the values, of course). So this interface might hold value, but I think the whole economy service needs to be looked at in tandem.

It can't do that. 
Again, you can create an account using an identifier (be it a UUID for UniqueAccounts or a string for VirtualAccounts) and that's about it.
The eco plugin can encode, add a prefix, add a suffix, do black magic with that identifier. We (or 3rd party plugins) don't care. What we care about is that said account can be gotten/deleted using the identifier we used to create the account.


> Right
> It would be better to have something like EconService#getAccounts(AccountHolder)
> because then the economy plugin has full control over what it does for a player, or a town, or whatever
> the control needs to be on the econ service as much as possible.

See point 1. in my previous post.

> If that's the case, why is there no way to identify the AccountHolder with just that interface? If it's not a player, the economy plugin may not have any clue as to whether it's a town, or whatever. There needs to be a way to identify that. There may also be a case for identifying whether the account holder is a player, or a town, or a world bank (if someone chose to go mega-economy on their server).

Good point.

> The major problem with what you've done is said that the implementation controls the identifer, and the econ plugin doesn't have any information beyond that - and that limits the control of the econ plugin. It actually leads to what is really a deficiency in the econ service itself, which is really the larger problem here. You've correctly noted that the getAccount(UUID) method only provides a suggestion that the UUID is a player's UUID - and that's not good.

The situation is not changed. An eco plugin doesn't have any information beyond the identifier. It never did. 
It seems that's what should be changed.

> Your alternative solution is better because the control properly goes through to the econ plugin - they have more information that they can work with. 

Keep in mind that from a 3rd party plugin perspective I just changed the identifier from a UUID to a Player object. Those plugins would query a player account by their player object.

> AccountHolder would need to provide identifying inforation

uh, so giving the owner of the AccountHolder control over the identifier, maybe specifying in the jd that it's only a hint and the eco plugin can just ignore the identifier and use its own.
That would make everyone happy. I think I'm going with that.

dualspiral:
My point wasn't "AccountHolder needs a couple of methods, also query by player, jobs a good'un", which is what I feel you're still edging towards. This is sidestepping a problem, not solving it. It's about the wider "okay, but what does this solve, and how can we improve the API to do what people want to do natively?"

At the end of the day, it boils down to this: if we basically want the economy service to be dumb, then getting the account for a player needs to basically state the player UUID is the key for players and everything else has a string identifier. Towny plugins specify in their docs what string identifiers to use for their accounts. Else, let's be bold and consider how this is really used, could be used, and make changes for that. Let's not just slap an interface on players and hope that solves everything.

Specific comments:

> Again, you can create an account using an identifier (be it a UUID for UniqueAccounts or a string for VirtualAccounts) and that's about it.

Why do we have both unique and virtual accounts? Surely that's partially arisen from this identification problem? I think there was some talk about removing the notion of these separate accounts, at least on a consumer facing level.

> >    Right
> >    It would be better to have something like EconService#getAccounts(AccountHolder)
> >    because then the economy plugin has full control over what it does for a player, or a town, or whatever
> >    the control needs to be on the econ service as much as possible.

> See point 1. in my previous post.

I did. It's something I talked about later in my post.

> If you were to then have a getAccount() on the player, I think it should be implemented as EconomyService#getAccount(this). AccountHolder would need to provide identifying inforation, alternatively, you could create a catalog type that converts an AccountHolder to UUID or String function that provides a way to perform identification if you wanted to (e.g. User -> getUniqueId().toString()), or register class types with the econ plugin and say "hey, this is a type of AccountHolder we support, and this is how you identify it".
> 
> You could make it a requirement to register AccountHolder types and define how accounts should operate - when withdrawing, should there be specific people in the cause? Do they need a permission? I seem to recall that there was the notion of an AccountArchetype in a previous PR which seemed to be superflous to requirements there, but may have a use here - register an account holder to have an archetype or some other control system to also give plugins some of that control.


> Keep in mind that from a 3rd party plugin perspective I just changed the identifier from a UUID to a Player object. Those plugins would query a player account by their player object.

No... as I said in my post it should at least be a `User` for the very reason that it needs to be something you can query offline. You even made that point...

> But what about offline players (User) ? Use a GameProfile instead ?

> The situation is not changed. An eco plugin doesn't have any information beyond the identifier. It never did.
It seems that's what should be changed.

Yes. This is an oppotunity to try to improve things on a larger scale.

ImMorpheus:
I made it pretty clear what problems this PR solves and what the aim is.

For others who want to chime in and have a discussion:

- I don't mind giving more control/knowledge to the eco plugin. For example by giving it the whole object instead of the identifier (in this PR the AccountHolder as an argument in the eco service).
- I don't want to give total control to the eco plugin. Read this as "AccountHolder is just a marker and does not provide any identifying inforation". (in this PR the owner of the AccountHolder can provide an hint on what identifier to use).¹
- We need to establish a standard way to query a player account since we are the owner of that class. (in this PR ServerPlayer is an AccountHolder an you can just query by AccountHolder) ²
- I'm ok with querying by AccountHolder and merge UniqueAccounts and VirtualAccounts (this is what the PR is doing now)

¹ If you want this, please provide a solution of how this would work. Note that having the eco plugin have a dependencies on each plugin that implement an AccountHolder is not a solution, it's madness.
² I said **query**. How the eco plugin internally identifies (UUID, string, name, inventory, encoded string, etc.) an account is none of our business. What we need to make sure is that 3rd party plugins can get a player account without taking guesses. Be it with by passing the UUID or the Player/User/AccountHolder instance.

ST-DDT:
> > AccountHolder would need to provide identifying inforation
>
> uh, so giving the owner of the AccountHolder control over the identifier, maybe specifying in the jd that it's only a hint and the eco plugin can just ignore the identifier and use its own.
That would make everyone happy. I think I'm going with that.

I think we are going too fast here.

If the identifier is just decoration then the plugins providing that identifier are troubled if they for example rename the AccountHolder and the eco plugin uses the name instead.

Also it does not cover for cases where the kingdoms balance is the same as the kings.
(The player pays for all expenses of the town).

Lets use reality as a guide here.
If I wish to send someone money. I ask that person "whats your Account(number)". Maybe we can slightly defiate from and just ask the economy plugin directly give me the account for `type:xyz`. I would even go as far as declaring `type` to a catalog type.
For players this would look like this `sponge:player:12334-1234-134...`
For those towns it could be `towny:town:my-awesome-city2k20`.
How the economy plugin stores those keys? We dont care. We also don't care whether the town returns a player account or a guild account or whatever.

The following code  is used to express my concept in more detail. It does not cover all aspects related to economy only the X -> account mapping. So please take it with a grain of salt.

````java
interface AccountOwnerType extends CatalogType {
}
````

````java
final class AccountRef { // IBAN

static AccountRef of(String accountId)

static AccountRef of(AccountOwnerType type, String ownerId)

String getAccountId() {
    return getOwnerType() + ":" + getOwnerId();
}

AccountHolderType getOwnerType();

String getOwnerId();

}
````

````java
interface AccountHolder { // Optional

AccountRef/Account getAccount(); // Return type does not matter

}
````

````java
interface AccountRefResolver {

Optional<AccountRef> resolve(Object source);

}
````

````java
class PlayerAccountRefResolver implements AccountRefResolver {

Optional<AccountRef> resolve(Object source) {
    if (source instanceof Player p) {
        return AccountRef.of(AccountOwnerTypes.PLAYER, p.getUuid().toString());
    }
    return null;
}

}
````

````java
class MyShopAccountRefResolver implements AccountRefResolver {

Optional<AccountRef> resolve(Object source) {
    if (source instanceof InventoryShop is) {
        return AccountRef.of(AccountOwnerTypes.PLAYER, is.getOwnerUuid());
    }
    if (source instanceof Merchant m) {
        return myPlugin.getShop(m).getAccountRef();
    }
    return null;
}
````

````java
interface EconomyService {

Account getAccount(AccountRef)

Account resolveAccount(Object target)

AccountRef resolve(Object target)

void register(AccountRefResolver)

}
````

**Comment by dualspiral** [comment]
I think this is something we now need to revisit before we wind down for freezing API 8.

The ultimate issue that really needs to be considered is the fact that there are many different objects that may very well have an account on them. Thinking over this some more, I came to a similar conclusion as @ST-DDT above, that we may be better served with the notion of an `AccountHolderType` or something to that effect.

It could be something like this:

```java
interface AccountHolderType<T> {

    String identifier(T type);    

}
```

add this `AccountHolderType` to the registry so Econ plugins can query it, and use it to get the unique identifier, using something like:

```
Optional<Account> findAccount(AccountHolderType<T> type, T entity);
```

etc. The use of an `AccountRef` may also be useful in the way that it links a type and entity together (but it may not be necessary)

We could have two standard types:

```
AccountHolderType.GENERIC
AccountHolderType.USER
```

and other plugins can support town banks, even if the plugin is an addon to a town plugin and said town plugin doesn't make itself economy aware. And of course, we need to keep hold of a generic type in case something does just want a one off account.

There is one large drawback to this idea, however: what if a plugin providing a town type is removed - how can we access it after removal? Do we provide a method such as

```
Optional<Account> findAccount(String type, String identifier);
```

to account for this scenario? I'm not entirely sure what to suggest there.

(I do not think we need to complicate matters with resolvers. By forcing the use of types plugin developers are unambiguously stating what their intention is).

---

Where do we think we want to go with this?

dualspiral:
This stalled a long time ago and is no longer eligible for API-8 as it's breaking. I'm going to close this.

I would say that as far as economy goes, it might be worth starting to plan out what is needed from a economy service rather than just throw out some code, and work out where we're gonna go from there. I think there was a huge clash of ideals here and it unfortunately petered out.

Issue #970, opened 2015-12-20
Allow Plugins to Mutate XP for Block Breaks & Entity Deaths


DarkArc:
There are currently no good ways to mutate the amount of experience dropped when a block is broken, or an entity dies. 

Note: This may also effect other areas of the API.

Zidane:
@DarkArc 

Entity XP orbs are just, well, entities. IF the cause/contextual data is correct and you know WHY the XP orb is spawning, this is a non-issue (as you'd grab the Entity and adjust its data).

Is that not the case or?

DarkArc:
@Zidane It's not quite the case, as increasing/decreasing the XP may result in more XP orbs in Vanilla which is a different effect than increasing the XP value of the orb.

PR #2172, opened 2020-07-15
Allow Plugins to Mutate XP for Block Breaks & Entity Deaths

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/970

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/DropExperienceEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/DropExperienceEvent.java
@@ -1,0 +1,89 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/entity/DropExperienceEvent.java
```diff
@@ -0,0 +1,89 @@
...
+     * Sets the new amount of experience.
+     *
+     * @param experience The new amount of experience to change to
+     */
+    void setExperience(int experience);
+
+    /**
+     * An event that is fired when a block drop experience on when broken
+     */
+    interface TargetBlock extends DropExperienceEvent {
```
Zidane, line 64: We've moved away (or rather are) from TargetXXXX events.

Check the cause.

PR #2173, opened 2020-07-15
UserManager#match now returns a Stream

dualspiral:
As title. [See common PR for more details](https://github.com/SpongePowered/SpongeCommon/pull/3067).

### DIFF
--- a/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
@@ -33,6 +33,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/profile/GameProfileCache.java
```diff
@@ -243,7 +243,7 @@ default boolean add(GameProfile profile, @Nullable Instant expiry) {
      * <p>This collection may also contain profiles of players who never played
      * on the server!</p>
      *
-     * <p>Use {@link UserManager#match(String)} for a collection that
+     * <p>Use {@link UserManager#streamOfMatches(String)} for a collection that
```
ST-DDT:
```suggestion
     * <p>Use {@link UserManager#streamOfMatches(String)} for a stream that
```

PR #2174, opened 2020-07-15
Use a double for damage related keys.

Cybermaxke:
API | [Common](https://github.com/SpongePowered/SpongeCommon/pull/3095)

Using an int is an impl detail, damage is represented as doubles.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -585,7 +585,7 @@ public final class Keys {
      * The damage absorbed by an armor {@link ItemStack}.
      * Readonly
      */
-    public static final Supplier<Key<Value<Integer>>> DAMAGE_ABSORPTION = Sponge.getRegistry().getCatalogRegistry().provideSupplier(Key.class, "DAMAGE_ABSORPTION");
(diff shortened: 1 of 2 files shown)


ImMorpheus:
This should apply to `Keys#DAMAGE_ABSORPTION` as well

Zidane:
@Cybermaxke Please update this.

src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -1758,7 +1758,7 @@
     /**
      * The maximum damage a {@link FallingBlock} can deal.
      */
-    public static final Supplier<Key<BoundedValue<Integer>>> MAX_FALL_DAMAGE = Sponge.getRegistry().getCatalogRegistry().provideSupplier(Key.class, "MAX_FALL_DAMAGE");
+    public static final Supplier<Key<BoundedValue<Double>>> MAX_FALL_DAMAGE = Sponge.getRegistry().getCatalogRegistry().provideSupplier(Key.class, "MAX_FALL_DAMAGE");
```
gabizou: Is it treated differently compared to how the implementation uses this integer? I figure the implementation stores this value to nbt, so would we be modifying the implementation to store a double instead?

Cybermaxke: For now, I was thinking about just internally converting it to int and back. But storing could also be an option.

Cybermaxke:
@Zidane Done

Issue #2178, opened 2020-07-18
KickPlayerEvent does not register kick due to spam and is not cancellable 


codestripper:
**I am currently running**
- SpongeForge version: .12.2-2838-7.2.4-RC4054
- Java version: 1.8.0_251-b08
- Operating System: Windows 10

**Issue Description**
KickPlayerEvent does not register a player being kicked for spamming chat, and is not cancellable. I believe it should be cancellable to allow certain players the ability to bypass being kicked for various reasons.

One such example would be to prevent certain users from being kicked due to chat spamming. As it stands this does not seem possible to do.

```
    @Listener
    public void onPlayerKick(KickPlayerEvent event) {
        plugin.getLogger().info("Player kicked");
    }
```

dualspiral:
`KickPlayerEvent` just isn't fired.

There are a couple of issues here:

* If this event was cancelled, the `Player#kick` event returns a `void` so there is no way for a user of this event to know if the kick was successful. We'll need to introduce a new method.
* The event *MessageChannel methods (from `MessageChannelEvent`) shouldn't exist, really. What use is a message channel on an event that sends a message to a disconnected client?

This can all be fixed properly in API 8, however there are challenges for API 7. We'd like to add this in 7.3, but we do need to consider how we work around the problems above - I'm noting them here so I can get some feedback from a couple of others.

ImMorpheus:
PR merged

PR #2179, opened 2020-07-18
Make KickPlayerEvent cancellable

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/2178

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
@@ -191,14 +191,14 @@ public interface ServerPlayer extends Player, Subject {
      * Kicks the player, showing the default kick reason (the translation key
      * {@code disconnect.disconnected}).
      */
-    void kick();
+    boolean kick();
 
     /**
      * Kicks the player given a reason.
      *
      * @param reason The reason for the kick
      */
-    void kick(Component reason);
+    boolean kick(Component reason);
 
     /**
      * Gets the {@link Scoreboard} displayed to the player.
--- a/src/main/java/org/spongepowered/api/event/advancement/AdvancementEvent.java
+++ b/src/main/java/org/spongepowered/api/event/advancement/AdvancementEvent.java
@@ -29,6 +29,7 @@ import org.spongepowered.api.advancement.criteria.AdvancementCriterion;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.entity.living.player.server.ServerPlayer;
 import org.spongepowered.api.event.Event;
+import org.spongepowered.api.event.message.MessageCancellable;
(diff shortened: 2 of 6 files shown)


Zidane:
@ImMorpheus 

This can target 7.

ImMorpheus:
The event is not implemented on stable-7.

dualspiral:
With regards to 7: https://github.com/SpongePowered/SpongeAPI/issues/2178#issuecomment-667669811 - this needs to be thought about when it's done.

With regards to 8: remove the `MessageChannelEvent` implementation (or whatever the adventure equivalent is/will be). I do not see the use of having an audience attached to the method.

ImMorpheus:
Done

EDIT: while it's true channels are not used we still use the message of MessageEvent.

dualspiral:
> while it's true channels are not used we still use the message of MessageEvent.

Add it on manually, don't keep an interface on if you don't use all of the elements of it.

ImMorpheus:
> Add it on manually, don't keep an interface on if you don't use all of the elements of it.

They are two different events.

`MessageChannelEvent` != `MessageEvent`

`KickPlayerEvent` makes full use of the latter

dualspiral:
Also, while we're at it, make `Player#kick` return a boolean, because if a kick is cancelled, a comsumer of the event should be notified of it. This also needs an implementation update, so that needs doing in Common too.

dualspiral:
> KickPlayerEvent makes full use of the latter

[No it doesn't](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/event/message/MessageEvent.java#L42-L56) - these are not used.

ImMorpheus:
> [No it doesn't](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/event/message/MessageEvent.java#L42-L56) - these are not used.

True, `KickPlayerEvent#isMessageCancelled` is not used but shouldn't it be ?

`ServerSideConnectionEvent.Auth` and `ServerSideConnectionEvent.Login` are the same (both are `MessageEvent` whose message refers to the "disconnect message"). 
If the message is cancelled then the default `disconnect.disconnected` message is used. 
Why shouldn't the same apply to `KickPlayerEvent` ?

dualspiral:
> If the message is cancelled then the default disconnect.disconnected message is used.

I'd rather that was ditched entirely - cancelling a message doesn't make sense in that scenario (all it's doing is basically using the original message instead). The message event should just be original/get/set if you ask me - then it's totally appropriate to put here.

ImMorpheus:
> I'd rather that was ditched entirely - cancelling a message doesn't make sense in that scenario (all it's doing is basically using the original message instead)

Done

> The message event should just be original/get/set if you ask me - then it's totally appropriate to put here.

Done and added a separate interface to mark the event as "message cancellable" (just like Event and Cancellable).

Issue #1526, opened 2017-03-29
A few issues with SpongeEventFactory


simon816:
The following events are unable to be created with the factory:
`achievement.GrantAchievementEvent`
`data.ChangeDataHolderEvent`
`entity.ExpireEntityEvent`
`entity.HarvestEntityEvent`
`statistic.ChangeStatisticEvent`

The following events have `createX()` methods but shouldn't (these are 'base' events):
`action.InteractEvent`
`action.CollideEvent.Impact`
`block.TargetBlockEvent`
`entity.AffectEntityEvent`
`entity.TargetEntityEvent`
`entity.explosive.TargetExplosiveEvent`
`entity.explosive.TargetFusedExplosiveEvent`
`entity.item.TargetItemEvent`
`entity.living.TargetAgentEvent`
`entity.living.TargetLivingEvent`
`entity.living.humanoid.HandInteractEvent`
`entity.living.humanoid.TargetHumanoidEvent`
`entity.living.humanoid.player.TargetPlayerEvent`
`entity.projectile.TargetProjectileEvent`
`game.state.GameStateEvent`
`item.inventory.AffectItemStackEvent`
`item.inventory.AffectSlotEvent`
`item.inventory.TargetContainerEvent`
`item.inventory.TargetInventoryEvent`
`server.ClientPingServerEvent.Response`
`server.ClientPingServerEvent.Response.Players`
`user.TargetUserEvent`
`world.TargetWorldEvent`
`world.chunk.TargetChunkEvent`

liach:
Which ones are still valid? `GrantAchievementEvent`, for example, no longer exists.

liach:
Given `server.ClientPingServerEvent.Response` and `server.ClientPingServerEvent.Response.Players` are not even events, are they still created by the factory?

ImMorpheus:
PR merged

PR #1977, opened 2019-03-02
Correct event gen, take two

liach:
Replaces #1759
Fixes #1526, tweak generation for sponge event factory
Signed-off-by: liach <liach@users.noreply.github.com>

PR #2185, opened 2020-07-22
Fix event-gen factory

ImMorpheus:
Fix https://github.com/SpongePowered/SpongeAPI/issues/1526

See https://github.com/SpongePowered/SpongeAPI/pull/1977

I've updated the PR (most of those events are gone).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -111,6 +111,7 @@ tasks {
         exclude("org/spongepowered/api/event/lifecycle/StartedEngineEvent.java")
         exclude("org/spongepowered/api/event/lifecycle/StoppingEngineEvent.java")
         inclusiveAnnotations = setOf("org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod")
+        exclusiveAnnotations = setOf("org.spongepowered.api.util.annotation.eventgen.NoFactoryMethod")
     }
 
     jar {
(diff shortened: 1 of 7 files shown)


Zidane:
@ImMorpheus Put these on the lifecycle file exclusions please.

dualspiral:
How is `@NoFactoryMethod` different from an `exclude` directive in the Gradle file? If there is no difference, I'd rather we just add things to the exclude directive, especially as we want to move the factory generation to Common, reducing the exposure of the generator to the API itself.

Zidane:
@dualspiral That is a good point, didn't even think about that.

ImMorpheus:
> How is @NoFactoryMethod different from an exclude directive in the Gradle file? If there is no difference, I'd rather we just add things to the exclude directive, especially as we want to move the factory generation to Common, reducing the exposure of the generator to the API itself.

The exclude directive targets a java file. The annotation targets a type.
This means that 
https://github.com/SpongePowered/SpongeAPI/blob/dc2186f59d37dd277ef8d02b0aef72f3fde2f61a/src/main/java/org/spongepowered/api/event/action/CollideEvent.java#L37-L47
can't be implemented using the exclude directive AFAIK.

I'll move the other ones since this doesn't apply to them.

gabizou (approved):
LGTM

PR #2188, opened 2020-07-23
Improve the chunk data stream.

Cybermaxke:


src/main/java/org/spongepowered/api/world/storage/WorldStorage.java
```diff
@@ -42,14 +43,54 @@
...
      *
-     * <p>Usage of a {@link ChunkDataStream} should be limited to asynchronous
-     * tasks to avoid hanging the main thread.</p>
+     * <p>Usage of this stream should be limited to asynchronous tasks to
+     * avoid hanging the main thread.</p>
      *
      * @return An iterator of generated chunks
      */
-    ChunkDataStream getGeneratedChunks();
+    Stream<ChunkEntry> getGeneratedChunks();
```
gabizou, line 93: I'm beginning to think this should be a CompletableFuture if asynchronous threads are intended to be used, because this can have implications of needing to wait on the main thread to sync up anyways.

PR #2192, opened 2020-07-27
Set the default value for sharedRoot to false in ConfigDir and DefaultConfig

alexstaeding:
The majority of plugins do not put their configuration in the shared root configuration directory. This PR makes this normal case less verbose but leaves the option to put a configuration in the root directory if needed.

There have been talks in the 1.14 updates channel of potentially removing the option to set sharedRoot at all. However, I believe that setting a default false value is a better alternative. It provides both the convenience of not having to specify it and keeps the option available for those who want it.

### DIFF
--- a/src/main/java/org/spongepowered/api/config/ConfigDir.java
+++ b/src/main/java/org/spongepowered/api/config/ConfigDir.java
@@ -55,6 +55,6 @@ public @interface ConfigDir {
      * @see ConfigRoot#getConfigPath() For information on what the
      *     shared root is
      */
-    boolean sharedRoot();
+    boolean sharedRoot() default false;
 
 }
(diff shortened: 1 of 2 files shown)


dualspiral:
I'm not accepting this. Such a change is something that needs more discussion and, quite honestly, I think that it's important that there is a larger discussion around this when we come to it. You've taken what's been said in 1dot14updates and jumped the gun.

> It provides both the convenience of not having to specify it and keeps the option available for those who want it.

This is a really weak argument. `sharedRoot = true` or `sharedRoot = false` is not a burden on anybody and it makes the annotation's purpose more discoverable. Most plugin devs of established plugins aren't going to go change this either.

> There have been talks in the 1.14 updates channel of potentially removing the option to set sharedRoot at all.

Yes, talk, and mostly fleeting at that. Right now we're focussed on getting the core product working and deciding these things are going to be almost plumb last as we wire up the final bits. 

I get that this isn't necessarily a breaking change as it stands, but what if we then decide that actually, we want all configs to be in the root? Then we've backed the wrong horse and we're stuck with it. I'd rather not do that.

`@DefaultConfig(sharedRoot = false)` is not onerous and is clearer to the developer that there are options involved. We may well revisit it in 8 but as for now, it stays, and it remains mandatory.

PR #2194, opened 2020-07-30
Add ability to programmatically create selectors

dualspiral:
This isn't for creating new selector types (i.e. `@a` alternatives), this is simply for if you want a selector outside of commands. I chose to ditch the `ArgumentTypes` system because really that seemed to be a reimplementation of selectors and I'd prefer to sit on top of what Vanilla has.

I did actually start trying to redo the ArgumentTypes system, but it was a rabbit hole. I'm open to doing so in the future, but for now, this should cover most (if not all) of the current options.

Also added `Range` utility. Might be useful in advancements because of the MinMaxBounds in the advancement package.


src/main/java/org/spongepowered/api/command/selector/Selector.java
```diff
@@ -0,0 +1,474 @@
...
+        /**
+         * Adds a {@link GameMode} constraint to the selector, requiring players
+         * be in the given game mode
+         *
+         * <p>Cannot be used with {@link #notGamemode(GameMode)}.</p>
+         *
+         * @param mode The gamemode
+         * @return This builder, for chaining
+         */
+        Builder gamemode(GameMode mode);
```
Cybermaxke: `gameMode`

src/main/java/org/spongepowered/api/command/selector/Selector.java
```diff
@@ -0,0 +1,474 @@
...
+        /**
+         * Adds a {@link GameMode} constraint to the selector, requiring that
+         * players are not in the given game mode
+         *
+         * <p>Cannot be used with {@link #gamemode(GameMode)}.</p>
+         *
+         * @param mode The gamemode
+         * @return This builder, for chaining
+         */
+        Builder notGamemode(Supplier<GameMode> mode);
```
Cybermaxke: `notGameMode`

PR #2200, opened 2020-08-09
Portal API

Zidane:


src/main/java/org/spongepowered/api/effect/Viewer.java
```diff
@@ -46,6 +48,17 @@
  */
 public interface Viewer extends Audience {
 
+    /**
+     * Sends the effect of being in a particular Vanilla world environment, such as the Nether,
+     * as an effect to the viewer.
+     *
+     * <p>For example, specifying {@link DimensionTypes#THE_NETHER}</p> will create a red skybox and
```
gabizou: Why close the paragraph mid sentence?

gabizou (approved):
A small nitpick

src/main/java/org/spongepowered/api/world/portal/PortalTypes.java
```diff
@@ -22,19 +22,20 @@
...
 import org.spongepowered.api.Sponge;
 
 import java.util.function.Supplier;
 
-public final class PortalAgentTypes {
+public final class PortalTypes {
 
-    // The default agent used to handle traveling to nether and the end
-    public static final Supplier<PortalAgentType> DEFAULT = Sponge.getRegistry().getCatalogRegistry().provideSupplier(PortalAgentType.class, "DEFAULT");
+    public static final Supplier<PortalType> NETHER = Sponge.getRegistry().getCatalogRegistry().provideSupplier(PortalType.class, "NETHER");
```
Cybermaxke, line 33: `THE_END` and `THE_NETHER` to be consistent with dimension and generator types naming?

dualspiral: I don't think that makes sense here. We generate "the nether" and "the end" (there is one nether dimension type and one nether generator, for example) - but we use "nether" and "end" portals, not "the nether" or "the end" portals - of which there may be multiple. I think it's correct as is.

src/main/java/org/spongepowered/api/effect/Viewer.java
Zidane: Whoops missed that.

src/main/java/org/spongepowered/api/event/entity/MoveEntityEvent.java
```diff
@@ -70,102 +72,14 @@
...
     @GenerateFactoryMethod
-    interface Position extends MoveEntityEvent {}
+    interface Position extends MoveEntityEvent {
+
+    }
 
     /**
-     * Fired when an {@link Entity}'s position changes for reasons other than
-     * normal movement.
+     * Fired when an {@link Entity entity's} position goes across {@link ServerWorld worlds}.
```
dualspiral: I don't think it should really be named "Teleport" if this is the condition, because plugin devs associate a teleport as a non-natural, such as `/tp`, movement, and they often don't involve a world change.

A `ChangeWorldEvent` is a specialised `MoveEntityEvent` really. I would argue that if you are changing world, you should _always_ be teleporting to a specific position meaning a standalone event makes no sense. So I think that should be a sub-event, and add a `MovementType` to the event itself so plugins can inspect what type of movement caused this.

src/main/java/org/spongepowered/api/event/cause/entity/teleport/MovementTypes.java
```diff
@@ -28,24 +28,24 @@
...
-    public static final Supplier<TeleportType> ENTITY_TELEPORT = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "ENTITY_TELEPORT");
+    public static final Supplier<MovementType> ENDER_PEARL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "ENDER_PEARL");
 
-    public static final Supplier<TeleportType> PLUGIN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "PLUGIN");
+    public static final Supplier<MovementType> NATURAL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "NATURAL");
 
-    public static final Supplier<TeleportType> PORTAL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "PORTAL");
+    public static final Supplier<MovementType> PLUGIN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "PLUGIN");
 
-    public static final Supplier<TeleportType> UNKNOWN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "UNKNOWN");
```
dualspiral: Is there a situation where we won't be able to determine what the movement type was? We might want to keep `UNKNOWN` if that's the case.

src/main/java/org/spongepowered/api/event/entity/MoveEntityEvent.java
```diff
@@ -70,102 +72,14 @@
      * Fired when an {@link Entity}'s position changes.
      */
     @GenerateFactoryMethod
-    interface Position extends MoveEntityEvent {}
+    interface Position extends MoveEntityEvent {
```
dualspiral: This was added to be able to separate position and rotation movement - but the rotation event is actually separate. So why do we need this? It would make more sense if this event contained entity rotation... but it does not.

I think we should specify that the `MoveEntityEvent` refers to position only, and that the rotation event is rotation only and just separate the movement and rotation logic completely.

src/main/java/org/spongepowered/api/event/entity/MoveEntityEvent.java
```diff
@@ -27,36 +27,38 @@
...
 import org.spongepowered.api.util.annotation.eventgen.GenerateFactoryMethod;
 import org.spongepowered.api.world.server.ServerWorld;
-import org.spongepowered.api.world.teleport.PortalAgent;
 import org.spongepowered.math.vector.Vector3d;
 
 /**
  * Called when an {@link Entity} performs movement.
  */
-@GenerateFactoryMethod
 public interface MoveEntityEvent extends Event, Cancellable {
```
dualspiral, line 36: I think this should contain `getServerWorld()` or `getServerLocation()` as many plugins will be interested in the entity's location, which includes the world they are in, and will want to act upon that. Don't need a setter, but a getter will support plugin devs better.

dualspiral (commented):
Mostly have thoughts about the events - I think they need a bit of rejigging.

src/main/java/org/spongepowered/api/event/entity/MoveEntityEvent.java
Zidane: There is the problem though.

Portal repositioning logic is VERY expensive. I fire the ChangeEntityWorld right away and allow you to change the world.

Not only that but I can't allow the Teleport to set the world as it breaks' the Mojang Teleporter so I compromised by letting you do that here (in-case you only care about the World you are going to) and then if the dev cares about the position you must cancel and do it yourself.

src/main/java/org/spongepowered/api/event/cause/entity/teleport/MovementTypes.java
Zidane: If it is unknown it should be treated as plugin. We control exactly when this is used so we can set the movement. Unknown, by the merits, is unhelpful.

src/main/java/org/spongepowered/api/event/cause/entity/teleport/MovementTypes.java
```diff
@@ -28,24 +28,28 @@
...
+    public static final Supplier<MovementType> COMMAND = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "COMMAND");
 
-    public static final Supplier<TeleportType> ENTITY_TELEPORT = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "ENTITY_TELEPORT");
+    public static final Supplier<MovementType> END_GATEWAY = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "END_GATEWAY");
 
-    public static final Supplier<TeleportType> PLUGIN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "PLUGIN");
+    public static final Supplier<MovementType> ENDER_PEARL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "ENDER_PEARL");
 
-    public static final Supplier<TeleportType> PORTAL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(TeleportType.class, "PORTAL");
+    public static final Supplier<MovementType> ENTITY_TELEPORT = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MovementType.class, "ENTITY_TELEPORT");
```
Cybermaxke: Just `TELEPORT`?

dualspiral:
7bc9064cf5c2e23cac42bcfcb8d98c1f0841b6e7

PR #2209, opened 2020-08-23
Remove EconomyService#deleteAccount default implementation

ImMorpheus:
`EconomyService#deleteAccount` was default implemented back in api 7 to avoid a breaking change. 
This PR removes the default implementation.
Input wanted on whether `AccountDeletionResultTypes#UNSUPPORTED` is worth keeping.


dualspiral:
> Input wanted on whether AccountDeletionResultTypes#UNSUPPORTED is worth keeping.

Do we expect econ services to not be able to delete their own data, or do we want to infer that they must provide that ability? I'd lean towards the latter, so it should disappear in that case.

ImMorpheus:
Done

PR #2215, opened 2020-08-28
Add Audiences#withPermission

ImMorpheus:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3138)

### DIFF
--- a/src/main/java/org/spongepowered/api/adventure/Audiences.java
+++ b/src/main/java/org/spongepowered/api/adventure/Audiences.java
@@ -47,7 +47,18 @@ public final class Audiences {
      * @return An audience
      */
     public static Audience onlinePlayers() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/adventure/Audiences.java
```diff
@@ -47,7 +47,18 @@ public static Audience server() {
...
+    }
+
+    /**
+     * Gets an {@link Audience} that targets all online players
+     * with the specified permission.
+     *
+     * @param permission The permission
+     * @return An audience
+     */
+    public static Audience withPermission(String permission) {
```
Zidane: `final` and `Objects.requireNonNull` in the method to fail out as soon as possible if there is a problem.

PR #2219, opened 2020-08-29
States cannot be cataloged so a backup system.

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3141)

Schematics need fixing as `BlockState`s are no longer `CatalogType`s.


dualspiral:
I think we're mostly there with this. I've folded the `BlockStateMatcher` into the `StateMatcher` and replaced the block state serialiser with the in built block state serialiser/deserialiser (not the DFU one). Fluid state is also serialisable and matching is available. Please take a look at the common impl in particular to see what I've done to make sure we're on the right track.

The one big thing left is the fact that the matchers are data serialisable. Why? I don't know how we want to serialise them but I'm not sure why they need to be in the first place - if we don't need them I can remove it, if we do, we need to work out how the container serialisation will work.

dualspiral:
> The one big thing left is the fact that the matchers are data serialisable. Why? I don't know how we want to serialise them but I'm not sure why they need to be in the first place - if we don't need them I can remove it, if we do, we need to work out how the container serialisation will work

We're not doing this.

src/main/java/org/spongepowered/api/state/State.java
```diff
@@ -35,7 +36,7 @@
 import java.util.Optional;
 import java.util.function.Supplier;
 
-public interface State<S extends State<S>> extends SerializableDataHolder.Immutable<S> {
+public interface State<S extends State<S>> extends SerializableDataHolder.Immutable<S>, DataSerializable {
```
gabizou: Would it make sense to have `State<S extends State<S> & DataSerializable>` instead? I don't want the rabbit hole of the state being serializable by itself (unless we want

PR #2222, opened 2020-08-31
Bring back ProjectileLauncher

RedNesto:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3149)

API's `ProjectileLauncher` was completely useless, even as a marker because `Living` extends `EntityProjectileSource`.


Faithcaio:
manually merged

PR #2225, opened 2020-09-05
Refactor the GameProfileManager and make GameProfile immutable.

Cybermaxke:
**API** | [Common](https://github.com/SpongePowered/SpongeCommon/pull/3153)

This PR refactors the `GameProfile`, `GameProfileManager` and `GameProfileCache`. 

- It's no longer possible to add entries to the `GameProfileCache`, only remove them to invalidate.
- `GameProfileCache` is now just a cache, it was previously also a lookup service.
- `GameProfileCache` is no longer replaceable.
- `GameProfile` is now immutable, but has transformation methods.
- "filled" profiles are now just "profile", and non filled ones are "basic profile", e.g. `getProfile` and `getBasicProfile`.
- "filled" was previously a bit confusing, because `isFilled` meant that the profile has a name, although filling a profile through `GameProfileManager` also added properties. This is now renamed to `hasName`.

### DIFF
--- a/src/main/java/org/spongepowered/api/profile/GameProfileManager.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileManager.java
@@ -36,7 +36,8 @@ public interface GameProfileManager extends GameProfileProvider {
      * Gets the {@link GameProfileCache} of this provider which will be
      * used to cache game profiles.
      *
-     * <p>In vanilla, properties do not persist between restarts.</p>
+     * <p>In vanilla, properties do not persist between restarts and profiles
+     * will expire after a month.</p>
      *
      * @return The cache
      */
--- a/src/main/java/org/spongepowered/api/profile/GameProfileManager.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileManager.java
@@ -24,201 +24,29 @@
  */
 package org.spongepowered.api.profile;
 
-import org.checkerframework.checker.nullness.qual.Nullable;
-import org.spongepowered.api.profile.property.ProfileProperty;
-
-import java.io.IOException;
-import java.util.Collection;
-import java.util.Map;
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/api/profile/GameProfileProvider.java
```diff
@@ -0,0 +1,201 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.time.Instant;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A service which handles how game profile data is requested.
```
dualspiral: It's not a service.

dualspiral (approved):
Barring my little javadocs comment, this looks fine.

src/main/java/org/spongepowered/api/profile/GameProfileManager.java
```diff
@@ -24,201 +24,30 @@
...
-    void setCache(GameProfileCache cache);
-
-    /**
-     * Gets the default cache.
-     *
-     * @return The default cache.
-     */
-    GameProfileCache getDefaultCache();
-
+    GameProfileProvider uncached();
```
Zidane, line 52: getUncached?

getDynamic?

With that said, I am not sure how wise it is to expose a provider that doesn't cache. As only due to the fact that profiles can be slow to query and Mojang does rate limit an IP from querying so many per minute I think.

Cybermaxke: The uncached one can still cache to consider rate limits, it just bypasses the default cache which retains profiles for a month.

Zidane:
Manually merged, thanks!

PR #2237, opened 2020-09-21
Update and implement checkstyle 

ImMorpheus:
xml copied from https://github.com/checkstyle/checkstyle/blob/checkstyle-8.36.1/src/main/resources/google_checks.xml

Updated to reflect https://docs.spongepowered.org/stable/en/contributing/implementation/codestyle.html
Some rules are also suppressed to reflect sponge coding style.


gabizou:
@ImMorpheus give an idea of what steps to set up IntelliJ + Gradle with the new configs. I updated the PR against api-8 for you so you don't have to deal with that shenanigans.

ImMorpheus:
File -> Settings -> Tools -> Checkstyle

Add a new `Configuration File`

2020-12-01-211147_502x403_scrot

It will ask you about the properties defined in the config:
set the suppressions.xml

2020-12-01-211223_505x412_scrot

Open a SpongeAPI class

Go in the CheckStyle tab, make sure you are using the `Sponge` rules and run `check module`

2020-12-01-212207_520x210_scrot

Note: the suppressions rules may not be updated

PR #2239, opened 2020-10-02
Update to licenser 0.5

jamierocks:
[Common pr](https://github.com/SpongePowered/SpongeCommon/pull/3171)


zml2008:
https://github.com/SpongePowered/SpongeAPI/blob/api-8/settings.gradle.kts#L17

jamierocks:
Aha. Thanks, was only testing in a SC environment so that explains it.

PR #2244, opened 2020-10-05
Genericize Palettes

gabizou:
**SpongeAPI** | [SpongeCommon]()

# Palettes and their evolution

Mostly as a request for WorldEdits' better handling, we seek to improve palettes to be more generic outside the purview of being indexing on `CatalogType`s alone. As Minecraft evolves, the handling of what a `BlockState` is allowed to contain increasingly becomes more and more data driven. As such, we need to less focus on a `ResourceKey` based object and more of a "here's a palette of object definitions that are defined by the `Palette`'s `PaletteType`.

## PaletteType and the encoder/decoder

Previously, we made `Palette` intrinsically serializable by building it up based on an index of `int`s to either `BlockState` or some other `CatalogType`. Now, we need to be able to serialize these `Palette`s  based on a predefinition from their `PaletteType` that already defines the *type* of object the `Palette` will support. As such, the `String` encoder/decoder will be maintained at the `PaletteType` level, and less so on the `Palette` itself.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/schematic/Palette.java
+++ b/src/main/java/org/spongepowered/api/world/schematic/Palette.java
@@ -27,8 +27,9 @@ package org.spongepowered.api.world.schematic;
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.util.annotation.CatalogedBy;
 
-import java.util.Collection;
 import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.stream.Stream;
 
 /**
  * Represents a mapping for types to a local identifier. Can be used for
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/world/schematic/Palette.java
```diff
@@ -72,30 +73,52 @@
...
-     */
-    boolean remove(T type);
 
     /**
      * Gets all {@code type T}s contained in this palette.
      *
      * @return All contained types
      */
-    Collection<T> getEntries();
+    Stream<T> getEntries();
```
Zidane: Just call it stream

Zidane: Actually, entryStream.

PR #2246, opened 2020-10-09
Removing Guava usage

gabizou:
Pretty basic premise, we've got issues with relying on guava, and as Configurate is migrating away from Guava altogether, so shall we (on the basis that we don't really use too much of guava, this is a good chance for cleanup as well).

@zml2008 has got more to do with migrating our usage of TypeTokens.

### DIFF
--- a/src/main/java/co/aikar/timings/Timings.java
+++ b/src/main/java/co/aikar/timings/Timings.java
@@ -24,11 +24,11 @@
  */
 package co.aikar.timings;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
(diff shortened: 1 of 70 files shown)


Lignium:
What about plugins? If each plugin will add Guava to JAR manually, there will be version conflicts.

gabizou:
> What about plugins? If each plugin will add Guava to JAR manually, there will be version conflicts.

We're not removing guava as part of the API set of libraries, we're just removing our usage of them in the API. There's also an overall issue of "we're stuck with the versions of guava based on the MC implementation" but other implementations (like lantern) are irrespective of Minecraft's version dependencies.

PR #2247, opened 2020-10-09
Replace Duration with Ticks where it makes sense to do so.

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3183)

I'm aware there is #2107 but as I've used the API, I've found that a purely `Duration` based API is cumbersome and confusing to use. From discussion in Discord with @Zidane and @kashike, we've decided to look at introducing the `Ticks` class and disassociating the idea of a minecraft tick from a temporal unit.

[Note I've argued this before but kept an open mind](https://github.com/SpongePowered/SpongeAPI/pull/2007#issuecomment-494107226). Now I've given it a go - I stuggled badly with it and had to look at the common code to work out what was going on.

This came about when I was trying to create a set time command in Nucleus using the `WorldProperties#setDayTime(Duration)` method. It became apparent _very_ quickly that the method itself was confusing. For example:

* What did the `Duration` represent? Days in Minecraft time? Real time? It turned out to be the latter, but nearly everyone will simply want to say "set the time to noon".
* How would you set the time to noon? In our impl, that would mean setting the duration to a couple of minutes or so. That's just for vanilla backed engines - different implementations might have different day cycles.
* There was `TemporalUnits.MINECRAFT_DAY` and `TemporalUnits.HALF_MINECRAFT_DAY` tied to `TemporalUnits.MINECRAFT_TICKS`. Basically - the way you're meant to use this method is to do `setDayTime(Duration.of(18000, TemporalUnits.MINECRAFT_TICKS));` to get midnight - completely defeating the purpose of removing ticks from the API...
* ... but then, you can't use `Duration.get(TemporalUnits.MINECRAFT_TICKS)` when you get a `Duration` back _because the javadocs of [Duration.get](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#get-java.time.temporal.TemporalUnit-) only accepts `SECONDS` and `NANOS`_! So the return value is especially useless because you now need to know what the day length is wall clock time to be able to convert!

**Minecraft is a tick based game** and while I'm a proponent of abstracting as much as we can, I don't think we can abstract this out. If we're still including ticks in some form (`TemporalUnits.MINECRAFT_TICKS`, additions to #2107 add the notion of TPS) then there is no point shoehorning it into Duration because ticks are not a duration. They do not have a reliable clock time - they are units of work. Rarely will they take 50ms, they'll fluctuate wildly above and below this time and we need to stop treating ticks as these fixed time units - just that we expect them to fit **within** a 50ms window. The more important thing about ticks is that they more of a guarantee of order rather than a guarantee of time - if you scheduled task A for 5 ticks and task B for 10 ticks, even if the engine drops the four ticks in the middle to try to recover performance, B will never happen before A (but _may_ happen in the same tick as A - effectively at the same time).

As a result, @kashike suggested the notion of a `Ticks` interface in Discord, to which @Zidane and I agreed with. This way, we don't propose any tick time, and if an implemention does not run on a tick based system, they can emulate what a tick time would be using it. By having a concrete type that replaces `Duration` that indicates some engine dependent time-frame and ordering, we're not muddying the waters and, more importantly, we're not using `Duration` where confusion may occur.

At the end of the day, this API needs to be clear to its consumers - using `Duration` with `TemporalUnits.MINECRAFT_TICKS` (or, as proposed, `Duration.Ticks` when ticks aren't a time unit) was only going to serve to be confusing and, ultimately, give the wrong impression when very few servers are really going to run at 50ms per tick (for vanilla) all the time. If it's going to run for a tick rather than 50ms, we should at least convey that clearly - and if a non-tick based system is built, they can just specifiy a tick to be whatever unit they want it to be and build the `Ticks` interface around that.

(As an aside, for world time, we should perhaps consider some other construct to indicate a specific time of minecraft day for the `get/setDayTime` that might be more user friendly and would allow for disassociation with a tick based time, but that's ~~another idea for another time~~ included too.)

---

* Added Ticks to shield the notion of a tick from consumers
* Removed TemporalUnits
* Renamed Task methods delayTicks and intervalTicks to delay and interval as they now take a Ticks object


src/main/java/org/spongepowered/api/util/Ticks.java
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Returns a {@link Ticks} that would correspond to the supplied number of
+     * in game hours, rounded to the nearest tick.
+     */
+    static Ticks ofMinecraftHours(final double hours) {
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).ofMinecraftDays(hours / 24.0);
+    }
+
+    /**
+     * Gets the expected {@link Duration} of this {@link Ticks} if the server
```
kashike: Not sure if this should mention overloading?

dualspiral: It makes sense to do so in a way as the time will change if there is any overloading - this is the whole problem of why doing this is necessary in the first place. However, I wrangled with the naming of this method for ages and I admit, I'm still not happy with it. Some iterations I went for were:

* `getMinimumAllottedDuration`
* `getMaximumExpectedDuration`
* `getExpectedDuration` (which is what I've got here)

but none of these are particularly clear because really, the point is that a tick isn't really this time unit but people are using it as one.

dualspiral:
I added the notion of a `MinecraftDayTime` after thinking about it overnight/this morning so minecraft times can be disassociated with ticks. It's not a duration, more akin to an Instant, so I haven't done this for weather - if we want to do this sort of thing for weather we'll likely want to make a `MinecraftDayDuration` too.

We may also not want this - if not I'll just throw away the last two commits. Just wanted to proof of concept it.

dualspiral:
@kashike made a good point to me - the client and server may have different tick rates. I think it's probably _unlikely_ that this is going to be the case but it does need to be thought about (and how we logically think about that).

It might just have to be `Ticks<E extends Engine>` and have the `of` method specify an engine...

src/main/java/org/spongepowered/api/util/MinecraftDayTime.java
```diff
@@ -0,0 +1,259 @@
...
+    int minute();
+
+    /**
+     * Adds the time given by the provided {@link Ticks} object and returns a
+     * new object with the result.
+     *
+     * @param ticks The {@link Ticks} to add.
+     * @return A new {@link MinecraftDayTime}
+     */
+    MinecraftDayTime plus(final Ticks ticks);
```
Zidane, line 166: No need for finals there (in a blueprint)

Zidane:
Manually merged.

PR #2250, opened 2020-10-20
API8 DataStores

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3192)

DataStores can now be built for
 - custom data (saved under "ForgeData.SpongeData.CustomManipulators") which is what plugins should use
or
- vanilla data (saved on the root dataview corresponding the base CompoundNBT) which is used by Sponge to set data on Archetypes and Snapshots

The DataStoreBuilder features Steps forcing you to call required builder methods at least once.

Adds convenience methods for building very simple dataregistration & datastores.
A dataregistration for a `String` value on an `ItemStack` is now as simple as:
```java
        this.mySimpleDataKey = Key.of(this.plugin, "mysimpledata", TypeTokens.STRING_VALUE_TOKEN);
        event.register(DataRegistration.of(this.mySimpleDataKey, ItemStack.class));
```

### DIFF
--- a/src/main/java/org/spongepowered/api/data/DataRegistration.java
+++ b/src/main/java/org/spongepowered/api/data/DataRegistration.java
@@ -27,12 +27,14 @@ package org.spongepowered.api.data;
 import com.google.common.reflect.TypeToken;
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.Sponge;
+import org.spongepowered.api.data.persistence.DataQuery;
 import org.spongepowered.api.data.persistence.DataStore;
 import org.spongepowered.api.data.value.Value;
 import org.spongepowered.api.data.value.ValueContainer;
 import org.spongepowered.api.util.CatalogBuilder;
 import org.spongepowered.plugin.PluginContainer;
 
+import java.util.Collection;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/data/DataRegistration.java
```diff
@@ -118,6 +120,21 @@ static Builder builder() {
...
+     * Creates a DataRegistration for a single key with a DataStore for given data-holders.
+     *
+     *
+     * @param key the data key
+     * @param dataHolders the data-holders
+     * @param <T> the keys value type
+     *
+     * @return The built data registration
+     */
+    static <T> DataRegistration of(Key<Value<T>> key, Class<? extends DataHolder>... dataHolders) {
```
gabizou: Are you intending to be able to call `DataRegistration.of(Keys.FOO)`? because the varargs can have a length of `0` by omitting the argument altogether in this case (this is why Cause historically always had an `Object, Object...` method signature, to ensure there's at least one argument passed in.

src/main/java/org/spongepowered/api/data/persistence/DataStore.java
```diff
@@ -115,18 +117,36 @@ default DataView serialize(DataManipulator dataManipulator) {
...
+    static <T> DataStore of(Key<Value<T>> key, DataQuery dataQuery, TypeToken<? extends DataHolder>... typeTokens) {
+        return builder().pluginData(key.getKey()).holder(typeTokens).key(key, dataQuery).build();
+    }
+
+    /**
+     * Provides a {@link DataStore} for a single {@link Key}.
+     * <p>
+     *     Note that default deserializers do not support {@link Collection}, {@link Map} or Array types!
+     *     Use {@link Builder.SerializersStep#key(Key, BiConsumer, Function)} for these.
+     * </p>
```
gabizou, line 139: Much like the comment above, may want to mention that the varargs of types is optional to specify what types the datastore will support?

gabizou (approved):
Aside from the varargs, the changes look good :)

PR #2252, opened 2020-11-10
Ray Trace API

dualspiral:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3197)

Needs to be tested, but otherwise heralds the return of ray tracing.


src/main/java/org/spongepowered/api/util/blockray/RayTrace.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /**
+     * Returns a {@link Predicate} for use in
+     * {@link #continueWhileLocation(Predicate)} that only allows a trace to
+     * continue if the {@link LocatableBlock} at the given position is air.
+     *
+     * @return The predicate
+     */
+    static Predicate<ServerLocation> onlyAir() {
```
Zidane: The switch of returns (this method then next method) are a hair confusing here. What is the reasoning for that?

dualspiral: Actually, this is a good catch in itself as this really should be `LocatableBlock` for the purpose that I was intending to use it for.

The intention behind this method and the next one (which should be `notAir`) is to create a ray that selects the first non-air block. For that to work given the javadocs, this would need to be set in `continueWhileBlock`, while `select` would then take a `notAir()` `Predicate<LocatableBlock>` - that is, continue the ray until we hit a non air block, selecting the non air block (`select` occurs before `continueWhileBlock`). There may also be another helper method for stopping a ray if an `Entity` is hit, though that is as simple as calling `continueWhileEntity(r -> false);`.

My development of this went through a few iterations and this looks like it was left behind. These will get adjusted.

src/main/java/org/spongepowered/api/util/blockray/RayTrace.java
```diff
@@ -0,0 +1,339 @@
...
+        return Sponge.getRegistry().getFactoryRegistry().provideFactory(Factory.class).selectNonAirBlock();
+    }
+
+    /**
+     * Sets the {@link ServerWorld} to perform the ray trace in.
+     *
+     * @param serverWorld The {@link ServerWorld}
+     * @return This, for chaining
+     */
+    RayTrace<T> world(ServerWorld serverWorld);
```
Zidane, line 119: Would there be a reason for ray tracing on the client thread at all?

dualspiral: Quite possibly for if you want to do some client rendering stuff. It would probably require me to split the ray trace up into a `Client` and `Server` component as specifying a world is (probably) unneeded on the client, for example - it certainly doesn't have a notion of a world tied to a key at the very least.

Worth doing, do you think, even if I just lay the groundwork to enable writing a client based version later on down the road?

dualspiral:
This and the corresponding implementation works. Two questions remain:

* Do we split this up slightly to make it possible to have a client version without breaking this?
* Do we rename this as `RayCast` (like always, I don't really care what we call it but I'd rather we choose what is easy to understand for the consumers and `RayTrace` is used in MCP, even though this is mostly independent from it.)

Zidane:
Forget the client for now, I can't justify it.

RayCast would be the more correct term I guess but most people just always treat it as RayTrace. I don't care either way.

Zidane:
Manually merged.

PR #2260, opened 2020-11-22
Api8/map api

tyhdefu:
SpongeAPI | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3208)

SpongeAPI of the map api. I think this is mostly complete. However lots of javadocs need work/completing.


tyhdefu:
Thinking about it now - I think there should be more information provided about MapBanners. In vanilla you can right click on a banner in the world to add it to a map. (With the api you can also add map banner decorations that are not attached to any in-world blocks). However the only access to use currently is `toggleBanner(...)` (or something similar to that) on MapInfo. This should probably be able to be accessed via a Key, and be a Collection<MapBanner> or something similar to that. This allows access to see all mapbanners, whether a banner is added etc, etc. I will try to work on adding this ASAP.

Faithcaio:
Merged in https://github.com/SpongePowered/SpongeAPI/commit/081d45961756316dfc6bb8deb55cdfc04d2fdd41
Feel free to make a new PR for additional API.

PR #2263, opened 2020-11-26
1.16.4

kashike:
https://github.com/SpongePowered/SpongeCommon/pull/3211

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/sound/music/MusicDiscs.java
+++ b/src/main/java/org/spongepowered/api/effect/sound/music/MusicDiscs.java
@@ -38,19 +38,21 @@ public final class MusicDiscs {
 
     public static final Supplier<MusicDisc> CHIRP = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "chirp");
 
-    public static final Supplier<MusicDisc> ELEVEN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "eleven");
-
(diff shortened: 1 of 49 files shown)


src/main/java/org/spongepowered/api/fluid/FluidTypes.java
```diff
@@ -31,14 +31,17 @@
 public final class FluidTypes {
```
kashike, line 31: Fluid stuff in the API is a mess. What even is going on?

src/main/java/org/spongepowered/api/effect/sound/music/MusicDiscs.java
```diff
@@ -38,20 +38,22 @@
...
 
-    public static final Supplier<MusicDisc> ELEVEN = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "eleven");
-
     public static final Supplier<MusicDisc> FAR = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "far");
 
     public static final Supplier<MusicDisc> MALL = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "mall");
 
     public static final Supplier<MusicDisc> MELLOHI = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "mellohi");
 
+    public static final Supplier<MusicDisc> MUSIC_DISC_11 = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MusicDisc.class, "11");
```
kashike, line 47: This class is weird too - Minecraft does not expose music disc types, but we do, and so we have this weirdness. The official names of these are "11" and "13", not "eleven" and "thirteen".

PR #2264, opened 2020-11-28
Expose portal agents

thibaulthenry:
**[SpongeCommon ](https://github.com/SpongePowered/SpongeCommon/pull/3212) | SpongeAPI**

In this PR I'm trying to expose the management of portal agents directly from the API.

See the PR in SpongeCommon for more information about implementation.

Thanks for reading this PR.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
+++ b/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
@@ -227,16 +227,15 @@ public interface WorldProperties extends DataSerializable {
      *
      * @return The portal agent map of this world
      */
-    Map<String, String> getPortalAgents();
+    Map<PortalType, String> getPortalAgents();
 
     /**
-     * Set the destination world for a {@link PortalType}
-     * on this world.
+     * Get the destination world properties for a {@link PortalType},
+     * if known.
      *
      * @param portalType The type of the portal
-     * @return The name of the destination world
      */
-    String getPortalDestination(PortalType portalType);
+    Optional<WorldProperties> getPortalDestination(PortalType portalType);
 
     /**
      * Set the destination world for a {@link PortalType}
--- a/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
+++ b/src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
@@ -36,6 +36,7 @@ import org.spongepowered.api.entity.living.player.gamemode.GameMode;
 import org.spongepowered.api.world.DimensionType;
 import org.spongepowered.api.world.GeneratorType;
 import org.spongepowered.api.world.PortalAgentType;
+import org.spongepowered.api.world.PortalType;
 import org.spongepowered.api.world.SerializationBehavior;
 import org.spongepowered.api.world.World;
 import org.spongepowered.api.world.difficulty.Difficulty;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -219,6 +220,33 @@
...
     PortalAgentType getPortalAgentType();
 
+    /**
+     * Gets the portal agent map of this world.
+     * Keys represent dimensions names.
+     * Values represent destination world names.
+     *
+     * @return The portal agent map of this world
+     */
+    Map<String, String> getPortalAgents();
```
dualspiral: I'm gonna be blunt here, I know what's been said in the past, but I _do not_ like this being a `Map<String, String>`, when the rest of the API tries to use `World` and `WorldProperties` as much as possible.

Don't get me wrong, I understand the reasoning. The idea is that it mirrors the config file, which might have invalid entries that you don't wish to lose. However, the rest of the API doesn't expose the config as strings, so we shouldn't here.

In my view, this should be a `Map<WorldProperties, PortalType>`, or something to that effect, dynamically gemerated when requested. Invalid entries should just be ignored, this should be a representation of what the running server will act upon, and **not** what is in the config file.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -219,6 +220,33 @@
...
+
+    /**
+     * Set the destination world for a {@link PortalType}
+     * on this world.
+     *
+     * @param portalType The type of the portal
+     * @return The name of the destination world
+     */
+    String getPortalDestination(PortalType portalType);
```
dualspiral: I think this should be an `Optional<WorldProperties>`, again mimicing what the server sees.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -219,6 +220,33 @@
...
+
+    /**
+     * Set the destination world for a {@link PortalType}
+     * on this world.
+     *
+     * @param portalType The type of the portal
+     * @param destinationWorldName The destination world name of the portal
+     */
+    void setPortalDestination(PortalType portalType, String destinationWorldName);
```
dualspiral, line 247: Again, with my above comments. I think this should be a `WorldProperties` destination world, which is fixed up by the implementation.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
Zidane: @dualspiral Completely fair points but it is dangerous to hand back a map reference with WorldProperties in it as stale references can be destructive.

Maybe a Map<PortalType, String> as a compromise?

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
Zidane: Agreed.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
Zidane: This one can still be a String else you would query the server for a world just to set a string and then be done.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
dualspiral: My original thought is that we'd use a `WeakHashMap`... but then I look at the javadocs and what it really should say is the keys are the portal types, not the dimensions. 

So, sure.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
Zidane: Also, it SHOULD allows String values for unknown worlds currently.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
thibaulthenry: You talked about unknown worlds in another conversation
It's not possible to retrieve them anymore if I return Optional<WorldProperties> instead

dualspiral: As API 7 basically knows about all it's loaded and unloaded worlds, that's not really an issue - it simply infers that the portal agent doesn't have a valid target in this context.

The point I was making about these methods it shouldn't mirror the config, but mirror what the server will do/knows.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
thibaulthenry: It's not clear here. Do we have a compromise on String or WorldProperties as parameter for destination ?

dualspiral:
Unfortunately, I'm going to close this. We're done with API changes on API 7 and, partially my fault, we never got to finishing this. API 8 does not have the notion of portal agents, so won't be needed there, but thank you for this regardless.

PR #2265, opened 2020-11-28
Add keys for data from client settings

zml2008:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3213)

First time implementing data in the API8 scheme, so putting this up for some review

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
@@ -33,6 +33,7 @@ import org.spongepowered.api.advancement.AdvancementProgress;
 import org.spongepowered.api.advancement.AdvancementTree;
 import org.spongepowered.api.data.Keys;
 import org.spongepowered.api.data.type.SkinPart;
+import org.spongepowered.api.data.value.SetValue;
 import org.spongepowered.api.data.value.Value;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
```diff
@@ -128,29 +128,37 @@ default boolean isViewingInventory() {
...
+    }
 
     /**
      * Sets the view distance setting of the player. This value represents the
      * radius (around the player) in unit chunks.
      *
      * @param distance The player's view distance
      */
-    void setViewDistance(int distance);
+    default void setViewDistance(final int distance) {
```
NickImpact: I see you were just updating this, but I also see that this view distance key is documented as read only. So I would imagine we shouldn’t be able to set a player’s view distance here as well.

zml2008: yeah -- this api method was unimplemented, but I preserved it since it existed, and documented what's currently easily implementable. Can update either way if I get confirmation on what'll be implemented.

Zidane: @zml2008 Additionally, data methods follow a certain style. Look at `Entity` to see.

PR #2266, opened 2020-11-28
Expose entity tags

thibaulthenry:
**[SpongeCommon ](https://github.com/SpongePowered/SpongeCommon/pull/3214)| SpongeAPI**

Hey,

Here is a small PR trying to expose entity tags which can be used as selector arguments in commands.
Until now they were not exposed in the API.

Learn more about tags: https://minecraft.gamepedia.com/Scoreboard#Tags

Thanks for reading this PR.


thibaulthenry:
Also needs an implementation in API-8

dualspiral:
Manually merged, thanks!

PR #2268, opened 2020-11-30
ResourceKey improvements

ItsDoot:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/3217)

- Cleaned up and added more javadocs
- Added new `resolve` methods that let you specify a custom default namespace

EDIT: Also keep in mind now that `ResourceKey.Builder#namespace` must always be set.

### DIFF
--- a/src/main/java/org/spongepowered/api/ResourceKey.java
+++ b/src/main/java/org/spongepowered/api/ResourceKey.java
@@ -102,7 +102,7 @@ public interface ResourceKey extends Key {
      * @return A new resource key
      */
     static ResourceKey of(final String namespace, final String value) {
-        return builder().namespace(namespace).value(value).build();
+        return ResourceKey.builder().namespace(namespace).value(value).build();
     }
 
     /**
@@ -113,7 +113,7 @@ public interface ResourceKey extends Key {
      * @return A new resource key
      */
     static ResourceKey of(final PluginContainer container, final String value) {
-        return builder().namespace(container).value(value).build();
+        return ResourceKey.builder().namespace(container).value(value).build();
     }
 
     /**
@@ -126,7 +126,7 @@ public interface ResourceKey extends Key {
      * @return A new resource key
      */
     static ResourceKey resolve(final String formatted) {
-        return resolve(formatted, MINECRAFT_NAMESPACE);
+        return ResourceKey.resolve(formatted, MINECRAFT_NAMESPACE);
     }
 
     /**
@@ -140,7 +140,7 @@ public interface ResourceKey extends Key {
      * @return A new resource key
      */
     static ResourceKey resolve(final String formatted, final PluginContainer defaultNamespace) {
-        return resolve(formatted, defaultNamespace.getMetadata().getId());
+        return ResourceKey.resolve(formatted, defaultNamespace.getMetadata().getId());
     }
 
     /**
--- a/src/main/java/org/spongepowered/api/ResourceKey.java
+++ b/src/main/java/org/spongepowered/api/ResourceKey.java
@@ -64,20 +64,20 @@ public interface ResourceKey extends Key {
     String SPONGE_NAMESPACE = "sponge";
 
     /**
-     * Creates a catalog key with a namespace of {@link #MINECRAFT_NAMESPACE minecraft}.
+     * Creates a resource key with a namespace of {@link #MINECRAFT_NAMESPACE minecraft}.
      *
      * @param value The value
-     * @return A new catalog key
+     * @return A new resource key
      */
     static ResourceKey minecraft(final String value) {
         return ResourceKey.of(ResourceKey.MINECRAFT_NAMESPACE, value);
     }
 
     /**
-     * Creates a catalog key with a namespace of {@link #SPONGE_NAMESPACE sponge}.
+     * Creates a resource key with a namespace of {@link #SPONGE_NAMESPACE sponge}.
      *
      * @param value The value
-     * @return A new catalog key
+     * @return A new resource key
      */
     static ResourceKey sponge(final String value) {
         return ResourceKey.of(ResourceKey.SPONGE_NAMESPACE, value);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/ResourceKey.java
```diff
@@ -95,38 +95,66 @@ static Builder builder() {
...
+     * Resolves a resource key from a string.
+     *
+     * <p>If no namespace is found in {@code formatted} then
+     * the specified default namespace will be used.</p>
+     *
+     * @param formatted The formatted string to parse
+     * @param defaultNamespace The default namespace to use
+     * @return A new resource key
+     */
+    static ResourceKey resolve(final String formatted, final PluginContainer defaultNamespace) {
```
dualspiral: I don't agree with this. Either you have the namespace in the formatted string (use `resolve(String)`) or you don't (use `resolve(String namespace, String value)` or `resolve(PluginContainer container, String value)`).

This method is just going to encourage lazy programming. Unless there is a compelling reason, this method should not exist.

src/main/java/org/spongepowered/api/ResourceKey.java
```diff
@@ -95,38 +95,66 @@ static Builder builder() {
...
+     * Resolves a resource key from a string.
+     *
+     * <p>If no namespace is found in {@code formatted} then
+     * the specified default namespace will be used.</p>
+     *
+     * @param formatted The formatted string to parse
+     * @param defaultNamespace The default namespace to use
+     * @return A new resource key
+     */
+    static ResourceKey resolve(final String formatted, final String defaultNamespace) {
```
dualspiral: As above.

src/main/java/org/spongepowered/api/ResourceKey.java
```diff
@@ -95,38 +95,66 @@ static Builder builder() {
...
+     * Creates a resource key.
      *
      * @param namespace The namespace
      * @param value The value
-     * @return A new catalog key
+     * @return A new resource key
      */
     static ResourceKey of(final String namespace, final String value) {
-        return Sponge.getRegistry().getBuilderRegistry().provideBuilder(Builder.class).namespace(namespace).value(value).build();
+        return builder().namespace(namespace).value(value).build();
```
dualspiral: Prefix static methods with the class name, even if it is the same class.

src/main/java/org/spongepowered/api/ResourceKey.java
kashike: I agree - this should not exist.

ItsDoot: These resolve methods would exist to make user input parsing easier.

For example, my Bending plugin has abilities that players can equip via a command: `/bending bind <some-ability>`.
With a usage being something like `/bending bind bending:air_blast`.

It would be nice if players could type plainly `/bending bind air_blast`, instead, without the namespace.
If I were to use the current existing `resolve` method, I could only get `minecraft:air_blast`, which is incorrect.
Therefore, these new resolve methods would make it a single line to be able to parse both `bending:air_blast` and `air_blast` correctly.

Zidane: @dualspiral @kashike 

I respectfully agree with @ItsDoot on the exact point he made.

In preparing for state of sponge, all the catalogs added originated from the Royale plugin. I can't tell you how annoying it was to type out the namespaces constantly, even with tab complete.

Maybe it isn't realistic to do it due to how the command system functions but it would sure be nice to parse a value string as something not always `minecraft`

kashike: I definitely don't like `resolve` as the name for this method, then.

zml2008: Yeah, imo this functionality is most useful for commands, and it would work but client completions would be tricky without falling back to requesting from the server.

If you aren't doing this in commands, you might as well just have a

```java
@Plugin
public class MyPlugin {
    // [...]
    public static ResourceKey key(final String value) {
        return ResourceKey.of(ProjectData.ID, value);
    }
    // [...]
}
```

style method in your plugin class -- that's what I use on a few platforms.

dualspiral: > I can't tell you how annoying it was to type out the namespaces constantly, even with tab complete.

[I actually have a way to default the namespace for catalogs for this exact use case.](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java#L183-L227).

dualspiral: [I have made it easier to specify a default namespace](https://github.com/SpongePowered/SpongeAPI/commit/4172d6b19cd1cc7dcca1bbbcae24e22e919e58fb) and re-added completions for non-namespaces values. That covers the use case here and is the right place for it to go - this sort of fuzzy-ish matching for commands belongs on a command parser. It was always intended that commands could forgo default namespaces. I agree, and always have done, that you should be able to specify default namespaces for commands - that has always existed. 

This method should not exist on ResourceKey itself, and I stand firm on that. As it turned out, this PR is an XY problem, the solution was presented without the problem it was trying to solve. This has now been solved in the correct way.

(zml: It already went to the server due to custom registries that the client doesn't know about, so it was trivial to fix)

zml2008: Ah perfect -- then I'm with dual here, the style tweaks are good but let's not add the custom default namespace methods here.

Zidane:
My concerns are now resolved as well

src/main/java/org/spongepowered/api/ResourceKey.java
kashike: :+1: with @dualspiral's comment: `This method should not exist on ResourceKey itself`

kashike (changes requested):

ItsDoot:
@dualspiral @kashike Removed the other resolve methods.

IMO, we should still keep the factory even though its a single method, because using the builder for resolving is a weird special case.

kashike (approved):

PR #2270, opened 2020-12-04
[WIP] Just brainstorming ideas for further registry work.

Zidane:
Perks of the changes, so far:

- Types go back to being enums (as long as they do not have generic parameters) which brings back the ease of use to plugin devs
- Suppliers now available for all registries and under any key, not just for Sponge
- Dynamic registration now available after lifecycle

Much more to come, seeking feedback before applying these changes on a wide scale.

### DIFF
--- a/src/main/java/org/spongepowered/api/registry/Registry.java
+++ b/src/main/java/org/spongepowered/api/registry/Registry.java
@@ -22,24 +22,46 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.data.type;
+package org.spongepowered.api.registry;
 
 import org.spongepowered.api.ResourceKey;
-import org.spongepowered.api.registry.Registries;
 
-import java.util.Locale;
+import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
-public enum MatterTypes implements Supplier<MatterType> {
+public interface Registry<V> extends Iterable<RegistryEntry<V>> {
 
-    GAS,
-    LIQUID,
-    SOLID;
+    ResourceKey getKey();
 
-    private final Supplier<MatterType> supplier = Registries.MATTER_TYPE.getSupplier(ResourceKey.sponge(this.name().toLowerCase(Locale.ROOT)));
+    Optional<ResourceKey> getKey(V value);
 
-    @Override
-    public MatterType get() {
-        return this.supplier.get();
-    }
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/registry/Registry.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public interface Registry<V> extends Iterable<RegistryEntry<V>> {
+
+    ResourceKey getKey();
+
+    Optional<ResourceKey> getKey(V value);
+
+    Optional<RegistryEntry<V>> getValue(ResourceKey key);
```
ItsDoot, line 39: Alternatively:
```java
     Optional<V> getValue(ResourceKey key);

     Optional<RegistryEntry<V>> getEntry(ResourceKey key);
```

src/main/java/org/spongepowered/api/data/type/MatterTypes.java
```diff
@@ -24,23 +24,22 @@
...
-public final class MatterTypes {
+public enum MatterTypes implements Supplier<MatterType> {
 
-    // SORTFIELDS:ON
+    GAS,
+    LIQUID,
+    SOLID;
 
-    public static final Supplier<MatterType> GAS = Sponge.getRegistry().getCatalogRegistry().provideSupplier(MatterType.class, "gas");
+    private final Supplier<MatterType> supplier = Registries.MATTER_TYPE.getSupplier(ResourceKey.sponge(this.name().toLowerCase(Locale.ROOT)));
```
kashike, line 39: This approach works here, but would not work with `SoundTypes` for example - I think it would be much better to pass the key into the constructor:
```java
GAS(ResourceKey.sponge("gas"))
```
Especially since some need special casing as well.

src/main/java/org/spongepowered/api/registry/RegistryManager.java
```diff
@@ -0,0 +1,44 @@
...
+     * <p>If no registry is found, this will throw an {@link IllegalStateException}.
+     * Mostly provided for cleaner access in various API classes and when plugin
+     * developers are 100% sure the registry will always exist. See {@link Registries}
+     * for an example.</p>
+     *
+     * @param key The key
+     * @param <V> The value type
+     * @return The registry
+     */
+    <V> Registry<V> findRegistry(ResourceKey key);
```
kashike, line 43:
```suggestion
    <V> Registry<V> requireRegistry(ResourceKey key);
```

kashike (changes requested):

PR #2272, opened 2020-12-05
Update Chunk API elements

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3222)

I took a look at updating the Chunk API (stepping out of comfort zone!) from invalid and... it's been more of a journey than I thought. This is not near being done but I haven't really touched this area of the code before so I want to make sure I'm doing the right things.

The biggest thing (at the moment!) is that I've changed `HeightTypes` and `ChunkState` to mirror the MCP entries. There will likely need to be some rejigging of the ProtoChunk and PrimitiveChunk interfaces. I also don't think `loadChunk` and `unloadChunk` should be on this interface - that should be on the associated `World`.

See the associated Common PR for the implementation.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/chunk/ProtoChunk.java
+++ b/src/main/java/org/spongepowered/api/world/chunk/ProtoChunk.java
@@ -24,11 +24,12 @@
  */
 package org.spongepowered.api.world.chunk;
 
-import org.spongepowered.api.Game;
-import org.spongepowered.api.Server;
 import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.util.Ticks;
+import org.spongepowered.api.util.annotation.DoNotStore;
 import org.spongepowered.api.world.ProtoWorld;
 import org.spongepowered.api.world.World;
+import org.spongepowered.api.world.generation.GenerationRegion;
 import org.spongepowered.api.world.volume.biome.BiomeVolume;
 import org.spongepowered.api.world.volume.block.BlockVolume;
 import org.spongepowered.api.world.volume.block.entity.BlockEntityVolume;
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/api/world/chunk/ProtoChunk.java
```diff
@@ -45,7 +45,11 @@
  * chunk loading from storage, or empty chunks on clients. Traditionally, a
  * usable "live" proto chunk instance will be a {@link Chunk} with a valid
  * {@link World} instance.</p>
+ *
+ * <p>A chunk may not be attached to a world if it is in the process of being
+ * generated.</p>
```
gabizou:
```suggestion
 * <p>A chunk may not be attached to a {@link World}, or a {@link org.spongepowered.api.world.generation.GenerationRegion}  if it is in the process of being
 * generated.</p>
```

src/main/java/org/spongepowered/api/world/chunk/ProtoChunk.java
```diff
@@ -98,49 +104,22 @@
...
-     * <p>It can be inferred however, that if {@link #state()} returns
-     * {@link ChunkStates#WORLD_READY}, the {@link ProtoWorld} would be a
-     * {@link World} instance. Inversely, if {@link #state()} returns
-     * {@link ChunkStates#EMPTY}, the {@link ProtoWorld} would not be a
-     * valid {@link World} object.</p>
-     *
-     * @return The parented world
+     * @return The number of ticks
      */
-    ProtoWorld<?> world();
```
gabizou, line 114: What happened to the reference to the world?

gabizou: Why shouldn't we be able to soft guarantee that any chunk has a world? Or was it that generating chunks may not actually have the links to the generation regions?

dualspiral: Generating chunks (ProtoChunks) have no link to thier target world or anything that links to something like a `GenerationRegion`.

PR #2274, opened 2020-12-06
Start of a Chunk Manager

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3228)

Creates a `ChunkManager`, currently just fronts chunk loading tickets.

### DIFF
--- a/src/main/java/org/spongepowered/api/registry/RegistryTypes.java
+++ b/src/main/java/org/spongepowered/api/registry/RegistryTypes.java
@@ -145,6 +145,7 @@ import org.spongepowered.api.world.gamerule.GameRule;
 import org.spongepowered.api.world.generation.structure.Structure;
 import org.spongepowered.api.world.portal.PortalType;
 import org.spongepowered.api.world.schematic.PaletteType;
+import org.spongepowered.api.world.server.TicketType;
 import org.spongepowered.api.world.teleport.TeleportHelperFilter;
 import org.spongepowered.api.world.weather.WeatherType;
 
@@ -349,6 +350,8 @@ public final class RegistryTypes {
 
     public static final DefaultedRegistryType<PortionType> PORTION_TYPE = RegistryTypes.spongeKeyInGame("portion_type");
 
+    public static final DefaultedRegistryType<QueryType> QUERY_TYPE = RegistryTypes.spongeKeyInGame("query_type");
+
     public static final DefaultedRegistryType<RabbitType> RABBIT_TYPE = RegistryTypes.spongeKeyInGame("rabbit_type");
 
     public static final DefaultedRegistryType<RaidStatus> RAID_STATUS = RegistryTypes.spongeKeyInGame("raid_status");
@@ -379,13 +382,13 @@ public final class RegistryTypes {
 
     public static final DefaultedRegistryType<TeleportHelperFilter> TELEPORT_HELPER_FILTER = RegistryTypes.spongeKeyInGame("teleport_helper_filter");
 
+    public static final DefaultedRegistryType<TicketType<?>> TICKET_TYPE = RegistryTypes.spongeKeyInGame("ticket_type");
+
     public static final DefaultedRegistryType<TransactionType> TRANSACTION_TYPE = RegistryTypes.spongeKeyInGame("transaction_type");
 
     public static final DefaultedRegistryType<Trigger<?>> TRIGGER = RegistryTypes.spongeKeyInGame("trigger");
 
     public static final DefaultedRegistryType<TropicalFishShape> TROPICAL_FISH_SHAPE = RegistryTypes.spongeKeyInGame("tropical_fish_shape");
-
-    public static final DefaultedRegistryType<QueryType> QUERY_TYPE = RegistryTypes.spongeKeyInGame("query_type");
 
     public static final DefaultedRegistryType<Visibility> VISIBILITY = RegistryTypes.spongeKeyInGame("visibility");
 
--- a/src/main/java/org/spongepowered/api/world/server/ChunkManager.java
+++ b/src/main/java/org/spongepowered/api/world/server/ChunkManager.java
@@ -1,0 +1,171 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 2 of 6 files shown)


src/main/java/org/spongepowered/api/world/server/TicketType.java
```diff
@@ -0,0 +1,49 @@
...
+@CatalogedBy(TicketTypes.class)
+public interface TicketType<T> extends CatalogType {
+
+    /**
+     * Gets the lifetime of any {@link Ticket tickets} of this type.
+     *
+     * @return The number of {@link Ticks} any {@link Ticket tickets} of this
+     *         type will be valid for.
+     */
+    Ticks getTicketLifetime();
```
Zidane: `getLifetime()`?

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.spongepowered.api.world.chunk.Chunk;
+import org.spongepowered.math.vector.Vector3i;
+
+import java.util.Optional;
+
+/**
+ * Manages {@link Chunk chunks} for a {@link ServerWorld}.
+ */
+public interface ChunkManager {
+
```
Zidane, line 41: `ServerWorld getWorld()`?

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param ticket The ticket to check.
+     * @return true if so
+     */
+    boolean isValid(Ticket<?> ticket);
+
+    /**
+     * Gets the default lifetime, in {@link Ticks}, of any ticket requested via
+     * this ticket manager.
+     */
+    Ticks getDefaultTicketLifetime(TicketType<?> ticketType);
```
Zidane: `getDefaultLifetime()`

src/main/java/org/spongepowered/api/world/server/TicketTypes.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Types of {@link Ticket tickets} that can be requested via the
+ * {@link ChunkManager}.
+ */
+public class TicketTypes {
+
+    /**
+     * Represents {@link Ticket tickets} that forces a chunk to stay loaded.
```
Zidane: that force chunk(s) ...

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Request a {@link Ticket} for the given {@link TicketType}.
+     *
+     * @param type The type of ticket to request.
+     * @param value The value to register the ticket with.
+     * @param <T> The type of the supplied {@code value}.
+     * @return The ticket, if granted.
+     */
+    <T> Optional<Ticket<T>> requestTicket(TicketType<T> type, Vector3i chunkPosition, T value);
```
Zidane: Radius?

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Releases the provided {@link Ticket}, allowing the chunk position
+     * represented by the given ticket to be unloaded (if it is not being kept
+     * loaded by other means).
+     *
+     * @param ticket The ticket to release.
+     */
+    boolean releaseTicket(Ticket<?> ticket);
+
+}
```
Zidane, line 171: `<T> Collection<Ticket<T>> getTickets(TicketType<T> type)` ?

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
```diff
@@ -203,4 +203,12 @@ default Chunk getChunkAtBlock(int bx, int by, int bz) {
      * @return The raid at that location, if present
      */
     Optional<Raid> getRaidAt(Vector3i blockPosition);
+
+    /**
+     * Gets the {@link ChunkManager} for this world.
+     *
+     * @return The ticket manager.
```
Zidane: the chunk manager

src/main/java/org/spongepowered/api/world/server/Ticket.java
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.server;
+
+/**
+ * Represents a chunk loading ticket.
+ *
+ * @param <T> The type of value this ticket holds.
+ */
+public interface Ticket<T> {
```
Zidane, line 32: `TicketType<T> getType()`

src/main/java/org/spongepowered/api/world/server/TicketTypes.java
```diff
@@ -0,0 +1,43 @@
...
+import org.spongepowered.api.Sponge;
+import org.spongepowered.math.vector.Vector3i;
+
+import java.util.function.Supplier;
+
+/**
+ * Types of {@link Ticket tickets} that can be requested via the
+ * {@link ChunkManager}.
+ */
+public class TicketTypes {
```
Zidane: `final`

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
Zidane: Additional, the chunkPosition really is the chunkOrigin that the radius operates off of.

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
dualspiral: Actually just removed this, it's on the `TicketType` instead.

pop4959:
Zidane asked me to look over this, and I'll say that I think it looks pretty nice so far.

However, there doesn't seem to be any way for a plugin to be notified as to when the chunk(s) for a particular ticket have been loaded, at a given chunk state (if not just loaded vs generated). I think this would be a worthwhile addition as many plugins will not care about the tickets themselves, but really just want access to chunks. In API 7 this would have been possible using something such as World#loadChunkAsync since it returns a future (and yes, I'm aware that API was never really functional, but it gets the point I'm trying to make across).

src/main/java/org/spongepowered/api/registry/RegistryTypes.java
```diff
@@ -367,6 +368,8 @@
 
     public static final RegistryType<TeleportHelperFilter> TELEPORT_HELPER_FILTER = RegistryTypes.spongeKey("teleport_helper_filter");
 
+    public static final RegistryType<TicketType<?>> TICKET_TYPE = RegistryTypes.spongeKey("ticket_types");
```
Lignium: ticket_types -> ticket_type

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -0,0 +1,108 @@
...
+     * Checks if the provided {@link Ticket} is valid for the world this manager
+     * represents.
+     *
+     * @param ticket The ticket to check.
+     * @return true if so
+     */
+    boolean isValid(Ticket<?> ticket);
+
+    /**
+     * Gets the {@link Ticks} remaining on the supplied ticket.
```
gabizou, line 59: What's the behavior of a ticket that isn't valid?

dualspiral: Doesn't do anything. The distance manager has exponged all knowledge of it so the chunks it refers to may unload.

dualspiral:
For anyone who is even _thinking_ about merging, please don't. While I think I've gotten most of the issues I'm still in the middle of testing, chunk tickets are a little weird.

> However, there doesn't seem to be any way for a plugin to be notified as to when the chunk(s) for a particular ticket have been loaded, at a given chunk state (if not just loaded vs generated). I think this would be a worthwhile addition as many plugins will not care about the tickets themselves, but really just want access to chunks. In API 7 this would have been possible using something such as World#loadChunkAsync since it returns a future (and yes, I'm aware that API was never really functional, but it gets the point I'm trying to make across).

I've actually spent a lot of time looking at this and it's not going to happen, at least short term. The problem is that the implementation doesn't really hold a link between a ticket and the chunk loading lifecycle so much - the ticket and actual action are so far removed from each other that a ticket is nothing more than a level to Minecraft - there is no "completion" notification.

The best I'll be able to do (and I'm not doing it as part of this) will be to just have our tickets listen for a chunk load and notify based on that - but there is no guarantee that a chunk load would happen as part of a ticket.

Getting the tickets working with the manager as we expect is the goal for this PR. There is no reason why such a mechanism, should one be found, couldn't be added in a non-breaking change manner to a future revision of API 8 if it misses the original cut.

pop4959:
Thanks for taking a look. I agree it probably doesn't make too much sense as part of the ticket system for the reasons you've just described. The original idea I had was that a loadChunk method returning a future could be re-added in API 8 (as mentioned in #2287). Implementations of this already exist on other platforms, though another mechanism could also be used. Perhaps after some further discussion this can be considered for a revision of API 8 (since the freeze is approaching).

PR #2275, opened 2020-12-06
Some Text helpers for discoverability

kashike:
@Faithcaio @Zidane Moving this commit to a PR, as per @gabizou.

PR will be updated as we handle things with Adventure.


zml2008:
For initial discussion, see 4e0ec9b

dualspiral:
Buckle up, for this is going to be a long one. You are free to turn parts of these into Adventure issues as you see fit. None of the following is meant to belittle what Adventure is, but to point out some of my major misgivings with it.

This commit (and subsequent fallout) is borne out of one thing: ease of use and discoverability of the Adventure library when used with Sponge. While I'm going to try to approach my comments from a Sponge point of view, these are **my** comments, **my** opinions. You are free to disagree with them, but I must be clear that you would be disagreeing with me, not the Sponge team. Disagree with them here too - this is something that everyone will benefit from - a private discussion helps no-one.

With that in mind, it's worth keeping in mind my preference for style, which _do_ conflict with some of the design decisions of Adventure. I'm not saying I'm right, rather I am trying to give my viewpoint so you can see where my head is coming from. I also freely admit that I sometimes get it wrong and contravene my own guidelines... anyhow - these are things that I feel Adventure doesn't do that I would:

* Verbosity over brevity: a method name should be clear as to its function - often using verbs to provide the action (for example, I prefer the use of `get`, `set`, `with` etc.).
* Distinctiveness: two methods on the same object that do not do the same thing should not be named the same thing.
* Static factory methods should be associated with the type they create where possible: this aids the comprehension of what the method creates.

There are valid reasons as to why you've done some of these things and I'm _not_ saying that you should do some of these things, however, I'll explain why we felt the need to create the `Text` class that is now in this PR, whereby some of these principles do come into play.

# Yes, discoverability is a problem

Some of us know the library, and we know it fairly well. Unfortunately, the authors of a library or API are the _worst_ people to comment on discoverability. Having used it now, there are things that I still struggle to get right and they need to be thought about. That is absolutely why the commit appeared in the first place - there has been a lot of fumbling trying to do things and so the class came out to try to support us and our plugin developers.

## Naming can be confusing

The biggest one is your static factory methods. Take the following two methods from your library in isolation - what someone booting an IDE and typing `Component.text(` might see.

```java
Component.text();
Component.text(String);
```

They both look like factory methods to me: my assumption is that they would both create `Component`s, one empty, one containing the string. However, it turns out that the first is a `TextComponent.Builder`, and one is a `TextComponent`. What's bad, however, is that my assumption can actually hold - if I use these as `ComponentLike`s in something that accepts `ComponentLike`s, then I will indeed end up with `Component.empty()` (a method that's actually okay but totally eclipsed by this) and my text component. So now we have two methods that don't do the same thing, but can be treated as such, but are actually totally different ways to do the same thing.

That's not discoverablilty. It's actually ignorance of what the two methods do that means people get away with it. While both can actually be used in the same way, with the same methods names (something I'll come back to under convenience methods), they really shouldn't be.

A comment on the commit said:

> These fall into the same issue as above -- it's not super clear which type of component is being created

You have the same problem to a lesser degree, the naming and the placing of the methods does hurt discoverabilty, even if the return type is `TextComponent`. Putting the `TextComponent` factory methods on `Component` rather than `TextComponent` doesn't make sense to me _for that very reason_. Yes, they return `TextComponent` and they are `Component`s, but `TextComponent.text(...)` rather than `Component.text(String...)` reinforces that the method creates and returns a `TextComponent`. Futher, from the Sponge point of view, where `Text.of()` in API 7 was a (admittedly bad) thing, that is the first place that people are going to start looking. Putting the factory and (renamed) builder methods back on `TextComponent` will aid IDE autocompletion and we're onto a winner - people **will** look for that.

I was personally willing to try run with it, but my ultimate reaction is that `Component.text(...)` should return a `Component`, and not a sub type - indeed, that's how I treat it.

Note that I agree that it's not always appropriate to use `builder()` as the name for a builder. Take the `Parameters` class, where we have a _lot_ of builder methods. The key thing is that they are all unique names, there is no ambiguity as to what they might return if two methods are the same.

For what it's worth, I fell into the same trap with command parameters, particularly the `firstOf` and `seq` methods, some of which returned builders, some didn't. After zml used it, they suggested that it would be better if we renamed the builders to include the word "builder". I was hesitant at first but thought on it. [They were right](https://github.com/SpongePowered/SpongeAPI/commit/8a63de257940fd45af6c5a77462216634d2a3230). It was a slightly different situation, but the naming just makes it clear you're getting something different back. I really think this needs to be done here, and `Component.text()` deprecated - a similar situation with other areas where that is the case.

## Sponge just does things differently

Part of the problem is a different philosophy between Sponge and Adventure. They use `of()` and `builder()` as the standard venacular for creating instances and builders. Adventure has not - and I personally have no problem with having a factory method named `text()` (though I don't like the reasoning of it being used for static imports, but that's a purely personal preference so isn't something I should complain about). The problem is simply that of the previous section - `text()` and `text(...)` are two different things, yet can be used the same way.

Our `Text` class is completely borne out of that. Name it what you will, it pulls some of these discoverable methods into the Sponge namespace so we can cater for Sponge developers who may have trouble breaking with the Sponge way of doing these things (though I don't like that it doesn't then produce "`Text`" interfaces...). 

It is important to note that it is entirely in the right of the Sponge project to do this if they see fit, if Sponge ends up deciding they need to do this, that's up to them. Ultimately, I would much prefer for Sponge interfaces to front some of the Adventure ones to reduce the step for the majority who have no interest in cross platform development, while keeping Adventure interfaces for those of us that want to do cross-platform stuff, but it's not something that I think is entirely realistic at this point in time.

## Serialisers are just not discoverable at all - there is no central place to find them

This isn't really a problem with Adventure itself because I know your serialisers are basicaly separate projects. They are, however, difficult for people to find. zml's suggestion on the commit is a good one and I think we need to do that in Sponge.

(As they are options, rather than constructing objects, I do think they can just sit in a `AdventureSerializers` class somewhere, rather than having their own class... somewhat backwards to the above!)

## Numerous convenience methods does not mean they are discoverable - in fact, they may be harmful

A comment that I saw made earlier is that Adventure has a lot of convenience methods. All well and good - but either _people don't know they are there_ or _they are potentially allowing people to do the wrong thing_.

An example of a difficult to discover method is `LinearComponents.linear(...)`. It's obviously designed to ease the burden of creating a tree like structure, but who is going to go looking for it? You've lost the benefit of this method because it's just not obvious to them. Remember who a lot of the user base is. I only know of it because I was told about it after suggesting such a method would be useful. This was the basis of the `Text.of(...)` method - we actually didn't want to emulate exactly what that did but it was a useful vehicle to allow for the joining of components. Intended or not, that is how the method can be used and it is indeed how I'm using it myself - using it to join multiple `Component`s together without having the `Component.text().append(...).append(...).build()` wrapping.

Let's go back to the `text()` methods from before. This returns two objects with a very important difference, one is mutable, one is not. And yet, both of these are valid `ComponentLike`s that can be used in many areas of Adventure:

```java
Component.text().content("Text").color(NamedTextColor.GREEN).style(Style.style().decorate(TextDecoration.BOLD).decorate(TextDecoration.ITALIC).build());
Component.text("").content("Text").color(NamedTextColor.GREEN).decorate(TextDecoration.BOLD).style(Style.style(TextDecoration.BOLD).decoration(TextDecoration.ITALIC, TextDecoration.State.TRUE)); // slighly facetious with the starting method, but it proves a point
```

**They look pretty much the same, and yet have completely different semantics.** One is created from mutable builders, one is created from immutable objects. There is **nothing** to tell me which one I should use (I know it should be the first one) - part of why the lack of verbs is a problem. In the first method, I create three objects, the text component builder, the style builder and the style object (a fourth will be created upon use). The second method creates five - maybe more, I haven't looked at the implementation. For more complicated structures, this is only going to get worse. There should be _some_ way to differentiate between what happens in both of these methods, and ideally, the latter should not exist at all. But there isn't. This is **bad**, and I am concerned it's going to hit a lot of people, and potentially going to cause a lot of mistakes.

Some thought needs to go into what is the "right" way to do things, and these need to be promoted. I do not know what the right way to do things is as it stands.

# There are just too many options to do the same thing

On a `TextComponent.Builder`, I can set a colour using `.color(...)` or `.style(Style.style(TextColor))` or `.style(Style.style().color(TextColor).build())`. I can set a colour and a decoration using `.decorate(...).color(...)` or `.style(Style.style(TextColor, TextDecoration))` or `.style(Style.style().color(TextColor).decorate(TextDecoration).build())` or `.style(Style.style().color(TextColor).decoration(TextDecoration, TextDecoration.State).build())` or `.style(Style.style(TextColor).decoration(TextDecoration, TextDecoration.State))`.

There isn't inherently anything wrong with options, but there are a lot of ways to do things - when gabi was writing a text yesterday, three of us came up with three different ways of doing things. Kind of in hand with the previous statement, some of these ways feel more correct than others - I feel like the library has been written to please everybody and cater to everyone's style. There is not a single consistent workflow for people using the library, for reasons I've given above.

I think keeping the builders is fine, and the convenience methods of `color` and `decorate` on the base builder. Beyond that - particularly the "with" methods - I'd seriously consider whether they add value (and I don't think they do).

# Documentation is a big problem

I know you're the first person to say documentation is a difficult thing, but it's a _really important one_. The front page of your docs still says that the docs are for an unreleased version of Adventure. They have some basic examples but are sparse. The javadocs are also rather sparse, and this is really hindering some of the usabilty of the library. While I could say "sure, I'll get my elbows out and muck in", in reality, the problem is yours. Javadocs are the second port of call for a developer using their IDE to develop. I honestly think a fair amount of time needs to go into the documentation so that things are much clearer - and to sign post some of the convenience methods where appropriate.

# Don't (just) design the API to the platform, design it to the user

As a final thought: I feel that Adventure for text is less an abstraction layer on top of Minecraft text components, and more just a layer that mimics the Minecraft system. That's okay, but it's not how people think. Take the `Text.of(Component...)` method, for example. The comment made saying "but this means you can't add colours and style in a linear fashion" actually misses the mark. It wasn't to expose the linear components function for all its glory, the point was to make people create the components and allow them to piece them together easily, rather than have to think about children which can add needless complexity. `linear` allows us to do that - and it works - the idea is to get people to use components properly and join them together without having to go `Component.text().content("1").append(Component.text().context(".").append(...).build()).build()).build()` (which is what I think has been suggested before), or even just what I suggested above (`Component.text().append(...).append(...).build()`).

At the end of the day, a good percentage of developers for the Minecraft platforms are relatively new, green developers. Think about what they want to do, not how they should do things in a Minecraft way - and API is meant to turn a machine way of doing things into a human way of doing things. They want a coloured text, maybe one with italics, and perhaps a click event - and they want a simple, discoverable way to do so. 

And that's what this commit was all about - the recognition that we needed the bridge the gap. I really want to see Adventure as a library that simplifes text development and is easy to use and discover, and hard to make mistakes with - but Sponge is always going to have its own take on things - if that spawns classes like this to aid developers, so be it.

---

A final reminder - these are my personal opinions, not those of the Sponge team.

dualspiral:
As a note, I have had a brief exchange with @kashike and the appropriate points will go into issues on the Adventure project.

Zidane:
@kashike @dualspiral is this still necessary or can this be closed out?

dualspiral:
Most of this has been raised in Adventure issues (above), but most have also not seen too much traction. Honestly, I still think there are usability and API design problems in Adventure (I clearly make no secret about that) but I've said my piece about them and I have nothing to add in that regard.

As for this PR, we can close it, but we reserve the right to re-introduce a PR like this if we find that people are generally struggling with Adventure (and by that, I mean it becomes a common question) when using Sponge. If this happens, I must absolutely stress - much like Adventure is not Sponge, Sponge is not Adventure. If we need to put utility methods into Sponge to soften the curve, so be it.

We'll monitor and see what happens.

PR #2277, opened 2020-12-09
Scoped Registry Framework

Zidane:
Needs some heavy documentation to explain the methods but assume any method that doesn't return an Optional will throw an exception should the value not exist.

Feel free to wait to review this until that exists but much of this is likely self explanatory.

To help with demonstration, `MatterTypes` has been updated.


src/main/java/org/spongepowered/api/registry/RegistryEntry.java
```diff
@@ -0,0 +1,49 @@
...
+package org.spongepowered.api.registry;
+
+import org.spongepowered.api.ResourceKey;
+
+/**
+ * Represents an entry within a {@link Registry}.
+ *
+ * @param <V> The type of the registry
+ */
+public interface RegistryEntry<V> {
```
dualspiral: I think this is going to be the key change here and we need to be totally clear about it.

Is this a replacement for the `CatalogType` interface? Given that, in theory, objects could be in multiple registries with different keys, this makes total sense because an object can no longer be identified by key alone - it's a combination of key and registry (or two keys if you _really_ want), and with scoping, as the server/worlds cycle, so too will the registry entries.

I think, in this case, this should contain the `RegistryKey`.

src/main/java/org/spongepowered/api/registry/RegistryEntry.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Gets the {@link ResourceKey key} this entry exists under in a {@link Registry}.
+     *
+     * @return The key
+     */
+    ResourceKey key();
```
dualspiral, line 41: `RegistryKey`? Could have this too, but this isn't sufficient for identification purposes.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterRegistryValueEvent.java
```diff
@@ -24,18 +24,18 @@
...
-import org.spongepowered.api.CatalogType;
+import org.spongepowered.api.ResourceKey;
 import org.spongepowered.api.event.GenericEvent;
 import org.spongepowered.api.registry.DuplicateRegistrationException;
+import org.spongepowered.api.registry.RegistryEntry;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.ScopedRegistryHolder;
 
-public interface RegisterCatalogEvent<T extends CatalogType> extends GenericEvent<T>, LifecycleEvent {
+public interface RegisterRegistryValueEvent<H extends ScopedRegistryHolder, T> extends GenericEvent<T>, LifecycleEvent {
```
dualspiral: Are we assuming that every registry holder can only have one registry of type `T`?

There is also going to need to be another form of `GenericEvent` that accepts two generic parameters - or we turn this into three events, one for each scope.

src/main/java/org/spongepowered/api/item/recipe/RecipeRegistration.java
```diff
@@ -26,9 +26,10 @@
 
 import org.spongepowered.api.CatalogType;
 import org.spongepowered.api.data.persistence.DataSerializable;
+import org.spongepowered.api.event.lifecycle.RegisterRegistryValueEvent;
 
 /**
- * A Sponge plugin recipe registration use {@link org.spongepowered.api.event.lifecycle.RegisterCatalogEvent<RecipeRegistration>} to register the recipe.
+ * A Sponge plugin recipe registration use {@link RegisterRegistryValueEvent <RecipeRegistration>} to register the recipe.
```
dualspiral: This has two type parameters, not one!

src/main/java/org/spongepowered/api/registry/Registry.java
```diff
@@ -0,0 +1,91 @@
...
+     * {@link IllegalStateException}.</p>
+     *
+     * <p>If the key already exists in the registry, a {@link DuplicateRegistrationException}
+     * will be thrown.</p>
+     *
+     * @param key The key to register under
+     * @param value The value to register
+     * @return The added {@link RegistryEntry}
+     */
+    RegistryEntry<T> register(ResourceKey key, T value);
```
dualspiral: I'm not sure I necessarily agree with what I'm saying, but I'll say it anyway as an option: `Optional<RegistryEntry<T>> registerIfDynamic(ResourceKey key, T value);`? Returns empty if not dynamic - and a plugin can go on with its day.

Not wedded to that though.

src/main/java/org/spongepowered/api/registry/RegistryHolder.java
```diff
@@ -0,0 +1,41 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.registry;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public interface RegistryHolder {
+
+    <R> Registry<R> registry(RegistryKey<R> key);
```
dualspiral: I guess this will throw an `IllegalArgumentException` on failure? (edit: and I should really read what you put in your initial comment before commenting... it will throw an exception)

src/main/java/org/spongepowered/api/registry/RegistryScopes.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RegistryScopes {
+
+    RegistryScope[] scopes();
```
dualspiral, line 40: `value()`

dualspiral (commented):
Initial thoughts within.

(General thought: This is going to be paaaainnnnffffuuullll.)

src/main/java/org/spongepowered/api/registry/RegistryEntry.java
Zidane: Negative, CatalogType isn't changing. This is simply a useful Key + Value struct for `Registry`. Could actually be tossed but would make `stream` on `Registry` annoying to use.

src/main/java/org/spongepowered/api/registry/RegistryEntry.java
Zidane: `RegistryKey` could be done here but if you have an entry, you at some point had/know of the actual `Registry` it came from.

Plugin devs are not expected to ever build entries.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterRegistryValueEvent.java
Zidane: No, in fact it is entirely possible for a `RegistryHolder` to have multiple registries of type `T`. This is why this event gives you the `RegistryKey` of the registry being invoked. Unfortunately, plugin devs will need to check against the API `Registries` constants if they want to register against a built-in registry. I don't see any clean away around it.

src/main/java/org/spongepowered/api/registry/Registry.java
Zidane: Could gate it with an `Optional` and likely needs to be an `Optional` return regardless as there are other reasons failure could occur that won't necessitate an exception thrown.

src/main/java/org/spongepowered/api/registry/RegistryHolder.java
Zidane: These methods are really meant for the API use case and very confident plugin devs that know certain content is assured to exist.

That said, this new API is designed around the fact that you are guaranteed one thing and one thing alone: no value in a registry within a holder is ever guaranteed to exist. That will be the pain point of the new system but hopefully RegistryReference will help with this.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterRegistryValueEvent.java
dualspiral: No, neither do I. The only thing I can think of is adding a `Optional<RegistryEntry<T>> registerIf(RegistryKey<T> registryKey, ResourceKey key, T value)` if we want to allow plugin devs to be lazy about it - or perhaps make it the only register method to enforce devs to think about it.

There isn't a nice solution though, I agree.

src/main/java/org/spongepowered/api/registry/RegistryEntry.java
dualspiral: Okay, got it.

gabizou:
Was manually merged in https://github.com/SpongePowered/SpongeAPI/commit/7ae0c2e79483511708130984a24861b3cf34bab3

PR #2278, opened 2020-12-10
enhance InteractEvents [API8]

Faithcaio:
**SpongeAPI** | [SpongeCommon](https://github.com/SpongePowered/Sponge/pull/3242)

Issues solved by this PR:
- `InteractBlockEvent.Primary` getting called multiple times for survival players without a way to differentiate between the actions.
  - by intruducing `InteractBlockEvent.Primary.Start` `InteractBlockEvent.Primary.Cancel` `InteractBlockEvent.Primary.Finish`
- `InteractEntityEvent.Secondary` getting called twice times for each entity.
  - by intruducing `InteractEntityEvent.Secondary.At` `InteractEntityEvent.Secondary.On`

@ImMorpheus

### DIFF
--- a/src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java
+++ b/src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java
@@ -28,11 +28,13 @@ import org.spongepowered.api.block.BlockSnapshot;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.action.InteractEvent;
 import org.spongepowered.api.item.inventory.ItemStack;
 import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.util.Tristate;
 import org.spongepowered.api.world.ServerLocation;
+import org.spongepowered.math.vector.Vector3d;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/event/item/inventory/InteractItemEvent.java
```diff
@@ -25,6 +25,7 @@
 package org.spongepowered.api.event.item.inventory;
 
 import org.spongepowered.api.event.action.InteractEvent;
+import org.spongepowered.api.event.entity.living.AnimateHandEvent;
```
ImMorpheus, line 31: Unused import

src/main/java/org/spongepowered/api/event/block/InteractBlockEvent.java
```diff
@@ -29,6 +29,8 @@
 import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.entity.living.player.Player;
 import org.spongepowered.api.event.action.InteractEvent;
+import org.spongepowered.api.event.entity.living.AnimateHandEvent;
```
ImMorpheus: Unused import

ImMorpheus (approved):
LGTM, just a few nitpicks.

dualspiral:
I'm really not sure we should have a cancellable event that doesn't actually do anything on cancel - I understand that you don't want to put `Cancellable` on all events but one, but given impls are generated, I don't think it's that big of a deal to do so.

PR #2285, opened 2020-12-25
Feature/datapack/dimension

Zidane:
Simply to follow my changes though feedback is welcomed.


src/main/java/org/spongepowered/api/world/generation/settings/NoiseGeneratorSettings.java
```diff
@@ -0,0 +1,66 @@
...
+
+    BlockState defaultBlock();
+
+    BlockState defaultFluid();
+
+    int bedrockRoofY();
+
+    int bedrockFloorY();
+
+    int sealLevel();
```
dualspiral: soa

Faithcaio: but it's fate was sealed

PR #2286, opened 2021-01-01
Pushes default Parameter handling to the CommandContext Implementation

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3247)

### DIFF
--- a/src/main/java/org/spongepowered/api/command/parameter/CommandContext.java
+++ b/src/main/java/org/spongepowered/api/command/parameter/CommandContext.java
@@ -45,8 +45,7 @@ import java.util.Optional;
  * is available (see {@link #getCause()}. Some popular tasks that operate
  * on the {@link CommandCause} are also directly available on this context,
  * namely permission checks (via {@link SubjectProxy}) and sending a message to
- * the {@link CommandCause}'s {@link Audience} (via
- * {@link #sendMessage(Component)}).</p>
+ * the {@link CommandCause}'s {@link Audience} (via {@link #sendMessage}).</p>
  */
 public interface CommandContext extends SubjectProxy {
 
@@ -113,9 +112,7 @@ public interface CommandContext extends SubjectProxy {
      * @param parameter the {@link Parameter} associated with the argument
      * @return whether there are any values present
      */
-    default boolean hasAny(final Parameter.Value<?> parameter) {
-        return this.hasAny(parameter.getKey());
-    }
+    boolean hasAny(final Parameter.Value<?> parameter);
 
     /**
      * Returns whether this context has any value for the given argument key.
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -1412,12 +1427,25 @@ static Parameter seq(@NonNull final Iterable<Parameter> parameters) {
 
     }
 
+    /**
+     * A parameter that consists of multiple parameters
+     */
+    interface MultiParameter extends Parameter {
```
dualspiral: `Multi` - otherwise this is `Parameter.MultiParameter`

src/main/java/org/spongepowered/api/command/parameter/CommandContext.java
```diff
@@ -152,6 +150,7 @@ default boolean hasAny(final Parameter.Value<?> parameter) {
     /**
      * Gets the value for the given key if the key has only one value,
      * otherwise, throws a {@link NoSuchElementException}.
+     * <p>Also takes {@link Parameter.Value#getDefault} into consideration.</p>
```
dualspiral: For context, this has come about because it's aspired that there are no real good ways to do defaults with Brig involved, and that really, the best control should come from the executor, with the parameters to be default set as optional. That way, the executor can decide what to do with the information sent from the parsers, nothing more, nothing less.

The idea here is that if you provide a parameter a default, it doesn't actually get inserted into the context, but if you call `requireOne(Parameter)`, it will insert it here - so you can still attach defaults to the value.

I've been thinking about this and to be honest, I don't like this as much as I thought I would:

* `orDefault` remains on `Parameter.Value.Builder` and is it concievable that people will just assume that the parameter will put a default into the context when it will not.
* If the developer uses the `Parameter.Key<?>` to get the value out instead of the `Parameter.Value<?>`, then we get a different behaviour, and thinking about it, I'm not sure that's right when the only difference is the signature and a line of JDs.
* `requireOne` et. al. no longer represent whether a parameter is in the context, and it's not obvious where the parameter came from<sup>1</sup>.
* `getOne(...).orElseGet(() -> )` could/should be used as an alternative to this in the executor. I _could_ be pursuaded to add a `getOneOrDefault(Key, Function<CommandCause, T>)` but that seems overkill when we're already returning optionals.

I know I was originally okay with this in my conversations with Faith, but I'm honestly leaning towards not having this behaviour because it adds a potential layer of confusion and I just feel like using the `getOne` method is a better solution - let the executor provider deal with it. _They_ know what a sane default for them is and it removes strain on trying to wok out when a default is corect, or when a trailing required argument should be called. [Indeed, that's what they do with Brig in Minecraft](https://github.com/Mojang/brigadier/pull/18#issuecomment-425859748).

--- 
<sup>1</sup> This was kind of a problem before, but in the previous case, defaults _were_ in the context, though there was no way to know where the result came from still.

PR #2288, opened 2021-01-22
Some minor tweaks to the Commands API

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3262)

This is a companion to the Sponge PR where I spent some time looking through the commands. My main goal here was to add a bit to some documentation that felt unclear and look at some other minor things as I came across them.

Some more detail is present on the Sponge side of this PR.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/Command.java
+++ b/src/main/java/org/spongepowered/api/command/Command.java
@@ -30,6 +30,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.command.exception.ArgumentParseException;
 import org.spongepowered.api.command.exception.CommandException;
+import org.spongepowered.api.command.parameter.ArgumentReader;
 import org.spongepowered.api.command.parameter.CommandContext;
 import org.spongepowered.api.command.parameter.Parameter;
 import org.spongepowered.api.command.parameter.managed.Flag;
@@ -95,7 +96,7 @@ public interface Command {
      * @return The result of a command being processed
      * @throws CommandException Thrown on a command error
      */
-    CommandResult process(CommandCause cause, String arguments) throws CommandException;
+    CommandResult process(CommandCause cause, ArgumentReader.Mutable arguments) throws CommandException;
 
     /**
      * Gets a list of suggestions based on input.
@@ -108,7 +109,7 @@ public interface Command {
      * @return A list of suggestions
      * @throws CommandException Thrown if there was a parsing error
      */
-    List<String> getSuggestions(CommandCause cause, String arguments) throws CommandException;
+    List<String> getSuggestions(CommandCause cause, ArgumentReader.Mutable arguments) throws CommandException;
 
     /**
      * Test whether this command can probably be executed given this
@@ -273,7 +274,7 @@ public interface Command {
          * @param arguments The argument {@link String}
          * @return The {@link CommandContext}
          */
-        CommandContext parseArguments(CommandCause cause, String arguments) throws ArgumentParseException;
+        CommandContext parseArguments(CommandCause cause, ArgumentReader.Mutable arguments) throws ArgumentParseException;
 
         /**
          * Gets the {@link CommandExecutor} for this command, if one exists.
@@ -286,7 +287,7 @@ public interface Command {
          * Processes the command by parsing the arguments, then
          * executing command based on these arguments.
          *
-         * <p>By default, this will call {@link #parseArguments(CommandCause, String)}
+         * <p>By default, this will call {@link #parseArguments(CommandCause, ArgumentReader.Mutable)}
          * and pass the resulting {@link CommandContext} to
          * {@link CommandExecutor#execute(CommandContext)}, if this command has
          * an executor. If it does not, this will throw a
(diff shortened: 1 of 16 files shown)


src/main/java/org/spongepowered/api/command/parameter/managed/clientcompletion/ClientCompletionTypes.java
```diff
@@ -51,9 +51,13 @@
...
     /**
-     * Indicates to the client that the {@link ValueParser} will parse Json.
+     * Indicates to the client that the {@link ValueParser} will parse SNBT
+     * (the string format for NBT).
+     *
+     * <p>This format somewhat resembles <em>JSON</em>, but supports unquoted
+     * keys, and can unambiguously distinguish every NBT tag type.</p>
      */
-    public static final DefaultedRegistryReference<ClientCompletionType> JSON = ClientCompletionTypes.key(ResourceKey.sponge("json"));
+    public static final DefaultedRegistryReference<ClientCompletionType> SNBT = ClientCompletionTypes.key(ResourceKey.sponge("snbt"));
```
zml2008, line 60: Unsure about how to refer to this. While SNBT is a name for the format, and used on the MC wiki, the closest API type would be `DataContainer` -- would it make sense to call it `DATA_CONTAINER` instead?

gabizou: I wouldn’t say datacintainer is quite up to snuff, it does not currently restrict that offering anything is within the limits of being compatible with a tag compound (some parts are translated specially). But it’s been on my todo list to revamp the api to be closer to that of having DataView like a ConfigNode while it’s implementation being closer to tags.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterCommandEvent.java
```diff
@@ -63,6 +66,19 @@
...
+     *
+     * <p>This allows for registering additional commands between game
+     * registration cycles.</p>
+     *
+     * <p><strong>Caution:</strong> commands registered outside of this event
+     * will not be visible to data packs, and may not be shown to clients.</p>
+     *
+     * @return the command registrar
+     */
+    CommandRegistrar<C> registrar();
```
dualspiral: Why can't we get the registrar out of the server's `CommandManager` instead? As registrars are scoped to the datapack lifecycle, encouraging a plugin to hold on to a registrar is a bad idea.

I'm not 100% against exposing the registrar in the event, but this can't be the reason to do so.

`CommandRegistrar`s, as a result, should be `@DoNotStore`d.

src/main/java/org/spongepowered/api/command/Command.java
```diff
@@ -95,7 +96,7 @@ static Builder builder() {
      * @return The result of a command being processed
      * @throws CommandException Thrown on a command error
      */
-    CommandResult process(CommandCause cause, String arguments) throws CommandException;
+    CommandResult process(CommandCause cause, ArgumentReader.Mutable arguments) throws CommandException;
```
dualspiral, line 99: A comment rather than a request:

I'm very "eh" on this. I'm not really in favour of forcing the use of the reader on commands, but on the other hand, I understand the reasons why you'd want to provide it. I'd prefer being to create a reader from a string for use elsewhere too (if someone wanted to parse a chat message or something, for example), and keep this as a string so developers can decide, but it's a design/preference sort of thing here. 

I don't care enough to argue about it though because this may very well be the preferred direction anyway. I just see "Raw" as being as low level as possible, which this effectively prevents, but as you can get the string out of the reader, it's not _really_ an issue.

src/main/java/org/spongepowered/api/command/parameter/managed/clientcompletion/ClientCompletionTypes.java
dualspiral: Ideally, it shouldn't refer to NBT. I'd probably just call it "DATA" because that's what it is, it _is_ a data blob.

zml2008: I think portraying the format as anything as vague as `DATA` provides a false sense of abstraction. The client does parse this as a SNBT compound for the purposes of completion, so it cannot be any arbitrary string, and knowning the format used is important for plugins understanding the reasons for using it. I think this could *become* something like `DATA_CONTAINER` if gabe has the time to make the changes he wants to though. For now I'll keep it as `SNBT` since that is, while not ideal, not wrong, but I'm not opposed to changing it in the future.

src/main/java/org/spongepowered/api/command/Command.java
zml2008: hm, this was something I was unsure about as well at first. Why I eventually settled on exposing the argument reader as a method parameter was that we already allow Raw commands to generate a command tree, and an argument reader is an easy way to parse input in a way that is compatible with the types usable for command trees.

Because the raw string is available in `arguments.getRemaining()` this doesn't lock raw commands out of any ways of working, but does give them a pointer towards the right direction to go when seeking low-level access.

PR #2298, opened 2021-02-12
Introduce services for observability

pschichtel:
Impl PR: SpongePowered/Sponge#3296

Some of you may remember the discussions in Discord. I've recently implemented a prometheus metrics exporter as a plugin (actually a CubeEngine module: [observe](https://github.com/CubeEngine/modules-extra/tree/master/observe)). A few days ago I also added  health checking to the module to get better supervision of the server and allow the orchestrator to automatically restart the container and I'm currently looking at OpenTracing and if/how that might come in handy (considering intrumenting @Faithcaio's terra module).

This PR is a direct copy from the observe code base minus the implementations. All service interfaces accept PluginContainer instances to track plugin ownership and I do support unregistration, not sure if that is something that is necessary/useful.

## Health Checks

In observe's implementation, healthchecks are exposed as an HTTP endpoint which will be polled by docker in our case. The API has the concept of a health probe, which is practically a `Suppler<HealthState>`, so an arbitrary function that, when asked, exposes the current health state of whatever it is probing. The simple probe I implemented simply has a task running on every tick recording the last tick timestamp and depending on how long that last tick was during the probe call, it will either return HEALTHY or BROKEN. The API differentiates between "sync probes" and "async probes", meaning probes that must be run on the server main thread and must provide their state in a synchronous way and probes that may be called from any thread and provide their result in an asynchronous way. The implementation in observe schedules them separately and joins the results.

## ~~Metrics~~

~~The metrics service is merely a managing layer on top of Prometheus' java libraries. The dependencies that's being pulled and exposed, is the core library that provides the registries, basic collector types (Gauges, Counters, Histograms). On top of that I added a `PullGauge` helper, that makes it simple to define stateless gauges, that extract their value on-demand (e.g. while being scraped by prometheus) from some target object. I use these to pull information from the Server and ServerWorld's, without requiring me to create a timer (blackbox monitoring basically). The API, similar to the health check service, differentiates between standard `Collector` instances, that can be scraped from any thread, and `SyncCollectors` (just a marker), that must be scraped from the server main thread. The implementation in observe schedules them separately and joins the results.~~

~~We use these interfaces and the exporter implementation behind it (netty based webserver, scheduling scraping jobs using the sponge schedulers) for weeks now without issues. We measured the overhead of the sync scraping using aikar timings to confirm the overhead is insignificant. The metrics are powering a Grafana dashboard which some of you might have seen.~~

## ~~Tracing~~

~~This one is the latest addition, which I assume would be useful to have, but I haven't actually applied this one directly. The service interface simply provides an opentracing `Tracer` instance, which is backed by whatever opentracing implementation. I chose jaeger in my implementation, because I have experience with that, but I could even make this easily configurable in the plugin. I don't really see much more that needs to be covered by the service interface here. I could however imagine having a set of helpers to ease the instrumentation of code. Since OpenTracing is especially useful for distributed tracing, it might be useful to provide e.g. pre-instrumented web-clients, scheduling interfaces/callback wrapping and so on.~~


src/main/java/org/spongepowered/api/service/metrics/pullgauge/PullGauge.java
```diff
@@ -0,0 +1,168 @@
...
+        private final BiFunction<String, String, PullGauge<T>> ctor;
+
+        private Builder(String unit, String help, BiFunction<String, String, PullGauge<T>> ctor) {
+            this.unit = unit;
+            this.help = help;
+            this.ctor = ctor;
+        }
+
+        public Builder<T> unit(String unit) {
+            return new Builder<>(unit, help, ctor);
```
kashike: Should not be returning a new instance here - field should be updated and `this` should be returned.

src/main/java/org/spongepowered/api/service/metrics/pullgauge/SimplePullGauge.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.prometheus.client.Collector.MetricFamilySamples.Sample;
+
+import java.util.List;
+import java.util.function.ToDoubleFunction;
+
+import static java.util.Collections.singletonList;
+
+public final class SimplePullGauge<T> extends PullGauge<T> {
+    private final ToDoubleFunction<T> f;
```
kashike: More descriptive field name, please.

src/main/java/org/spongepowered/api/service/metrics/pullgauge/SimplePullGauge.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public SimplePullGauge(String name, String unit, String help, ToDoubleFunction<T> f) {
+        super(name, unit, help);
+        this.f = f;
```
kashike: More descriptive parameter name, please.

kashike (changes requested):

pschichtel:
I think I'll drop the TracingService from this PR until I have a clearer idea of what is required for that.

src/main/java/org/spongepowered/api/service/metrics/pullgauge/PullGauge.java
pschichtel: It's an immutable builder

kashike: We don't use immutable builders - that defeats the usual purpose of a builder.

pschichtel: I start to remember why I don't usually touch java any more :D Anyway: Question now becomes: Do I align the builder with other collector builders from prometheus or do I align it with builders from sponge?

pschichtel:
I've been thinking about what @gabizou said in discord about wanting to trace information on a per chunk level. I agree, that the information would definitely be useful, but I don't think prometheus is necessarily the right tool for that job. I've been toying with the idea of generating world heatmaps to visualize various high level game play aspects, but I guess the same could be done with low-level information.

This makes me reconsider the idea of directly exposing prometheus/openmetrics and instead abstracting over it in order to be more flexible with the provided information.

pschichtel:
Just to give an update: I'm working on another design in the observe plugin. I will update this branch once I have something useful

pschichtel:
@gabizou @kashike reduced this one to just healthchecks

pschichtel:
@gabizou moved the health check stuff over to observer as well. Nothing remains in the PR, so I close it.

PR #2301, opened 2021-02-27
API8 ChunkEvent

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3305)

### DIFF
--- a/src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
@@ -24,27 +24,19 @@
  */
 package org.spongepowered.api.event.world.chunk;
 
+import org.spongepowered.api.ResourceKey;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
```diff
@@ -54,25 +46,51 @@
...
          */
-        Chunk getTargetChunk();
+        interface Pre extends Save, Cancellable {
+
+            /**
+             * Gets the {@link Chunk} being changed.
+             *
+             * @return The Chunk
+             */
+            Chunk getTargetChunk();
```
gabizou, line 71: If the chunk is available, is this a LevelChunk that has a live reference to the world?

Faithcaio: Could be ProtoChunk too I think.

PR #2303, opened 2021-03-07
adventure: Expose component resolution to API

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3312)

This exposes the component resolution functionality needed to display score, selector, and NBT components to plugins, in a way that is fairly close to what Vanilla provides.

Some things I'm still thinking about:
- Is `Entity` the most appropriate type for a receiver? Perhaps `Audience` would make more sense for the API
- Not sure I like the name `resolve` -- my original choice was `render` to match `ComponentRenderer` in the adventure API, but that makes it sound like a client-side graphics thing.
- Would it make sense to allow plugins to register other renderers?

### DIFF
--- a/src/main/java/org/spongepowered/api/registry/RegistryTypes.java
+++ b/src/main/java/org/spongepowered/api/registry/RegistryTypes.java
@@ -242,8 +242,6 @@ public final class RegistryTypes {
 
     public static final DefaultedRegistryType<ComparatorMode> COMPARATOR_MODE = RegistryTypes.spongeKeyInGame("comparator_mode");
 
-    public static final DefaultedRegistryType<ResolveOperation> RESOLVE_OPERATION = RegistryTypes.spongeKeyInGame("resolve_operation");
-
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/registry/RegistryTypes.java
```diff
@@ -238,6 +238,8 @@
 
     public static final DefaultedRegistryType<ComparatorMode> COMPARATOR_MODE = RegistryTypes.spongeKeyInGame("comparator_mode");
 
+    public static final DefaultedRegistryType<ResolveOperation> RESOLVE_OPERATION = RegistryTypes.spongeKeyInGame("resolve_operation");
```
kashike: Did you forget how the alphabet works? :smile:

zml2008: ah, i renamed it a few times and forgot to re-order lol

PR #2304, opened 2021-03-07
api: Introduce NotificationTickets

gabizou:
**API** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3314)

This revamps the previously meticulous brain-ache
NotifyNeighborBlockEvent into a more
streamlined event similar to ChangeBlock event. As
the batching system has proven useful, this can
safely expose directional suppliers
without explicitly relying on legacy logic of
"Always notify every neighbor on a target position"

Signed-off-by: Gabriel Harris-Rouquette <gabizou@me.com>

### DIFF
--- a/src/main/java/org/spongepowered/api/block/transaction/NotificationTicket.java
+++ b/src/main/java/org/spongepowered/api/block/transaction/NotificationTicket.java
@@ -22,37 +22,28 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.event.block;
+package org.spongepowered.api.block.transaction;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/block/transaction/NotificationTicket.java
```diff
@@ -0,0 +1,49 @@
...
+
+    BlockSnapshot target();
+
+    default Vector3i targetPosition() {
+        return this.target().position();
+    }
+
+    boolean valid();
+
+    void valid(boolean valid);
```
dualspiral: This should be `setValid`, we only dropped `get` prefixes.

dualspiral (commented):
This (+ impl) looks fine, can't see anything untoward. One thing...

ImMorpheus (approved):

PR #2307, opened 2021-03-08
Add CompletableFutures to Ban/Whitelist Service

### LINKED ISSUE #2186: Add CompletableFutures to BanService
The current `BanService` interface effectively requires all bans to be loaded in memory. While this is appropriate for the vanilla implementation in SpongeCommon ([SpongeBanService](https://github.com/SpongePowered/SpongeCommon/tree/stable-7/src/main/java/org/spongepowered/common/service/ban)) it does not match the behaviour of more elaborate plugins which store punishments in a backend requiring IO operations, such as a MySQL database. Should a ban's plugin receive an API request, would it need to perform a database call on the main thread?

I'm wondering whether the Sponge team would be open to revising the *BanService* interface to return `CompletableFuture`s of its current return types, as was similarly done for `PermissionService` per #1524 .

Should this be undertaken, it would be necessary to change some of the assumptions underlying the *BanService*. The [documentation](https://docs.spongepowered.org/stable/en/plugin/bans.html) suggests checking `isBanned` before calling `getBanFor`, but not only would this lead to 2 database queries, it is liable to race conditions. Accordingly, using *getBanFor* and checking if the returned Optional is present is more proper.

With this in mind, the use cases of `isBanned` dwindle to only determining whether a profile or address is banned without caring in the slightest for the `Ban` itself. Combined with the immense potential for abuse of *isBanned*, it should rather be scrapped entirely in favour of the presence of the *getBanFor* Optional. If it is worried amateur developers may call *getBanFor* twice, the documentation should emphasise the circumstances; if it is still worried, method names could even be changed to *findBanFor* to further hammer the point that such method is not necessarily a simple getter. The objective is to avoid wasteful API calls.

ImMorpheus:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3318)
Fix https://github.com/SpongePowered/SpongeAPI/issues/2186

### DIFF
--- a/src/main/java/org/spongepowered/api/service/ban/BanService.java
+++ b/src/main/java/org/spongepowered/api/service/ban/BanService.java
@@ -29,6 +29,7 @@ import org.spongepowered.api.profile.GameProfile;
 import java.net.InetAddress;
 import java.util.Collection;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Represents a service with which to ban things, such as {@link GameProfile}s or IP addresses.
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/service/ban/BanService.java
```diff
@@ -48,77 +49,77 @@
...
+    CompletableFuture<Optional<Ban.IP>> getBanFor(InetAddress address);
 
     /**
      * Checks if a {@link GameProfile} has a ban.
      *
      * @param profile The profile
      * @return True if the profile has a ban, false otherwise
      */
-    boolean isBanned(GameProfile profile);
+    CompletableFuture<Boolean> isBanned(GameProfile profile);
```
jamierocks, line 90: A compeling case to remove this method was made. GH-2186

jamierocks:
The documentation for these methods really needs to reflect this change.

PR #2311, opened 2021-03-12
Update Guice 4.1.0 -> 5.0.1

### LINKED ISSUE #2305: Guice 4.1.0 cannot process Java 11+ classes
This only affects plugins compiled for versions of Java > 8, so will only be hit rarely.

Because Guice shades an outdated version of ASM (see https://github.com/google/guice/blob/4.1/pom.xml#L157), it can fail to process classes compiled with a newer Java target.

This results in errors like: https://gist.github.com/ItsDoot/3e590d870f2ab236e10e762d17414844

Unfortunately, there is not any easy way to fix this. Our options forward appear to be:

1. **Updating to Guice 5.0.1**: This cannot happen because newer Guice requires newer Guava than is shipped with Minecraft
2. **Forking Guice 4.1.0**: This is inconvenient and adds some extra maintenance cost, but would allow us to update ASM (or disable shading it entirely) without needing to break API or hope that Mojang updates Guava.

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3333)

This allows us to be somewhat future-proofed, but does mean implementations may have to use a fork of Guice in order to remain compatible with our Guava version

There are some drawbacks on the implementation side of this -- see that PR for a rundown of the issues.

This fixes #2305

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,9 +57,10 @@ dependencies {
     }
 
     // Dependency injection
-    api("com.google.inject:guice:4.1.0") {
+    api("com.google.inject:guice:5.0.1") {
+        exclude(group ="com.google.code.findbugs", module = "jsr305") // We don't want to use jsr305, use checkerframework
         exclude(group = "javax.inject", module = "javax.inject")
-        exclude(group = "com.google.guava", module = "guava") // We bump the version compared to guice
+        exclude(group = "com.google.guava", module = "guava") // We use an older version than Guice does
     }
 
     // High performance cache + guava - shaded guava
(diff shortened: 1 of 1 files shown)


jamierocks:
We depend on Guava 21.0 here too, so I think its best to just specify the Guice fork here too.

zml2008:
While it would be strictly correct to specify the fork here, it would cause a few issues:

- Cross-platform plugins that use upstream guice would have duplicate classes in their classpaths, since there's nothing that tells gradle that `ca.stellardrift.guice-backport:guice` and `com.google.inject:guice` provide the same API.
- We add API in our fork (by reverting a commit that removed deprecated API) since no suitable replacement exists yet. Currently, we can drop that revert if upstream does eventually add the necessary API. If the fork were visible in SpongeAPI, we would have to maintain that change until a breaking API revision.

Because we explicitly exclude Guava from the Guice dependency, the newer version shouldn't pollute plugin classpaths in most cases, so upstream seems to be the least bad choice for API imo.

jamierocks:
This potentially affects unit testing for plugins because of incompatible dependencies though, and means that third-party implementations need to replace it in their implementation too.

zml2008:
Yes -- it's not at all an ideal solution, but we don't really have any good options. We were forced to upgrade because guice 4.1.0 shaded an ancient version of ASM that was incompatible with any Java version newer than Java 8, and needed workarounds to allow its outdated method of class generation to work. Guice 5 is the earliest version that fixes the class generation issues, but also removes the API we need on the implementation side.

The issues with third-party implementations has been brought up on the implementation PR -- hoping to hear from @Cybermaxke on how suitable this route will be for Lantern.

PR #2312, opened 2021-03-14
Add delegating methods for InteractEvent

CDFN:


ImMorpheus:
See https://github.com/SpongePowered/SpongeAPI/issues/1975 for the discussion on this and a possible approach to the method doc

src/main/java/org/spongepowered/api/event/action/InteractEvent.java
```diff
@@ -24,14 +24,69 @@
...
  * Base event for all interactions.
  */
 public interface InteractEvent extends Event {
-
+    /**
+     * Gets the fake {@link Player} who performed interaction.
+     *
+     * @return Fake player performing interaction
+     */
+    default Optional<Player> fakePlayer() {
```
kashike, line 47: Is this really something that should be here? If so, this probably needs to either have (or point to) more documentation of what a fake player actually is, when one is available, etc.

kashike (changes requested):

Zidane:
@CDFN I am going to put this PR onto the back burner because you have, intentionally or not, shined a light onto a much bigger issue with our events.

The Cause system means our event metadata is completely dynamic...but the fact that it is cuts us both ways. Namely it makes it difficult for plugin devs to know what to reasonably expect from an event without simply listening to it and printing a cause. We are discussing as a team how best to tackle this issue without compromising the dynamic attributes that truly sets Sponge a part from the other platforms.

ImMorpheus:

___

PR #2314, opened 2021-03-14
Remove BanService#hasBan and BanService#isBanned 

ImMorpheus:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3384)

See discussion on https://github.com/SpongePowered/SpongeAPI/issues/2186

### DIFF
--- a/src/main/java/org/spongepowered/api/service/ban/BanService.java
+++ b/src/main/java/org/spongepowered/api/service/ban/BanService.java
@@ -39,8 +39,7 @@ import java.util.concurrent.CompletableFuture;
  * service API (e.g. writing them to a database). In essence, expired bans
  * should be treated the same as user-removed bans.
  *
- * <p>For example, {@link #bans()} would not include any expired bans,
- * and {@link #hasBan(Ban)} would return <code>false</code>.</p>
+ * <p>For example, {@link #bans()} would not include any expired bans.</p>
  */
 public interface BanService {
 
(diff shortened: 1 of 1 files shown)


A248 (commented):
Looks good to me! Amazing to see discussion and API progress happen so quickly.

One thing I was further wondering was, should the documentation stress that `getBanFor` should not be used before calling `pardon` or `addBan`? Those methods, as you mentioned, might be open to abuse as well. I'm not clear on the degree to which you want to pursue correct API usage but I see it as an advantage that Sponge generally steers developers in the right direction.

ImMorpheus:
> One thing I was further wondering was, should the documentation stress that `getBanFor` should not be used before calling `pardon` or `addBan`? Those methods, as you mentioned, might be open to abuse as well. I'm not clear on the degree to which you want to pursue correct API usage but I see it as an advantage that Sponge generally steers developers in the right direction.

I was leaning towards explaining that in the docs, similar to how we have a "best practice" section for the economy service: https://docs.spongepowered.org/stable/en/plugin/economy/practices.html

dualspiral:
This is going to need changes in the impl to support this because the ban entry lists use these methods. Before pulling this I'd like to see an impl PR.

It might be a good time to investigate properly using the `CompletableFuture` returning methods during the login phase rather than just `join`ing - if we have to sync on the main thread every time someone logs in anyway then having futures seems rather pointless because the main use of the service is going to be on login anyway.

PR #2315, opened 2021-03-15
Expose auto spin attack ticks

RedNesto:
SpongeAPI | [Sponge](https://github.com/SpongePowered/Sponge/pull/3336)


Grinch:
Merged in a7d1ce8

PR #2331, opened 2021-04-20
Add more Selector API

jpenilla:
[Sponge](https://github.com/SpongePowered/Sponge/pull/3370) | SpongeAPI

Adds methods to `Selector` to retrieve
- Whether it is able to select players only or any entity type
- The limit for number of entities it can select

These APIs are particularly useful for external command frameworks when combined with the `ENTITY` `ClientCompletionKey`.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/selector/Selector.java
+++ b/src/main/java/org/spongepowered/api/command/selector/Selector.java
@@ -104,6 +104,20 @@ public interface Selector {
     Collection<Entity> select(CommandCause cause);
 
     /**
+     * Get the limit for how many entities this Selector can select.
+     *
+     * @return The limit for the amount of entities this selector may select
+     */
+    int limit();
+
+    /**
+     * Get whether this selector can only select players, or can select any entity type.
+     *
+     * @return Whether this selector can only select players
+     */
+    boolean playersOnly();
+
+    /**
      * Creates a {@link Selector} based on the provided criteria.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/command/selector/Selector.java
```diff
@@ -103,6 +104,31 @@ static Builder builder() {
...
+    /**
+     * Get an optional which contains the input string that was parsed to create
+     * this selector, if it exists.
+     *
+     * <p>If this selector was not created through
+     * parsing a string, an empty Optional will be returned.</p>
+     *
+     * @return An Optional which may contain the input string
+     */
+    Optional<String> inputString();
```
dualspiral: No - a selector should have no link with an input string whatsoever. This is something that should be handled by the external parser itself - they should store the string if its of interest, not this selector.

dualspiral (changes requested):

src/main/java/org/spongepowered/api/command/selector/Selector.java
jpenilla: I've removed the input string API

dualspiral:
I've reworded the Javadocs slightly and made them conform to our style - figured it'd be better to do that rather than get you to do it and waste time doing so.

https://github.com/SpongePowered/SpongeAPI/commit/9273ea9a33c7c239ec95862b07a31a5e7a87e397

Issue #2320, opened 2021-03-21
[API 8] Resolve removal of get prefix on certain methods that didn't accept the change too nicely


NickImpact:
Aiming to use this issue as a general repository of method names in the API that, following the get prefix removal, received confusing renames.

Please append any issues you spot to this ticket rather than creating a new one, so we can have a global repository and one singular issue aimed at fixing the problem. Also, please provide possible renames of the methods so they can be properly discussed.

---

### EconomyService
* getOrCreateAccount --> orCreateAccount 
    *  accountOrCreate (zml) or createAccountIfAbsent (mbaxter)
 
### Objective
* getOrCreateScore --> orCreateScore
    * scoreOrCreate or createScoreIfAbsent

kashike:
https://github.com/SpongePowered/SpongeAPI/commit/e5717aca256ffd1a1b9b87fac1cb701b4c14bf2a

dualspiral:
https://github.com/SpongePowered/SpongeAPI/commit/aee5e7c7915d930914228538e2f100c8cfce60a5 for economy service.

Will leave this issue open for the time being for any other weird results from dropping `get`.

dualspiral:
`GameProfileCache` now has methods such as `byId`, `byName` etc. I feel like these should be named `find` and overloaded if possible, or at least `findById(...)` etc. Thoughts?

ImMorpheus:
> `GameProfileCache` now has methods such as `byId`, `byName` etc. I feel like these should be named `find` and overloaded if possible, or at least `findById(...)` etc. Thoughts?

+1

Can we rename similar methods for consistency ?
`StateContainer#statePropertyByName`
`Scheduler#tasksByName`
`Scheduler#tasksById`
`State#statePropertyByName`

dualspiral:
We should - open for comments but I don't want to leave this proposal too long.

zml2008:
For `GameProfileCache`, how about a scheme like `withName` or `named`?

NickImpact:
I'm more on the side of `withX`, as I can imagine `id` not coming across as straight as it should.

dualspiral:
I'm not sure about `with`, as `with` is generally used to return an immutable copy of itself with a modified attribute. 

Note that I have used `find` for the user manager - it's worth discussing that too as I'd argue is the same thing here. Also probably the same with `account`.

[...]

PR #2332, opened 2021-04-22
Rename methods that don't do well without "get" to "findX/findXOrCreate"

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3374)

See https://github.com/SpongePowered/SpongeAPI/issues/2320. Basically here to ensure that we're okay with the proposed renames.

### DIFF
--- a/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
+++ b/src/main/java/org/spongepowered/api/profile/GameProfileCache.java
@@ -75,7 +75,7 @@ public interface GameProfileCache {
      * @return The profile, if present, or {@link Optional#empty()} if
      *     the cache did not contain a profile with the provided id
      */
-    Optional<GameProfile> byId(UUID uniqueId);
+    Optional<GameProfile> findById(UUID uniqueId);
 
     /**
      * Gets {@link GameProfile}s in bulk by their unique id.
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/api/profile/GameProfileCache.java
```diff
@@ -75,15 +75,15 @@
      * @return The profile, if present, or {@link Optional#empty()} if
      *     the cache did not contain a profile with the provided id
      */
-    Optional<GameProfile> byId(UUID uniqueId);
+    Optional<GameProfile> findById(UUID uniqueId);
```
dualspiral, line 78: I originally wanted to drop "byId" etc. but the problem is that findByIds and findByNames have the same erasure so wouldn't work out so well.

We could have `find` for the singular returns, and keep the current names for the... well, current values!

src/main/java/org/spongepowered/api/state/State.java
```diff
@@ -69,7 +69,7 @@
      * @param name The state property name
      * @return The state property, if available
      */
-    Optional<StateProperty<?>> statePropertyByName(String name);
+    Optional<StateProperty<?>> findStateProperty(String name);
```
dualspiral, line 72: Could just be `findProperty()` instead?

There is the additional complication that there are other `stateProperty` methods that also return `Optional`s, but they are more tied into getting a property from a key. It's less of a search than using a name is. Is there a differentiation here, or should all of these methods be `findStateProperty`?

I can see the argument for saying "find" is for use by a free form identifier, lack of find when passing some key like object in, but we should decide this now.

PR #2333, opened 2021-04-24
Add recipe, biome, and sound ClientCompletionKeys

jpenilla:
[Sponge](https://github.com/SpongePowered/Sponge/pull/3375) | SpongeAPI

Adds `ClientCompletionKey`s for `all_recipes`, `available_biomes`, and `available_sounds`. These are not specific argument types, only suggestion providers. In Vanilla, these suggestion providers are always used in combination with the `ResourceLocationArgument`.

### DIFF
--- a/src/main/java/org/spongepowered/api/command/registrar/tree/ClientCompletionKeys.java
+++ b/src/main/java/org/spongepowered/api/command/registrar/tree/ClientCompletionKeys.java
@@ -40,6 +40,12 @@ public final class ClientCompletionKeys {
 
     // SORTFIELDS:ON
 
+    public static final DefaultedRegistryReference<ClientCompletionKey<CommandTreeNode.Basic>> ALL_RECIPES = ClientCompletionKeys.key(ResourceKey.minecraft("all_recipes"));
+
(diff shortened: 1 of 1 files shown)


dualspiral:
This isn't the right way to go about it.

CCKs map to Brig's `ArgumentType`, so the keys that are provided match to those argument types (why they aren't in a registry, I'll never know, but I digress). 

What you're doing here is adding suggestions to existing argument types. If (and hopefully, when) Mojang create a registry for their argument types, this is what will effectively map on to it. What you're trying to do here actually completely muddies that - "suggestion providers" are different.

I understand what you're trying to do (and what you do in the impl with the Entity Summon argument), but CCKs are meant to represent what you're parsing, not what the completions are. `ALL_RECIPES` still actually parses a `ResourceKey`, regardless of whether you pass a recipe to it or not, and so really it shouldn't be its own key.

What we really need is a separate registry for these suggestion providers and a way to link them up to the nodes.

dualspiral:
(I'm working on it)

dualspiral:
Supersceded by https://github.com/SpongePowered/SpongeAPI/pull/2334. Thanks for taking a stab, unfortunately it was the wrong approach.

PR #2340, opened 2021-05-01
API8 - Improve PlayerChatEvent

### LINKED ISSUE #2259: Rethink PlayerChatRouter
I'm putting it here because I don't have a single solution to this right now, but we _need_ to think about this, when I posted it in Discord a few days ago it got lost amongst the noise.

Basically - the notion of the `PlayerChatRouter` is a step back in our chat event based on what we have in API 7 and I think we need to properly address it.

## API 7 - Message Channels

In API 7 we have the notion of a message channel. They are a bit of a mess and have caused all sorts of compatibility issues. For example in Nucleus 1, I used Message Channels for things like staff chat, because they have the benefit of formatting chat as well as specifying who a message goes to. However, they very quickly got messy.

### Message Channels could only be used one at a time

A lot of users who used Nucleus at the time also used a plugin called Boop, which _also_ used message channels for their per player formatting functionality. It very quickly became apparent that there was no good way to chain message channels together on the formatting side.

### Message Channels hid the final form of a chat message

Another issue, particularly for protection plugins, is that while the chat event could alter a chat message, the message you were manipulating was, in fact, **not** the final message that would be sent to a player. This was necessarily the case because a message channel could customise the message for each player! However, there was no way for a plugin to inspect this, all a message channel could do was send the message - no event would be fired so the final message is shrouded in mystery.

For these reasons, I stopped using message channels in Nucleus as much as possible for formatting - both Boop and GriefDefender were having massive issues integrating with Nucleus because of these issues and I did not want the reduction in compatibility.

## API 8 - PlayerChatRouter

`PlayerChatRouter` inherits these problems, plus one more:

### PlayerChatRouters do not let plugins inspect which `Audience`s a message is going to

While Minecraft 1.16.4 is including the notion of ignoring player chat, 1.15.2 doesn't have this. This is a big issue for protection plugins and chat management plugins who want to know who a message is getting sent to and thus want to block or modify the chat destination outright, but not necessarily change the other formatting - again, think of something that has chat channels, and a plugin that highlights player names for a specific player. The `PlayerChatRouter` can only be used by one plugin at a time - if one plugin just wants to alter the destination audiences and another the formatting, you're stuck. At least with `MessageChannel`s you could chain them to that degree.

---

With this in mind, I think we need to either ditch or severely rework the notion of the chat router. Either:

* We create another event which fires on `ServerPlayer#sendMessage(sender, message, MessageType.CHAT)` that plugins can intercept with the finished chat result, that can then be cancelled. This could get spammy for larger servers, however.
* We split the router into two: chat targeting (collection of audiences?) and chat formatting ((sender, target, message) -> Optional.of(message)), such that they can play nicely together (chaining the formatting would be easier, plugins can see where their messages are going).

Thoughts?

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3391)
#2259
  - adds an easy way to filter Audiences for `MessageChannelEvent`.
  - `PlayerChatRouter` -> `PlayerChatFormatter` now only handles message formatting. Which is applied to each audience of a `ForwardingAudience`
  - `PlayerChatEvent` extends `MessageChannelEvent` to handle the audience target replacing `PlayerChatRouter`

### DIFF
--- a/src/main/java/org/spongepowered/api/adventure/Audiences.java
+++ b/src/main/java/org/spongepowered/api/adventure/Audiences.java
@@ -25,9 +25,14 @@
 package org.spongepowered.api.adventure;
 
 import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.audience.ForwardingAudience;
 import org.spongepowered.api.Sponge;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Predicate;
 
 /**
  * {@link Audience}s.
@@ -62,6 +67,31 @@ public final class Audiences {
     public static Audience withPermission(final String permission) {
         Objects.requireNonNull(permission);
         return Audiences.factory().withPermission(permission);
+    }
+
+    /**
+     * Filters an audience with given predicate.
+     * <p>For {@link net.kyori.adventure.audience.ForwardingAudience}s the predicate is tested on each {@link ForwardingAudience#audiences()}</p>
+     *
+     * @param predicate The predicate
+     *
+     * @return The filtered audience.
+     */
+    public static Optional<Audience> filtered(final Audience audience, final Predicate<Audience> predicate) {
+        if (!(audience instanceof ForwardingAudience)) {
+            if (predicate.test(audience)) {
+                return Optional.of(audience);
+            }
+            return Optional.empty();
+        }
+        final List<Audience> list = new ArrayList<>();
+        for (Audience subAudience : ((ForwardingAudience) audience).audiences()) {
+            Audiences.filtered(subAudience, predicate).ifPresent(list::add);
+        }
+        if (list.isEmpty()) {
+            return Optional.empty();
+        }
+        return Optional.of((ForwardingAudience) () -> list);
     }
 
     /**
--- a/src/main/java/org/spongepowered/api/entity/living/player/PlayerChatFormatter.java
+++ b/src/main/java/org/spongepowered/api/entity/living/player/PlayerChatFormatter.java
@@ -25,16 +25,15 @@
 package org.spongepowered.api.entity.living.player;
 
 import net.kyori.adventure.audience.Audience;
-import net.kyori.adventure.audience.MessageType;
 import net.kyori.adventure.text.Component;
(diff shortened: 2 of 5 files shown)


src/main/java/org/spongepowered/api/adventure/Audiences.java
```diff
@@ -64,6 +70,29 @@ public static Audience withPermission(final String permission) {
...
+                return Optional.of(audience);
+            }
+            return Optional.empty();
+        }
+        final List<? extends Audience> list = Streams.stream(((ForwardingAudience) audience).audiences())
+                .filter(predicate).collect(Collectors.toList());
+        if (list.isEmpty()) {
+            return Optional.empty();
+        }
+        return Optional.of((ForwardingAudience) () -> list);
```
ItsDoot: Should this be recursive instead?

Faithcaio: it is now

dualspiral:
I've been talking to Faith about this as it's been done and this is a step up from where we were.

I do have a couple of thoughts on a slightly different approach and any feedback/thoughts on this would be welcome.

---

For this, the intention is that chat formatters are chained. Would a list of formatters that are then executed in order, rather than chained, be better here. My thought was that plugins could inspect the formatters that other plugins have applied and make decisions based on that, so if I detected a chat router that came from UltimateChat, via some heuristic, for example, I might not want to apply mine at all. It should also be possible for a plugin to clear formatters from this list too

That might be complicating it a bit though. I don't mind not doing this, it's just an alternative, but it might not be all the useful. 

---

Protection plugins, such as GriefDefender, may only care about the end result of a message. Rather than having to go and execute the chat formatter on every audience manually, is there merit to create a second "post but not post" event after the formatting has been resolved? Something like:

```java
interface PlayerChatEvent extends Cancellable {

    interface Format extends PlayerChatEvent {
        // this would be what we have now
    }

    interface Formatted extends PlayerChatEvent {

        /*
         * The Audiences that will recieve a message, and their respective message
         */
        Map<Audience, Component> messages();

        /*
         * Protection plugins could stop certain messages getting through to specific
         * audiences
         */
        void filter(Audience audience);        

    }

}

```

Protection plugins etc. that really only care to cancel something may appreciate this.

PR #2347, opened 2021-05-02
Add CommandCompletion and associated builder

### LINKED ISSUE #2343: Rich command suggestion tooltips
In Brigadier, suggestions are able to have a `Message` tooltip, which will be shown to users when hovering over the suggestion. Minecraft's `Component` implements `Message`, so to get a `Message` from a `Component` is as simple as `SpongeAdventure.asVanilla(Component)`.

This capability is not exposed to API however. In fact, when using the `BrigadierCommandRegistrar`, these tooltips will be lost: https://github.com/SpongePowered/Sponge/blob/0ca0e644f4e3b233f127ee936f5cfee29f0ba62c/src/main/java/org/spongepowered/common/command/registrar/BrigadierCommandRegistrar.java#L198-L199

A possible way to implement this would be replacing String based suggestion APIs with something like a `Completion`, which would be a holder for a string suggestion, and a nullable `Component` tooltip. The implementation can then decide how to deal with these `Completion`s.

In addition it would be nice to expose an API for converting `Component` to `Message`, for plugin developers who want to use the Brigadier registrar.

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3398)

See #2343 - supports tooltips in suggestions. I used tooltips in `/sponge plugins info <plugin>` to show the name of the plugin - we can expand on that if we like - figured it'd be a nice little demonstration.

Showing off the tooltip

### DIFF
--- a/src/main/java/org/spongepowered/api/command/CommandCompletion.java
+++ b/src/main/java/org/spongepowered/api/command/CommandCompletion.java
@@ -1,0 +1,93 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.command;
+
+import net.kyori.adventure.text.Component;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.Sponge;
+
+import java.util.Optional;
+
+/**
+ * A potential completion of a command argument.
+ */
+public interface CommandCompletion {
+
+    /**
+     * Creates a basic completion that only contains {@link String} that a given
+     * input may be completed to.
+     *
+     * @param completion The potential completion
+     * @return The {@link CommandCompletion} object
+     */
+    static CommandCompletion of(final @NonNull String completion) {
+        return CommandCompletion.of(completion, null);
+    }
+
+    /**
+     * Creates a completion that contains {@link String} that a given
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/api/command/CommandCompletion.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Creates a basic completion that only contains {@link String} that a given
+     * input may be completed to.
+     *
+     * @param completion The potential completion
+     * @return The {@link CommandCompletion} object
+     */
+    static CommandCompletion of(final @NonNull String completion) {
```
kashike: Should have a method that accepts `(String, Component)` too, in my opinion.

kashike (changes requested):

src/main/java/org/spongepowered/api/command/CommandCompletion.java
```diff
@@ -0,0 +1,109 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.util.ResettableBuilder;
+
+import java.util.Optional;
+
+/**
+ * A potential completion of a command argument.
+ */
+public interface CommandCompletion {
```
kashike, line 37: I'd argue this should be called `CommandSuggestion` (or similar with "Suggestion") instead.

kashike (changes requested):

src/main/java/org/spongepowered/api/command/CommandCompletion.java
dualspiral: Use the builder. This is only here because this is going to be by far the most common operation.

kashike: I don't see how having another convenience method would hurt?

src/main/java/org/spongepowered/api/command/CommandCompletion.java
dualspiral: Why? We use completion in other parts of the API (e.g. ValueCompleter) - the only places we use "suggest" is as a verb to suggest completions.

kashike: We should be using "suggestion" everywhere, really. We've called them "suggestions" in the API for a long time, so it is what people are used to, especially those dealing with Brigadier and other command libraries now too. Now we have a split of terminology.

dualspiral: > We should be using "suggestion" everywhere, really. We've called them "suggestions" in the API for a long time, so it is what people are used to, especially those dealing with Brigadier and other command libraries now too. Now we have a split of terminology. 

@kashike This really should have been said some time ago. It's been like this for some time now and just before we're looking to freeze the API is not really an ideal time to say this.

@gabizou @Zidane What do we want to do - rename all "Completers" to "Suggestors" or whatever, or are we going to stick with what we have now. If we're going to do it, we really should do it now.

src/main/java/org/spongepowered/api/command/CommandCompletion.java
zml2008: imo a builder is excessive given there are only two possible parameters, and no expectation that more will be added in the immediate future. Two methods (String), and (String, Component) would cover the possibilities for a Completion, and make it immediately obvious which combinations of parameters are possible.

src/main/java/org/spongepowered/api/command/CommandCompletion.java
zml2008: I've gone back and forth a few times looking at various solutions to this naming scheme, but I think that consistently calling everything `Completions` makes sense.

Colloquially, this operation is known as "tab completion", so any name mentioning completions is likely to be familiar to users. JLine implements a similar operation using a `Completer` that provides `Candidates`. Brigadier chooses to go a bit differently by using pure `Suggestion`/`SuggestionProvider` terminology.

The only thing I think should change if we go with `Completion` terminology is `Command.suggestions` -- it clashes with the completion-focused language the API uses elsewhere.

dualspiral: Yeah, I had noticed that too - either way we go we should make that sort of thing consistent and so in that case, it would become `Command.complete`.

src/main/java/org/spongepowered/api/command/CommandCompletion.java
dualspiral: Fair point. I can get behind that - we can always re-introduce a builder later if it comes to it I guess.

PR #2295, opened 2021-02-04
Rename SIDEBAR_TEAM_NO_COLOR to SIDEBAR

thibaulthenry:
**[Sponge ](https://github.com/SpongePowered/Sponge/pull/3283) | SpongeAPI**

This has been discussed on Discord a while ago, with Masa and Zidane.
The important code changes are on Sponge repository.

Thanks for reading the PRs

PR #2350, opened 2021-05-10
Rename SIDEBAR_TEAM_NO_COLOR to SIDEBAR, allow getting DisplaySlot by team colour

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3405)

This superscedes #2295 (thanks to @thibaulthenry for the initial commit!)

Makes `withTeamColor` static and renames it to `findByTeamColor`. I have used a factory to provide the mapping from colour to slot because I couldn't really think of any logical place to put an instance method that does this mapping (maybe `Scoreboard` could house such a method, but it felt a little out of place).

Are we okay with this being a factory, or should we look to make this an instance method somewhere?

### DIFF
--- a/src/main/java/org/spongepowered/api/scoreboard/displayslot/DisplaySlot.java
+++ b/src/main/java/org/spongepowered/api/scoreboard/displayslot/DisplaySlot.java
@@ -26,6 +26,7 @@ package org.spongepowered.api.scoreboard.displayslot;
 
 import net.kyori.adventure.text.format.NamedTextColor;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.Sponge;
 import org.spongepowered.api.registry.DefaultedRegistryValue;
 import org.spongepowered.api.scoreboard.Team;
 import org.spongepowered.api.util.annotation.CatalogedBy;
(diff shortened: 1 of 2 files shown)


Zidane:
Yeah that is an odd name isn't it. That isn't a factory. It doesn't create instances, correct?

How does that actually work, in concept, when you don't provide a Scoreboard to query?

dualspiral:
> Yeah that is an odd name isn't it. That isn't a factory. It doesn't create instances, correct?

That's right - which is why I'm icky about it, it just doesn't feel right.

> How does that actually work, in concept, when you don't provide a Scoreboard to query?

Not sure I'm 100% sure what you mean here, but all a display slot is in our impl is something that wraps around an integer. That integer is used to get an element from an Objectives array. So, for all intents and purposes, a `DisplaySlot` is just a fancy way to abstract the magic numbers away - no connection to a scoreboard what so ever.

kashike:
A factory doesn't always need to *create* instances, technically - it can simply provide already-created things too.

Zidane:
@dualspiral 

Ah that makes sense, thanks for the clarification. You are also correct @kashike , based on that terminology that works here. This can remain then.

PR #2351, opened 2021-05-14
Add ItemRarity API

jpenilla:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3409)

Adds `ItemRarity` and APIs to get the rarity of `ItemStack`s and `ItemType`s.

### DIFF
--- a/src/main/java/org/spongepowered/api/item/ItemRarities.java
+++ b/src/main/java/org/spongepowered/api/item/ItemRarities.java
@@ -1,0 +1,61 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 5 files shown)


src/main/java/org/spongepowered/api/item/ItemRarities.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryScope;
+import org.spongepowered.api.registry.RegistryScopes;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
```
Zidane: Omit the comment on this class, needs to be removed from the others. Doesn't add much value.

jpenilla: updated

Faithcaio:
https://github.com/SpongePowered/SpongeAPI/commit/2d1cf182da004649a16a62507868c99990568f4b

PR #2352, opened 2021-05-15
Purge the TypeTokens constant holder class

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3411)

This class is mostly replaced with Class literals, and minimal use of
TypeTokens directly created where no alternative makes sense.


src/main/java/org/spongepowered/api/data/Key.java
```diff
@@ -94,8 +99,54 @@
...
     static <E, V extends Value<E>> Key<V> of(final ResourceKey resourceKey, final TypeToken<V> type) {
-        return Key.builder().key(Objects.requireNonNull(resourceKey, "resourceKey")).type(Objects.requireNonNull(type, "type"))
-                .build();
+        return Key.builder()
+            .key(Objects.requireNonNull(resourceKey, "resourceKey"))
+            .type(Objects.requireNonNull(type, "type"))
+            .build();
+    }
+
+    static <V> Key<Value<V>> valued(final PluginContainer plugin, final String value, final Class<V> type) {
```
dualspiral: Why `valued` when other similar methods use `of`? I'm guessing it's because `listOf`, `mapOf` wouldn't make sense - but we should at least be consistent.

That said, `valued` doesn't seem right here either. Keys do not hold values, they are not "valued". Perhaps `for`, `forSet`, `forMap` etc.?

zml2008: Yeah -- the naming here is a bit iffy, my main focus was on the API structure here :)

I think `for*` would be a good scheme -- but it has the issue of `for` being a language keyword. I've used `forElement`, but I'm not entirely satisfied with that either -- that seems to suggest the other methods should follow the pattern of `forSetElement`, etc, which is imo overly verbose.

dualspiral: Yeah, I was struggling to think of a good name, truth be told. I also forgot `for` was a reserved word. Maybe `forScalar`? You use Scalar in configurate, so it'd match here, maybe?

kashike: `from`?

zml2008: I don't think `scalar` would be the appropriate type here -- since Configurate combines the concept of an object with multiple fixed fields, and a map with arbitrary keys, which the data API does not. I would accept `from` just fine though. At that point it would probably also make sense to remove the convenience overloads that take a general `TypeToken<X extends Value<Y>>`, leaving that for only those who directly use the builder.

dualspiral: Fair enough. That all seems reasonable, I think we should go with all that.

PR #2354, opened 2021-05-19
Adventure: ComponentSerializer services

kashike:
https://github.com/KyoriPowered/adventure/pull/343


kashike:
@dualspiral We should probably remove these methods instead of just deprecating them - do you agree?

dualspiral:
If it's done before freeze. If we find we need to re-add them later, we can do so, can't remove them after freezing.

PR #2355, opened 2021-05-23
Support Schematic Specification v3

gabizou:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3300) | [Specification](https://github.com/SpongePowered/Schematic-Specification/pull/20)

DO NOT MERGE

Accompanying changes for enabling Schematic supporting the Spec. Some of the changes are necessitated to represent a Schematic from deserialization.

I'll jot some mental notes here as time permits, but there's some ramifications that arise from having Biomes being Server scoped. I can't imagine that the game scope would change before 1.18, but in effect, biomes are re-created every server instance, so it's very well possible that a Schematic cannot exist between server restarts (like on the client going between single player worlds).

### DIFF
--- a/src/main/java/org/spongepowered/api/world/schematic/PaletteTypes.java
+++ b/src/main/java/org/spongepowered/api/world/schematic/PaletteTypes.java
@@ -28,6 +28,7 @@ import org.spongepowered.api.ResourceKey;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.block.entity.BlockEntityType;
 import org.spongepowered.api.registry.DefaultedRegistryReference;
 import org.spongepowered.api.registry.RegistryKey;
 import org.spongepowered.api.registry.RegistryScope;
@@ -57,6 +58,8 @@ public final class PaletteTypes {
      * {@link BlockState biome} is registered via {@link Palette.Mutable#orAssign (Object)}
      */
     public static final DefaultedRegistryReference<PaletteType<BlockState, BlockType>> BLOCK_STATE_PALETTE = PaletteTypes.key(ResourceKey.sponge("block_state_palette"));
+
+    public static final DefaultedRegistryReference<PaletteType<BlockEntityType, BlockEntityType>> BLOCK_ENTITY_PALETTE = PaletteTypes.key(ResourceKey.sponge("block_entity_type_palette"));
 
     // SORTFIELDS:OFF
 
--- a/src/main/java/org/spongepowered/api/world/volume/archetype/block/entity/BlockEntityArchetypeVolume.java
+++ b/src/main/java/org/spongepowered/api/world/volume/archetype/block/entity/BlockEntityArchetypeVolume.java
@@ -91,9 +91,7 @@ public interface BlockEntityArchetypeVolume extends BlockVolume {
 
     }
 
-    interface Mutable<M extends Mutable<M>> extends Streamable<M>,
-        BlockVolume.Mutable<M>,
-        MutableVolume {
+    interface Modifiable<M extends Modifiable<M>> extends Streamable<M>, BlockVolume.Modifiable<M>, MutableVolume {
 
         default void addBlockEntity(final Vector3i pos, final BlockEntity blockEntity) {
             this.addBlockEntity(pos.x(), pos.y(), pos.z(), blockEntity);
@@ -114,6 +112,10 @@ public interface BlockEntityArchetypeVolume extends BlockVolume {
         }
 
         void removeBlockEntity(int x, int y, int z);
+    }
+
+    interface Mutable extends Modifiable<Mutable> {
+
 
     }
 
--- a/src/main/java/org/spongepowered/api/world/volume/biome/BiomeVolumeFactory.java
+++ b/src/main/java/org/spongepowered/api/world/volume/biome/BiomeVolumeFactory.java
@@ -22,25 +22,27 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.api.world.volume.block;
+package org.spongepowered.api.world.volume.biome;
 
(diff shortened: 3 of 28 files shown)


src/main/java/org/spongepowered/api/command/parameter/Parameter.java
```diff
@@ -713,6 +715,32 @@ static Parameter seq(@NonNull final Iterable<Parameter> parameters) {
         return Parameter.builder(type, vvp.build());
     }
 
+    @SafeVarargs
+    static Function<CommandContext, RegistryHolder> registryFromOptions(final Function<CommandContext, Optional<RegistryHolder>>... functions){
```
gabizou: @dualspiral I still don't know how we want to have this documented, but there's the use case that we already discussed before with the testplugin.

src/main/java/org/spongepowered/api/world/schematic/PaletteTypes.java
```diff
@@ -58,6 +59,8 @@
      */
     public static final DefaultedRegistryReference<PaletteType<BlockState, BlockType>> BLOCK_STATE_PALETTE = PaletteTypes.key(ResourceKey.sponge("block_state_palette"));
 
+    public static final DefaultedRegistryReference<PaletteType<BlockEntityType, BlockEntityType>> BLOCK_ENTITY_PALETTE = PaletteTypes.key(ResourceKey.sponge("block_entity_type_palette"));
```
gabizou, line 62: Needs to be removed.

src/main/java/org/spongepowered/api/world/volume/biome/BiomeVolumeFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+    BiomeVolume.Mutable copyFromRange(BiomeVolume.Streamable<@NonNull ?> existing, Vector3i newMin, Vector3i newMax);
+
+    BiomeVolume.Mutable copy(BiomeVolume.Streamable<@NonNull ?> existing);
+
+    BiomeVolume.Immutable immutableOf(BiomeVolume.Streamable<@NonNull ?> existing);
+
+    BiomeVolume.Immutable immutableOf(BiomeVolume.Streamable<@NonNull ?> existing, Vector3i newMin, Vector3i newMax);
+
+    UnrealizedBiomeVolume.Mutable empty(Vector3i min, Vector3i max);
```
gabizou, line 46: I don't know if this should exist as a separate class or whether it should be nested within BiomeVolume? @Zidane @dualspiral opinions

src/main/java/org/spongepowered/api/world/volume/virtual/UnrealizedBiomeVolume.java
```diff
@@ -0,0 +1,38 @@
+package org.spongepowered.api.world.volume.virtual;
+
+import org.spongepowered.api.world.biome.Biome;
+import org.spongepowered.api.world.volume.biome.BiomeVolume;
+
+public interface UnrealizedBiomeVolume<B extends BiomeVolume> extends Virtualized<Biome, B> {
```
gabizou, line 30: This isn't necessarily made use of, but it can be potentially made use by the implementation and/or plugins for "game scoped" volumes that can exist between reloads of entire registries. I did have the feeling that Schematic/ArchetypeVolume should really make use of the "virtualization" by this to further emphasize that they can be deserialized at any given point of the game scope.

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
dualspiral: I actually need to think about where that should go full stop - I always said here was just a temporary home after all.

src/main/java/org/spongepowered/api/world/volume/archetype/block/entity/BlockEntityArchetypeVolume.java
```diff
@@ -91,9 +91,7 @@
 
     }
 
-    interface Mutable<M extends Mutable<M>> extends Streamable<M>,
-        BlockVolume.Mutable<M>,
-        MutableVolume {
+    interface Modifiable<M extends Modifiable<M>> extends Streamable<M>, BlockVolume.Modifiable<M>, MutableVolume {
```
dualspiral, line 94: What's the difference between "Mutable" and "Modifiable"?

dualspiral:
Given we're now in "soft freeze", we need to consider whether some of the changes are necessary. Most of the `Mutable` -> `Modifiable` changes seem to be ABI rather that pure API breaks, but there are a couple of functions that move - I think mostly `empty` calls on `BlockVolume` were what I saw.

If we absolutely _must_ do this then this needs to be done ASAP, but if we don't _have_ to we shouldn't now.

src/main/java/org/spongepowered/api/command/parameter/Parameter.java
dualspiral: I've handled this by splitting apart getting the holder (of which multiple can be specified, choosing the first one it comes across) and `RegistryType` to get.

For reference (some of these already existed)

https://github.com/SpongePowered/SpongeAPI/blob/a5b9d9fb5dd5b84a07a8557e9c2348788112b05a/src/main/java/org/spongepowered/api/command/parameter/managed/standard/VariableValueParameters.java#L231-L263

for use with 

https://github.com/SpongePowered/SpongeAPI/blob/a5b9d9fb5dd5b84a07a8557e9c2348788112b05a/src/main/java/org/spongepowered/api/command/parameter/Parameter.java#L750-L761

gabizou:
@dualspiral Some things we gotta do:
- Entities and BlockEntities need their positions relative to the [0,0,0] of the volume not the offset (currently min) so their positions are purely additive to the block min of the volume itself (Since we know some of the details about the minimum on copying of the entity/blockentity -> archetype, we can do this math on copy, maybe an adjustment on the volume applicator to take it into consideration?
- It feels dirty that we have a pseudo position on an archetype, maybe we should make the stream of archetype entries instead so that position is not being stored within the object?
  -  This would resolve doing the transform on position separately from the archetype being transformed
- Pass transformation around through the transformation referents and verify their transforms (there's some problems with the inversion, it's slightly going off if you try to rotate 180 across the center)

PR #2357, opened 2021-05-24
Initial update of block states & block entities for 1.17

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3414)

wip -- seems there are some oddities with the current ways BlockState API is exposed, so this is waiting for @gabizou to resolve some of that.


ImMorpheus (dismissed):
Missing RegistryScopes annotation

PR #2280, opened 2020-12-14
WorldBorder work

kashike:

PR #2359, opened 2021-05-30
World Borders

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3418)

This superscedes #2280 

The basic premise is that world borders require a `WorldBorder` object to be passed to the `MutableWorldBorder` object to update it. This is to enable proper ChangeWorldBorderEvent events to be fired for both worlds and players. I've not quite finished sorting out the listeners in the test plugin, but basic commands are there.

Having multiple set* functions made the event difficult to implement as you never really had the same border, which is why I've gone down this route.

Comments are needed quickly because this is a break! I'll merge tomorrow early afternoon UK time if there are no objections.


src/main/java/org/spongepowered/api/world/border/MutableWorldBorder.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * A {@link WorldBorder} that is mutable.
+ *
+ * <p>Mutable world borders <em>may</em> be attached to a region or player. If
+ * this is the case, any changes that are made will occur immediately.</p>
+ *
+ * @see WorldBorder for more about world borders.
+ */
+public interface MutableWorldBorder extends WorldBorder {
```
gabizou: Could this have just been WorldBorder.Mutable?

src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
```diff
@@ -239,20 +239,29 @@ default boolean isViewingInventory() {
...
-     * border. If {@code null} is passed, the world border is unset.
+     * Sets the {@link WorldBorder} that this player sees. If {@code null}, the
+     * border is unset, reverting to the border of the world the player is
+     * currently in.
+     *
+     * <p>A {@code false} return does not mean a border was not set, rather,
+     * that the resultant border that was set is not the one that was supplied.
+     * If false is returned, it is suggested that you inspect
+     * {@link #worldBorder()} to investigate what was changed.
+     * </p>
```
gabizou: This seems to be a case of we should return the Optional<WorldBorder> object that was set (if it wasn't refused)

gabizou: Like, asking the developer to re-inspect after attempting to set seems to me that we should just provide it back a a result (Transaction if you will).

src/main/java/org/spongepowered/api/world/border/WorldBorder.java
```diff
@@ -0,0 +1,234 @@
...
+    /**
+     * Gets the time when a contracting world border will warn a player for whom
+     * the world border will reach in {@code time} seconds.
+     *
+     * <p>In Minecraft, the warning is displayed in the form of a reddish
+     * tint.</p>
+     *
+     * @return The warning time
+     */
+    Duration warningTime();
```
gabizou, line 141: Isn't this more of ticks rather than duration?

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
```diff
@@ -77,4 +79,12 @@
      * @return The difficulty
      */
     Difficulty difficulty();
+
+    /**
+     * Gets the {@link MutableWorldBorder}.
```
gabizou: Is it mutable? The return here seems to say it's just generic

src/main/java/org/spongepowered/api/world/border/MutableWorldBorder.java
dualspiral: Yeah, it can be.

src/main/java/org/spongepowered/api/world/storage/WorldProperties.java
dualspiral: Hangover from the previous PR, I'll change it.

src/main/java/org/spongepowered/api/world/border/WorldBorder.java
dualspiral: I was under the impression it was actually seconds - but I'll check that properly before it goes in.

src/main/java/org/spongepowered/api/entity/living/player/server/ServerPlayer.java
dualspiral: I've just made it an `Optional<WorldBorder>`. If it fails, it will return the current border. If it wasn't changed, the supplied object and the returned object (in the optional) will be the same reference. I'll add that to the docs.

src/main/java/org/spongepowered/api/world/border/WorldBorder.java
dualspiral: It is actually seconds, not ticks.

dualspiral:
I ended up dropping the notion of a mutable border completely, it turned out to be useless fluff. Instead, we just use immutable `WorldBorder` objects to set/get the state of the border (it's backed by `WorldBorder.Settings` in the impl code).

It seems to me that to code works, but in my limited testing, I found that the world border wasn't being displayed on the client for some reason, though if I walked into a world border it did start damaging me. It looks like packets are being sent though so I'm not sure exactly what to think at the moment, it could be the `PerWorldBorderListener` needs something doing to it.

As far as API goes, I think this is what we want.

(Edit: testing with the server shows the world border, still have the problem of it not saving though...)

src/main/java/org/spongepowered/api/event/world/ChangeWorldBorderEvent.java
```diff
@@ -48,6 +48,16 @@
...
 
+    /**
+     * Gets the border that was originally meant to be used as an {@link Optional}.
+     * For some targets, the border may be unset and therefore they may not always
+     * have a border. It is in these instances that {@link Optional#empty()} is
+     * returned.
+     *
+     * @return The border that will be in use, if there is one.
+     */
+    Optional<WorldBorder> originalNewBorder();
```
gabizou, line 59: At this point, should we use Transaction? Since it manages the state of original, custom, default, and can always view the final replacement (along with invalidate, isValid, etc. )

dualspiral: Given that an entry here can be null and Transactions do not allow for null values - no.

dualspiral:
I got it all working and saving, turns out I had an and check instead of an or check.

I think this is ready. I thought about throwing DataSerializable onto WorldBorder, but that's really an addition at this stage and can be done after this. If we do adapt Transaction so that it accepts null values, then we can just deprecate some of the current methods and have them redirect to a Transaction - we'll just have to manually implement the event if we choose to do that. However, if we're going to go down that route let's think about simply making `WorldBorder`s part of the Data API for API 9 (or even possibly 8.1, I'm sure it could be done).

PR #2365, opened 2021-06-08
[API-8] Add Tags to sponge

tyhdefu:
SpongeAPI | [Sponge](https://github.com/SpongePowered/Sponge/pull/3422)
Add a Tag API into sponge, to make life easier for plugin developers.


tyhdefu:
This API does not yet have a way of adding to existing tags. I think I would need a custom event for that. Adding your own tags should be possible via RegisterRegistryValue event, using Tag.builder to build  them.
I am also unsure currently on whether to force adding a .uniqueKey() on Tag.Builder or to make it optional.

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,171 @@
...
+    Collection<T> all();
+
+    /**
+     * Check whether this tag
+     * contains the given value
+     *
+     * @param value Value to check
+     * @return Whether the value is a part of this tag
+     */
+    default boolean contains(T value) {
```
Zidane: I don't see much value in this method, most people know how to use collections.

Zidane:
@tyhdefu Example plugin code? Ignoring lack of impl.

src/main/java/org/spongepowered/api/tag/BlockTypeTags.java
```diff
@@ -0,0 +1,48 @@
...
+import org.spongepowered.api.registry.RegistryScope;
+import org.spongepowered.api.registry.RegistryScopes;
+import org.spongepowered.api.registry.RegistryTypes;
+
+@RegistryScopes(scopes = RegistryScope.GAME)
+public class BlockTypeTags {
+
+    // SORTFIELDS:ON
+
+    public static final DefaultedRegistryReference<Tag<BlockType>> WOOL = BlockTypeTags.key(ResourceKey.minecraft("wool"));
```
zml2008, line 214: This should be added to the `generator` source set, so tag catalogs are automatically generated

tyhdefu:
Now has a test plugin, but no ability to retrieve Tag instances in plugins. But its proven working by looking at the F3 menu ingame and observing its effects.

src/main/java/org/spongepowered/api/tag/TagType.java
```diff
@@ -0,0 +1,42 @@
...
+@CatalogedBy(TagTypes.class)
+public interface TagType<T extends Taggable> {
+
+    /**
+     * The directory where the tag is stored,
+     * (data/namespace/xxx/(name).json
+     * @see <a href="https://minecraft.fandom.com/wiki/Tag#Folder_structure">https://minecraft.fandom.com/wiki/Tag#Folder_structure</a>
+     * @return Directory name
+     */
+    String datapackDirectory();
```
dualspiral: If we're going to have this (and I'm not sure if we necessarily should because an alternative impl might not want to use a flat file structure), this should be a `Path`.

Why is this useful anyway?

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,117 @@
...
+         * exists.</p>
+         *
+         * <p>By default, if the {@link ResourceKey#value()} is the same as
+         * another data pack / plugin, the two will be combined, and their
+         * tags appended to each other.</p>
+         *
+         * @param replace Whether to replace instead of merge.
+         * @return This builder, for chaining.
+         */
+        Builder<T> replace(boolean replace);
```
dualspiral: Replace _what_ exactly? When does the replacement occur?

I admit, I don't like this being here because the tag shouldn't really be the one to make a decision as to whether to make the replacement (it really should be done on registration instead), but it does represent what is in the datapack JSON...

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,117 @@
...
+
+        Builder<T> addValues(Collection<T> values);
+
+        /**
+         * A nicely generified tag, <b>That must be registered</b> in {@link org.spongepowered.api.event.lifecycle.RegisterDataPackValueEvent}.
+         *
+         * @return The built tag.
+         */
+        @Override
+        @NonNull TagRegistration build();
```
dualspiral: ...so what's the point of the separation between a `TagRegistration` and a `Tag`?

Also, we're not registering the tag at this point, right? So this isn't a registration...

src/main/java/org/spongepowered/api/tag/TagTypes.java
```diff
@@ -0,0 +1,61 @@
...
+
+    // @formatter:off
+
+    public static final DefaultedRegistryReference<TagType<BlockType>> BLOCK_TYPE = TagTypes.key(ResourceKey.sponge("block_type"));
+
+    public static final DefaultedRegistryReference<TagType<EntityType<?>>> ENTITY_TYPE = TagTypes.key(ResourceKey.sponge("entity_type"));
+
+    public static final DefaultedRegistryReference<TagType<FluidType>> FLUID_TYPE = TagTypes.key(ResourceKey.sponge("fluid_type"));
+
+    public static final DefaultedRegistryReference<TagType<Taggable>> FUNCTIONS = TagTypes.key(ResourceKey.sponge("functions")); // TODO?
```
dualspiral: Don't include it unless we know how deal with it.

This will probably have to wait on us getting functions working in the first place.

src/main/java/org/spongepowered/api/tag/Taggable.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.spongepowered.api.tag;
+
+import org.spongepowered.api.data.DataHolder;
+import org.spongepowered.api.registry.DefaultedRegistryValue;
+
+/**
+ *
+ */
+public interface Taggable extends DataHolder, DefaultedRegistryValue {
```
dualspiral: I feel like this needs utility methods to get the tags associated with this object. 

If a tag is going to have a generic, should this have the same generic?

src/main/java/org/spongepowered/api/tag/Tag.java
tyhdefu: A TagRegistration is representative of the actual json file. TagRegistration is a addition or replacement of tags by a particular plugin. A tag has all of these combined, into a useable object.
I called it TagRegistration to make it like RecipeRegistration - which also hasn't been registered yet. Maybe a change of that too is in order?

src/main/java/org/spongepowered/api/tag/Tag.java
Faithcaio: It is replacing tags based on load order of datapacks.
So no it does not happen during registration but later when the datapack is (re)loaded.

That said we currently have no API way to specify that priority and all datapack values are grouped into one datapack for each type.

src/main/java/org/spongepowered/api/tag/Tag.java
Faithcaio: Registration because we `RegisterDataPackValueEvent#register` those `DataPackSerializable`.

In case of recipes I mirrored how vanilla creates its recipes. 
`net.minecraft.data.recipes.FinishedRecipe` = `RecipeRegistration` which knows how to serialize itself
`net.minecraft.world.item.crafting.Recipe` = `Recipe` which knows how to craft/cook etc.

For Tags we also need this distinction because datapacks can append to or replace existing tags.
What a plugin registers in its datapack is not necessarily the final state when all datapacks are loaded.

src/main/java/org/spongepowered/api/tag/TagType.java
Faithcaio: We need it in Impl but I also don't see a use for it in API.

src/main/java/org/spongepowered/api/tag/Tag.java
dualspiral: Well, that's _not a registration_. It's not _registered_ at that point which is what my point was there.

However, my bigger point is that these things are not clear. I have delved into the source code to understand what is going on - this is how I understand it:

* A `Tag` is the basic "here is a category and you can get everything that fits in this category from this". My guess is that it is the immutable in-game representation. In general, you won't create this.
* A `TagRegistration` is "here is how and what you register". It just so happens to be a superset of what `Tag` is.

Have I got this right here? If so, then this builder needs to be an inner interface to the `TagRegistration`, and the `register` method should be exposed on said registration because the behaviour should be knowable. (On `Tag`, that wouldn't matter because it has no futher role to play in registration so the `register` method is useless).

The docs need to be clearer as to what they do too.

Zidane: @dualspiral I am likely going to institute that all the `data pack` like classes become `XXXXTemplate` as that is what they are and likely this will jive a lot better in your head then.

Observe how `WorldTypeTemplate` and `WorldType` exist and a `WorldType` may make a `WorldTypeTemplate` but a template only, eventually, becomes a type after the `RegisterDataPackValueEvent`

Faithcaio: also keep in mind that Datapacks can be reloaded and so the plugin must register it every time. `RegisterDataPackValueEvent` is called

dualspiral: > @dualspiral I am likely going to institute that all the `data pack` like classes become `XXXXTemplate` as that is what they are and likely this will jive a lot better in your head then.
> 
> Observe how `WorldTypeTemplate` and `WorldType` exist and a `WorldType` may make a `WorldTypeTemplate` but a template only, eventually, becomes a type after the `RegisterDataPackValueEvent`

Sensible enough to me. Consistency is important. In that regard I'm not sure if `TagTemplate` or whatever should implement `Tag` altogether, if we're going to make that distinction.

> also keep in mind that Datapacks can be reloaded and so the plugin must register it every time. `RegisterDataPackValueEvent` is called

Right - but that doesn't change anything here.

src/main/java/org/spongepowered/api/tag/Tag.java
dualspiral: Yes - the docs need to make this super clear.

src/main/java/org/spongepowered/api/tag/TagType.java
dualspiral: It should be removed then.

src/main/java/org/spongepowered/api/tag/Tag.java
Zidane: @dualspiral You are correct. a `TagTemplate` eventually transforms into a `Tag` but they are definitely not the same thing. Should be kept separate.

@Faithcaio Can this happen the same for recipes? Consistency as dual put it is very important. For plugin devs, they should know that they build `RecipeTemplates` and submit it to our event. Getting the actual `Recipe` later is looking into the registry system.

Basically the `data pack template` classes are immutable things you build once, submit, Vanilla/Sponge ops on it to write to filesystem then discarded.

src/main/java/org/spongepowered/api/tag/TagType.java
Zidane: Agreed, remove it. In fact, the rest of the API for data pack content so far gives no assumption that we're writing anything IO wise.

src/main/java/org/spongepowered/api/tag/BlockTypeTags.java
Zidane: Agreed, was going to comment this exact thing.

src/main/java/org/spongepowered/api/tag/Taggable.java
Zidane: This will definitely need utility methods and due to that, a self-referencing generic to fulfill `Tag`

src/main/java/org/spongepowered/api/tag/Tag.java
Faithcaio: Not exactly the same but yes thats how they work. Recipes don't use "normal" registries but `net.minecraft.world.item.crafting.RecipeManager`

It is not consistent with Advancements.

src/main/java/org/spongepowered/api/tag/BlockTypeTags.java
tyhdefu: The issue is that it isn't an actual registry in Minecraft. We need to wait for how we are doing registries for data packs first. But once this is established, I will make sure this is done.

src/main/java/org/spongepowered/api/tag/TagType.java
tyhdefu: Its needed in the api if you wanted to add your own type of tag, but there's no reason it has to be the name of the directory, I should rename to something like id or name.

src/main/java/org/spongepowered/api/tag/Tag.java
dualspiral: In that vein then, `AdvancementTemplate` should be a thing, separated from `Advancement`.

src/main/java/org/spongepowered/api/tag/BlockTypeTags.java
tyhdefu: Took a bit of work, but they now auto-generate.

src/main/java/org/spongepowered/api/tag/TagType.java
tyhdefu: I've changed it to `id`. Most likely I also want to add a builder for TagType soon so plugins can more easily add a new type of tag.

src/main/java/org/spongepowered/api/tag/TagTypes.java
tyhdefu: Removed

src/main/java/org/spongepowered/api/tag/TagType.java
```diff
@@ -0,0 +1,41 @@
...
+
+@CatalogedBy(TagTypes.class)
+public interface TagType<T extends Taggable> {
+
+    /**
+     * The id of this tag type.
+     */
+    String id();
+
+    DefaultedRegistryType<T> taggableRegistry();
```
Zidane: Daring to state we'll always know the registry we can default the type from.

I would just do `RegistryType<T>`

src/main/java/org/spongepowered/api/tag/TagRegistration.java
```diff
@@ -0,0 +1,36 @@
...
+package org.spongepowered.api.tag;
+
+import org.spongepowered.api.datapack.DataPackSerializable;
+
+/**
+ * A registration that creates a or modifies a {@link Tag}.
+ *
+ * <p>All registrations through the API will generate into the Vanilla data pack system</p>
+ */
+public interface TagRegistration extends DataPackSerializable {
```
Zidane: Rest of API will be changed, this will become `TagTemplate`.

Additionally, the builder stuff needs to go over onto this interface. Templates are built, not types.

src/main/java/org/spongepowered/api/tag/TagType.java
```diff
@@ -0,0 +1,41 @@
...
+import org.spongepowered.api.registry.RegistryType;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(TagTypes.class)
+public interface TagType<T extends Taggable> {
+
+    /**
+     * The id of this tag type.
+     */
+    String id();
```
dualspiral: Is this just the ID that this tag type is registered with under the appropriate registry? What does this ID look like? _Why_ do we need this to create tag types?

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Collection;
+
+/**
+ * Represents a Tag for a given type.
+ * Tags group {@link TagTypes}.
+ *
+ * @param <T> Type that this tag groups
+ */
+public interface Tag<T> extends DefaultedRegistryValue, ResourceKeyed, TagTemplate {
```
dualspiral: I thought we said that a `Tag` and `TagTemplate` should be separated?

src/main/java/org/spongepowered/api/tag/TagType.java
tyhdefu: No - Its effectively a "key" that doesn't have a namespace. Its used to the folders in vanilla:
e.g `blocks` `entity_types`, `items`. In the registry they are registered under different keys.
These need to comply with minecraft's folder system - cannot be namespaced.
I thought it was best to call them `id`s otherwise doing a `Path` assumes that its stored in a file structure.
Open to suggestions though.

Zidane: For clarity, what would be the filepath of a self-added tag?

That will help determine if really necessary.

src/main/java/org/spongepowered/api/tag/Tag.java
tyhdefu: Whoops, fixed

src/main/java/org/spongepowered/api/tag/EntityTypeTags.java
```diff
@@ -0,0 +1,59 @@
...
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.EntityType;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
+ * <!-- This file is automatically generated. Any manual changes will be overwritten. -->
+ */
+@SuppressWarnings("unused")
```
ImMorpheus, line 39: Missing `@RegistryScopes` annotation

src/main/java/org/spongepowered/api/tag/BlockTypeTags.java
```diff
@@ -0,0 +1,220 @@
...
+import org.spongepowered.api.block.BlockType;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
+ * <!-- This file is automatically generated. Any manual changes will be overwritten. -->
+ */
+@SuppressWarnings("unused")
+public final class BlockTypeTags {
```
ImMorpheus, line 41: Missing `@RegistryScopes` annotation

src/main/java/org/spongepowered/api/tag/FluidTypeTags.java
```diff
@@ -0,0 +1,52 @@
...
+import org.spongepowered.api.fluid.FluidType;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
+ * <!-- This file is automatically generated. Any manual changes will be overwritten. -->
+ */
+@SuppressWarnings("unused")
+public final class FluidTypeTags {
```
ImMorpheus, line 41: Missing `@RegistryScopes` annotation

src/main/java/org/spongepowered/api/tag/ItemTypeTags.java
```diff
@@ -0,0 +1,156 @@
...
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
+ * <!-- This file is automatically generated. Any manual changes will be overwritten. -->
+ */
+@SuppressWarnings("unused")
+public final class ItemTypeTags {
```
ImMorpheus, line 41: Missing `@RegistryScopes` annotation

ImMorpheus (dismissed):

src/main/java/org/spongepowered/api/tag/TagType.java
tyhdefu: datapackname/data/namespace/tags/tagtype (the id)/xxx.json

Datapacks / plugins are within their rights to add json files under namespaces other than their own.
Infact I have just realised that it is probably a good idea to have seperate datapacks per plugin. Otherwise you're won't be able to specify the priority of datapacks, or disable individual datapacks, and possibly your checking of whether a datapack is still valid may be wrong too.

dualspiral: Let's take it from a different angle.

I don't care if this is in the implementation as an implementation detail. Why is this in the API?

Zidane: @dualspiral I see, in the case if someone was to add their own `TagType`...basically if someone was to make something else `Taggable`. It just so happens to correlate to the sub directory under `tags`

That said, @tyhdefu I don't think we can support this. While this is ultimately an implementation detail, the client will die if it gets unknown types sent through the protocol. I...think, perhaps I am wrong.

and maybe not able to support isn't entirely right...a system would need to go into place for "logic"/"server-only" tag types.

src/main/java/org/spongepowered/api/tag/Tag.java
tyhdefu: The docs are now:
```
/**
 * <p>Whether to replace instead of append if the tag already
 * exists. This replaces any minecraft tag data and any data packs
 * that are lower priority than this one (the plugin data pack)</p>
 *
 * <p>By default, if the {@link ResourceKey#value()} is the same as
 * another data pack / plugin, the two will be combined, and their
 * tags appended to each other.</p>
 *
 * @param replace Whether to replace instead of merge.
 * @return This builder, for chaining.
 */
```

src/main/java/org/spongepowered/api/tag/Taggable.java
tyhdefu: I've added the utility methods - I hid them in the implementation as they were to cluttery/confusing for `default` in my opinion. Furthermore we might want to change how we retrieve them, or cache them etc.

src/main/java/org/spongepowered/api/tag/TagType.java
tyhdefu: Removed from the API.

src/main/java/org/spongepowered/api/datapack/DataPackManager.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Collection;
+
+/**
+ * An interface for all information about data packs.
+ *
+ * Everything in here is subject to change throughout the
+ * lifetime of the server.
+ */
+public interface DataPackManager extends ScopedRegistryHolder {
```
Zidane: So this I feel is outside the scope of this PR.

There is no need for the `DataPack` registry scope (and frankly it wouldn't change anything with the impl in Sponge of this PR). In fact, data pack content is ENGINE scoped (SERVER namely)

That said, there is value in knowing all the loaded packs currently. I would put this into a separate PR where we can discuss what further support in the API for the data pack system would look like.

tyhdefu:
Just pulled in Zidane's most recent data pack commits. Currently in an non-working state, as I didn't bother merging some files properly, as those files were most likely going to be reverted later. Will fix when we have a decision on how to handle everything.

src/main/java/org/spongepowered/api/datapack/DataPackManager.java
tyhdefu: Removed

src/main/java/org/spongepowered/api/registry/RegistryScope.java
```diff
@@ -31,5 +31,5 @@
 public enum RegistryScope {
     GAME,
     ENGINE,
-    WORLD
+    WORLD,
```
dualspiral: Remove the comma.

src/main/java/org/spongepowered/api/registry/Registry.java
```diff
@@ -37,7 +37,7 @@
  * @param <T> The type
  */
 @DoNotStore
-public interface Registry<T> extends Iterable<RegistryEntry<T>> {
+public interface Registry<T> {
```
dualspiral: I mostly want to call this out as a break - which is fine, but it needs visibility. Given the problems we know we have with this, I may make this change a direct commit on `api-8` so that it is clear to anyone tracking this.

src/main/java/org/spongepowered/api/tag/TagTemplate.java
```diff
@@ -0,0 +1,108 @@
...
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.datapack.DataPackSerializable;
+
+import java.util.Collection;
+
+/**
+ * A template that creates a or modifies a {@link Tag}.
+ *
+ * <p>All registrations through the API will generate into the Vanilla data pack system</p>
```
dualspiral: Remove this line - it's an implementation detail.

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.spongepowered.api.tag;
+
+import org.spongepowered.api.ResourceKeyed;
+import org.spongepowered.api.registry.DefaultedRegistryValue;
+
+import java.util.Collection;
+
+/**
+ * Represents a Tag for a given type.
```
dualspiral, line 36: I feel like this needs more explanation because I don't think it's clear what a tag is in the first place. It doesn't help that Minecraft calls these things "tags" when they seem to work the wrong way round for me - I'd expect to tag an object, not for an object to be included in a tag...

Anyhow:

> Represents a Tag for a given type

What's a tag?

> "Tags group TagTypes" 

This means nothing to me - and is argubly wrong in the first place because it _doesn't_ group tag types themselves.

I think this needs to read something a bit more like (maybe not exactly):

```
A {@link ResourceKey resource keyed} collection of {@link Taggable} values (of type {@code T}).

<p>While any number of tags may exist and the values each tag contains is arbitary, vanilla Minecraft generally uses pre-defined tags for one of the following reasons:</p>

<ul>
    <li>To define a common material that {@link BlockType blocks} are made of, for example, {@link BlockTypeTags#ACACIA_LOGS}</li>
    <li>To define a common behavior that the {@link Engine engine} should apply to a block, such as specifying a block as a log that can burn via {@link BlockTypeTags#LOGS_THAT_BURN}</li>
</ul>
```

src/main/java/org/spongepowered/api/tag/TagTemplate.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        /**
+         * Adds a child tag to this tag.
+         * This adds all of its values to this tag.
+         *
+         * @param childTag Child tag to add.
+         * @return This builder, for chaining
+         */
+        Builder<T> addChild(Tag<T> childTag);
```
dualspiral: At the point you're creating a `TagTemplate`, there aren't any `Tag`s available, are there? Should this be a `TagTemplate` instead?

src/main/java/org/spongepowered/api/tag/TagTemplate.java
```diff
@@ -0,0 +1,108 @@
...
+         *
+         * @param value Value to add
+         * @return This builder, for chaining
+         */
+        Builder<T> addValue(T value);
+
+        Builder<T> addValues(Collection<T> values);
+
+        /**
+         * A nicely generified tag, <b>That must be registered</b> in {@link org.spongepowered.api.event.lifecycle.RegisterDataPackValueEvent}.
```
dualspiral:
```
Creates a {@link TagTemplate} that should be registered during the {@link RegisterDataPackValueEvent}.
```

Import Javadocs references as well.

src/main/java/org/spongepowered/api/tag/TagTypes.java
```diff
@@ -0,0 +1,59 @@
...
+import org.spongepowered.api.fluid.FluidType;
+import org.spongepowered.api.item.ItemType;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryScope;
+import org.spongepowered.api.registry.RegistryScopes;
+import org.spongepowered.api.registry.RegistryTypes;
+
+@RegistryScopes(scopes = RegistryScope.GAME)
+public class TagTypes {
```
dualspiral: `final`

```
Represents a type that a {@link Tag} can represent.
```

src/main/java/org/spongepowered/api/tag/Taggable.java
```diff
@@ -0,0 +1,47 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.tag;
+
+import org.spongepowered.api.registry.DefaultedRegistryValue;
+
+import java.util.Collection;
+
+/**
+ * A type that can be tagged inside a {@link Tag}
```
dualspiral:
```
A type that may be included in one or more {@link Tag} collections.
```

src/main/java/org/spongepowered/api/tag/Tag.java
```diff
@@ -0,0 +1,54 @@
...
+ * @param <T> Type that this tag groups
+ */
+public interface Tag<T> extends DefaultedRegistryValue, ResourceKeyed {
+
+    /**
+     * Get every value in this Tag.
+     *
+     * @return All tag values
+     */
+    Collection<T> all();
```
dualspiral: You're adding `values` in the template, so should this be `values` to be consistent?

dualspiral (commented):
Something that needs to be improved is the documentation - put my thoughts below. Also, I suspect there is a mistake in `TagTemplate` - it probably shouldn't be accepting child `Tag`s, right?

src/main/java/org/spongepowered/api/tag/TagTemplate.java
Zidane: This will need be a template, too early in the bootstrap otherwise.
In fact, can you take a RegistryReference<TagTemplate>? Notice how in `WorldTemplate` I do not point at a `WorldType` directly.

src/main/java/org/spongepowered/api/registry/Registry.java
tyhdefu: Yeah I was planning to put a comment on here saying I removed it. Its needed for this PR to implement a registry that implements sponge's registry and extends minecraft's registry.

src/main/java/org/spongepowered/api/tag/Tag.java
tyhdefu: I've done it exactly, as I think this description is good, but I'll wait for your confirmation.

src/main/java/org/spongepowered/api/tag/TagTemplate.java
tyhdefu: Correct. Infact it needs to be a RegistryKey to allow types that are registered by plugins/datapacks other than itself. Im not using a RegistryReference because what we actually need is a ResourceKey, guaranteed to be from the correct registry - we don't actually want the Tag object itself.

tyhdefu:
I've also realised that you can specify an entry that allows specifying if the tag or taggable is required for the tag to load or not. Now added to the API.

src/main/java/org/spongepowered/api/tag/Tag.java
dualspiral: It's fine, just wanted to make sure I got it right as well.

src/main/java/org/spongepowered/api/registry/Registry.java
dualspiral: It's been committed directly.

tyhdefu:
I seem to have fixed the Registry remove Iterable<T> diff, but now git thinks i renamed TwoDimensionalPortal and changed all its contents. Interesting.

tyhdefu:
Now complete / ready for further review, the same as the implementation.

dualspiral:
https://github.com/SpongePowered/SpongeAPI/commit/119a9e97384465416f0d377871df4c633dd6fb09
https://github.com/SpongePowered/SpongeAPI/commit/9e9f8c2c54ef82905ba9e6cc3f7791f2d08cc56c

PR #2369, opened 2021-06-29
[API-8] Make TabListEntry serializable, untether from TabList

ejm:
SpongeAPI | Sponge (TODO)

The functionality previously in TabList now exists as a Key for entries and as methods in Audience.

Essentially, the idea is that when you offer a player these entries, the implementation will compare to the current state of a player's tab list and [determine what changes need to be sent](https://gist.github.com/ejm/00b2a377df6eef66e668164619d23851) to the client. Additionally, by making `TAB_LIST_ENTRIES` a key, we can take advantage of `ChangeDataHolderEvent.ValueChange` to regulate changes to the tab list by Vanilla or other plugins.


dualspiral:
We'll take the break on this one - full speed ahead with impl!

ejm:
I've done some more thinking, and while it might make sense to serialize as if the entries are just a list (or set?), being able to reference by UUID internally is just too useful to get rid of. Making any operation that gets an entry O(n) feels like a regression. Regardless, the entry set still won't be tied to any individual player.

Also, should we track a server-wide `TabListEntrySet` that plugins could use to restore the state of the player's tab list?

dualspiral:
> Also, should we track a server-wide TabListEntrySet that plugins could use to restore the state of the player's tab list?

Rather, you should treat the absence of the data on a player as "defer to server" (you can remove data so use that to your advantage).

Nothing wrong with having the "server" list accessible, you might want to see what the server says though, but offering this to a player should be a snapshot in time and not "auto-updating".

ejm:
I think at this point, then, it makes sense to store two entry sets per-player. One is regularly updated by vanilla/mods, etc. and the other is the one offered with the Data API. If the second entry is removed, just revert to the first entry and continually update it as if there was never any custom list to begin with.

src/main/java/org/spongepowered/api/entity/living/player/tab/TabListEntrySet.java
```diff
@@ -24,81 +24,20 @@
...
 import java.util.Collection;
 import java.util.Optional;
 import java.util.UUID;
 
 /**
- * Represents a {@link Player}'s tab list.
+ * Represents a {@link Player}'s tab list entries.
  */
-public interface TabList {
+public interface TabListEntrySet extends DataSerializable {
```
kashike: Maybe just `TabListEntries`? I don't like "Set" in the name

kashike (changes requested):

src/main/java/org/spongepowered/api/entity/living/player/tab/TabListEntrySet.java
ejm: Yeah, that's fine! I really wasn't sure what to call it. I'll change the name when I'm home.

ejm:
> > Also, should we track a server-wide TabListEntrySet that plugins could use to restore the state of the player's tab list?
> 
> Rather, you should treat the absence of the data on a player as "defer to server" (you can remove data so use that to your advantage).
> 
> Nothing wrong with having the "server" list accessible, you might want to see what the server says though, but offering this to a player should be a snapshot in time and not "auto-updating".

I just want to confirm here, if a plugin offers a tab list, the server should not otherwise touch it at all, correct? Just want to make sure I write this in a way consistent to that!

dualspiral:
> I just want to confirm here, if a plugin offers a tab list, the server should not otherwise touch it at all, correct? Just want to make sure I write this in a way consistent to that!

I went to read the implementation to see how it works and I'd argue this is more complicated than meets the eye.

I think before we make any more decisions, make an implementation of what you think should happen in the _simplest_ case. That is:

* What do you think should happen if someone gets the data (include both the "never set" case and "previous set" cases)?
* What do you think should happen if someone sets the data?
* Later on: if I set the data, what happens if a new player joins or leaves? (this is where I was thinking about removing data, and why your suggesting of a "default" tab list on the server might be a good idea)

Get that built. Then, we can think more about the problem of whether something should be updated or not. We save no time by just sitting around with an API, having an impl to talk over will make it easier to frame the problems.

However, there is going to be a bit of a problem (and this isn't your fault or anything!) in that I think this is going to affect things like tab completions for players and other things where the client uses the local player list - from the tab completion point of view this may be a real problem. I guess we'll see what happens...

dualspiral:
Well, SS API-8 has sailed.

I think that, given the above, we might be better going back to basics and having a think about what is wrong with tab lists at the moment, and how we might want to fix it. I think I derailed this somewhat - that's on me.

I'd welcome a new PR for API-9 (though we may end up better thinking about versions later), or even just an issue with a core problem statement and what is deficient about the current API.

(I also have to add, in impl, tab lists are a pain in the ass...).

ejm:
That "pain in the ass"-ness is the reason I ultimately never finished this.
I don't think you really derailed this one as much as it just not being
trivial to track what's going on.

I'll think about this more for a future version, though!

PR #2372, opened 2021-07-04
[WIP] Add short-hand registry getters

Zidane:
For Game/Server scope, a REGISTRY field

For world scope, a static util method that is passed a ServerWorld

https://github.com/SpongePowered/Sponge/commit/cb8be18c052acb545a1d2c6fc463e3af238898ae


Zidane:
@zml2008 Be kind :)

Additionally, could probably elect to not do a method for GAME/SERVER scope but was having issues keeping it "clean"

src/main/java/org/spongepowered/api/world/WorldTypes.java
```diff
@@ -40,6 +41,8 @@
 public final class WorldTypes {
 
     // @formatter:off
+    public static final Registry<WorldType> REGISTRY = WorldTypes.registry();
```
dualspiral: If the game is capable of restarting the server (i.e. the client), this is going to hold a stale reference (or if called at the wrong time, cause the class to not load altogether).

I'm not sure why this needs to be a field. Unlike the other fields, this isn't a member of the world types (same thing for other classes). I think the registry access should be a method.

Zidane:
@Grinch will continue this one.

PR #2381, opened 2021-07-21
Adjustments to Chunk and World interfaces

dualspiral:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3481)

Been discussed in Discord - sorting out confusing naming and such.


src/main/java/org/spongepowered/api/event/world/chunk/WorldChunkEvent.java
```diff
@@ -28,54 +28,54 @@
 import org.spongepowered.api.event.Cancellable;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.world.World;
-import org.spongepowered.api.world.chunk.Chunk;
+import org.spongepowered.api.world.chunk.WorldChunk;
 import org.spongepowered.math.vector.Vector3i;
 
-public interface ChunkEvent extends Event {
+public interface WorldChunkEvent extends Event {
```
dualspiral: @Zidane I've opted to make this rename here because all event really are on `WorldChunk`s, but we might for simplicity not want this change. If not - easy to revert.

Zidane: I'd leave it as ChunkEvent and refactor it so that the base event doesn't supply the World's key.

src/main/java/org/spongepowered/api/event/world/chunk/WorldChunkEvent.java
```diff
@@ -86,14 +86,14 @@
     /**
      * Called when a new {@link WorldChunk} was generated.
      */
-    interface Generated extends ChunkEvent {
+    interface Generated extends WorldChunkEvent {
```
gabizou: Is this actually being called right now? I have the feeling that Generated is somewhat stepping onto the point of being loaded since a chunk isn't "generated" until it flows through the pipeline, but then it's not a WorldChunk yet.

src/main/java/org/spongepowered/api/world/generation/GenerationRegion.java
```diff
@@ -51,8 +51,13 @@
...
  */
-public interface GenerationRegion extends ProtoWorld<GenerationRegion> {
+public interface GenerationRegion extends WorldLike<GenerationRegion> {
 
+    /**
+     * The {@link ServerWorld} that this region will be placed into.
+     *
+     * @return The world
+     */
     ServerWorld world();
```
gabizou: Maybe add a note that the world may not be safe to access asynchronously still? (Even Mojang locks this down a bit internally)

src/main/java/org/spongepowered/api/world/generation/GenerationRegion.java
```diff
@@ -65,37 +70,74 @@ default Difficulty difficulty() {
         return this.world().difficulty();
     }
 
-    Vector3i centerChunkPos();
+    /**
+     * The minimum chunk co-ordinate of this region.
+     *
+     * @return The chunk co-ordinate as a {@link Vector3i}
+     */
+    Vector3i chunkMin();
```
gabizou, line 87: Is this based on the chunk layout of the world (chunk pos vs world block pos)? Maybe we might want to provide the ChunkFormat or whatever that API object is that provides that conversion?

One thing to note is that GenerationRegion inherits the `blockMin()` and `blockMax()` from `Volume` by definition.

src/main/java/org/spongepowered/api/world/generation/GenerationRegion.java
dualspiral: I think the better option here is to just provide the world key to try to discourage that further.

src/main/java/org/spongepowered/api/world/generation/GenerationRegion.java
dualspiral: Fairly sure that this is world co-ordinates, but something to check.

src/main/java/org/spongepowered/api/event/world/chunk/WorldChunkEvent.java
dualspiral: https://github.com/SpongePowered/Sponge/blob/e84a9c53f48f2a3a9c2b485b5ed7176e9ed6020b/src/mixins/java/org/spongepowered/common/mixin/core/server/level/ChunkMapMixin.java#L137-L149

It's done so here and uses an `ImposterProtoChunk` which wraps a `LevelChunk`. I haven't personally looked at this code though.

PR #2384, opened 2021-07-29
Desupply the API

Zidane:
Most of the supplier methods existed when types were going to be fed by
explicit suppliers (and not by lookups that references do right now).
Whereas references are suppliers, that was only done for passing to
other java constructs easier. The plugin dev needs to always call get
themselves to terminate the callstack at the call site.

It also doesn't help that the only consistent thing regarding suppliers
were we were inconsistent which makes for a messier API. This cleans
this up.

@dualspiral 
@gabizou


gabizou:
If we can retain suppliers throughout, that’s be better (since DefaultedRegistryReference naturally implements it) than removing all supplier usage since it does make the various api calls a fair bit cluttered for the various places.

Zidane:
Closing this as discussion was made and the intent will be to go and standardize the supplier calls throughout the API.

PR #2391, opened 2021-09-04
Add .entities in EntityVolume, lift from ServerWorld

tyhdefu:
SpongeAPI | [Sponge](https://github.com/SpongePowered/Sponge/pull/3520)

This PR adds the ability to get all entities from chunks as well as worlds. It is not breaking.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/volume/entity/EntityVolume.java
+++ b/src/main/java/org/spongepowered/api/world/volume/entity/EntityVolume.java
@@ -78,6 +78,7 @@ public interface EntityVolume extends Volume {
     /**
      * Gets all the loaded entities in this entity volume, in no particular
      * order.
+     * The returned collection is unmodifiable.
      *
      * @return All entities in this volume.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/world/volume/entity/EntityVolume.java
```diff
@@ -75,6 +75,14 @@
      */
     Optional<Entity> entity(UUID uuid);
 
+    /**
+     * Gets all the loaded entities in this entity volume, in no particular
+     * order.
+     *
+     * @return All entities in this volume.
+     */
+    Collection<? extends Entity> entities();
```
gabizou: Can we determine whether the collection is thread safe? Is it unmodifiable? Reusable? Any sort of information about the collection itself? (I realize that the original `entities` isn't even documented, but this is something I'd like to have some solid understanding on, whether we want the API to be returning some concrete definition of what is a returned collection)

tyhdefu: I'll make it unmodifiable - some could be re-used but its best to have a uniform front.

dualspiral:
https://github.com/SpongePowered/SpongeAPI/commit/b5076e9420e4377f6bd18df81be5dcb48831170a

PR #2393, opened 2021-09-05
feature/resources

Zidane:
For tracking purposes


src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,146 @@
...
+     * @return The metadata or empty if it doesn't exist
+     * @see NamedMetadataSections
+     */
+    <T> Optional<T> metadata(MetadataSection<T> section);
+
+    /**
+     * Creates a new {@link BufferedReader} from this resource's
+     * {@link InputStream}.
+     *
+     * @param charset The charset to use, usually utf-8
```
dualspiral: `{@link StandardCharsets#UTF_8 utf-8}`, just to give a bit of a pointer.

Are all Mojang resources UTF-8? (I mean, I'd hope so...)

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Loads the {@link Resource resource} at the given path, or throws an
+     * exception if it doesn't exist.
+     *
+     * <p>Resource implements {@link Closeable}, so remember to close it.</p>
+     *
+     * <p>Example:</p>
+     *
+     * <pre>
+     *     try (Resource res = resourceManager.load(path)) {
```
dualspiral: For this to be true, `Resource` should actually implement `AutoClosable`.

...though do we really need to provide Java 102 in our javadocs?

src/main/java/org/spongepowered/api/resource/ResourceReloadListener.java
```diff
@@ -0,0 +1,194 @@
...
+ *                 .thenAcceptAsync(thing -> {
+ *                      // apply things sync in game executor
+ *                 }, gameExecutor);
+ *         }
+ *     }
+ * </pre>
+ *
+ * @see RegisterResourceReloadListenerEvent
+ */
+public interface ResourceReloadListener {
```
dualspiral: Reading the javadocs, should it not contain two methods, one for the async stage and one for the sync stage, then just handle calling both methods elsewhere as appropriate? It seems to me that the javadocs say how reloading works, only to instruct you to do it yourself using `thenAccept` etc.

That said, I feel like we should just tell devlopers to use the factories and not to implement this themselves.

src/main/java/org/spongepowered/api/resource/metadata/NamedMetadataSection.java
```diff
@@ -0,0 +1,33 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.resource.metadata;
+
+import org.spongepowered.api.util.Nameable;
+import org.spongepowered.api.util.annotation.CatalogedBy;
+
+@CatalogedBy(NamedMetadataSections.class)
+public interface NamedMetadataSection<T> extends MetadataSection<T>, Nameable {
```
dualspiral: I think this was created simply for catalogued types, now objects (generally) don't have knowledge of their own keys, so with that in mind, is this necessary, or can we just stick `MetadataSection`s into the registry?

dualspiral (commented):
A couple of thoughts

src/main/java/org/spongepowered/api/resource/pack/PackDiscoverer.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * A pack discoverer will populate the {@link PackInformationManager} with {@link PackInformation}s
+ * and provide it a supplier to create a new {@link Pack}.
+ *
+ * @see RegisterPackDiscovererEvent
+ */
+public interface PackDiscoverer {
```
dualspiral: Before delving any further into this, to echo a sentiment on Discord - why might a plugin programmatically want to define a pack? If we can't give a solid answer to that, we should ditch this notion altogether.

src/main/java/org/spongepowered/api/resource/metadata/MetadataSection.java
```diff
@@ -22,38 +22,43 @@
...
- * Sponge {@link PluginContainer plugins}. The asset manager will attempt to find the
- * asset of the specified name at: <code>assets/&lt;plugin_id&gt;</code>
+ * A meta section is a well-known resource metadata object. Metadata is typically
+ * stored in a file with the resource and {@code .mcmeta} appended to the name.
+ * The file contains a json object with each key associated with a section.
+ *
+ * @param <T> The object type
  */
-public interface AssetManager {
+public interface MetadataSection<T> {
```
dualspiral: The thing that strikes me about this is that it's basically wrapping data stuff to access the metadata - we have a query and a deserialise method. Can this be `DataSerializable`, or perhaps more to the point, should it?

I'm not arguing for this case so much, more that I'm seeing parallels that we may want to exploit.

src/main/java/org/spongepowered/api/resource/pack/PackOrder.java
```diff
@@ -0,0 +1,47 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.resource.pack;
+
+import org.spongepowered.api.Sponge;
+
+public interface PackOrder {
```
dualspiral: We're not allowing custom packs to be loaded, right? Is there any need for this? "first" and "last" only have limited value when you have more than two packs.

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
```diff
@@ -0,0 +1,95 @@
...
+     * Gets the version compatibility of the pack.
+     *
+     * @return The version compatibility
+     */
+    PackVersion version();
+
+    /**
+     * @return The {@link PackOrder order}
+     */
+    PackOrder order();
```
dualspiral: See `PackOrder` comment - not sure if there is much value to have this here.

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Gets whether this pack is forced to be enabled at all times.
+     *
+     * @return True if forced enabled, false if disable-able
+     */
+    boolean forced();
```
dualspiral: Is there value to this?

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Gets whether this pack is order locked on top or bottom.
+     *
+     * @return True if order locked, false if order-able
+     * @see #order()
+     */
+    boolean locked();
```
dualspiral: Is there value to this?

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+package org.spongepowered.api.resource.pack;
+
+import net.kyori.adventure.text.Component;
+import org.spongepowered.api.resource.metadata.PackMetadata;
+import org.spongepowered.api.util.Nameable;
+
+/**
+ * Holds informational data about a {@link Pack}. It also functions as a
+ * factory to create a new instance.
```
dualspiral: > It also functions as a factory to create a new instance.

Not sure what this is referring to - a new `Pack`?

src/main/java/org/spongepowered/api/resource/pack/Pack.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * A resource pack or data pack can contain several {@link Resource resources}.
+ * A pack can contain both assets and data.
+ *
+ * @see <a href=http://minecraft.gamepedia.com/Resource_pack#Contents>
+ * Minecraft Wiki/Resource Packs
+ * </a>
+ */
+public interface Pack extends Nameable, Closeable {
```
dualspiral: `AutoClosable`

src/main/java/org/spongepowered/api/resource/Resource.java
```diff
@@ -0,0 +1,128 @@
...
+    ResourcePath path();
+
+    /**
+     * Returns the {@link InputStream} of this resource. Multiple calls to this
+     * method will not return a new object. To get a new object, get a new
+     * resource.
+     *
+     * @return The input stream
+     */
+    InputStream inputStream();
```
dualspiral, line 53: I think this is going to cause some confusion.

Basically, what you're saying here is that this resource is backed by the input stream here - that's okay. What's a little more complicated here is that, at least as I read it, other methods are going to affect this stream, such as `newReader`, `readString` etc. and this may well trip people up.

I'm not sure as of the best way to convey this. It may be better to split this interface into two interfaces, perhaps `Managed` (with the `read*` methods that can rewind the stream if necessary) and `Unmanaged` (this method), and perhaps drop the `newReader` method altogether, if people want that they can get that from an "unmanaged" stream by creating the reader by themselves. 

Even if the impl implements both interfaces, allowing for casting by advanced users, I feel like making this separation may benefit everyone, because you're separating the idea of using a stream from the idea of using managed methods.

Alternatively, we just go all out on documentation. That may be preferred.

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
Zidane: I don't see a harm in exposing this data personally.

src/main/java/org/spongepowered/api/resource/pack/PackInformation.java
Zidane: See comment below

src/main/java/org/spongepowered/api/resource/pack/PackOrder.java
Zidane: Basically the thought process in Mojang's structure is the first pack will load all the assets (and additional ones mind you) and the last pack will get final say.

They are both important in their own way...granted the first pack not as much as the last.

Unsure, maybe a plugin dev would want to know what the last pack is? (and not so much the first...)

We could toss it, read the updated javadocs and this comment and see what you think. I am thinking removal as well.

src/main/java/org/spongepowered/api/resource/Resource.java
Zidane: So you'll find in the next commit I tossed most of the methods.

They are heavy handed in the API and anyone can easily do it on their own (wrap the stream in a reader/etc). The bigger problem, I feel, is Vanilla.

Vanilla caches the input stream on `SimpleResource` which is the only implementation. That said...SimpleResource is created by demand from the `ResourceManager` when queried. The API is strange because `PackResources` in Vanilla does not give you `Resource` objects (it really ought to) and instead just hands you an `InputStream` while the API `PackContentsContainer` does.

I feel like we're close to beating this API framework in place, just some final bits and bobs to iron out.

dualspiral: It honestly feels daft to me that the `InputStream` is tied to the `Resource` in the way that it is... but that's the hand we're dealt.

I'm guessing closing the resource closes the stream. It would probably be good to note that on the class docs.

Zidane: @dualspiral I've considered tossing Resource out completely.

Due to the nature of the re-creation and the fact that using it in a try-with-resources quite literally nukes the actual worthwhile contents (namely the stream), I am finding less and less usage of the actual object.

src/main/java/org/spongepowered/api/resource/pack/PackOrder.java
dualspiral: > read the updated javadocs

Which ones?

I think I'll address the notion of removal here along with what I said about other methods. I fully accept there is no harm in leaving these informational methods, so if we chose to keep them, we keep them! However at the same time, I still am not sure on what is particularly useful about them on a consumer level -- it's quite possible I'm missing something though.

I say this as my mindset is "make minimal, then expand" rather than provide things that end up being not useful and possibly changed down the road. What if Mojang changes these packs to have a numeric order, or `isLocked` is suddenly ignored. Now, I'm not sure that will happen, but erring on the side of caution is often the way I look at things.

The likelihood of this generally being an issue is low, but it's a consideration that is what leads me down to this suggestion. If it turns out someone will find it useful, we can always add it in a point release.

src/main/java/org/spongepowered/api/resource/Resource.java
Zidane: Hmm or maybe not...bah this is difficult.

The issue is...`SimpleResource` doesn't open the stream up upon creation. While implementation shouldn't dictate API, we're dealing with nasty IO issues.

Thoughts?

dualspiral: I was thinking about that. `Resource` contains methods to packs and paths... but I'm guessing those were used to get the resource in the first place?

src/main/java/org/spongepowered/api/resource/ResourceManager.java
```diff
@@ -0,0 +1,75 @@
...
+public interface ResourceManager {
+
+    /**
+     * Loads the {@link Resource resource} at the given path, or throws an
+     * exception if it doesn't exist.
+     *
+     * @param path The path to the resource
+     * @return The resource
+     * @throws IOException If the resource could not be read
+     * @throws java.io.FileNotFoundException If the file does not exist
```
Lignium: `java.io.FileNotFoundException` -> `FileNotFoundException `

src/main/java/org/spongepowered/api/resource/ResourcePath.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Creates a resource path.
+     *
+     * @param namespace The namespace
+     * @param value     The value
+     * @return A new resource path
+     * @see ResourceKey#of(String, String)
+     */
+    static ResourcePath of(final String namespace, final String value) {
```
Lignium: Value or path? Below there is a method that names the key value as path.

PR #2394, opened 2021-09-15
Add EntityCategory

gabizou:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3536)

### DIFF
--- a/src/main/java/org/spongepowered/api/entity/EntityCategories.java
+++ b/src/main/java/org/spongepowered/api/entity/EntityCategories.java
@@ -1,0 +1,58 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/api/entity/EntityCategories.java
```diff
@@ -0,0 +1,54 @@
...
+package org.spongepowered.api.entity;
+
+import org.spongepowered.api.ResourceKey;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.registry.DefaultedRegistryReference;
+import org.spongepowered.api.registry.RegistryKey;
+import org.spongepowered.api.registry.RegistryTypes;
+
+public final class EntityCategories {
+
```
Zidane, line 34: Wrap these static lines in `@formatter` entries.

gabizou: done

PR #2401, opened 2021-11-27
Minor changes to AffectEntityEvent and ExplosionEvent.Detonate

Yeregorix:
Impl PR : [Sponge#3564](https://github.com/SpongePowered/Sponge/pull/3564)

Changes:
- Make `entities` and `filterAffectedLocations` unmodifiable.
- Make `filterEntities` and `filterAffectedLocations` return void.
- Fix event impl class hierarchy (enforces the presence of `SPAWN_TYPE` context key)
- Exclude `SpawnEntityEvent` from `SpongeEventFactoryTest#testCreate` because the context key is not mocked.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/entity/AffectEntityEvent.java
+++ b/src/main/java/org/spongepowered/api/event/entity/AffectEntityEvent.java
@@ -35,8 +35,6 @@ import org.spongepowered.api.util.annotation.eventgen.PropertySettings;
 import org.spongepowered.api.world.explosion.Explosion;
 import org.spongepowered.api.world.server.ServerLocation;
 
-import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.function.Predicate;
 
@@ -73,6 +71,7 @@ public interface AffectEntityEvent extends Event, Cancellable {
      *
      * @return The List
      */
+    @PropertySettings(requiredParameter = true, generateMethods = false)
     List<Entity> entities();
 
     /**
@@ -86,17 +85,7 @@ public interface AffectEntityEvent extends Event, Cancellable {
      * @return The entities removed from {@link #entities()}
      */
     default List<Entity> filterEntityLocations(Predicate<ServerLocation> predicate) {
-        final List<Entity> removedEntities = new ArrayList<>();
-
-        final Iterator<Entity> i = this.entities().iterator();
-        while (i.hasNext()) {
-            final Entity entity = i.next();
-            if (!entity.location().onServer().map(predicate::test).orElse(false)) {
-                i.remove();
-                removedEntities.add(entity);
-            }
-        }
-        return removedEntities;
+        return filterEntities(entity -> entity.location().onServer().map(predicate::test).orElse(false));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/event/entity/AffectEntityEvent.java
```diff
@@ -73,6 +71,7 @@
      *
      * @return The List
```
gabizou: Just clarify what this list is returning, whether it's the original entities, the new entities affected, or the filtered entities that won't be affected.

src/main/java/org/spongepowered/api/event/entity/AffectEntityEvent.java
```diff
@@ -109,17 +98,5 @@
...
-        while (i.hasNext()) {
-            final Entity entity = i.next();
-            if (!predicate.test(entity)) {
-                i.remove();
-                removedEntities.add(entity);
-            }
-        }
-        return removedEntities;
-    }
+    List<Entity> filterEntities(Predicate<Entity> predicate);
```
Yeregorix: To optimize I suggest returning void. I doubt many plugins need this list. For the few ones that need it, elements can still be collected in the predicate.

Yeregorix:
I just saw that [ExplosionEvent.Detonate](https://github.com/SpongePowered/SpongeAPI/blob/api-8/src/main/java/org/spongepowered/api/event/world/ExplosionEvent.java#L99) is in similar situation (`affectedLocations()` and `filterAffectedLocations(Predicate)`). Should I make this list unmodifiable too?

gabizou:
> Should I make this list unmodifiable too?

yep!

Yeregorix:
Looks like I have an issue with `event-impl-gen`. I ran `clean` then `genEventImpl` gradle tasks multiple times but the generated event doesn't inherit `AbstractDetonateEvent`.

Test code:
```java
    @Listener
    public void onDetonate(ExplosionEvent.Detonate event) {
        System.out.println(event.getClass().getName());
        System.out.println(event.getClass().getSuperclass().getName());
    }
```

Result:
```
org.spongepowered.api.event.ExplosionEvent$Detonate$Impl
org.spongepowered.api.event.impl.entity.AbstractAffectEntityEvent
```

Same problem for `SpawnEntityEvent`:
```
org.spongepowered.api.event.SpawnEntityEvent$Pre$Impl
org.spongepowered.api.event.impl.entity.AbstractAffectEntityEvent
```

It seems that the first superclass is always ignored despite the `@ImplementedBy` annotation.

Yeregorix:
I have probably identified the problem. This sounds like a priority issue with the `ImplementedBy` annotation. 

I can easily fix it by increasing priority of the annotation on the proper interface but I find it counterintuitive that `event-impl-gen` chooses the furthest annotation by default.

https://github.com/SpongePowered/event-impl-gen/blob/6f28baf367d2e95cbdd314cd6160468e2a0209d6/src/main/java/org/spongepowered/eventimplgen/processor/EventImplWriter.java#L184
`priority > max` would make more sense here IMO.

src/main/java/org/spongepowered/api/event/entity/AffectEntityEvent.java
dualspiral: I'm with you there I think. I dunno if @gabizou or @Zidane have an opposing view, but in general I think plugins are going to care about what _is_ going to be affected, not what _isn't_. I think we provide the original list of entities that are affected anyway, right? Plugins could also diff that if they need all entities affected by the event regardless.

Yeregorix: There is `List<EntitySnapshot> entitySnapshots()` which is immutable but must be called at `Order#PRE` to be initialized.

dualspiral:
https://github.com/SpongePowered/SpongeAPI/commit/3481b71ed1d7d9f9e5bcf8c55628ff1905da2ad6

PR #2403, opened 2021-12-23
Improve PotionEffect(.Builder) javadocs

qixils:
Fixes outright issues in some of the builder javadocs and elaborates on some non-obvious methods/parameters/etc

Some of these could also be backported to API 7 but I believe that's officially finished? Would be happy to PR to it anyway if desired

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
+++ b/src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
@@ -105,8 +105,8 @@ public interface PotionEffect extends DataSerializable {
     int amplifier();
 
     /**
-     * Gets if the potion effect is an ambient effect. The particles emitted
-     * by ambient effects are less opaque than otherwise.
+     * Gets whether or not this potion effect is an ambient effect.
+     * Ambient potions emit more translucent particles.
      *
      * @return If the effect should be ambient.
      */
@@ -175,8 +175,8 @@ public interface PotionEffect extends DataSerializable {
         Builder amplifier(int amplifier) throws IllegalArgumentException;
 
         /**
-         * Sets the potion effect to be ambient or not. The particles emitted
-         * by ambient effects are less opaque than otherwise.
+         * Sets whether the potion effect is ambient.
+         * Ambient potions emit more translucent particles.
          *
          * @param ambient Whether the potion effect is ambient
          * @return This builder, for chaining
@@ -184,7 +184,7 @@ public interface PotionEffect extends DataSerializable {
         Builder ambient(boolean ambient);
 
         /**
-         * Sets the potion effect to show particles when applied or not.
+         * Sets whether the potion effect should show particles when applied.
          *
          * @param showParticles Whether the potion effect will show particles
          * @return This builder, for chaining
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
```diff
@@ -57,11 +57,12 @@ static Builder builder() {
      * duration in ticks.
      *
      * @param type The potion type
-     * @param amplifier The amplifier
+     * @param amplifier The zero-indexed amplifier
```
dualspiral, line 60: While this is true, I don't think this itself is particularly helpful. It's zero indexed, so what does that mean for the amplifier itself? It's really an implementation note, not a descriptive note.

Would we be better served either:

* making this one-indexed and handling the change to zero based ourselves, or
* being a bit more descriptive about what the amplifier is, e.g the first level of a potion is amplifier zero, which each potion level having a higher amplifier (level-1, to be precise).

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
```diff
@@ -90,22 +92,23 @@ static PotionEffect of(Supplier<? extends PotionEffectType> type, int amplifier,
      * Gets the duration in ticks for which this potion effect
      * will apply for.
      *
-     * @return The duration.
+     * @return The duration in ticks.
```
dualspiral, line 95: This is a good spot - in API-8 this should really be returning `Ticks`.

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
```diff
@@ -169,7 +175,8 @@ default Builder potionType(Supplier<? extends PotionEffectType> potionEffectType
         Builder amplifier(int amplifier) throws IllegalArgumentException;
 
         /**
-         * Sets the potion effect to be ambient or not.
+         * Sets the potion effect to be ambient or not. The particles emitted
```
dualspiral: While you're here:

```
Sets whether the potion effect is ambient. Ambient potions emit particles that are more translucent than those from non-ambient potions.
```

dualspiral (commented):
What you did is totally valid and can be pulled as is, but it does raise some musings.

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
qixils: At the very least it should definitely be more descriptive, I'll get to that in a little bit. It would probably be sensible to make it one-indexed though it does sound like a kinda unnecessary breaking change when some better javadoc wording would probably suffice

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
qixils: I haven't worked with the Sponge impl before but I'd be happy to try and take a look into doing that :)

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
dualspiral: I agree with you there - just needs a bit more of a descriptive wording. I'll pull this in as is and quickly update it as I'm trying to get around tying up the API today, but thank you for the updates as is!

src/main/java/org/spongepowered/api/effect/potion/PotionEffect.java
dualspiral: I'll make the update myself as I'm going to push to get breaking changes all done today, but thanks for the offer!

PR #2404, opened 2022-01-01
[API9] World Generation

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3572)

API I did not add but to consider for the future:
 - Building custom `SurfaceRule`s and exposing all of the vanilla provided SurfaceRules.
   -  I only exposed the SurfaceRules for the main dimensions for now.
 - Building custom terrain `Shaper`s based on CubicSplines like vanilla.


src/main/java/org/spongepowered/api/world/generation/config/SurfaceRules.java
```diff
@@ -0,0 +1,32 @@
+package org.spongepowered.api.world.generation.config;
+
+import org.spongepowered.api.Sponge;
+
+public interface SurfaceRules {
```
Zidane: Just `SurfaceRule`

src/main/java/org/spongepowered/api/world/generation/config/noise/Shaper.java
```diff
@@ -22,35 +22,52 @@
...
-import java.util.Objects;
+public interface Shaper {
 
-public interface MultiNoiseConfig {
+    static Shaper overworld() {
+        return Sponge.game().factoryProvider().provide(Shaper.Factory.class).overworld(false);
+    }
+
+    static Shaper amplifiedOverworld() {
+        return Sponge.game().factoryProvider().provide(Shaper.Factory.class).overworld(true);
```
Zidane, line 36: Why does toggling overworld true mean amplified? If so, rename this, this is very confusing.

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
```diff
@@ -86,29 +94,31 @@ static Builder builder() {
 
         Builder defaultFluid(BlockState fluid);
 
+        Builder surfaceRule(SurfaceRules rule);
+
         Builder seaLevel(int y);
 
+        Builder mobGeneration(boolean disableMobGeneration);
```
Zidane: Inversed boolean?

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
```diff
@@ -76,6 +82,8 @@ static Builder builder() {
 
     boolean noodleCaves();
 
+    boolean legacyRandomSource();
```
Zidane, line 174: Is there any point to expose this?

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
```diff
@@ -86,29 +94,31 @@ static Builder builder() {
...
         Builder aquifers(boolean enableAquifers);
 
         Builder noiseCaves(boolean enableNoiseCaves);
 
         Builder oreVeins(boolean enableOreVeins);
 
-        Builder noodleCaves(boolean noodleCaves);
+        Builder noodleCaves(boolean enableNoodleCaves);
+
+        Builder randomSource(boolean useLegacyRandomSource);
```
Zidane, line 207: See question above regarding the getter.

src/main/java/org/spongepowered/api/world/generation/config/SurfaceRules.java
```diff
@@ -0,0 +1,32 @@
+package org.spongepowered.api.world.generation.config;
```
Zidane: header

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
Faithcaio: it mirrors what vanilla has in its datapacks

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
Faithcaio: From what I can see e.g. nether and end are using this but overworld does not.

src/main/java/org/spongepowered/api/world/generation/config/NoiseGeneratorConfig.java
Zidane: Perhaps but is still so confusing...I'd rather have the API be straightforward then graduate up this silly implementation detail.

PR #2414, opened 2022-02-12
Make ItemStack also implement HoverEventSource

zml2008:
**API** | [Impl](https://github.com/SpongePowered/Sponge/pull/3628)

This matches `ItemStackSnapshot`

### DIFF
--- a/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
+++ b/src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
@@ -25,6 +25,8 @@
 package org.spongepowered.api.item.inventory;
 
 import net.kyori.adventure.text.ComponentLike;
+import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.event.HoverEventSource;
 import org.spongepowered.api.Sponge;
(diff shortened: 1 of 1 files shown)


dualspiral (approved):

gabizou (approved):
LGTM

src/main/java/org/spongepowered/api/item/inventory/ItemStack.java
```diff
@@ -57,7 +59,7 @@
  * use {@link ValueContainer#get(Key)} in order to retrieve information regarding
  * this item stack.</p>
  */
-public interface ItemStack extends SerializableDataHolder.Mutable, ComponentLike {
+public interface ItemStack extends SerializableDataHolder.Mutable, ComponentLike, HoverEventSource<HoverEvent.ShowItem> {
```
gabizou, line 62: Not that I'm expecting people to be implementing an ItemStack, but is there a way to default implement the method itself being added?

zml2008: Not in any complete manner -- the hover event includes SNBT-serialized data from the stack's tag, which we don't reeealy expose in API

PR #2425, opened 2022-04-23
Expose BlockState string representation

Yeregorix:
[Sponge](https://github.com/SpongePowered/Sponge/pull/3668) | SpongeAPI


src/main/java/org/spongepowered/api/block/BlockState.java
```diff
@@ -157,6 +157,13 @@ default BlockState mirror(final Supplier<? extends Mirror> mirror) {
         return this.mirror(mirror.get());
     }
 
+    /**
+     * Gets a string representation of this block state, including the block type and the properties.
```
Zidane, line 161: I would add a further note to not confuse this with an `Object`'s traditional `toString` and to use the `toString` to the specific implementation's descriptor of this `Object`.

Something like that.

PR #2429, opened 2022-05-16
Add ServerWorld#saveAndFlush() method

Lignium:
**SpongeAPI** |  [Sponge](https://github.com/SpongePowered/Sponge/pull/3686)

Add separate method for saving world with forced flushing.
Required to avoid memory leaks during intensive modification of a huge number of chunks.
By analogy with the `/save-all [flush]` command, you can perform both simple save and save with flush here and now.


Zidane:
Need a lot more docs on this method to explain exactly what it will do and exactly what implementations are expected to honor.

Further, I'd prefer just a single `save` method that is given a `SaveMode` (name not final) (which could be an enum, similar to SerializationBehavior). The current save method can be deprecated and default implemented to call the new version with a value that is just save.

PR #2435, opened 2022-07-21
Refactor statistics to be implementable

Zidane:
- Remove Criterion from being in the registry as no registration process
exists to properly register new criterion. A factory is provided to
acquired Vanilla types
- Correct the registry for STATISTIC to be the keys (So a Registry of
Keys by Keys). This now follows Vanilla spec
- Expose the StatisticFormatter to the API so that mod formatters on
SpongeForge are properly provided to plugins. This is not complete, an
implementation issue on Sponge needs to be addressed to allow this to
work completely
- Correctly dictate that a Statistic IS a Criterion which now bridges
over into Scoreboard code correctly
- Introduce new API onto a StatisticCategory to allow adding new
statistics to a category
- Fixup javadocs in a few spots (this is not complete)


src/main/java/org/spongepowered/api/registry/RegistryTypes.java
```diff
@@ -200,9 +199,9 @@ public final class RegistryTypes {
 
     public static final DefaultedRegistryType<SoundType> SOUND_TYPE = RegistryTypes.minecraftKeyInGame("sound_event");
 
-    public static final DefaultedRegistryType<Statistic> STATISTIC = RegistryTypes.minecraftKeyInGame("custom_stat");
+    public static final DefaultedRegistryType<ResourceKey> STATISTIC = RegistryTypes.minecraftKeyInGame("custom_stat");
```
dualspiral, line 202: Can we just deprecate this and use `CUSTOM_STATISTIC` as a new registry type?

src/main/java/org/spongepowered/api/statistic/Statistic.java
```diff
@@ -24,45 +24,37 @@
...
+public interface Statistic<T> extends Criterion {
 
     /**
-     * Returns the objective {@link Criterion} for this statistic.
+     * Gets the {@link StatisticCategory category} this statistic resides under.
      *
-     * @return Objective criterion
+     * @return The category
      */
-    Optional<Criterion> criterion();
```
dualspiral, line 44: I'm kind of in two minds about this - we're breaking so yes, I see the logic as to why, but to minimise pain we could deprecate and return `Optional.of(this)`.

Then again, other stuff is breaking...

src/main/java/org/spongepowered/api/statistic/StatisticCategory.java
```diff
@@ -24,48 +24,48 @@
...
      *
-     * @param <T> The type
+     * @param value The value
+     * @return The instance
      */
-    interface Typed<T> extends StatisticCategory {
-
-        @Override
-        Collection<Statistic.TypeInstance<T>> statistics();
+    Statistic<T> statisticOrCreate(T value);
```
dualspiral, line 51: I don't like this name - `nounOrVerb` doesn't read well and I complained about it before - I think it was when I was doing the user manager.

`find`/`findOrCreate` feels better if we're trying to not do `get`, but is it clear it's to get a statistic then? In the `UserManager`, we used `loadOrCreate` - but that's correct here.

src/main/java/org/spongepowered/api/registry/RegistryTypes.java
Zidane: We could I suppose. However the deprecated field would literally do nothing. It can't even reference an actual registry object. The code would return a null type.

src/main/java/org/spongepowered/api/statistic/StatisticCategory.java
Zidane: getOrCreate? Doesn't say you're getting a statistic though. And the non-create method doesn't exist either. Tough situation, I had a hard time naming this.

ImMorpheus:

___

PR #2437, opened 2022-07-29
Add WallConnectionState

Lignium:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3729)

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -86,6 +86,7 @@ import org.spongepowered.api.data.type.StairShape;
 import org.spongepowered.api.data.type.StructureMode;
 import org.spongepowered.api.data.type.TropicalFishShape;
 import org.spongepowered.api.data.type.VillagerType;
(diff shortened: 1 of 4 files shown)


Faithcaio:
I did add those in API9 as Properties:

https://github.com/SpongePowered/SpongeAPI/blob/api-9/src/main/java/org/spongepowered/api/data/type/WallType.java
https://github.com/SpongePowered/SpongeAPI/blob/api-9/src/main/java/org/spongepowered/api/state/EnumStateProperties.java#L99

Lignium:
@Faithcaio I find the name `WallType` extremely unfortunate in this case. This is not about the wall as such, but about its connection, and there are 4 such connections. And of course, such an implementation is not complete. It does not allow the use of `BlockState#mergeWith()` and is also inconsistent with the generic `Keys#CONNECTED_DIRECTIONS`.

ImMorpheus:
Thanks for your contribution! This is being addressed in api-10 with a more generic approach.

Superseded by https://github.com/SpongePowered/SpongeAPI/pull/2440

PR #2438, opened 2022-07-30
Add Keys#HANGING for lanterns

Lignium:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3730)

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -961,11 +961,6 @@ public final class Keys {
     public static final Key<Value<Integer>> GROWTH_STAGE = Keys.key(ResourceKey.sponge("growth_stage"), Integer.class);
 
     /**
-     * Whether a lantern block is hanging.
-     */
-    public static final Key<Value<Boolean>> HANGING = Keys.key(ResourceKey.sponge("hanging"), Boolean.class);
-
-    /**
      * Whether an {@link ArmorStand}'s arms are visible.
      */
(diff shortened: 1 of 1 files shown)


gabizou (approved):

Faithcaio:
API9 https://github.com/SpongePowered/SpongeAPI/blob/api-9/src/main/java/org/spongepowered/api/state/BooleanStateProperties.java#L81

Zidane:
@Faithcaio 

Going to merge this for api-8 just to allow this version to have this feature. When you merge this upstream, you'll need to omit this commit. Or deprecate the data entry and mention to use the state properties?

Lignium:
@Zidane If I understand you correctly, why are you suggesting to omit this in API 9? Isn't the Data API the preferred way to work with data? The State Property API is, and has always been, a secondary (fallback) way to get data when the data is not reflected in the Data API. Let me remind you that the more data is implemented using the Data API, the better methods such as `mergeWith`, `values`, etc. will work.

You also forgot to merge the [implementation](https://github.com/SpongePowered/Sponge/pull/3730).

Zidane:
@Lignium 

I didn't forget to merge it, just haven't had the chance to merge your impl PR and bump the API ref. If you wanna help me out further, have your impl PR bump the API ref then it is a clean pull.

Also I only pointed it out to @Faithcaio in-case there was some reason this cannot remain in the data system. If no such reason exists, this will be a clean merge upstream.

Faithcaio:
Its more work to have the properties as Data atm. as the properties get generated automatically.
Also `Keys` becomes quite cluttered.

So the best option would be to improve how to register Properties as Data and then deprecate the properties entirely.

PR #2442, opened 2022-09-02
Add Keys#PREVIOUS_GAME_MODE

Lignium:
[**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3757)]

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -934,7 +934,7 @@ public final class Keys {
     public static final Key<Value<Ticks>> FUSE_DURATION = Keys.key(ResourceKey.sponge("fuse_duration"), Ticks.class);
 
     /**
-     * The {@link GameMode} a {@link Humanoid} has.
+     * The {@link GameMode} a {@link ServerPlayer} has.
      */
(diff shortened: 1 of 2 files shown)


Zidane:
@Lignium 

I wonder if we should allow this key to be set? I only bring this up because setting it effectively feels like breaking the contract of what it is?

Lignium:
@Zidane My authorization plugin manipulates game mode, adventure is set for the duration of authorization. And there was a problem, the key combination F3 + N broke, the previous game mode turned out to be the mode that temporarily set the authorization plugin, not at all what was expected. Instead of switching back and forth between creative and spectator, I get thrown into adventure mode and falling from a height.

Zidane:
My point is I'd expect setting of GAME_MODE via key would change PREVIOUS_GAME_MODE but no one can rely on PREVIOUS_GAME_MODE actually being truthful if anyone can simply go set it via API contract.

Lignium:
One option is to make it read-only, and with two Keys#GAME_MODE set in a row, I can set both values.

gabizou (approved):

gabizou:
The way I see it is that while game modes can change, we easily provide the DataTransaction including the Keys.PREVIOUS_GAME_MODE, since it can very well be changing.

PR #2443, opened 2022-10-05
Add FluidState to LocatableBlock

gabizou:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3768)

At some point I had this in my workspace, not sure why, but saw an opportunity to improve the quality of usage for fluid states during various ChangeBlockEvent.Pre


ImMorpheus (dismissed):
LGTM! @gabizou can you target this to api-10 ? do you need help ?

PR #2444, opened 2022-10-16
Expose adventure's new ComponentLogger to plugins

zml2008:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3779)

As a consequence, SLF4J is an API dependency again (it's always existed at some level in the implementations). We don't really want that, but how long can we avoid it if MC's switch to slf4j leads to developers expecting SLF4J to be present?

Now, we currently expose component logging functionality on SpongeVanilla by injecting into log4j -- that allows passing components to a logger the exact same way from the plugin developer side. This takes a fair bit of mucking about with Log4J internals, though so it is vulnerable to being disrupted by changes on Mojang's side. This functionality also hasn't been implemented on SpongeForge, and I am unsure exactly how much work it would be to do so.

Use of Adventure's ComponentLogger wrapper results in less work on our side, but adds more places for plugin developers to be confused about which logging interfaces should be used.

Thoughts?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,6 +33,7 @@ dependencies {
     val junitVersion: String by project
     val mockitoVersion: String by project
     val pluginSpiVersion: String by project
+    val slf4jVersion: String by project
 
     // Directly tied to what's available from Minecraft
     api("org.apache.logging.log4j:log4j-api:$log4jVersion")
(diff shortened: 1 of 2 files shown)


ImMorpheus:

___

PR #2446, opened 2022-11-01
Fix contribution guidelines link

QuickWrite:
The previous link was liking to a dead page. This is now changed to always point to the newest version of the contribution guidelines.

Even though this is currently merging into `api-8` this problem exists in the other branches as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,4 +2,4 @@
 
 We will always have a need for developers to help us improve SpongeAPI. There is no such thing as a perfect project and things can always be improved. If you are a developer and are interested in helping then please do not hesitate. Just make sure you follow our guidelines.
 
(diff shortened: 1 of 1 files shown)


QuickWrite:
I am currently realizing that this should rather be an issue. If it is needed I will recreate this as an issue.

ImMorpheus:
Sorry for the late response,

This was reported in the minor issue list at https://github.com/SpongePowered/SpongeAPI/issues/1000#issuecomment-963911267

There are a few more links that need to be updated in the readme.
Thanks for your contribution!

QuickWrite:
> Sorry for the late response,
> 
> This was reported in the minor issue list at [#1000 (comment)](https://github.com/SpongePowered/SpongeAPI/issues/1000#issuecomment-963911267)
> 
> There are a few more links that need to be updated in the readme. Thanks for your contribution!

No problem. I haven't seen the issue itself so it was my mistake. 😄

PR #2448, opened 2023-01-08
build: require java 17+ and update readme

ImMorpheus:
MC 1.18+ require java 17+

Update readme accordingly and replace old links

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -22,6 +22,6 @@ plugins {
     id("org.gradle.toolchains.foojay-resolver-convention") version ("0.3.0")
 }
 
-if (JavaVersion.current() < JavaVersion.VERSION_11) {
-    throw GradleException("SpongeAPI requires at least Java 11 to build, but you have ${JavaVersion.current()}.")
+if (JavaVersion.current() < JavaVersion.VERSION_17) {
+    throw GradleException("SpongeAPI requires at least Java 17 to build, but you have ${JavaVersion.current()}.")
 }
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -21,7 +21,3 @@ pluginManagement {
 plugins {
     id("org.gradle.toolchains.foojay-resolver-convention") version ("0.3.0")
 }
-
-if (JavaVersion.current() < JavaVersion.VERSION_17) {
-    throw GradleException("SpongeAPI requires at least Java 17 to build, but you have ${JavaVersion.current()}.")
-}
--- a/.github/workflows/check-spotless.yaml
+++ b/.github/workflows/check-spotless.yaml
@@ -9,4 +9,6 @@ on:
 jobs:
   call-check:
     uses: SpongePowered/.github/.github/workflows/shared-check-spotless.yaml@master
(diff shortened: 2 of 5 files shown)


settings.gradle.kts
```diff
@@ -22,6 +22,6 @@ plugins {
     id("org.gradle.toolchains.foojay-resolver-convention") version ("0.3.0")
 }
 
-if (JavaVersion.current() < JavaVersion.VERSION_11) {
-    throw GradleException("SpongeAPI requires at least Java 11 to build, but you have ${JavaVersion.current()}.")
+if (JavaVersion.current() < JavaVersion.VERSION_17) {
```
zml2008: The Java version required to run gradle does not necessarily have to be the Java version used to compile -- this is just for buildscript plugins (iirc originally checkstyle) to run.

zml2008 (changes requested):

zml2008:
This does however bring up the issue of the CI running gradle with Java 11, which is a bit redundant since 17 will have to be downloaded by Gradle anyways.

You should be able to add a `runtime_version: 17` to the `with:` block on all the shared workflow actions used in SpongeAPI to avoid that. (as defined [in the shared workflows](https://github.com/SpongePowered/.github/blob/master/.github/workflows/shared-check-spotless.yaml#L7))

settings.gradle.kts
ImMorpheus: Should we optimize for the majority ? I'm thinking most people clone the repo to tinker/contribute so they probably want to build it. Any thoughts ?

ImMorpheus:
> This does however bring up the issue of the CI running gradle with Java 11, which is a bit redundant since 17 will have to be downloaded by Gradle anyways.
> 
> You should be able to add a `runtime_version: 17` to the `with:` block on all the shared workflow actions used in SpongeAPI to avoid that. (as defined [in the shared workflows](https://github.com/SpongePowered/.github/blob/master/.github/workflows/shared-check-spotless.yaml#L7))

Thanks, had to dig into the common workflows :smile:

settings.gradle.kts
zml2008: Well, you can build either way -- if you're running on an older Java version, Gradle will try to detect the Java 17 JVM or if necessary download it -- it's easier to be more lenient rather than making devs muck around with their environment variables.

ImMorpheus: Ah, didn't know that. Followup question: what is the check used for ? Woudn't it be best to remove and let gradle handle the java version then ? I'm not sure why we are specifically enforcing 11

zml2008: I believe I added that check because newer versions of Checkstyle require J11 and Gradle didn't yet support toolchains on the checkstyle task at the time (it does now, so that could be configured to use the project toolchain instead and this check could be removed)

ImMorpheus: I see we are not using the checkstyle plugin directly but we're relying on indra which has the toolchain version set (https://github.com/SpongePowered/SpongeAPI/blob/api-10/build.gradle.kts#L258). Shouldn't it take care of it ?

PR #2452, opened 2023-01-29
Fix login NPE with fluid states 

ImMorpheus:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3824)


ImMorpheus:
Revert https://github.com/SpongePowered/SpongeAPI/commit/3e6014a0e43b1b4fb7fe435de300ded452e75486

PR #2456, opened 2023-04-14
Expose protocol version for status pings

### LINKED ISSUE #2251: Add option to set a custom version in the ClientPingServerEvent
Ok so i was looking through the API lately and noticed that the option to set a minecraft servers version is missing. Doing this would allow us as an example to let the client think he is outdated and so always displaying the current minecraft version. And also setting a custom version name for the client to display. (This can be even coloured instead of the default red colour). 

Here is an example, which did make use of this functionality of minecraft: 

I hope this feature can get added to this awesome API. 

You can by the way already get the version that is sent, but not edit it yet:
https://github.com/SpongePowered/SpongeAPI/blob/66629378af29e796271ec09ef4b18ca46e8300ef/src/main/java/org/spongepowered/api/network/status/StatusClient.java#L50 

Useful links for devs:
https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/event/server/ClientPingServerEvent.java
https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/network/status/StatusClient.java

stephan-gh:
Historically we considered the Minecraft protocol versions as "implementation detail" that should not be exposed in SpongeAPI.

However, since the addition of the status ping API 8 years ago (#367) the protocol version still exists exactly the same way, and there are use cases for checking it (identifying the exact client version) and modifying it (making the server appear as incompatible to clients).

Right now plugins have to resort to using implementation-specific code for this, which is complicated and now hopelessly broken for api-10 (due to internal changes in the Minecraft code).

Make it possible to check and modify the server version to fix this once and for all.

Closes #2251

### DIFF
--- a/src/main/java/org/spongepowered/api/MinecraftVersion.java
+++ b/src/main/java/org/spongepowered/api/MinecraftVersion.java
@@ -45,6 +45,16 @@ public interface MinecraftVersion extends Comparable<MinecraftVersion> {
     String name();
 
     /**
+     * Gets the protocol version of this Minecraft version.
+     *
+     * @implNote This should be interpreted together with {@link #isLegacy()},
+     * since legacy clients use a different protocol version numbering scheme.
+     * @return The protocol version
+     * @see <a href="https://minecraft.fandom.com/wiki/Protocol_version">Protocol version (Minecraft Wiki)</a>
+     */
+    int protocolVersion();
+
+    /**
      * Returns whether this version is an older version that doesn't support
      * all of the features in {@link StatusResponse}. These versions are only
      * supported for the {@link ClientPingServerEvent}, normally they should not be
@@ -62,4 +72,13 @@ public interface MinecraftVersion extends Comparable<MinecraftVersion> {
      * @return The data version
      */
     OptionalInt dataVersion();
+
+    @Override
+    default int compareTo(MinecraftVersion o) {
+        final int result = Boolean.compare(this.isLegacy(), o.isLegacy());
+        if (result != 0) {
+            return result;
+        }
+        return this.protocolVersion() - o.protocolVersion();
+    }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/MinecraftVersion.java
```diff
@@ -62,4 +72,13 @@ public interface MinecraftVersion extends Comparable<MinecraftVersion> {
...
      */
     OptionalInt dataVersion();
+
+    @Override
+    default int compareTo(MinecraftVersion o) {
+        final int result = Boolean.compare(isLegacy(), o.isLegacy());
+        if (result != 0) {
+            return result;
+        }
+        return protocolVersion() - o.protocolVersion();
```
ImMorpheus: qualify method calls with `this`

src/main/java/org/spongepowered/api/event/server/ClientPingServerEvent.java
```diff
@@ -92,6 +93,9 @@ interface Response extends StatusResponse {
          */
         void setHidePlayers(boolean hide);
```
ImMorpheus, line 95: As mentioned in the implementation PR I think we should add a way to let plugins check if players are hidden.

Currently there is a setter (`setHidePlayers`) but no getter

ImMorpheus (changes requested):

src/main/java/org/spongepowered/api/event/server/ClientPingServerEvent.java
ImMorpheus: I looked at the implementation before reading the API :facepalm: 

This should be fine:

```
         * <p>Use {@link #players()}.{@link Optional#isPresent() isPresent()} to
         * check if the players are already hidden.</p>
```

src/main/java/org/spongepowered/api/MinecraftVersion.java
stephan-gh: Fixed, thanks

ImMorpheus (approved):

PR #2466, opened 2023-10-05
Add region files API

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/3903)

Adds API to handle region files,

as discussed, adding access to offline chunks instead.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/server/ChunkManager.java
+++ b/src/main/java/org/spongepowered/api/world/server/ChunkManager.java
@@ -30,9 +30,12 @@ import org.spongepowered.api.util.annotation.DoNotStore;
 import org.spongepowered.api.world.ChunkRegenerateFlag;
 import org.spongepowered.api.world.ChunkRegenerateFlags;
 import org.spongepowered.api.world.chunk.WorldChunk;
+import org.spongepowered.math.vector.Vector2i;
 import org.spongepowered.math.vector.Vector3i;
 
+import java.nio.file.Path;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


Yeregorix:
Aren't regions an implementation detail? A custom implementation could theoretically stores worlds in a different format (not region based). It could even be in memory or in a database, without files.

pschichtel:
@Yeregorix I fully agree, but the API already leaks this detail currently (ServerWorld.directory).

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
```diff
@@ -235,4 +236,15 @@ default boolean restoreSnapshot(final Vector3i position, final BlockSnapshot sna
...
+    /**
+     * Returns a stream of existing chunk positions for this world.
+     * <p>
+     *     Note that consuming the stream may be slow.
+     *     The stream must be closed or fully consumed otherwise file handles may stay open.
+     * </p>
+     *
+     * @return The stream of existing chunk positions.
+     */
+    Stream<Vector3i> chunkPositions();
```
Faithcaio, line 249: so how about this instead?

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
```diff
@@ -246,4 +249,65 @@ default CompletableFuture<Boolean> regenerateChunk(final Vector3i chunkPosition,
...
+     */
+    Path regionDirPath();
+
+    /**
+     * Returns the path of the region file containing the given {@link WorldChunk}.
+     *
+     * @param chunk The {@link WorldChunk} for which to acquire the region file path
+     * @return The path of the region file for the given chunk
+     */
+    Path regionFilePath(WorldChunk chunk);
```
gabizou, line 265: I'd almost want to have a StorageHandler interface to deal with this so that it's not directly exposed in the API, but can be something exposed for "vanilla" kinds.

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
gabizou: This seems fine

src/main/java/org/spongepowered/api/world/server/ChunkManager.java
gabizou: It's not always guaranteed that the world has region files (like mods) that could be storing chunks in a different region handler that we simply won't be able to directly interface with.

src/main/java/org/spongepowered/api/world/server/ServerWorld.java
```diff
@@ -235,4 +237,29 @@ default boolean restoreSnapshot(final Vector3i position, final BlockSnapshot sna
...
+     * <p>
+     * Note that consuming the stream is very slow, consider consuming it asynchronously.
+     * <p>
+     * The stream must be closed or fully consumed otherwise file handles may stay open.
+     * <p>
+     * You should not modify data of currently loaded chunks.
+     *
+     * @return The stream of existing chunks.
+     */
+    Stream<OfflineChunk> offlineChunks();
```
Faithcaio, line 263: and this if you want the chunk data too

PR #2480, opened 2023-12-18
Add Ticks#infinite

aromaa:
Added `Ticks#infinite` and updated the documentation where providing it would not make sense.


aromaa:
Some other alternatives that I don't quite like are that we could change the `PotionEffect#duration` type to `Optional<Ticks>` but that's quite vague. Would that imply that not specifying the duration in the builder mean that it creates a infinite one? That however could be solved with another method `PotionEffect.Builder#infinite`.

Adding new more concreate type like `TicksOrInfinite` is basically the same approach as the `Optional<Ticks>` one as we couldn't implement the `Ticks` interface or we are back at square one. Also adds more surface area for not much added clarity?

PR #2496, opened 2024-04-23
[NO MERGE] Contextual Data API

aromaa:
[Sponge](https://github.com/SpongePowered/Sponge/pull/3995) | **SpongeAPI**

Open questions?
- [ ] Server as DataPerspective? One could implement way to hide specific entity data like health from clients globally.
- [ ] This could also be used to change the world properties send to the client like weather by taking advantage of the World -> Entity data direction. Do we want that?
- [ ] How do we implement collection operations? What happen happen when you do `offerSingle` to the data perspective and the base entity already has elements. Append? Create new list with just that single item?
- [ ] Do we want to allow plugins to provide more info how to solve merge conflicts?
  - [ ] Provide types to provide the operation type? Like `Value<Integer>` could have addition, multiplication, etc.
  - [ ] Allow that same concept for values with collections?

### DIFF
--- a/src/main/java/org/spongepowered/api/data/DataPerspective.java
+++ b/src/main/java/org/spongepowered/api/data/DataPerspective.java
@@ -1,0 +1,13 @@
+package org.spongepowered.api.data;
+
+import org.spongepowered.api.data.value.ValueContainer;
+import org.spongepowered.plugin.PluginContainer;
+
+public interface DataPerspective {
+
+    Iterable<DataPerspective> perceives();
+
+    ValueContainer getDataPerception(DataPerspective perspective);
+
+    DataHolder.Mutable createDataPerception(PluginContainer plugin, DataPerspective perspective);
+}
--- a/src/main/java/org/spongepowered/api/data/DataPerspective.java
+++ b/src/main/java/org/spongepowered/api/data/DataPerspective.java
@@ -1,3 +1,27 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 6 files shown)


src/main/java/org/spongepowered/api/data/DataPerspective.java
```diff
@@ -0,0 +1,37 @@
...
+package org.spongepowered.api.data;
+
+import org.spongepowered.api.data.value.ValueContainer;
+import org.spongepowered.plugin.PluginContainer;
+
+public interface DataPerspective {
+
+    Iterable<DataPerspective> perceives();
+
+    ValueContainer getDataPerception(DataPerspective perspective);
```
Zidane, line 34: Should just be `dataPerception(...)` no?

src/main/java/org/spongepowered/api/data/DataPerspectiveResolver.java
```diff
@@ -0,0 +1,55 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.data;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.spongepowered.api.data.value.Value;
+
+public interface DataPerspectiveResolver<V extends Value<E>, E> {
```
Zidane, line 30: Is this interface meant to be implemented? (DataRegistration below gives that impression).

If so, maybe it is best to make this abstract? I do not like a method that returns the key that isn't enforced to always return the same key (the instance could change).

aromaa: Yes, this is at the moment required to light up contextual data, otherwise we refuse to apply the data. I followed the same style we use for `DataProvider` already, that also has getter for the key.

But it might be more beneficial to try to intergrade this whole concept to be part of `DataProvider` to make them always be contextual aware, but that then touches a lot of core data API surface area.

Zidane: it's up to you as it's a commitment on your part but if you are up to it, I think this is a spot where we go all in and make DataProvider and the data API as a whole contextual aware. It was always meant to be, no one sat down to do it was always the issue.

Until now :).

PR #2497, opened 2024-04-25
Rework ChunkEvent

aromaa:
[Sponge](https://github.com/SpongePowered/Sponge/pull/3997) | **SpongeAPI**

Reworks the ChunkEvent events to take in account that vanilla has split the block and entity storage. I kept the `BlockChunk` as read-only because it needs more work to implement to be able to mutate it at that state, likely not something we are going to address anytime soon.

### DIFF
--- a/src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
+++ b/src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
@@ -82,9 +82,9 @@ public interface ChunkEvent extends Event {
             /**
              * Gets the {@link WorldChunk chunk} block volume.
              *
-             * @return The block volume
+             * @return The block chunk
              */
-            BlockChunk blockVolume();
+            BlockChunk chunk();
         }
 
         /**
@@ -96,14 +96,14 @@ public interface ChunkEvent extends Event {
              * Called before the {@link WorldChunk chunk} block data is saved. Cancelling this
              * will prevent any of the chunk's block data being written to it's storage container.
              */
-            interface Pre extends Save, Cancellable {
+            interface Pre extends Blocks.Save, Cancellable {
 
                 /**
                  * Gets the {@link WorldChunk chunk} block volume.
                  *
-                 * @return The block volume
+                 * @return The block chunk
                  */
-                BlockChunk blockVolume();
+                BlockChunk chunk();
             }
 
             /**
@@ -112,7 +112,7 @@ public interface ChunkEvent extends Event {
              * <p>
              * Depending on the implementation, this event may be called off-thread.
              */
-            interface Post extends Save {}
+            interface Post extends Blocks.Save {}
         }
     }
 
@@ -133,9 +133,9 @@ public interface ChunkEvent extends Event {
             /**
              * Gets the {@link WorldChunk chunk} entity volume.
              *
-             * @return The entity volume
+             * @return The entity chunk
              */
-            EntityChunk entityVolume();
+            EntityChunk chunk();
         }
 
         /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
```diff
@@ -64,58 +66,131 @@ interface WorldScoped extends ChunkEvent {
...
+             * will prevent any of the chunk's entity data being written to it's storage container.
+             */
+            interface Pre extends Save, Cancellable {
+
+                /**
+                 * Gets the {@link WorldChunk chunk} entity volume.
+                 *
+                 * @return The entity volume
+                 */
+                EntityChunk entityVolume();
```
Zidane: Maybe just `chunk`? Same for the above.

src/main/java/org/spongepowered/api/world/chunk/BlockChunk.java
```diff
@@ -0,0 +1,30 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.chunk;
+
+import org.spongepowered.api.world.volume.block.BlockVolume;
+
+public interface BlockChunk extends BlockVolume {
```
gabizou, line 29: What is the significance of this? Why not just a BlockVolume?

src/main/java/org/spongepowered/api/world/chunk/EntityChunk.java
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world.chunk;
+
+import org.spongepowered.api.world.volume.entity.EntityVolume;
+
+/**
+ * An entity chunk is a portion of a {@link WorldChunk}.
+ */
+public interface EntityChunk extends EntityVolume.Modifiable<EntityChunk> {
```
gabizou, line 32: We may as well split modifiable if it means it's not spawnable, or am I missing something?

src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
```diff
@@ -64,58 +66,131 @@ interface WorldScoped extends ChunkEvent {
...
+    interface Entities extends WorldScoped {
+
+        /**
+         * Called when an entity data of {@link WorldChunk chunk} is loaded.
+         * This can be called outside the {@link World#engine() main} {@link Thread}.
+         * It is NOT safe to perform modifications to the {@link World} or via
+         * {@link org.spongepowered.api.world.server.ServerLocation} as this could
+         * result in a deadlock.
+         */
+        interface Load extends Entities {
```
gabizou, line 131: What is the rammification of calling `spawnEntity` on the `EntityChunk`?

src/main/java/org/spongepowered/api/world/chunk/BlockChunk.java
aromaa: Just to align with the `EntityChunk` and it allows us to easily change it to be `BlockVolume.Modifiable` in the future without it being binary breaking change.

src/main/java/org/spongepowered/api/world/chunk/EntityChunk.java
aromaa: Entities are modifiable, thats implemented. Blocks are special case due to all of your pipeline tracking and its not obvious how we would want to alter that to make it possible.

src/main/java/org/spongepowered/api/event/world/chunk/ChunkEvent.java
aromaa: That works and is safe. The entity is appended to be part of the list of entities to be loaded.

PR #2504, opened 2024-05-15
Open container from type & some fixes

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4010)


src/main/java/org/spongepowered/api/block/entity/carrier/chest/Chest.java
```diff
@@ -64,7 +64,7 @@ public interface Chest extends NameableCarrierBlockEntity {
      * @return The attachment type of this chest.
      */
     default Value.Mutable<ChestAttachmentType> attachmentType() {
-        return this.requireValue(Keys.CHEST_ATTACHMENT_TYPE).asMutable();
+        return this.block().requireValue(Keys.CHEST_ATTACHMENT_TYPE).asMutable();
```
gabizou: I know this is subtle, but why?

MrHell228: Because `Keys.CHEST_ATTACHMENT_TYPE` is registered for `BlockState`, not `BlockEntity`. And currently this method simply throw an exception

Faithcaio: Note that we have code (`SpongeDataHolder#impl$delegateDataHolder`) that could make this work on BlockEntity too.
e.g. 
https://github.com/SpongePowered/Sponge/blob/api-12/src/main/java/org/spongepowered/common/world/server/SpongeServerLocation.java#L280

MrHell228: Oh, that's cool. I guess it would be better to use this method then

MrHell228: Reverted change in `Chest` file, in impl made `BlockEntity` delegate `DataHolder` to its `#block()` and `#block().type()`.
Also reverted `Container#type()` change since it's breaking (and in api-12 Faithcaio fixed this).

PR #2513, opened 2024-06-02
Add Viewer#sendBlockProgress

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4023)

### DIFF
--- a/src/main/java/org/spongepowered/api/effect/Viewer.java
+++ b/src/main/java/org/spongepowered/api/effect/Viewer.java
@@ -161,4 +161,60 @@ public interface Viewer extends Audience {
      */
     void resetBlockChange(int x, int y, int z);
 
+    /**
+     * Sends a client-only block breaking progress.
+     *
+     * <p>In vanilla breaking progress will not be rendered if player
+     * is further than 32 blocks from given position.<br>
+     * Sent breaking progress expires on client 400 ticks after receiving.</p>
+     *
+     * @param position The position
+     * @param progress The breaking progress from 0 to 1 (1 excluded)
+     */
+    default void sendBlockProgress(final Vector3i position, final float progress) {
+        Objects.requireNonNull(position, "position");
+        this.sendBlockProgress(position.x(), position.y(), position.z(), progress);
+    }
+
+    /**
+     * Sends a client-only block breaking progress.
+     *
+     * <p>In vanilla breaking progress will not be rendered if player
+     * is further than 32 blocks from given position.<br>
+     * Sent breaking progress expires on client 400 ticks after receiving.</p>
+     *
+     * @param x The x position
+     * @param y The y position
+     * @param z The z position
+     * @param progress The breaking progress from 0 to 1 (1 excluded)
+     */
+    void sendBlockProgress(int x, int y, int z, float progress);
+
+    /**
+     * Resets the client's view of the provided position to actual
+     * breaking progress.
+     *
+     * <p>This is useful for resetting what the client sees
+     * after sending a {@link #sendBlockProgress block progress}.</p>
+     *
+     * @param position The position
+     */
+    default void resetBlockProgress(final Vector3i position) {
+        Objects.requireNonNull(position, "position");
+        this.resetBlockProgress(position.x(), position.y(), position.z());
+    }
+
+    /**
+     * Resets the client's view of the provided position to actual
+     * breaking progress.
+     *
+     * <p>This is useful for resetting what the client sees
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/effect/Viewer.java
```diff
@@ -161,4 +161,60 @@ default void resetBlockChange(final Vector3i position) {
...
+     * breaking progress.
+     *
+     * <p>This is useful for resetting what the client sees
+     * after sending a {@link #sendBlockProgress block progress}.</p>
+     *
+     * @param x The x position
+     * @param y The y position
+     * @param z The z position
+     */
+    void resetBlockProgress(int x, int y, int z);
```
ImMorpheus, line 210: This is leaking impl details, please remove this method

src/main/java/org/spongepowered/api/effect/Viewer.java
```diff
@@ -161,4 +161,60 @@ default void resetBlockChange(final Vector3i position) {
...
+        Objects.requireNonNull(position, "position");
+        this.sendBlockProgress(position.x(), position.y(), position.z(), progress);
+    }
+
+    /**
+     * Sends a client-only block breaking progress.
+     *
+     * <p>In vanilla breaking progress will not be rendered if player
+     * is further than 32 blocks from given position.<br>
+     * Sent breaking progress expires on client 400 ticks after receiving.</p>
```
ImMorpheus: Same as below, this is impl details

ImMorpheus (changes requested):

src/main/java/org/spongepowered/api/effect/Viewer.java
MrHell228: Wdym it's impl detail. It's must have method if we make limits (from 0 to 1) for progress arg in sendBlockProgress

ImMorpheus: The server doesn't need any knowledge of the current progress. This is leaking impl details by forcing implmentations to track the progress

MrHell228: Server doesn't have any knowledge of the current progress. It resets only progress that was sent by `sendBlockProgress()` so the only knowledge server has is that some plugins used `sendBlockProgress()`

src/main/java/org/spongepowered/api/effect/Viewer.java
MrHell228: Sponge doesn't (and can't) do anything with this because it's client stuff. But for plugin devs this info could be useful so they don't have to go to sources and I'm pretty sure api has a bunch of these `by default` additions.
From what I just found:
https://github.com/SpongePowered/SpongeAPI/blob/22c65048cba23a68f12e83a6cf98225be5ab7088/src/main/java/org/spongepowered/api/data/Keys.java#L742
https://github.com/SpongePowered/SpongeAPI/blob/42269e2110de472a40447c27d0f41a542900432f/src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageType.java#L49
https://github.com/SpongePowered/SpongeAPI/blob/b2216ba4ebe77b96e059441fbd2a067746e948bb/src/main/java/org/spongepowered/api/block/entity/StructureBlock.java#L67

src/main/java/org/spongepowered/api/effect/Viewer.java
Yeregorix: I think `resetBlockProgress` is necessary because we need a way to remove the crack.
Notes:
1. setting progress to 0 does not remove, it just means you began cracking the block.
2. as discussed on discord, we will not track progress, however we need to keep somewhere a map of blockpos to destruction id

src/main/java/org/spongepowered/api/effect/Viewer.java
```diff
@@ -161,4 +161,60 @@ default void resetBlockChange(final Vector3i position) {
...
+     *
+     * @param position The position
+     * @param progress The breaking progress from 0 to 1 (1 excluded)
+     */
+    default void sendBlockProgress(final Vector3i position, final float progress) {
```
Yeregorix: Should we use double instead of float?

MrHell228: Maybe just for comfort so you could do 0.1 instead of 0.1F

Faithcaio (approved):

PR #2516, opened 2024-06-04
Change BlockEntity#setValid() to remove()

MrHell228:


src/main/java/org/spongepowered/api/block/entity/BlockEntity.java
```diff
@@ -56,31 +56,16 @@
...
-     * <p>If the block entity is invalid, no processing will be done on this
-     * {@link BlockEntity} until it either becomes valid or is reset on the next
-     * tick.</p>
-     *
-     * <p>If the block entity is valid, then processing can continue and this
-     * {@link BlockEntity} will not be reset on the next tick.</p>
-     *
-     * @param valid True if the block entity should be validated, or false if
-     *     it should be invalidated
+     * Removes this block entity from world
```
aromaa:
```suggestion
     * Removes this block entity from the world and its corresponding block.
```

aromaa (approved):
LGTM, just small NIT.

PR #2517, opened 2024-06-04
Add BlockEntity tick methods

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4027)

### DIFF
--- a/src/main/java/org/spongepowered/api/block/entity/BlockEntity.java
+++ b/src/main/java/org/spongepowered/api/block/entity/BlockEntity.java
@@ -68,6 +68,27 @@ public interface BlockEntity extends SerializableDataHolder.Mutable, Locatable {
     void remove();
 
     /**
+     * Returns whether this block entity can tick.
+     *
+     * @return True if this block entity can tick
+     */
+    boolean canTick();
+
+    /**
+     * Returns whether this block entity is ticking.
+     *
+     * @return True if this block entity is ticking
+     */
+    boolean isTicking();
+
+    /**
+     * Sets if this block entity will naturally tick.
+     *
+     * @param ticking The ticking state
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/block/entity/BlockEntity.java
```diff
@@ -67,6 +67,27 @@ public interface BlockEntity extends SerializableDataHolder.Mutable, Locatable {
...
+     * @return True if this block entity is ticking
+     */
+    boolean isTicking();
+
+    /**
+     * Sets if this block entity will naturally tick.
+     *
+     * @param ticking The ticking state
+     */
+    void setTicking(boolean ticking);
```
gabizou: Can we just clarify in some more detail if we can actually override whether a block entity can tick? And if a block entity normally doesn't tick, that this would do nothing? Perhaps having the return value `boolean` signify the success or failure of being able to set this block entity to tick/not?

MrHell228: The point here is that if block entity `canTick()` and not `isRemoved()` then we can change it's ticking state (in impl it's simply cancelling current tick if this state is false). Impl currently throws exceptions if these checks are not passed (probably worth mentioned this in api). So which way is better: throw if dev didn't checked these conditions or have `boolean` result?

gabizou: I think having a boolean result (imo) would be ideal. It still allows passive setting, and then developers can elect to provide feedback for the failure mode.

MrHell228: Done

PR #2518, opened 2024-06-17
Update particles

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4031)
<br>

ParticleOptions:
- Removed `DIRECTION` and `POTION_EFFECT_TYPE` options due to them being legacy
- Added `OPACITY` option to support `ParticleTypes.ENTITY_EFFECT` (it uses ARGB int as color, so it's represented as `COLOR` and `OPACITY` options)
- Added `DESTINATION` option to support `ParticleTypes.VIBRATION`
- Made `ITEM_STACK_SNAPSHOT` option to be explicitly not empty because minecraft throws exception on empty itemstack anyway
- Updated options description
<br>

PositionSource:
- Exposed minecraft interface (gabizou said it's fine to expose it)
- Can be reference to `Entity` with some vertical offset or reference to `BlockPos` (it's packed into vibration particle and sent to client)
- If made of entity, vibration particles will chase entity
- Probably not the best name because it sounds very similar with our `Locatable` but I can't come up with good name (maybe something like `ResolvablePosition`, idk)
- The only method it could have is `Optional<Vector3d> position(World)` (mirrored from original `PositionSource`) but I'm not sure if it should be exposed

### DIFF
--- a/src/main/java/org/spongepowered/api/world/PositionSource.java
+++ b/src/main/java/org/spongepowered/api/world/PositionSource.java
@@ -28,6 +28,10 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.math.vector.Vector3i;
 
+/**
+ * In some particle implementations, they can "follow" a source that moves over
+ * several ticks such that a singular position is not enough information to follow.
+ */
 public interface PositionSource {
 
     /**
--- a/src/main/java/org/spongepowered/api/world/PositionSource.java
+++ b/src/main/java/org/spongepowered/api/world/PositionSource.java
@@ -1,0 +1,86 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/api/world/PositionSource.java
```diff
@@ -0,0 +1,86 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.api.world;
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.math.vector.Vector3i;
+
+public interface PositionSource {
```
gabizou, line 35: Can you document what the point of a PositionSource is? The factory functions are easy to understand, but as mentioned in discord, what the reasoning is for a PositionSource.

gabizou:
Let's target `api-12` since there's some breaking changes here (type changes and a kind removal).

src/main/java/org/spongepowered/api/world/PositionSource.java
MrHell228: How should I document it? My only thought on how to describe this is "PositionSource represents something that can have position in world" (I'm kinda bad at this :d) and maybe add  `{@link ParticleOptions#DESTINATION}` because for now it's the only usage of it

gabizou: Well, I guess the point of it is to describe what is the purpose of PositionSource over a Vector3i? 
> In some particle implementations, they can "follow" a source that moves over several ticks such that a singular position is not enough information to follow.

gabizou: The goal should be to describe to someone who is looking at this class for the first time, "What do I do with this? Why do I need this?"

MrHell228: Yeah, I understand what the goal is, I'm just not good at expressing thoughts even on my language, not to mention English :D So I added what you suggested

PR #2519, opened 2024-06-17
Update particles (api12)

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4032)
<br>

ParticleOptions:
- Removed `DIRECTION` and `POTION_EFFECT_TYPE` options due to them being legacy
- Added `OPACITY` option to support `ParticleTypes.ENTITY_EFFECT` (it uses ARGB int as color, so it's represented as `COLOR` and `OPACITY` options)
- Added `DESTINATION` option to support `ParticleTypes.VIBRATION`
- Made `ITEM_STACK_SNAPSHOT` option to be explicitly not empty because minecraft throws exception on empty itemstack anyway
- Updated options description
<br>

PositionSource:
- Exposed minecraft interface (gabizou said it's fine to expose it)
- Can be reference to `Entity` with some vertical offset or reference to `BlockPos` (it's packed into vibration particle and sent to client)
- If made of entity, vibration particles will chase entity
- Probably not the best name because it sounds very similar with our `Locatable` but I can't come up with good name (maybe something like `ResolvablePosition`, idk)
- The only method it could have is `Optional<Vector3d> position(World)` (mirrored from original `PositionSource`) but I'm not sure if it should be exposed

### DIFF
--- a/src/main/java/org/spongepowered/api/world/PositionSource.java
+++ b/src/main/java/org/spongepowered/api/world/PositionSource.java
@@ -28,6 +28,10 @@ import org.spongepowered.api.Sponge;
 import org.spongepowered.api.entity.Entity;
 import org.spongepowered.math.vector.Vector3i;
 
+/**
+ * In some particle implementations, they can "follow" a source that moves over
+ * several ticks such that a singular position is not enough information to follow.
+ */
 public interface PositionSource {
 
     /**
--- a/src/main/java/org/spongepowered/api/world/PositionSource.java
+++ b/src/main/java/org/spongepowered/api/world/PositionSource.java
@@ -1,0 +1,86 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


MrHell228:
Moved to api12 (maybe it could be done on previous PR but I don't know if it's possible)

src/main/java/org/spongepowered/api/world/PositionSource.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.math.vector.Vector3i;
+
+/**
+ * In some particle implementations, they can "follow" a source that moves over
+ * several ticks such that a singular position is not enough information to follow.
+ */
+public interface PositionSource {
```
Faithcaio, line 35: Have not looked at impl. yet, but would it be possible to have a version with Provider<Vectro3i>? That way a plugin could maybe create a PositionSource that can be moved without an entity.

Faithcaio (approved):

src/main/java/org/spongepowered/api/world/PositionSource.java
MrHell228: It's not possible. PositionSource is serialized by codec and sent to client. So abstract Provider<Vector3i> can not go the same way

Issue #2520, opened 2024-06-18
Structure and Feature placement callbacks


octylFractal:
### Major SpongeAPI version

12

### Is this likely to be a breaking change?

No

### What are you requesting?

Currently, `Structure` and `Feature` have `boolean place(ServerWorld, Vector3i)`, which does not allow a mod like WorldEdit to capture or modify the changes. It would be nice if we could get a callback for each block/entity so that WorldEdit may place it, rather than placing it directly into the world. This allows undo, masking, and better control of block placement than using Minecraft's code directly.

Yeregorix:
I think the proper way to do it would be:
```java
try (CauseStackManager.StackFrame frame = Sponge.server().causeStackManager().pushCauseFrame()) {
  frame.addContext(MY_CONTEXT_KEY, "hello");
  Features.TREES_PLAINS.get().place(location);
}
```
and
```java
@Listener
public void onBlockChange(ChangeBlockEvent e) {
  if (e.context().get(MY_CONTEXT_KEY).isPresent()) {
    // filter blocks
  }
}
```
However I tested that a few days ago and it didn't work so there is likely a bug in the implementation currently.
The implementation needs to be fixed, but I don't think the API needs any additions for that.

aromaa:
I have fixed the implementation for the suggested and recommended approach mentioned above in the commit https://github.com/SpongePowered/Sponge/commit/49fe76a0fb90c1365e90879285f1ed8a2e88c878. This should unblock any scenarios where you would want to modify or record the block changes.

I will leave this open however because we most likely would like to still implement some improvements around this where you only want to record the outcome. Something along the lines of https://github.com/SpongePowered/SpongeAPI/commit/63edbff7320403dfd001f34c3604d61b775a6b80#diff-18baa7adfaf0c306829fb40fb9aeccd21a7885830d19064a07504aaf3d0d361fR106. We have introduced `CompositeEvent` in API 14 which seems to be the perfect candidate here.

PR #2564, opened 2024-10-05
BlockPlacerCallback

nmacl:
Created a callback method to allow plugins to intercept the place method on Feature & Structure

https://github.com/SpongePowered/SpongeAPI/issues/2520

### DIFF
--- a/src/main/java/org/spongepowered/api/world/generation/feature/Feature.java
+++ b/src/main/java/org/spongepowered/api/world/generation/feature/Feature.java
@@ -55,22 +55,35 @@ public interface Feature extends DefaultedRegistryValue {
     DataView toContainer();
 
     /**
-     * Places the feature at given position and world
-     *
-     * @param world The world
-     * @param pos The position
-     *
-     * @return true when the feature was successfully placed
-     */
-    boolean place(ServerWorld world, Vector3i pos);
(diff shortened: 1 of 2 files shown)


Yeregorix:
I'm skeptical about using a callback instead of the blockchange event.
Also, the code is duplicated and doesn't even build.

avaruus1:
> I'm skeptical about using a callback instead of the blockchange event. Also, the code is duplicated and doesn't even build.

Seems that there's also no PR for the implementation.

src/main/java/org/spongepowered/api/world/generation/feature/Feature.java
```diff
@@ -55,22 +55,35 @@ public interface Feature extends DefaultedRegistryValue {
...
 
     /**
-     * Places the feature at given position and world
-     *
-     * @param world The world
-     * @param pos The position
-     *
-     * @return true when the feature was successfully placed
-     */
-    boolean place(ServerWorld world, Vector3i pos);
```
gabizou, line 65: I don't see a reason why this couldn't remain, if a callback can be optionally an overloaded method.

Yeregorix:
As I stated on the issue, the API already has everything needed, the real work is on the implementation.

PR #2566, opened 2024-10-20
Rewrite attack and damage events

Yeregorix:
[Sponge](https://github.com/SpongePowered/Sponge/pull/4138) | SpongeAPI

This PR attempts to solve long standing issues with the attack and damage events. The previous API was attempting to represent the damage calculation as a chain of mathematical operations which can be modified or replayed with a different base damage. This was problematic because it implies the implementation must be capable to capture all operations to be able to replay. However, due to the nature of modded platforms, it is impossible to capture everything as a mathematical operation. This results in a incorrect final damage value and thus angry pvp players.

Let's take an example: the shield modifier. Modded platforms have a shield event where mods can just set the current damage value. Was it an addition, a multiplication or any other mathematical operation? We can't know. We could just capture it as a constant, but being able to replay with a different base damage becomes useless. We could instead capture a modifier that would re-fire the event each time we replay the chain, however this would certainly break mods that have side-effects.

1. That's why we must give up the ability to replay everything. 

Another problem is that we represent the damage calculation as a chain. First of all, it is not. It has branches, early returns, splits and merges. Even if it was a chain, a mod could do some bytecode modification and add new operations in the middle of the "chain". Since we can't capture these new operations, our representation of the chain will be incomplete.

2. That's why we must give up the idea that the damage calculation is chain.

Finally, due to having a single event firing after the whole damage calculation is done, the implementation has to cancel all side-effects during the damage calculation, then replay them after the event. Again, this will not work with mod side-effects but even for vanilla this is a brittle and cumbersome approach

3. That's why it's preferable to modify the "steps" and their side-effects before they actually occur.

Note that even if it seems that these events were mostly problematic on modded platforms (indeed they were completely broken), I have also seen a few mistakes in the vanilla implementation, especially for the invulnerability time mechanic. This is because the API can hardly represent this mechanic correctly.

---

Now let's talk about the proposed solution. 

Forget about the previous concept of `DamageModifier` and let's introduce the concept of `DamageStep` which represent a step in the damage calculation that we want to expose to the API. It no longer captures the mathematical operation, but just the values before and after the step. (consequence of 1.) 

These steps are just some important key points in the damage calculation, there are not an exhaustive representation of all the things that may happen during the damage calculation. (consequence of 2.) 

Now, we split the damage event into two parts: Pre and Post.
Pre happens before the damage calculation and all its side-effects, at this time, we only know the base damage.
Post happens after the damage calculation, all steps have been captured.
During the Pre event, plugins can add some callbacks that will modify the damage value just before or just after a step. These callbacks are called `DamageModifier`. These callbacks are executed after the Pre event, in the right place during the damage calculation, ensuring that mods and side-effects will adapt to the modified value. (consequence of 3.).

As a bonus, the attack and damage events now share most of their logic in a parent class.


src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageModifier.java
```diff
@@ -24,255 +24,87 @@
...
 
     /**
-     * Gets the {@link DamageModifierType} for this {@link DamageModifier}.
+     * Gets the consumer that will modify the cause frame.
      *
-     * @return The damage modifier type
+     * @return The cause frame modifier
      */
-    DamageModifierType type();
+    Optional<Consumer<CauseStackManager.StackFrame>> frame();
```
aromaa:
```suggestion
    Optional<Consumer<CauseStackManager.StackFrame>> frameModifier();
```

src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageModifier.java
```diff
@@ -24,255 +24,87 @@
...
-     * Returns the damage modifier group.
-     * <p>Grouped modifiers calculate their damage independently from each other</p>
+     * Gets the function that will modify the damage.
+     * The function may be absent if the sole purpose of this modifier is to apply children steps.
      *
-     * @return The damage modifier group
+     * @return the damage modifier
      */
-    String group();
+    Optional<Function> damage();
```
aromaa:
```suggestion
    Optional<Function> damageFunction();
```

src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageStep.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Gets whether this step is skipped.
+     * When skipped, only the step itself and its side effects are ignored, children are still applied.
+     * A modifier willing to ignore every previous children should revert the damage to {@link #damageBeforeChildren()},
+     * or call {@link #skip} on each child.
+     *
+     * @return Whether this step is skipped
+     */
+    boolean isSkipped();
```
aromaa, line 64: Make this enum?
```java
enum SkipOptions {
    NONE,
    SELF,
    CHILDREN_ONLY,
    ALL
}
````

**Review by aromaa — APPROVED** [review]
LGTM, this iteration has much potential, good work! Added few nits.

File: src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageStep.java (hunk 3 above)
Comment (Yeregorix) [reply src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageStep.java:64]: I'm worried that setting skip option NONE on a child after setting skip option ALL on a parent might be ignored.
In the other hand, being able to skip a whole sub-tree is very handy.

If we go this way I suggest this approach :

```java
enum TreePart {
  SELF, CHILDREN_BEFORE, CHILDREN_AFTER
}
```

and then using an `EnumSet` we can describe more combinations, as well as none.

aromaa: Fair enough. Generally uncancelling is advised against because its problematic but I guess semantically its hard to explain that you might need to walk the tree upwards to figure out if the parent is cancelled.

Yeregorix: Actually, it is problematic to call setSkipped with CHILDREN_BEFORE or CHILDREN_AFTER on a parent because some children might already have been traversed and some not.

I think I will make setSkipped(boolean) fail silently when the node has already been traversed and add helper methods like setChildrenBeforeSkipped(boolean) etc that will just call setSkipped recursively. So nodes already traversed are not affected, and other nodes are skipped.

aromaa: Better approach would be that we implement it as two pass system. First pass filters out nodes and on the second pass the tree structure is immutable.

Yeregorix: Plugins may wanna know the damage to decide whether some steps should be skipped or not.

aromaa: We can also run the pass after the damage calculation has been done too. But running two passes basically solves the problem.

Yeregorix: We cannot skip a step after the damage calculation has been done. We either skip the damage calculation and the side-effects or none. For platform steps, these two are linked and cannot be separated.

aromaa: We can do it as part of the root pass, but not to the whole tree, so its not different from what you could already do.

But running the skip pass first would be the most straightforward schematically too. Want to modify the damage based on the chain? Use the modifiers.

Yeregorix: Imagine step B wanna skip the side-effects of step A when a certain condition depending on the damage value is met.
This would not be possible with the two pass approach.
I don't see what's wrong with the one pass approach, it is simpler and allows combining skip, damage, and side-effects without restrictions.

aromaa: > I don't see what's wrong with the one pass approach

You brought up that
> Actually, it is problematic to call setSkipped with CHILDREN_BEFORE or CHILDREN_AFTER on a parent because some children might already have been traversed and some not.

Which with two pass approach there would be no such problem.

Yeregorix: Yeah my plan was not to introduce a second pass, just keeping the current approach and adding helper methods

Yeregorix: I will put it on code so it's clearer, maybe my message wasn't describing my idea properly

aromaa: Generally these implicit behaviors or silent suppressions are not a great design approach. Someone is going to end up shooting their own leg so it would be better if we had an approach where that is not possible, hence the suggestion.

The order of the modifiers is also not guaranteed so there is a possibility that other modifications might result to different behavior depending on whatever there are other plugins present or not.

Yeregorix: Well, having setSkipped fail-silently was just to make things easier, I can still write helper methods without that

aromaa: If we can fail fast and document the cases where its not possible, that's okay. It would be good to document some other use cases that people might try to accomplish.

For example, they might want to emulate skip `CHILDREN_AFTER` by appending a node that skips all the following modifiers and resets the damage to the original value.

src/main/java/org/spongepowered/api/event/cause/entity/damage/DamageStep.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    /**
+     * Sets whether the children before this step are skipped.
+     * This has no effect on steps already finished.
+     *
+     * @see #isSkipped()
+     */
+    default void trySetChildrenBeforeSkipped(boolean skipped) {
```
aromaa: Do we actually gain something with these instead of just having a param with a `EnumSet`?

Yeregorix: No
Indeed, I can switch to EnumSet, that will clarify things.

PR #2568, opened 2024-10-25
feat: add armor trim key

gabizou:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4143)

### DIFF
--- a/src/main/java/org/spongepowered/api/item/recipe/smithing/ArmorTrim.java
+++ b/src/main/java/org/spongepowered/api/item/recipe/smithing/ArmorTrim.java
@@ -1,0 +1,58 @@
+/*
+ * This file is part of SpongeAPI, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/api/item/recipe/smithing/ArmorTrim.java
```diff
@@ -0,0 +1,58 @@
...
+
+    TrimPattern pattern();
+
+    interface Factory {
+
+        ArmorTrim create(TrimMaterial material, TrimPattern pattern);
+
+        default ArmorTrim create(Supplier<? extends TrimMaterial> material, Supplier<? extends TrimPattern> pattern) {
+            return create(material.get(), pattern.get());
+        }
```
MrHell228, line 47: Shouldn't these defaulted methods be in `ArmorTrim` and not in `Factory`? All convience with not calling `.get()` is gone if you need to call factory for it

gabizou: You're right, adding the static methods to ArmorTrim as well

PR #2572, opened 2024-11-07
Add key for item model

avaruus1:
[SpongeAPI | [Sponge](https://github.com/SpongePowered/Sponge/pull/4147)]

In 1.21.2 Mojang added an item component for the model of an ItemStack. This provides a more straightforward way of setting the model of an item compared to custom model data.

### DIFF
--- a/src/main/java/org/spongepowered/api/data/Keys.java
+++ b/src/main/java/org/spongepowered/api/data/Keys.java
@@ -2439,6 +2439,8 @@ public final class Keys {
 
     /**
      * Location for the model of an {@link ItemStack}
+     *
+     * @see <a href="https://minecraft.wiki/w/Data_component_format#item_model">The item_model data component</a>
      */
     public static final Key<Value<ResourceKey>> MODEL = Keys.key(ResourceKey.sponge("model"), ResourceKey.class);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -2437,6 +2437,11 @@ public final class Keys {
      */
     public static final Key<Value<Ticks>> MIN_SPAWN_DELAY = Keys.key(ResourceKey.sponge("min_spawn_delay"), Ticks.class);
 
+    /**
+     * Location for the model of an {@link ItemStack}
```
gabizou: Can we link the wikidocs for what this model means? Since it's a texture model identifier that's available in a possible resource pack, it's a good idea to make it clear here.

src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -2437,6 +2437,13 @@ public final class Keys {
...
+     *
+     * @see <a href="https://minecraft.wiki/w/Data_component_format#item_model">The item_model data component</a>
```
aromaa: While this does clarify of *what* this is actually referring to, its still quite vague of the *actual* use cases. This is also leaking the implementation detail to the documentation. I think its better to link to the actual model related documentation that lies in https://minecraft.wiki/w/Model.

PR #2573, opened 2025-01-03
Adds key to recipe

mosemister:
API | [Impl](https://github.com/SpongePowered/Sponge/pull/4156)

Adds key to recipe

### DIFF
--- a/src/main/java/org/spongepowered/api/item/recipe/Recipe.java
+++ b/src/main/java/org/spongepowered/api/item/recipe/Recipe.java
@@ -60,13 +60,18 @@ import java.util.Optional;
  */
 public interface Recipe<T extends RecipeInput> {
 
+    /**
+    * Provides the registered key for this recipe
+    *
+    * @return The key the recipe is registered with
+    */
     Optional<ResourceKey> key();
 
     /**
      * Checks if the given inventory fits the required constraints to make a valid recipe
      *
      * @param inventory The inventory to check for validity
-     * @param world     The world this recipe would be used in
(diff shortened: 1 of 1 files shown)


mosemister:
First implementation PR for Sponge so if there is any PR standards then please comment below. 

As this PR itself, I couldn't think of a situation whereby a recipe wouldn't have a key, but as I don't play modded Minecraft and this is the first time looking at the minecraft code, I played it safe with a optional

aromaa:
Recipes are part of registry so they follow the pattern where only the registry knows its key, not the item itself, so this is by design.

The problem is that we are missing recipes from `RegistryTypes` because we were not accounting for reloadable registries which the Spongepowered/Sponge#4145 is trying to resolve. Internally vanilla has layers for their registries where different layer builds upon another layer. I don't remember all the details but there were some concerns around the bootstrapping.

mosemister:
> Recipes are part of registry so they follow the pattern where only the registry knows its key, not the item itself, so this is by design.
> 
> The problem is that we are missing recipes from `RegistryTypes` because we were not accounting for reloadable registries which the [SpongePowered/Sponge#4145](https://github.com/SpongePowered/Sponge/pull/4145) is trying to resolve. Internally vanilla has layers for their registries where different layer builds upon another layer. I don't remember all the details but there were some concerns around the bootstrapping.

Thanks for the information. Ill change the implementation to just make it so you can call RegistryManager methods again without exception (the original point of me digging - just found that recipe had keys along the way)

PR #2582, opened 2025-02-26
Registry rework

aromaa:
[Sponge](https://github.com/SpongePowered/Sponge/pull/4179) | **SpongeAPI**

Notable changes:
- Template's have been removed.
  - Builders have been moved to the registry types.
  - If someone wishes, they can still serialize (if possible) the types marked as `DataPackSerializable` to a `DataView`.
- `DataPack` has been removed.
  - Replacement can be implemented in follow-up PR if desired.
- `RegisterDataPackValueEvent` removed
  - New types are registered directly to the registries.
    - Everything is purely runtime only and transient.
- Commands now operate at the server registry layer.
  - `CommandRegistrarType` takes in `RegistryHolder`.
  - `CommandTreeNodeTypes` is now only available in the server layer.
    - `Command.Raw#commandTree` takes in `RegistryHolder`.
  - `ResourceKeyedValueParameters` remains in the game layer.
    - Parameters that require registry data can be defined using `ValueParameter.Bindable`.
- `RegisterRegistryEvent` & `RegisterRegistryValueEvent` improvements.
  - Registries can now depend on other registries, allowing chaining.
  - `RegisterRegistryValueEvent` may be called multiple times in the same layer as registries are being appended or reloaded.
    - Plugins wanting to modify specific registry call the `#registry` method and the implementation consumes the consumer if appropriate.
    - Helper type `RegistryRegistrationSet<T>` avoids invalid use cases.
- Registries that have no vanilla equivalent, a Sponge one has been added in-place to iterate and add new entries.
  - Recipes
  - Advancements
- Loading of custom worlds is now explicitly required as `WorldTemplete` is gone.
  - Loading and creation of new world is made easier using the `ServerWorldProperties.LoadOptions` type.
- `AdvancementTreeEvent.GenerateLayout` now exposes the `RegistryHolder` that contains the updated registry data.


src/main/java/org/spongepowered/api/event/lifecycle/RegisterRegistryValueEvent.java
```diff
@@ -27,23 +27,93 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Lifecycle event for registering entries to a {@link Registry}.
+ *
+ * <p>This event may be called multiple times for the same layer
+ * as the registry is being appended and as dependencies become
+ * available. Plugins should only execute code inside the consumers.</p>
+ *
+ * <p><strong>Note:</strong> Layers might be reloadable!</p>
```
Zidane: This would mean that the consumers can potentially be called multiple times, correct? I think this should be further elaborated on.

src/main/java/org/spongepowered/api/event/lifecycle/RegisterRegistryValueEvent.java
```diff
@@ -27,23 +27,93 @@
...
+     *
+     * <p>The consumer will be called if it matches the current set of
+     * registries being created.</p>
+     *
+     * @param registryType The registry type to append.
+     * @param consumer The consumer to be called if found.
+     * @param dependencies The list of registries to create before
+     *                     invoking the consumer.
+     */
+    <T> void registry(RegistryType<T> registryType, BiConsumer<RegistryHolder, RegistryStep<T>> consumer, final RegistryType<?>... dependencies);
```
Zidane, line 100: Can you provide reasoning in the documentation as to why registry dependencies would be useful? Also, if I am registering 3 registries with B and C dependent on A, how do I have their types at this point?

src/main/java/org/spongepowered/api/registry/RegistryHolder.java
```diff
@@ -57,6 +57,13 @@ public interface RegistryHolder {
      */
     <T> Optional<Registry<T>> findRegistry(final RegistryType<T> type);
 
+    /**
+     * Gets a {@link Stream} of the {@link Registry registries} in this holder.
+     *
+     * @return The stream
+     */
+    Stream<Registry<?>> streamRegistries();
```
Zidane, line 65: Maybe just `registries`? Rename the below to that but has the parameter

aromaa: We seem to prefer the stream prefix, leaving this as is as other relevant files have it too. We could revisit that at some point and change them all at once.

MrHell228:
Due to datapack entries being added only at runtime now, I guess [this builder](https://github.com/SpongePowered/SpongeAPI/blob/api-14/src/main/java/org/spongepowered/api/advancement/criteria/trigger/Trigger.java#L104) shouldn't require providing a way to (de)serialize anymore. Don't know if it would be better to make it optional or completely remove though

aromaa:
The `Trigger` stuff has been runtime only-ish since its introduction. There seems to be some overlap and would need more investigation on whats happening there. It would be more ideal as follow up than trying to squeeze it here.

PR #2583, opened 2025-03-12
Make HandPreference a StringRepresentable

MrHell228:
It's StringRepresentable in impl, so this just missed in API


aromaa (approved):
Thank you :)

PR #2584, opened 2025-03-26
introduce scoped registry holder using PhaseTracker

Faithcaio:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4194)

Adds `RegistryKey#asScopedReference` 
and `RegistryType#asScopedType`
which uses the current Cause to determine an appropiate `RegistryHolder`
if none is found it falls back to `Server` if available else to `Game`

This allows early access to registries available after Game but before Server.
E.g. EnchantmentTypes during registration for recipes.


MrHell228:
Server `DefaultedRegistryType`s should probably use scoped registry holder as well

PR #2585, opened 2025-03-28
Add ItemActions

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4195)

Exposes `ConsumeEffect`s, adds`Keys.CONSUME_EFFECTS` and `Keys.DEATH_PROTECTION_EFFECTS` to apply them to `ItemStack`.

Currently `Keys.APPLICABLE_POTION_EFFECTS` does nothing and its `WeithedTable` approach doesn't make much sense anymore so I think it would be better to just delete it.


MrHell228:
Actually, ConsumeEffect might be not the best name because DEATH_PROTECTION component also uses them. Can't think of a different name though

aromaa:
More fitting names could be `ItemActionEffect` or `EntityActionEffect`. I'm not really sure would we like to actually generalize this and then Mojang throws a curveball at us but food for thought.

MrHell228:
`ItemActionEffect` sounds fine, changed. Also removed `World` arg for `ItemActionEffect#apply` since it's only used for sounds and not really excepted to differ from entity world.
I'm not sure what you mean by not generalizing this. If it's about having like `Keys<List<Double>>` and `Key<List<SoundType>>` for teleports and sounds, then there is still no guarantee that mojang won't make it, for example, single teleport and single sound effect at some point.
So I guess nothing can be considered as a totally right choice for future here.

src/main/java/org/spongepowered/api/data/type/ConsumeEffectTypes.java
```diff
@@ -0,0 +1,63 @@
...
+import org.spongepowered.api.registry.RegistryScope;
+import org.spongepowered.api.registry.RegistryScopes;
+import org.spongepowered.api.registry.RegistryTypes;
+
+/**
+ * <!-- This file is automatically generated. Any manual changes will be overwritten. -->
+ */
+@SuppressWarnings("unused")
+@RegistryScopes(scopes = RegistryScope.GAME)
+public final class ConsumeEffectTypes {
```
Yeregorix, line 41: Should be `ItemActionEffectTypes`.

src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -398,11 +399,6 @@ public final class Keys {
      */
     public static final Key<Value<Integer>> ANGER_LEVEL = Keys.key(ResourceKey.sponge("anger_level"), Integer.class);
 
-    /**
-     * The set of {@link PotionEffect}s applied on use of an {@link ItemStack}.
-     */
-    public static final Key<WeightedCollectionValue<PotionEffect>> APPLICABLE_POTION_EFFECTS = Keys.weightedKey(ResourceKey.sponge("applicable_potion_effects"), PotionEffect.class);
```
Yeregorix: Because API-14 have been released, this would be a breaking change. Rebase the PR to API-15 or keep this key for now and remove it in later versions.

Yeregorix:
`ItemActionEffect` sounds a bit redundant to me. Wouldn't `ItemAction` or `ItemEffect` be enough? I have a slight preference for `ItemAction` as it is easier to distinguish from `PotionEffect`.

MrHell228:
I don't mind renaming it to `ItemAction`. I guess it does reduce the understanding of the class purpose from the name but I believe that's what docs are for

Yeregorix:
Well, that's just my opinion, I'm not 100% sure

MrHell228:
I prefer shorter names overall though here I'm not 100% sure as well :/

MrHell228:
Renamed to `ItemAction` (new keys got renamed as well) and returned deleted key (it had no implementation anyway)

Yeregorix:
Merged by 34622032e7bda623d9eff36cd00e7d1503cb2af6

PR #2587, opened 2025-04-01
DefaultedRegistryValue, Taggable & StringRepresentable expansion

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4198)

QOL PR with big amount of lines but very small amont of actual changes.

List of changes:
- Currently `Taggable` has `registryType` method, and I think it would be nice to have it for `DefaultedRegistryValue` too to be able to retrieve registry key just from value without need to pass registry. So old `DefaultedRegistryValue` methods moved to `RegisryValue` interface (i guess it makes sense).
- `#tags` & `#is` in `Taggable` now defaulted so interface would be more friendly for plugin devs to use.
- All `DefaultedRegistryValue` and `Taggable` subtypes now explicitly implement `#registryType`.
- All vanilla registry types (except `WorldType` & `WorldArchetypeType`) implement `Taggable` because minecraft allows adding tags for them with data packs
- A lot of registry types implement `DefaultedRegistryValue` because they were missing it
- A lot of registry types implement `StringRepresentable` because they were missing it
- Added missing builtin tag groups

If override `RegistryValue` methods in `DefaultedRegistryValue`, this whole PR shouldn't cause even binary breaking change. Don't know if it's needed though.


src/main/java/org/spongepowered/api/tag/Taggable.java
```diff
@@ -24,37 +24,35 @@
...
 
-import org.spongepowered.api.registry.DefaultedRegistryType;
 import org.spongepowered.api.registry.DefaultedRegistryValue;
-import org.spongepowered.api.registry.RegistryType;
 
 import java.util.Collection;
 
 /**
- * A type that may be included in one or more {@link Tag} collections.
+ * A {@link DefaultedRegistryValue} that may be included in one or more {@link Tag} collections.
```
Zidane: This feels like implementation creep onto the API where taggables must be defaulted registry values.

src/main/java/org/spongepowered/api/registry/RegistryValue.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * A Utility marker that assists in getting a {@link ResourceKey} for values that generally can be
+ * within a {@link DefaultedRegistryType defaulted registry}.
+ */
+@SuppressWarnings("unchecked")
+public interface RegistryValue<T extends RegistryValue<T>> {
```
Zidane: Give me "What is an example of why Java needs extension methods for $200 Alex".

In all seriousness, I detest stuff like this. Since we can't fix the JDK, maybe this needs a better name. Because not all things that are RegistryValues can extend this...

src/main/java/org/spongepowered/api/tag/Taggable.java
MrHell228: ah, I forgot to revert this line, sorry. (`Taggable` interface extends `DefaultedRegsistryValue` even without this PR though)

src/main/java/org/spongepowered/api/registry/RegistryValue.java
MrHell228: Maybe something like `KnownRegistryValue` or `RegistryValueCandidate`?

And honestly, I don't know if these "old" methods are even needed with `DefaultedRegistryValue` being enhanced. I guess having 2 registries for the same type is unnecessary because the same goal can be achieveved with `Tag`s

MrHell228:
Nuked `RegistryValue` interface.
Added overloads that take `RegistryHolder` for methods in `DefaultedRegistryValue` and `Taggable`.
Changed return type of `Taggable#tags` from `Collection` to `Stream`

PR #2589, opened 2025-04-05
Add missing StringRepresentable

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4204)


Yeregorix:
PushReaction is not a StringRepresentable in MC internals. Do we want it to be StringRepresentable in our API?
The rest LGTM.

MrHell228:
Huh, I believe I looked through internals while doing this PR. I'm not sure how this happened, will remove it

PR #2592, opened 2025-04-06
Add DefaultedTag

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4206)

Changes:
- Adds `DefaultedTag` which allows to remove hardcoded registries from `Taggable` and keep convenient `Taggable#is(Tag)` (which was changed to `#is(DefaultedTag)`).
- All generated built-in tags are now `DefaultedTag`s instead of just `Tag`s.
- `Taggable#tags` requires `DefaultedRegistryType`
- Changes return type of `Taggable#tags` and `Registry#taggedValues` to `Stream`.


aromaa:
Removed some defaulted methods because their time complexity was O(n) and that's not appropriate. The implementation can do more efficient operation under the hood and I did not feel like we want to promote the fact that tags are looked up from the registry too much and just try to hide it in most cases.

 Reduces the API surface a bit too before we add too many ways to do the same thing.

aromaa:
Also mentioning that I feel like its natural that the defined tags are part of the registry itself, hence why `DefaultedTag<T>` feels right. However that the instances don't hold their tags but need to lookup it from the registry is not something we would like to promote.

PR #2594, opened 2025-04-22
Add SignalAwareVolume

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4210)

Adds volume that allows getting redstone signal.

There are 4 different ways of getting this signal from a block and each of them has a set of similar methods.
So I think the most sane way would be to add registry with those `SignalType`s.
The only exception is that analog signal doesn't support direction-based ouput so in impl passed direction is ignored.


src/main/java/org/spongepowered/api/world/volume/game/SignalAwareVolume.java
```diff
@@ -0,0 +1,322 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A volume aware of the signal that can "power" blocks. <br>
+ * This signal is often referred to as "redstone signal".
+ */
+public interface SignalAwareVolume extends PrimitiveGameVolume {
+
+    /**
+     * Returns whether the given position can conduct the signal.
```
gabizou, line 41: Is there any implicit direction being mentioned here? Does direction actually matter in this case?

PR #2596, opened 2025-05-28
Expose ArgumentReader#skip()

MrSlimeDiamond:
[Sponge](https://github.com/SpongePowered/Sponge/pull/4214) | **SpongeAPI**

### DIFF
--- a/src/main/java/org/spongepowered/api/command/parameter/ArgumentReader.java
+++ b/src/main/java/org/spongepowered/api/command/parameter/ArgumentReader.java
@@ -140,6 +140,11 @@ public interface ArgumentReader {
     interface Mutable extends ArgumentReader {
 
         /**
+         * Moves the cursor to the next character.
+         */
+        void skip();
+
+        /**
          * Moves the cursor to the next non-whitespace character. The cursor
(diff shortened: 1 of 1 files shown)


aromaa:
This should follow the convention we already use for the other skip methods, so `skipChar`. You most likely also want to couple this with `peekChar` to have this to be more usably.

I'm also not opposed to just have a method to set the cursor position directly.

MrSlimeDiamond:
It looks like there's already an `ArgumentReader#peekCharacter` method, but `Mutable`'s methods use a `char` suffix instead of `character`. (for example `parseChar`). Perhaps this should be made consistent?

I'm happy to rename it to `skipChar`, but this means I'd need to open a PR to Sponge

aromaa:
Sigh, there's also `peekString` inside the `Mutable` instead of being in the base interface which I was glancing at for other similar functions. We can't accept breaking changes in 12 so that shouldn't be done here. Let's just rename it to `skipChar` to keep it consistent with the `Mutable` inteface.

MrSlimeDiamond:
@aromaa Updated, there's now a [Sponge PR](https://github.com/SpongePowered/Sponge/pull/4214) to implement this

PR #2602, opened 2025-09-24
Add item attributes

MrHell228:
**SpongeAPI** | [Sponge](https://github.com/SpongePowered/Sponge/pull/4231)

Exposes EquipmentCondition and ItemAttribute with corresponding key.
Also adds factory-like methods for AttributeModifier creation.


src/main/java/org/spongepowered/api/data/Keys.java
```diff
@@ -2181,6 +2182,11 @@ public final class Keys {
      */
     public static final Key<Value<Boolean>> IS_WET = Keys.key(ResourceKey.sponge("is_wet"), Boolean.class);
 
+    /**
+     * The {@link ItemAttribute}s an {@link ItemStack} can apply.
```
gabizou: StackLike?

MrHell228: Oh, right, changed to ItemStackLike.
Though I looked through keys and almost every stack-related key mentions ItemStack. Probably could be another PR to update all of this

PR #2610, opened 2026-05-07
Change world border timeToTargetDiameter to ticks

abola2:
https://www.minecraft.net/en-us/article/minecraft-java-edition-1-21-11

> World Border interpolation is now dependent on game ticks instead of real time


aromaa (approved):
LGTM. Thank you

PR #2614, opened 2026-08-16
Introduce TeleportBehavior

whimxiqal:
This is to introduce teleport behavior on PortalLogic so that it has an understanding of what to do with the block that is discovered as part of the destination portal. In vanilla, this block is used to linearly interpolate the spot in the portal that matches relatively to the entry point in the entrance portal. This is the default behavior of this PortalLogic in my followup implementation PR.

### DIFF
--- a/src/main/java/org/spongepowered/api/world/portal/PortalLogic.java
+++ b/src/main/java/org/spongepowered/api/world/portal/PortalLogic.java
@@ -54,6 +54,8 @@ public interface PortalLogic {
 
     Optional<PortalFinder> finder();
 
+    Optional<TeleportBehavior> teleporter();
+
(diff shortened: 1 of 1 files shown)


whimxiqal:
I'm open to suggestions here. I've the API for the PortalLogic and its Builder to be a bit obtuse. The "composite" nature of what gets ultimately built makes it so as a consumer, I have to build a builder and then tell it to fallback to the appropriate portal logic depending on which type of portal I'm going through, even if I want to just override a single PortalLogic component. In what cases would you have more than two portals added to the rules set of `SpongePortalLogicBuilder`, potentially one for overridden functionality and the other as a fallback? And the fallback should be discoverable by the event depending on which type of portal kicked off the event.

